PR #2, opened 2015-01-20
Move to Prebuild-based buildsystem.  Compiles on VS2010.

N3X15:
# Summary:

This change removes all Visual Studio buildsystem remnants and replaces it with Prebuild.

Prebuild is capable of generating a wide range of buildscripts, including VS, nant, MonoDevelop.  Prebuild uses prebuild.xml, obviously.
# Other Changes:
- Removed a bunch of crap that VS generated for the key bindings XSD, thinking it was a DataSet.
- Renamed Reference to Third-Party.
- Added MS Diagnostics API DLL, since it's now distributable.
- All binaries are now dumped in Bin/.
# Test Plan:
- Run prebuild2010.cmd
- Compile using SpaceStation14.sln
- Run stuff in Bin/.
# Reviewers: #ss14_developers

Differential Revision: http://phab.nexisonline.net/D14

Not on phabricator because phab can't handle all these binaries.


YotaXP:
**Testing with VS 2013 Community**
1. It is making `ClientGameObject` the startup project.  This project isn't even runnable.
2. I'm not a fan of the `Client` project being renamed to `SpaceStation13`, and `Server` to `SS13_Server`.
3. Can we organize these into solution folders?
4. `EntityEditor` and `GlyphSizeTest` should be removed from the solution. (May want to keep the project folder though.)

N3X15:
The idea was to just move the project over to prebuild first.  A lot of those are design-y things.  Also, volundr's merge covered some of that already.

PR #3, opened 2015-01-20
Move to Prebuild-based buildsystem. Compiles on VS2010.  (FIXD)

volundr-:
This change removes all Visual Studio buildsystem remnants and replaces it with Prebuild.

Prebuild is capable of generating a wide range of buildscripts, including VS, nant, MonoDevelop.  Prebuild uses prebuild.xml, obviously.
# Other Changes:
- Removed a bunch of crap that VS generated for the key bindings XSD, thinking it was a DataSet.
- Renamed Reference to Third-Party.
- Added MS Diagnostics API DLL, since it's now distributable.
- All binaries are now dumped in Bin/.
- Removed reference to MiniDumper -- was crashing client when not run with debugger -- can be readded later
- Fixed Media paths to point to Resources
- Made Server/Client build to Bin/Server and Bin/Client respectively so they dont stomp on each other's shit
# Test Plan:
- Run prebuild2010.cmd
- Compile using SpaceStation14.sln
- Run stuff in Bin/.
# Reviewers: #ss14_developers

Differential Revision: http://phab.nexisonline.net/D14

Not on phabricator because phab can't handle all these binaries.


prebuild2010.cmd
```diff
@@ -0,0 +1,3 @@
+mkdir %~dp0\Bin
+call Tools\_prebuild_common.cmd
```
N3X15, line 1: You call Tools_prebuild_common.cmd but the file's actually named prebuild_common.cmd
Also, first line dupes prebuild_common.cmd functionality

N3X15:
You also forgot to nuke all the .csproj files and .sln files.

volundr-:
+1 and merge?

PR #12, opened 2015-04-07
New Tile Grid Implementation

YotaXP:
- Replaced old tile system with a new grid implementation.
- Removed atmos, as it is not compatable.
- Merged project `SS14.Shared.IoC` into `SS14.Shared`.
- Added some more helper methods to `ClientWindowData`.
- Updated prebuild files for previous organ commit.
## Screenshot

Pro spriter
## inb4 IT'S NOT OBLIQUE

The new wall rendering code is going to be very complex.  There is no point in implementing it until the SFML branch is merged in.  Deal with these ugly walls for now.


CorruptComputer:
I thought you were supposed to us plab?

Iamgoofball:
>removed atmos

best commit ever imo

YotaXP:
@CorruptComputer Plab is only being used for tasks now.  Also, thanks for reminding me.

optimumtact:
You removed all of atmos?

Thats a bit of a step backwards

edit:other than that this looks really good

SS14.Client.Services/Map/MapManager.cs
```diff
@@ -21,483 +18,299 @@ namespace SS14.Client.Services.Map
...
+
+        //    if (tileType.GetInterface("ICollidable") != null)
+        //        _collisionManager.AddCollidable((ICollidable) created);
+
+        //    return created;
+        //}
+        
+        // An actual modulus implementation, because apparently % is not modulus.  Srsly
+        // Should probably stick this in some static class.
+        [System.Diagnostics.DebuggerStepThrough]
```
optimumtact, line 309: % is the modulo operator though?

YotaXP:
@optimumtact Perhaps, but atmos needed to be reimplemented with threading in mind.  Also, I don't think the current build has a functioning atmos anyway.

SS14.Client.Services/Map/MapManager.cs
YotaXP, line 309: Yet it does not perform the modulo operation.
In C# (and most other languages even) `-5 % 4 == -1`, however [-5 MOD 4 = 3](https://www.wolframalpha.com/input/?i=-5+MOD+4).
`%` is actually a remainder operator, which is something slightly different.
I was devastated when I learned this.  I needed the actual modulo operation for this.  I would like for the function to be moved to some static class in `SS14.Shared` eventually.

volundr-:
Merged manually via 523af198d170fb017a96c97c10585c32038d3d3f

PR #14, opened 2015-04-09
Cleanup

YotaXP:
Seems like a good point to do some tidying.

This PR sorts and removes unused using statements, reformats code files, and removes some prebuild/gorgon/arcanist files that we no longer use.


ghost:
:+1:

YotaXP:
Actually, ima hold off on this for a later date.

PR #16, opened 2015-04-09
Re-enabled collision and hitbox debug overlay.

YotaXP:
It apparently just works, so let's turn it back on!


ghost:
Huh, conflict?
https://github.com/space-wizards/space-station-14/pull/15

YotaXP:
Fixed

PR #19, opened 2015-04-18
Fixed the cursor-invisible-over-window-chrome problem.

YotaXP:
**Before:**
Before Animation

**After:**
After Animation


as334:
This good to merge?

PR #20, opened 2015-04-22
Linux SFML work

harikattar:
The big one is the change to SpriteBatch, and I'd like to get more people testing that.  From what I can gather, we're lucky it happened to work on any machine, since C# handling of raw C++ arrays is nontrivial.  I think it was just because .net is 32-bit, and used 4-byte alignment, where mono is 64-bit and uses 8-byte.

C++ mandates arrays be contiguous, even if that means alignment guarantees may be violated.  C#, on the other hand, handles arrays such that every element has the same alignment, even if that produces gaps.

When feeding a C# array to C++, it gets confused after the first element, picking either null bytes (all the vertices at 0,0) or 0xffff from the color.white field.

Instead of trying to fix that, I used the SFML.net VertexArray which has already taken all of that into account.


harikattar:
I'd like a specific review on the GC impact of this change, since every time a spritebatch is updated it destroys and recreates multiple native-code VertexArrays, and I'm not comfortable signing off completely on GC that interacts across the managed boundry.

SS14.Client.Graphics/Sprite/SpriteBatch.cs
```diff
@@ -106,8 +87,8 @@ public void Draw(BaseSprite sprite)
         public unsafe void Draw(Texture texture, Vector2f position, IntRect rec, Color color, Vector2f scale,
```
YotaXP: Could remove the `unsafe` flag from this method now, and disable unsafe code in the project.

YotaXP:
Looks and works fine for me.

Ideally, SpriteBatch would implement the disposable pattern, and dispose of all the VertexArrays.  Should dispose or reuse them when the list is cleared in Begin() as well. (Don't dispose the textures, obviously.)

This should be fine for now though.  SFML implements the disposable pattern on everything that holds references to unmanaged memory, so the GC will eventually take care of it.  Making SpriteBatch disposable would just allow us to be more aggressive with freeing memory.

PR #26, opened 2015-08-11
Viewport

patrikturi:
- Fixed window resizing issues by using SFML's built-in functionality
- Please use CluwneLib.Camera:
  - SetWorldView(): render the world, you can set which part of the world will be visible
  - SetInterfaceView(): render the UI
  - ViewSize: the bounds where you have to render the UI, this won't change on window resizing
- New terminology to make things consistent:
  - Pixel Position: the pixel that was clicked physically, this is returned by the SFML MouseButton/MouseMove events
  - Window Position: coordinates in original window size (not resized), this is used by the UI
  - World Position: coordinates used by the gameplay
- Graphics/MapUtil is just a temporary solution for replacing the remaining part of ClientWindow, the problem is that Services/Map/MapManager is not static, moreover TileSize is used all around the code, but MapManager can't be seen in every project because it would cause circular dependency


ninjacha:
yus pls, merge this or I will commit shooter crime,

PR #34, opened 2015-12-25
Makes the RenderImages and RenderWindow resize properly

YotaXP:
It appears that `CluwneLib.CurrentClippingViewport` doesn't get updated.  May need to fix that later, or remove it.


elwray:
God bless you @YotaXP.

PR #37, opened 2015-12-30
Fix some of the worst perf issues.

volundr-:
SetActive does not need to be called on an offscreen target.
GaussianBlur doesn't create and dispose rendertargets every time it runs now.


volundr-:
yolo

PR #38, opened 2015-12-30
This is how you look on linux

Szunti:
More specifically it's on Arch Linux.

ss14

The options menu doesn't look right though:
options

It took me a day because I'm new to C#, mono, monodevelop and there was a couple of problems.
- UnitTesting is not working in Monodevelop. It sais "This project type is not supported by Monodevelop". Ignored it.
- [NEED HELP] Added dllmap configs for sfml.net libraries. I just put them in Third-Party/ now. I have to copy them to bin/Client after compile. I don't know how can I do it with the project file.
- Had to compile a newer version of sfml.net. I have csfml 2.3 but the included sfml.net was for an older version probably. Without the new version the mouse cursor was stucked. (Maybe this was the input problem I read about on reddit)

### DIFF
--- a/Resources/shaders/reductionEffect/HorizontalReduction.frag
+++ b/Resources/shaders/reductionEffect/HorizontalReduction.frag
@@ -1,4 +1,4 @@
-#version 130
+#version 120
 uniform sampler2D texture;
 uniform float TextureDimensions;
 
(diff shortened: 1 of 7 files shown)


volundr-:
This is great work! That actually looks much the same as it does on Windows. Please come to #spacebus on rizon IRC net and we can chat :) I will review this PR and hopefully merge today.

PR #46, opened 2016-04-21
Adds basic Lua scripting.

PJB3005:
```
PJB3005 shakes SS14 trying to wake them up!
PJB3005 shakes SS14 trying to wake them up!
PJB3005 shakes SS14 trying to wake them up!
```

Powered by NLua.

Made a new component, ScriptComponent for this purpose. It has 2 parameters:
- Code: Lua code to execute on creation.
- File: file in `SS14.Server/Scripts/` to run as Lua.

If a file is loaded more than once it will be loaded from an internal cache to prevent overhead from file I/O.

If the component receives any messages and has a function with the same name in the lua instance, that function will be called.

NLua.dll and KopiLua.dll are both under the MIT license.


SS14.Server.GameObjects/Component/Script/Script.cs
```diff
@@ -0,0 +1,89 @@
...
+		public ScriptComponent()
+		{
+			//Family = ComponentFamily.Script;
+
+			lua = new Lua();
+			lua.LoadCLRPackage();
+
+			
+			/* Would these be important?
+			RegisterSVar("Code", typeof(string));
```
PJB3005: Should I use these or not?

SS14.Shared/GO/ComponentEnums.cs
```diff
@@ -34,6 +34,7 @@ public enum ComponentFamily
         Physics,
         PlayerActions, //Holds and manages player actions (abilities).
         Renderable, // Can be rendered -- sprite or particle system
+		//Script, // Does Lua scripts.
```
PJB3005: For some reason having this here (and the `Family` thing on my `ScriptComponent`) causes clients to crash when they load in, any idea?

ike709:
keeping the dream alive :+1:  x :100:

PJB3005:
Alright just tested and no issues running on Linux.

YotaXP:
This is a good start, but it's gonna need more stuff before it really becomes useful.  Such as exposing interfaces to other components, and maybe some sandboxing.  That can come later, though.

PJB3005:
Currently it's all server side so I doubt sandboxing would be useful.

PR #48, opened 2016-04-22
Implements SFML.Window.Event.TextEvent.

PJB3005:
Uses this to fix TextBoxes.

Also fixes the console being unable to be opened, new key is Home.

KNOWN BUG: bound keys still happen, even if textboxes are focused.


YotaXP:
I hate how we propagate events, but the PR looks good nevertheless.

PJB3005:
Yeah it's pretty bad but it's very modular.

Could use some unspaghettification though.

PR #52, opened 2016-04-24
Command parsing rework.

PJB3005:
I thought having functional console commands was a good idea to code.

Added multiple interfaces and abstracts to rework client / chat / server console commands.

Chat commands and server commands are sort of integrated, client commands are not yet.

On a related note do we really not have shared interfaces?


YotaXP:
What am I doing wrong?


Also, what about client commands?  Shouldn't commands be processed only on the client?

PJB3005:
@YotaXP client console commands aren't implemented yet.

Chat console commands are somewhat, but not properly due to the `ChatManager` being ass.

For some reason the console on the client can also execute specific commands on the server so I have to sync the available ones etc... which is why it's taking a while to code.

Issue #47, opened 2016-04-22
Lighting issues.


PJB3005:
**SFML branch**

https://dl.dropboxusercontent.com/u/107028026/ShareX/2016/04/2016-04-21_21-25-32.webm

Also pictured: really poor performance.

GPU is a GTX 660m, driver version 353.62

PJB3005:
Gonna have to retest this.

ZoldorfTheWizard:
#63

PR #54, opened 2016-04-24
Fixes Lighting

ZoldorfTheWizard:
Grr Conflicts, too many of them 

Lighting works (for now) 
https://imgur.com/jsln41w

screenShadows was not being updated correctly and therefore causing that light bug where the light moved around

Currently working on a new bug: the first six lights render properly. After that, lights get rendered but screenShadows seems to not be getting updated with the new lights


ike709:
Did you forget to close #47...?

ZoldorfTheWizard:
its still open due to a bug that causes all lights after the first six to render improperly. Currently 
working on it

Issue #51, opened 2016-04-24
Starting server with empty config file causes crash.


PJB3005:
```
System.InvalidOperationException was unhandled
  HResult=-2146233079
  Message=There is an error in XML document (0, 0).
  Source=System.Xml
  StackTrace:
       at System.Xml.Serialization.XmlSerializer.Deserialize(XmlReader xmlReader, String encodingStyle, XmlDeserializationEvents events)
       at System.Xml.Serialization.XmlSerializer.Deserialize(TextReader textReader)
       at SS14.Server.Services.Configuration.ConfigurationManager.Initialize(String ConfigFileLoc) in C:\Users\Simon\Documents\SS13 Repos\space-station-14\SS14.Server.Services\Configuration\ConfigurationManager.cs:line 25
       at SS14.Server.SS14Server..ctor() in C:\Users\Simon\Documents\SS13 Repos\space-station-14\SS14.Server\SS14Server.cs:line 100
       at SS14.Server.EntryPoint.Main(String[] args) in C:\Users\Simon\Documents\SS13 Repos\space-station-14\SS14.Server\Program.cs:line 24
       at System.AppDomain._nExecuteAssembly(RuntimeAssembly assembly, String[] args)
       at System.AppDomain.ExecuteAssembly(String assemblyFile, Evidence assemblySecurity, String[] args)
       at Microsoft.VisualStudio.HostingProcess.HostProc.RunUsersAssembly()
       at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
       at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
       at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
       at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
       at System.Threading.ThreadHelper.ThreadStart()
  InnerException: 
       HResult=-2146232000
       LineNumber=0
       LinePosition=0
       Message=Root element is missing.
       Source=System.Xml
       SourceUri=""
       StackTrace:
            at System.Xml.XmlTextReaderImpl.Throw(Exception e)
            at System.Xml.XmlTextReaderImpl.ParseDocumentContent()
            at System.Xml.XmlTextReaderImpl.Read()
            at System.Xml.XmlTextReader.Read()
[...]

PR #55, opened 2016-04-25
Empty config bug fixed (Issue #51)

xecollons:
Hello,

Since this is my first commit and still don't have all the scope of the project in my head, I wanted to start with a pretty easy thing.

PersistentConfiguration and PlayerConfiguration classes have default values for their properties, so, when there is no config file, a new instance of the corresponding class will be created.

There is also a default path string in server ConfigurationManager for later saving purposes.


PJB3005:
Uh, this won't actually fix the issue reported.

I mean it's definitely an improvement but the actual issue is that the xml parser throws am exception with an invalid config file.

Issue #56, opened 2016-04-26
tilebuildoverlay sprite missing "_south" suffix


NickPowers43:
After trying to use the tile building tool an exception is thrown inside AlignTileAny.Update. Client.Services.Placement.GetDirectionalSprite(string baseSprite) is being used to get the UI sprites but the problem is the UI sprites dont have a direction. In my case the sprite that was trying to be accessed is "tilebuildoverlay_south" which should just be "tilebuildoverlay".

NickPowers43:
Should the sprite be renamed with a "_south" suffix or should I define a GetSprite function next to GetDirectionalSprite and use it instead?

NickPowers43:
I created a pull request that fixes this issue on my end.

PR #57, opened 2016-04-26
Fixed missing non-directional sprite

NickPowers43:


ike709:
Good shit if it works, but you can just put "Fixes #issuenumber" or in your case "Fixes #56" (No quotes) in the PR description and it'll auto-close the issue when this is merged rather than you having to close it yourself. Usually better to do this in case the PR isn't merged for some reason so the issue will still be there. "Fixes" can be replaced with "Closes" or "Resolves". Even more alternatives probably exist, but I don't know them all off the top of my head.

ZoldorfTheWizard:
GetSprite and GetDirectionalSprite should not be returning null as that will cause a nullpointerexecption, they should be returning the nosprite image instead.

change `return null` to `return pManager.ResourceManager.getnosprite();`  and ill merge it :D
Good work!

NickPowers43:
I removed the `return null` statements

PR #78, opened 2017-05-01
Gitignore update

d34d10cc:
Makes git ignore files under the .vs folder which Visual Studio uses.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,4 @@ _ReSharper*/
 /Media
 /setenv.bat
 /.nuget
+.vs/


d34d10cc:
Will pitch updated version.

.gitignore
```diff
@@ -55,3 +55,4 @@ _ReSharper*/
 /Media
 /setenv.bat
 /.nuget
+.vs/
```
psykzz, line 58: Can `.vs` files appear in other files and folders?

PR #79, opened 2017-05-01
Reduces warning count

d34d10cc:
What it says on the tin. Also, ignores all metafiles Visual Studio creates under the .vs folder.

Ignore the 'to 1' in the commit name, I forgot to build.


psykzz:
Any reason to use the beta runtime?

d34d10cc:
Well, it's the first one that popped up, sooo... no, not really.

I'm not even sure it's needed, but since it was added, I figured it's going to be used or had been used in the past. It compiles well both on Windows and Linux, so I don't see any issue with using the beta.

ZoldorfTheWizard (approved):

PR #84, opened 2017-05-02
Slide when colliding with Colliders

psykzz:
Currently, if you collide with a wall moving diagonally you will stop hard.
This will allow you to slide against the wall if there is the space.

### DIFF
--- a/SS14.Client.GameObjects/EntitySystems/TransformSystem.cs
+++ b/SS14.Client.GameObjects/EntitySystems/TransformSystem.cs
@@ -21,6 +21,37 @@ namespace SS14.Client.GameObjects.EntitySystems
             EntityQuery.Exclusionset.Add(typeof(SlaveMoverComponent));
         }
 
+        private Vector2f? calculateNewPosition(Entity entity, Vector2f newPosition, TransformComponent transform)
+        {
+            //Check for collision
+            var collider = entity.GetComponent<ColliderComponent>(ComponentFamily.Collider);
+            bool collided = collider.TryCollision(newPosition - transform.Position, true);
+            if (!collided)
+            {
+                return newPosition;
+            }
+
+            Vector2f newPositionX = newPosition;
+            newPositionX.X = transform.Position.X;
+            bool collidedX = collider.TryCollision(newPositionX - transform.Position, true);
+            if (!collidedX)
+            {
+                return newPositionX;
+            }
+
+            Vector2f newPositionY = newPosition;
+            newPositionY.Y = transform.Position.Y;
+            bool collidedY = collider.TryCollision(newPositionY - transform.Position, true);
+            if (!collidedY)
+            {
+                return newPositionY;
+            }
+
+            return null;
+
+
+        }
+
         public override void Update(float frametime)
         {
             var entities = EntityManager.GetEntities(EntityQuery);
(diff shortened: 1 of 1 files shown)


SS14.Client.GameObjects/EntitySystems/TransformSystem.cs
```diff
@@ -21,6 +21,37 @@ public TransformSystem(EntityManager em, EntitySystemManager esm)
             EntityQuery.Exclusionset.Add(typeof(SlaveMoverComponent));
         }
 
+        private Vector2f? calculateNewPosition(Entity entity, Vector2f newPosition, TransformComponent transform)
```
ComicIronic, line 24: Correct me if I'm wrong: the `?` makes this nullable?

psykzz: That's right

PJB3005 (approved):

d34d10cc (approved):

SS14.Client.GameObjects/EntitySystems/TransformSystem.cs
```diff
@@ -101,16 +132,12 @@ public override void Update(float frametime)
...
-                            var collider = entity.GetComponent<ColliderComponent>(ComponentFamily.Collider);
-                            bool collided = collider.TryCollision(newPosition - transform.Position, true);
-                            if (!collided)
-                                doTranslate = true;
-                            else
+                            Vector2f? _newPosition = calculateNewPosition(entity, newPosition, transform);
+                            if (_newPosition != null)
                             {
-                                //Debugger.Break();
+                                newPosition = (Vector2f) _newPosition;
```
ComicIronic: I'm surprised casting works, but maybe `_newPosition.Value` is better - it removes the runtime check.

SS14.Client.GameObjects/EntitySystems/TransformSystem.cs
RemieRichards: Why have you made it nullable? that's not required to return null, and you're not using anything specific to nullables, just checking ```!= null``` when you use the value.

also @ComicIronic yes, ```T?``` is sugar for ```Nullable<T>```

PJB3005: @RemieRichards it is for structs I believe.

PJB3005: afbeelding

yep

RemieRichards: interesting that classes are nullable but structs aren't. guess it's the whole "struct smells like an object but is actually a value" thing.

ComicIronic: It's also a syntax error to make an object type nullable, I believe.

SS14.Client.GameObjects/EntitySystems/TransformSystem.cs
psykzz: Im good with that change, I don't have an environment here at work, so i'll make the change at home later.

SS14.Client.GameObjects/EntitySystems/TransformSystem.cs
PJB3005: Structs are value types so they're presented in memory literally in the same way as things like integers IIRC, whereas classes are references types and managed by references. It makes sense.

RemieRichards: @PJB3005 
>guess it's the whole "struct smells like an object but is actually a value" thing.

but yeah.

psykzz:
Ready for merge.

ComicIronic (approved):

PR #88, opened 2017-05-03
Removes Interaction System Exception

clusterfack:
This code would crash servers testing out interaction code whenever they click on an empty tile needlessly. This replaces the code with a simple return 1.

### DIFF
--- a/SS14.Server.GameObjects/EntitySystems/InteractionSystem.cs
+++ b/SS14.Server.GameObjects/EntitySystems/InteractionSystem.cs
@@ -112,6 +112,7 @@ namespace SS14.Server.GameObjects.EntitySystems
 
         private bool DoEmptyHandToActorInteraction(Entity user, Entity obj)
         {
+            // TODO Implementation for this
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


PJB3005:
Should maybe make a log warning that it's unimplemented.

CrAzYPiLoT-SS13:
This still feels like a hacky fix. The exception is there for the very reason it's named after: to notify you something is unimplemented. I don't see a reason for it not to throw an exception until it is properly implemented. 

That said, it may be approved in this case since accidental clicking and all that, and it's fairly obvious if it doesn't work, but this shouldn't become a trend IMO.

clusterfack:
Its the epitome of stupidity that clicking on nothing, which at this point should have no behavior anyway, would cause a server crash. You can interact with certain objects just fine.

PJB3005:
Yeah it shouldn't happen, but I think a warning (with line number) is still in order.

ComicIronic:
`LogManager.Log(String.Format("{0} doesn't implement (whatever the method name is.)", this.GetType()));`

ComicIronic:
Wait, why didn't you just add the logging?

PR #89, opened 2017-05-03
Signal handlers, clear terminal title on shutdown.

PJB3005:
The signal handling code only gets compiled under Mono. The `Console.CancelKeyPress` code always does though.


SS14.Server.Services/SS14Server.cs
```diff
@@ -47,6 +47,8 @@ public void Restart()
             instance.Restart();
         }
 
+        public void Shutdown(string reason=null) => instance.Shutdown(reason);
```
ComicIronic: Don't break consistency, make this a full method body.

psykzz:
Why does the title need to be cleared?

PJB3005:
@psykzz if we don't clear it explicitly, it stays after the process exits:

afbeelding

See here my terminal tab title *and* window title (due to being the selected tab) being the title from the non-running server.

psykzz:
Makes sense, thanks.

psykzz:
@PJB3005 conflicts

SS14.Server/Signals.cs
```diff
@@ -0,0 +1,44 @@
...
+            UnixSignal[] signals = new UnixSignal[] {
+                new UnixSignal (Mono.Unix.Native.Signum.SIGTERM),
+                new UnixSignal (Mono.Unix.Native.Signum.SIGINT)
+            };
+
+            SignalThread = new Thread(() =>
+            {
+                while (true)
+                {
+                    Console.WriteLine("hrm?");
```
PJB3005: Wait shit I forgot to remove this

ZoldorfTheWizard:
@PJB3005  SS14.Server.Services/ServerConsole/ConsoleManager.cs conflicts

PR #91, opened 2017-05-03
Basic command line argument handling for the server.

PJB3005:
Also removes the NuGet packages from the repo.


d34d10cc:
You monster.

SS14.Server/Program.cs
```diff
@@ -37,17 +34,14 @@ private static void Main(string[] args)
...
-                switch(args[i])
-                {
-                    case "--fulldump":
-                        fullDump = true;
-                        break;
-                }
-            }
+            var options = new CommandLineArgs();
+            bool result = CommandLine.Parser.Default.ParseArguments(args, options);
+            if (!result)
```
ComicIronic: Yech. Blech. Why.

ComicIronic:
You a real one. Let's see this break something package-wise.

PJB3005:
People might have to restore packages but that should be default in Visual Studio IIRC. Even if it isn't it's a simple context menu option I'm pretty sure.

PR #101, opened 2017-05-05
Python-based, cross platform buildResourcePack.py

PJB3005:
Alright we're good here.


PJB3005:
So everything works, but for some reason the game code can't read the zip file and sees it as completely empty.

The hell.

Resources/buildResourcePack.py
```diff
@@ -0,0 +1,232 @@
...
+
+    # For debugging on non-Windows systems.
+    # You're probably gonna want to pass --sprite-renderer SpriteDummy too.
+    parser.add_argument("--force-animations-build",
+                        dest="force_animations",
+                        action="store_true",
+                        help=argparse.SUPPRESS)
+    
+
+    args = parser.parse_args()
```
psykzz, line 94: After parsing the args, can you just pass them into another function that actually does the work.

I know you are doing more arg parsing below, but it looks more like sanity check and cross platform stuff.
Can you move the following code into its own function instead.

psykzz:
Can you add a `requirements.txt` to list all the dependencies you'll need to install?

`pip freeze > requirements.txt`

PJB3005:
@psykzz onlyuses first party modules in Python 3.5. If `colorama` is installed it'll output colored text but that's completely optional.

Resources/buildResourcePack.py
PJB3005: Well I don't really see the need to move it to its own function personally.

ComicIronic:
PIL is first-party?

PJB3005:
@ComicIronic the sprite dummy is for debugging the script. It doesn't get ran unless you set the sprite renderer executable to it.

psykzz:
And yet, if someone did do that it wouldn't work, because they don't have PIL.
Either remove the DummyScript or add its dependencies.

What harm is a requirements.txt file anyway?

PJB3005:
Alright I'll put a requirements.txt next to the dummy.

Resources/buildResourcePack.py
```diff
@@ -0,0 +1,270 @@
...
+    # You're probably gonna want to pass --sprite-renderer SpriteDummy/DummySpriteRenderer.py too.
+    parser.add_argument("--force-animations-build",
+                        dest="force_animations",
+                        action="store_true",
+                        help=argparse.SUPPRESS)
+    
+
+    args = parser.parse_args()
+
+    animations = not args.animations
```
ComicIronic: This is just confusing - why not make the parser values of the negative form i.e. `animations = not args.no_animations`?

PJB3005: Good idea.

PJB3005:
Should be done. Works on Windows. Requires Python 3.5.

Resources/buildResourcePack.py
```diff
@@ -0,0 +1,273 @@
...
+
+    if not path.endswith("/"):
+        path += "/"
+
+    try:
+        info = file.getinfo(path)
+
+    except:
+        info = zipfile.ZipInfo()
+        info.date_time = (2001, 9, 11, 8, 46, 0)
```
d34d10cc, line 260: M E M E S

ZoldorfTheWizard:
Approved

[Approved with PullApprove](https://pullapprove.com/space-wizards/space-station-14/pull-request/101/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=space-wizards/space-station-14)

PR #102, opened 2017-05-05
Fix diagonal movement being slower than normal

### LINKED ISSUE #96: You move slower when diagonally moving against a wall.
When your direction is corrected from diagonally moving into a wall the speed is slower than if you just walk in one direction manually.

psykzz:
Fixes #96


ZoldorfTheWizard:
approved

Issue #67, opened 2016-08-02
TODO: GUI Fixes 


ZoldorfTheWizard:
Client 
Screen does not resize.
Scrollbars are rendering black
Spawn Entity menu renders the background of the items black

psykzz:
Should all be fixed.

PR #119, opened 2017-05-07
Fix entity spawnlist background

psykzz:
Fixes the issue in #67 
> Spawn Entity menu renders the background of the items black
image


SS14.Client.Services/UserInterface/Components/EntitySpawnSelectButton.cs
```diff
@@ -91,8 +91,7 @@ public override void Render()
         {
            CluwneLib.drawRectangle(ClientArea.Left, ClientArea.Top, ClientArea.Width, ClientArea.Height,
                                                        selected ? new SFML.Graphics.Color(34, 139, 34) : new SFML.Graphics.Color(255, 250, 240));
-           CluwneLib.drawRectangle(ClientArea.Left, ClientArea.Top, ClientArea.Width, ClientArea.Height,
```
ZoldorfTheWizard, line 94: wow the debug rect was causing this? good catch

ZoldorfTheWizard (approved):

PR #124, opened 2017-05-07
Cleanup chat rendering

psykzz:
Cleans up the chat logging in console (set to debug, otherwise its going to be way too much logging).
Also changed the font as the pixel count for word wrapping assumes a different font and was causing issues.

Also fixed incorrectly set vertical scrollbar on chatboxes.

image

image


PJB3005:
In reality the pixel count shouldn't be hard coded.

psykzz:
I agree, but we don't have a better solution just yet.

PJB3005 (approved):

psykzz:
Added the fix for command output

image

PR #128, opened 2017-05-07
Disable keybindings when GUI has focus

### LINKED ISSUE #104: GUIObject with focus does not stop background keybindings
When you are searching within a text box we do not handle focus correctly.

psykzz:
Fixes #104 
Currently has a small bug, where if you close the MenuWindow with the [x] in the top right, focus is not restored to keybindings.

You can press ESC to clear it and keep moving, but this isn't ideal.

Video for demonstration - https://www.youtube.com/watch?v=bMHi0X2gvQ0


BobdaBiscuit:
I don't think disabling movement with menus open is a good gameplay decision from all standpoints.

Is there any way to make this only happen when a text box is in focus?

psykzz:
Updated to now block chat, and also fixed bug of keybindings not enabling again.
Note: Focus can be set on disposing elements, call RemoveFocus after Dispose

ComicIronic:
I have some things to say about how this is implemented, but my main issue is with the type-checking. It's totally inappropriate in an OO language.

SS14.Client.Services/State/States/GameScreen.cs
```diff
@@ -373,6 +374,9 @@ private void InitializeGUI()
             _menuButton.Update(0);
             _menuButton.Clicked += menuButton_Clicked;
             UserInterfaceManager.AddComponent(_menuButton);
+
+            // Ensure nothing has focus
+            UserInterfaceManager.RemoveFocus();
```
ComicIronic: Why would anything have focus at this point anyways? Does the UserInterfaceManager give new components focus?

ZoldorfTheWizard: Everything in userinterfacemanager starts with no focus so its unnecessary in my opinion

psykzz: It was more a precaution and a way to ensure state.

ComicIronic: It adds more side-effects to the method and mixes up responsibility. I feel like this could cause confusion in the future.

psykzz:
@ComicIronic how would you do the type-checking? Do you have an example?

ComicIronic:
Do it polymorphically - either extend the current interface or make a new one, and give it `EnableKeyBindings { get; }`, and set it based on that.

psykzz:
@ComicIronic something like this?
https://github.com/space-wizards/space-station-14/pull/128/files#diff-e4cabca85f5f0808511b024499fd6d8eL108

SS14.Client.Services/UserInterface/UserInterfaceManager.cs
```diff
@@ -378,6 +384,34 @@ public virtual bool KeyDown(KeyEventArgs e)
...
+                if (hasMenuWindow > 0)
+                {
+                    DisposeAllComponents<MenuWindow>(); //Remove old ones.
+                    RemoveFocus();
+                    return true;
+                }
+                if (_currentFocus != null)
+                {
+                    // Handle special case where we are mid-typing.
+                    if (_currentFocus.GetType() == typeof(Chatbox)
```
ComicIronic: You still have this, which is what I have a problem with. The UI manager shouldn't have to care about special rules for each part of the UI.

psykzz:
Im going to work on completing the rest of this PR then later tonight.

psykzz:
This should be ready for merge now

psykzz:
Conflicts resolved for now

PR #129, opened 2017-05-08
Added SonarQube to Travis build

ComicIronic:
SonarQube is a linter which automatically analyses projects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: csharp
 
+# Beta distribution for running SonarQube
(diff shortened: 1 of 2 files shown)


ComicIronic:
The result of these changes will be that new build analyses should go to https://sonarqube.com/organizations/space-wizards/projects . It might end up being worthwhile to make a dev branch to bundle changes upstream for checking, but that would be premature right now.

ZoldorfTheWizard (approved):

PR #130, opened 2017-05-08
Fixes exception with map saving

clusterfack:
This exception would occur due to the map saving system (F5) taking the current name of the server and saving over the current server's map. This would cause an accessdenied exception due to the map being in use while the server is running presumably. 

This fixes that problem and adds a timestamp added when saving maps using the debug verb (to prevent any overwriting of maps)

### DIFF
--- a/SS14.Server.Services/Map/MapManager.cs
+++ b/SS14.Server.Services/Map/MapManager.cs
@@ -13,6 +13,8 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
+
 
 namespace SS14.Server.Services.Map
 {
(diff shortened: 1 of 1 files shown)


SS14.Server.Services/Map/MapManager.cs
```diff
@@ -316,6 +318,12 @@ public void NetworkUpdateTile(TileRef tile)
 
         public void SaveMap(string mapName)
         {
+            mapName = Regex.Replace(mapName, @"-\d\d-\d\d_\d\d-\d\d-\d\d", ""); //Strip timestamp, same format as below
+            DateTime date = DateTime.Now;
+            mapName = string.Concat(mapName, date.ToString("-MM-dd_HH-mm-ss")); //Add timestamp, same format as above
+
+            LogManager.Log(string.Format("We are attempting to save map with map named {0}", mapName));
```
RemieRichards: "save map with map"?
How about, "save map as {0}"?

SS14.Server.Services/Map/MapManager.cs
```diff
@@ -316,6 +318,12 @@ public void NetworkUpdateTile(TileRef tile)
 
         public void SaveMap(string mapName)
         {
+            mapName = Regex.Replace(mapName, @"-\d\d-\d\d_\d\d-\d\d-\d\d", ""); //Strip timestamp, same format as below
```
ComicIronic, line 321: Why does the mapName being passed in have a timestamp on it in the first place? I feel like this is a hack fix for a logic error.

PJB3005: @ComicIronic I believe the reason is if you load a map file with time stamp in the filename, so it doesn't get a double time stamp.

clusterfack: Yeah, it's so you swap to a version you are editing (which logically would have a timestamp) and save it, without it getting multiple.

clusterfack:
There was another exception too, that I accidentally fixed and thought it was related to the first one fixed by this PR so I removed it, but it comes from not having the folder maps are saved into. This properly builds the directory so the exception is fixed.

PR #138, opened 2017-05-09
Fix string.Format errors that would cause exceptions.

gbasood:
String format in C# does not take empty curly braces.
Also someone broke appveyor


ComicIronic:
Yeah, I broke Appveyor. Still works, though.

PR #139, opened 2017-05-09
remove an unnecessary, probably remnant calculation

gbasood:
`x - x = 0`


ComicIronic:
This actually looks like an error in a previous refactor. Someone replaced two different values with the same one, which is why it's `x - x`.

The second `bounds` should be `_equipBg.GetLocalBounds()` instead, as seen in https://github.com/gbasood/space-station-14/commit/653e2deab21da267e057c515323ab2c69051a37c

ComicIronic (changes requested):
Actually fix the maths.

gbasood:
This entire file is a mess, by the way. New vectors are defined over and over and over again when they don't really need to be.

ZoldorfTheWizard (approved):

PR #151, opened 2017-05-10
Refactor IoC.

PJB3005:
Is now attribute driven and more compile time enforced (with interfaces). IoC can now resolve types (and create them with a helper to reduce boilerplate).

Should allow for better modularization and stuff later down the line.


PJB3005:
@ComicIronic Should I make `[IoCTarget]` required or optional? Right now it's optional.

SS14.Client.Services/Configuration/ConfigurationManager.cs
```diff
@@ -2,9 +2,11 @@
 using SFML.Window;
 using System.IO;
 using System.Xml.Serialization;
+using SS14.Shared.IoC;
 
 namespace SS14.Client.Services.Configuration
 {
+    [IoCTargetAttribute]
```
ComicIronic: `IoCTarget`?

ComicIronic:
Optional for what, to actually get the type? It should definitely be mandatory for that, I don't want to shim a test implementation and it gets used in the game by accident.

PJB3005:
@ComicIronic alright then.

SS14.UnitTesting/SS14.Shared/IoC/IoCManager_Test.cs
```diff
@@ -38,6 +38,30 @@ public void IoCTestAttributes()
...
+                {
+                    did1 = true;
+                }
+                else if (!did2 && type == typeof(IoCTestPriorities2))
+                {
+                    did2 = true;
+                }
+                else
+                {
+                    Assert.Fail("IoCManager returned too much types.");
```
ComicIronic: many* , minor point.

psykzz:
Has the usage changed with this PR?

If so, can you do a small example for clarity?

PJB3005:
@psykzz yes.

To allow IoC to resolve an interface, the interface needs to inherit `IIoCInterface`:
```cs
public interface IFoo : IIoCInterface { ... }
```

To allow IoC to find an implementation, said implementation needs the `[IoCTarget]` attribute:
```cs
[IoCTarget]
public class Bar : IFoo { ... }
```

psykzz:
Do you still use `IoCManager.Resolve<T>()` ?

PJB3005:
Yes.

There's also `ResolveType<T>()` which returns a type instead of an instance, `NewType<T>()` which returns a *new* instance (no services), and `ResolveEnumerable<T>()` which returns an `IEnumerable<Type>` over all implementors.

SS14.UnitTesting/SS14.Shared/IoC/IoCManager_Test.cs
ZoldorfTheWizard: code good too much types :P

PR #152, opened 2017-05-10
Build the resource pack on CI anyways.

PJB3005:
Won't be useful anyways.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: csharp
 
+dist: trusty
+sudo: false
+
 os:
 - linux
 - osx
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,11 +12,14 @@ addons:
     packages:
     - nunit-console
 
+before_install:
+  - if [ $TRAVIS_OS_NAME = osx ]; then brew update && brew install python3; fi
```
ComicIronic, line 16: I want to yell at you for not quoting this sh variable, but we have bigger problems if it ever breaks something.

ComicIronic (approved):

PR #157, opened 2017-05-11
Disconnect when quitting

### LINKED ISSUE #156: Clean up connections when a client disconnects
When you disconnect from a server, the resource is not cleaned up.

psykzz:
Closes #156 
Close connections when the client clicks `quit`.

### DIFF
--- a/SS14.Client/GameController.cs
+++ b/SS14.Client/GameController.cs
@@ -85,6 +85,7 @@ namespace SS14.Client
                 CluwneLib.RunIdle (this, _frameEvent);
                 CluwneLib.Screen.Display();
             }
+            _networkManager.Disconnect();
             CluwneLib.Terminate();
             Console.WriteLine("Gameloop terminated.");
         }


PJB3005 (approved):
Won't clear when a client gets killed incorrectly but better than nothing.

ComicIronic:
What are we going to do if the client does get killed incorrectly? Have the server try and ping it and kill the connection after a certain number of failures?

psykzz:
I was working on that last night, 
I was keeping track of messages from the client and if the client doesn't send a message within a timeout, we manually disconnect them.

The only issue im having at the moment is if you connect multiple times from the same address, the disconnect kills all the clients. So im fixing that before i do a pull request for it.

psykzz:
@ComicIronic for how i was planning to deal with connections that need to be cleaned up i pushed that to this PR - https://github.com/space-wizards/space-station-14/pull/158

psykzz:
Closing this one in favor of #158

PR #158, opened 2017-05-12
Disconnect clients that stop sending messages

psykzz:
This PR is incomplete, just putting it here to gather feedback and some direction.


PJB3005 (approved):

psykzz:
Does anyone know how in ligdren to not disconnect clients on the same IP?
Calling `Disconnect` on a `NetConnection` is causing all clients to disconnect.

PJB3005:
Shouldn't `NetConnection`s be IP + port? That sounds weird.

clusterfack:
To solve that I'd investigate how it distinguishes clients in the first place

PR #167, opened 2017-05-13
add a splash screen

### LINKED ISSUE #166: Add a "Splash Screen"

psykzz:
Fixes #166 
Adds a splash screen while loading the client.


psykzz:
We can always update the path as part of some content file but for now having a direct image is fine.

PR #189, opened 2017-05-20
Cleanup. Fixes a ton of compiler warnings.

PJB3005:


SS14.Client.Graphics/Render/RenderImage.cs
```diff
@@ -233,13 +233,12 @@ public void ResetCurrentRenderTarget()
             CluwneLib.ResetRenderTarget();
         }
 
-            
+
         /// <summary>
         /// Deconstructs and disposes this instance
         /// </summary>
-        public void Dispose()
+        protected override void Destroy(bool disposing)
```
psykzz, line 240: Are you sure a Dispose function isn't needed?

PJB3005: `Destroy` gets called by `SFML.System.ObjectBase.Dispose()`. The compiler was correctly saying that this `Dispose()` definition was hiding the one in `SFML.ObjectBase` so it wouldn't ever have been called anyways IIRC.

psykzz (approved):

PR #196, opened 2017-05-21
New Splash screen

PJB3005:
afbeelding


Uqualm:
The four looks like something an autistic child would draw it for his art project.

PJB3005:
@Uqualm 
1. We already have that as a splash screen, this just adds a background and NT logo.
2. That's good though.

psykzz (approved):

PR #199, opened 2017-05-21
Relative paths, and better logging

### LINKED ISSUE #185: Make the server and client working dir independent.
All file paths should be relative to the executable directory *unless* they're passed through environment variables or command line arguments.

psykzz:
Fixes #185 

Made use of execRelativePaths to ensure paths work based on the exe.
Potentially not relevant anymore as we restructured, but useful regardless.

Also moved logs into timestamp'd files for better handling.


SS14.Server/SS14Server.cs
```diff
@@ -89,6 +88,9 @@ public static SS14Server Singleton
 
         public SS14Server(ICommandLineArgs args)
         {
+
+            IoCManager.Resolve<ISS14Server>().SetServerInstance(this);
```
PJB3005: The need for this line should be gone since I made this `SS14Server` be the thing IoC resolves to. Probably best to cut the instance crap out of the interface, it was a terrible hack to get around the lack of IoC outside Services.

psykzz: Im good to get rid of it, it was there previously and im not sure 100% what you cut originally, will remove it.

PJB3005 (approved):

PR #201, opened 2017-05-22
Fix movement

### LINKED ISSUE #192: Saved entities being spawned at (0, 0).
All saved map entities appear to be spawning at (0, 0), the start location for the player.

### LINKED ISSUE #193: Player cannot move
Likely related to #192 , but I don't think it's due to collision with the spawned entities (if you remove all entries from `SavedEntities.xml` the player still can't move).

psykzz:
Entity systems were not being populated on server / client.

This hacks them back in via the similar method that they had before.
Ideally this can be replace at some point, but at least we have our movement back.

Fixes #193 
Fixes #192

### DIFF
--- a/SS14.Shared/GameObjects/EntitySystemManager.cs
+++ b/SS14.Shared/GameObjects/EntitySystemManager.cs
@@ -23,16 +23,14 @@ namespace SS14.Shared.GameObjects
         public EntitySystemManager(EntityManager em)
         {
             _entityManager = em;
-            List<Type> _systemTypes = new List<Type>();
-
-            _systemTypes.AddRange(
-                Assembly.GetEntryAssembly().GetTypes().Where(
-                    t => typeof(EntitySystem).IsAssignableFrom(t)));
+            List<Type> _systemTypes = Assembly.GetEntryAssembly().GetTypes().Where(
+                t => typeof(EntitySystem).IsAssignableFrom(t))
 
             foreach (Type type in _systemTypes)
             {
                 if (type == typeof(EntitySystem))
                     continue; //Don't run the base EntitySystem.
+                
                 //Force initialization of all systems
                 var instance = (EntitySystem)Activator.CreateInstance(type, _entityManager, this);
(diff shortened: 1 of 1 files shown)


SS14.Shared/GameObjects/EntitySystemManager.cs
```diff
@@ -13,6 +13,7 @@ namespace SS14.Shared.GameObjects
 {
     public class EntitySystemManager
     {
+        private readonly List<Type> _systemTypes;
```
PJB3005: This isn't used anywhere except the constructor, remove it.

SS14.Shared/GameObjects/EntitySystemManager.cs
```diff
@@ -23,6 +24,24 @@ public class EntitySystemManager
         public EntitySystemManager(EntityManager em)
         {
             _entityManager = em;
+            _systemTypes = new List<Type>();
+
+            _systemTypes.AddRange(
```
PJB3005: Do something like 
```cs
var types = Assembly.GetEntryAssembly().GetTypes().Where(t => typeof(EntitySystem).IsAssignableFrom(t));

foreach (var type in types)
{
```

PJB3005 (changes requested):

psykzz:
@PJB3005 can you review again?

SS14.Shared/GameObjects/EntitySystemManager.cs
```diff
@@ -23,6 +23,22 @@ public class EntitySystemManager
         public EntitySystemManager(EntityManager em)
         {
             _entityManager = em;
+            List<Type> _systemTypes = Assembly.GetEntryAssembly().GetTypes().Where(
```
PJB3005: You need to make this an `IEnumerable<Type>` or a `var` or it won't compile.

psykzz: This actually compiles fine as is.

PJB3005 (approved):

Issue #206, opened 2017-05-23
Client crashes when you run around


h3half:
If you load up the client in VS and monitor process memory, you'll see that it will quickly (10-15sec) reach 2GB if you're in-game running around.

![image](https://cloud.githubusercontent.com/assets/1471542/26336598/70d4335a-3f41-11e7-8e3e-52d184a83f00.png)

Image is what's shown on the client's command window before crash. Once that starts being reported, the client turns all black and never recovers. Every movement input is responded to with another line of `Failed to create an OpenGL context for this window`.

Eventually VS brings up `SpriteExt.Draw` as the source of the error, saying `System.AccessViolationException` on the one line inside `SpriteExt.Draw`.

My quick guess is something to do with player movement sprites. You don't actually have to leave the initial box area to crash the game, and it seems to happen faster if you run.

h3half:
~~@Silvertorch5 found the issue: [This code he wrote](https://github.com/space-wizards/space-station-14/blob/master/SS14.Client/State/States/GameScreen.cs#L1226) to un-fuck some SFML thing where SFML decides to vertically flip textures every even frame.~~

~~ROBUSTENGINE will fix this, someday.~~

This is probably totally wrong, as disabling lighting didn't fix the issue

Pangogie:
Visual Studio says something funky in [here](https://github.com/space-wizards/space-station-14/blob/master/SS14.Shared/GameStateDelta.cs#L45-L47) is happening.

`toStream` does not seem to be getting garbage collected.

d34d10cc:
This is related to a memory leak in the BsDiffLib library, which is a .NET wrapper of a binary file patching utility. I still want to look at it more, but this seems to be a very odd bug up to now.

PJB3005:
Jesus christ.

So apparently the way sending state to clients works is by serializing two gamestate objects, diffing their binary result, and sending that to the client.

Of course this diffing library is now shitting the bed because this a ridiculously retarded way to handle network syncing of data.

This entire thing needs a complete rewrite.

ZoldorfTheWizard:
https://github.com/space-wizards/space-station-14/blob/master/SS14.Client/State/States/GameScreen.cs#L1226 

This portion of the code is responsible for the System.AccessViolation in SFML. 

I have disabled lighting and shadows in 
https://github.com/space-wizards/space-station-14/commit/0730545fbd201596f5e4f679a5c55e3b3ea5abcf

The opengl context issue i am still investigating.

[...]

PR #209, opened 2017-05-25
Temporary fix for #206

h3half:
Disables shadows by default, toggle them with F6. This is a Bad Solution because it doesn't solve the issues but it at least stops the crashing.


ZoldorfTheWizard:
This re enables shadowcalculation but it does not display the shadows.

PJB3005:
Compile failiure @h3half

psykzz:
Shadows were removed in @PJB3005 's last big merge. So gonna close this for now, but opened an issue https://github.com/space-wizards/space-station-14/issues/216 to track progress.

PR #215, opened 2017-05-28
Clean up the options menu

psykzz:
Also add support for setting a label's fontSize.

An example of the options menu now
image

### DIFF
--- a/SS14.Client/State/States/OptionsMenu.cs
+++ b/SS14.Client/State/States/OptionsMenu.cs
@@ -99,9 +99,9 @@ namespace SS14.Client.State.States
 
         private void UpdateGUIPosition()
         {
-            var SECTION_PADDING = 50;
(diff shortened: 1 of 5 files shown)


SS14.Client/State/States/OptionsMenu.cs
```diff
@@ -44,31 +46,103 @@ public class OptionsMenu : State, IState
...
+            _btnApply = new Button("Apply Settings", ResourceManager);
+            _btnApply.Clicked += _applybtt_Clicked;
+            UserInterfaceManager.AddComponent(_btnApply);
+
+            UpdateGUIPosition();
+        }
+
+        private void UpdateGUIPosition()
+        {
+            var SECTION_PADDING = 50;
```
PJB3005: Make these `const`

PJB3005 (approved):

PR #218, opened 2017-05-28
Clicking fix

d34d10cc:
This fixes clicking. Here, fixes is a weak term. This fixes it to the extent that you can again click somewhere near objects to interact with them (such as the mop or the suit), but it doesn't fix the, e.g, issue that clicks are offset from the cursor and such.

Shoutout to @psykzz for pointing out that there was something wrong.


psykzz (approved):
Ignoring .sln

PR #222, opened 2017-06-07
Lighting de-lagging

### LINKED ISSUE #216: Fix shadows
Shadows have been turned off / removed until we can figure out what their problem is.

d34d10cc:
The client no more has extra lag when running with lights enabled, so I reenabled them, only the 'normal' amount of lag from the delta thing.

Fixes #216.

Please no more Texture.CopyToImage(), thanks.


SS14.Client/State/States/GameScreen.cs.rej
```diff
@@ -0,0 +1,10 @@
+diff a/SS14.Client/State/States/GameScreen.cs b/SS14.Client/State/States/GameScreen.cs	(rejected hunks)
```
d34d10cc: what

psykzz (changes requested):
Any chance we can move to using a class to abstract away your huge messy dictionary?

We can still use the string name as key but instead of saving an opacity map and an image, just store a class that handles all that for us.

This is almost impossible to maintain as it is.

PJB3005:
Yeah this is getting ridiculous.

psykzz (approved):

ZoldorfTheWizard (approved):
Its another hacky fix but it does fix the performance loss caused by reading the texture then immediately writing the changes

PR #224, opened 2017-06-13
GameObjects IoC refactor.

PJB3005:
# What it does

* Reworks all of `GameObjects` to use `IoC` instead of dumb hacks. `Entity` variables and alike have all been changed to `IEntity`, Things such as entity managers correctly use `IoC`. As a result we can now simply do `IoCManager.Resolve<IEntityManager>()` to get the manager, instead of `IoCManager.Resolve<IEntityManagerContainer>().EntityManager` on the client and `IoCManager.Resolve<ISS14Server>().EntityManager` on the server.

* Makes some tweaks to the entry paths for the client and server to be cleaner DI.


PJB3005:
Alright got it to start on the client.

I'm gonna do some more generic cleanup around the place now.

PJB3005:
Speedmerging this. The MacOS build'll probably be fine and I want to work on other things that depend on this.

PR #234, opened 2017-06-15
Commits the line endings that should exist but never got commited

gbasood:


gbasood:
Actually we should just do #235

PR #235, opened 2017-06-15
Fix EOLs

PJB3005:


gbasood:
As for the reasoning behind this:
On Windows, I see modified files as soon as I clone the repo. `autocrlf` changes don't help.
This is caused by our .gitattributes file *and also someone never commited new EoLs after they added that file*

PR #253, opened 2017-06-26
Unified Config

Acruid:
The client and server config managers were two duplicated classes of hard-coded configuration values. I merged these two classes together and created a new shared Configuration Manager. I also made an extendable system where any code can register, fetch, and set variables. These variables are global, and in the future can be synced between client/server. Think of how the IoC system works, but with variables.

NOTE: This adds the Nett (TOML library) dependency through nuget.

Pros of new system:
Uses a TOML file to store the server and client settings. each in their own files, with shared variables going in both.
Each module defines their own variables, and sets their default value. No master Configuration class that has to be updated, and no worrying about default values.
CVarFlags that can be added to each variable for extra functionality, like syncing between client/server, only server can change, etc. (WIP, a remade network manager is required.)
Future: Integrate CVars with existing console/commands?

Here is a list of the variables and descriptions. A few of them may not have been ported to the new system.
https://pastebin.com/cXgash6A


SS14.Shared/Configuration/ConfigurationManager.cs
```diff
@@ -0,0 +1,191 @@
...
+                    if(!cVar.Registered)
+                        continue;
+
+                    var name = kvCVar.Key;
+                    var cVarVal = cVar.Value;
+
+                    //runtime unboxing, either this or generic hell... ¯\_(ツ)_/¯
+                    if (cVarVal is int || cVarVal is Enum)
+                        tblCfg.AddValue(name, (int)cVarVal);
+                    else if (cVarVal is long)
```
PJB3005: You can use C# 7 pattern matching here:
```cs
else if (cVarVal is long val)
{
    tblCfg.AddValue(name, val);
}
```

SS14.Shared/Configuration/ConfigurationManager.cs
```diff
@@ -0,0 +1,191 @@
...
+                    throw new Exception($"[CFG] Cannot convert {tmlType}.");
+            }
+        }
+
+        /// <inheritdoc />
+        public void SaveFile()
+        {
+            if (_configFile == null)
+            {
+                Logger.Log("[CFG] Cannot save the config file, because one was never loaded.", LogLevel.Warning);
```
PJB3005: Use `Logger.Warning("")` here.

In general clean this up for all the log calls, even when `Logger.Log()` defaults to info I still think it should be `Logger.Info()`

SS14.Shared/Configuration/ConfigurationManager.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                Logger.Log($"[CFG] Cannot save the config file '{_configFile}'.\n {e.Message}", LogLevel.Warning);
+            }
+        }
+
+        /// <inheritdoc />
+        public void RegisterCVar(string name, object defaultValue, CVarFlags flags = CVarFlags.NONE)
+        {
+            ConfigVar cVar;
+            if (_configVars.TryGetValue(name, out cVar))
```
PJB3005: You can put the `ConfigVar cVar` into the `out` like this thanks to C# 7:
```cs
if (_configVars.TryGetValue(name, out ConfigVar cVar))
```

SS14.Shared/Configuration/ConfigurationManager.cs
```diff
@@ -0,0 +1,191 @@
...
+            if (_configVars.TryGetValue(name, out cVar) && cVar.Registered)
+            {
+                //TODO: Make flags work, required non-derpy net system.
+                if(cVar.DefaultValue != value)
+                {
+                    cVar.Value = value;
+                }
+            }
+
+            Logger.Log($"[CVar] Trying to set unregistered variable '{name}'", LogLevel.Fatal);
```
PJB3005: I don't think logging is appropriate here because the exception will do most of the "oh crap".

SS14.Shared/Configuration/ConfigurationManager.cs
```diff
@@ -0,0 +1,191 @@
...
+        public T GetCVar<T>(string name)
+        {
+            ConfigVar cVar;
+            if (_configVars.TryGetValue(name, out cVar) && cVar.Registered)
+            {
+                //TODO: Make flags work, required non-derpy net system.
+                return (T) (cVar.Value ?? cVar.DefaultValue);
+            }
+
+            Logger.Log($"[CVar] Trying to get unregistered variable '{name}'", LogLevel.Fatal);
```
PJB3005: Again, logging is kinda redundant.

PJB3005 (changes requested):
Minor styling nitpicks, otherwise there's one kinda big elephant in the room though.

Currently the config files look like this:

```toml
[Configuration]
key_console = 64
res_pack = "..\\..\\Resources\\ResourcePack.zip"
res_password = ""
display_width = 1280
display_height = 720
display_fullscreen = false
display_refresh = 60
display_vsync = false
net_server = "127.0.0.1"
net_updaterate = 20
net_cmdrate = 30
net_interpolation = 0.100000001490116
net_rate = 10240
net_fakelag = false
net_fakeloss = 0.0
net_fakelagmin = 0.0
net_fakelagrand = 0.0
ply_name = "Joe Genero"
ent_logging = false
```

Which isn't exactly great toml, It's look much better like this:

```toml
[input.keyboard]
console = 64

[resourcepack]
pack = "..\\..\\Resources\\ResourcePack.zip"
password = ""

[display]
width = 1280
height = 720
fullscreen = false
refresh = 60
vsync = false

[net]
server = "127.0.0.1"
updaterate = 20
cmdrate = 30
interpolation = 0.100000001490116
rate = 10240
fakelag = false
fakeloss = 0.0
fakelagmin = 0.0
fakelagrand = 0.0

[player]
name = "Joe Genero"

[debug]
ent_logging = false
```

Which would be much more sane on the config file end and actually make toml have some advantages. That said you'll probably need some extra code for fetching nested config keys. What I personally did for this in my Discord bot was making `.` separated keys so `display.vsync` would retrieve the vsync key there.

This is already a huge improvement though.

SS14.Shared/Configuration/ConfigurationManager.cs
Acruid: If anyone touches the logger API they should mark Logger.Log as Obsolete.

SS14.Shared/Configuration/ConfigurationManager.cs
Acruid: ~~After further discussion this will be declined, since the val would be defined outside the scope of the IF block, which would not simplify the code.~~

EDIT: The entire block is replaced with a Switch statement, which solves all problems.

SS14.Shared/Configuration/ConfigurationManager.cs
PJB3005: Not necessarily, since having it based on enum value is still useful when dealing with dynamic values.

PJB3005:
Toml supports nested dictionaries so I don't think doing it like that is all that clean.

PJB3005 (approved):

PR #259, opened 2017-07-05
Remove buildResourcePack* batch files.

PJB3005:
The files have been obsolete with the advent of the (faster) buildResourcePack.py.
There's no reason to use the old ones, and this sets Python 3.6 as a hard build dependency.


clusterfack:
👎 for adding more depedencies

elwray:
Agree with @clusterfack.

PJB3005:
It's this or somebody volunteering to maintain batch scripts. I sure am not gonna write batch scripts and most of the time I literally can't as a matter of fact (MacOS). Requiring that we keep supporting batch scripts will just make it harder because then I'd need to ask somebody for help every time I write a new tool or whatever for the repo, slowing down development.

It's not worth it to support it and installing Python 3 is not hard.

ZoldorfTheWizard:
+1 python

theOperand:
Having another dependence as minor as Python is much better than needing to
write all of our build scripts in multiple language

Op 7-jul.-2017 01:22 schreef "Silver" <notifications@github.com>:

+1 python

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/space-wizards/space-station-14/pull/259#issuecomment-313545576>,
or mute the thread

gbasood:
@clusterfack haha yeah dude python is so bad :^)
hey remember this: https://github.com/gbasood/vgstation13/blob/Bleeding-Edge/tools/travis/dm.sh

Jokes aside, if we're not dependent on Python, we're dependent on someone knowing and maintaining the sort of script I just linked for all possible platforms every time something about the pipeline changes. I'd rather have an easily maintainable pipeline than something built on arcane shell scripts that no one understands years from now and is too afraid to change them for fear of breaking it.

And honestly, Python is probably the nearest to universal choice for that. If you're worried about users eventually needing to use Python to play the game, though, the answer to that is what we'd probably want to do eventually, which is an automated build packaging system, either through a script and uploading results to GitHub as releases or through something external like Jenkins and a CDN. Or better yet have resource pack generation happen on startup, which we can try to get going after robust

Alternatively we could write a separate C# app for it but that's messier and gross.

SaphireLattice:
>Alternatively we could write a separate C# app for it but that's messier and gross.

You... do realize that you'll have to compile it? And if for some reason it won't work without re-co mpilation for someone, it could be VERY messy.
The `/bin/bash` script you linked is fine, most of it is just shitcode to make it run under MinGW/Windows, some arguments checking and that somewhat messy map selection thing. If you don't understand a bash script, it's either too much perl or sed magic that must be documented, or it's just poorly written in first place and you should slap idiot who wrote it.

PJB3005:
Nah, bash in general has a lot of awfulness no matter what, and equivalent Python code is 10x more readable no matter what.

Yes you can write good bash scripts, but Python is still better in most cases.

ZoldorfTheWizard:
Python has been chosen. Move along :)

PR #264, opened 2017-07-09
General Bugfixes

Acruid:
Fixed SFML Fonts throwing StreamClosed errors.
Fixed misspelled font name causing chat box to use default font.
Fixed crash when trying to quit the client from inside a round.
Fixed? backslash issue with the resource pack path.
Removed obsolete unused Player_config.xml.


Acruid:
Ok, so I'm going to leave the tests broken for tonight so we can decide how to proceed. The tests are correct for the function, but the function is not compatible with cluwnelib. In order to fix the player moonwalking bug, you need to flip the Y axis the player is moving in. This can be in the DetermineDirection() function, or something can be changed [in Cluwnelib.](https://github.com/space-wizards/space-station-14/blob/master/SS14.Client/GameObjects/Component/Renderable/SpriteComponent.cs#L225) The tickets  https://github.com/space-wizards/space-station-14/issues/144  and https://github.com/space-wizards/space-station-14/issues/178 open about problems with the animation system.

The 'proper' solution to this issue would be to have cluwnelib do the correct transforms on the direction so that it displays correctly, and leave the DetermineDirection() function alone. Right now I am presenting a quick fix, you guys can decide which path to choose.

Acruid:
https://github.com/space-wizards/space-station-14/pull/265 is fixing the path issues, and the Moonwalk bug is more than a simple fix if you want to do it properly. This PR should now be complete.

PJB3005:
Ok, nice.

PJB3005 (approved):

PR #269, opened 2017-07-10
Network system

Acruid:
There were three goals to this. The first was to combine the client and server network managers into one shared manager that can be accessed from anywhere. The second was to remove the lidgren networking code from SS14Server. The third was to encapsulate the server packets so that some form of sanitizing can be done.

Things that were not in the scope of this were changes to the client/server communication protocol, and converting the client/shared to the new network system. This is currently designed to work with minimal changes to the client, and minimal changes outside of SS14Server (now named BaseServer, maybe the content assembly could inherit from it?), SS14.Server.Networking*, and SS14.Server.Player*.

The public interfaces for the NetManager are pretty much what you would expect, check out the doc comments on them. Packets are now their own object that serializes and deserializes itself. The big benefit to this is the system can enforce a template and easily detect malformed packets.

Systems define their own packets and register them with the NetManager and a callback. Currently since Client code has not been migrated, they still rely on the static shared NetworkMessages enum. This should be removed in the future. You can currently find the existing NetMessages for the game inside SS14.Shared.Network.Messages. Yes, the protocol is a mess.

I'm not super satisfied with how this turned out but it does meet my goals, and I think it is way better than what we had before. I do believe it would be a good base to keep building off of. I have been coding this in a vacuum though, there may be something major I missed. Comments are always welcome.


SS14.Shared/Interfaces/Network/INetClientManager.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using Lidgren.Network;
+using SS14.Shared.Network;
+
+namespace SS14.Shared.Interfaces.Network
+{
+    /// <summary>
+    ///     The Client version of the INetManager.
+    /// </summary>
+    public interface INetClientManager : INetManager
```
PJB3005: Probably better to rename this to `IClientNetManager`

SS14.Shared/Interfaces/Network/INetManager.cs
```diff
@@ -0,0 +1,161 @@
...
+{
+    /// <summary>
+    ///     A network server that listens for connections, relays packets, and manages channels.
+    /// </summary>
+    public interface INetManager
+    {
+        /// <summary>
+        ///     Is this a server, or a client?
+        /// </summary>
+        bool IsServer { get; }
```
PJB3005, line 16: Using an enum here is probably a better idea. Though neither is probably best, having code need to check whether it's client or server like this seems like poor design.

SS14.Shared/Interfaces/Network/INetServerManager.cs
```diff
@@ -0,0 +1,10 @@
+﻿
+namespace SS14.Shared.Interfaces.Network
+{
+    /// <summary>
+    /// The server version of the INetManager.
+    /// </summary>
+    public interface INetServerManager : INetManager
```
PJB3005: `IServerNetManager`

SS14.Shared/Network/NetManager.cs
```diff
@@ -0,0 +1,537 @@
...
+
+        /// <summary>
+        ///     Default constructor.
+        /// </summary>
+        public NetManager()
+        {
+            _callbacks = new Dictionary<Type, ProcessMessage>();
+            _channels = new Dictionary<NetConnection, NetChannel>();
+
+            _config = IoCManager.Resolve<IConfigurationManager>();
```
PJB3005: IoC resolution from constructors like this is an implementation detail and incredibly prone to race conditions. Implement `SS14.Shared.IoC.IPostInjectInit` and do this from `PostInject()` with a hard `[Dependency]` to the manager.

SS14.Shared/Network/Messages/MsgAdmin.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        #region REQUIRED
+        public const NetMessages ID = NetMessages.RequestEntityDeletion;
+        public const MsgGroups GROUP = MsgGroups.ENTITY;
+        public static readonly string NAME = ID.ToString();
+        
+        public MsgAdmin(INetChannel channel) : base(NAME, GROUP, ID) { }
+        #endregion
+
+        public int EntityId;
```
PJB3005: Sonar is gonna kill you over these public fields, just an FYI. Make them auto implemented properties (`{ get; set; }`) and it'll be happy.

SS14.Shared/Network/NetManager.cs
```diff
@@ -0,0 +1,537 @@
...
+            {
+                config.RegisterCVar("net.server", "127.0.0.1", CVarFlags.ARCHIVE);
+                config.RegisterCVar("net.updaterate", 20, CVarFlags.ARCHIVE);
+                config.RegisterCVar("net.cmdrate", 30, CVarFlags.ARCHIVE);
+                config.RegisterCVar("net.interpolation", 0.1f, CVarFlags.ARCHIVE);
+                config.RegisterCVar("net.rate", 10240, CVarFlags.REPLICATED | CVarFlags.ARCHIVE);
+            }
+
+#if DEBUG
+            config.RegisterCVar("net.fakelag", false, CVarFlags.CHEAT);
```
PJB3005: Seems inconsistent to register part here and part in the constructors.

SS14.Shared/Network/NetManager.cs
```diff
@@ -0,0 +1,537 @@
...
+                return retVal;
+            }
+        }
+
+        /// <summary>
+        ///     Gets the NetChannel of a peer NetConnection.
+        /// </summary>
+        /// <param name="connection">The raw connection of the peer.</param>
+        /// <returns>The NetChannel of the peer.</returns>
+        private INetChannel GetChannel(NetConnection connection)
```
PJB3005, line 227: Should probably make this throw an exception if it can't be found.

SS14.Shared/Network/NetManager.cs
```diff
@@ -0,0 +1,537 @@
...
+        {
+            get
+            {
+                INetChannel retVal;
+
+                if (_netPeer.ConnectionsCount <= 0)
+                    return null;
+                try
+                {
+                    retVal = GetChannel(_netPeer.Connections[0]);
```
PJB3005: How could this throw an exception?

SS14.Shared/Network/NetMessage.cs
```diff
@@ -0,0 +1,87 @@
...
+{
+    /// <summary>
+    /// The group the message belongs to, used for statistics and packet channels.
+    /// </summary>
+    public enum MsgGroups
+    {
+        /// <summary>
+        /// Error state, the message needs to set a different one.
+        /// </summary>
+        ERROR = 0,
```
PJB3005: Are all caps names like these even standard in .NET land? Shouldn't they simply be PascalCase?

PJB3005 (commented):
Pretty good overall, definitely a good idea.

ghost:
holy fuck nice work

SS14.Shared/Interfaces/Network/INetManager.cs
Acruid: After further discussion, it was decided that these fields will stay unmodified. If they are not needed by shared code in the future, they can be removed at a later date. An enum is overkill because there are only 2 states, being a client or a server.

PJB3005 (approved):

ZoldorfTheWizard:
👍

PR #274, opened 2017-07-11
Fix server process crash when shut down.

Acruid:
Fixes Timer cleanup, so the server process does not crash when closed.


PJB3005 (approved):

ZoldorfTheWizard:
👍 pbj closing these so fast you cant even put a bangin donk on it

PR #275, opened 2017-07-12
Bsdiff replacement, submodule handling framework.

PJB3005:
Replaced the `bsdiff.NET` library we were using with a Rust-based library. Should be MUCH more transparent about memory usage (Rust has no GC and everything is allocated and deallocated inside its calls) and prevent the GC from getting overworked. Thus fixing #206, probably.

Part of this is a git hooks system for managing submodules, because `SS14.Shared.Bsdiff` is a submodule.

Don't merge this yet there's some other things I want to do.


readme.md
```diff
@@ -4,6 +4,10 @@ Space Station 14 is a revived attempt at an SS13 remake.
 
 **If you want to run the game and contribute, see our [wiki](https://github.com/space-wizards/space-station-14/wiki).**
 
+## DOWNLOADING ##
+
+Before you go to build SS14, you should run the `RUN_THIS.py` Python script using Python 3.5+. If you don't do this, it will happen on the first build, **but the first build will fail**. This is necessary to initialize Git submodules and hooks.
```
ZoldorfTheWizard, line 9: Simplify "If you don't do this, it will happen on the first build, **but the first build will fail**." to If you don't do this, **the first build will fail**.

ZoldorfTheWizard (changes requested):

PJB3005:
Ok so no matter what this literally doesn't fix it the memory leak is caused by something else entirely WHAT THE HELL.

PJB3005:
I'm closing this. When we need the submodule script, we can salvage it.

d34d10cc:
Oh, the iron E.

PR #279, opened 2017-07-20
Replaces all TestFixture constructors with OneTimeSetUp attribute

### LINKED ISSUE #245: Rework unit tests to use proper test fixture attributes.
Currently unit test fixtures are set up by using constructors on `SS14UnitTest`. They should be moved to using `[TestFixtureSetup]` and the alike inside NUnit.

Cyberboss:
Closes #245

Not sure why NUnit.Framework lines were edited, but w/e

### DIFF
--- a/SS14.UnitTesting/Client/Graphics/Shaders/TestShader_Test.cs
+++ b/SS14.UnitTesting/Client/Graphics/Shaders/TestShader_Test.cs
@@ -1,4 +1,4 @@
-using NUnit.Framework;
+﻿using NUnit.Framework;
 using SFML.Graphics;
 using SS14.Client.Graphics;
 using SS14.Client.Graphics.Render;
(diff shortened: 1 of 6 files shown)


PJB3005:
Did you miss the constructor for the base `SS14UnitTest` or was that intentional?

The changes to the `using` directives were probably the BOM being removed. My guess is your editor isn't listening to the `.editorconfig` file. Wish there was a sane way to tell VS to not use UTF-8-BOM so we could just use UTF-8 like sane people.

The rest looks good though.

Cyberboss:
I'm actually on VS 2017, oddly enough. I'll fix the base constructor as well.

PJB3005:
Actually turns out I'm dumb. The line change was *adding* a BOM, just verified by fetching your branch. We're good there then.

PJB3005 (approved):

PJB3005:
No idea why neither Travis nor AppVeyor decided to trigger, but I manually checked the PR locally and the tests pass so I'm merging this.

ZoldorfTheWizard:
:+1:

PR #280, opened 2017-07-21
Server Main Loop

Acruid:
I implemented a proper game loop for the server. I removed the nasty Win32 Timer interop, and created the GameTiming class that holds common timer and profiling info. I attempted to switch the client over to use the new class, but because of the state of the client architecture I don't even know if it works. I have SS14.Client.Graphics using its own instance of the class instead of IoC because I assume we want a separate timer for the rendering (or use the OpenTK one), and the client main loop.

The GameTiming class is kinda a WIP, I have not fully figured out how clientside interpolation is going to be done. The Pause and Timescale properties are not working, i need to do more research on how to properly implement them with relation to CurTime and RealTime. I kept them so people have something to work with in their code. If you want me to remove them say so.

I would also like someone to test this on an non-windows system to make sure everything works as expected.


PJB3005 (approved):

ZoldorfTheWizard (commented):
testing codeowners

PR #295, opened 2017-07-31
Content System + Content Assembly Loading

Acruid:
You now can dynamically load file directories and content packs into the ContentManager. Each pack you load into the manager is relative to the root directory "./". So if you load a zip file with the file ./Assemblies/Content.Server.dll in it, you use that relative location in ContentManager.GetFile(). If two packs have the same file, the file from the first pack loaded will be used. Content packs can be loaded at any time into the ContentManager. I have set it up so that the executable directory is loaded before the content pack, so people can override anything in the content pack by dropping it in the relative location in the projects bin folder.

The Content assemblies are now loaded through the Content System. Before they are loaded, their referenced types are scanned through a whitelist. I also added an option to dump all referenced types to the log so that people can see how they are filtered through the whitelist. W = whitelisted, G = greylist, B = Blacklist. If anything ends up on the black list or grey list, the assembly will be prevented from loading. Feel free to modify the lists in SS14.Shared.ContentPack.AssemblyTypeChecker.cs. Remember that this is only a front-line defense against malicious assemblies, and we should be doing other things with OS security to prevent bad things happening to clients. The last thing we need is this project to be spreading malware to players.

I also started to switch over the old content cache system in SS14.Client.Resources.ResourceCache. The new system is built into the old class, and does work, but is unfinished. Fonts were converted to the new system, you can see how they are accessed. The idea was to use something like the code side of XNA's ContentManager system. When we switch over to RobustEngine we can decide if that is a good idea or not.

The embedded resources in the SS14.Client were moved to the EngineContentPack. This pack is required for the engine to function. It should include things like UI resources and data files for the splash screen/menu system.

Another random thing: Vsync is now enabled and the Thread.Sleep() was removed.


psykzz:
Could also add support for loading resources from directory outside of the resource.pak?
That way i can work on assets without having to pack them for testing.

Also would mean we can support clientside UI mods.

PJB3005:
I don't think client side UI mods is something we should allow. It's just a gateway to cheating. Least we could do is not encourage it.

psykzz:
We can disable it in release builds if it becomes an issue, but for debug builds it should allow them.
Would mean we'd need to also allow servers to block `debug` clients

SS14.Client/ResourceManagement/BaseResource.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        /// <summary>
+        ///     Fallback resource if this one does not exist.
+        /// </summary>
+        public virtual string Fallback => null;
+
+        /// <summary>
+        ///     Disposes this resource.
+        /// </summary>
+        public virtual void Dispose()
```
PJB3005, line 20: Shouldn't you have a destructor along with this?

SS14.Client/ResourceManagement/BaseResource.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+        }
+
+        /// <summary>
+        ///     Deserializes the resource from the VFS.
+        /// </summary>
+        /// <param name="cache">ResourceCache this resource is being loaded into.</param>
+        /// <param name="path">Path of the resource relative to the root of the ResourceCache.</param>
+        /// <param name="stream">Stream of the resource that was opened.</param>
+        public abstract void Load(ResourceCache cache, string path, Stream stream);
```
PJB3005: Shouldn't this be `IResourceCache`

SS14.Client/ResourceManagement/BaseResource.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace SS14.Client.ResourceManagement
+{
+    /// <summary>
+    ///     Base resource for the cache.
+    /// </summary>
+    public abstract class BaseResource : IDisposable
+    {
+        /// <summary>
+        ///     Fallback resource if this one does not exist.
```
PJB3005: Might be a good idea to specify that this is a fallback *file path*.

SS14.Client/ResourceManagement/ResourceCache.cs
```diff
@@ -690,16 +648,135 @@ public Texture GetTexture(string key)
...
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Adds a resource to the cache.
+        /// </summary>
+        /// <param name="resource"></param>
+        /// <param name="path"></param>
+        /// <returns></returns>
+        public void CacheResource<T>(string path, T resource)
```
PJB3005, line 750: This feels like something that should be private, unless I'm misunderstanding.

SS14.Shared/ContentPack/AssemblyTypeChecker.cs
```diff
@@ -0,0 +1,147 @@
...
+            "System.Xml",
+
+            // Sockets Stuff
+            "System.Net"
+        };
+
+        /// <summary>
+        ///     Completely disables type checking, allowing everything.
+        /// </summary>
+        public static bool DisableTypeCheck { get; set; }
```
PJB3005: FYI you should probably default initialize these:
```cs
public static bool DisableTypeCheck { get; set; } = false
```

SS14.Shared/ContentPack/IContentRoot.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// </summary>
+        /// <returns>If the content was mounted properly.</returns>
+        bool Mount();
+
+        /// <summary>
+        ///     Gets a file from the content root using the relative path.
+        /// </summary>
+        /// <param name="relPath">Relative path from the root directory.</param>
+        /// <returns>A stream of the file loaded into memory.</returns>
+        MemoryStream GetFile(string relPath);
```
PJB3005, line 21: Wouldn't it make more sense to make this return a generic stream, so you can make the `DirLoader` not have to pre-allocate?

SS14.Shared/ContentPack/ResourceManager.cs
```diff
@@ -0,0 +1,121 @@
...
+        [Dependency] private readonly IConfigurationManager _config;
+
+        private readonly List<IContentRoot> _contentRoots = new List<IContentRoot>();
+
+        /// <summary>
+        ///     Default constructor.
+        /// </summary>
+        public ResourceManager()
+        {
+            _config = IoCManager.Resolve<IConfigurationManager>();
```
PJB3005: You're already using `[Dependency]`, why this?

Acruid:
It does support loading resources from any directory, sorry I was not clear on that. Right now it is adding the exe's directory, so it can load anything in the ./space-station-14/bin/Client/ folder as if it was in a resource pack.

PJB3005 (approved):

ZoldorfTheWizard (approved):

PR #298, opened 2017-08-01
Fix the esc menu

### LINKED ISSUE #296: Closing the Menu (Esc Menu) With X Button Prevents Reopening
When closing this menu by using esc multiple times, it properly toggles the visibility. However if you use the X button on the top right to close this menu, it gets deleted and will never reopen itself.

clusterfack:
The esc menu would previously dispose of itself if you used the X button, however this menu is meant simply to toggle invisbility on and off when you press X or esc.
Fixes #296

### DIFF
--- a/SS14.Client/UserInterface/Components/MenuWindow.cs
+++ b/SS14.Client/UserInterface/Components/MenuWindow.cs
@@ -73,25 +73,21 @@ namespace SS14.Client.UserInterface.Components
             //Create a new one.
             Dispose();
         }
+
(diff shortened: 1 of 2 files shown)


psykzz:
Does this properly handle focus?
What if it had a text box would it receive the inputs?

clusterfack:
The changes I made are only for the explicit menu you get when you press esc, it has no text box. The helper proc togglevisibility() seems to handle focus correctly on its own.

PR #299, opened 2017-08-02
Upgrade and Include OpenTK in all projects.

Acruid:
Changes SS14.Client.Graphics to use OpenTK through NuGet.
Upgrades OpenTK 1.6 to 3.0pre in SS14.Client.Graphics project.
Adds OpenTK to all projects so that the math library is available.
Adds helper functions to SFML and OpenTK vectors to convert between them for better compatibility.


SS14.Client.Graphics/SS14.Client.Graphics.csproj
```diff
@@ -160,6 +159,8 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="app.config" />
+    <None Include="OpenTK.dll.config" />
```
PJB3005: Surely this should be copied to output, no?

PJB3005 (approved):

PJB3005:
Merging this since I need it to fix https://github.com/space-wizards/space-station-14-content/issues/3

PR #309, opened 2017-08-05
Client Timing

Acruid:
Adds a main loop to the client, just like the server. I also solved some TODO issues in the GameTiming class. Fixes some sonar issues. Should fix https://github.com/space-wizards/space-station-14/issues/301.

Because the TransformComponent prediction is bad, this PR will break it even more, and cause your player to rubber band a bit. The code reviewers can decide if they want to hold off merging this until I have fixed the TransformComponent or not.


PJB3005:
> The code reviewers can decide if they want to hold off merging this until I have fixed the TransformComponent or not.

Nah I don't care I'm 👍 here.

SS14.Client/GameController.cs
```diff
@@ -159,6 +213,49 @@ public void Run()
...
+
+        /// <summary>
+        /// Runs a tick of the simulation.
+        /// </summary>
+        /// <param name="e">Current GameTiming.FrameTime</param>
+        private void Update(FrameEventArgs e)
+        {
+            _networkManager.ProcessPackets();
+            CluwneLib.RunIdle(this, e);
+            _stateManager.Update(e);
```
psykzz, line 233: Shouldn't state be handled first last item being render things to screen? Otherwise screen rendered, state changes is bad.

PJB3005 (approved):

SS14.Client/GameController.cs
Acruid: CluwneLib.RunIdle does not render the scene. It used to update state, i moved that code into the Update function. Nothing is even using that event now.

psykzz: Should we not just remove it then?

PR #312, opened 2017-08-07
Deprecate resource pack temporarily.

PJB3005:
Temporarily being until we actually need content downloading from server.
Resources now get copied to the output directories directly.


clusterfack:
Perhaps make this a debug/release setting?

PJB3005:
Nah, we're not gonna have a need for ZIP resource packs for a long time.

PR #315, opened 2017-08-08
Fix SFML memory leak.

### LINKED ISSUE #206: Client crashes when you run around

PJB3005:
This was my fault.

Fixes #206 probably.

### DIFF
--- a/SS14.Client.Graphics/Render/RenderImage.cs
+++ b/SS14.Client.Graphics/Render/RenderImage.cs
@@ -233,15 +233,6 @@ namespace SS14.Client.Graphics.Render
             CluwneLib.ResetRenderTarget();
         }
 
-
-        /// <summary>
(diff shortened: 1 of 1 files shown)


ZoldorfTheWizard (approved):
2 years of this and we just figured it out lol

PR #322, opened 2017-08-11
Inverse entity queries.

PJB3005:
They now effectively behave like `Func<IEntity, bool>`. The actual matching logic is now done on entity queries.

They've also been abstracted, as such there are now multiple entity query types to pick from.

Also generic clean up.


SS14.Shared/GameObjects/EntityQuery.cs
```diff
@@ -1,20 +1,77 @@
...
+    /// An entity query that checks based on the components of an entity.
+    /// </summary>
+    /// <remarks>
+    /// The query can be controlled with <see cref="AllSet" />, <see cref="ExclusionSet" /> and <see cref="OneSet" />.
+    /// If all of these are empty, the query is effectively equal to <see cref="AllEntityQuery" />.
+    /// </remarks>
+    public class ComponentEntityQuery : IEntityQuery
+    {
+        /// <summary>
+        /// A list of component reference types, all of which an entity must have to pas.
```
RemieRichards: pass?

PJB3005: Oops

SS14.Shared/GameObjects/EntityQuery.cs
```diff
@@ -1,20 +1,77 @@
...
+        public bool Match(IEntity entity) => true;
+    }
+
+    /// <summary>
+    /// An entity query which will match entities based on a predicate.
+    /// </summary>
+    public class PredicateEntityQuery : IEntityQuery
     {
-        public EntityQuery()
+        public readonly Func<IEntity, bool> Predicate;
```
RemieRichards, line 69: you could use an actual ```Predicate<IEntity>``` here I believe.
Edit: Apparently Microsoft hates predicates explicitly used, up to you whether that's a + or a - :P

PJB3005: TIL.

PR #325, opened 2017-08-11
Component Rework

Acruid:
I merged a bunch of components together and generally tried to clean up the mess. The commits are pretty atomic, so you can read through and figure out which ones got merged/removed.

The first difference is that Lerping between states was removed on the client, meaning if the server tick rate is low you can see the client jump to each tick. Once the Components for Physics get agreed upon in https://github.com/space-wizards/space-station-14/issues/286, and component networking get finalized, features like Lerp and Prediction can be re-added.  The second difference is that the player movement is also mirrored, which is also caused by the same issue as https://github.com/space-wizards/space-station-14/issues/323. The third issue is that there are no grid physics, because Acruid gets sidetracked and the MapManager *still* needs to be remade, currently https://github.com/space-wizards/space-station-14/issues/150. Once this is completed it will be accessible from shared, and can be hooked back up to the physics system.


SS14.Server/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -1,65 +1,56 @@
...
 
-        public float X
-        {
-            get => Position.X;
-            set => Position = new Vector2f(value, Position.Y);
-        }
+        /// <summary>
+        ///     Current rotation offset of the entity.
+        /// </summary>
+        public Vector2 Rotation { get; set; }
```
PJB3005: Wouldn't it make more sense to store this as rotation in radians, instead of a unit vector?

PJB3005 (approved):

PR #327, opened 2017-08-12
Fixes Exceptions on Rejoining

clusterfack:
This fixes two different exceptions you received when rejoining

1. A server exception that occurred due to the fact that the gamestatemanager would get cleared if everyone disconnected, but the list that it uses as keys to itself would not get cleared. Now both are cleared.

2. A client exception that occurred due to the component references list being completely cleared. This doesn't get regenerated upon reconnect, @PJB3005 may have to say whether or not this is wise but you cannot reconnect with it there.

Client still looks like shit when you reconnect though, and you cant move your player character, and the console has a line that says player.name variable is already defined.

### DIFF
--- a/SS14.Shared/GameObjects/Entity.cs
+++ b/SS14.Shared/GameObjects/Entity.cs
@@ -421,7 +421,9 @@ namespace SS14.Shared.GameObjects
             }
             _components.Clear();
             _netIDs.Clear();
-//          _componentReferences.Clear();
+            _componentReferences.Clear();
+            var componentmanager = IoCManager.Resolve<IComponentManager>();
+            componentmanager.Cull();
         }
 
         public IEnumerable<IComponent> GetComponents()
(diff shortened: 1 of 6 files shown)


SS14.Shared/GameObjects/Entity.cs
```diff
@@ -421,7 +421,7 @@ public virtual void Shutdown()
             }
             _components.Clear();
             _netIDs.Clear();
-            _componentReferences.Clear();
+//          _componentReferences.Clear();
```
PJB3005: Wait what? Are these entities being re-used after you reconnect?

clusterfack: It seems so, if a client reconnects after this list gets cleared it getcomponent which uses this list immediately has an exception from trying to read it.

PJB3005: Ok these entities shouldn't be getting reused. Fix that instead of "solving" the exception without fixing it.

clusterfack:
Alright, I implemented PJB's suggestion and it seems to have fully brought the functionality back to what it used to be before.

PR #329, opened 2017-08-12
Adds a HSL -> RGB conversion helper.

PJB3005:


PJB3005:
Well turns out OpenTK has `Color4.FromHsl`.

PR #336, opened 2017-08-14
Fix issues with Menu Window

ghost:
- Opening Tile or Entity window will no longer kill the Menu Window, causing further issues
- Menu Window will no longer respond to ESC prompts if not focused. This works, surprisingly enough, when you close all the windows you're dealing with the menu window will respond to ESC prompts again. I guess this is normal behavior for a pause menu

Was tested, dunno what is up with the extra changes in Window and GUIComponent

### DIFF
--- a/SS14.Client/UserInterface/Components/EntitySpawnPanel.cs
+++ b/SS14.Client/UserInterface/Components/EntitySpawnPanel.cs
@@ -269,8 +269,13 @@ namespace SS14.Client.UserInterface.Components
 
         public override bool KeyDown(KeyEventArgs e)
         {
-            if (base.KeyDown(e)) return true;
-            return false;
+            if (e.Code != Keyboard.Key.Escape)
+            {
+                return false;
+            }
+
+            Dispose();
+            return true;
         }
     }
 }
--- a/SS14.Client/UserInterface/Components/EntitySpawnPanel.cs
+++ b/SS14.Client/UserInterface/Components/EntitySpawnPanel.cs
@@ -269,13 +269,12 @@ namespace SS14.Client.UserInterface.Components
 
         public override bool KeyDown(KeyEventArgs e)
         {
-            if (e.Code != Keyboard.Key.Escape)
+            if (e.Code == Keyboard.Key.Escape)
             {
-                return false;
+                Dispose();
+                return true;
             }
-
-            Dispose();
-            return true;
+            return false;
         }
     }
 }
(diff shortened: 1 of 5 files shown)


ghost:
And yeah, you're reading this right. "If the ESC menu is not visible, don't take ESC prompts"

It works exactly as I wanted it to. Pressing ESC opens and closes pause menu alone. If another menu is open, ESC closes it, THEN another ESC press opens the menu again

Magic

SS14.Client/UserInterface/Components/MenuWindow.cs
```diff
@@ -86,7 +89,12 @@ public override bool KeyDown(KeyEventArgs e)
             {
                 return false;
             }
-            
+
+            if(IsVisible() == false)
```
PJB3005: So, `!IsVisible()`

PJB3005 (changes requested):

SS14.Client/UserInterface/Components/EntitySpawnPanel.cs
```diff
@@ -269,8 +269,13 @@ public override bool MouseWheelMove(MouseWheelEventArgs e)
 
         public override bool KeyDown(KeyEventArgs e)
         {
-            if (base.KeyDown(e)) return true;
-            return false;
+            if (e.Code != Keyboard.Key.Escape)
```
RemieRichards: I feel it would be better to Dispose if the key IS Escape, rather than Not Disposing if it Isn't Escape.
(Typically 'KeyDown()' events handle a key, not handle not-that-key)

SS14.Client/UserInterface/Components/MenuWindow.cs
RemieRichards: or change it more:

```
if(IsVisible()){
    ToggleVisible();
}
```
(basically incorporating the line below)

SS14.Client/UserInterface/Components/TileSpawnPanel.cs
```diff
@@ -169,8 +169,13 @@ public override bool MouseWheelMove(MouseWheelEventArgs e)
 
         public override bool KeyDown(KeyEventArgs e)
         {
-            if (base.KeyDown(e)) return true;
-            return false;
+            if (e.Code != Keyboard.Key.Escape)
```
RemieRichards: see what I said before.

SS14.Client/UserInterface/Components/MenuWindow.cs
ghost: Change pushed

PJB3005 (approved):

PR #338, opened 2017-08-14
Map Manager Refactor

Acruid:
Cleans up the MapManager interface, and moves it to shared. Now supports multiple grids per map. Chunks are pretty much abstracted away, so the game logic does not have to worry about them.

This has been sitting in my repo long enough, and I don't really want to look at it anymore. There are prob some missing features/bugs with it, but it is working with no noticeable bugs. This should be a decent base to build off of.


ZoldorfTheWizard (changes requested):
.

ZoldorfTheWizard (approved):

ZoldorfTheWizard:
MacOS build failed. Windows/Linux succeeded

PR #339, opened 2017-08-15
Containers

clusterfack:
Creates an abstract container class and a container manager that entity storage is based on.

- Any component can create one (1) container which it can freely insert and remove entities from.
- Once a container has been created, the entity itself will have a ContainerManagerComponent which holds references to every container and allows removal of an entity from the entity regardless of which container it is in
- After inserting, the entity that was inserted has its transform's parent set to the transform of what it is inserted to. This creates a tree of parents that lead to the object that is placed on the map.
- This also leads to that we can getmaptransform() to find what is on the map, and get position can be rerouted until it gets to position at the top of the tree.

I left out two features I planned due to @PJB3005 thinking they shouldn't be implemented until we start expressly needing them. The two being
- A method that causes the map transform to call an event similar to onmoved for all objects nested inside the map transform when it is moved.
- Event handlers for insertion/removal that a component can add delegates to when it uses its container. At this point in time it seems unnecessary as people may find something they like better later.

Side notes:
- The logic behind this requires that entitys have transforms otherwise things get a bit whacky. If necessary it can be tweaked to not need transform components but then I can no longer guarantee non-circular storage.


SS14.Client/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -16,7 +16,7 @@ public class TransformComponent : Component, ITransformComponent
     {
         public Vector2 Position { get; private set; }
         public Angle Rotation { get; private set; }
-        public IEntity Parent { get; private set; }
+        public ITransformComponent Parent { get; set; }
```
PJB3005: Why did you make this publically accessible?

SS14.Client/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -67,13 +67,39 @@ private void DetachParent()
...
+            }
+            return this;
+        }
+
+        /// <summary>
+        ///     Does this entity contain the entity in the argument
+        /// </summary>
+        public bool ContainsEntity(ITransformComponent entity)
+        {
+            if(entity.Parent != null) //Is the entity on the map
```
PJB3005: Use braces for this if statement because this is incredibly ambiguous.

SS14.Shared/GameObjects/Components/Container/Container.cs
```diff
@@ -0,0 +1,82 @@
...
+            IContainerManager containermanager = holder.Owner.GetComponent<IContainerManager>();
+            if (containermanager == null)
+            {
+                var factory = IoCManager.Resolve<IComponentFactory>();
+                holder.Owner.AddComponent(factory.GetComponent<ContainerManagerComponent>());
+            }
+            Owner = holder.Owner;
+            if(containermanager.EntityContainers[holder.GetType()] != null)
+            {
+                throw new NotImplementedException(); //This system is designed to have singular containers per component, make a new component dummy.
```
PJB3005: In that case, throw an `InvalidOperationException`.

SS14.Shared/GameObjects/Components/Container/Container.cs
```diff
@@ -0,0 +1,82 @@
...
+using SS14.Shared.IoC;
+
+namespace SS14.Server.GameObjects.Components.Container
+{
+    public class Container : List<IEntity>, IContainer
+    {
+        public Container(IComponent holder)
+        {
+            IContainerManager containermanager = holder.Owner.GetComponent<IContainerManager>();
+            if (containermanager == null)
```
PJB3005: This can't even be true, as `GetComponent` throws an exception instead of returning `null`. You want to use `if (!holder.Owner.TryGetComponent<IContainerManager>(out var containermanager))` instead.

SS14.Client/GameObjects/Components/Transform/TransformComponent.cs
Acruid: This needs to be changed back to the more general IEntity. If you need to get the TransformComponent of the Parent, use `Parent.GetComponent<ITransformComponent>();` Also, this needs to be changed back to a private setter. The client does not decide if something is parented or not, it asks the server to parent things.

SS14.Client/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -67,13 +67,39 @@ private void DetachParent()
         /// Sets another entity as the parent entity.
         /// </summary>
         /// <param name="parent"></param>
-        private void AttachParent(IEntity parent)
+        public void AttachParent(ITransformComponent parent)
```
Acruid: Change this back to IEntity. This also needs to stay private, and is only called by the getter/setter of the public Parent property.

SS14.Client/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -54,10 +54,10 @@ public override void HandleComponentState(ComponentState state)
         /// <summary>
         /// Detaches this entity from its parent.
         /// </summary>
-        private void DetachParent()
+        public void DetachParent()
```
Acruid: Change back to private, this should only be called by the public Parent property.

SS14.Server/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -16,7 +16,7 @@ public class TransformComponent : Component, ITransformComponent
         /// <summary>
         ///     Current parent entity of this entity.
         /// </summary>
-        public IEntity Parent { get; set; }
+        public ITransformComponent Parent { get; set; }
```
Acruid, line 20: Change back to IEntity.

SS14.Shared/Interfaces/GameObjects/Components/ITransformComponent.cs
```diff
@@ -23,5 +23,24 @@ public interface ITransformComponent : IComponent
         ///     Event that gets invoked every time the position gets modified through properties such as <see cref="Rotation" />.
         /// </summary>
         event EventHandler<VectorEventArgs> OnMove;
+
+        /// <summary>
+        ///     Reference to the transform of the container of this object if it exists, can be nested several times.
+        /// </summary>
+        ITransformComponent Parent { get; set; }
```
Acruid: This needs to be the more general IEntity instead of a ITransformComponent.

SS14.Shared/GameObjects/Components/Transform/TransformComponentState.cs
```diff
@@ -14,7 +15,7 @@ public class TransformComponentState : ComponentState
         /// <summary>
         ///     Current parent entity of this entity.
         /// </summary>
-        public readonly IEntity Parent;
+        public readonly ITransformComponent Parent;
```
Acruid, line 18: Change back to IEntity.

SS14.Shared/GameObjects/Components/Transform/TransformComponentState.cs
```diff
@@ -31,8 +32,8 @@ public class TransformComponentState : ComponentState
         /// </summary>
         /// <param name="position">Current position offset of the entity.</param>
         /// <param name="rotation">Current direction offset of the entity.</param>
-        /// <param name="parent">Current parent entity of this entity.</param>
-        public TransformComponentState(Vector2 position, Angle rotation, IEntity parent)
+        /// <param name="parent">Current parent transform of this entity.</param>
+        public TransformComponentState(Vector2 position, Angle rotation, ITransformComponent parent)
```
Acruid, line 36: Back to IEntity.

SS14.Shared/GameObjects/Components/Container/Container.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Threading.Tasks;
+using SS14.Shared.GameObjects;
+using SS14.Shared.Interfaces.GameObjects;
+using SS14.Server.Interfaces.GameObjects;
+using SS14.Shared.Interfaces.GameObjects.Components;
+using SS14.Shared.IoC;
+
+namespace SS14.Server.GameObjects.Components.Container
+{
+    public class Container : List<IEntity>, IContainer
```
PJB3005: Inheriting a basic type like `List<>` seems like a terriblee idea, because you can't control the accessibility of yourself. Make the list be a private field instead.

SS14.Shared/GameObjects/Components/Container/Container.cs
```diff
@@ -0,0 +1,82 @@
...
+        public Container(IComponent holder)
+        {
+            IContainerManager containermanager = holder.Owner.GetComponent<IContainerManager>();
+            if (containermanager == null)
+            {
+                var factory = IoCManager.Resolve<IComponentFactory>();
+                holder.Owner.AddComponent(factory.GetComponent<ContainerManagerComponent>());
+            }
+            Owner = holder.Owner;
+            if(containermanager.EntityContainers[holder.GetType()] != null)
```
PJB3005: Doesn't this throw a `KeyNotFoundException` instead of returning null?

SS14.Shared/GameObjects/Components/Container/Container.cs
```diff
@@ -0,0 +1,82 @@
...
+        public bool CanInsert(IEntity toinsert)
+        {
+            if (toinsert.GetComponent<ITransformComponent>().ContainsEntity(Owner.GetComponent<ITransformComponent>())) //Crucial, prevent circular insertion
+            {
+                return false;
+            }
+            return true;
+        }
+
+        public bool Insert(IEntity toinsert)
```
PJB3005: Don't use boolean return values to say whether something was successful or not. Return void and throw an exception on error.

SS14.Shared/GameObjects/Components/Container/Container.cs
```diff
@@ -0,0 +1,82 @@
...
+        public bool CanRemove(IEntity toremove)
+        {
+            if(!this.Contains(toremove))
+            {
+                return false;
+            }
+            return true;
+        }
+
+        public new bool Remove(IEntity toremove)
```
PJB3005: Same deal, throw exception if things are wrong instead of returning a boolean.

SS14.Shared/GameObjects/Components/Container/ContainerManagerComponent.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace SS14.Server.GameObjects.Components.Container
+{
+    public class ContainerManagerComponent : Component, IContainerManager
+    {
+        public override string Name => "ContainerContainer";
+
+        Dictionary<Type, Container> IContainerManager.EntityContainers { get; set; }
+        public Dictionary<Type, Container> EntityContainers = new Dictionary<Type, Container>();
+
+        public bool Remove(IEntity entity)
```
PJB3005: No boolean error codes.

SS14.Shared/GameObjects/Components/Container/ContainerManagerComponent.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Threading.Tasks;
+
+namespace SS14.Server.GameObjects.Components.Container
+{
+    public class ContainerManagerComponent : Component, IContainerManager
+    {
+        public override string Name => "ContainerContainer";
+
+        Dictionary<Type, Container> IContainerManager.EntityContainers { get; set; }
+        public Dictionary<Type, Container> EntityContainers = new Dictionary<Type, Container>();
```
PJB3005: Uh, what? You defined the interface explicitly and didn't use it, and used a separate declaration instead?

SS14.Shared/SS14.Shared.csproj
```diff
@@ -136,13 +136,17 @@
...
+    <Compile Include="GameObjects\Components\Container\Container.cs" />
+    <Compile Include="GameObjects\Components\Container\ContainerManagerComponent.cs" />
     <Compile Include="Interfaces\Configuration\IConfigurationManager.cs" />
     <Compile Include="ConsoleCommand.cs" />
     <Compile Include="EntityEvents.cs" />
     <Compile Include="Exceptions\TypeArgumentException.cs" />
     <Compile Include="GameObjects\Entity.cs" />
     <Compile Include="Interfaces\GameObjects\Components\IClickableComponent.cs" />
     <Compile Include="Interfaces\GameObjects\Components\ICollidableComponent.cs" />
+    <Compile Include="Interfaces\GameObjects\Components\Icontainer.cs" />
```
PJB3005: You messed up the capitalization here.

SS14.Shared/SS14.Shared.csproj
```diff
@@ -136,13 +136,17 @@
...
+    <Compile Include="Interfaces\GameObjects\Components\iContainerManager.cs" />
```
PJB3005: Captialization messed up.

SS14.Shared/GameObjects/Components/Container/Container.cs
ZoldorfTheWizard: Yes it does

SS14.Shared/GameObjects/Components/Container/Container.cs
clusterfack: Alright good

SS14.Shared/GameObjects/Components/Container/Container.cs
clusterfack: No that's incredibly stupid

SS14.Shared/GameObjects/Components/Container/Container.cs
clusterfack: Same

SS14.Shared/GameObjects/Components/Container/ContainerManagerComponent.cs
clusterfack: Still dumb

SS14.Shared/GameObjects/Components/Container/Container.cs
PJB3005: How is that stupid? You're doing something obviously wrong and 99% of the time this would be an error. Failing silently is never the solution because it just sets everything up for different bugs and issues later, that are much harder to debug.

SS14.Client/GameObjects/Components/Transform/TransformComponent.cs
clusterfack: There's not much reason for it to be entity, I get the point about the private setter though

clusterfack: Yeah making it private again

SS14.Shared/GameObjects/Components/Container/Container.cs
clusterfack: Thanks I was looking for a more appropriate error to throw

SS14.Shared/GameObjects/Components/Container/Container.cs
clusterfack: Yeah, I was basing it on another class that I've seen but that makes more sense

clusterfack:
I think I've brought @Acruid around to keeping it as ITransformComponents, for now at least until such a distinction becomes important or necessary.

However someone will have to explain to me how we will attach and detach parents if we want to hide it from the client, but the interface we need to reference from is in shared. IE we use the ITransformComponent interface to attach parents, the client implements the ITransformComponent interface, therefore the client can attach parents under the model necessary for compile to succeed.

PJB3005:
@Clusterfack make an `IServerTransformComponent`

SS14.Shared/GameObjects/Components/Container/Container.cs
PJB3005: Ok I'm fine with keeping this boolean return, because user code can block it, but I do still think that the infinite loop prevention should throw an exception.

clusterfack: Alright I'll make it throw an exception

clusterfack:
Alright, I'll do that

SS14.Client/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -67,13 +67,48 @@ private void DetachParent()
...
+            }
+            return true;
+        }
+
+        /// <summary>
+        ///     Does this entity contain the entity in the argument
+        /// </summary>
+        public bool ContainsEntity(ITransformComponent transform)
+        {
+            if(IsMapTransform(transform)) //Is the entity on the map
```
PJB3005: Still not using braces here.

SS14.Server/GameObjects/Components/Container/ContainerManagerComponent.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Threading.Tasks;
+
+namespace SS14.Server.GameObjects.Components.Container
+{
+    public class ContainerManagerComponent : Component, IContainerManager
+    {
+        public override string Name => "ContainerContainer";
+
+        public Dictionary<Type, IContainer> EntityContainers = new Dictionary<Type, IContainer>();
+        Dictionary<Type, IContainer> IContainerManager.EntityContainers => EntityContainers;
```
PJB3005: This is still messed up. You can just remove this line and make the top one an auto-implemented property like this:
```cs
public Dictionary<Type, IContainer> EntityContainers { get; }= new Dictionary<Type, IContainer>();

SS14.Client/GameObjects/Components/Transform/TransformComponent.cs
clusterfack: Oh I see what you mean, I thought you mean semantically ambiguous and not syntactically ambiguous

SS14.Server/GameObjects/Components/Container/ContainerManagerComponent.cs
clusterfack: Could have said that from the start tbh

PJB3005: Uh... Actually using the declaration in an interface is pretty normal that doesn't need saying...

clusterfack: Fuck off

PR #341, opened 2017-08-16
Replace SFML math types with OpenTK and custom ones.

PJB3005:
Removing of dependency on SFML to make replacing it easier.

Also removes the redundant dead vector implementations from the code. That's the biggest part of the deletions in the diff.


ZoldorfTheWizard (approved):
Looks good. I use OpenTK math anyway so the math will integrate flawlessly

PR #346, opened 2017-08-18
Refactor entity initialization and fix doc comments.

PJB3005:
Entities now initialize in a few stages. First all their components are added, then `entity.PreInitialize()` is called, then all components are initialized, then `entity.Initialize()` is called.

This multi-level initialization system should hopefully be more powerful and make things like prefetching components easier.

Also what the crap WAS the old init system even?

Also removes the use of constructors completely, thus making it not painful to override entities and make initialization fully interface bound and not rely on type-unsafe constructor arguments.

### DIFF
--- a/SS14.Shared/GameObjects/EntityManager.cs
+++ b/SS14.Shared/GameObjects/EntityManager.cs
@@ -39,18 +39,6 @@ namespace SS14.Shared.GameObjects
 
         #region IEntityManager Members
 
-        public virtual void InitializeEntities()
-        {
-            foreach (IEntity e in _entities.Values.Where(e => !e.Initialized))
-            {
-                e.Initialize();
-            }
-        }
-
-        public virtual void LoadEntities()
-        {
-        }
-
         public virtual void Shutdown()
         {
             FlushEntities();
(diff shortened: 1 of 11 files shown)


UristMcDorf:
So if I had a component that wants to get certain info from other components near the beginning of an entity's lifetime I'd use Initialize(), correct?

PJB3005:
Yes.

UristMcDorf:
Okay, gotcha.

SS14.Shared/GameObjects/EntityManager.cs
```diff
@@ -148,23 +136,54 @@ public void FlushEntities()
...
+                InitializeEntity(entity);
+            }
+
+            return entity;
+        }
+
+        private void InitializeEntity(IEntity entity)
+        {
+            entity.PreInitialize();
+            foreach (var component in entity.GetComponents())
```
psykzz, line 171: A general question we might need to raise to an issue, but will we ever need components to be initialized in a specific order?

psykzz:
Any thoughts on `PostInitialize()`?

SS14.Shared/GameObjects/EntityManager.cs
UristMcDorf: From a component coder's viewpoint, I say no. If a component needs another component, then it can do that part of the setup after all components are initialised.

PJB3005:
Honestly I'm not gonna add extra init methods until we actually need it. This many is already plenty and it's just adding cost otherwise.

PR #348, opened 2017-08-18
Fix F2 on Unix, collision bug and join crash.

PJB3005:


PJB3005:
Linux and Windows passed, merging this because I can't wait 10 minutes on MacOS builds.

PR #358, opened 2017-08-20
Implement MsgEntity

wixoaGit:
Implement the MsgEntity NetMessage, and use it.


SS14.Shared/Network/Messages/MsgEntity.cs
```diff
@@ -16,16 +21,202 @@ public MsgEntity(INetChannel channel)
         { }
         #endregion
 
-        public NetIncomingMessage Output { get; set; }
+        [Dependency]
```
PJB3005: `[Dependency]` doesn't work outside of things created via IoC.

PJB3005 (approved):

PR #370, opened 2017-08-23
Tiles Helper Functions

clusterfack:
Creates a helper function for making tile indices into tile refs using the grid system, and creates a function to get a tileref one step in any particular direction of another tileref.

### DIFF
--- a/SS14.Shared/Map/Chunk.cs
+++ b/SS14.Shared/Map/Chunk.cs
@@ -64,6 +64,14 @@ namespace SS14.Shared.Map
             var indices = ChunkTileToGridTile(new MapGrid.Indices(xTile, yTile));
             return new TileRef(_mapManager, _grid.Index, indices.X, indices.Y, _tiles[xTile, yTile]);
         }
+        public TileRef GetTile(MapGrid.Indices indices)
+        {
+            // array out of bounds
+            if (indices.X >= ChunkSize || indices.Y >= ChunkSize)
+                throw new ArgumentException("Tile indices out of bounds.");
+            
+            return new TileRef(_mapManager, _grid.Index, indices.X, indices.Y, _tiles[indices.X, indices.Y]);
+        }
 
         /// <inheritdoc />
         [Obsolete("Enumerate over the chunk instead.")]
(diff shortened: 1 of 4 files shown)


SS14.Shared/Map/Chunk.cs
```diff
@@ -64,6 +64,14 @@ public TileRef GetTile(ushort xTile, ushort yTile)
             var indices = ChunkTileToGridTile(new MapGrid.Indices(xTile, yTile));
             return new TileRef(_mapManager, _grid.Index, indices.X, indices.Y, _tiles[xTile, yTile]);
         }
+        public TileRef GetTile(MapGrid.Indices indices)
+        {
+            // array out of bounds
+            if (indices.X >= ChunkSize || indices.Y >= ChunkSize)
```
PJB3005: Can't these also be negative?

SS14.Shared/Map/Chunk.cs
```diff
@@ -64,6 +64,14 @@ public TileRef GetTile(ushort xTile, ushort yTile)
...
+                throw new ArgumentException("Tile indices out of bounds.");
```
PJB3005: Use `ArgumentOutOfRangeException` instead.

SS14.Shared/Map/Chunk.cs
clusterfack: No, they're 0 to chunksize

SS14.Shared/Map/Chunk.cs
clusterfack: Alright, I'll replace it in the other place its used as well

SS14.Shared/Map/Chunk.cs
PJB3005: Well yes that's why you verify the input to prevent issues, but while you checked to make sure they're not too big, you didn't check whether they were too small.

clusterfack: Yeah thats fair, someone else might fuck up and use it wrong

YotaXP:
I'm a little late to the party, but I don't see the purpose of `MapGrid.IndicesToTile`.  `GetTile` does the exact same thing, since the `TileRef` struct was designed to be able to point to non-existent tiles.  Also, it should be named `TryGetTile` for consistency with similar .Net methods.

PR #373, opened 2017-08-24
Makes Walls Entities, removes tile walls

clusterfack:
So this makes walls into entities, they still have shadowcasting and collision. Lights and everything work

### DIFF
--- a/SS14.Client/GameObjects/ClientEntityManager.cs
+++ b/SS14.Client/GameObjects/ClientEntityManager.cs
@@ -33,6 +33,26 @@ namespace SS14.Client.GameObjects
             }
         }
 
+        public IEnumerable<IEntity> GetEntitiesIntersecting(Box2 position)
+        {
+            foreach (var entity in _entities.Values)
+            {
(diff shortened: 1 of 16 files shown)


clusterfack:
That appveyor error aint me, that's something else entirely.

SS14.Client/GameObjects/ClientEntityManager.cs
```diff
@@ -33,6 +33,47 @@ public IEnumerable<IEntity> GetEntitiesInRange(Vector2 position, float Range)
...
+                    {
+                        yield return entity;
+                    }
+                }
+            }
+        }
+
+        public bool AnyEntitiesIntersecting(Box2 position)
+        {
+            foreach (var entity in _entities.Values)
```
PJB3005, line 58: This could be shortened to `GetEntitiesIntersecting(position).Any()`, though that might perform a *little* bit worse. It's your call really.

Silvertorch5:
Rebuild success.

PR #374, opened 2017-08-24
Shifts down the mob's bounding box by 0.6 tiles

clusterfack:
This centers it just slightly above the feet of the mob. Otherwise the bounding box of the mob allows it to directly have its legs over walls when colliding with them, and colliding at walls from the bottom at chest height.

This is how it used to look


YotaXP:
Shouldn't the sprite just be shifted further up?  The collision box should totally be around the entity's origin.

PJB3005:
That actually raises an interesting question. Is the coordinate of a mob the center or the bottom left?

PR #375, opened 2017-08-24
Move client GameState management

wixoaGit:
Moves the client's GameState message handling to GameStateManager, and creates MsgFullState and MsgUpdateState.


SS14.Client/GameStates/GameStateManager.cs
```diff
@@ -1,57 +1,95 @@
...
-        private readonly Dictionary<long, uint> ackedStates = new Dictionary<long, uint>();
-        private uint currentStateSeq;
-        public GameState CurrentState { get; private set; }
+        public Dictionary<uint, GameState> GameStates;
 
-        #region IGameStateManager Members
+        private List<uint> AckedGameStates;
 
-        public void Cull()
+        public GameState CurrentState
```
PJB3005: Use anonymous functions here for simple getters:
```cs
public GameState CurrentState => GameStates.Values.Last();
```

That said, doesn't this make performance a lot worse, *and* doesn't the C# documentation say that the order of dictionary iteration is undefined?

SS14.Client/GameStates/GameStateManager.cs
```diff
@@ -1,57 +1,95 @@
...
+        {
+            foreach (uint v in GameStates.Keys.Where(v => v < sequence).ToList())
+                GameStates.Remove(v);
+        }
 
-        public void CullAll()
+        private void AckGameState(uint sequence)
         {
-            Clear();
+            IClientNetManager networkManager = IoCManager.Resolve<IClientNetManager>();
```
PJB3005: Because this is an IoC-managed class, you can store the network manager in a field as dependency like so:

```cs
[Dependency]
private readonly IClientNetManager networkManager;
``` 

In fact, do that for all IoC resolutions in this class.

PJB3005 (changes requested):

PJB3005 (approved):

clusterfack:
This causes the client to get stuck in a neverending memory state patching loop if you pause and unpause the client process

PR #380, opened 2017-08-25
Fixes moonwalk bug and inverted up/down controls.

Acruid:
It is a bit hacky for my taste, but it should do.


PJB3005 (approved):

Acruid:
This fixes https://github.com/space-wizards/space-station-14/issues/323.

PR #387, opened 2017-08-27
Entity Component Shutdown Fix.

Acruid:
Now when an Entity is Shutdown(), components are actually Shutdown() and OnRemoved()'d properly.
I also moved various component cleanup code from OnRemove to Shutdown, because most of the cleanup code was referencing other components that *could* already be removed.

This fixes issues with particles, collidables, and lights not being cleaned up when their controlling component was removed. This should fix https://github.com/space-wizards/space-station-14/issues/179.

### DIFF
--- a/SS14.Client/GameObjects/Components/Light/PointLightComponent.cs
+++ b/SS14.Client/GameObjects/Components/Light/PointLightComponent.cs
@@ -135,12 +135,12 @@ namespace SS14.Client.GameObjects
                 ModeClass = LightModeClass.Constant;
             }
         }
-
-        public override void OnRemove()
+        
+        public override void Shutdown()
         {
             Owner.GetComponent<ITransformComponent>().OnMove -= OnMove;
             IoCManager.Resolve<ILightManager>().RemoveLight(Light);
-            base.OnRemove();
+            base.Shutdown();
         }
 
         private void OnMove(object sender, VectorEventArgs args)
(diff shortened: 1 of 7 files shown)


ZoldorfTheWizard:
My hero

PJB3005 (approved):

SS14.Client/GameObjects/Components/Light/PointLightComponent.cs
```diff
@@ -135,12 +135,12 @@ public override void LoadParameters(YamlMappingNode mapping)
...
             }
         }
-
-        public override void OnRemove()
+        
+        public override void Shutdown()
         {
             Owner.GetComponent<ITransformComponent>().OnMove -= OnMove;
             IoCManager.Resolve<ILightManager>().RemoveLight(Light);
-            base.OnRemove();
```
ZoldorfTheWizard, line 143: Grumble Grumble Cluster Grumble...

ZoldorfTheWizard (approved):
I n s t a m e r g e

SS14.Client/GameObjects/Components/Light/PointLightComponent.cs
clusterfack: I never touched this file tho

ZoldorfTheWizard: grumble

PR #388, opened 2017-08-27
Adds rendering color parameter to sprite components

clusterfack:
- You can load up a color from the system color names as a parameter
- You can alternatively load a color through a hex color parameter which takes hex strings

This causes the sprite to display as the color when being rendered, it changes the color to the color variable, renders it, and resets it afterwards.

Unfortunately it does not affect when you generically use getsprite(spritename). Might be nice to implement getting a fully descriptive sprite from the entity prototype system later that includes color.

### DIFF
--- a/SS14.Client/GameObjects/AnimatedSprite.cs
+++ b/SS14.Client/GameObjects/AnimatedSprite.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using SS14.Shared.Maths;
+using OpenTK.Graphics;
 
 namespace SS14.Client.Graphics.Sprite
 {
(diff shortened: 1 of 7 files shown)


SS14.Client/GameObjects/Components/Renderable/AnimatedSpriteComponent.cs
```diff
@@ -194,6 +196,16 @@ public override void LoadParameters(YamlMappingNode mapping)
                 SetDrawDepth(node.AsEnum<DrawDepth>());
             }
 
+            if (mapping.TryGetNode("color", out node))
```
PJB3005, line 199: You could make these the same field. Read the node as string, see if it's a hex color, if not try to resolve from name.

SS14.Client/GameObjects/AnimatedSprite.cs
```diff
@@ -135,10 +136,12 @@ public SFML.Graphics.Sprite GetCurrentSprite()
             return _currentSprite;
         }
 
-        public void Draw()
+        public void Draw(Color4 Color)
         {
             _currentSprite.Scale = new SFML.System.Vector2f(HorizontalFlip ? -1 : 1, 1);
+            _currentSprite.Color = Color.Convert();
             _currentSprite.Draw();
+            _currentSprite.Color = Color4.White.Convert();
```
ZoldorfTheWizard, line 144: This line is not needed as we convert Color above on line 142

PR #389, opened 2017-08-27
Creates a Snap Checking System

### LINKED ISSUE #395: YAML Parenting only uses parented component values
PushInheritance uses a component specific code to give inheritance down to children. 

In particular this means that other prototype specific values such as placement or snapping that are not on components do not get inherited by child prototypes.

clusterfack:
Status: yaml works, snapgrid placement works, tryspawnentityat works, testing showed everything is fine

Creates a system that attempts to sanitize placement of objects when attempting to spawn them on snapping grids. This checks whether an object is trying to spawn specifically on a snap grid, and if it is checks against a snapflag of the type to see if other exclusive snappable objects are on that snap grid point.

Separates the spawnentityat into tryspawnentityat which checks for snappables, and forcespawnentityat which completely ignores that.
Fixes #395


SS14.Shared/Prototypes/SnapFlags.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SS14.Shared.Prototypes
+{
+    [Flags]
+    public enum SnapFlags : int
```
PJB3005: The problem is that these flags can't be extended in the slightest, so this ruins the content assembly.

I would recomment using a set of strings instead of bitflags.

SS14.Shared/Prototypes/IPrototype.cs
```diff
@@ -16,6 +16,11 @@ public interface IPrototype
         /// Load data from the YAML mappings in the prototype files.
         /// </summary>
         void LoadFrom(YamlMappingNode node);
+
+        /// <summary>
+        /// Bitflag to hold snapping categories that this object has applied to it such as pipe/wire/wallmount
+        /// </summary>
+        SnapFlags SnapFlags { get; set; }
```
PJB3005: `IPrototype` is for generic prototypes, including stuff such as recipes. This should go under `EntityPrototype`

SS14.Shared/Prototypes/PrototypeManager.cs
```diff
@@ -60,6 +64,10 @@ public interface IPrototypeManager
         /// Syncs all inter-prototype data. Call this when operations adding new prototypes are done.
         /// </summary>
         void Resync();
+        /// <summary>
+        /// Tests whether the prototype is going to conflict with anything previously placed in this location on the snap grid
+        /// </summary>
+        bool CanSpawnAt(string EntityType, IMapGrid grid, Vector2 position);
```
PJB3005: This functionality shouldn't be part of the prototype manager itself. It should be on entity prototypes.

SS14.Shared/Prototypes/EntityPrototype.cs
```diff
@@ -135,6 +140,16 @@ public void LoadFrom(YamlMappingNode mapping)
             {
                 ReadPlacementProperties(placementMapping);
             }
+
+            // Reads snapping flags that this object holds that describe its properties to such as wire/pipe/wallmount, used to prevent certain stacked placement
+            if (mapping.TryGetNode<YamlSequenceNode>("snap", out var snapsequence))
```
PJB3005: Put this under `ReadPlacementProperties` so that the syntax becomes this:
```yml
- type: entty
  ...
  placement:
    snap:
    - Pipe
    ...
```

SS14.Shared/GameObjects/EntityManager.cs
```diff
@@ -95,6 +97,18 @@ public IEnumerable<IEntity> GetEntities(IEntityQuery query)
             return _entities.Values.Where(e => e.Match(query));
         }
 
+        public IEnumerable<IEntity> GetEntitiesAt(Vector2 position)
+        {
+            foreach (var entity in _entities.Values)
+            {
+                var transform = entity.GetComponent<ITransformComponent>();
+                if (transform.Position == position)
```
PJB3005: Doing hard equality checking on floating point numbers like this is highly error prone. Allow a small range of error and use `<` and such.

SS14.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -40,12 +42,65 @@ public void SaveEntities()
...
         /// <param name="position"></param>
         /// <returns></returns>
-        public IEntity SpawnEntityAt(string EntityType, Vector2 position)
+        public bool TrySpawnEntityAt(string EntityType, IMapGrid grid, Vector2 position, out IEntity entity)
         {
-            IEntity e = SpawnEntity(EntityType);
-            e.GetComponent<TransformComponent>().Position = position;
-            e.Initialize();
-            return e;
+            var protomanager = IoCManager.Resolve<IPrototypeManager>();
```
PJB3005: These resolves should be replaced with `[Dependency]` fields on the manager.

SS14.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -40,12 +42,65 @@ public void SaveEntities()
...
+            if(protomanager.CanSpawnAt(EntityType, grid, position))
+            {
+                entity = SpawnEntity(EntityType);
+                entity.GetComponent<TransformComponent>().Position = position;
+                entity.Initialize();
+                return true;
+            }
+            entity = new Entity(); //This is a bad
```
PJB3005: If it's bad why are you doing it? Can't you just set the field to `null`?

SS14.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -40,12 +42,65 @@ public void SaveEntities()
...
+            return false;
+        }
+
+        /// <summary>
```
PJB3005: These doc comments should go in the interface declaration, not the implementation.

SS14.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -40,12 +42,65 @@ public void SaveEntities()
...
+        /// </summary>
+        /// <param name="EntityType"></param>
+        /// <param name="position"></param>
+        /// <returns></returns>
+        public bool TrySpawnEntityAt(string EntityType, Vector2 position, out IEntity entity)
+        {
+            var mapmanager = IoCManager.Resolve<IMapManager>(); //TODO: wait we only have one map? this only supports one map
+            if(mapmanager.TryFindGridAt(position, out IMapGrid gridlocation))
+            {
+                bool returnvalue = TrySpawnEntityAt(EntityType, gridlocation, position, out IEntity returnoutvar);
```
PJB3005: You can shorten this to `return TrySpawnEntityAt(EntityType, gridlocation, position, out entity);`

SS14.Shared/Prototypes/IPrototype.cs
clusterfack: This is true

SS14.Shared/Prototypes/SnapFlags.cs
clusterfack: Ehhhh I really think strings are bad but I'm not really sure how you could run your own snapflags in content otherwise

SS14.Shared/GameObjects/EntityManager.cs
clusterfack: Well it works currently, if we're going to do that for checking float equality we should standardize it.

SS14.Server/GameObjects/ServerEntityManager.cs
clusterfack: No because out vars force a return value

SS14.Server/GameObjects/ServerEntityManager.cs
clusterfack: What, putting a reference to it on the entity manager with a deodency attribute or..?

SS14.Server/GameObjects/ServerEntityManager.cs
clusterfack: Cool I'll try that

SS14.Shared/Prototypes/EntityPrototype.cs
clusterfack: Sure

SS14.Shared/Prototypes/PrototypeManager.cs
clusterfack: Are we implying people are going to override entity prototypes are some point or what

SS14.Server/GameObjects/ServerEntityManager.cs
PJB3005: Yes.

https://github.com/space-wizards/space-station-14/wiki/IoC-(Inversion-of-Control)#field-injection-and-you

SS14.Shared/Prototypes/PrototypeManager.cs
PJB3005: No, but it should be part of the entity prototypes because the prototype manager is generic and designed for more than just entities.

SS14.Server/GameObjects/ServerEntityManager.cs
PJB3005: Sure about that?

https://dotnetfiddle.net/PzHqPt

RemieRichards: null IS a return value.
it's just a null one.

PJB3005:
I made a PR: https://github.com/clusterfack/space-station-14/pull/1

SS14.Server/GameObjects/ServerEntityManager.cs
clusterfack: And not all types are nullable, though I guess typecasting downwards like PJB suggests shouldn't make a difference as long as people are checking the bool sanely.

clusterfack:
Alright besides that float equals standardization function that we need to put in somewhere, this is donezo

clusterfack:
funcode fixing #https://github.com/space-wizards/space-station-14/issues/395 incoming

PR #396, opened 2017-08-28
Fixes Bounding/Clicking Bugs

clusterfack:
This fixes a number of bugs with the clicking system.

- Wearable animated components were fixed so that they properly report their AABB 
- Wearable animated components draw their AABB in the AABB debug
- Wearable animated components were fixed so that they properly use their nonwornsprite or wornsprite for checkclicking
- Fixes the clicking system only choosing the player character
- Fixes the eraser only working within a 1.5 tile range

### DIFF
--- a/SS14.Client/GameObjects/ClientEntityManager.cs
+++ b/SS14.Client/GameObjects/ClientEntityManager.cs
@@ -53,6 +53,26 @@ namespace SS14.Client.GameObjects
             }
         }
 
+        public IEnumerable<IEntity> GetEntitiesIntersecting(Vector2 position)
+        {
+            foreach (var entity in _entities.Values)
+            {
+                if (entity.TryGetComponent<BoundingBoxComponent>(out var component))
+                {
+                    if (component.WorldAABB.Contains(position))
+                        yield return entity;
+                }
+                else
+                {
+                    var transform = entity.GetComponent<ITransformComponent>();
+                    if (position == transform.Position)
+                    {
+                        yield return entity;
+                    }
+                }
+            }
+        }
+
         public bool AnyEntitiesIntersecting(Box2 position)
         {
             foreach (var entity in _entities.Values)
--- a/SS14.Client/GameObjects/ClientEntityManager.cs
+++ b/SS14.Client/GameObjects/ClientEntityManager.cs
@@ -65,7 +65,7 @@ namespace SS14.Client.GameObjects
                 else
                 {
                     var transform = entity.GetComponent<ITransformComponent>();
-                    if (position == transform.Position)
+                    if (FloatMath.CloseTo(transform.Position.X, position.X) && FloatMath.CloseTo(transform.Position.Y, position.Y))
                     {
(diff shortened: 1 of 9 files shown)


SS14.Client/GameObjects/ClientEntityManager.cs
```diff
@@ -53,6 +53,26 @@ public IEnumerable<IEntity> GetEntitiesIntersecting(Box2 position)
...
+            {
+                if (entity.TryGetComponent<BoundingBoxComponent>(out var component))
+                {
+                    if (component.WorldAABB.Contains(position))
+                        yield return entity;
+                }
+                else
+                {
+                    var transform = entity.GetComponent<ITransformComponent>();
+                    if (position == transform.Position)
```
PJB3005: No hard float equality checking, it's too unreliable.

SS14.Client/GameObjects/ClientEntityManager.cs
```diff
@@ -53,6 +53,26 @@ public IEnumerable<IEntity> GetEntitiesIntersecting(Box2 position)
...
         }
 
+        public IEnumerable<IEntity> GetEntitiesIntersecting(Vector2 position)
+        {
+            foreach (var entity in _entities.Values)
+            {
+                if (entity.TryGetComponent<BoundingBoxComponent>(out var component))
+                {
+                    if (component.WorldAABB.Contains(position))
+                        yield return entity;
```
PJB3005, line 63: Wrap this in braces. No brace-less if statements.

SS14.Client/GameObjects/Components/Renderable/WearableAnimatedSpriteComponent.cs
```diff
@@ -43,6 +44,13 @@ public void SetCarriedSprite(string spritename)
             CarriedSprite = spritename;
         }
 
+        public override Sprite GetCurrentSprite()
+        {
+            if (!IsCurrentlyWorn)
```
PJB3005, line 49: No brace-less ifs.

SS14.Client/GameObjects/ClientEntityManager.cs
clusterfack: Sure we can make that a standard, but it makes no sense to start making it a requirement when nothing in the codebase follows it.

Like we always do make a PR to change it everywhere and then lets standardize it.

SS14.Client/GameObjects/ClientEntityManager.cs
clusterfack: Need a standard function for this

SS14.Client/GameObjects/Components/Renderable/WearableAnimatedSpriteComponent.cs
clusterfack: Ditto above

clusterfack:
The braces will end up getting changed once we PR the regexed brace changes

PR #404, opened 2017-09-01
Texture Loading Performance

Acruid:
Reworked the alpha clicking, removed the useless clickmaps, and gained about 20% faster load times according to the profiler.

The Texture ctor in ResourceCache.LoadTextureFrom is still taking up about 55% of the total loading time of the client. Do we really need to load every texture in the VFS into GPU memory before the client window opens?


psykzz:
The opacity map was to fix issue where when running across the screen, if you were to spam left click, you'd have some noticeable lag.

Are you sure this isn't going to cause a regression?

Acruid:
I have tested this PR again, and it does not cause noticeable fps lag when clicking on an object. The GetPixel function is never even called unless the bounding box checks pass and the entity has a ClickableComponent.

PR #407, opened 2017-09-04
Add our own Vector2 type.

PJB3005:


SS14.Shared/Maths/Vector2.cs
```diff
@@ -0,0 +1,265 @@
...
+        ///     Clamps the components of a vector to minimum and maximum vectors.
+        /// </summary>
+        /// <param name="vector">The vector to clamp.</param>
+        /// <param name="min">The lower bound vector.</param>
+        /// <param name="max">The upper bound vector.</param>
+        public static Vector2 Clamp(Vector2 vector, Vector2 min, Vector2 max)
+        {
+            return new Vector2(
+                vector.X.Clamp(min.X, max.X),
+                vector.X.Clamp(min.Y, max.Y)
```
YotaXP: Shouldn't this be `vector.Y.Clamp(...)`?

PJB3005: Yes it definitely should be. Good catch.

psykzz (approved):

SS14.Shared/Maths/Vector2.cs
```diff
@@ -0,0 +1,265 @@
...
+
+        /// <summary>
+        ///     Returns the hash code for this instance.
+        /// </summary>
+        /// <returns>A unique hash code for this instance.</returns>
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                return (X.GetHashCode() * 397) ^ Y.GetHashCode();
```
psykzz, line 261: why the random number here?

Acruid: It is a prime number that reduces collisions in the hash.

gbasood: Still, it's a magic number. Make it a const somewhere, and comment on why it exists so we don't have people asking this question a year from now.

PR #413, opened 2017-09-12
Sprite Clicking Rework

Acruid:
I completely remade the WasClicked() functions for the sprite components. Bounding boxes and Transparency click filtering properly works now. I added lots of comments and simple steps to help whoever is going to be looking at the code a year from now. This fixes https://github.com/space-wizards/space-station-14/issues/411.

Box2 get a new Scale() helper function.

Entity deleting bug was fixed on the client, so the GUI entity eraser works again.

### DIFF
--- a/SS14.Client/GameObjects/Components/Renderable/AnimatedSpriteComponent.cs
+++ b/SS14.Client/GameObjects/Components/Renderable/AnimatedSpriteComponent.cs
@@ -42,14 +42,10 @@ namespace SS14.Client.GameObjects
 
         public override Type StateType => typeof(AnimatedSpriteComponentState);
 
-        public float Bottom
-        {
-            get
-            {
-                return Owner.GetComponent<ITransformComponent>().Position.Y +
-                       (sprite.TextureRect.Height / 2);
-            }
-        }
+        /// <summary>
+        ///     Center of the Y axis of the sprite bounds in world coords.
+        /// </summary>
+        public float Bottom => Owner.GetComponent<ITransformComponent>().Position.Y + sprite.LocalAABB.Height / 2;
 
         public Box2 AverageAABB
         {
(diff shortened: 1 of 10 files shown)


SS14.Client/State/States/GameScreen.cs
```diff
@@ -635,7 +635,7 @@ .Y ascending
             // Check whether click is outside our 1.5 tile range
             float checkDistance = 1.5f * MapManager.TileSize;
             var dist = PlayerManager.ControlledEntity.GetComponent<ITransformComponent>().Position - entToClick.GetComponent<ITransformComponent>().Position;
-            if (dist.Length > checkDistance)
+            if (dist.LengthSquared > checkDistance * checkDistance)
```
clusterfack: Why this though

SS14.Client/GameObjects/Components/Renderable/AnimatedSpriteComponent.cs
```diff
@@ -45,7 +47,7 @@ public float Bottom
             get
             {
                 return Owner.GetComponent<ITransformComponent>().Position.Y +
-                       (sprite.AABB.Height / 2);
+                       (sprite.TextureRect.Height / 2);
```
PJB3005: Shouldn't this be the AABB because that's using world coordinates?

SS14.Client/State/States/GameScreen.cs
PJB3005: This is faster because it doesn't have to square root inside the `Length` getter.

clusterfack: Meh I let the compiler handle trivialities like that

PJB3005: I highly doubt the .NET AOT is smart enough to optimize this in any way.

Hell even C compilers would have a hard time getting this.

ZoldorfTheWizard (approved):

PR #414, opened 2017-09-13
Multiple Map System

clusterfack:
I've been working on and off on this system for awhile. Here are its primary functionalities

- Implements a map which holds all grids, similar to what the mapmanager does now, the map has its own index and can be retrieved from the map manager through this index
- Implements a coordinate system which is heavily tied into the map and transform systems, this stores the vector position, the grid, and the map, so that information can be passed around as a value without being lost and ambiguously retrieved
- Things on different zlevels do not render at once, they render based on the current viewable zlevel which right now is simply the zlevel the player is on
- The coordinate system move also allowed movement towards a local positioning system based on the current grid you are on, instead of always being based on world coordinates, this is still a relatively untested feature as we need more QoL tools to create grids and play around with them ingame

There are a few known bugs but I've fixed all the ones that I can currently fix until clicking/deletion are fixed. I need those to test some of the remaining issues primarily
~~- The zlevel rendering thing doesn't allow rendering off the grid number 1, but since you cant even leave grid 1 without deleting the walls its just a hassle to test until clicking gets fixed so I'm unsure why exactly this is occurring~~ [Fixed]

There are a few things that should be added in the future to finish fleshing out the system but aren't necessarily for a feature complete PR here
- The default grid having set tile overriden so that it doesn't do anything, so that the tiles on the default grid always remain the default tile and that it only returns tiles which aren't completely covered by another grid on gettilesintersecting
- Tools to create and manage grids, create and switch between maps, and etc ingame
- Entity loading that takes into account the map/grid system, most likely this would end up being what we would call a 'map saving/loading' system when that starts getting done


clusterfack:
Some of my testing is going to rely on https://github.com/space-wizards/space-station-14/pull/413 so I'll wait until that is merged until seriously trying to fix the conflicts here

Silvertorch5:
@clusterfack merged #413 rip you

clusterfack:
Alright, that's all the bugs I know about. I can tell you the system looks and acts almost identically (and better in some cases) than the current one.

If anyone wants to review it first that's fine, or if they want me to explain some of the finer implementation details, other than that though requesting merge of the system change.

SS14.Shared/Map/Coordinates.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Threading.Tasks;
+
+namespace SS14.Shared.Map
+{
+    [Serializable]
+    public struct LocalCoordinates
+    {
+        public readonly int GridID;
+        public readonly int MapID;
+        public Vector2 Position;
```
PJB3005, line 17: This should probably be readonly.

clusterfack: Ah yeah I missed that when converting it from a class, good catch it wouldn't work properly to set on a struct

SS14.Shared/Map/Coordinates.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        public float X
+        {
+            get => Position.X;
+            set => new LocalCoordinates(value, Position.Y, GridID, MapID);
```
PJB3005, line 22: This doesn't work. I'd be amazed if your IDE doesn't throw a warning for unused expression.

PR #419, opened 2017-09-15
Fix git_helper.py on Python 3.5

PJB3005:


psykzz:
This seems like `path.join` could help more than pathlib

PR #420, opened 2017-09-17
Refactor and clean up containers.

PJB3005:
Containers are no longer arbitrarily bound to a single component, they use string IDs instead.

The API to interact with containers has been documented with doc comments.

Container creation now uses a static method.

Containers are sealed instead of abstract. Custom logic doesn't go on containers.

Fixed bugs so that containers work now.

Added unit tests for containers too.


PJB3005:
Also obligatory 🔥

ZoldorfTheWizard (approved):

PR #422, opened 2017-09-25
Add more whitelisted types.

PJB3005:


PJB3005:
Screw it I aint got all day and those MacOS builds are gonna take 10 minutes easily because America is waking up.

PR #435, opened 2017-10-11
Lighting Performace Changes

Acruid:
Two simple changes to make the lighting render loop run faster.

### DIFF
--- a/SS14.Client/State/States/GameScreen.cs
+++ b/SS14.Client/State/States/GameScreen.cs
@@ -992,10 +992,6 @@ namespace SS14.Client.State.States
                 source.Blit(0, 0, source.Width, source.Height);
                 screenShadows.EndDrawing();
             }
-
-            Image texunflipx = screenShadows.Texture.CopyToImage();
-            texunflipx.FlipVertically();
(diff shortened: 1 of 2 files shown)


SS14.Client/State/States/GameScreen.cs
```diff
@@ -874,8 +874,6 @@ private void RenderLightsIntoMap(IEnumerable<ILight> lights)
             }
 
             //Step 2 - Set up the render targets for the composite lighting.
-            screenShadows.Clear(Color.Black);
```
ZoldorfTheWizard, line 877: but but muh r e d u n d a n c y

(nice catch)

ZoldorfTheWizard (approved):

SS14.Client/State/States/GameScreen.cs
Acruid: Gotta overwrite that texture with 0's multiple times to make sure everything is erased ;)

theOperand:
Silver’s Boot And Nuke

PJB3005 (approved):

PR #439, opened 2017-10-20
More Batch in This Patch

Serkket:
More a u t o m a t i o n

### DIFF
--- a/Setup.bat
+++ b/Setup.bat
@@ -1,0 +1,5 @@
+cd %CD%
+python RUN_THIS.py
+cd %CD%\Resources
+python buildResourcePack.py
+pause


ghost:
there must be an issue with default python being 2
python3 is safe

PJB3005 (changes requested):
Use `py -3` instead.

psykzz:
Instead of `py -3` add the shebang `#!/usr/local/bin/python python3` which should work universally.

`py` is a windows python launcher, and will be able to read and understand the shebang and run it via python3 if its installed.

PJB3005:
That still won't work if you invoke python directly, like this PR is doing.

psykzz:
Ah your right, for windows to use the shebang you need to use `py` so might as well just do `py -3`

ZoldorfTheWizard:
Please update and create a new pull request.

PR #457, opened 2017-10-23
Catch port in use exception

### LINKED ISSUE #454: Server crashes if port is already in use.
To reproduce, open two instances of the server exe. The second one will crash when it tries to bind to the same port as the first.

The error needs to be caught and displayed in the console. The program then needs to be gracefully shut down.

psykzz:
Fixes #454


PJB3005:
Does this actually *tell* the user that the reason it can't start is the blocked port, though?

psykzz:
Yeah my bad i didn't push the 2nd part. @PJB3005

PJB3005 (approved):

ZoldorfTheWizard (approved):

PR #463, opened 2017-10-25
UI Rework

Acruid:
This is the first phase in https://github.com/space-wizards/space-station-14/issues/459. Controls now inherit most of their functionality from parent Controls, so a lot less code/field duplication. Controls are now laid out in a tree structure instead of an array, fixing a lot of the previous relative positioning and event consuming issues. Child controls can easily be placed relative to their parent controls in the layout, so no more having to do confusing absolute screen space calculations with control width and position.

This fixes the focus system. It works pretty much as you expect, clicking on certain controls gives them keyboard focus, clicking on the game screen or pressing escape removes it. As an added bonus, the GameScreen now controls the KeyBindingManager, so you can easily block game input with the UI input events.

**KNOWN ISSUES**
Many of the custom controls manage network messages and state of the game. This needs to be refactored out into a BaseClient class, along with most of the code in the GameScreen. Because of this, I did not spend too much time testing them because they are going to be heavily modified.


PJB3005:
Is it actually good design to make things be children in the tree just to do relative positioning? Doesn't the tree influence focus and such too?

PR #480, opened 2017-11-18
Fix lighting on MacOS.

PJB3005:
Also adds rendering debug console commands.

### DIFF
--- a/SS14.Client/State/States/GameScreen.cs
+++ b/SS14.Client/State/States/GameScreen.cs
@@ -39,6 +39,7 @@ using Vector2i = SS14.Shared.Maths.Vector2i;
 using SS14.Client.UserInterface;
 using SS14.Client.UserInterface.Controls;
 using SS14.Client.UserInterface.CustomControls;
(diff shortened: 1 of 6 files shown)


SS14.Client/State/States/GameScreen.cs
```diff
@@ -825,10 +826,19 @@ private void RenderLightsIntoMap(IEnumerable<ILight> lights)
...
+            #if !MACOS
             source.Clear(Color.Black);
-
-            RenderImage desto = shadowIntermediate;
+            #else
+            // For some insane reason, SFML does not clear the texture on MacOS.
+            // unless you call CopyToImage() on it, which we can't due to performance.
+            // This works though!
+            source.BeginDrawing();
+            CluwneLib.drawRectangle(0, 0, source.Width, source.Height, Color.Black);
```
PJB3005: This is the meat of the fix.

🙏 SFML.

ZoldorfTheWizard (approved):
Sfml wild ride

PR #482, opened 2017-12-13
ClientSession/BaseClient Rework

Acruid:
The main parts of this PR are the new BaseClient class, and the reworked the client PlayerManager. The BaseClient class is similar to the BaseServer, where it controls the overall state of the client. In the future when the GameScreen god class on the client gets refactored, the BaseClient class will be very useful. The client PlayerManager now keeps a synced collection of clients connected to the server, and better manages   data and events about the local client with its LocalPlayer class. Because of these changes, multiple clients can now properly connect and be identified on a server, and the game lobby can now properly show a list of clients in the server that is updated in realtime.

**Potential Improvements**
Right now the client syncing between PlayerManagers is done with both the MsgPlayerList packet and GameStates. This is a candidate for refactoring in the future. The PlayerJoinedServer/Lobby/Game functions in BaseClient are good candidates for events, and sound very useful to the content assembly. The MsgServerInfoReq and MsgServerInfo packets that swap info between client and server when building the session should be refactored into the new Console/Chat/ConCmd/CVar system I have been promising for six months.

**Known Issues**
The scrollbar in the PlayerListTab of the lobby screen has a position bug. This will be fixed in https://github.com/space-wizards/space-station-14/issues/459.

The BaseClient switches from the LobbyMenu to the GameScreen immediately after the joingame verb is sent, and before the entities and tilemaps are sent to the client (tilemap/entity info is not sent to clients while in the lobby). The result is that the first frame drawn is usually of empty space and half of the entities. The screen is only redrawn when the player moves, so the screen will stay like this for awhile. It looks terrible. Whoever does the GameScreen god class refactor may want to come up with a better solution.


SS14.Client/Player/LocalPlayer.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            ControlledEntity.GetComponent<ITransformComponent>().OnMove += OnPlayerMoved;
+
+            EntityAttached?.Invoke(this, EventArgs.Empty);
+        }
+
+        /// <summary>
+        ///     Detaches the client from an entity.
+        /// </summary>
+        public void DetatchEntity()
```
PJB3005: You typo'd this method's name here. `DetachEntity()` instead of `DetatchEntity()`

SS14.Client/BaseClient.cs
```diff
@@ -0,0 +1,239 @@
...
+        private void OnConnected(object sender, NetChannelArgs args)
+        {
+            // request base info about the server
+            var msgInfo = _net.CreateNetMessage<MsgServerInfoReq>();
+            msgInfo.PlayerName = "Joe Hello";
+            _net.ClientSendMessage(msgInfo, NetDeliveryMethod.ReliableOrdered);
+        }
+
+        /// <summary>
+        ///     Player session is fully built, player is an active member of the server. Player is prepaired to start
```
psykzz: I think you mean `prepared`

ZoldorfTheWizard (approved):

PR #485, opened 2018-01-07
Add timers.

### LINKED ISSUE #456: System for timed callbacks (sleeping)
Because everything runs synchronously in a single thread, sleeping ruins everything. Thus I propose we code a scheduling system to allow you to do a delayed callback. Similar to TG's timers I believe.

wixoaGit:
Adds the ability to create timers. This resolves #456.

An example for a 2 second repeating timer:
```cs
Timer timer = new Timer(2000, true, () => Logger.Log("Timer Fired."));
timerManager.AddTimer(timer);
```

### DIFF
--- a/SS14.Shared/Timers/Timer.cs
+++ b/SS14.Shared/Timers/Timer.cs
@@ -1,0 +1,66 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SS14.Shared.Timers
+{
+    public class Timer
+    {
+        /// <summary>
+        /// Counts the time (in milliseconds) before firing again.
+        /// </summary>
+        private int _timeCounter { get; set; }
+
+        /// <summary>
+        /// Time (in milliseconds) between firings.
+        /// </summary>
+        public int Time { get; set; }
+
+        /// <summary>
+        /// Whether or not this timer should repeat.
+        /// </summary>
+        public bool IsRepeating { get; set; }
+
+        /// <summary>
+        /// Whether or not this timer is still running.
+        /// </summary>
+        public bool IsActive { get; set; }
+
+        /// <summary>
+        /// Called when the timer is fired.
+        /// </summary>
+        public event EventHandler Fired;
+
+        public Timer(int time, bool isRepeating)
+        {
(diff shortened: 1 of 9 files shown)


SS14.Shared/Timers/Timer.cs
```diff
@@ -0,0 +1,66 @@
...
+            IsRepeating = isRepeating;
+            _timeCounter = Time;
+            IsActive = true;
+        }
+
+        public void Update(float frameTime)
+        {
+            if (IsActive)
+            {
+                _timeCounter -= (int)Math.Floor(frameTime * 1000);
```
RemieRichards: you don't need the ```Math.Floor()```, since the cast to int will already floor the value.

SS14.Shared/Timers/Timer.cs
```diff
@@ -0,0 +1,66 @@
...
+
+                if (_timeCounter <= 0)
+                {
+                    Fired(this, new EventArgs());
```
RemieRichards: this seems iffy, why not use an Action or a Task as a backing? or give users the choice?

Cyberboss: An Action makes much more sense imo. If all you're passing is sender (always a timer manager or some test mock) and standard EventArgs (which contain literally nothing) no one is ever going to use those args.

I don't think a Task is a good idea to pass in. IDK how you guys handle threading but if you want timers to be Task based the way to do it would be to call the passed Action with Task.Run()

SS14.Shared/Timers/Timer.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
Cyberboss: All you need here is System

SS14.Shared/Timers/Timer.cs
Cyberboss: After reading the issue, Tasks are pretty much out of the question for what you want to accomplish

wixoaGit: I'll change it to an action, then.

ZoldorfTheWizard (approved):

PR #486, opened 2018-01-12
Sandbox Gamemode

Acruid:
Adds new Sandbox game dlls to the solution. These are required for the engine to work without the SS14-content repo. The main focus was to get basic player state events working on the server, and to get basic multi-map functionality working.

The `PlayerJoined*` events on the server are now based on the `IPlayerSession.PlayerState` enum instead of callbacks from the network handlers. The `SandBox.Server` project is a good example of haw to set up an event to manage players on the server.

It is now optional to load the demo grid, the engine won't crash if you chose not to. You can change the default spawn point to a specific mapID and location. If you do not load any grids/maps, the player will be placed in the default Nullspace map at (0, 0). You can use the `addmap` console command to create a new mapId, and then use the new `tp` command to teleport to it. Spawning tiles and entities should work as you expect. New `MapID` and `GridID` types were added for additional type safety.

This PR breaks lighting on maps. Map/Light rendering needs work before it can handle multiple maps.


ZoldorfTheWizard (approved):
10 years passed

Issue #448, opened 2017-10-21
Disconnecting, then reconnecting causes client to complaint about the CVar player.name being registered twice.


PJB3005:

Acruid:
This is because the CVar registration for `player.name` is inside the `SS14.Client.State.States.GameScreen.Startup()` function, which is called every time the state gets focus. It needs to be moved to the `GameScreen.InitializeGUI()` function which is only called once.

Acruid:
Fixed in https://github.com/space-wizards/space-station-14/pull/488.

PR #488, opened 2018-01-16
Embrace NetMessages

Acruid:
No more `NetMsg.ID`, no more `NetDeliveryMethod`. ALL networking has been converted to the new system using net messages. The Lidgren assembly reference has been removed from everything except `SS14.Shared`, and only `SS14.Shared.Network.*` should be referencing it.

`LocalCoordinates.IsValidLocation()` was added to test if a `LocalCoordinate` is usable. Issues came up where the client would receive entity positions before maps were registered, causing the map and grid lookups to throw an exception. This function can be used to prevent that.

This fixes https://github.com/space-wizards/space-station-14/issues/448.


optimumtact:
You know if you put fixes #448 in the PR body it will autoclose the issue when the pr is merged right?

SS14.Client.Graphics/CluwneLib.cs
```diff
@@ -435,9 +435,11 @@ public static Vector2 TileToWorld(Vector2 point)
         /// </summary>
         public static LocalCoordinates ScreenToCoordinates(ScreenCoordinates point)
         {
+            // world pos in current map
             var pos = (point.Position - Window.Viewport.Size / 2) / Camera.PixelsPerMeter + Camera.Position;
-            var grid = IoCManager.Resolve<IMapManager>().GetMap(point.MapID).FindGridAt(pos);
-            return new LocalCoordinates(pos, grid);
+
+            // world coords on map
+            return new LocalCoordinates(pos, GridId.DefaultGrid, point.MapID);
```
clusterfack: Why the default grid rather than the particular grid it is actually on?

Acruid: Because the map/grid is not guaranteed to exist at that point. The MapID is taken off the transform on the local Session.LocalPlayer.ControlledEntity in GameScreen.Update. ControlledEntity can be set before the maps have been received. The GetMap call will crash the game, so you won't be able to find the grids.

clusterfack: Well that leads to a separate question, how to make sure that if a grid gets added or removed the local coordinates pointing to that position or grid don't get updated.

PJB3005:
@Acruid conflicts.

ZoldorfTheWizard:
Fix the conflicts pls

ZoldorfTheWizard:
Comment

PR #489, opened 2018-01-17
Sprite Offset Property

Acruid:
Adds a property to `SpriteComponent` to offset the location of sprite rendering in world units.

This fixes https://github.com/space-wizards/space-station-14/issues/451.


SS14.Server/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -4,7 +4,12 @@
 using SS14.Shared.Interfaces.GameObjects;
 using SS14.Shared.IoC;
 using System.Collections.Generic;
-using System; //TODO: Remove this when all NotImplementedExceptions are dealt with
+using System;
+using SS14.Shared.Maths;
+using SS14.Shared.Utility;
+using YamlDotNet.RepresentationModel;
+
+//TODO: Remove this when all NotImplementedExceptions are dealt with
```
Cyberboss, line 12: This applied only to the System using i think

PR #490, opened 2018-01-18
Updates content compatibility

clusterfack:
Changes way content loads instead of using config, attempts content then loads sandbox instead
Fixes some compiler warnings
Adds a field needed for content player mob


Silvertorch5:
Subscribe

PR #499, opened 2018-01-20
Content Build Targets

Acruid:
Changed projects so that Sandbox content is built in its own folder, and then copied to `./Resources/Assemblies` in the client/server bin folders. Hopefully this fixes Sonar.


PJB3005 (changes requested):
Put the `SS14.Sandbox.targets` in the `/MSBuild` folder.

PJB3005 (approved):

PR #501, opened 2018-01-20
Fixes appveyor clone

Cyberboss:
GitHub absolutely despises it when you shallow clone then fetch. See here: https://github.com/cocoapods/cocoapods/issues/4989


PJB3005:
Oh that's the thing I was looking for. Nice.

Issue #288, opened 2017-07-25
Cannot send anything but primitives over component network messages.


PJB3005:
Components can send messages to their counterparts over the wire. These messages are limited to what's serializable [here](https://github.com/space-wizards/space-station-14/blob/master/SS14.Client/GameObjects/ClientEntityNetworkManager.cs#L96), so using custom data structures is out of the question.

Of course, we should probably make things like structs possible. Problem is *how*?

Easiest would probably to use `NetSerializer` for it, with a custom network data type for types serialized like this. This might be slow and bloated however.

The other option is to require registration of the message types ahead of time, using a shared interface so that these custom messages can be understood on both ends. I think this might be the best solution. Any input @Acruid?

YotaXP:
Could put methods on the struct to serialize to or from an `IEnumerable<object>`.  Instead of sending the type information through the network, just have the deserialization code verify that all the primitive types are in the correct order.  Maybe include a `NetworkDataType.EndOfStruct` in the message to help catch edge cases.

PJB3005:
The problem is that then you need to have a decoding schema server side.

On second thought maybe we need that anyways the current system's a mess.

PJB3005:
We could fix this with `NetSerializer`. Would instantly solve all our trouble.

Acruid:
Resolved by #528, component messages now use NetSerializer.

PR #528, opened 2018-02-17
Component Messaging Rework

Acruid:
Component messaging system now sends classes instead of using `MessageType` Enum + object param array. It now works very similar to C# events.

Some component initialization was changed so that there is no point in a components life where it is not owned by an entity. Previously a component was not owned until Component.Initialize was called. Since components cannot change entities, `IEntity.AddComponent(IComponent)` was replaced with `IEntity.AddComponent<Component>()`

Oldcode UI Context system was removed, because it was broken and poorly designed.

ECS Entity class is now sealed so that nobody gets any ideas.


PJB3005 (approved):

Acruid:
This resolves #288.

PR #531, opened 2018-02-27
InputSystem Messages

Acruid:
This moves input message sending from the `KeyBindingInputComponent` to the ECS `InputSystem`. Also cleaned up a bunch of misc input messaging OldCode.

This PR is dedicated to @clusterfack.


SS14.Client/GameObjects/Components/Input/KeyBindingInputComponent.cs
```diff
@@ -1,142 +1,11 @@
...
-            //Delete false states from the dictionary so they don't get reprocessed and fuck up other stuff.
-            foreach (var state in _keyStates.ToList())
-            {
-                if (!state.Value)
-                    _keyStates.Remove(state.Key);
-                else
-                    SendMessage(new BoundKeyRepeatMsg(state.Key, BoundKeyState.Repeat));
-            }
-        }
+        // Does nothing client side.
```
PJB3005, line 9: In that case you can mark it as ignored in the component factory so it doesn't even get created.

PR #533, opened 2018-03-06
Directional Placement Manager Fix

clusterfack:
Makes directional sprites show up in the placement manager upon rotation

### DIFF
--- a/SS14.Client/Placement/PlacementManager.cs
+++ b/SS14.Client/Placement/PlacementManager.cs
@@ -139,6 +139,11 @@ namespace SS14.Client.Placement
                     Direction = Direction.North;
                     break;
             }
+
+            if (CurrentMode != null)
+            {
+                CurrentMode.SetSprite();
+            }
         }
 
         public void HandlePlacement()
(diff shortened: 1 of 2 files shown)


SS14.Client/Placement/PlacementManager.cs
```diff
@@ -139,6 +139,11 @@ public void Rotate()
                     Direction = Direction.North;
                     break;
             }
+
+            if (CurrentMode != null)
+            {
+                CurrentMode.SetSprite();
```
RemieRichards, line 145: How about ```CurrentMode?.SetSprite();```

clusterfack: It's the same thing

RemieRichards: yes, but 1 line instead of 4. It's stylistic, hence 'How about' not something more definite like, 'you should use'

clusterfack: I don't think we're going to reach a limit on our lines of code anytime soon

PR #534, opened 2018-03-08
Full interaction system

clusterfack:
Raises an event on click, regardless of whether we clicked on something or not. Sends it to the server to be parsed in the content interaction system.

### DIFF
--- a/SS14.Client/GameObjects/EntitySystems/InputSystem.cs
+++ b/SS14.Client/GameObjects/EntitySystems/InputSystem.cs
@@ -62,6 +62,7 @@ namespace SS14.Client.GameObjects.EntitySystems
                 return; //Don't repeat keys that are already down.
 
             SetKeyState(e.Function, true);
+
             var message = new BoundKeyChangedMessage(e.Function, e.FunctionState);
             RaiseEvent(message);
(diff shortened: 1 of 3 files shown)


SS14.Client/GameObjects/EntitySystems/InputSystem.cs
```diff
@@ -117,5 +118,11 @@ public virtual void UpdateKeys(float frameTime)
                     _keyStates.Remove(state.Key);
             }
         }
+
+        public virtual void RaiseClick(ClickEventMessage message)
+        {
+            //TODO: Check modifiers here and modify the click type of the message?
```
RemieRichards, line 124: Maybe ```ClickEventMessage``` could contain a bitfield of modifier keys? Since there'd only be Ctrl/Shift/Alt usually.

clusterfack: It seems we don't really store which modifier keys are being held down right now, I asked acruid because I thought this inputsystem might have such a feature but it doesn't. 

So I'm waiting for godot on that one since afaik godot handles that fairly well natively and I'll just pass it from the click later.

PR #537, opened 2018-03-12
Arbitrary Effects System

clusterfack:
This a system to make temporary clientside effects without the creation and syncing of full entities. 

You can define how you want the effect to look, move, and change over time by specifying a number of variables in the effect creation message. After creating one with the prerequisite variables you can pass it to the server effect entity system and it will send it to clients, and the client effect entity system will render them.

### DIFF
--- a/SS14.Client/GameObjects/EntitySystems/EffectSystem.cs
+++ b/SS14.Client/GameObjects/EntitySystems/EffectSystem.cs
@@ -1,0 +1,270 @@
+﻿using SS14.Client.Graphics;
+using SS14.Client.Graphics.Sprites;
+using SS14.Client.Interfaces.Resource;
+using SS14.Shared.GameObjects;
+using SS14.Shared.GameObjects.EntitySystemMessages;
+using SS14.Shared.GameObjects.System;
+using SS14.Shared.Interfaces.Network;
+using SS14.Shared.Interfaces.Timing;
+using SS14.Shared.IoC;
+using SS14.Shared.Log;
+using SS14.Shared.Map;
+using SS14.Shared.Maths;
+using System;
+using System.Collections.Generic;
+
+namespace SS14.Client.GameObjects
+{
+    public class EffectSystem : EntitySystem
+    {
+        private List<Effect> _Effects = new List<Effect>();
+        private TimeSpan lasttimeprocessed = TimeSpan.Zero;
+
+        public override void RegisterMessageTypes()
+        {
+            base.RegisterMessageTypes();
+
+            RegisterMessageType<EffectSystemMessage>();
+        }
+
+        public override void HandleNetMessage(INetChannel channel, EntitySystemMessage message)
+        {
+            base.HandleNetMessage(channel, message);
+
+            switch (message)
+            {
+                case EffectSystemMessage msg:
+                    CreateEffect(msg);
+                    break;
+            }
+        }
+
+        public void CreateEffect(EffectSystemMessage message)
+        {
+            var gametime = IoCManager.Resolve<IGameTiming>().CurTime;
+            if (gametime > message.DeathTime) //Did we already die in transit? That's pretty troubling isn't it
+            {
+                Logger.Warning(string.Format("Effect using sprite {0} died in transit to the client", message.EffectSprite), message);
+                return;
+            }
+
+            //Create effect from creation message
+            var effect = new Effect(message);
+
+
+            //Remove this
+            effect.Age = gametime;
+            effect.Deathtime = effect.Age + TimeSpan.FromSeconds(4);
+
+            //Age the effect through a single update to the previous update tick of the effect system
+            //effect.Update((float)((lasttimeprocessed - effect.Age).TotalSeconds));
+
+            _Effects.Add(effect);
+        }
+
+        public override void Update(float frameTime)
+        {
+            lasttimeprocessed = IoCManager.Resolve<IGameTiming>().CurTime;
+
+            for (int i = 0; i < _Effects.Count; i++)
+            {
+                var effect = _Effects[i];
+
+                //Update variables of the effect via its deltas
+                effect.Update(frameTime);
+
+                //These effects have died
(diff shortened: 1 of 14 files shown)


SS14.Client/Program.cs
```diff
@@ -62,6 +62,7 @@ private static void Main()
         {
             RegisterIoC();
             RegisterComponents();
+            IoCManager.Resolve<IEntitySystemManager>().Initialize();
```
Acruid: This should go somewhere inside `SS14.Client.GameController` with the rest of the initialization calls.

SS14.Shared/GameObjects/EntitySystemMessages/EffectMessage.cs
```diff
@@ -0,0 +1,95 @@
+﻿using SS14.Shared.Map;
+using SS14.Shared.Maths;
+using System;
+
+namespace SS14.Shared.GameObjects.EntitySystemMessages
+{
+    [Serializable]
+    public class EffectSystemMessage : EntitySystemMessage
```
Acruid, line 8: This is a large packet to be sending to all 200 clients for every visual effect created. Is there any way to make it smaller? Do we really need `Velocity`, `RadialVelocity`, and `TangentialVelocity` sent separately?

SS14.Server/GameObjects/EntitySystems/EffectSystem.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// Comparer that keeps the device dictionary sorted by powernet priority
+        /// </summary>
+        public class EffectMessageComparer : IComparer<EffectSystemMessage>
+        {
+            public int Compare(EffectSystemMessage x, EffectSystemMessage y)
+            {
+                int compare = y.DeathTime.CompareTo(x.DeathTime);
+
+                ////If the comparer returns 0 sortedset will believe it is a duplicate and return 0, so return 1 instead
+                //if (compare == 0 && !x.Equals(y))
```
Acruid: Commented out code needs to be removed. The doc comment summary for this class also refers to powernet priority?

SS14.Client/GameObjects/EntitySystems/EffectSystem.cs
```diff
@@ -0,0 +1,270 @@
...
+            //Create effect from creation message
+            var effect = new Effect(message);
+
+
+            //Remove this
+            effect.Age = gametime;
+            effect.Deathtime = effect.Age + TimeSpan.FromSeconds(4);
+
+            //Age the effect through a single update to the previous update tick of the effect system
+            //effect.Update((float)((lasttimeprocessed - effect.Age).TotalSeconds));
```
Acruid, line 60: No commented out code.

clusterfack: This is how its supposed to work, if the timing system worked the code a line above it would be removed

SS14.Server/GameObjects/EntitySystems/EffectSystem.cs
clusterfack: When I used the priority comparer in powernets having it return 0 would cause it to be considered a duplicate and removed so I was on the fence here about whether thats also necessary for this priority queue or not

SS14.Client/Program.cs
clusterfack: Okay

SS14.Shared/GameObjects/EntitySystemMessages/EffectMessage.cs
clusterfack: Sent separately as opposed to what? Do you save space putting them all in a vector or something?

Acruid: I was hoping some of these values could be derived client side to reduce the size of the message, but I don't think that will be possible since I realized we can have particles spiraling outwards while moving linearly through the world.

SS14.Client/GameObjects/EntitySystems/EffectSystem.cs
```diff
@@ -0,0 +1,270 @@
...
+
+            /// <summary>
+            /// Rate of change of effect's size change
+            /// </summary>
+            public float SizeDelta = 0f;
+
+            /// <summary>
+            /// Effect's current color
+            /// </summary>
+            public Vector4 Color = new Vector4(255,255,255,255);
```
PJB3005, line 177: Why is this being stored as a `Vector4` instead of a `Color`?

SS14.Client/GameObjects/EntitySystems/EffectSystem.cs
```diff
@@ -0,0 +1,270 @@
...
+
+            /// <summary>
+            /// Effect's tangential acceleration
+            /// </summary>
+            public float TangentialAcceleration = 0f;
+
+            /// <summary>
+            /// Effect's spin about its center in radians
+            /// </summary>
+            public float Rotation = 0f;
```
PJB3005, line 157: Use `Angle` maybe?

SS14.Client/GameObjects/EntitySystems/EffectSystem.cs
clusterfack: Because it can store wildly abnormal values

SS14.Client/GameObjects/EntitySystems/EffectSystem.cs
clusterfack: That's just bikeshedding my dude

PJB3005: How is that bikeshedding? We have a standardized type for storing angles, so use it. You're not gonna store a vector as a tuple `(float, float)` either are you?

clusterfack: I looked at this and its fairly pointless because its meant to be replicated on the client, so silly storing it in Angle form on the client as its a struct is silly since it gets continuously remade in an update() function from the system

PR #552, opened 2018-04-11
RSI support

PJB3005:
**Not implemented into `SpriteComponent` yet because that turd needs a refactor anyways.**

The loader itself works.

Also includes a refactor of the resource cache and VFS:
* `ResourcePath` is an immutable class used for the handling and standardization of resource paths.
* `BaseResource` is now given a path to load and that's it. It's in charge of opening file streams and shit.


RemieRichards:
My Schema! Wooooo!

PR #572, opened 2018-04-20
Detaches from parent when addtoscreen

clusterfack:
Like it says, detaches from the current parent before adding to a new screen

### DIFF
--- a/SS14.Client/UserInterface/CustomControls/SS14Window.cs
+++ b/SS14.Client/UserInterface/CustomControls/SS14Window.cs
@@ -287,6 +287,10 @@ namespace SS14.Client.UserInterface.CustomControls
 
         public void AddToScreen()
         {
+            if(Parent != null)
+            {
+                Parent.RemoveChild(this);
+            }
             UserInterfaceManager.WindowRoot.AddChild(this);
         }
 


SS14.Client/UserInterface/CustomControls/SS14Window.cs
```diff
@@ -287,6 +287,10 @@ public bool IsAtFront()
 
         public void AddToScreen()
         {
+            if(Parent != null)
```
PJB3005, line 290: Would probably be better to just make the function no-op instead.

clusterfack: what

PJB3005: Make the function do nothing if it already has a parent. That or make it only detach if the parent isn't the window root.

PJB3005: Actually I'd say I'm in favor of just making this throw an exception. Why are you calling it multiple times?

clusterfack: Why would you want this to throw an exception

PJB3005:
You're absolutely never supposed to call this twice, why are you?

If you want it for that "remove from window root completely" idea, make a
removefromscreen too and then I'll be ok with making it silently quit.

PJB3005:
> If you wanted that to work you'd have to either initialize and hide the screen for everything immediately

This is exactly how I designed the system to work. That's why I said that if you want to change it, you should add a `RemoveFromScreen()` and make hiding the window use it.

Acruid:
It has been over 2 months since last activity in this PR, it appears to be abandoned. Closing.

PR #580, opened 2018-05-06
Placement Systems

clusterfack:
~~This PR is about 90% completed, it still has two bugs to fix with the placement system, the collision check and finding out why tile spawning is misaligned by like half a tile.~~ This PR is now complete.

In this PR we add alternate modes to spawn placement, line placement and grid placement

Line placement: https://cdn.discordapp.com/attachments/310555209753690112/442622598753550346/2018-05-06_04-43-49.mp4

Grid placement:
https://cdn.discordapp.com/attachments/310555209753690112/442625755499986944/2018-05-06_04-56-24.mp4

You use shift click to start a line placement, and after stretching out the line to where you want you click to spawn the items or right click to break out of the placement mode. The grid placement is the same, except it uses control click as its modifier.

This reworks a lot of the placement modes to use some simpler code, namely separating out alignment code from valid positioning code as well as removing some now unnecessary vars such as MouseScreen. And it adds a lot of comments to the placement system to make it much more understandable.

### Going forward

This will make mapping large areas of the station much easier now, creating long lines of walls or wires very rapidly with little chance for misclick


SS14.Client/Placement/Modes/AlignTileAny.cs
```diff
@@ -5,34 +5,39 @@ namespace SS14.Client.Placement.Modes
...
-                    CurrentTile.Y + tileSize / 2,
+                MouseCoords = new LocalCoordinates(CurrentTile.X + tileSize,
+                    CurrentTile.Y + tileSize,
                     MouseCoords.Grid);
             }
             else
             {
-                MouseCoords = new LocalCoordinates(CurrentTile.X + tileSize / 2 + pManager.CurrentPrototype.PlacementOffset.X,
-                    CurrentTile.Y + tileSize / 2 + pManager.CurrentPrototype.PlacementOffset.Y,
+                MouseCoords = new LocalCoordinates(CurrentTile.X + tileSize * 1.5f + pManager.CurrentPrototype.PlacementOffset.X,
```
RemieRichards: 1.5? That smells wrong.

clusterfack: I literally said in the OP that I hadn't fixed the alignment issues

RemieRichards: Yes, which got me looking for weirdness in your coordinates.
But fuck me right?

clusterfack: You look at something I had changed, when it wasn't working in the first place otherwise why the fuck would I have changed it. Which is rather dumb of you.

clusterfack: And it turns out that *1.5f is actually closer to the actual answer than anything because the reason this was broken was due to the fact that tileSize is a ushort and it was dividing it by 2 which was a broken operation

clusterfack:
Alright I fixed this up and got it working. Placement modes now align properly, they were because tilesize was dividing a ushort and just truncating the value instead of converting to float first.

Additionally now when a prototype is loaded into the placement manager it uses the yaml deserializer to get the aabb value for the bounding box if it has collision and bounding boxes, otherwise it just creates an empty point box to test collision checks.

Additionally I changed the placement system to use much more vibrant and clear placement ghosts by making the placement colors sharply increase the green and red much more than the rest. And by making the placement ghosts render above everything else and use a material shader that ignores lighting. Combining all these things placement ghosts are much clearer to use than they ever were before.

Issue #558, opened 2018-04-13
Abrupt client disconnect memory leaks server.


PJB3005:
The server doesn't time out clients on debug builds, to prevent disconnects while debugging (useful).

Problem is that if the client crashes and never tells the server it's disconnected, the server will keep piling on entity state deltas for the client, causing it to basically memory leak slowly. Only fix is a restart.

Acruid:
This is not really a "leak", as the memory is not being lost. The client stops acknowledging the game state deltas, so the server has to keep a history of every state that is sent since the last ack'd state. The main problem is the client is not timing out (because debug), and there is no coded limit to how much state history the server holds for a client since the last ack.

The 'proper' solution is to set a limit of how large the backlog of states can get. This way if the client stops ack'ing states for a few seconds, the server will just stop creating delta states and flush the old ones. If the client ever starts responding again the server can just start fresh with a full state and generate deltas from there. This solution will also help prevent abuse where client stays connected, but does not ack states.

A short term work-around is to use the server `kick` command on the timed out client.

PJB3005:
This still happens even after #586, though this time I suspect it's just net messages piling up and Lidgren waiting for them to be acked.

It's much less significant though.

Acruid:
It needs to be verified that the client cannot DOS the server by not sending state ACKs. See the second part of my response above for a way to resolve this issue.

captkirk88:
Couldn't you also solve that with a good keep alive message every 5 seconds (as a example)?

If there is no keep alive, disconnect the client since they are ded.

I think you guys are using Lidgren still, if so, it has that ability you just have to enable it.

Acruid:
The problem is that the client stops acknowledging the game states every tick, so the server has to store every state for every tick between the last ack'd state and the current tick. Currently there is no coded limit to how large the buffer can be, so the client can DOS the server by not ack'ing states and let millions of states pile up. This is unrelated to lidgren's keep alive messages, which are disabled atm for debug purposes, but will be enabled for releases.

captkirk88:
I see your point.  I was going with the assumption that if the client was disconnected and you were disposing of the buffer related to that client when the disconnect event was fired (as a result of not receiving a keep alive), then there would be no buffer to worry about because it was disposed of when the server discovered the client was inactive.

PJB3005:
I'm gonna drop the priority because we're not all too worried about DOS attacks yet. We're not going for stability yet.

Acruid:
[...]

PR #586, opened 2018-05-10
Game State Refactor

### LINKED ISSUE #505: BSDiff removal and replacement
As it was said in the chat, usage of BSDiff is a huge hack, and should not be a thing. Proposed solution was to mark things "dirty".

Something something issue to keep track of this?

PJB3005:
Fixes #505

Hoo boy.

Removes the usage of bsdiff for game states, instead using a proper dirty flag system. Performance is much better now although bandwidth usage is slightly higher.

Helps a *lot* with #558 but it's not fixed yet. I suspect it's just Lidgren trying to ack messages now.

TODO:

* ~~Deletions of entities and components~~ Done.
* ~~Optimize bandwidth usage more?~~ Not doing this.
* ~~Optimize player states. They're not using a dirty system yet.~~ Not doing this.


PJB3005:
@Acruid this is mostly done. I haven't removed player state syncing though because the client needs it to connect for some reason. If that's fine go ahead and merge it.

PR #620, opened 2018-07-24
ECS - ComponentManager Restructure

### LINKED ISSUE #618: ECS - Component Storage Restructure
Restructuring the ECS system for performance.
- [x] Switch from storing the components on each entity to a central database inside of ComponentManager.
- [x] Better deferred component remove.
- [x] Turn the IEntity component functions into proxies of the ComponentManager functions.
- [x] Remove Update() base function from Components. If components need to update, an EntitySystem needs to do it.

Acruid:
* Switch from storing the components on each entity to a central database inside of ComponentManager.
* Better deferred component remove.
* Turn the IEntity component functions into proxies of the ComponentManager functions.
* Remove Update() base function from Components. If components need to update, an EntitySystem needs to do it.

Resolves #618.


SS14.Server/Maps/MapLoader.cs
```diff
@@ -253,7 +253,7 @@ private void LoadEntNode(IMap map, GridId? gridId, YamlSequenceNode entNode)
 
                 try
                 {
-                    var entity = _entityMan.SpawnEntity(protoName);
+                    var entity = _entityMan.CreateEntity(protoName);
```
Acruid, line 256: @PJB3005 Don't forget to change this in your branch.

SS14.Client/GameObjects/EntitySystems/AppearanceSystem.cs
```diff
@@ -17,16 +18,32 @@ sealed class AppearanceSystem : EntitySystem
         public AppearanceSystem()
         {
             IoCManager.InjectDependencies(this);
+
+            EntityQuery = new ComponentEntityQuery
+            {
+                OneSet = new List<Type>
+                {
+                    typeof(AppearanceComponent),
+                    typeof(AppearanceTestComponent),
```
PJB3005: `AppearanceTest` is a complete separate thing from `Appearance`. It just sets a test key on/off periodically.

It shouldn't be handled by this system, that or make it a separate query that gets ran before. Right now the way you've got that trygetcomponent set up it'll break and never be ran.

SS14.Client/GameObjects/EntitySystems/SpriteSystem.cs
```diff
@@ -13,15 +16,23 @@ public class SpriteSystem : EntitySystem
         public SpriteSystem()
         {
             IoCManager.InjectDependencies(this);
+
+            EntityQuery = new ComponentEntityQuery
+            {
+                OneSet = new List<Type>
+                {
+                    typeof(ISpriteComponent),
```
PJB3005: Can't this use `TypeEntityQuery` too?

SS14.Shared/GameObjects/Component.cs
```diff
@@ -46,30 +47,70 @@ public bool NetSyncEnabled
...
         }
 
         /// <summary>
         ///     Called when the component gets added to an entity.
         /// </summary>
-        public virtual void OnAdd() { }
+        public virtual void OnAdd()
+        {
+            if(Initialized)
+                throw new InvalidOperationException("Cannot Add an Initialized entity!");
```
PJB3005: You mean component instead of entity here?

PJB3005 (changes requested):

PJB3005 (approved):

PR #633, opened 2018-07-30
NetManager Shutdown Race Fix

### LINKED ISSUE #628: Race condition on shutting down client while connected.
Sometimes if you close the client while connected (with the red x), the client will crash. This is due to an exception inside Lidgren's threaded code because it wasn't fully shut down correctly, so when it does it throws an exception which is unhandled and causes a hard crash.

Two solutions:

* Wrap a giant try catch around the offending code.
* Wait for lidgren's thread to shut down before continuing with the shutdown of the main client.

Acruid:
NetManager now blocks the main thread and waits for the network thread to shut down.

This SHOULD resolve #628, but i can't prove it.

### DIFF
--- a/SS14.Shared/Network/NetManager.cs
+++ b/SS14.Shared/Network/NetManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Lidgren.Network;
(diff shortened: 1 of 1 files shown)


PJB3005 (commented):
Modifying Lidgren so we can wait on the thread to shut down properly might be a better idea than this.

PR #661, opened 2018-08-20
Move more Player things into content.

PJB3005:
Spawning of players is now 100% up to content.
IPlayerData is now purely for content.


RemieRichards:
remove sandbox

PR #676, opened 2018-08-30
Entity-bound UI system.

PJB3005:
Keeps track of which clients have a GUI open and such.


RemieRichards:
Holy heck sweetpea, you're on fire!

PR #681, opened 2018-09-03
Client name change command

### LINKED ISSUE #660: Console command to change client username before connecting, without modifying config file.
#659 added a proper session maintaining system. It lets the server auto resolve usernames.

It should be possible that the username that will be requested from the server is changed with a console command, so you can more precisely decide which session to re-join.

Code for starting the connection is in `MainMenu.cs`

Centronias:
Adds member field to BaseClient to hold onto override name string. Figured that was the closest place to the usage, so why not stick it there.
On connecting to the server, the client will prefer to use that name, if it's set, over the name from the configuration.

Adds client console command to set that field.

Fixes #660

### DIFF
--- a/SS14.Client/Console/Commands/Debug.cs
+++ b/SS14.Client/Console/Commands/Debug.cs
@@ -19,8 +19,6 @@ using System.Globalization;
 using SS14.Shared.Interfaces.Map;
 using SS14.Shared.GameObjects.Components.Transform;
 using System.Linq;
-using SS14.Client.Interfaces.GameObjects;
-using SS14.Shared.Interfaces.Configuration;
 
 namespace SS14.Client.Console.Commands
 {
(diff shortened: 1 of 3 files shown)


SS14.Client/Console/Commands/Debug.cs
```diff
@@ -292,4 +291,30 @@ public bool Execute(IDebugConsole console, params string[] args)
...
+    {
+        public string Command => "overrideplayername";
+        public string Description => "Changes the name used when attempting to connect to the server.";
+        public string Help => Command + " <name>";
+
+        public bool Execute(IDebugConsole console, params string[] args)
+        {
+            var client = IoCManager.Resolve<IBaseClient>();
+
+            if (client is BaseClient bc)
```
PJB3005: In that case, just add it to `IBaseClient` too.

SS14.Client/Console/Commands/Debug.cs
```diff
@@ -292,4 +291,30 @@ public bool Execute(IDebugConsole console, params string[] args)
...
+            {
+                var oldName = bc.PlayerNameForDebug;
```
PJB3005: You're not using this var anywhere, are you?

PJB3005 (changes requested):

PJB3005 (approved):

PR #682, opened 2018-09-08
Adds a var to forcibly show a godot transform.

RemieRichards:


SS14.Client/GameObjects/Components/Transform/GodotTransformComponent.cs
```diff
@@ -15,6 +15,8 @@ internal class GodotTransformComponent : TransformComponent, IGodotTransformComp
 
         IGodotTransformComponent IGodotTransformComponent.Parent => (IGodotTransformComponent)Parent;
 
+        public bool ForceVisible = false;
```
PJB3005: 1. This should be made a property, not a field (and when you do, it'd be nice if the setter would run `UpdateSceneVisibility` again)
2. I would prefer it if this were something like `VisibleWhileParented`
3. Might be a good idea to put it on `ITransformComponent`, because it's not accessible from content as `GodotTransformComponent` is an internal class.

PJB3005 (changes requested):

SS14.Client/GameObjects/Components/Transform/GodotTransformComponent.cs
RemieRichards: Okily dokily. Give me a moment.

clusterfack:
I cant wait to add a variable ForceForceInvisible to force something to be invisible 😏

SS14.Shared/Interfaces/GameObjects/Components/ITransformComponent.cs
```diff
@@ -63,6 +63,11 @@ public interface ITransformComponent : IComponent
         /// </summary>
         bool IsMapTransform { get; }
 
+        /// <summary>
+        /// Whether or not this entity is visible while parented to another entity.
+        /// </summary>
+        bool VisibleWhileParented { get; }
```
PJB3005: This isn't `set;` so you still can't actually modify this property.

RemieRichards: Doh.

PJB3005 (approved):

PR #690, opened 2018-09-12
Project file maintenance, C# 7.2

PJB3005:
1. Updates all the NuGet packages (except CommandLineParser because they ruined their API)
2. Makes all projects use C# 7.2 explicitly. (not latest)
3. Use some C# 7.2 features like readonly structs and default literals.


SS14.Client/SS14.Client.csproj
```diff
@@ -57,20 +58,29 @@
   </PropertyGroup>
   <Import Project="..\MSBuild\SS14.DefineConstants.targets" />
   <ItemGroup>
+    <Reference Include="ICSharpCode.SharpZipLib, Version=1.0.0.999, Culture=neutral, PublicKeyToken=1b03e6acf1164f73">
```
Cyberboss: oof, upgrade to [PackageReferences](https://docs.microsoft.com/en-us/nuget/reference/migrate-packages-config-to-package-reference) my dude

PJB3005: Do those even work for old style .NET Framework projects?

PR #696, opened 2018-10-25
Client IPlayerSession Interface.

Acruid:
Most of this work was already done in the InputSystem rework.

Resolves https://github.com/space-wizards/space-station-14/issues/617.


SS14.Shared/Players/IBaseSession.cs
```diff
@@ -7,6 +7,9 @@ namespace SS14.Shared.Players
     /// </summary>
     public interface IBaseSession
     {
+        /// <summary>
+        ///     The GUID of this session.
+        /// </summary>
         NetSessionId SessionId { get; }
```
PJB3005, line 13: Uh. A `NetSessionId` isn't a [GUID](https://en.wikipedia.org/wiki/Universally_unique_identifier).

PJB3005 (approved):

PR #700, opened 2018-10-29
Adds exception handling for the main loop.

### LINKED ISSUE #596: Add exception handling to common events.
Right now, a single exception and the entire server dies. At least you can't sneeze it to death I guess.

Anyways, we need multiple levels of exception catching and logging. Something in the main loop as a last resort for one, then try catches around events such as input should all be nice.

Client could probably use some too. It doesn't crash but it can make it flat miss an entire update tick.

ShiftyRail:
Fixes #596. I think.

I have no idea if this is correct or not. Also, I don't really know how to send the return value in a meaningful way.


SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SS14.Shared.Exceptions
+{
+    class RuntimeLog
```
PJB3005: I'd just go ahead and make this class public to be honest.

SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        private List<Exception> exceptions { get; set; }
```
PJB3005: You don't need the `{ get; set; }`, it's a private property after all.

PJB3005: Also, you're never actually initializing these lists so they'll throw null refs.

SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,38 @@
...
+        private Dictionary<Type, int> exp_count { get; set; }
```
PJB3005: This really isn't as useful as you think it'll be.

SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,38 @@
...
+        public void AddException(Exception E)
+        {
+            exceptions.Add(E);
+            exp_count.Add(E.GetType(), exp_count[E.GetType()] + 1);
```
PJB3005: This will throw an exception reliably because `exp_count` won't contain the type the first time this is called.

SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+        public string Display()
+        {
+            var ret = "";
```
PJB3005: Use a [`StringBuilder`](https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder?view=netframework-4.7.2) here.

SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,38 @@
...
+            foreach (Exception E in exceptions)
+            {
+                ret += "Exception in " + E.Source.ToString() + ", " + E.TargetSite.ToString() + ". \n";
+                ret += "Message: " + E.Message + "\n";
+                ret += "Stack Trace: " + E.StackTrace + "\n";
+                if (E.Data.Count > 0)
+                {
+                    ret += "Additional data:";
+                    foreach (Object x in E.Data)
```
PJB3005: Looping over a dictionary returns a `KeyValuePair<object, object>`, not an enumerator over the keys like you clearly intended here. You want to do `foreach (var x in E.Data.Keys)`

SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            exceptions.Add(E);
+            exp_count.Add(E.GetType(), exp_count[E.GetType()] + 1);
+        }
+        public string Display()
+        {
+            var ret = "";
+            foreach (Exception E in exceptions)
+            {
+                ret += "Exception in " + E.Source.ToString() + ", " + E.TargetSite.ToString() + ". \n";
```
PJB3005: Pretty sure the source is included inside the stack trace so this is probably redundant.

You may be better off just running `E.ToString()` and using that.

PJB3005: Also, you can do format strings with `$"Exception in: {E.Source} ..."`

SS14.Server/BaseServer.cs
```diff
@@ -255,8 +260,14 @@ public void MainLoop()
...
             // set GameLoop.Running to false to return from this function.
-            _mainLoop.Run();
-
+            try
+            {
+                _mainLoop.Run();
+            }
+            catch (Exception exp)
+            {
+                runtimeLog.AddException(exp);
```
PJB3005: Right now this just silently eats the exceptions with no way to be aware of them. The general idea behind the system is definitely a step in the right direction but taking out the logging is not a good idea.

PJB3005 (changes requested):

ShiftyRail:
@psykzz What could end execution early, the throw ?

psykzz:
Yeah.

ShiftyRail:
I see. I tried it on a small project and you're right.
I am used to Byond's exception handling only killing the current method.

SS14.Server/BaseServer.cs
```diff
@@ -255,8 +260,14 @@ public void MainLoop()
             _mainLoop.Tick += (sender, args) => Update(args.DeltaSeconds);
 
             // set GameLoop.Running to false to return from this function.
-            _mainLoop.Run();
-
+            try
+            {
+                _mainLoop.Run();
```
PJB3005: It just occurred to me that because this try catch wraps the main `Run` method, you're still gonna have the game crash because the main loop will die on an exception.

SS14.Server/BaseServer.cs
```diff
@@ -333,6 +344,9 @@ private void Cleanup()
             // shutdown entities
             _entities.Shutdown();
 
+            // Wrtie down exception log
+            System.IO.File.WriteAllText($"Runtime-{DateTime.Now.ToShortDateString()}", runtimeLog.Display());
```
PJB3005: This location is working directory relative so this won't always be next to the executable. See `PathHelpers.ExecutableRelativeFile`

SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+public struct ExceptionAndTime
```
PJB3005: Don't put this outside the namespace.

SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,68 @@
...
+}
+namespace SS14.Shared.Exceptions
+{
+    public class RuntimeLog
+    {
+        private Dictionary<Type, List<ExceptionAndTime>> exceptions;
+
+        public RuntimeLog()
+        {
+            this.exceptions = new Dictionary<Type, List<ExceptionAndTime>>();
```
PJB3005: You can put this initializer inside the field declaration like in BYOND.

captkirk88:
Your going to want to put the Cleanup(); in a finally block.

SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                ExceptionAndTime EandD = new ExceptionAndTime();
+                EandD.exp = E;
+                EandD.time = D;
+                exceptions[E.GetType()].Add(EandD);
+            }
+            else // Doesn't contain the element so let's instanciate it
+            {
+                exceptions[E.GetType()] = new List<ExceptionAndTime>();
+                ExceptionAndTime EandD = new ExceptionAndTime();
```
PJB3005: You're pretty much duplicating these 4 lines.

What I would do:

```cs
if (!exceptions.TryGetValue(E.GetType(), out var list))
{
    list = new List<ExceptionAndTime>();
    exceptions[E.GetType()] = list;
}

list.Add(new ExceptionAndTime
{
    exp = E,
    time = D,
});
```

SS14.Shared/Exceptions/RuntimeLog.cs
```diff
@@ -0,0 +1,68 @@
...
+                exceptions[E.GetType()].Add(EandD);
+            }
+
+        }
+        public string Display()
+        {
+            StringBuilder ret = new StringBuilder();
+            foreach (Type T in exceptions.Keys)
+            {
+                ret.AppendLine($"{exceptions[T].Count().ToString()} exception {((exceptions[T].Count() > 1) ? "s" : "")} {T.ToString()}");
```
PJB3005: You don't need to explicitly call `ToString()` for arguments to format strings.

PJB3005 (changes requested):

SS14.Server/BaseServer.cs
```diff
@@ -82,7 +83,7 @@ public class BaseServer : IBaseServer
         private GameLoop _mainLoop;
         private ServerRunLevel _runLevel;
 
-        private Shared.Exceptions.RuntimeLog runtimeLog;
```
PJB3005: You can put `[Dependency]` in front of this field, and then it'll automatically be filled in. You won't have to resolve it explicitly down below.

SS14.Shared/Timing/GameLoop.cs
```diff
@@ -66,72 +68,80 @@ public void Run()
 
             while (Running)
             {
-                var accumulator = _timing.RealTime - _lastTick;
-
-                // If the game can't keep up, limit time.
-                if (accumulator > maxTime)
+                try
```
PJB3005: I think you put this `try` a bit to broad. If one of the handlers throws an exception the main loop's timing system could still mess up. You should wrap the `Update?.Invoke(this, simFrameEvent);` and such instead.

PJB3005 (changes requested):
You need to register the runtime log class next to all the other IoC registrations. Check `GameController.cs`, `SS14.Server/Program.cs` and `SS14UnitTest.cs`

SS14.Server/BaseServer.cs
```diff
@@ -333,6 +338,10 @@ private void Cleanup()
             // shutdown entities
             _entities.Shutdown();
 
+            // Wrtie down exception log
+            var logPath = _config.GetCVar<string>("log.path");
+            System.IO.File.WriteAllText($"{PathHelpers.ExecutableRelativeFile(logPath)}/Runtime-{DateTime.Now.ToShortDateString()}", runtimeLog.Display());
```
PJB3005: You should use `System.IO.Path.Join` instead of string concat to make the file path.

PJB3005:
@ShiftyRail Any chance of this being fixed?

PJB3005:
Ding dong @ShiftyRail do you still want to finish this? I can take it off your hands if you'd prefer.

ShiftyRail:
Sorry school got me and then I forgot this
I'll get back to it tommorow and hopefully finish it this week-end

SS14.Server/BaseServer.cs
ShiftyRail: I couldn't find that method (my IDE didn't like it), but there is `Combine()` which apparently does the job.
https://docs.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=netcore-2.2

PJB3005: Ah hell you're absolutely right, I was mixing it up with Python.

SS14.Shared/Timing/GameLoop.cs
```diff
@@ -90,48 +93,55 @@ public void Run()
 
                 realFrameEvent.SetDeltaSeconds((float)_timing.RealFrameTime.TotalSeconds);
 
-                // process Net/KB/Mouse input
-                Input?.Invoke(this, realFrameEvent);
+                try {
```
PJB3005: You still have this try placed way too broadly, put it around the `Invoke()` calls themselves only.

ShiftyRail: How about defining a `TryInvoke` method that wraps a try/catch around it ?
```csharp
public object TryInvoke (object obj, object[] parameters) {
          try
          {
                This?.Invoke(obj, parameters);
          }
          catch (Exception exp)
          {
                IRuntimeLog runtimeLog = IoCManager.Resolve<IRuntimeLog>();
                runtimeLog.AddException(exp, DateTime.Now);
          }
}
```

Problem is that I'm not sure where I should I implement that method. In DM it could be anywhere but this is a sane language

PJB3005: I mean, sure, if it makes the code cleaner.

ShiftyRail:
I swear I had pushed something on this

SS14.Shared/Timing/GameLoop.cs
```diff
@@ -93,55 +93,78 @@ public void Run()
...
 
-                    // update out of the simulation
-                    simFrameEvent.SetDeltaSeconds((float)_timing.FrameTime.TotalSeconds);
+                // update out of the simulation
+                simFrameEvent.SetDeltaSeconds((float)_timing.FrameTime.TotalSeconds);
+                try
+                {
                     Update?.Invoke(this, simFrameEvent);
+                }
+                catch (Exception)
```
PJB3005, line 147: Hrm? Not gonna catch this one?

PJB3005:
Again, I'm totally fine with doing this myself.

ShiftyRail:
Then please do
Sorry for being slow and dumb

PJB3005:
Superseded by #759

PR #702, opened 2018-10-30
UI Range Check

Acruid:
Clients are now properly unsubscribed from entities when they move out of range. This prevents clients from being able to control entities from across the map.


PJB3005 (approved):
This is fine, although I would prefer if the system were more modular as to allow other closing conditions (for example, getting knocked unconscious).

PR #705, opened 2018-11-20
IPv6 & multi-peer NetManager support.

PJB3005:
1. IPv6 support. Woo!
2. NetManager can now explicitly bind to adresses and work with multiple peers. ~~I thought I needed this for IPv6 because I forgot about IPv4-mapped adresses, but the code's written so let's keep it.~~ OpenBSD doesn't support IPv4-mapped IPv6 addresses so effort not wasted.

### DIFF
--- a/Lidgren.Network/NetPeer.Internal.cs
+++ b/Lidgren.Network/NetPeer.Internal.cs
@@ -114,7 +114,7 @@ namespace Lidgren.Network
 				iep = new IPEndPoint(m_configuration.LocalAddress, m_configuration.Port);
 				EndPoint ep = (EndPoint)iep;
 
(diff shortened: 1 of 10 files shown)


Lidgren.Network/NetPeer.Internal.cs
```diff
@@ -386,6 +392,7 @@ private void Heartbeat()
 				int bytesReceived = 0;
 				try
 				{
+					var honk = this;
```
Cyberboss, line 395: nani?

PJB3005: Oh jeez.

For some reason Rider is failing to resolve `this` references in the debugger so I had to do that to get one. Forgot to remove it.

Cyberboss: toppest

PR #709, opened 2018-11-25
Server status framework.

PJB3005:
Adds a HTTP server to the server that exposes `/status` to fetch the server status.

### DIFF
--- a/SS14.Server/ServerStatus/StatusHost.cs
+++ b/SS14.Server/ServerStatus/StatusHost.cs
@@ -1,0 +1,156 @@
+using System;
+using System.IO;
+using System.Net;
+using System.Text;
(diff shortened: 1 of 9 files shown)


SS14.Server/ServerStatus/StatusHost.cs
```diff
@@ -0,0 +1,178 @@
...
+                response.ContentType = "text/plain";
+                _respondText(response, "400 Bad Request", false);
+                return;
+            }
+
+            var head = request.HttpMethod == "HEAD";
+            try
+            {
+                var uri = request.Url;
+                if (uri.AbsolutePath == "/teapot")
```
Simyon264, line 100: i hate you

PR #718, opened 2018-12-10
Convert PushInheritance to iteration in EntityPrototype

ghost:
fixes the "TODO: remove recursion somehow" comment

### DIFF
--- a/SS14.Shared/Prototypes/EntityPrototype.cs
+++ b/SS14.Shared/Prototypes/EntityPrototype.cs
@@ -278,14 +278,50 @@ namespace SS14.Shared.GameObjects
                     {
                         break;
                     }
-                    foreach (EntityPrototype child in Children)
-                    {
-                        PushInheritance(this, child);
-                    }
-
+                    PushInheritanceAll();
                     break;
             }
             return false;
+        }
+
+        /// <summary>
+        /// Iteratively pushes inheritance down to all children, children's children, etc. breadth-first.
+        /// </summary>
+        private void PushInheritanceAll()
+        {
+            if (Children == null)
+            {
+                return;
+            }
+            var sourceTargets = new List<Tuple<EntityPrototype, List<EntityPrototype>>>();
+            var newSources = new List<EntityPrototype>();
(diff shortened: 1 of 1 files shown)


ghost:
compiles, runs the server, runs the client until I crashed from trying to pun a gun into a backpack while sprinting which is probably an unrelated bug

SS14.Shared/Prototypes/EntityPrototype.cs
```diff
@@ -278,14 +278,50 @@ public bool Sync(IPrototypeManager manager, int stage)
...
+        /// <summary>
+        /// Iteratively pushes inheritance down to all children, children's children, etc. breadth-first.
+        /// </summary>
+        private void PushInheritanceAll()
+        {
+            if (Children == null)
+            {
+                return;
+            }
+            var sourceTargets = new List<Tuple<EntityPrototype, List<EntityPrototype>>>();
```
PJB3005: Is there any reason you're using reference tuples here? They're ugly as hell.

SS14.Shared/Prototypes/EntityPrototype.cs
```diff
@@ -278,14 +278,52 @@ public bool Sync(IPrototypeManager manager, int stage)
...
+            if (Children == null)
+            {
+                return;
+            }
+
+            var sourceTargets = new List<(EntityPrototype, List<EntityPrototype>)> {(this, Children)};
+            var newSources = new List<EntityPrototype>();
+            while (true)
+            {
+                foreach (var sourceTargetTuple in sourceTargets)
```
PJB3005: You can do `var (source, targetList) in sourceTargets` here I believe.

RemieRichards: you can.

PJB3005: See I knew you can but if I write "I believe" you can't hold me accountable if I messed up the exact positioning of the syntax.

RemieRichards: Haha, I do that too.
And to that end, I think it might need to be ```(var source, var targetList)``` since source and targetList are of differing types.

PJB3005 (approved):

PR #761, opened 2019-03-22
[Ready] Fix server runtimes log exception

Nirnael:
Fixes missing directory exception when writing the server runtimes log. Not sure if you intended to create a new directory for runtimes or just create a file name with this structure, but since it has a dash in it I assumed it's the name.


Nirnael:
This wafflebot

PR #767, opened 2019-03-27
Establish standard format for runtime log naming and future-proof it until the human race perishes

Nirnael:
This PR establishes the format "Runtime-yyyy-MM-dd-THH-mm-dd.txt" for runtime logs from the server, preventing locale from messing with string escape characters and causing directory runtimes at next line. ~I've replaced the dash with an underscore as they can cause problems in some shells such as windows CLI.~ Dashes are staying because whoever has a shell that complains needs to get a new one.

~~**_I have most importantly made sure that SS14 can live on until the year of our Lord God Emperor and veritable saviour 99,999 Anno Domini of the Gregorian Calendar by extending the runtime-log-naming  year range to 5 digits._**~~ 
~~_Let the meme PRs begin if they haven't already_~~

4 digits so logs will display until 9999, godspeed future maintainers.

### DIFF
--- a/SS14.Server/BaseServer.cs
+++ b/SS14.Server/BaseServer.cs
@@ -291,7 +291,7 @@ namespace SS14.Server
 
             // Wrtie down exception log
             var logPath = _config.GetCVar<string>("log.path");
-            var pathToWrite = Path.Combine(PathHelpers.ExecutableRelativeFile(logPath), "Runtime_" + DateTime.Now.ToString("yyyyy_MM_dd") + ".txt");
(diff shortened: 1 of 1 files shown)


Nirnael:
_Should I extend it to 999,999? Vote with your thumbs_ Will the human race even last that long?

clusterfack:
Putting at 1 y would probably be sufficiently optimistic

PJB3005 (changes requested):
Just store it with ISO 8601. You can get that by formatting with `"o"` as parameter I believe.

PJB3005 (approved):

PR #769, opened 2019-03-28
Adds a frame counter to timing debug monitor

Nirnael:
Use case explained in its documentation. I could probably make it a long in case someone runs at thousands of fps for days, or make it debug only.

### DIFF
--- a/SS14.Shared/Interfaces/Timing/IGameTiming.cs
+++ b/SS14.Shared/Interfaces/Timing/IGameTiming.cs
@@ -49,6 +49,14 @@ namespace SS14.Shared.Interfaces.Timing
         TimeSpan RealFrameTimeStdDev { get; }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


SS14.Shared/Interfaces/Timing/IGameTiming.cs
```diff
@@ -48,6 +48,14 @@ public interface IGameTiming
         /// </summary>
         TimeSpan RealFrameTimeStdDev { get; }
 
+        /// <summary>
+        ///     Current graphics frame since init OpenGL which is taken as frame 1. Useful to set a conditional breakpoint on specific frames, and
+        ///     syncrhonize with OGL debugging tools that capture frames. Depending on the tools used, this frame
```
PJB3005: I'm sorry but I can't not point out that typo.

Nirnael: sigh

PJB3005 (approved):

PR #770, opened 2019-03-28
Initial Discord Rich Presence support

occune:
but we still need to tweak cosmetic aspects (like gamestates etc) and change the timer
huge thanks to @Injazz

### DIFF
--- a/BuildChecker/BuildChecker.csproj
+++ b/BuildChecker/BuildChecker.csproj
@@ -33,6 +33,19 @@ https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild
   <PropertyGroup>
     <StartupObject />
   </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="DiscordRPC, Version=1.0.77.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\DiscordRichPresence.1.0.77\lib\net35\DiscordRPC.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=11.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <HintPath>..\packages\Newtonsoft.Json.11.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
   <Target Name="Build">
     <Exec Command="$(Python) git_helper.py" CustomErrorRegularExpression="^Error" />
   </Target>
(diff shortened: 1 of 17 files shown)


BuildChecker/BuildChecker.csproj
```diff
@@ -33,6 +33,19 @@ https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild
   <PropertyGroup>
     <StartupObject />
   </PropertyGroup>
+  <ItemGroup>
```
PJB3005, line 36: I think you added these references to the RPC assembly in a bit too much projects there. It should only be in `SS14.Client`

SS14.Client/GameController.cs
```diff
@@ -155,6 +157,8 @@ private void Startup()
             _placementManager.Initialize();
             _viewVariablesManager.Initialize();
 
+            RichPresenceClient presence = new RichPresenceClient();
```
PJB3005, line 160: This should be made an IoC dependency:

https://github.com/space-wizards/space-station-14/wiki/IoC-(Inversion-of-Control)

SS14.Client/Utility/RichPresenceClient.cs
```diff
@@ -0,0 +1,59 @@
...
+            timer.Start();
+
+            //Connect
+            client.Initialize();
+
+            //Send a presence. Do this as many times as you want
+            client.SetPresence(new RichPresence()
+            {
+                Details = "Gamemode: Valid Hunt Racing",
+                State = "/vg/station",
```
PJB3005, line 36: Not sure whether referring to /vg/ directly is a good idea.

I think a more agnostic name might be better.

SS14.Client/Utility/RichPresenceClient.cs
```diff
@@ -0,0 +1,59 @@
...
+                Assets = new Assets()
+                {
+                    LargeImageKey = "vgstation-logo2",
+                    LargeImageText = "Meta Station",
+                    SmallImageKey = "logo",
+                },
+                Timestamps = Timestamps.FromTimeSpan(10)
```
PJB3005, line 43: Shouldn't this value actually be incremented periodically? If it isn't wouldn't it be better to leave it out?

SS14.Client/Utility/RichPresenceClient.cs
```diff
@@ -0,0 +1,59 @@
...
+            });       
+        }
+
+        public void Free()
```
PJB3005, line 47: You should implement `IDisposable` and rename this to `Dispose()`.

PJB3005 (changes requested):
I guess it would be nice to have some things like changing the status when connecting to a server but oh well.

io-june:
I also want to note that there is a bug in DiscordRPC causing discord pipe to fail connection on Mono.
https://github.com/Lachee/discord-rpc-csharp/issues/54
you need fresh nuget version (which gonna update soon), or do it manually

SS14.Client/GameController.cs
ZoldorfTheWizard: +1 ^

PJB3005:
Superseded by #776

PR #782, opened 2019-04-06
Implemented IEquatable on Angle

tentekal:
Babby's first commit pt ii, I actually didn't have my forked engine repo as a remote so at least that's fixed.

### DIFF
--- a/SS14.Shared.Maths/Angle.cs
+++ b/SS14.Shared.Maths/Angle.cs
@@ -6,7 +6,7 @@ namespace SS14.Shared.Maths
     ///     A representation of an angle, in radians.
     /// </summary>
     [Serializable]
-    public readonly struct Angle : IApproxEquatable<Angle>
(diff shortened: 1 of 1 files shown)


SS14.Shared.Maths/Angle.cs
```diff
@@ -129,6 +129,38 @@ private static double Reduce(double theta)
...
+            if (ReferenceEquals(null, obj)) return false;
+            return obj is Angle angle == Equals(Theta);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                return (Theta.GetHashCode() * 397) ^ Theta.GetHashCode();
```
PJB3005: Pretty sure you only need to `return Theta.GetHashCode();` here and you can then remove the `unchecked` too.

tentekal: Updated code to reflect this.

PR #784, opened 2019-04-11
Remove InjectDependencies Calls

Acruid:
This PR focuses on removing the static `IoCManager.InjectDependencies(this)` calls from all engine code. Along the way, any files touched also had any static `IoCManager.Resolve()` calls removed. Some of the UI classes have a lot of dependencies injected into their constructors now, which we may want to address at a later time. Redesigning the entire thing to actually enforce SRP and remove the logic-in-constructor anti-pattern in UI controls was outside the scope of this PR.

`States` (game scenes like main menu), `Components`, and `EntitySystems` now have dependencies injected into them from the DI container. There is no reason to use any static IoC calls in these classes, or any of their child objects.


PJB3005:
Controls are pretty yikes material, yeah. Honestly their entire construction system should be re-worked.

Otherwise seems fine.

PR #785, opened 2019-04-11
Locker shoving

PrPleGoo:
Added a bool to ICollidable to indicate friction with the floor, for thrown items down the line.
Added a bool to PhysicsComponent to indicate if movement is possible.
Added list of PhysicsComponent that are slowing "this" down.
PhysicsComponent handles BumpEntMsg by adding the bumping entity to that list.
Implemented ICollideSpecial to prevent collision with items in that list.
Reworked movement code:
- Separated calculating movement from doing movement;
- Calculating movement now involves several recalculations to see who is on "that" list;
- Friction of a tile now slows movement down depending on an object's mass;
Optimized collision code by building a Dictionary with a snapped grid location for key, reducing the amount of bodies that need to be checked significantly.


SS14.Shared.Maths/Vector2.cs
```diff
@@ -8,7 +8,7 @@ namespace SS14.Shared.Maths
     /// </summary>
     [StructLayout(LayoutKind.Sequential)]
     [Serializable]
-    public readonly struct Vector2 : IEquatable<Vector2>, IApproxEquatable<Vector2>
+    public readonly struct Vector2 : IEquatable<Vector2>, IApproxEquatable<Vector2>, IComparable<Vector2>
```
PJB3005: I'm not sure it makes much sense to make `Vector2`s automatically comparable. If you need magnitude comparison, couldn't you use a specific `IComparer` instead?

SpaceStation14.sln
```diff
@@ -153,12 +153,12 @@ Global
 		{8AF31169-49B1-4A12-B8F4-2A0674A9E7CB}.Release|x64.Build.0 = Release|Any CPU
 		{8AF31169-49B1-4A12-B8F4-2A0674A9E7CB}.Release|x86.ActiveCfg = Release|Any CPU
 		{8AF31169-49B1-4A12-B8F4-2A0674A9E7CB}.Release|x86.Build.0 = Release|Any CPU
-		{8AF31169-49B1-4A12-B8F4-2A0674A9E7CB}.Tools|Any CPU.ActiveCfg = Tools|Any CPU
```
PJB3005: uh, this on accident?

PrPleGoo: Yes

SS14.Shared.Maths/Vector2.cs
PrPleGoo: I'll check it out thanks.

SpaceStation14.sln
PrPleGoo: Changed it back.

SS14.Shared.Maths/Vector2.cs
PrPleGoo: Removed the implementation. Where I needed it I now instead order the list.

PJB3005 (approved):

PR #786, opened 2019-04-11
Rendering order now done by who was last dirty instead of Uid

PrPleGoo:


SS14.Client/Graphics/Clyde/Clyde.Rendering.cs
```diff
@@ -225,7 +225,7 @@ public void Render(FrameEventArgs args)
                     return cmp;
                 }
 
-                return a.Owner.Uid.CompareTo(b.Owner.Uid);
```
PJB3005, line 228: Seems like a bad idea, since last modified tick is only intended for net code.

It could randomly change, not to mention that it's not held up to standards on client side.

PJB3005 (changes requested):

PrPleGoo:
Relies on https://github.com/space-wizards/space-station-14/pull/194 for actually setting the value.

PR #787, opened 2019-04-12
Actual lockers

PrPleGoo:
Adds storing entities into lockers the way we all know and love.


ZoldorfTheWizard (approved):

SS14.Shared/Physics/CollisionGroup.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+namespace SS14.Shared.Physics
+{
+    /// <summary>
+    ///     Defined collision groups for the physics system.
+    /// </summary>
+    [Flags]
+    public enum CollisionGroup
```
PJB3005: See the problem is that now this is defined as an actual enum in the engine, it can't have things re-organized in content.

PrPleGoo: I don't really see how that's a problem.
Defining it as an int server side while it is in fact a flag seems like a bad idea.
Not to mention all the int casting.
If you like I can set up a content side Enum that copies the values over but I'm not sure what that would add.

PJB3005: The reason it's a problem that if content wants to make their own collision groups (and they will) it'd be even more ugly than the alternative (integer constants)

PrPleGoo:
Fixed

PJB3005 (approved):

PR #789, opened 2019-04-12
Entity Interpolation

Acruid:
* Adds entity interpolation to the game. [Here is an explanation](https://www.youtube.com/watch?v=DoexBWfjWUg).
* Added the `net.interp` cvar to enable/disable the entire feature. It is disabled by default because it causes `1/Tickrate` movement lag for the player. Clientside movement prediction will solve this.
* Added the `net.interp_ratio` cvar. This is an integer of how many extra states to buffer. Other games call this "network smoothing". This also causes `1/Tickrate` input lag for every state buffered.
* The IGameTiming.InSimulation field can now be used to enable/disable interpolation transparently. This is set automatically by the game loop.
* Tickrate is now officially 10. This is to ease up of the network bandwidth, and the interpolation hides the jitter when moving.

### DIFF
--- a/SS14.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/SS14.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -8,7 +8,8 @@ using SS14.Shared.Interfaces.GameObjects.Components;
 using SS14.Shared.Interfaces.Map;
  using SS14.Shared.Interfaces.Timing;
  using SS14.Shared.IoC;
-using SS14.Shared.Map;
+ using SS14.Shared.Log;
+ using SS14.Shared.Map;
 using SS14.Shared.Maths;
 using SS14.Shared.Serialization;
  using SS14.Shared.Utility;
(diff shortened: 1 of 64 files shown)


SS14.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -382,81 +389,154 @@ public override void ExposeData(ObjectSerializer serializer)
...
 
         /// <inheritdoc />
-        public override void HandleComponentState(ComponentState state)
+        public override void HandleComponentState(ComponentState curState, ComponentState nextState)
         {
-            var newState = (TransformComponentState)state;
+            var curTrans = curState as TransformComponentState;
+            var nextTrans = nextState as TransformComponentState;
+
+            Logger.DebugS("ent", $"cTick={(curTrans != null? curTrans.LocalPosition.ToString():"-")}, nTick={(nextTrans!=null?nextTrans.LocalPosition.ToString():"-")}");
```
PJB3005: Stray debug log.

ZoldorfTheWizard (approved):

PR #795, opened 2019-04-20
Tick Logic Timing and Bandwidth Improvements

Acruid:
Code was changed so that time starts at tick zero, and all entities are initialized in the space between tick 0 to 1. This was a solution to the fact fromSequence was including entities with lastStateUpdate tick one less than the actual sequence number. This is purely an issue about when exactly the game does its logic relative to the tick signals, ie positive or negative edge of the tick. Right now the game is set up to run the game logic on the positive edge of a Tick (enforced by GameLoop).

The second big part is that an empty collection takes 5ish bytes more than a null collection in the NetSerializer, even though they are conceptually the same thing in a game state. This PR puts null checks around everything and allows GameStates to have whole sections of itself null. This reduced the payload size of an empty game state from 49B to 15B.

This PR also fixes the bug where the physics system marks every movable entity as dirty every tick.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Physics/PhysicsComponentState.cs
+++ b/Robust.Shared/GameObjects/Components/Physics/PhysicsComponentState.cs
@@ -11,9 +11,9 @@ namespace Robust.Shared.GameObjects
     public class PhysicsComponentState : ComponentState
     {
         /// <summary>
-        ///     Current mass of the entity.
+        ///     Current mass of the entity, stored in grams.
         /// </summary>
-        public readonly float Mass;
+        public readonly int Mass;
 
         /// <summary>
         ///     Current velocity of the entity.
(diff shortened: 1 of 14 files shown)


Robust.Shared/GameObjects/Components/Physics/PhysicsComponentState.cs
```diff
@@ -11,9 +11,9 @@ namespace Robust.Shared.GameObjects
     public class PhysicsComponentState : ComponentState
     {
         /// <summary>
-        ///     Current mass of the entity.
+        ///     Current mass of the entity, stored in grams.
         /// </summary>
-        public readonly float Mass;
+        public readonly int Mass;
```
PJB3005, line 16: I get that the point of this was to reduce the size of the game state, but aren't you still sending 4 bytes?

PJB3005:
A few unit tests are failing, just so you know.

Robust.Shared/GameObjects/Components/Physics/PhysicsComponentState.cs
Acruid: No, NetSerializer packs ints.

PR #797, opened 2019-04-22
ItemLists now have a scrollbar and can be scrolled with the mouse wheel as well.

gradientvera:


Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -1,19 +1,29 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Drawing;
```
PJB3005: What did you include `System.Drawing` for here?

Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -56,6 +70,10 @@ public void AddItem(string text, Texture icon = null, bool selectable = true)
             {
                 var i = new Item() {Text = text, Icon = icon, Selectable = selectable};
                 _itemList.Add(i);
+                if (_isAtBottom && ScrollFollowing)
```
PJB3005, line 86: You're not re-calculating `_totalContentHeight` here before running this so this won't work correctly.

Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -426,7 +451,22 @@ protected internal override void Draw(DrawingHandleScreen handle)
                 }
 
                 separation += (0, itemHeight);
+                _itemListHeight += itemHeight;
+            }
+
+            _totalContentHeight = (int)_itemListHeight;
```
PJB3005: Setting this inside the drawing loop is a bad idea.

Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -453,6 +493,56 @@ protected void DrawTextInternal(DrawingHandleScreen handle, string text, UIBox2
             }
         }
 
+        protected override Vector2 CalculateMinimumSize()
+        {
+            return (ActualBackground?.MinimumSize.X ?? 0, _itemListHeight);
```
PJB3005: You need to accommodate for the the horizontal size of the background too.

PJB3005 (changes requested):

Robust.Client/UserInterface/Controls/ItemList.cs
gradientvera: Honestly, I'm not even sure myself. Rider must have screwed something up.
I'll remove it

Robust.Client/UserInterface/Controls/ItemList.cs
gradientvera: Hmm, I'm not too sure what you mean here... Isn't MinimumSize.X enough, or should I do something else?

Robust.Client/UserInterface/Controls/ItemList.cs
gradientvera: Should be fixed now.

Robust.Client/UserInterface/Controls/ItemList.cs
gradientvera: Should be fixed now.

Robust.Client/UserInterface/Controls/ItemList.cs
PJB3005: I meant vertical size, durr.

gradientvera: Should be good now, then

Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -509,7 +509,9 @@ protected void DrawTextInternal(DrawingHandleScreen handle, string text, UIBox2
 
         protected override Vector2 CalculateMinimumSize()
         {
-            return (ActualBackground?.MinimumSize.X ?? 0, _itemListHeight);
+            if (_itemListHeight > ActualBackground?.MinimumSize.Y)
```
PJB3005: Wouldn't you just.. need to do ```cs
return (ActualBackground?.MinimumSize.X ?? 0, _itemListHeight + ActualBackground?.MinimumSize.Y ?? 0);
```

Actually I guess if it weren't obvious, Godot uses `MinimumSize` on styleboxes as the "extra width you need to add" which perhaps is *kinda* annoying and should be fixed, but oh well.

gradientvera: Oh, I didn't know about that... Sorry, should be fixed now.

PJB3005 (approved):

PR #818, opened 2019-05-24
Add StyleBox overrides to ProgressBar

Moneyl:
This adds foreground and background style box overrides to `ProgressBar` primarily so it's colors can be changed easily. The overrides are `StyleBoxFlat` types instead of `StyleBox` as `StyleBox` doesn't have a color property. `ProgressBar._getBackground` and `ProgressBar._getForeground` were also changed to return `StyleBoxFlat` instead of `StyleBox` to keep things simple rather than interchangeably using the two types throughout the class.

### DIFF
--- a/Robust.Client/UserInterface/Controls/ProgressBar.cs
+++ b/Robust.Client/UserInterface/Controls/ProgressBar.cs
@@ -43,26 +43,26 @@ namespace Robust.Client.UserInterface.Controls
         }
 
         [Pure]
-        private StyleBoxFlat _getBackground()
+        private StyleBox _getBackground()
         {
             if (BackgroundStyleBoxOverride != null)
             {
                 return BackgroundStyleBoxOverride;
             }
 
-            TryGetStyleProperty(StylePropertyBackground, out StyleBoxFlat ret);
+            TryGetStyleProperty(StylePropertyBackground, out StyleBox ret);
             return ret;
         }
 
         [Pure]
-        private StyleBoxFlat _getForeground()
+        private StyleBox _getForeground()
         {
             if (ForegroundStyleBoxOverride != null)
             {
                 return ForegroundStyleBoxOverride;
             }
 
-            TryGetStyleProperty(StylePropertyForeground, out StyleBoxFlat ret);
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/Controls/ProgressBar.cs
```diff
@@ -19,17 +22,47 @@ public ProgressBar(string name) : base(name)
...
+            get => _foregroundStyleBoxOverride;
+            set
+            {
+                _foregroundStyleBoxOverride = value;
+                MinimumSizeChanged();
+            }
+        }
+
         [Pure]
-        private StyleBox _getBackground()
```
PJB3005: Keep all this as `StyleBox`, don't change it to `StyleBoxFlat`.

PJB3005 (changes requested):

Robust.Client/UserInterface/Controls/ProgressBar.cs
Moneyl: Done.

Robust.Client/UserInterface/Controls/ProgressBar.cs
```diff
@@ -19,16 +22,46 @@ public ProgressBar(string name) : base(name)
         {
         }
 
+        public StyleBoxFlat BackgroundStyleBoxOverride
```
PJB3005: This needs to be `StyleBox` too.

Moneyl: People will need to cast this to `StyleBoxFlat` to access `BackgroundColor`. Why is `StyleBox` preferred in this case?

PJB3005: Because you can pass any type of `StyleBox`, not only a `StyleBoxFlat`. If you want to avoid having to cast it you should keep a direct reference somewhere else.

Moneyl: I see. It's done.

PJB3005 (approved):

PR #823, opened 2019-05-31
Circle Ranges

io-june:
Needed for https://github.com/space-wizards/space-station-14/pull/247

### DIFF
--- a/Robust.Shared/Map/MapGrid.cs
+++ b/Robust.Shared/Map/MapGrid.cs
@@ -242,19 +242,17 @@ namespace Robust.Shared.Map
         public IEnumerable<TileRef> GetTilesIntersecting(Circle worldArea, bool ignoreEmpty = true, Predicate<TileRef> predicate = null)
         {
             var aabb = new Box2(worldArea.Position.X - worldArea.Radius, worldArea.Position.Y - worldArea.Radius, worldArea.Position.X + worldArea.Radius, worldArea.Position.Y + worldArea.Radius);
-            var tiles = new List<TileRef>();
 
             foreach(var tile in GetTilesIntersecting(aabb, ignoreEmpty))
             {
-                if (worldArea.Contains(new Vector2(tile.X, tile.Y)))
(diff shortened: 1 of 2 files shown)


Robust.Shared/Map/MapGrid.cs
```diff
@@ -238,6 +238,25 @@ public IEnumerable<TileRef> GetTilesIntersecting(Box2 worldArea, bool ignoreEmpt
             return tiles;
         }
 
+        /// <inheritdoc />
+        public IEnumerable<TileRef> GetTilesIntersecting(Circle worldArea, bool ignoreEmpty = true, Predicate<TileRef> predicate = null)
+        {
+            var aabb = new Box2(worldArea.Position.X - worldArea.Radius, worldArea.Position.Y - worldArea.Radius, worldArea.Position.X + worldArea.Radius, worldArea.Position.Y + worldArea.Radius);
+            var tiles = new List<TileRef>();
```
PJB3005: You can remove this intermediate list and use `yield return` instead of `tiles.Add()`. Then the C# compiler makes an enumerable for you.

Robust.Shared/Map/MapGrid.cs
```diff
@@ -238,6 +238,25 @@ public IEnumerable<TileRef> GetTilesIntersecting(Box2 worldArea, bool ignoreEmpt
...
+
+            foreach(var tile in GetTilesIntersecting(aabb, ignoreEmpty))
+            {
+                if (worldArea.Contains(new Vector2(tile.X, tile.Y)))
```
PJB3005: I realize your math is off for two reasons here:

1. Tile indices are not grid coordinates. You need a method like `GridTileToLocal`
2. You are using the bottom left of a tile to do the contains check, but that would easily miss many tiles since the method fetches *intersecting*.

io-june: 1. OK
2. i can do something like that instead
```
if (GridTileToLocal(tile.GridIndices).Distance(_mapManager, new GridCoordinates(worldArea.Position,tile.GridIndex)) <= worldArea.Radius)
```

io-june:
fucked my own tree through submodule who could guessed

PR #824, opened 2019-06-03
Add gradient helper function, Color.HueShiftedGradient

Moneyl:
Moves the gradient logic from the APC and SMES into this helper function. Should be useable as a general gradient helper.

### DIFF
--- a/Robust.Shared.Maths/Color.cs
+++ b/Robust.Shared.Maths/Color.cs
@@ -871,6 +871,34 @@ namespace Robust.Shared.Maths
                 FloatMath.CloseTo(A, other.A);
         }
 
+        /// <summary>
+        ///     Calculates a gradient by lerping between 3 different hue values. The gradient has uniform saturation and brightness
+        /// </summary>
+        /// <param name="leftHue">The hue at the far left side of the gradient. When gradientPosition = 0.0</param>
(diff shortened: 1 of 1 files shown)


PJB3005:
This approach is a bit too specific.

First of all I don't care too much about having it in the engine like this. I would've been fine with you putting the code for that specific gradient for the power bar in some static in content too. The point was to remove the glaring copy paste, not to make it useful for the engine.

Secondly, I think this method is too specific to really be useful. A method for interpolating over a gradient with N points and then giving the rest of the job to content is much better.

Moneyl:
Yeah, I see what you mean. I pushed changes that let it take an arbitrary amount of hues to interpolate over so it's more useful in any situation. Here's an example gif interpolating between 6 hues. Some of the color shifts are more rapid since the hues I selected are not equally spaced. 

WBvALVmqB8

PJB3005:
I meant that it would be better to have a standard gradient interpolation function for any float, then use that combined with fromhsv from content.

PJB3005:
Any update on this?

Moneyl:
Sorry about the lack of updates. I got caught up in another project and life. For the moment I'm going to close my existing PRs since I don't see myself getting to them anytime soon.

PR #827, opened 2019-06-04
Box2Rotated & Vector2 Rotation

Acruid:
* Adds the Box2Rotated struct for holding a box rotated in worldspace. IDK how useful this will be, we will see.
* Adds a function to rotate a Vector2 by an Angle. Now you don't have to explicity build a rotation matrix to rotate a vector!

### DIFF
--- a/Robust.UnitTesting/Shared/Maths/Angle_Test.cs
+++ b/Robust.UnitTesting/Shared/Maths/Angle_Test.cs
@@ -148,5 +148,17 @@ namespace Robust.UnitTesting.Shared.Maths
 
             Assert.That(target.GetCardinalDir(), Is.EqualTo(test.Item3));
         }
+
+        [Test]
+        public void TestAngleRotateVec()
+        {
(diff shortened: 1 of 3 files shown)


Robust.UnitTesting/Shared/Maths/Angle_Test.cs
```diff
@@ -148,5 +148,17 @@ public void TestAngleToCardinal([ValueSource(nameof(CardinalSources))] (float, f
...
+
+        [Test]
+        public void TestAngleRotateVec()
+        {
+            var angle = new Angle(MathHelper.Pi / 6);
+            var vec = new Vector2(0.5f, 0.5f);
+
+            var result = angle.RotateVec(vec);
+
+            Assert.That(FloatMath.CloseTo(result.X, 0.183013f), Is.True, result.X.ToString);
```
PJB3005:
```suggestion
            Assert.That(result, new ApproxEqualityConstraint(new Vector2(0.183013f, 0.683013f));
```

PR #839, opened 2019-08-04
Fix missing semicolon

DamianX:


DamianX:
Why is that test failing?

DamianX:
From what I understand https://github.com/space-wizards/RobustToolbox/commit/85a4375905e37a218f77a54525563c4410e53305 introduced an optimization: instead of running Stream.ToArray(), the stream is copied to a MemoryStream, and then the MemoryStream's internal buffer is returned.
The buffer, however, may have a different size than the original stream, and so would return something like a byte array with `256-stream.Length` zeroes at the end.

My commit changes GetBuffer to ToArray, so that the size of the returned array is the correct one, but this probably defeats the purpose of the optimization.
Thoughts? @PJB3005 

Also, Travis just shat the bed.

PJB3005:
The optimization is still better than it was before, though yes `GetBuffer` was probably a bad call.

PR #842, opened 2019-08-09
Fix for minimizing the client causing a crash

ShadowCommander:
Added a check for zero when _window.Resize gets called.


PJB3005:
This really isn't the best way to solve this. If window size ends up at 0 for any reason it'll still break. The code should be able to handle a zero-sized window OR should cleanly provide default like 1,1 if it happens.

PR #845, opened 2019-08-15
Implemented RobustRandom

DamianX:


Robust.Shared/Utility/RandomString.cs
```diff
@@ -1,22 +1,20 @@
...
+using Robust.Shared.Interfaces.Random;
+using Robust.Shared.IoC;
 
 namespace Robust.Shared.Utility
 {
     public class RandomString
     {
         private const string _chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
-        private static Random _rand;
+        private static readonly IRobustRandom Rand = IoCManager.Resolve<IRobustRandom>();
```
PJB3005: You can't use a static initializer here because it's totally possible for the runtime to initialize this in another thread/before IoC initializes.

Just do an IoC resolve inside the generate method every time.

PJB3005 (changes requested):

PJB3005 (approved):

PR #846, opened 2019-08-16
Allowed multiple Pseudo Classes per element.

CatTheSystem:
Needed NanoStyle.cs update so need to accept pull request on Space-station-14

### DIFF
--- a/Robust.Client/UserInterface/Control.Styling.cs
+++ b/Robust.Client/UserInterface/Control.Styling.cs
@@ -30,9 +30,9 @@ namespace Robust.Client.UserInterface
             }
         }
 
-        public bool HasStylesPseudoClass(ICollection<string> collection)
+        public bool HasStylePseudoClass(string className)
         {
-            return _stylePseudoClass.IsSubsetOf(collection);
+            return _stylePseudoClass.Contains(className);
         }
 
         protected void AddStylePseudoClass(string className)
@@ -58,6 +58,7 @@ namespace Robust.Client.UserInterface
 
             if (className == null)
             {
+                Restyle();
                 return;
             }
 
--- a/Robust.Client/UserInterface/Control.Styling.cs
+++ b/Robust.Client/UserInterface/Control.Styling.cs
@@ -13,7 +13,7 @@ namespace Robust.Client.UserInterface
         private readonly HashSet<string> _styleClasses = new HashSet<string>();
         private readonly HashSet<string> _stylePseudoClass = new HashSet<string>();
         public ICollection<string> StyleClasses { get; }
-        public ICollection<string> StylePseudoClass { get => _stylePseudoClass; }
+        public IReadOnlyCollection<string> StylePseudoClass { get => _stylePseudoClass; }
 
         private string _styleIdentifier;
 
@@ -30,12 +30,12 @@ namespace Robust.Client.UserInterface
             }
         }
 
-        public bool HasStylePseudoClass(string className)
+        public bool HasStylesPseudoClass(ICollection<string> collection)
         {
-            return _stylePseudoClass.Contains(className);
+            return _stylePseudoClass.IsSubsetOf(collection);
         }
 
-        public void AddStylePseudoClass(string className)
+        protected void AddStylePseudoClass(string className)
         {
             if (_stylePseudoClass.Contains(className))
             {
@@ -46,20 +46,21 @@ namespace Robust.Client.UserInterface
             Restyle();
         }
 
-        public void RemoveStylePseudoClass(string className)
+        protected void RemoveStylePseudoClass(string className)
         {
             _stylePseudoClass.Remove(className);
             Restyle();
         }
 
-        public void SetOnlyStylePseudoClass(string className)
+        protected void SetOnlyStylePseudoClass(string className)
         {
-            if (_stylePseudoClass.Contains(className))
+            _stylePseudoClass.Clear();
+
+            if (className == null)
             {
                 return;
             }
 
-            _stylePseudoClass.Clear();
             _stylePseudoClass.Add(className);
             Restyle();
         }
--- a/Robust.Client/UserInterface/Control.Styling.cs
+++ b/Robust.Client/UserInterface/Control.Styling.cs
@@ -11,7 +11,9 @@ namespace Robust.Client.UserInterface
 
         private readonly Dictionary<string, object> _styleProperties = new Dictionary<string, object>();
         private readonly HashSet<string> _styleClasses = new HashSet<string>();
+        private readonly HashSet<string> _stylePseudoClass = new HashSet<string>();
         public ICollection<string> StyleClasses { get; }
+        public ICollection<string> StylePseudoClass { get => _stylePseudoClass; }
 
         private string _styleIdentifier;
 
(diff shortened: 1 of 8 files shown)


Robust.Client/UserInterface/Control.Styling.cs
```diff
@@ -11,7 +11,9 @@ public partial class Control
 
         private readonly Dictionary<string, object> _styleProperties = new Dictionary<string, object>();
         private readonly HashSet<string> _styleClasses = new HashSet<string>();
+        private readonly HashSet<string> _stylePseudoClass = new HashSet<string>();
         public ICollection<string> StyleClasses { get; }
+        public ICollection<string> StylePseudoClass { get => _stylePseudoClass; }
```
PJB3005: Make this read only collection. The `StyleClasses` is the way it is to work with object initializers, but that doesn't apply here as pseudo classes should be protected set.

Robust.Client/UserInterface/Stylesheet.cs
```diff
@@ -148,7 +148,7 @@ public sealed class SelectorElement : Selector
             Type elementType,
             IReadOnlyCollection<string> elementClasses,
             string elementId,
-            string pseudoClass)
+            string[] pseudoClass)
```
PJB3005: This should be an `IReadOnlyCollection<string>` too.

Robust.Client/UserInterface/Stylesheet.cs
```diff
@@ -197,7 +198,7 @@ public override bool Matches(Control control)
                 }
             }
 
-            if (PseudoClass != null && PseudoClass != control.StylePseudoClass)
+            if (PseudoClass.Count > 0 && !PseudoClass.SetEquals(control.StylePseudoClass))
```
PJB3005: You shouldn't be doing a set equals here. You should be checking that it's a subset, like the regular style classes do.

Robust.Client/UserInterface/Control.Styling.cs
```diff
@@ -28,24 +30,39 @@ public string StyleIdentifier
...
 
-        private string _stylePseudoClass;
+        public bool HasStylePseudoClass(string className)
+        {
+            return _stylePseudoClass.Contains(className);
+        }
 
-        [ViewVariables]
-        public string StylePseudoClass
+        public void AddStylePseudoClass(string className)
```
PJB3005: This should be protected.

Robust.Client/UserInterface/Control.Styling.cs
```diff
@@ -28,24 +30,39 @@ public string StyleIdentifier
...
-                _stylePseudoClass = value;
-                Restyle();
+                return;
             }
+
+            _stylePseudoClass.Add(className);
+            Restyle();
         }
 
+        public void RemoveStylePseudoClass(string className)
```
PJB3005: Should be protected.

Robust.Client/UserInterface/Control.Styling.cs
```diff
@@ -28,24 +30,39 @@ public string StyleIdentifier
...
+        {
+            _stylePseudoClass.Remove(className);
+            Restyle();
+        }
+
+        public void SetOnlyStylePseudoClass(string className)
```
PJB3005: Should be protected.

Robust.Client/UserInterface/Control.Styling.cs
```diff
@@ -28,24 +30,39 @@ public string StyleIdentifier
...
+        {
+            if (_stylePseudoClass.Contains(className))
```
PJB3005: This won't give expected behavior if there's multiple elements in the pseudo class collection. Just remove this if.

Robust.Client/UserInterface/Control.Styling.cs
```diff
@@ -28,24 +30,39 @@ public string StyleIdentifier
...
+            {
+                return;
+            }
+
+            _stylePseudoClass.Clear();
+            _stylePseudoClass.Add(className);
```
PJB3005, line 65: This will insert a null if `className` is null. Passing null should just clear the list.

Robust.Client/UserInterface/Stylesheet.cs
```diff
@@ -166,13 +166,14 @@ public sealed class SelectorElement : Selector
...
-            PseudoClass = pseudoClass;
+            PseudoClass = pseudoClass != null ? new HashSet<string>(pseudoClass) : new HashSet<string>();
         }
 
         public Type ElementType { get; }
         public IReadOnlyCollection<string> ElementClasses { get; }
         public string ElementId { get; }
-        public string PseudoClass { get; }
+
+        public HashSet<string> PseudoClass;
```
PJB3005: I don't think it's that useful to use a hash set for this. Just do what the regular style classes do and store an `IReadOnlyCollection<string>`.

Also make it plural.

Robust.Client/UserInterface/Stylesheet.cs
```diff
@@ -208,7 +209,7 @@ public override bool Matches(Control control)
         public override StyleSpecificity CalculateSpecificity()
         {
             var countId = ElementId == null ? 0 : 1;
-            var countClasses = (ElementClasses?.Count ?? 0) + (PseudoClass == null ? 0 : 1);
+            var countClasses = (ElementClasses?.Count ?? 0) + (PseudoClass.Count == 0 ? 0 : 1);
```
PJB3005: You should take the amount of pseudo classes, not 0/1 depending on whether there are any.

Robust.Server/GameObjects/Components/Light/PointLightComponent.cs
```diff
@@ -66,7 +66,7 @@ public override void ExposeData(ObjectSerializer serializer)
 
             serializer.DataField(ref _enabled, "enabled", true);
             serializer.DataField(ref _color, "color", new Color(200, 200, 200));
-            serializer.DataField(ref _radius, "radius", 512);
+            serializer.DataField(ref _radius, "radius", 10);
```
PJB3005, line 69: Yeah I think you definitely didn't intend to commit this.

When working with git you're supposed to make a new branch for each pull request, so that your local `master` doesn't get messed up.

PJB3005 (changes requested):

Robust.Client/UserInterface/Control.Styling.cs
```diff
@@ -28,24 +30,40 @@ public string StyleIdentifier
...
+        {
+            _stylePseudoClass.Remove(className);
+            Restyle();
         }
 
+        protected void SetOnlyStylePseudoClass(string className)
+        {
+            _stylePseudoClass.Clear();
+
+            if (className == null)
```
PJB3005, line 59: You're still not running a restyle before returning here.

Robust.Client/UserInterface/Control.Styling.cs
```diff
@@ -28,24 +30,40 @@ public string StyleIdentifier
             }
         }
 
-        private string _stylePseudoClass;
+        public bool HasStylesPseudoClass(ICollection<string> collection)
```
PJB3005: Just keep this accepting a single string.

Robust.Client/UserInterface/Stylesheet.cs
```diff
@@ -208,7 +210,7 @@ public override bool Matches(Control control)
         public override StyleSpecificity CalculateSpecificity()
         {
             var countId = ElementId == null ? 0 : 1;
-            var countClasses = (ElementClasses?.Count ?? 0) + (PseudoClass == null ? 0 : 1);
+            var countClasses = (ElementClasses?.Count ?? 0) + (PseudoClass.Count == 0 ? 0 : PseudoClass.Count);
```
PJB3005:
```suggestion
            var countClasses = (ElementClasses?.Count ?? 0) + PseudoClass.Count;
```

Robust.Client/UserInterface/Stylesheet.cs
```diff
@@ -197,7 +199,7 @@ public override bool Matches(Control control)
                 }
             }
 
-            if (PseudoClass != null && PseudoClass != control.StylePseudoClass)
+            if (PseudoClass.Count > 0 &&  !control.HasStylesPseudoClass(_pseudoclass))
```
PJB3005: Just use a loop like the regular style classes do...

PJB3005 (changes requested):

Robust.Client/UserInterface/Control.Styling.cs
```diff
@@ -11,7 +11,9 @@ public partial class Control
 
         private readonly Dictionary<string, object> _styleProperties = new Dictionary<string, object>();
         private readonly HashSet<string> _styleClasses = new HashSet<string>();
+        private readonly HashSet<string> _stylePseudoClass = new HashSet<string>();
         public ICollection<string> StyleClasses { get; }
+        public IReadOnlyCollection<string> StylePseudoClass { get => _stylePseudoClass; }
```
PJB3005:
```suggestion
        public IReadOnlyCollection<string> StylePseudoClass => _stylePseudoClass;
```

Robust.Client/UserInterface/Stylesheet.cs
```diff
@@ -166,13 +166,15 @@ public sealed class SelectorElement : Selector
             ElementType = elementType;
             ElementClasses = elementClasses;
             ElementId = elementId;
-            PseudoClass = pseudoClass;
+            _pseudoclass = pseudoClass != null ? new HashSet<string>(pseudoClass) : new HashSet<string>();
```
PJB3005: There's no real reason to turn this into a hash set, since those aren't better at enumeration (which it's primarily used for). Just store the input directly.

PJB3005 (approved):

PR #847, opened 2019-08-17
Physics Shapes

### LINKED ISSUE #678: One bounding box is not cutting it.
Right now, collision and click bounding boxes are shared.

This is already problematic for humans, tables and doors.

They should be separated into their own deal.

Acruid:
This PR completely removes `BoundingBoxComponent` (BBComp). Currently the `ClickableComponent` (ClickComp) as well as the `CollidableComponent`(CollidableComp) share the same AABB in BBComp. This PR changes the design so that the collision AABB is now a property of the CollidableComp. ClickComp by default now uses the AABB of CollidableComp, or you can override this by providing an AABB to ClickComp through yaml or code. This feature resolves https://github.com/space-wizards/RobustToolbox/issues/678.

Previously a default CollidableComp was set to the Mob collision group, now by default it won't collide with anything.

Now the CollidableComp holds a PhysicsBody class, which is the representation of an entity in the physics system.

Currently a Physics Body was represented as a single AABB that resided on BoundingBoxComponent. This PR adds the concept of Physics Shapes, which are the physical 2D representation of the entity in the physics engine. These shapes can be any sort of primitive shape like a circle or rectangle, or meshes, or the collision volumes of a map grid. There can be multiple shapes on a single entity, and these shapes can rotate along with the entity. The Physics Body uses these shapes to calculate a bounding box for the entity.

Here are some examples of entities with a PhysBody (the red AABB), and a collection of PhysShapes (the green rotatable boxes) that make up the body:
PhysBody with single Shape PhysBody with two Shapes PhysBody with One rotated Shape

The CollisionGroup and CollisionMask are now a property of a physics shape, previously they were a property of the entire PhysBody.

Here is an example of how to set up the mob example above, inside YAML:
```yaml
  - type: Collidable # CollidableComponent, that holds the PhysBody
    shapes: # The set of PhysShapes that make up the PhysBody
    - !type:PhysShapeAabb # First shape, defines which shape class this is.
      bounds: "0.05,-0.25,0.5,0.25" # Local bounds of first shape (bottom, left, top, right).
      mask: 19 # Collision mask for first shape
      layer: 2 # Collision layer for first shape
    - !type:PhysShapeAabb # Second shape, defines which shape class this is.
      bounds: "-0.5,-0.25,-0.05,0.25" # Local bounds of second shape (bottom, left, top, right).
      mask: 19 # Collision mask for second shape
      layer: 2 # Collision layer for second shape
```

Things that were outside the scope of this PR:
* Narrow Phase collision detection. All collisions still use the PhysShape AABB, and the combination of all collision masks and groups.
* More primitive PhysShapes. Things like circles and convex hulls need debug drawing support in Clyde before I will implement them.


Robust.Server/Maps/MapLoader.cs
```diff
@@ -108,7 +108,15 @@ public IMapGrid LoadBlueprint(IMap map, string path)
...
 
                 var context = new MapContext(_mapManager, _tileDefinitionManager, _serverEntityManager, _pauseManager, (YamlMappingNode)data.RootNode, map);
-                context.Deserialize();
+                try
+                {
+                    context.Deserialize();
+                }
+                catch (Exception e)
+                {
+                    Logger.ErrorS("map", $"    Failed to deserialize map.");
```
PJB3005: This is really bad. This exception should 100% propagate, let the code doing the map load decide how to handle load failures.

Acruid: This was removed in commit 7daefbc, I guess Github did not detect it because of the rebase.

PJB3005 (approved):

PR #849, opened 2019-08-22
Added KeyBindClick and fix for KeyBind OnContextChanged

ShadowCommander:
Fixes a crash when clicking Quit from the Esc menu.
Shuffled some code around to implement KeyBindClick which is when a KeyBind is pressed and released on the same Control.


PJB3005:
I'm not sold on the click keybind. Seems like very little controls would be able to make good use of it.

PR #851, opened 2019-08-22
Add a check in OnContextChanged to fix some crashes and refactored BaseButton

ShadowCommander:
Fixes crashes due to double input. (e.g. the Quit button in the Esc menu.)


ShadowCommander:
Changed BaseButton to only trigger OnPressed when KeyBindDown and KeyBindUp are triggered on the same control.

PR #852, opened 2019-08-25
Things for wires

DamianX:


gradientvera (approved):

Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
```diff
@@ -272,4 +272,15 @@ private void AssertContains(IPlayerSession session)
             }
         }
     }
+
+    public class ServerBoundUserInterfaceMessage : BoundUserInterfaceMessage
+    {
+        public BoundUserInterfaceMessage Message;
```
PJB3005: Make these `{ get; }`

Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
```diff
@@ -272,4 +272,15 @@ private void AssertContains(IPlayerSession session)
             }
         }
     }
+
+    public class ServerBoundUserInterfaceMessage : BoundUserInterfaceMessage
```
PJB3005: I'm not sure there's a good reason to make these a subtype of `BoundUserInterfaceMessage`. Just seems like it's more confusing.

Robust.Shared.Maths/Color.cs
```diff
@@ -1795,5 +1796,18 @@ public enum BlendFactor
         };
 
         #endregion
+
+        [CanBeNull]
+        public string Name()
+        {
+            foreach (var entry in GetAllDefaultColors())
```
PJB3005: You should absolutely construct a dictionary for this. There's a few hundred colors so it would be extremely expensive to loop through all of them.

PJB3005 (changes requested):

DamianX:
- Removed Aqua and Magenta, as they were duplicates of Cyan and Fuchsia, respectively.
- There was no reason for `ServerBoundUserInterfaceMessage` to implement `BoundUserInterfaceMessage`
- Made the fields get-only properties

PJB3005 (approved):

PR #863, opened 2019-09-07
Improved ItemLists

gradientvera:


Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -62,143 +69,146 @@ private void RecalculateContentHeight()
             _updateScrollbarVisibility();
         }
 
-        public void AddItem(string text, Texture icon = null, bool selectable = true)
+        public void Add(Item item)
```
PJB3005, line 72: These `Item` instances should be made by the `ItemList` itself. With the current implementation you can theoretically add the same item to multiple lists at once, which will absolutely go haywire.

So more something like this:
```cs
public Item AddItem(string text, Texture icon = null, bool selectable = true)
```

Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -548,7 +548,58 @@ public enum ItemListSelectMode
...
+
+    public sealed class Item
+    {
+        public event Action<Item> OnSelected;
+        public event Action<Item> OnDeselected;
+
+        private bool _selected = false;
+        private bool _disabled = false;
+
+        public string Text;
```
PJB3005: All of these should be properties.

PJB3005 (changes requested):

Robust.Client/UserInterface/Controls/ItemList.cs
Zumorica: Done.

Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -69,7 +69,7 @@ private void RecalculateContentHeight()
             _updateScrollbarVisibility();
         }
 
-        public void Add(Item item)
+        void ICollection<Item>.Add(Item item)
```
PJB3005: This doesn't actually prevent the item from being added by external code.

gradientvera: Hmm, should I make Add do nothing, or just outright remove the IList<Item> interface from ItemList?

Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -548,7 +565,63 @@ public enum ItemListSelectMode
         {
             None,
             Single,
-            Multiple
+            Multiple,
+            Button,
+        }
+    }
+
+    public sealed class Item
```
PJB3005: Keep this as a nested class on `ItemList`

gradientvera: Done! As I've mentioned in a comment, apparently [indexers compile into a property called "Item" by default](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/indexers/using-indexers#remarks) so I had to use an attribute there.

PJB3005 (approved):

PR #866, opened 2019-09-10
Implemented SpinBox and added a filter to LineEdit

ShadowCommander:
Implemented SpinBox which is a number input using a LineEdit and a variable amount of Buttons.
Added a filter for input to LineEdits.
SpinBox.SetButtons takes two lists of ints and makes them into buttons surrounding the LineEdit.

I'd like some feedback on this.

### DIFF
--- a/Robust.Client/UserInterface/Controls/LineEdit.cs
+++ b/Robust.Client/UserInterface/Controls/LineEdit.cs
@@ -41,6 +41,11 @@ namespace Robust.Client.UserInterface.Controls
             KeyboardFocusOnClick = true;
         }
 
+        /// <summary>
+        ///     Determines whether the LineEdit text gets changed by the input text.
+        /// </summary>
+        public Func<string, bool> IsValid;
+
         public AlignMode TextAlign { get; set; }
 
         public string Text
@@ -53,7 +58,10 @@ namespace Robust.Client.UserInterface.Controls
                     value = "";
                 }
 
-                _text = value;
+                if (!SetText(value))
+                {
+                    return;
+                }
                 _cursorPosition = 0;
                 _updatePseudoClass();
             }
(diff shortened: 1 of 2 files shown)


Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -41,6 +41,11 @@ public LineEdit()
             KeyboardFocusOnClick = true;
         }
 
+        /// <summary>
+        ///     Determines whether the LineEdit text gets changed by the input text.
+        /// </summary>
+        public Func<string, bool> IsValid;
```
PJB3005: Make this a property.
```suggestion
        public Func<string, bool> IsValid { get; set; }
```

Robust.Client/UserInterface/Controls/SpinBox.cs
```diff
@@ -0,0 +1,115 @@
...
+using Robust.Shared.Maths;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+
+namespace Robust.Client.UserInterface.Controls
+{
+    /// <summary>
+    ///     Input for numbers.
+    /// </summary>
+    public class SpinBox : BoxContainer
```
PJB3005: Make this inherit from `HBoxContainer` instead and remove the `Vertical` override.

Robust.Client/UserInterface/Controls/SpinBox.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            ClearButtons();
+            AddLeftButton(-1, "-");
+            AddRightButton(1, "+");
+            UpdateButtonOrder();
+        }
+
+        public void AddRightButton(int num, string text)
+        {
+            var button = new Button { Text = text };
```
PJB3005, line 53: You can add these to the container and then change the order with `SetPositionInParent()`. Then you can add them immediately in this method instead of requiring an easy-to-miss `UpdateButtonOrder()` to be called.

Robust.Client/UserInterface/Controls/SpinBox.cs
```diff
@@ -0,0 +1,115 @@
...
+                AddChild(button);
+            }
+            AddChild(_lineEdit);
+            foreach (var button in _rightButtons)
+            {
+                AddChild(button);
+            }
+        }
+
+        protected internal override void MouseWheel(GUIMouseWheelEventArgs args)
```
PJB3005: I'm not sure I like the idea of being able to control these with mousewheel. It always carries the risk of "scrolling through a page and accidentally changing something" which is really bad.

PJB3005 (changes requested):

Robust.Client/UserInterface/Controls/SpinBox.cs
ShadowCommander: Ah, yeah, I forgot to add a check so that it only captures MouseWheel while the LineEdit is selected.

PJB3005 (approved):

PR #873, opened 2019-09-26
Adds Midi playback, Midi input, etc.

gradientvera:
Sister PR: https://github.com/space-wizards/space-station-14/pull/368
~Requires #877~

Also:
- Added OnClosed event to ServerUserInterfaceComponent
- Added CloseAll method to ServerUserInterfaceComponent

Special thanks to PJB for helping me so much with this project.


Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,220 @@
...
+    public interface IMidiManager
+    {
+        IMidiRenderer GetNewRenderer();
+        bool IsMidiFile(string filename);
+        bool IsSoundfontFile(string filename);
+    }
+
+    public class MidiManager : IPostInjectInit, IDisposable, IMidiManager
+    {
+        public static readonly string[] LinuxSoundfonts =
```
PJB3005: Make all of these private, and also make the implementation `internal`. You can keep the interface public though.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,220 @@
...
+        public static readonly string WindowsSoundfont = "c:\\WINDOWS\\system32\\drivers\\gm.dls";
+
+        public static readonly string OSXSoundfont =
+            "/System/Library/Components/CoreAudio.component/Contents/Resources/gs_instruments.dls";
+        public static readonly string FallbackSoundfont = "/Resources/Midi/fallback.sf2";
+
+#pragma warning disable 169
+        [Dependency] private IClydeAudio _clydeAudio;
+        [Dependency] private IResourceCache _resourceCache;
+#pragma warning enable 169
```
PJB3005: You're disabling the wrong warning here. CS0169 (the one you disabled) is "unused field". The one you want to disable is CS0649 (unassigned readonly field).

Muting the wrong warning here means that it won't be apparent if the field is unused at all, instead of simply ignoring the false positive caused by the compiler not recognizing `[Dependency]`.

Oh, also, the syntax for "enable" is actually "restore", which is probably why the syntax highlighter choked here.

It also can't hurt to mark these fields as readonly.

So apparently GitHub implemented multi-line review comments but didn't make multi-line suggestions, *sigh*:

```cs
#pragma warning disable 649
        [Dependency] private readonly IClydeAudio _clydeAudio;
        [Dependency] private readonly IResourceCache _resourceCache;
#pragma warning restore 649
```

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,220 @@
...
+            _settings["synth.threadsafe-api"].IntValue = 1;
+            _settings["audio.driver"].StringValue = "file";
+            _settings["midi.autoconnect"].IntValue = 1;
+            _settings["player.reset-synth"].IntValue = 0;
+
+            _midiThread = new Thread(ThreadUpdate);
+            _midiThread.Start();
+        }
+
+        /// <summary>
```
PJB3005: Put these doc comments in the interface declaration. Down here they're not so useful.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,220 @@
...
+        ///     This method returns a midi renderer ready to be used.
+        ///     You only need to set the <see cref="IMidiRenderer.MidiProgram"/> afterwards.
+        /// </summary>
+        public IMidiRenderer GetNewRenderer()
+        {
+            var renderer = new MidiRenderer(_settings, _soundFontLoader);
+            _renderers.Add(renderer);
```
PJB3005: `List<T>` is not thread safe. You should `lock(_renderers)` the list while mutating and reading it to prevent data races.

Also, it's probably best if you added the renderer to the list when all the sound font loading is done, not before. So the renderer thread won't start rendering before the sound fonts are correctly loaded.

As an example of what can go wrong, with the implementation of `List<T>` in CoreFX, adding to the list while you are doing a classic for loop over the list like down below can end with you landing [perfectly between the increment of the size and the actual setting of the element](https://source.dot.net/#System.Private.CoreLib/shared/System/Collections/Generic/List.cs,205), potentially causing the render thread to pull a null and throw a nullref. Yay!

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,220 @@
...
+        ///     Main method for the thread rendering the midi audio.
+        /// </summary>
+        private void ThreadUpdate()
+        {
+            while (_alive)
+            {
+                for (var i = 0; i < _renderers.Count; i++)
+                {
+                    var renderer = _renderers[i];
+                    if(renderer != null && !renderer.Rendering)
```
PJB3005: What could cause `renderer.Rendering` to be true here? As far as I can tell the only thing changing it is `Render` itself, which doesn't get called from any other thread, so it seems impossible to me for this to ever matter.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,220 @@
...
+                        renderer.Render();
+                }
+
+                Thread.Sleep(1);
+            }
+        }
+
+        public void Dispose()
```
PJB3005, line 204: It would probably be best wait on the renderer thread to stop with `.Join()` after setting `_alive` here, before running the rest of the shutdown.

Otherwise you might run into weird race conditions with the disposal of the renderers while they are being used.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,220 @@
...
+        bool IsMidiFile(string filename);
+        bool IsSoundfontFile(string filename);
+    }
+
+    public class MidiManager : IPostInjectInit, IDisposable, IMidiManager
+    {
+        public static readonly string[] LinuxSoundfonts =
+        {
+            "/usr/share/soundfonts/default.sf2",
+            "/usr/share/soundfonts/FluidR3_GM.sf2",
```
PJB3005, line 62: Does Fluidsynth smartly avoid loading the entire soundfont if you only need a few instruments?
`FluidR3_GM.sf2` is 130 MB which is absolutely *insane* to load into the game at once.

Probably best to just exclude it from this list, IMO.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,220 @@
...
+                    return IntPtr.Zero;
+                }
+
+                _file = new BinaryReader(stream);
+                _fileData = _file.ReadBytes((int) stream.Length);
+                _file.BaseStream.Position = 0;
+                fixed (byte* ptr = _fileData)
+                {
+                    return (IntPtr) ptr;
+                }
```
PJB3005: This is not at all how you should be using `fixed`. The point of `fixed` is that you can safely access the data via the pointer INSIDE the block. Once your code goes outside the fixed block again (the return), the GC could decide to move the array and your pointer would get invalidated... then again you're not actually using that pointers later anyways it seems so cool..

You need to either pin the array with `GCHandle` to ensure the pointer remains valid, or copy the data to unmanaged memory (allocated/freed via `Marshal.AllocCoTaskMem`/`Marshal.FreeCoTaskMem`) where this isn't an issue.

That said, none of those solutions are ideal either. You're loading the ENTIRE sound font from disk in one read call. Depending on how the disk feels this can take quite a while, and the GC won't be very happy with you either. This loader callbacks API is designed explicitly so that you can wrap the underlying `Stream` provided by the OS with the read/seek calls, so fluidsynth can request the data in smaller chunks and avoid the aforementioned issues.

TL;DR keep the `Stream` around, map the `Read`/`Seek`/`Tell` callbacks onto the C# stream, get good performance.

By the way, the handle you're returning here is something for how C APIs work. C doesn't have capturing closures like C# does so they're necessary. They're not necessary with C# however so you can just return `IntPtr.Zero` here and then keep ignoring `IntPtr sfHandle` like you are below.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+        ///     Start playing a midi file.
+        /// </summary>
+        /// <param name="filename">Path to the midi file</param>
+        void OpenMidi(string filename);
+
+        /// <summary>
+        ///     Start playing a midi file.
+        /// </summary>
+        /// <param name="stream">Bytes of the midi file</param>
+        void OpenMidi(Span<byte> stream);
```
PJB3005: Slight nitpick: `Span<byte>` is a buffer, not a stream.

Also, you can make this use `ReadOnlySpan`, looking at the implementation.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+
+        internal bool Free { get; set; } = false;
+
+        internal MidiRenderer(Settings settings, SoundFontLoader soundFontLoader)
+        {
+            _settings = settings;
+            _soundFontLoader = soundFontLoader;
+            _synth = new Synth(_settings);
+            _synth.AddSoundFontLoader(_soundFontLoader);
+            _source = AL.GenSource();
```
PJB3005: The OpenAL calls should really be isolated to Clyde.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+            AL.SourcePlay(_source);
+        }
+
+        private unsafe void EmptyBuffers()
+        {
+            var length = (SampleRate / Buffers) * (Mono ? 1 : 2);
+
+            for (var i = 0; i < Buffers; i++)
+            {
+                var empty = new Span<ushort>(new ushort[length]);
```
PJB3005: Move the array creation and pinning out of the loop, so it only has to happen once. It'll improve performance and reduce GC pressure.

Also, you don't need to wrap the array in a `Span<>` to be able to pin it.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+                fixed (ushort* ptr = empty)
+                {
+                    AL.BufferData(_buffers[i], Mono ? ALFormat.Mono16 : ALFormat.Stereo16, (IntPtr) ptr, length*sizeof(ushort), SampleRate);
+                    AL.SourceQueueBuffers(_source, 1, new []{_buffers[i]});
```
PJB3005: Can't you do this for every buffer at once, after the for loop?

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+
+        /// <summary>
+        ///     Whether the renderer is listening for midi input or not.
+        /// </summary>
+        bool IsInputOpen { get; }
+
+        /// <summary>
+        ///     Whether a midi file is being played or not.
+        /// </summary>
+        bool IsMidiOpen { get; }
```
PJB3005: Are you ever supposed to be able to play both a midi file and from input with the same renderer? If not, make these two properties represented with an enum and make the methods like `OpenInput` fail if something is already running.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+
+        internal MidiRenderer(Settings settings, SoundFontLoader soundFontLoader)
+        {
+            _settings = settings;
+            _soundFontLoader = soundFontLoader;
+            _synth = new Synth(_settings);
+            _synth.AddSoundFontLoader(_soundFontLoader);
+            _source = AL.GenSource();
+            _buffers = AL.GenBuffers(Buffers);
+            EmptyBuffers();
```
PJB3005: Wouldn't filling all the buffers initially with empty just cause the audio to be delayed for one second, or does that not happen?

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+                        AL.BufferData(buffer, Mono ? ALFormat.Mono16 : ALFormat.Stereo16, (IntPtr) ptr, Mono ? length * sizeof(ushort) : length*sizeof(ushort)*2, SampleRate);
+                    }
+                }
+
+                AL.SourceQueueBuffers(_source, buffersProcessed, buffers);
+            }
+
+            if (IsMidiOpen && _player.Status == FluidPlayerStatus.Done)
+            {
+                OnMidiPlayerFinished?.Invoke();
```
PJB3005: Right now this will be invoked from the audio thread. It would probably be wise to run this on the main thread with `ITaskManager.RunOnMainThread`.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+                    fixed (ushort* ptr = left)
+                    {
+                        AL.BufferData(buffer, Mono ? ALFormat.Mono16 : ALFormat.Stereo16, (IntPtr) ptr, Mono ? length * sizeof(ushort) : length*sizeof(ushort)*2, SampleRate);
+                    }
+                }
+
+                AL.SourceQueueBuffers(_source, buffersProcessed, buffers);
+            }
+
+            if (IsMidiOpen && _player.Status == FluidPlayerStatus.Done)
```
PJB3005: Here's a fun race condition for you: Since this block of code doesn't lock the player, it's possible for somebody to run `CloseMidi()` between `IsMidiOpen` and `_player.Status == FluidPlayerStatus.Done`, thus causing a null reference exception.

You should be able to solve this one by just wrapping the if statement in a `lock (_player)`. C#'s lock statement is re-entrant so it won't cause a problem with the lock happening inside `CloseMidi`.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+                {
+                    var buffer = buffers[i];
+
+                    ushort[] left;
+                    if (Mono)
+                    {
+                        left = new ushort[length];
+                        var right = new ushort[length];
+                        _synth?.WriteSample16( left, 0, 1, right, 0, 1);
+                        left = left.Zip(right, (x, y) => (ushort) (x + y)).ToArray();
```
PJB3005: Shouldn't you also be dividing these by two to average them, or is that not how audio works?

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+                var buffers = AL.SourceUnqueueBuffers(_source, buffersProcessed);
+
+                for (var i = 0; i < buffers.Length; i++)
+                {
+                    var buffer = buffers[i];
+
+                    ushort[] left;
+                    if (Mono)
+                    {
+                        left = new ushort[length];
```
PJB3005: You should probably *really* cut down on the amount of allocations here. Right now this code allocates 200,000 bytes per second for a single MIDI renderer, and it has to be low latency too. That's not accounting for the LINQ either.

The easiest way is probably to use `stackalloc` to just allocate a buffer, since they're quite small anyways. You can re-use it outside the for loop too.

Also note, in both the Mono and Stereo case your buffers add up to the same length, so you can always `Span<ushort> samples = stackalloc ushort[length*2]`.

For the Mono case, just give the right audio channel the top half and then merge it with a plain for loop, and only give OpenAL the bottom half when you're done. For the stereo case you can just interleave it like you're doing right now.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+                    default:
+                        if (_notesPlaying.Contains(midiEvent.Key))
+                        {
+                            _synth.NoteOff(ch, midiEvent.Key);
+                            _notesPlaying.Remove(midiEvent.Key);
+                        }
+                        break;
+                }
+            }
+            catch (FluidSynthInteropException interopException)
```
PJB3005: Why are you ignoring exceptions here?
If you have a good reason, please do elaborate in a comment.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+                    // NoteOn
+                    case 144:
+                        // NoteOn with 0 velocity is the same as NoteOff
+                        if (midiEvent.Velocity == 0)
+                        {
+                            if (_notesPlaying.Contains(midiEvent.Key))
+                            {
+                                _synth.NoteOff(ch, midiEvent.Key);
+                                _notesPlaying.Remove(midiEvent.Key);
+                            }
```
PJB3005: You can replace this with `goto case 128;`

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,383 @@
...
+                                return;
+
+                            _synth.NoteOn(ch, midiEvent.Key, midiEvent.Velocity);
+                            if (!_notesPlaying.Contains(midiEvent.Key))
+                                _notesPlaying.Add(midiEvent.Key);
+                        }
+
+                        break;
+                    default:
+                        if (_notesPlaying.Contains(midiEvent.Key))
```
PJB3005: This seems identical to the note off? In that case you should probably just remove the `case 128` entirely and let this default handle it.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
```diff
@@ -138,6 +138,8 @@ private void _updateAudio()
         public IClydeAudioSource CreateAudioSource(AudioStream stream)
         {
             var source = AL.GenSource();
+            if (_audioSources.ContainsKey(source))
```
PJB3005: Uuuh, what? How could this ever pass sanely?

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -39,7 +39,7 @@ public override void RegisterMessageTypes()
         public override void FrameUpdate(float frameTime)
         {
             // Update positions of streams every frame.
-            foreach (var stream in PlayingClydeStreams)
+            foreach (var stream in PlayingClydeStreams.ToArray())
```
PJB3005: Why was this change necessary? The contents of the loop here really shouldn't be adding or removing any streams.

PJB3005 (dismissed):

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Nope, dividing by two would mess up the sound. It's a bit unintuitive, but this has to do with the superposition principle since audio is essentialy a wave

Robust.Client/Audio/Midi/MidiManager.cs
gradientvera: Done.

Robust.Client/Audio/Midi/MidiManager.cs
gradientvera: Ah, fuck. I screwed up here, yeah. (And in the end I didn't even need those dependencies...)

Robust.Client/Audio/Midi/MidiManager.cs
gradientvera: Ow. I locked the list of renderers, and also added the new renderer after the soundfont has been loaded. Should be good now.

Robust.Client/Audio/Midi/MidiManager.cs
gradientvera: Done.

Robust.Client/Audio/Midi/MidiManager.cs
gradientvera: Yeah, I'm not actually using 'rendering' anymore... ...But I forgot to remove it. Whoops. 
I removed it now tho.

Robust.Client/Audio/Midi/MidiManager.cs
gradientvera: Done!

Robust.Client/Audio/Midi/MidiManager.cs
gradientvera: Uh, seems I misunderstood what fixed does, then... Should be _mostly_ good now, I ended up using the code you wrote for this. (However there's still the "double linked list corruption crash" issue whenever you delete a few instruments... I fear it has to do with this class, somehow)

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: That's funny! The argument in the implementation is called "buffer", yet the one in the interface is called "stream"... At first, OpenMidi used `Stream` instead of `Span`, so that might be the reason. Done.

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Hmm, how should I go about this? Should I wrap some OpenAL stuff in Clyde for "low-level access" to the audio system?

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Why didn't I think of that before... Done now

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Oh, indeed. Done now.

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Yeah, you're supposed to either be playing from a midi file, playing from midi input or none. Done now.

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: I think it adds a very minor delay, but unless I do this the source automatically stops before any audio is rendered at all...

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Oh gosh, I wish I knew about that much sooner... Done.

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Ow, seems I still need to play some more Deadlock Empire! Done now.

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: This is now done, thanks to the changes you made to WriteSample16!

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: At first I was ignoring all exceptions because trying to send a NoteOff event for a key that's not on threw an exception... However, after adding the list of played notes this isn't the case anymore. Furthermore, I think a fluidsynth exception shouldn't crash the game, so I think I'll add a log message here instead of throwing the exception. Is that fine?

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Done.

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Indeed it is. I moved the case 128 so it's the same as default now:
`case 128:`
`default:`
`[...]`

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
gradientvera: ...multi-threading. Yeah, I'll change these back since I don't even use clyde.audio directly anymore.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
gradientvera: Changed this one as well.

Robust.Client/Audio/Midi/MidiManager.cs
gradientvera: The problem is that FluidR3_GS.sf2 doesn't include most instruments, as it's an extension to FluidR3_GM.sf2... And freepats is 190mb. Honestly, this is a bit of a mess. Got any ideas for this?

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,258 @@
...
+            }
+
+            public override int Close(IntPtr sfHandle)
+            {
+                var stream = _openStreams[(int) sfHandle];
+                stream.Dispose();
+                _openStreams.Remove((int) sfHandle);
+                return 0;
+            }
+        }
```
gradientvera, line 289: Something is seriously wrong with this class. Disposing a few instruments triggers a double-linked list corruption when using this.

Robust.Client/Audio/Midi/MidiRenderer.cs
PJB3005: Yeah, that's fine.

Robust.Client/Audio/Midi/MidiManager.cs
PJB3005: Damnit...

At that point I'd rather dynamically download the sound font from our server and default to fallback.

Robust.Client/Audio/Midi/MidiManager.cs
PJB3005: I did some googling. Looks like an issue related to double frees etc?

I have no idea why the code would be doing double frees here, though..

elwray: Maybe because IntPtr has a 64 bit size and you cast it to 32 bit integer?

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,258 @@
...
+                }
+
+                var id = _nextStreamId++;
+
+                _openStreams.Add(id, stream);
+
+                return (IntPtr) id;
+            }
+
+            public override unsafe int Read(IntPtr buf, long count, IntPtr sfHandle)
```
elwray, line 247: Does methods `Read` and `Seek` should return 0 or actual number of bytes that read / skipped?

gradientvera: Both return 0 ([FLUID_OK](http://www.fluidsynth.org/api/misc_8h.html#ae4efb1c3ce0d550c922504adfb0fb886)) if the read/seek operation executed correctly. ([API reference](http://www.fluidsynth.org/api/sfont_8h.html#a5ad574cbab9c9413932d51faaa5d10a8))

Robust.Client/Audio/Midi/MidiManager.cs
gradientvera: Well, seems the problem wasn't this class, but SoundfontLoader being deleted by Synth on their own deletion. I'm resolving this one, for now.

Robust.Client/Audio/Midi/MidiRenderer.cs
PJB3005: Probably? As long as it's semi sane and OpenAL is specific to Clyde I'm fine with it.

gradientvera: Resolving this in favor of #877

gradientvera:
I guess the only thing missing from this now is... Solving the "issue" with soundfonts on linux.

We've already discussed this a bit, and we concluded that shipping a big soundfont with the repo isn't ideal, and that the common soundfonts used on linux are way too big. 
Since the soundfont needs to be loaded into memory at least once, it would be ideal if it was small. Maybe less than 10mb.

Does anyone have any suggestions as to what to do with this issue?

io-june:
ship soundfonts ourselves

gradientvera:
> ship soundfonts ourselves

Maybe the launcher could download a soundfont from the builds server or something?

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -0,0 +1,262 @@
...
+        /// </summary>
+        IMidiRenderer GetNewRenderer();
+
+        /// <summary>
+        ///     This method is used to dispose a renderer.
+        ///     You should use this instead of disposing the renderer manually.
+        /// </summary>
+        /// <param name="renderer"></param>
+        /// <returns>Whether it could be returned or not</returns>
+        bool DisposeRenderer(IMidiRenderer renderer);
```
PJB3005: You really shouldn't have this separate dispose method. It should all be handled by `MidiRenderer.Dispose`

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,424 @@
...
+                _synth.SoundFontSelect(i, 1);
+        }
+
+        public event Action<Shared.Audio.Midi.MidiEvent> OnMidiEvent;
+        public event Action OnMidiPlayerFinished;
+
+        internal void Render(int length = SampleRate/1000)
+        {
+            if(Mono && Position != null)
+                lock(Position)
```
PJB3005: This lock is not at all enough to avoid thread safety issues here, since no other code actually locks entities (and it shouldn't, it'd kill performance).

You should update the position from a frame update from the main loop instead, like how `AudioSystem` does it.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,424 @@
...
+                    }
+
+                    _audioSource.WriteBuffer(buffer, audio);
+                }
+
+                _audioSource.QueueBuffers(buffers);
+            }
+
+            if(_player != null)
+                lock(_player)
```
PJB3005: You still aren't correctly locking the state related to the synth or the player here.

You should make a separate plain `object`, and use that as a guard to lock all the state related to the player and the synth.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -0,0 +1,424 @@
...
+                                _synth.NoteOff(ch, midiEvent.Key);
+                                _notesPlaying.Remove(midiEvent.Key);
+                            }
+
+                        break;
+                }
+            }
+            catch (FluidSynthInteropException e)
+            {
+                _taskManager.RunOnMainThread(() => Logger.ErrorS("midi",
```
PJB3005: Logging is thread safe actually, if you store the `ISawmill` to log to somewhere when creating the midi renderer.

PJB3005 (changes requested):

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Should be good now.

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Done, I think.

Robust.Client/Audio/Midi/MidiManager.cs
gradientvera: I removed this method, but I still needed to remove disposed/null renderers from the renderer list, so I let the thread method handle that. Should be alright, but please let me know if you can think of a better solution!

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: This is done now.

PJB3005 (approved):

PR #877, opened 2019-10-13
Add BufferedAudioSource to Clyde.Audio

gradientvera:
Required by #873

### DIFF
--- a/Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
+++ b/Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
@@ -33,6 +33,7 @@ namespace Robust.Client.Graphics.Clyde
         // so we need to properly send them off in the main thread.
         private readonly ConcurrentQueue<int> _sourceDisposeQueue = new ConcurrentQueue<int>();
         private readonly ConcurrentQueue<uint> _bufferedSourceDisposeQueue = new ConcurrentQueue<uint>();
+        private readonly ConcurrentQueue<uint> _bufferDisposeQueue = new ConcurrentQueue<uint>();
 
         private bool _alcHasExtension(string extension) => _alcExtensions.Contains(extension);
         private bool _alContentHasExtension(string extension) => _alContextExtensions.Contains(extension);
@@ -153,6 +154,13 @@ namespace Robust.Client.Graphics.Clyde
                 AL.DeleteSource((int)handle);
                 _checkAlError();
                 _bufferedAudioSources.Remove(handle);
+            }
+
+            // Clear out finalized audio buffers.
+            while (_bufferDisposeQueue.TryDequeue(out var handle))
+            {
+                AL.DeleteBuffer((int)handle);
+                _checkAlError();
             }
         }
 
@@ -318,6 +326,11 @@ namespace Robust.Client.Graphics.Clyde
         private void DeleteBufferedSourceOnMainThread(uint bufferedSourceHandle)
         {
             _bufferedSourceDisposeQueue.Enqueue(bufferedSourceHandle);
+        }
+
+        private void DeleteAudioBufferOnMainThread(uint bufferHandle)
+        {
+            _bufferDisposeQueue.Enqueue(bufferHandle);
         }
 
         private sealed class AudioSource : IClydeAudioSource
@@ -527,11 +540,14 @@ namespace Robust.Client.Graphics.Clyde
                 {
                     // We can't run this code inside the finalizer thread so tell Clyde to clear it up later.
                     _master.DeleteBufferedSourceOnMainThread(SourceHandle.Value);
+                    for (var i = 0; i < BufferHandles.Length; i++)
+                        _master.DeleteAudioBufferOnMainThread(BufferHandles[i]);
                 }
                 else
                 {
 
                     AL.DeleteSource((int)SourceHandle.Value);
+                    AL.DeleteBuffers(BufferHandles);
                     _master._bufferedAudioSources.Remove(SourceHandle.Value);
                     _checkAlError();
                 }
--- a/Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
+++ b/Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
@@ -555,18 +555,15 @@ namespace Robust.Client.Graphics.Clyde
                 return buffersProcessed;
             }
 
-            public unsafe Span<uint> GetBuffersProcessed()
+            public unsafe void GetBuffersProcessed(Span<uint> handles)
             {
                 _checkDisposed();
-                var span = new Span<uint>();
-                fixed(uint* ptr = span)
+                fixed(uint* ptr = handles)
                     // ReSharper disable once PossibleInvalidOperationException
-                    AL.SourceUnqueueBuffers(SourceHandle.Value, GetNumberOfBuffersProcessed(), ptr);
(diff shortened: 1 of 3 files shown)


Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
```diff
@@ -390,5 +426,143 @@ private void _checkDisposed()
...
+                }
+            }
+
+            public int GetNumberOfBuffersProcessed()
+            {
+                AL.GetSource(SourceHandle, ALGetSourcei.BuffersProcessed, out var buffersProcessed);
+                return buffersProcessed;
+            }
+
+            public int[] GetBuffersProcessed()
```
PJB3005: I would really prefer if there was a variant of this API that does not allocate, for example something that writes into a `Span<int>`.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
```diff
@@ -390,5 +426,143 @@ private void _checkDisposed()
...
+            public unsafe void WriteBuffer(int handle, ReadOnlySpan<ushort> data)
+            {
+                fixed (ushort* ptr = data)
+                {
+                    AL.BufferData(handle, _mono ? ALFormat.Mono16 : ALFormat.Stereo16, (IntPtr) ptr,
+                        _mono ? data.Length/2 * sizeof(ushort) : data.Length * sizeof(ushort), SampleRate);
+                }
+            }
+
+            public void QueueBuffers(int[] handles)
```
PJB3005: Again, this should be doable without allocations by passing a `ReadOnlySpan<int>`

Robust.Client/Graphics/Clyde/ClydeHeadless.cs
```diff
@@ -143,6 +148,73 @@ public void SetVolume(float decibels)
             }
         }
 
+        private sealed class DummyBufferedAudioSource : IClydeBufferedAudioSource
```
PJB3005: Make this inherit `DummyAudioSource` maybe?

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
gradientvera: Done, although I changed all those to use `Span<uint>` instead.

Robust.Client/Graphics/Clyde/ClydeHeadless.cs
gradientvera: Done as well.

gradientvera:
I also mapped all buffer handles internally so we don't leak any OpenAL buffer handles, and added some simple checks for those.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
```diff
@@ -505,56 +522,89 @@ public void Dispose()
...
             {
-                AL.GetSource(SourceHandle, ALGetSourcei.BuffersProcessed, out var buffersProcessed);
+                _checkDisposed();
+                // ReSharper disable once PossibleInvalidOperationException
+                AL.GetSource(SourceHandle.Value, ALGetSourcei.BuffersProcessed, out var buffersProcessed);
                 return buffersProcessed;
             }
 
-            public int[] GetBuffersProcessed()
+            public unsafe Span<uint> GetBuffersProcessed()
```
PJB3005: You can't actually return a span like this. A span is just a section of memory. Note how you aren't actually allocating any memory to put the data in here.

You will need to have this function *take in* a `Span<uint>` and write `min(span.Length, GetNumberOfBuffersProcessed())` into this span. The user of the API can use `GetNumberOfBuffersProcessed()` themselves to correctly allocate enough space for all the buffers.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
```diff
@@ -505,56 +522,89 @@ public void Dispose()
...
                 }
             }
 
-            public void QueueBuffers(int[] handles)
+            public unsafe void QueueBuffers(ReadOnlySpan<uint> handles)
             {
-                AL.SourceQueueBuffers(SourceHandle, handles.Length, handles);
+                _checkDisposed();
+
+                var realHandles = new Span<uint>();
```
PJB3005: Again, creating a new `Span<>` won't allocate any memory for you, so this will fail.

What you want to do here is create `realHandles` like so, using `stackalloc` to actually allocate space:

```cs
Span<uint>() realHandles = stackalloc uint[handles.Length];
```

This is significantly cheaper than a full array allocation still, and works with dynamic sizes.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
```diff
@@ -505,56 +522,89 @@ public void Dispose()
...
+
+                for (var i = 0; i < realHandles.Length; i++)
+                {
+                    var handle = realHandles[i];
+                    if(handle >= BufferHandles.Length)
+                        throw new ArgumentOutOfRangeException(nameof(handles), $"Invalid handle with index {i}!");
+                    realHandles[i] = BufferHandles[handle];
+                }
+
+                fixed(uint* ptr = handles)
```
PJB3005: I believe you mean to pass `realHandles` here.

PJB3005 (changes requested):

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
gradientvera: Whoops. Fixed now...

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
gradientvera: Done! (I also updated the midi pr to use these changes)

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
gradientvera: Done as well!

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
```diff
@@ -555,18 +555,15 @@ public int GetNumberOfBuffersProcessed()
...
                     // ReSharper disable once PossibleInvalidOperationException
-                    AL.SourceUnqueueBuffers(SourceHandle.Value, GetNumberOfBuffersProcessed(), ptr);
-
-                for (var i = 0; i < span.Length; i++)
-                    span[i] = BufferMap[span[i]];
+                    AL.SourceUnqueueBuffers(SourceHandle.Value, Math.Min(handles.Length, GetNumberOfBuffersProcessed()), ptr);
 
-                return span;
+                for (var i = 0; i < handles.Length; i++)
```
PJB3005: You're only writing to the bottom `Math.Min(handles.Length, GetNumberOfBuffersProcessed())` of the buffers handles here, so you should only map that lower section.

gradientvera: Oh, you're right... Fixed now!

gradientvera:
I just made it delete buffers on dispose, since I'm pretty sure deleting the source isn't enough to free them.

PJB3005 (approved):

PR #884, opened 2019-11-10
Code smell for loop braces

Rockdtben:
Fixes 7 Code Smells

### DIFF
--- a/Robust.Client/Console/Commands/Debug.cs
+++ b/Robust.Client/Console/Commands/Debug.cs
@@ -268,7 +268,7 @@ namespace Robust.Client.Console.Commands
             var offset = SnapGridOffset.Center;
             if (args.Length == 3)
             {
-                offset = (SnapGridOffset) Enum.Parse(typeof(SnapGridOffset), args[2]);
+                offset = (SnapGridOffset)Enum.Parse(typeof(SnapGridOffset), args[2]);
             }
 
             var mapMan = IoCManager.Resolve<IMapManager>();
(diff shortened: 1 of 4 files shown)


Robust.Client/Console/Commands/Debug.cs
```diff
@@ -312,10 +312,10 @@ public bool Execute(IDebugConsole console, params string[] args)
 
             var type = reflection.LooseGetType(args[1]);
             var getResourceMethod =
-                resourceCache.GetType().GetMethod("GetResource", new[] {typeof(string), typeof(bool)});
+                resourceCache.GetType().GetMethod("GetResource", new[] { typeof(string), typeof(bool) });
```
PJB3005, line 315: Man we really need to configure IDE formatters to be consistent on this. I know my IDE auto format (Rider) will completely undo these changes.

Rockdtben: I think there is a way to configure the style settings per solution in Visual Studio.

Is there a standard auto formatting you want for this project?

PJB3005: I don't have too much of a preference on this kind of stuff, but it definitely should be consistent.

Rider can also pull formatting config from the solution (and also editorconfig). There are some general C# specific rules you can put into editorconfig which I believe VS should listen to aswell, but the one that would be relevant here is `csharp_space_within_single_line_array_initializer_braces` which only Rider/Resharper seem to care about.

PR #885, opened 2019-11-12
Fix sprite jittering.

PJB3005:
@Acruid I would really like your opinion on the way I fixed this, since personally I think this fix is ugly as hell, but I can't think of much better.

### DIFF
--- a/Robust.Client/GameObjects/Components/Eye/EyeComponent.cs
+++ b/Robust.Client/GameObjects/Components/Eye/EyeComponent.cs
@@ -111,7 +111,9 @@ namespace Robust.Client.GameObjects
         public void UpdateEyePosition()
         {
             var mapPos = Owner.Transform.MapPosition;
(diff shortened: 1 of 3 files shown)


d34d10cc:
Personally, I think this is as elegant as it's gonna get for this kind of problem.

Acruid:
The core problem is that the world coordinates have sub-pixel interval resolution (1/32m with no zoom, game currently uses 2x), so as the view moves the sprites get rounded between the two pixel intervals at different offsets. The rounding is required to keep pixel-perfect sprite rendering.

You can reproduce this effect by standing still and offsetting your Eye downwards by very small increments, the item sprites on the table will appear to move down with the view until they round up to the next pixel, the items round at different offsets.

The camera zoom, camera worldpos, camera rotation, RSI layer offset, sprite worldpos, and sprite rotation all work together to cause this issue.

This PR seems abandoned so I am closing it.

PR #898, opened 2019-11-21
Implement SpawnRepeating

ShadowCommander:
Allows repeating timer events to be used.

### DIFF
--- a/Robust.Shared/Timers/Timer.cs
+++ b/Robust.Shared/Timers/Timer.cs
@@ -128,12 +128,11 @@ namespace Robust.Shared.Timers
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Robust.Shared/Timers/Timer.cs
```diff
@@ -126,5 +126,30 @@ public static void Spawn(TimeSpan duration, Action onFired, CancellationToken ca
         {
             Spawn((int)duration.TotalMilliseconds, onFired, cancellationToken);
         }
+
+        /// <summary>
+        ///     Schedule an action to be fired after a certain delay.
```
PJB3005: This doc comment doesn't seem very correct.

PR #904, opened 2019-11-25
Shared IMap/Map class removal

Acruid:
This PR removes the Map class and the IMap interface from the game. Now a game Map in code is purely represented by a MapId value, instead of an object.

The existing IMap object was mainly being used as a proxy interface to the functions that already existed in IMapManager, and a convoluted way to pass around a MapId.

This is purely a refactoring PR, no logic should have been changed.


PJB3005:
Personally I quite like having an OOP API like that :man_shrugging:. That said I can definitely see why because it didn't contain all that many useful things.

Are you planning to do the same with MapGrid?

PJB3005 (approved):

Acruid:
So this is why I broke the removal into two commits, so that if we wanted the last commit can be reverted and the Map implementation can be used purely for a proxy to MapManager. One thing that makes the proxy class useful is that it is a way to pass around a (MapId, MapManager) tuple, though in the case of our code base there was no instance where MapManager was not available at the call site. This is similar to the Entity proxy class, which is basically a (EntityUid, EntityManager) tuple that gets passed around so that you don't need to have a handle to EntityManager at the call site.

Originally the Manager held a set of maps, and each map held a set of grids. This was changed awhile back for valid reasons, so that now the Manager holds a set of grids, and the MapId is a property of each grid (I think, that HashSet is sketchy!). `Defaultness` of a grid is a property on a grid, and with Map having no other properties besides the completely internal mapCreationTick, this is what caused Map to become a proxy/useless class. MapGrid still holds a set of MapChunks, and has properties like world position, nothing will be changed with it.

My other concern was that the Map class basically duplicates the entire MapManager grid API, meaning there are multiple ways to modify grids, which makes the whole thing more confusing to the average user trying to learn our API. In the ComponentManager/Entity proxy case, we are basically running on the idea that ComponentManager API does not exist (does anything even get the IComponentManager from IoC?), but it has the same underlying problem of causing API complexity.

Thank you for coming to my TED Talk, please give me all your money, have a nice day!

PR #909, opened 2019-12-05
EventBus Refactor

Acruid:
Extracts Entity Event code out of EntityManager into a new class. Adds XML docs and Full test coverage to the new class. Removed the old (now proxy) event methods from EntityManager. The new EventBus class is now a public member of the EntityManager.

This is a refactor, no logic should be changed.


Robust.Client/Player/LocalPlayer.cs
```diff
@@ -89,7 +90,8 @@ public void AttachEntity(IEntity entity)
             entity.SendMessage(null, new PlayerAttachedMsg());
 
             // notify ECS Systems
-            ControlledEntity.EntityManager.RaiseEvent(this, new PlayerAttachSysMessage(ControlledEntity));
+            EntityEventArgs toRaise = new PlayerAttachSysMessage(ControlledEntity);
+            ControlledEntity.EntityManager.EventBus.RaiseEvent((object) this, toRaise);
```
PJB3005: Why are you explicitly casting these to `object` and `EntityEventArgs`?

Acruid: Looks like an artifact of R# refactoring, I will clean up the redundant code.

PR #910, opened 2019-12-06
Main menu parallax

ghost:
- Main menu now has a nebula in the background that moves against your cursor's position

UI & rendering learning experience, looks pretty, will get deleted in the future anyway

https://gfycat.com/ScentedBlueLacewing

### DIFF
--- a/Robust.Client/State/States/MainMenu.cs
+++ b/Robust.Client/State/States/MainMenu.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Text.RegularExpressions;
 using Robust.Client.Graphics;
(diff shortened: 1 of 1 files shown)


ghost:
oh no that png is actually huge relative to the rest of the engine

ghost:
that's better

PJB3005:
This **really** shouldn't have been merged...

I'm pretty sure I stated *why* at the time on Discord, so I really should have mentioned it here on GitHub and just closed it...

1. That's a 1 MB PNG file, *in the engine*.
2. This is the engine, this kind of stuff should be part of content.

PR #913, opened 2019-12-12
Expose IWritableDirProvider.RootDir

DamianX:
I need this to provide a filesystem path to SQLite in https://github.com/space-wizards/space-station-14/pull/465

### DIFF
--- a/Robust.Shared/ContentPack/WritableDirProvider.cs
+++ b/Robust.Shared/ContentPack/WritableDirProvider.cs
@@ -9,7 +9,8 @@ namespace Robust.Shared.ContentPack
     /// <inheritdoc />
     internal class WritableDirProvider : IWritableDirProvider
     {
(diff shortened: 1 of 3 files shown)


Robust.Shared/ContentPack/WritableDirProvider.cs
```diff
@@ -9,7 +9,8 @@ namespace Robust.Shared.ContentPack
     /// <inheritdoc />
     internal class WritableDirProvider : IWritableDirProvider
     {
-        private readonly string _rootDirString;
+        /// <inheritdoc />
+        public string RootDir { get; private set; }
```
PJB3005: You can remove the `private set;` and it will act like a `readonly` like before.

PJB3005:
I wonder if it would make sense to hide this info from content on the client. Probably something to worry about later though.

PR #916, opened 2019-12-17
Map & Grid Entities

Acruid:
Adds entities to the scene tree that represent the Map and Grids. The Map entity is always the root node of the tree, by convention Grid entities are always parented to the map entity. This allows us to use the transform parenting system to easily manipulate large branches of objects around the tree. This PR removes the need for the "DefaultGrid", which was an artifact of the original coordinate system. Entire grids and all their attached entities can be moved between maps by just changing their parent.

A normal scene graph looks something like this now:
```
MapEntity
   |- GridEntity
   |    |- StationObjectEntity
   |    \- StationObjectEntity
   |- FloatingObjectEntity
   \- FloatingObjectEntity
```
With the old coordinate system, there was no concept of a root node of the scene tree, so most entities were orphan nodes without a parent. This does not work with the new system, and now entities are required to specify which map (root node) they are being placed on at the time of their creation. Currently it is not possible to spawn an entity into nullspace, this restriction can be removed if needed. It is now not possible to have a non-map entity without a parent.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -2,6 +2,7 @@
  using System.Collections.Generic;
  using System.Linq;
  using Robust.Shared.Animations;
+ using Robust.Shared.Containers;
  using Robust.Shared.Enums;
  using Robust.Shared.GameObjects.Components.Map;
  using Robust.Shared.GameObjects.EntitySystemMessages;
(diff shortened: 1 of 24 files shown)


Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -536,7 +512,9 @@ public override void HandleComponentState(ComponentState curState, ComponentStat
                 var rebuildMatrices = false;
                 if (Parent?.Owner?.Uid != newParentId)
                 {
+#if _DELME
```
PJB3005: Left this in?

PJB3005 (approved):

PR #918, opened 2019-12-25
TransformComponent.MapID returns null instead of throwing

DamianX:
Because I need to spawn an entity in nullspace for character preview.
`GridID` also throws currently, should I change that as well even if I don't need it?

### DIFF
--- a/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -89,7 +89,8 @@
                 if (_parent.IsValid())
                     return Parent.GridID;
 
-                throw new InvalidOperationException("Transform node does not exist inside scene tree!");
+                // Transform node does not exist inside scene tree
+                return GridId.Nullspace;
             }
         }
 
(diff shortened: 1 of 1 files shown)


PJB3005:
Making grid ID not throw is probably best yeah.

Acruid:
The only thing this PR does is allow someone to spawn an orphan entity outside of the scene tree. Removing the MapID exception is not a solution to anything. Returning the Nullspace MapID does not make the transform an entity inside the Nullspace scene graph.

The PR is not getting accepted at this point.

DamianX:
I don't know what I need to do.
This allows me to do `entityManager.SpawnEntity("HumanMob_Content", GridCoordinates.Nullspace)`.

Acruid:
That is OK, I will look into a better solution tomorrow.

PJB3005:
image

Acruid:
"Way better fix" is inside https://github.com/space-wizards/RobustToolbox/commit/58d0a771e8882efcfea245fe92485c9a53a96161.

Call `mapMan.CreateNewMapEntity(MapId.Nullspace);` and then spawn entities into MapId.Nullspace like any other map.

PR #924, opened 2020-01-03
Rotating Grids that kind of works

ghost:
Does a lot of things to make rotating grids work, including sprite rotation and improving vector math. It's very rough and the calculation to determine which gridchunks to render is broken. I give up on that one, maybe I'm retarded. Ask if you need any details on it.

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -996,15 +996,13 @@ namespace Robust.Client.GameObjects
             if (useWorldTransform)
             {
                 var angle = Rotation;
-                if (Directional)
+                if (!Directional)
+
                 {
-                    angle -= Owner.Transform.WorldRotation;
-                }
-                else
-                {
+                    angle += Owner.Transform.WorldRotation;
                     angle -= new Angle(MathHelper.PiOver2);
                 }
-
+                angle -= Owner.Transform.LocalRotation;
                 var mOffset = Matrix3.CreateTranslation(Offset);
                 var mRotation = Matrix3.CreateRotation(angle);
                 Matrix3.Multiply(ref mRotation, ref mOffset, out transform);
@@ -1388,7 +1386,7 @@ namespace Robust.Client.GameObjects
                 return RSI.State.Direction.South;
             }
 
-            var angle = new Angle(Owner.Transform.WorldRotation);
+            var angle = new Angle(Owner.Transform.LocalRotation);
             return angle.GetDir().Convert(type);
         }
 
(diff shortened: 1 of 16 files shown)


Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
```diff
@@ -61,9 +65,11 @@ private void _drawGrids(Box2 worldBounds)
                     continue;
                 }
 
-                var model = Matrix3.Identity;
+                Matrix3 model = Matrix3.Identity;
```
PJB3005, line 68: Why remove these `var` declarations :thinking:

Robust.Client/Graphics/ClientEye/Eye.cs
```diff
@@ -14,7 +14,10 @@ public class Eye : IEye
 
         /// <inheritdoc />
         public virtual MapCoordinates Position { get; internal set; }
-        
+
+        /// <inheritdoc />
+        public Angle Rotation { get; set; }
```
PJB3005, line 19: Didn't get around to using this?

Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
ghost: It's more readable. I was also probably planning to change it but went against it.

Robust.Client/Graphics/ClientEye/Eye.cs
ghost: Nope, this caused an absolute shitton of issues. Grid rotation is already buggy enough. Maybe in the future.

ghost:
And to add on, the normal rotation (0 deg) works fine. When a grid is rotated the calculations for determining what tiles to render is borked. I kind of gave up on it after getting frustrated with Box2Rotated.

Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
PJB3005: I disagree. It's completely obvious that this variable is a `Matrix3`. No need to write that again.

Same goes for the changes you made like this in `Clyde.Rendering.cs`.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -996,15 +996,13 @@ internal void OpenGLRender(DrawingHandleWorld drawingHandle, bool useWorldTransf
             if (useWorldTransform)
             {
                 var angle = Rotation;
-                if (Directional)
-                {
-                    angle -= Owner.Transform.WorldRotation;
-                }
-                else
+                if (!Directional)
```
PJB3005, line 999: Are these changes so sprites ignore the rotation of their parent, so they rotate with the grid?

If so, this is gonna break for more complex multi-level layouts (person in chair).

Might be better to just implement the angle snapping idea.

Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
PJB3005, line 68: Looking at this again, you could probably replace this manual matrix calculation by getting the world matrix from the grid's entity.

PR #934, opened 2020-01-21
Add delete key behavior to LineEdit

Moneyl:
Pressing delete will now delete the character in front of your LineEdit cursor like usual. Will create small PR on the content repo that completes this shortly after this one.

### DIFF
--- a/Robust.Client/UserInterface/Controls/LineEdit.cs
+++ b/Robust.Client/UserInterface/Controls/LineEdit.cs
@@ -346,7 +346,7 @@ namespace Robust.Client.UserInterface.Controls
                 }
                 else if (args.Function == EngineKeyFunctions.TextDelete)
                 {
-                    if (_cursorPosition + 1 > _text.Length || !Editable)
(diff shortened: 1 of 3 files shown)


Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -344,6 +344,15 @@ protected internal override void KeyBindDown(GUIBoundKeyEventArgs args)
                         InsertAtCursor(clipboard.GetText());
                     }
                 }
+                else if (args.Function == EngineKeyFunctions.TextDelete)
+                {
+                    if (_cursorPosition + 1 > _text.Length || !Editable)
```
PJB3005: This is probably a bit clearer.

```suggestion
                    if (_cursorPosition >= _text.Length || !Editable)
```

Moneyl: yup, you're right. applied

Issue #807, opened 2019-05-05
Implement alternate mouse cursors.


PJB3005:
For things such as resizing in game windows.

The problem is OpenTK does not expose the OS ones at all. I also looked at say GLFW, and while it does expose some, it doesn't expose all (notably, diagonal resize, because macOS does not document having it, even though it does have it, thanks Apple).

So we have two options here:

1. Draw our own mouse cursors and use them via OS hardware cursors.
2. Switch to GLFW for windowing and simply don't have a diagonal cursor.

Which.. is more preferable?

io-june:
use CC-licensed cursors, you don't even need to draw anything

tentekal:
After taking a look at OpenTK further, I think we should go into GLFW for windowing, it's a hell of a lot easier and cleaner for this feature. No idea how difficult it would be to convert it though.

PJB3005:
Well I'll be glad to let you know that as of 52e5afd27736f6485c068b3e619fa355b2323ab7 we *are* using GLFW for windowing and that with https://github.com/glfw/glfw/issues/427 GLFW 3.4 will have support for all the cursors we could need.

tentekal:
Nice, I got this idea from looking in OpenTK and seeing the cursor updates you pushed. This will be super useful for those RTS-like cursor indicators we talked about forever ago.

PJB3005:
For RTS-like cursor indicators it's probably better to just manually draw at the cursor position, because you can't really sanely render them to the OS.

tentekal:
I think GLFW supports small images, and animations wouldn't be necessary (would be pretty awful for trying to click fine detail). At the very least a cursor-recolor would be sufficient.

PJB3005:
Yeah but if you want to display something like a count next to your cursor you're out of luck.

PJB3005:
Done by 832a4e01f1fca7280ce94d973f554f5a8c8bf8e4

PR #943, opened 2020-01-27
Cursor styles

tentekal:
Please ignore hacky implementation of CreateCursor. I plan to fix some of these added empty lines and squash some unnecessary commits. 

Placeholder addressing #807


Robust.Client/Graphics/Clyde/Clyde.Windowing.cs
```diff
@@ -151,6 +158,21 @@ private void InitWindow()
             InitOpenGL();
         }
 
+        public override void CreateCursor(GlfwImage image, int x, int y)
```
PJB3005: 1. GLFW should not be exposed directly.
2. Just make this take an `Image<Rgba32>`

Robust.Client/Graphics/Clyde/Clyde.Windowing.cs
```diff
@@ -151,6 +158,21 @@ private void InitWindow()
...
+        {
+            List<Image<Rgba32>> cursorIcons = LoadCursorIcon("/Textures/UserInterface/cursorIcons");
```
PJB3005: Uh, why even take in an image then?

Robust.Client/Graphics/Clyde/Clyde.Windowing.cs
```diff
@@ -151,6 +158,21 @@ private void InitWindow()
...
+            image = SetCursorIcon(cursorIcons);
+            Logger.Info("Cursor style changed");
+
+            _cursor = GLFW.CreateCursor(image, x, y);
```
PJB3005: You need to destroy the previous cursor (if we have any) or else there's a memory leak here.

Robust.Client/Graphics/Clyde/Clyde.Windowing.cs
```diff
@@ -204,6 +226,53 @@ private void SetWindowIcon(IEnumerable<Image<Rgba32>> icons)
...
+                {
+                    var image = Image.Load(stream);
+                    cursorIcons.Add(image);
+                }
+            }
+
+            return cursorIcons;
+        }
+
+        private GlfwImage SetCursorIcon(IEnumerable<Image<Rgba32>> cursorIcons)
```
PJB3005, line 241: You don't have to do all this stuff with taking in a list. You're just returning the result for the first image aren't you?

Also, you're doing a memory violation here. You can't return the `GlfwImage` allocated since it contains a pointer to managed memory (that gets unpinned at the end of the method).

To actually get a `GlfwImage` from a *single* `Image<Rgba32>` you can just do the following:

```cs
fixed (Rgba32* pixPtr = image.GetPixelSpan())
{
	var glfwImage = new GlfwImage(image.Width, image.Height, (byte*) pixPtr);
	// DO NOT use glfwImage after leaving the `fixed` block. Call `glfwCreateCursor` in here or something. 
}
```

PJB3005 (changes requested):

Robust.Client/Graphics/Clyde/Clyde.Windowing.cs
tentekal: My mistake, I thought I read somewhere in documentation the CreateCursor() destroyed the previous cursor automatically. Not sure why I thought that.

Robust.Client/Graphics/Clyde/Clyde.Windowing.cs
tentekal: Understood, I'll expose via IClyde.cs.

Robust.Client/Graphics/Clyde/Clyde.Windowing.cs
tentekal: I intend to pass a PNG path from a component to CreateCursor. This was just a lazy way of making something render without worrying about components yet.

Robust.Client/Graphics/Clyde/Clyde.Windowing.cs
tentekal: Copy-paste of the method used for loading window icons, I originally had something similar to your method implemented but it was accidentally removed.  This is super helpful though, thanks.

PJB3005:
I am gonna do this myself since the PR is stale.

PR #947, opened 2020-02-02
Update Travis to Bionic and Python 3.6

SpaceManiac:


PJB3005:
Probably better to change the distro to Bionic or something

PR #948, opened 2020-02-03
Increase default vv window size for convenience

Moneyl:
Makes all tabs and information visible by default instead of needing to resize vv windows to see everything. 

Before:
Content Client_Oncn3KO80l

After:
Content Client_jIUBqruLlc

This is pretty subjective, and some windows will still have info cutoff, but I find it useful since I frequently have to resize vv windows.


gradientvera (approved):

PJB3005:
There's still quite a bit of things that can be done to improve the density of the window, but this is a nice start.

PR #955, opened 2020-02-06
Have PhysicsManager use DynamicTree

Tyler-IN:
IBroadPhase and BroadPhaseNaive are unused now, though.

Rushed.


PJB3005:
I think it'd probably be better if this was properly integrated with `IBroadPhase`, but I'll let @Acruid be the final judge on that.

Tyler-IN:
IBroadPhase would have to change drastically, I think it is a bit overcomplicated, maybe based on other physics abstraction approaches.

PhysicsManager is technically doing the broad phase collision operation, the DynamicTree is just a data structure.

Tyler-IN:
I went the more evil route and replaced IBroadPhase in order to separate out BroadPhase from DynamicTree from PhysicsManager.

PR #957, opened 2020-02-08
Correct ray casting

Tyler-IN:
hotfix for ray intersections invoking _extractAabb too early

fix raycasting, minor formatting

remove a Debug.WriteLine

### DIFF
--- a/Robust.Shared/Physics/DynamicTree.cs
+++ b/Robust.Shared/Physics/DynamicTree.cs
@@ -604,18 +604,19 @@ namespace Robust.Shared.Physics
 
                 var item = node.Item;
 
-                if (!approx)
-                {
-                    var preciseAabb = _extractAabb(item);
-
-                    if (!ray.Intersects(preciseAabb, out dist, out hit))
-                    {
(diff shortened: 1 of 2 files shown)


Tyler-IN:
Instead of a SortedList, could just compare dist and keep lowest... augh I'll make another commit on top.

Moneyl:
Tested this out and it's working. For future reference you should try to avoid force pushing with PRs. From what I understand it can make resolving merge conflicts a pain for whoever merges it.

PJB3005:
Nah the pain with force pushes is that it makes the PR history kinda confusing, since you can't incrementally look at the diff anymore.

For reference, I squash merge anything (unless it needs a regular merge due to having another PR depend on it) so you can make your PR diff as dirty as you like.

Tyler-IN:
I force pushed because the branch was out of sync, don't git preach to me I'm an senior architect lol

(I think I dragged and dropped a branch inadvertently causing the fork history to be out of sync w/ my local, force pushing fixed history; only have to force push if rewriting history)

PR #967, opened 2020-02-11
Container light occlusion

ghost:
https://github.com/space-wizards/space-station-14/pull/687


PJB3005 (commented):
Does this work correctly when you have nested containers? e.g. a box without occlusion inside a backpack?

ghost:
will check when I'm home

ghost:
haven't found a way to break it, and we don't have boxes to fit inside backpacks

Robust.Client/GameObjects/Components/Light/PointLightComponent.cs
```diff
@@ -100,6 +107,7 @@ public bool VisibleNested
         private Color _color = Color.White;
         private Vector2 _offset;
         private bool _enabled = true;
+        private bool _occluded = true;
```
PJB3005: This shouldn't default to true.

PJB3005 (dismissed):
I tested this by making the container of a locker occlude.

If you drop the flashlight into the locker and close it, it does not let light through.
If you hold the flashlight in your hand and close it on you, it does let light through.

This should be fixed.

ghost:
unsure how to solve this

PR #972, opened 2020-02-12
Adds before/after movement events to the physics system

ghost:
I have no idea if this is the optimal way to do this but there has to be a way to directly react to movement for pulling and probably other stuff

also unatomic formatting changes for clarity


ghost:
Ok this is definitely not the way to go about this

Issue #859, opened 2019-09-01
Add down arrow icon on OptionButton


ShadowCommander:
Just a small thing to make the OptionButton easier to recognize and use.

gradientvera:
Fixed by https://github.com/space-wizards/space-station-14/pull/700

PR #975, opened 2020-02-13
Implement ContainerButton and refactor buttons

ShadowCommander:
Implements ContainerButton which is a button that resizes to fit the Controls it is a parent of.

Refactors OptionButton to use ContainerButton so that an inverted triangle can be added to make it more obvious that the button is an OptionButton.

Content side is space-wizards/space-station-14#700

Part of fixing #859

### DIFF
--- a/Robust.Client/State/States/MainMenu.cs
+++ b/Robust.Client/State/States/MainMenu.cs
@@ -228,10 +228,7 @@ namespace Robust.Client.State.States
                 var layout = new LayoutContainer();
                 AddChild(layout);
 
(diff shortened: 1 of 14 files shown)


PJB3005 (commented):
Is it possible to make `Button` also inherit from `ContainerButton`? That would solve a lot of duplicate could I would imagine.

ShadowCommander:
Don't merge yet. I'm still working on styling.

ShadowCommander:
Ready for review again.

PJB3005 (approved):

PR #983, opened 2020-02-16
Implement client bubble PVS.

Tyler-IN:
You guys call it bubbling. PVS means Potentially Visible Set.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
@@ -78,11 +78,19 @@ namespace Robust.Client.GameObjects.EntitySystems
 
                 if (stream.TrackingCoordinates != null)
                 {
-                    stream.Source.SetPosition(stream.TrackingCoordinates.Value.ToMapPos(_mapManager));
+                    if (!stream.Source.SetPosition(stream.TrackingCoordinates.Value.ToMapPos(_mapManager)))
+                    {
+                        Logger.Warning("Interrupting positional audio, can't set position.");
+                        stream.Source.StopPlaying();
+                    }
                 }
                 else if (stream.TrackingEntity != null)
                 {
-                    stream.Source.SetPosition(stream.TrackingEntity.Transform.WorldPosition);
+                    if(!stream.Source.SetPosition(stream.TrackingEntity.Transform.WorldPosition))
+                    {
+                        Logger.Warning("Interrupting positional audio, can't set position.");
+                        stream.Source.StopPlaying();
+                    }
                 }
             }
 
@@ -148,10 +156,16 @@ namespace Robust.Client.GameObjects.EntitySystems
         /// <param name="stream">The audio stream to play.</param>
         /// <param name="entity">The entity "emitting" the audio.</param>
         /// <param name="audioParams"></param>
+        [CanBeNull]
         public IPlayingAudioStream Play(AudioStream stream, IEntity entity, AudioParams? audioParams = null)
         {
             var source = _clyde.CreateAudioSource(stream);
-            source.SetPosition(entity.Transform.WorldPosition);
+            if (!source.SetPosition(entity.Transform.WorldPosition))
+            {
+                Logger.Warning("Can't play positional audio, can't set position.");
+                return null;
+            }
+
             ApplyAudioParams(audioParams, source);
 
             source.StartPlaying();
@@ -170,6 +184,7 @@ namespace Robust.Client.GameObjects.EntitySystems
         /// <param name="filename">The resource path to the OGG Vorbis file to play.</param>
         /// <param name="coordinates">The coordinates at which to play the audio.</param>
         /// <param name="audioParams"></param>
+        [CanBeNull]
         public IPlayingAudioStream Play(string filename, GridCoordinates coordinates, AudioParams? audioParams = null)
         {
             if(_resourceCache.TryGetResource<AudioResource>(new ResourcePath(filename), out var audio))
@@ -187,11 +202,17 @@ namespace Robust.Client.GameObjects.EntitySystems
         /// <param name="stream">The audio stream to play.</param>
         /// <param name="coordinates">The coordinates at which to play the audio.</param>
         /// <param name="audioParams"></param>
+        [CanBeNull]
         public IPlayingAudioStream Play(AudioStream stream, GridCoordinates coordinates,
             AudioParams? audioParams = null)
         {
             var source = _clyde.CreateAudioSource(stream);
-            source.SetPosition(coordinates.ToMapPos(_mapManager));
+            if (!source.SetPosition(coordinates.ToMapPos(_mapManager)))
+            {
+                Logger.Warning("Can't play positional audio, can't set position.");
+                return null;
+            }
+
             ApplyAudioParams(audioParams, source);
 
             source.StartPlaying();
--- a/Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
+++ b/Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
@@ -3,6 +3,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Runtime.CompilerServices;
 using OpenTK;
 using OpenTK.Audio.OpenAL;
@@ -412,10 +413,16 @@ namespace Robust.Client.Graphics.Clyde
                 _checkAlError();
             }
 
-            public void SetPosition(Vector2 position)
+            public bool SetPosition(Vector2 position)
             {
                 _checkDisposed();
 
+                var (x, y) = position;
+
+                if (!ValidatePosition(x, y))
+                {
+                    return false;
+                }
 #if DEBUG
                 // OpenAL doesn't seem to want to play stereo positionally.
                 // Log a warning if people try to.
@@ -428,9 +435,19 @@ namespace Robust.Client.Graphics.Clyde
                 }
 #endif
 
-                var (x, y) = position;
                 AL.Source(SourceHandle, ALSource3f.Position, x, y, 0);
                 _checkAlError();
+                return true;
+            }
+
+            private static bool ValidatePosition(float x, float y)
+            {
+                if (float.IsFinite(x) && float.IsFinite(y))
+                {
+                    return true;
+                }
+
+                return false;
             }
 
             public void SetPitch(float pitch)
@@ -562,15 +579,32 @@ namespace Robust.Client.Graphics.Clyde
                 _checkAlError();
             }
 
-            public void SetPosition(Vector2 position)
+            public bool SetPosition(Vector2 position)
             {
                 _checkDisposed();
 
-                _mono = true;
                 var (x, y) = position;
+
+                if (!ValidatePosition(x, y))
+                {
+                    return false;
+                }
+
+                _mono = true;
                 // ReSharper disable once PossibleInvalidOperationException
                 AL.Source(SourceHandle.Value, ALSource3f.Position, x, y, 0);
                 _checkAlError();
+                return true;
+            }
+
+            private static bool ValidatePosition(float x, float y)
+            {
+                if (float.IsFinite(x) && float.IsFinite(y))
+                {
+                    return true;
+                }
+
+                return false;
             }
 
             public void SetPitch(float pitch)
--- a/Robust.Server/GameObjects/ServerEntityManager.cs
+++ b/Robust.Server/GameObjects/ServerEntityManager.cs
@@ -173,13 +173,16 @@ namespace Robust.Server.GameObjects
 
         private static readonly Vector2 Vector2NaN = new Vector2(float.NaN, float.NaN);
 
+        private IDictionary<EntityUid, GameTick> GetLastSeen(IPlayerSession player)
+        {
+            var lastSeen = GetLastSeen(player);
+
+            return lastSeen;
+        }
+
         private GameTick GetLastSeenTick(IPlayerSession player, EntityUid uid)
         {
-            if (!_playerLastSeen.TryGetValue(player, out var lastSeen))
-            {
-                lastSeen = new Dictionary<EntityUid, GameTick>();
-                _playerLastSeen[player] = lastSeen;
-            }
+            var lastSeen = GetLastSeen(player);
 
             if (!lastSeen.TryGetValue(uid, out var tick))
             {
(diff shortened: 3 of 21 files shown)


Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -610,6 +617,11 @@ private static Box2 GetWorldAabbFromEntity(in IEntity ent)
         }
 
         #endregion
+
+        public virtual void Update()
+        {
+        }
```
Clyybber, line 623: Should prolbably be removed

Robust.Shared/Interfaces/GameObjects/IEntityManager.cs
```diff
@@ -198,5 +198,7 @@ public interface IEntityManager
 
         #endregion
 
+        void Update();
```
Clyybber, line 201: Ditto

Robust.Shared/Network/NetManager.cs
```diff
@@ -641,6 +641,12 @@ private bool DispatchNetMessage(NetIncomingMessage msg)
             {
                 instance.ReadFromBuffer(msg);
             }
+            catch (InvalidCastException ice)
+            {
+                Logger.ErrorS("net",
+                    $"{msg.SenderConnection.RemoteEndPoint}: Wrong deserialization of {type.Name} packet: {ice.Message}");
+                throw ice;
```
Clyybber:
```suggestion
                throw;
```

Clyybber (commented):
Awesome

Tyler-IN:
FYI there are still some outstanding known errata w/ picking parents/children.

Needs more testing to narrow down when/where/how.

I will be making client changes to support position being NaN in state updates as "don't vanish until seen" mechanics so the client can't see things are being unseen.

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -104,7 +108,7 @@ public void SendGameStateUpdate()
 
             var oldestAck = GameTick.MaxValue;
 
-            var work = new List<(INetChannel channel, GameTick lastAck)>();
+            var work = new List<(IPlayerSession session, INetChannel channel, GameTick lastAck)>();
```
TYoung86:
```suggestion
```

Tyler-IN:
```suggestion
```

Robust.Shared/Interfaces/GameObjects/IEntityManager.cs
TYoung86: Turns out I do actually use it in ServerEntityManager.

Robust.Shared/GameObjects/EntityManager.cs
TYoung86: Turns out I do actually use it in ServerEntityManager.

Tyler-IN: Turns out I do actually use it in ServerEntityManager.

Robust.Shared/Interfaces/GameObjects/IEntityManager.cs
Tyler-IN: Turns out I do actually use it in ServerEntityManager.

Robust.Client/GameObjects/ClientEntityManager.cs
```diff
@@ -178,6 +178,17 @@ protected override EntityUid GenerateEntityUid()
...
+        {
+            if (Entities.TryGetValue(uid, out var ent))
+            {
+                return ent;
+            }
+
+            Logger.ErrorS("entity", $"Could not find entity {uid} on this client.");
+            return null;
+        }
+
```
Tyler-IN:
```suggestion
```

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -117,7 +117,7 @@ public virtual void FrameUpdate(float frameTime)
         /// </summary>
         /// <param name="uid"></param>
         /// <returns>Entity or null if entity id doesn't exist</returns>
-        public IEntity GetEntity(EntityUid uid)
+        public virtual IEntity GetEntity(EntityUid uid)
```
Tyler-IN:
```suggestion
        public IEntity GetEntity(EntityUid uid)
```

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -148,10 +156,16 @@ public IPlayingAudioStream Play(string filename, IEntity entity, AudioParams? au
...
         /// <param name="audioParams"></param>
+        [CanBeNull]
         public IPlayingAudioStream Play(AudioStream stream, IEntity entity, AudioParams? audioParams = null)
         {
             var source = _clyde.CreateAudioSource(stream);
-            source.SetPosition(entity.Transform.WorldPosition);
+            if (!source.SetPosition(entity.Transform.WorldPosition))
+            {
+                Logger.Warning("Can't play positional audio, can't set position.");
+                return null;
```
PJB3005, line 166: Not sure it's a good idea for this to ever return null. Even in the worst scenario it might just be better to have this return a dummy value that's seen as stopped?

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -187,11 +202,17 @@ public IPlayingAudioStream Play(string filename, GridCoordinates coordinates, Au
...
+        [CanBeNull]
         public IPlayingAudioStream Play(AudioStream stream, GridCoordinates coordinates,
             AudioParams? audioParams = null)
         {
             var source = _clyde.CreateAudioSource(stream);
-            source.SetPosition(coordinates.ToMapPos(_mapManager));
+            if (!source.SetPosition(coordinates.ToMapPos(_mapManager)))
+            {
+                Logger.Warning("Can't play positional audio, can't set position.");
+                return null;
```
PJB3005, line 213: Ditto on returning null here.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
```diff
@@ -412,10 +413,16 @@ public void SetPlaybackPosition(float seconds)
...
             }
 
-            public void SetPosition(Vector2 position)
+            public bool SetPosition(Vector2 position)
             {
                 _checkDisposed();
 
+                var (x, y) = position;
+
+                if (!ValidatePosition(x, y))
```
PJB3005, line 422: This *feels* like really bad API design to me.

Maybe make this throw on NaN and have the uses explicitly check for it before calling?

Robust.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -1,35 +1,62 @@
...
 #pragma warning disable 649
         [Dependency] private readonly IMapManager _mapManager;
+
         [Dependency] private readonly IPauseManager _pauseManager;
+
+        [Dependency] private readonly IConfigurationManager _configurationManager;
 #pragma warning restore 649
 
-        private int _nextServerEntityUid = (int)EntityUid.FirstUid;
+        private float? _maxUpdateRangeCache;
```
PJB3005, line 45: This thing doesn't actually get assigned outside of that null down there.

Robust.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -124,6 +154,509 @@ public List<EntityState> GetEntityStates(GameTick fromTick)
...
+            {
+                tick = GameTick.First;
+            }
+
+            return tick;
+        }
+
+        private GameTick UpdateLastSeenTick(IPlayerSession player, EntityUid uid, GameTick newTick)
+        {
+            if (!_playerLastSeen.TryGetValue(player, out var lastSeen))
```
PJB3005: This block is duplicated in quite a few methods.

TYoung86:
from discord, a rough explanation;
> so when a player joins, they get sent the entire world
> the last seen dictionary stores anything that they've seen that has changed since that tick
> as just a (uid,tick) pair
> if they come across something that has a different change tick than they've last seen that's inside their bubble, they get the state
> if it hasn't changed since they last saw it, no state is sent nor needs to be
> if something changes outside their bubble that they have previously seen, the NaN covers that situation that would produce a ghost
> the client can be updated (imo separate PR) to handle NaNs by hiding the object temporarily until the next position comes in, instead of updating the pos
> or the process can be changed to just send some new "unsee this" message instead of position NaN
> when they "unsee" something, it's removed from the last seen dictionary
> so it doesn't blow up and replicate the world inside the dictionary
> the lastSeen dictionary only gets queried or built based on their bubble
> so it should virtually always be a subset of the world, worst case they've somehow seen everything and it's the full set of uids but just ticks
> in a future PR I'd like to exclude sending container contents, and maybe re-work some of the handling so sending updates to parents can be more relaxed
> right now immobile objects like snap grid objects outside the bubble get NaN'd prematurely when they're destroyed if they're in some common parent/child heirarchy

Robust.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -124,6 +154,509 @@ public List<EntityState> GetEntityStates(GameTick fromTick)
...
+        }
+
+        private readonly PlayerSeenEntityStatesResources _playerSeenEntityStatesResources
+            = new PlayerSeenEntityStatesResources(false);
+
+        /// <inheritdoc />
+        public List<EntityState> UpdatePlayerSeenEntityStates(GameTick fromTick, IPlayerSession player, float range)
+        {
+            var playerEnt = player.AttachedEntity;
+            if (playerEnt == null)
```
PJB3005, line 316: Players in the lobby have no entity, so...

PJB3005 (commented):
Would it be possible to introduce an explicit off switch (that also would work if changed mid game?).

Robust.Server/GameObjects/ServerEntityManager.cs
TYoung86: ???

Robust.Server/GameObjects/ServerEntityManager.cs
TYoung86: That's correct... so the server can update _maxUpdateRangeCache cvar once per frame.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
TYoung86: Outside of the API it's used like this;
```cs
if (!renderer.Source.SetPosition(renderer.TrackingEntity.Transform.WorldPosition))
{
  Shared.Log.Logger.Warning("Interrupting positional audio, can't set position.");
  renderer.Source.StopPlaying();
}
```
Adding `[MustUseReturnValue]` makes the API require the bool to be checked or issue a warning on complying IDEs.
The caller should just stop playing the sound if the position is NaN, because that indicates it's an entity the client is unaware of the position. If this ever occurs, it's probably intentional.
I think Audio that's not relatively quiet or localized needs to be sent with a specifically with a position and not an entity. There should be no guarantee that a sound can play at any position.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
TYoung86: PlayAudioPositionalHandler as the only outer caller doesn't check the return value.
Do you want it to verify if a sound played?

Tyler-IN: PlayAudioPositionalHandler as the only outer caller doesn't check the return value.
Do you want it to verify if a sound played?

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
Tyler-IN: Outside of the API it's used like this;
```cs
if (!renderer.Source.SetPosition(renderer.TrackingEntity.Transform.WorldPosition))
{
  Shared.Log.Logger.Warning("Interrupting positional audio, can't set position.");
  renderer.Source.StopPlaying();
}
```
Adding `[MustUseReturnValue]` makes the API require the bool to be checked or issue a warning on complying IDEs.
The caller should just stop playing the sound if the position is NaN, because that indicates it's an entity the client is unaware of the position. If this ever occurs, it's probably intentional.
I think Audio that's not relatively quiet or localized needs to be sent with a specifically with a position and not an entity. There should be no guarantee that a sound can play at any position.

Robust.Server/GameObjects/ServerEntityManager.cs
Tyler-IN: That's correct... so the server can update `_maxUpdateRangeCache` related cvar once per frame.

Robust.Server/GameObjects/ServerEntityManager.cs
Tyler-IN: Sure I'll break it out.

Robust.Server/GameObjects/ServerEntityManager.cs
Tyler-IN: What does this mean?

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
Tyler-IN: The only outer caller doesn't evaluate the result of Play.

How should it handle this case?

Tyler-IN:
> Would it be possible to introduce an explicit off switch (that also would work if changed mid game?).

Added.

Robust.Server/GameObjects/ServerEntityManager.cs
Tyler-IN: Oh I read this wrong, I thought you meant it doesn't get assigned except for the null and where it gets assigned...
```cs
        public float MaxUpdateRange => _maxUpdateRangeCache
            ??= _configurationManager.GetCVar<float>("net.maxupdaterange");
```
^ It gets assigned there. It's just 2 lines down.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
Tyler-IN: Do you think it should be `TrySetPosition` instead?

Acruid (approved):
There are a few edge cases that need to be looked at, mainly around world parenting and shuttles. As far as I can tell these problems are because the TransformComponent is poorly implemented, and really needs some testing and API tweaks. The main features were successfully implemented and seem stable enough for the CM test at the end of the month, so I am approving this.

Robust.Server/GameObjects/ServerEntityManager.cs
PJB3005: Oh, my bad.

PJB3005 (approved):

PR #988, opened 2020-02-18
Adds default config for diagonal movement setting

CakeQ:
Shouldn't change anything beyond making the configuration option more apparent

### DIFF
--- a/Robust.Server/server_config.toml
+++ b/Robust.Server/server_config.toml
@@ -20,6 +20,7 @@ mapname = "stationstation"
 maxplayers = 64
 type = 1
 welcomemsg = "Welcome to the server!"
+diagonalmovement = true
 
 [console]
 width = 120


PJB3005:
Personally I'd rather this file be deleted and a default generated automatically by the engine for reference purposes like this. It's a constant source of pain of this file going out of sync.

io-june:
>Personally I'd rather this file be deleted and a default generated automatically by the engine for reference purposes like this. It's a constant source of pain of this file going out of sync.

that would require server host to compile from source code

PJB3005:
> that would require server host to compile from source code

1. no it wouldn't.
2. this would be a problem how?

CakeQ:
Surely you're asking me to do something out of scope? Unless someone plans on writing this supposed automatic config maker the config toml should either include all existing configurations or not exist at all

PJB3005:
I'd rather not merge content-specific cvars into the engine for obvious reasons. I was proposing an alternative method to solve the problem that's content-agnostic.

PJB3005:
Actually it turns out the `saveconfig` command already dumps default config entries. To auto generate the config file you can just run `saveconfig` in the server console.

Acruid:
I'm closing this, it would be better if the toml files were completely removed from the project. You can generate the default ones at any time with `saveconfig` in the console, or somebody improve it by generating default files if they are missing on startup.

PR #993, opened 2020-02-22
Make login command use host role instead of admin role

Moneyl:
Requires [content #683](https://github.com/space-wizards/space-station-14/pull/683). Makes login console command use new host role. Has a few commands that may be useful for debugging that the administrator role doesn't have.

### DIFF
--- a/Robust.Server/Console/ConGroupController.cs
+++ b/Robust.Server/Console/ConGroupController.cs
@@ -59,7 +59,7 @@ namespace Robust.Server.Console
                 var address = session.ConnectedClient.RemoteEndPoint.Address;
                 if (Equals(address, IPAddress.Loopback) || Equals(address, IPAddress.IPv6Loopback))
                 {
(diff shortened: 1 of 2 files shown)


PJB3005:
Uh, how can people log in as `admin` role then?

Moneyl:
I think I can change `login` to handle multiple groups/passwords. Is a hard-coded thing where it only works for the `admin` and `host` groups and has separate passwords for each fine for the moment? This will probably be scrapped once we have an account system anyways, right?

Acruid:
This entire `login` and password system will probably be scrapped once we have user accounts and the server database working. It only exists to prevent normal players from currently having full console access in the public server, and admins the ability to run restricted commands without rcon access.

PJB3005:
I made `loginlocal` log you in as host in 09a06c14c57916c9892088ed87f53261fca84886. I'd say this PR can be closed now.

PR #1000, opened 2020-02-25
Expression based binding for ObjectSerializer

Tyler-IN:
Needs review and testing.

### DIFF
--- a/Robust.Shared/Serialization/YamlObjectSerializer.cs
+++ b/Robust.Shared/Serialization/YamlObjectSerializer.cs
@@ -182,7 +182,8 @@ namespace Robust.Shared.Serialization
 
             var value = getter();
             DataField(ref value, name, defaultValue, alwaysWrite);
-            if (!Writing)
+
+            if (Writing)
             {
                 return;
             }
--- a/Robust.Shared/Serialization/YamlObjectSerializer.cs
+++ b/Robust.Shared/Serialization/YamlObjectSerializer.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
+using System.Linq.Expressions;
 using System.Reflection;
 using Robust.Shared.Interfaces.Reflection;
 using Robust.Shared.Interfaces.Serialization;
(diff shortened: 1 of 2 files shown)


Robust.Shared/Serialization/YamlObjectSerializer.cs
```diff
@@ -145,6 +146,59 @@ public override void DataField<T>(ref T value, string name, T defaultValue, bool
...
+
+            if (!(expr.Body is MemberExpression mExpr))
+            {
+                throw new NotImplementedException(expr.Body.GetType().FullName);
+            }
+
+            // the below looks like duplicate code, but PropertyInfo and FieldInfo have no common
+            // base or interface for GetValue/SetValue
+
+            object value;
```
PJB3005: Shouldn't this be a `T value`? The `DataField` below don't have the correct generic parameters like this do they?

Tyler-IN: Fixed.

Robust.Shared/Serialization/YamlObjectSerializer.cs
```diff
@@ -164,39 +164,33 @@ public override void DataField<T>(ref T value, string name, T defaultValue, bool
...
+                    getter = () => (T) pi.GetValue(o);
+                    setter = (v) => pi.SetValue(o, v);
                     break;
                 default:
                     throw new NotImplementedException(mExpr.Member.GetType().FullName);
             }
+
+            var value = getter();
+            DataField(ref value, name, defaultValue, alwaysWrite);
+            if (!Writing)
```
PJB3005: I think you have this backwards. `Writing` means the serialization is happening TO FILE. Right now you're running the setter while writing instead of while reading.

Tyler-IN: Inverted.

Robust.Shared/Serialization/YamlObjectSerializer.cs
```diff
@@ -145,6 +146,54 @@ public override void DataField<T>(ref T value, string name, T defaultValue, bool
...
+
+            var value = getter();
+            DataField(ref value, name, defaultValue, alwaysWrite);
+
+            if (Writing)
+            {
+                return;
+            }
+
+            if (!alwaysWrite && IsValueDefault(name, value, defaultValue))
```
PJB3005, line 191: `alwaysWrite` and `IsValueDefault` are both things intended for writing, not reading. Just run the setter.

PJB3005:
Merged after manually patching it in fd8e68091e8c780f893ff3ff330a57b4239d8b7f. Couldn't push to the PR branch.

PR #1004, opened 2020-02-26
Fix lockers making items uninteractable

Tyler-IN:


Tyler-IN:
NOTE:
Currently its not possible to drop an item without it being on a grid, and it's not possible to pick up an item that is parented to the map entity.
Once that is fixed, there is a TODO in this PR to cover changes necessary.

PR #1005, opened 2020-02-26
Automatic Partial Shader Reloading

Tyler-IN:
Makes shader load failures non-fatal.

Carries includes over to to ParsedShader.

### DIFF
--- a/Robust.Client/Console/Commands/Debug.cs
+++ b/Robust.Client/Console/Commands/Debug.cs
@@ -1,13 +1,15 @@
-﻿using System;
+using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime;
-using System.Runtime.Versioning;
+using System.Runtime.InteropServices;
 using System.Text;
 using Robust.Client.Input;
+using System.Threading;
 using Robust.Client.Interfaces;
 using Robust.Client.Interfaces.Console;
 using Robust.Client.Interfaces.Debugging;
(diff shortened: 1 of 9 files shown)


Robust.Client/Console/Commands/Debug.cs
```diff
@@ -735,25 +737,175 @@ public bool Execute(IDebugConsole console, params string[] args)
...
+                {
+                    if (_watchers != null)
+                    {
+                        console.AddLine("Already watching.");
+                        return false;
+                    }
+                    resC = IoCManager.Resolve<IResourceCache>();
+
+                    _watchers = new Dictionary<string, FileSystemWatcher>();
+                    var stringComparer = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? StringComparer.OrdinalIgnoreCase : StringComparer.Ordinal;
```
PJB3005: macOS is also case insensitive, and Linux can do it too (although you won't see it often). This selection isn't correct.

Robust.Client/ResourceManagement/ResourceTypes/ShaderSourceResource.cs
```diff
@@ -30,10 +32,24 @@ public override void Load(IResourceCache cache, ResourcePath path)
...
+                try
+                {
+                    using var stream = cache.ContentFileRead(path);
+                    using var reader = new StreamReader(stream, EncodingHelpers.UTF8);
+                    ParsedShader = ShaderParser.Parse(reader, cache);
+                    break;
+                }
+                catch (IOException ioe)
+                {
+                    if (!PathHelpers.IsFileInUse(ioe))
```
PJB3005, line 48: Is there no way to use file sharing modes to allow concurrent access?

PJB3005: Also maybe put a cap on the amount of attempts in this loop.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -172,10 +173,19 @@ private void LoadLightingShaders()
...
-                var shaderSource = _resourceCache.GetResource<ShaderSourceResource>(path);
-                return shaderSource.ClydeHandle;
+                try
+                {
+                    var shaderSource = _resourceCache.GetResource<ShaderSourceResource>(path);
+                    return shaderSource.ClydeHandle;
+                }
+                catch (Exception ex)
+                {
+                    Logger.Error($"Can't load shader {path}, {ex.GetType().Name}: {ex.Message}\n{ex.StackTrace}");
```
PJB3005: Can't you achieve all this `{ex.GetType().Name}: {ex.Message}\n{ex.StackTrace}` stuff by just... running `ToString()` on the exception?

PJB3005 (commented):
Copying over review comments from the first PR (and added some more):

Robust.Client/Console/Commands/Debug.cs
Tyler-IN: Got a means to determine case sensitivity?

Robust.Client/ResourceManagement/ResourceTypes/ShaderSourceResource.cs
Tyler-IN: Whoever gets the handle to the file first gets to determine the sharing mode.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
Tyler-IN: Yeah. I'll remove ex.StackTrace

Robust.Client/Console/Commands/Debug.cs
PJB3005: I think on Linux you have to check the directory attributes for a specific flag. On Windows/macOS you can just assume insensitive.

Assuming Linux is case sensitive is fine though. Even CoreCLR does it.

Issue #748, opened 2019-03-05
Improve LineEdit under Clyde.


PJB3005:
Features that are missing that kinda need implementing:

* Can't do text selection.
* Can't do copy/paste. This one's kinda icky because OpenTK does not expose a clipboard..
* Needs undo.

PJB3005:
Pasting is now implemented on Windows and Linux, doesn't work on macOS. Oh well.

ShadowCommander:
Another thing to add to this:
Scrolling text so that the cursor is always visible. (So that when the text entered is longer than the textbox it scrolls to the right to keep typed characters on screen.)

tentekal:
Adding myself to this because I'm assigned to the Content one

PR #1010, opened 2020-02-27
LineEdit Upgrades

tentekal:
Addresses issue space-wizards/space-station-14/#639 in the Content repo and #748 in RobustToolbox.

I'm planning to enhance further with text selection via cursor, copy-paste implemented, etc..

- [x] text scrolling
- [x] ctrl-arrow movement
- [x] copy/paste
~~- [ ] ctrl-shift selection~~ (I'd rather get this merged before doing bike shed priority)
- [x] cursor selection

### DIFF
--- a/Robust.Client/Input/EngineContexts.cs
+++ b/Robust.Client/Input/EngineContexts.cs
@@ -30,6 +30,10 @@ namespace Robust.Client.Input
             common.AddFunction(EngineKeyFunctions.TextCursorRight);
             common.AddFunction(EngineKeyFunctions.TextCursorBegin);
             common.AddFunction(EngineKeyFunctions.TextCursorEnd);
+            common.AddFunction(EngineKeyFunctions.TextCtrlLeft);
+            common.AddFunction(EngineKeyFunctions.TextCtrlRight);
+            common.AddFunction(EngineKeyFunctions.TextHome);
+            common.AddFunction(EngineKeyFunctions.TextEnd);
             common.AddFunction(EngineKeyFunctions.TextBackspace);
             common.AddFunction(EngineKeyFunctions.TextSubmit);
             common.AddFunction(EngineKeyFunctions.TextPaste);
--- a/Robust.Client/Input/EngineContexts.cs
+++ b/Robust.Client/Input/EngineContexts.cs
@@ -30,8 +30,10 @@ namespace Robust.Client.Input
             common.AddFunction(EngineKeyFunctions.TextCursorRight);
             common.AddFunction(EngineKeyFunctions.TextCursorBegin);
             common.AddFunction(EngineKeyFunctions.TextCursorEnd);
-            common.AddFunction(EngineKeyFunctions.TextCtrlLeft);
-            common.AddFunction(EngineKeyFunctions.TextCtrlRight);
+            common.AddFunction(EngineKeyFunctions.MoveWordLeft);
+            common.AddFunction(EngineKeyFunctions.MoveWordRight);
+            common.AddFunction(EngineKeyFunctions.SelectWordLeft);
+            common.AddFunction(EngineKeyFunctions.SelectWordRight);
             common.AddFunction(EngineKeyFunctions.TextHome);
             common.AddFunction(EngineKeyFunctions.TextEnd);
             common.AddFunction(EngineKeyFunctions.TextBackspace);
--- a/Robust.Client/Input/EngineContexts.cs
+++ b/Robust.Client/Input/EngineContexts.cs
@@ -36,6 +36,8 @@ namespace Robust.Client.Input
             common.AddFunction(EngineKeyFunctions.TextEnd);
             common.AddFunction(EngineKeyFunctions.TextBackspace);
             common.AddFunction(EngineKeyFunctions.TextSubmit);
+            common.AddFunction(EngineKeyFunctions.TextCopy);
+            common.AddFunction(EngineKeyFunctions.TextSelectAll);
             common.AddFunction(EngineKeyFunctions.TextPaste);
             common.AddFunction(EngineKeyFunctions.TextHistoryPrev);
             common.AddFunction(EngineKeyFunctions.TextHistoryNext);
--- a/Robust.Client/UserInterface/Controls/LineEdit.cs
+++ b/Robust.Client/UserInterface/Controls/LineEdit.cs
@@ -1,20 +1,15 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Linq;
 using System.Text.RegularExpressions;
 using JetBrains.Annotations;
 using Robust.Client.Graphics;
 using Robust.Client.Graphics.Drawing;
-using Robust.Client.Input;
-using Robust.Client.Interfaces.Input;
 using Robust.Client.Interfaces.UserInterface;
 using Robust.Shared.Input;
 using Robust.Shared.IoC;
-using Robust.Shared.Log;
 using Robust.Shared.Maths;
 using Robust.Shared.Timing;
-using SharpFont.TrueType;
 
 namespace Robust.Client.UserInterface.Controls
 {
@@ -28,7 +23,7 @@ namespace Robust.Client.UserInterface.Controls
         public const string StylePseudoClassPlaceholder = "placeholder";
 
         [NotNull] private string _text = "";
-        private int _overflowChars = 0;
+        private int _overflowChars;
         private bool _editable = true;
         [CanBeNull] private string _placeHolder;
         private int _cursorPosition;
@@ -83,25 +78,20 @@ namespace Robust.Client.UserInterface.Controls
 
                 if (!SetSelectionBuffer(value))
                 {
-                    return;
                 }
             }
         }
-
-        public int UnrenderedChars
+      
+        public int OverflowChars
         {
             get => _overflowChars;
             set
             {
-                if (value == null)
+                if (value < 0)
                 {
                     value = 0;
                 }
-
-                if (!SetUnrenderedChars(value))
-                {
-                    return;
-                }
+                _overflowChars = value;
             }
         }
 
@@ -171,16 +161,6 @@ namespace Robust.Client.UserInterface.Controls
             Text = "";
         }
 
-        public void StartSelect()
-        {
-            _startSelection = _cursorPosition;
-        }
-
-        public void EndSelect()
-        {
-            _endSelection = _cursorPosition;
-        }
-
         public void ClearSelection()
         {
             _selectedText = "";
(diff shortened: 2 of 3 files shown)


Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -63,6 +60,24 @@ public string Text
             }
         }
 
+        public string TextBuffer
```
PJB3005: You really should not be using a separate text buffer just to keep track of what to display.

Just automatically calculate an offset in the current text stored to keep track of where to start drawing.

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -297,13 +326,76 @@ protected internal override void KeyBindDown(GUIBoundKeyEventArgs args)
...
+
+
+                if (args.Function == EngineKeyFunctions.TextHome)
+                {
+                    if (!this.HasKeyboardFocus())
+                    {
+                        return;
+                    }
+
+                    _cursorPosition -= _text.Length;
```
PJB3005: Shouldn't you be able to achieve this by just setting current position to 0?

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -297,13 +326,76 @@ protected internal override void KeyBindDown(GUIBoundKeyEventArgs args)
...
+                }
+                else if (args.Function == EngineKeyFunctions.TextEnd)
+                {
+                    if (!this.HasKeyboardFocus())
+                    {
+                        return;
+                    }
+
+                    _cursorPosition += _text.Length;
```
PJB3005: You should set it to text.length, not add. Like this you'll have cursor position going out of bounds.

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -297,13 +326,76 @@ protected internal override void KeyBindDown(GUIBoundKeyEventArgs args)
...
+                        return;
+                    }
+
+                    int length = 0;
+                    int wordsLength = _text.Replace(" ", String.Empty).Length;
+                    var individualWords = _text.Split(" ");
+                    var numSpaces = individualWords.Length - 1;
+                    int wordEnd = 0;
+
+                    for (int index = 0; index < individualWords.Length; index++)
```
PJB3005: Uh, can't you just iterate over characters until you find two characters that form a word bound?

Robust.Client/UserInterface/Controls/LineEdit.cs
tentekal: Set this up, but left it as a property because each render cycle would reset the offset

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -207,7 +236,7 @@ protected internal override void Draw(DrawingHandleScreen handle)
             }
 
             var count = 0;
-            foreach (var chr in renderedText)
+            foreach (var chr in renderedText.Substring(startRenderPos))
```
PJB3005:
```suggestion
            foreach (var chr in renderedText.AsSpan(startRenderPos))
```
Use a span here so that you don't have to allocate the sub string. Decent performance boost.

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -176,6 +193,17 @@ protected bool SetText(string newText)
             return true;
         }
 
+        protected bool SetUnrenderedChars(int charCount)
+        {
+            if ( IsValid != null )
```
PJB3005: What?

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -63,6 +62,24 @@ public string Text
...
+                if (value == null)
+                {
+                    value = 0;
+                }
+
+                if (!SetUnrenderedChars(value))
+                {
+                    return;
+                }
+                _updatePseudoClass();
```
PJB3005: This should not modify the pseudo class. The text buffer is still the same.

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -63,6 +62,24 @@ public string Text
             }
         }
 
+        public int UnrenderedChars
+        {
+            get => _overflowChars;
+            set
+            {
+                if (value == null)
```
PJB3005, line 74: It's impossible to actually hit this if, as comparing an `int` with `null` will always fail.

PJB3005 (changes requested):

Robust.Client/UserInterface/Controls/LineEdit.cs
tentekal: Is this a pointer to a portion of the existing string? I didn't know that was possible, thanks.

Robust.Client/UserInterface/Controls/LineEdit.cs
tentekal: So that's what Rider was warning me about on compile...

Robust.Client/UserInterface/Controls/LineEdit.cs
tentekal: Removing this actually broke the ability to backspace and I have no idea why. Could you provide some insight?

tentekal:
I need to clean out some unneeded code, and I'm not happy with how highlighted characters are rendered (I don't see a performance hit, but I haven't profiled). 

Looking to apply best practices to this, but I believe it's ready to be out of draft as the functionality is there.

Robust.Shared/Input/KeyFunctions.cs
```diff
@@ -37,12 +37,20 @@ public static class EngineKeyFunctions
...
         public static readonly BoundKeyFunction TextBackspace = "TextBackspace";
         public static readonly BoundKeyFunction TextSubmit = "TextSubmit";
+        public static readonly BoundKeyFunction TextSelectAll = "TextSelectAll";
+        public static readonly BoundKeyFunction TextCopy = "TextCopy";
         public static readonly BoundKeyFunction TextPaste = "TextPaste";
         public static readonly BoundKeyFunction TextHistoryPrev = "TextHistoryPrev";
         public static readonly BoundKeyFunction TextHistoryNext = "TextHistoryNext";
         public static readonly BoundKeyFunction TextReleaseFocus = "TextReleaseFocus";
         public static readonly BoundKeyFunction TextScrollToBottom = "TextScrollToBottom";
+        public static readonly BoundKeyFunction TextShiftLeft = "TextShiftLeft";
```
PJB3005: You should not refer to `ShiftLeft` or `CtrlLeft` directly. Should instead use terms like "select left" or "move words left".

Also you need a combined select + words move version.

Robust.Client/Input/EngineContexts.cs
```diff
@@ -30,8 +30,14 @@ public static void SetupContexts(IInputContextContainer contexts)
             common.AddFunction(EngineKeyFunctions.TextCursorRight);
             common.AddFunction(EngineKeyFunctions.TextCursorBegin);
             common.AddFunction(EngineKeyFunctions.TextCursorEnd);
+            common.AddFunction(EngineKeyFunctions.TextCtrlLeft);
+            common.AddFunction(EngineKeyFunctions.TextCtrlRight);
+            common.AddFunction(EngineKeyFunctions.TextHome);
+            common.AddFunction(EngineKeyFunctions.TextEnd);
```
PJB3005, line 38: You're not registering ShiftLeft here.

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -25,11 +23,16 @@ public class LineEdit : Control
...
 
         [NotNull] private string _text = "";
+        private int _overflowChars;
         private bool _editable = true;
         [CanBeNull] private string _placeHolder;
         private int _cursorPosition;
         private float _cursorBlinkTimer;
         private bool _cursorCurrentlyLit;
+        private bool _mouseHeld;
+        private string _selectedText = "";
```
PJB3005, line 33: You should not have a separate selected text buffer. You should create a copy of the buffer (or better yet, create a `ReadOnlySpan<char>`) when necessary instead.

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -25,11 +23,16 @@ public class LineEdit : Control
...
+        private int _startSelection;
```
PJB3005, line 34: Your functions for modifying the selection don't modify these.

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -218,10 +270,24 @@ protected internal override void Draw(DrawingHandleScreen handle)
...
                 if (baseLine.X + metrics.Width + metrics.BearingX > contentBox.Right)
                 {
-                    break;
+                    _overflowChars += 1;
                 }
 
+                var selectionStart = Math.Min(_startSelection, _endSelection);
+
+                // If chr is selected, render a highlight over it
+                if ( _selectedText != "" && Enumerable.Range(selectionStart, _selectedText.Length).Contains(count) )
```
PJB3005, line 279: Keep track of the index of the current char while iterating and then just start rendering when said index is greater than `selectionStart`.

Also for performance it would be best if you stored the pixel position at which the selection starts and ends (by calculating it while rendering the main text) and then drew one rectangle after all text is drawn.

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -25,11 +23,16 @@ public class LineEdit : Control
         public const string StylePseudoClassPlaceholder = "placeholder";
 
         [NotNull] private string _text = "";
+        private int _overflowChars;
         private bool _editable = true;
         [CanBeNull] private string _placeHolder;
         private int _cursorPosition;
```
PJB3005, line 29: Does it still make sense to have a `_cursorPosition` when you have a selection start and end?

I know you'll still need to keep track of "which side of the selection is the cursor at" but otherwise it should match, shouldn't it?

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -292,18 +358,116 @@ protected internal override void KeyBindDown(GUIBoundKeyEventArgs args)
...
+                    {
+                        return;
+                    }
+
+                    var boundaryAheadPat = @"\b";
+                    var wordBoundaries = Regex.Matches(_text, boundaryAheadPat);
+            
+                    foreach (Match boundary in wordBoundaries)
+                    {
+                        if (boundary.Index < _cursorPosition)
```
PJB3005: Shouldn't this be `>`?

PJB3005 (changes requested):

PJB3005:
This PR is completely stale and I'd like to get this in.

PR #1012, opened 2020-02-29
Fix crash when using ControlMob on your owned mob when aghosting

### LINKED ISSUE #758: Server crashes if you try to control mob as adminghost.

gradientvera:
Should fix the crash.
Also, closes https://github.com/space-wizards/space-station-14/issues/758


gradientvera:
This wasn't the way to go about it

PR #1013, opened 2020-03-05
Entities can no longer collide with Entities that are inside their container.

### LINKED ISSUE #719: Container Items/Lockers are gunked
- If an item is placed in a locker the locker can no longer be pushed
- Sometimes when opening/closing a lock you can no longer pickup any items inside the locker
- Toolboxes can get in a weird state that cannot be picked up, instead they just open their inventory on click

PrPleGoo:
Fixes https://github.com/space-wizards/space-station-14/issues/719


PJB3005:
Wouldn't it make more sense to disable collision for things that are inside containers? That way if something with a large bounding box ends up in a locker it won't block *other* things.

PrPleGoo:
things bigger than a locker can't go inside one

PJB3005:
I feel like that's a nice thing in theory but won't work out in practice.

Exhibit A:

image

PrPleGoo:
Lockers not being able to eat objects bigger than themselves was broken (or fixed, depending on your definition) in:
https://github.com/space-wizards/space-station-14/commit/5aafe89d95c338ab1322e752f8bd9fa4021f6432#diff-602e515cf4fc47550c5fee59591617fb

I'd argue that things inside a locker bumping into things outside of it are out of scope for this PR which is aimed at fixing space-wizards/space-station-14#719

PrPleGoo:
>That way if something with a large bounding box ends up in a locker it won't block other things.

If the larger bounding box is disabled and the locker is placed snugly against a wall then opening the locker could lead to clipping the larger bounding box into the wall.
This is partially why lockers were implemented to not allow bigger entities inside them.

This PR fixes space-wizards/space-station-14#719
If you want to discuss whether or not bigger things should be inside a locker and how to handle their bounding box, I suggest making an issue labeled "Discussion" so it can be discussed and a decision can be made.

PrPleGoo:
Made an issue for the bigger things: https://github.com/space-wizards/RobustToolbox/issues/1022
When that is solved we can disable collision for things inside containers.
Until then the clipping issues would just be too common for the game to be playable.

PrPleGoo:
is this in purgatory?

PJB3005:
I guess I wasn't clear about this but since this doesn't fix "can move locker up to wall to get somebody stuck in wall", the approach of "disable colliders inside containers outright" would still be way better.

PrPleGoo:
>"disable colliders inside containers outright"

That would actually directly enable "can move locker up to wall to get somebody stuck in wall".

PJB3005:
No, because as we established on Discord you can do that even with this fix.

PrPleGoo:
That's because the thing inside the locker was not colliding with the wall.

PJB3005:
Well yeah but that doesn't mean I want to merge this fix. It makes the collision system assymetric (bad), it worsens collision performance, etc... All while a better content-side fix is possible.

PrPleGoo:
Don't you have a reject button for that?

PJB3005:
I guess? I kinda hate pressing it though.

PR #1014, opened 2020-03-05
Added unit tests for some collision code and a small rename

PrPleGoo:
Fixes nothing, just adds a unit test with the entire table of collision comparisons for a single bit


PJB3005:
I'm not sure this is a good way to fix this at all, unless @Acruid can confirm it's fine.

Collision being asymmetric between two bodies *feels* like a massive can of worms. Would another solution be to make items on the ground not collide? So only items actively being thrown would have collision.

PrPleGoo:
>I'm not sure this is a good way to fix this at all, unless @Acruid can confirm it's fine.

Reading the comments on CollisionLayer and -Mask in CollidableComponent this is how it should be implemented. I actually spent some time trying to fix the issue by changing the mask/layer for lockers and items only to find out that the comments lied to me. 
The change is a little controversial and I recognized that so I ran around colliding myself and a locker with different things without noticing any issues. Further testing may be required but all use-cases i knew of (shooting, doors, lasers, lockers, throwing) seem to still be working perfectly.

>Would another solution be to make items on the ground not collide? So only items actively being thrown would have collision.

Yes, but what about lamps?

PJB3005:
I'm not sure lamps should have collision at all that isn't for the purposes of attacks/clicks.

Acruid:
https://www.iforce2d.net/b2dtut/collision-filtering
Please read that. `IPhysBody` is our name for fixtures, the link explains and give examples of how the category and mask bits are supposed to work. We don't have group indexes, do we need them? The link's `ShouldCollide` predicate function is the `ICollideSpecial` interface in our code.

https://docs.godotengine.org/en/3.1/tutorials/physics/physics_introduction.html
Here Godot talks about how they implemented collision groups (they call them Layers) and masks. The `CollisionGroups` enum in our code is what is giving names/meaning to each bit.

The first link explains that the test needs to be symmetrical, your change is not correct, and this PR as-is will not be accepted.

Wall lights should only collide with bullets/thrown objects and be clickable, like most wall mounted entities. Lockers should collide with walls, low blockers like tables, and machines, and not collide with items. You have 32 different categories to put any entity in the game into, they can go in multiple categories, and they can collide with multiple categories.

PrPleGoo:
Changed the purpose of this PR, please rereview it.

PJB3005 (approved):

PR #1017, opened 2020-03-14
Fix issue651 and370

### LINKED ISSUE #651: Implement y-sorting
Right now, there doesn't seem to be any kind of sorting for mobs and other entities. This can make things look weird at times:
sort
Ideally, draw order for some (but not all, maybe?) entities should be relative to their Y position on the screen, making the ones that are further up drawn first, and the ones below drawn later.

PrPleGoo:
Fixes https://github.com/space-wizards/space-station-14/issues/651 by adding a Y comparison to the SpriteDrawingOrderComparer.

Switches Mobs and Items values in DrawDepth enum to solve https://github.com/space-wizards/space-station-14/issues/370
Fix to this relies on other PR (https://github.com/space-wizards/space-station-14/pull/788) that sets DrawDepth to Mobs in human entity.

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
@@ -838,6 +838,13 @@ namespace Robust.Client.Graphics.Clyde
                     return cmp;
                 }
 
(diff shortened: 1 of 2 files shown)


PJB3005:
I'd love for both of these issues to be fixable in a more flexible way in content but this is a decent bandaid for now.

PrPleGoo:
Well, 370 is fixed in content it just relies on the enum values being correct here.

PR #1028, opened 2020-04-19
Split HandleMessage into local and networked variants, add ICommonSession parameter to the latter

### LINKED ISSUE #704: Replace INetChannel with IPlayerSession instances in common events.
Net channels are pretty useless for common events like component messages. If you care where the message is from at all you'll always be resolving a player session.

Pretty annoying boilerplate and a leaky abstraction.

snowsignal:
Fixes #704. 
 
### What this PR does
Split the HandleMessage function for `IComponent` into two distinct functions, one for processing a local component message and one for processing a component message sent by a remote peer. It also adds a `Session` field to `NetworkComponentMessage`, which is null by default but can be specified by the `ServerEntityNetworkManager` as the session for the client who sent the component message.

The components which define a custom `HandleMessage` routine have also been updated. I still need to create a separate PR for the content repo that updates the components in there.

### DIFF
--- a/Robust.Client/GameObjects/Components/Light/PointLightComponent.cs
+++ b/Robust.Client/GameObjects/Components/Light/PointLightComponent.cs
@@ -7,6 +7,7 @@ using Robust.Shared.Interfaces.GameObjects;
 using Robust.Shared.Interfaces.Network;
 using Robust.Shared.IoC;
 using Robust.Shared.Maths;
+using Robust.Shared.Players;
 using Robust.Shared.Serialization;
 using Robust.Shared.ViewVariables;
 
@@ -117,10 +118,20 @@ namespace Robust.Client.GameObjects
         }
 
         /// <inheritdoc />
-        public override void HandleMessage(ComponentMessage message, INetChannel netChannel = null,
-            IComponent component = null)
+        public override void HandleMessage(ComponentMessage message, IComponent component)
         {
-            base.HandleMessage(message, netChannel, component);
+            base.HandleMessage(message, component);
+
+            if ((message is ParentChangedMessage msg))
+            {
+                HandleTransformParentChanged(msg);
+            }
+        }
+
+        /// <inheritdoc />
+        public override void HandleNetworkMessage(ComponentMessage message, INetChannel netChannel, ICommonSession session = null)
+        {
+            base.HandleNetworkMessage(message, netChannel, session);
 
             if ((message is ParentChangedMessage msg))
             {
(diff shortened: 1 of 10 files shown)


snowsignal:
Note: I don't think the components that are exclusive to the server-side or client-side need to implement `HandleNetworkMessage`, but I did just in case.

Robust.Client/GameObjects/Components/Light/PointLightComponent.cs
```diff
@@ -117,10 +118,20 @@ public float Radius
...
+            base.HandleMessage(message, component);
+
+            if ((message is ParentChangedMessage msg))
+            {
+                HandleTransformParentChanged(msg);
+            }
+        }
+
+        /// <inheritdoc />
+        public override void HandleNetworkMessage(ComponentMessage message, INetChannel netChannel, ICommonSession session = null)
```
PJB3005: Yeah this is an example of a message that only ever gets sent over local, so this doesn't need to override handle network message.

Robust.Client/GameObjects/Components/UserInterface/ClientUserInterfaceComponent.cs
```diff
@@ -44,10 +45,45 @@ public override void ExposeData(ObjectSerializer serializer)
             _interfaceData = interfaceData;
         }
 
-        public override void HandleMessage(ComponentMessage message, INetChannel netChannel = null,
-            IComponent component = null)
+        public override void HandleMessage(ComponentMessage message, IComponent component)
```
PJB3005: And this is an example of a message that only ever sent over the network, so the local variant isn't necessary.

Robust.Server/GameObjects/Components/Physics/PhysicsComponent.cs
```diff
@@ -108,9 +109,30 @@ public override ComponentState GetComponentState()
...
+                    if (!msg.Entity.TryGetComponent(out PhysicsComponent physicsComponent))
+                    {
+                        return;
+                    }
+                    physicsComponent.AddVelocityConsumer(this);
+                    break;
+            }
+        }
+
+        public override void HandleNetworkMessage(ComponentMessage message, INetChannel netChannel, ICommonSession session = null)
```
PJB3005: This one's only local so the network one can be removed.

Robust.Server/GameObjects/ServerEntityNetworkManager.cs
```diff
@@ -139,7 +139,9 @@ private void DispatchEntityNetworkMessage(MsgEntity message)
             switch (message.Type)
             {
                 case EntityMessageType.ComponentMessage:
-                    ReceivedComponentMessage?.Invoke(this, new NetworkComponentMessage(message));
+                    // Get player session
+                    var session = _playerManager.GetSessionByChannel(message.MsgChannel);
+                    ReceivedComponentMessage?.Invoke(this, new NetworkComponentMessage(message, session));
```
PJB3005: You already have the player session a few lines up, use that.

PR #1030, opened 2020-04-20
ServerEntityNetworkManager doesn't dispatch messages if the client got disconnected

### LINKED ISSUE #1029: ServerEntityNetworkManager throws if a client disconnects before their message is supposed to be handled
`ServerEntityNetworkManager` queues up incoming messages to play them later if they come in earlier than the client expected. The problem is that if you disconnect between the message coming in and it actually being handled, the server will throw while trying to fetch a player session for you.

Easy way to replicate this is to add `message.SourceTick += 60;` on [this line](https://github.com/space-wizards/RobustToolbox/blob/c770c2dd1195f6fb04bbafe00223898c9fe586ec/Robust.Server/GameObjects/ServerEntityNetworkManager.cs#L111). Then hit a button like move and immediately disconnect the client.

snowsignal:
Fixes #1029.

### What this PR does
Fixes a bug where the ServerEntityNetworkManager would attempt to retrieve the session of a disconnected client when dispatching messages.

### DIFF
--- a/Robust.Server/GameObjects/ServerEntityNetworkManager.cs
+++ b/Robust.Server/GameObjects/ServerEntityNetworkManager.cs
@@ -127,7 +127,7 @@ namespace Robust.Server.GameObjects
         private void DispatchEntityNetworkMessage(MsgEntity message)
         {
             // Don't try to retrieve the session if the client disconnected
-            if (!message.MsgChannel.NetPeer.IsConnected)
+            if (!message.MsgChannel.IsConnected)
             {
                 return;
             }
--- a/Robust.Server/GameObjects/ServerEntityNetworkManager.cs
+++ b/Robust.Server/GameObjects/ServerEntityNetworkManager.cs
@@ -126,6 +126,12 @@ namespace Robust.Server.GameObjects
 
         private void DispatchEntityNetworkMessage(MsgEntity message)
         {
+            // Don't try to retrieve the session if the client disconnected
+            if (!message.MsgChannel.NetPeer.IsConnected)
+            {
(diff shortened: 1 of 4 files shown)


Robust.Server/GameObjects/ServerEntityNetworkManager.cs
```diff
@@ -126,6 +126,12 @@ private void HandleEntityNetworkMessage(MsgEntity message)
 
         private void DispatchEntityNetworkMessage(MsgEntity message)
         {
+            // Don't try to retrieve the session if the client disconnected
+            if (!message.MsgChannel.NetPeer.IsConnected)
```
PJB3005: `NetPeer.IsConnected` on the server just reads whether ANY client is connected, not specifically the one that `MsgChannel` came from. You'll need to add one on `NetChannel` itself.

Robust.UnitTesting/RobustIntegrationTest.NetManager.cs
```diff
@@ -299,6 +299,8 @@ private sealed class IntegrationNetChannel : INetChannel
                 public int ConnectionUid { get; }
                 long INetChannel.ConnectionId => ConnectionUid;
 
+                public bool IsConnected { get; }
```
PJB3005, line 302: You're gonna have to initialize this to `true` for integration tests.

snowsignal: Alright, wasn't sure if I should leave it uninitialized since it isn't used within the tests.

PR #1031, opened 2020-04-21
Gravity

snowsignal:
Link to relevant content PR: https://github.com/space-wizards/space-station-14/pull/841

### What this PR does

In order to add Gravity to SS14, a few minor engine changes must be made:
- Each grid should keep track of whether or not it has gravity
- The physics system should not apply friction to moving objects in zero gravity

In addition, I fixed a bug involving asymmetric collisions. This bug would happen whenever two objects collided, one of which had a `PreventCollide` override that returned true, and the other that did not. The former would not collide with the later, but the later would still act as if it had collided with the former.

#### Completion Checklist
- [X] Grid has a gravity property
- [X] Friction is not applied to objects moving in zero gravity
- [X] Space will always have zero gravity

### DIFF
--- a/Robust.Shared/Physics/PhysicsManager.cs
+++ b/Robust.Shared/Physics/PhysicsManager.cs
@@ -99,10 +99,10 @@ namespace Robust.Shared.Physics
                 foreach (var mods in collisionmodifiers)
                 {
                     preventcollision |= mods.PreventCollide(otherCollidable);
-                    foreach (var othermods in othercollisionmodifiers)
-                    {
-                        preventcollision |= othermods.PreventCollide(collidable);
-                    }
+                }
+                foreach (var othermods in othercollisionmodifiers)
+                {
+                    preventcollision |= othermods.PreventCollide(collidable);
                 }
                 if (preventcollision)
                 {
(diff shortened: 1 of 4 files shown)


Robust.Shared/Physics/PhysicsManager.cs
```diff
@@ -92,11 +92,17 @@ public bool TryCollide(IEntity entity, Vector2 offset, bool bump = true)
...
                 }
 
+                var othercollisionmodifiers = otherCollidable.Owner.GetAllComponents<ICollideSpecial>();
+
                 //Provides component level overrides for collision behavior based on the entity we are trying to collide with
                 var preventcollision = false;
                 foreach (var mods in collisionmodifiers)
                 {
                     preventcollision |= mods.PreventCollide(otherCollidable);
+                    foreach (var othermods in othercollisionmodifiers)
```
PJB3005: Does this have to be inside the other loop? This doesn't have to process based on `mods` does it?

PJB3005 (changes requested):

Robust.Shared/Physics/PhysicsManager.cs
snowsignal: You're right, I'll fix that.

PJB3005 (approved):

PR #1040, opened 2020-05-01
Change Reference -> Ref in VV

### LINKED ISSUE #984: Use abbreviations on all vv variables & labels
We already have some abbreviations in the vv menu. Notably in the component lists:

Content Client_pS62ovGN1v
So, for example, `C.C.GO` = `Content.Client.GameObjects`. However, in some other places like on component members, menu labels, and server/client variables these abbreviations aren't used yet. Examples:

Content Client_kcr7O8OeEQ

Content Client_EGJVNRFyem

Content Client_SGJpC3CboK

ComicIronic:
Fixes #984.

Since `1ab703325fdd3a363fe29ab5421771cf45ef4803` already abbreviated types where possible, the only remaining part of the issue was shortening `Reference` to `Ref` in the menu.

### DIFF
--- a/Robust.Client/ViewVariables/Editors/ViewVariablesPropertyEditorReference.cs
+++ b/Robust.Client/ViewVariables/Editors/ViewVariablesPropertyEditorReference.cs
@@ -28,7 +28,7 @@ namespace Robust.Client.ViewVariables.Editors
 
             var button = new Button
             {
-                Text = $"Reference: {toString}",
+                Text = $"Ref: {toString}",
                 ClipText = true,
                 SizeFlagsHorizontal = Control.SizeFlags.FillExpand
             };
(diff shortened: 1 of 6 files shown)


PJB3005:
There's still a few other places where unabbreviated types are used, for example the header: 
image

Wouldn't say that the issue is quite fixed yet.

ComicIronic:
I can be confident that anything called `Pretty` is user-facing, and has yet to be abused by some brittle code, correct?

PJB3005:
Yup

ComicIronic:
That should be all the types that appear in window titles and the like.

I've noticed that generics get shortened incorrectly, (probably) because `FullName` for a generic type doesn't look like its `ToString`. Generic type arguments don't get the abbreviation either. That could be solved by making `Abbreviate` take a `Type` instead, and doing some custom printing logic.

PJB3005:
That'd probably be best, yeah

Issue #899, opened 2019-11-21
Compile-time checked functions for referring to prototypes/resources.


PJB3005:
The idea is that you could do something like 
```cs
SpawnEntity(EntityName("wrench"));
```
and a build step would automatically verify that an entity prototype with ID `wrench` exists in the prototypes.

There could be some other variations on this too, such as giving `SpawnEntity` a parameter attributes that does this automatically (for constants) and other such stuff.

PR #1043, opened 2020-05-02
Add Roslyn analyzer for checking prototype names

ComicIronic:
Based on #899, this introduces some static analysis to catch mistakes in the prototype name passed to a method, configurable by an attribute on method parameters.

Goals
  * [x] MVP: an analyzer that errors when the prototype name is wrong (loaded from a yaml file)
  * [x] Attribute to mark method argument as a prototype name
  * [ ] Automatic loading of prototypes from `<location>`


ComicIronic:
There's lots of room for improvement e.g. tagging method arguments in the engine as prototype names so that they get analyzed automatically, rather than just looking at `SpawnEntity`. This is mainly an attempt to produce an MVP for the analyzer itself.

PJB3005:
For what it's worth, I can build this on Linux just fine.

ComicIronic:
Ah good, I suppose the compiler SDK was only necessary for the initial project template then. I can't see  a reason for the Travis failure - `Robust.Shared` is referenced by `Robust.Analyzer`.

PJB3005:
Seems to be because you added an AnyCPU target to the solution, which the build commands are now defaulting to.

ComicIronic:
Yep, just pushed the fix for that. The sln file should be correct now.

Robust.Analyzer/CheckPrototypesExistAnalyzer.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            try
+            {
+                return prototypeManager.HasIndex<EntityPrototype>(prototypeName);
+            }
+            catch (UnknownPrototypeException e)
+            {
+                throw new Exception(String.Format("Exception prevented querying of entity prototypes for existence of \"{0}\".", prototypeName), e);
+            }
+        }
```
ComicIronic: @PJB3005 Is this the right way to interface with an `IPrototypeManager`? `HasIndex` is throwing an exception.

PJB3005: `PrototypeManager` needs to be able to discover the prototype-types, which is done via `IReflectionManager`. You're just instantiating one straight up though which means it's not getting dependencies injected which is problematic.

I wonder if it would make sense to just not use the prototype manager itself directly and parse only the `id`/`type` of prototype declarations in YAML. Then you also wouldn't have the issue of "would need to load `Content.Shared` to do good analysis"

ComicIronic: Wouldn't that make it more brittle? I'm conscious that tooling that has to be maintained by multiple people tends to get abandoned, so it would be good if this was more tied to how the prototype system works.

PJB3005: Unless we want to get into parsing the actual contents of prototypes it should be "good enough" I would think.

I'm not too familiar with Roslyn analyzers though. Could you set up a whole IoC instance? If you can then using the shared code is probably the best option.

ComicIronic: It seems  like an IoC instance is possible. The main remaining problem is where the analyzer should load prototype files from - currently I'm passing them in to `AdditionalDocuments` for testing, but it should probably reflect some project-set value in reality.

ComicIronic: Turns out it was possible.

ComicIronic:
@PJB3005 Would appreciate a double-check that there's no `[PrototypeName]` missing from any important methods.

Robust.Shared/Prototypes/PrototypeNameAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Robust.Shared.Prototypes
+{
+    /// <summary>
+    /// A metadata attribute which marks an argument as being a prototype name.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = true)]
+    public class PrototypeNameAttribute : Attribute
+    {
+        public PrototypeNameAttribute() { }
```
PJB3005: Might be worth it to make this take in a constructor argument for what kind of prototype to check for, so that in the future we can expand it to other prototypes like reagents.

ComicIronic: What kind of argument, a string? Having to go between the compilation types and the library types makes a lot of reflection impossible. Should I extend the `IPrototypeManager` with an (unsafe) method to try and load a prototype by passing in the type by name?

PJB3005: > What kind of argument, a string? Having to go between the compilation types and the library types makes a lot of reflection impossible.

Yeah, string is probably the best idea, specifically to keep the reflection down. 

> Should I extend the IPrototypeManager with an (unsafe) method to try and load a prototype by passing in the type by name?

Probably, yeah.

ComicIronic:
@PJB3005 The analyzer is limited by its ability to hook into useful information about the project it's compiling, like `ReflectionManager` prefixes and the `PrototypeManager` loading directories. It might be worthwhile to work on some kind of project config file that could also be fed to the analyzer.

PJB3005:
That's probably a good idea, it'd also solve a bunch of stuff like the need to register prototypes as ignored.

ComicIronic:
Blocked by #1044.

PR #1037, opened 2020-04-27
Rewrite the physics engine

snowsignal:
Corresponding content PR: https://github.com/space-wizards/space-station-14/pull/872

Fixes #1046 and fixes #982.

Note: client-side physics will probably have to be implemented separately, because otherwise this PR would take forever to review and test. But my redesign introduces the groundwork for making such an implementation easier.

### Completion checklist
- [X] CollidableComponent API has new functions to retrieve all entities colliding with it
- [X] Implement momentum/impulses in physics system
- [X] Implement working impulse resolution
- [X] Implement virtual forces
- [x] Make the system work with gravity
- [X] Make physics system cache collision results for increased performance
- [X] Update the rest of the codebase to match the new physics engine
- [X] Update implementation tests and add some new ones
- [x] Add a raycast that gets all entities in a certain path to IPhysicsManager

PR #1046, opened 2020-05-05
Add GetCollidingEntities to the API

### LINKED ISSUE #1036: CollidableComponent should be able to retrieve entities it is colliding with
Right now, `TryCollision` only returns a boolean value, not the entities that are causing the collision. It would be nice to have a `GetCollidingEntities` function that would retrieve all entities that are colliding with the object.

ComicIronic:
Fixes #1036.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Collidable/CollidableComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Collidable/CollidableComponent.cs
@@ -232,7 +232,7 @@ namespace Robust.Shared.GameObjects.Components
             if (!_collisionEnabled || CollisionMask == 0x0)
                 return new IEntity[] { };
 
-            return null;
+            return _physicsManager.GetCollidingEntities(Owner, offset, bump);
         }
 
         public bool UpdatePhysicsTree()
--- a/Robust.Shared/GameObjects/Components/Collidable/CollidableComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Collidable/CollidableComponent.cs
@@ -226,6 +226,15 @@ namespace Robust.Shared.GameObjects.Components
             return _physicsManager.TryCollide(Owner, offset, bump);
         }
 
+        /// <inheritdoc />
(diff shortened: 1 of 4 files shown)


Robust.Shared/GameObjects/Components/Collidable/CollidableComponent.cs
```diff
@@ -226,6 +226,15 @@ public bool TryCollision(Vector2 offset, bool bump = false)
             return _physicsManager.TryCollide(Owner, offset, bump);
         }
 
+        /// <inheritdoc />
+        public IEnumerable<IEntity> GetCollidingEntities(Vector2 offset, bool bump = false)
+        {
+            if (!_collisionEnabled || CollisionMask == 0x0)
+                return new IEntity[] { };
+
+            return null;
```
ComicIronic: Hwoops.

snowsignal:
Thanks for doing this, but this has already been added as part of my physics rewrite PR #1037. I should have added #1036 to the list of issues that the PR was going to fix.

Robust.Shared/Physics/PhysicsManager.cs
```diff
@@ -109,21 +120,21 @@ public bool TryCollide(IEntity entity, Vector2 offset, bool bump = true)
...
-                    continue;
+                    otherCollidable.Bumped(entity);
                 }
-
-                otherCollidable.Bumped(entity);
-                collidedwith.Add(otherCollidable.Owner);
             }
 
-            collidable.Bump(collidedwith);
+            if (bump)
```
ComicIronic, line 131: @InquisitivePenguin Does your PR also fix this missing check in some way?

snowsignal: My PR removes the whole 'bump' concept entirely, so... technically yes?

gradientvera:
Mind if I close this in favor of #1037 then? @ComicIronic @InquisitivePenguin

ComicIronic:
Sounds good.

PR #1047, opened 2020-05-05
Add system for using flag names in YAML

### LINKED ISSUE #974: Collision layers need to be human readable
They're bitflags serialized as decimals and there's no apparent definition for which layer is which bit at the moment. Probably should be defined as a list of strings in the YAML and converted to the bitflag on serialization.

ComicIronic:
It's now possible to hook an enum into the serialization and deserialization of a field in `DataField`, using an attribute. This allows for bitflags in the engine to be defined in the content, in a way that lets content creators write flag names in prototype YAML.

This works by annotating an enum with the [FlagsFor(<tagtype>)] attribute, which the YamlObjectSerializer picks up through the ReflectionManager to use as a special serializer for fields with the right invocation with that tag type.

Fixes #974.

### DIFF
--- a/Robust.Shared/Serialization/YamlCustomFormatSerializer.cs
+++ b/Robust.Shared/Serialization/YamlCustomFormatSerializer.cs
@@ -1,0 +1,96 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using YamlDotNet.Core;
+using YamlDotNet.RepresentationModel;
+
+namespace Robust.Shared.Serialization
+{
+    /// <summary>
+    /// Helper for doing custom value representation in YAML.
+    /// </summary>
+    /// <typeparam name="T">The type for which this gives a custom representation.</typeparam>
+    public class YamlCustomFormatSerializer<T> : YamlObjectSerializer.TypeSerializer
+    {
+        private readonly WithFormat<T> _formatter;
+
+        public YamlCustomFormatSerializer(WithFormat<T> formatter)
+        {
+            _formatter = formatter;
+        }
+
+        public override object NodeToType(Type _type, YamlNode node, YamlObjectSerializer serializer)
+        {
+            return _formatter.FromCustomFormat(serializer.NodeToType(_formatter.Format, node));
+        }
+
+        public override YamlNode TypeToNode(object obj, YamlObjectSerializer serializer)
+        {
+            var t = (T)obj;
+            return serializer.TypeToNode(_formatter.ToCustomFormat(t));
+        }
+    }
+
+    /// <summary>
+    /// Conveniece class for static access to custom formatters.
+    /// </summary>
+    public static class WithFormat
+    {
+        // This is concurrent because it can possibly be updated from multiple threads
+        // calling WithFormat at the same time.
+        private static readonly ConcurrentDictionary<Type, WithFormat<int>> _flagFormatters = new ConcurrentDictionary<Type, WithFormat<int>>();
+
+        public static WithFormat<int> Flags<T>()
+        {
+            return _flagFormatters.GetOrAdd(typeof(T), _type => new WithFlagRepresentation<T>());
+        }
+    }
+
+    /// <summary>
+    /// Interface for controlling custom value representation in an abstract storage medium.
+    /// </summary>
+    public abstract class WithFormat<T>
+    {
+        /// <summary>
+        /// The underlying type used for representation in the storage medium.
+        /// </summary>
+        public abstract Type Format { get; }
+
+        public abstract T FromCustomFormat(object obj);
+        public abstract object ToCustomFormat(T t);
+
+        /// <summary>
+        /// Get the corresponding YAML serializer for the custom representation.
+        /// </summary>
+        public virtual YamlObjectSerializer.TypeSerializer GetYamlSerializer()
+        {
+            return new YamlCustomFormatSerializer<T>(this);
+        }
+
+        private class DoNothing<T> : WithFormat<T>
+        {
+            public override Type Format => typeof(T);
+            public override T FromCustomFormat(object obj) { return (T)obj; }
+            public override object ToCustomFormat(T t) { return t; }
+
+            private static YamlCustomFormatSerializer<T> _serializer;
+
+            internal DoNothing()
+            {
+                if (_serializer == null)
+                    _serializer = new YamlCustomFormatSerializer<T>(this);
+            }
+
+            public override YamlObjectSerializer.TypeSerializer GetYamlSerializer()
+            {
+                return _serializer;
+            }
+        }
+
+        /// <summary>
+        /// The identity format i.e. the format which represents a value as itself.
+        /// </summary>
+        public static readonly WithFormat<T> NoFormat = new DoNothing<T>();
+    }
+}
--- a/Robust.Shared/Serialization/YamlCustomFormatSerializer.cs
+++ b/Robust.Shared/Serialization/YamlCustomFormatSerializer.cs
@@ -1,3 +1,5 @@
+﻿using Robust.Shared.Interfaces.Serialization;
+using Robust.Shared.IoC;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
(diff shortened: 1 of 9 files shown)


ComicIronic:
I'm not sure the matching on field name strings is necessarily the best choice, but I don't see a clear alternative. If someone is comfortable with `TypeSerializer`s, we might be able to make the `FlagsFor` attribute take a type argument as well.

Robust.Shared/Serialization/YamlObjectSerializer.cs
```diff
@@ -47,6 +49,26 @@ static YamlObjectSerializer()
                 { typeof(MapId), new MapIdSerializer() },
                 { typeof(SpriteSpecifier), new SpriteSpecifierSerializer() },
             };
+
+            var reflectionManager = IoCManager.Resolve<IReflectionManager>();
```
PJB3005: Doing this in the static constructor is a very bad idea for two reasons:

1. It means the data is shared between different IoC instances, which could be a problem for stuff such as integration tests which use it to split client/server instances.
2. There's no real good guarantee that the static constructor runs *after* content has been initialized, in which case this would break.

Robust.Shared/Serialization/YamlObjectSerializer.cs
```diff
@@ -460,15 +482,21 @@ public override bool TryGetCacheData<T>(string key, out T data)
...
             // special snowflake string
             if (type == typeof(String))
                 return node.ToString();
 
             // val primitives
             if (type.IsPrimitive || type == typeof(decimal))
+            {
+                if (nodeName != null && _flagNameSerializers.TryGetValue(nodeName, out var flagSerializer))
+                {
+                    return flagSerializer.NodeToFlags(node, this);
```
PJB3005: Tying this to the name of the node isn't gonna be sufficient. For example then we'd need to have multiple `[FlagsFor]` attributes on collision layers, which would then also mark ANY variable called `layer` and `mask` as collision layers.

PJB3005 (commented):
Would be nice if this could be extended to other types of values too, like `DrawDepth`.

Robust.Shared/Serialization/YamlObjectSerializer.cs
ComicIronic: I tried to avoid it with `IPostInjectInit`, but I didn't see how to get `YamlObjectSerializer` to be injected into by the IoC system. How should I do that?

Robust.Shared/Serialization/YamlObjectSerializer.cs
ComicIronic: Like I say, I don't immediately see a better alternative.

Unless you would prefer some kind of newtype wrapper around `int` in the engine, that then gets custom-serialization based on an attribute in the content? Which would solve the multiple-attributes/field-name problems, but would require rewriting a lot of engine code, probably.

The main complication for me is the use of `DataField`. Otherwise some kind of metadata attribute could go directly on the property being serialized.

PJB3005: I would love if serialization was instead to-object for flexibility and performance reasons like this, yeah.

Could maybe provide some simple options parameter to `DataField` and friends that can contain flexible options like this? Something like `serializer.DataField(..., SerializationOptions.EnumType("CollisionLayer")`

ComicIronic:
@PJB3005 How's this? It's made the PR a bit chunkier, but the new system should be extensible enough to allow for future custom serialization/deserialization, including using enum names as constants.

Robust.Shared/Serialization/YamlObjectSerializer.cs
PJB3005: You're gonna have to store this data on a manager *somewhere*, not sure where the best place would be. Probably ReflectionManager?

ComicIronic: The new code no longer does this in a static initializer. Do your comments still apply?

Robust.Shared/Serialization/YamlCustomFormatSerializer.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            public override Type Format => typeof(T);
+            public override T FromCustomFormat(object obj) { return (T)obj; }
+            public override object ToCustomFormat(T t) { return t; }
+
+            private static YamlCustomFormatSerializer<T> _serializer;
+
+            internal DoNothing()
+            {
+                if (_serializer == null)
```
PJB3005: Any reason you're not just initializing this in the declaration? It'd solve a theoretical race condition here and just be cleaner.

ComicIronic: Can't refer to `this` in a static initializer. I could lock it, but this constructor will be called a lot in serialization, and the race condition doesn't have any affect at the moment.

Robust.Shared/Serialization/YamlCustomFormatSerializer.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        public override YamlNode TypeToNode(object obj, YamlObjectSerializer serializer)
+        {
+            var t = (T)obj;
+            return serializer.TypeToNode(_formatter.ToCustomFormat(t));
+        }
+    }
+
+    /// <summary>
+    /// Conveniece class for static access to custom formatters.
```
snowsignal: Ultra-minor nitpick but this should be 'convinience'.

Robust.Shared/Serialization/YamlFlagSerializer.cs
```diff
@@ -0,0 +1,240 @@
...
+                        foreach (var flagsforAttribute in bitflagType.GetCustomAttributes<FlagsForAttribute>(true))
+                        {
+                            if (typeof(T) == flagsforAttribute.Tag)
+                            {
+                                if (_flagType != null)
+                                {
+                                    throw new NotSupportedException(
+                                        String.Format(
+                                            "Multiple bitflag enums declared for the tag {0}.",
+                                            flagsforAttribute.Tag));
```
snowsignal: Maybe use string interpolation instead, e.g. `"Multiple bitflag enums declared for the tag ${flagsforAttribute.Tag}"`

Robust.Shared/Serialization/YamlFlagSerializer.cs
```diff
@@ -0,0 +1,240 @@
...
+        public override Type Format => typeof(List<string>);
+
+        public override int FromCustomFormat(object obj)
+        {
+            var flagNames = (List<string>)obj;
+            var flags = 0;
+
+            foreach (var flagName in flagNames)
+            {
+                flags |= (int)Enum.Parse(_flagType, flagName);
```
snowsignal: Is this cast guarenteed to succeed?

snowsignal (changes requested):

Robust.Shared/Serialization/YamlFlagSerializer.cs
ComicIronic: For the default declaration of enums, yes,  they are backed by `int`. For enums which are not `int`, using this flag serialization is a mistake, since it only works for fields which are `int` underneath. I'll add a check for that to `WithFlagRepresentation`, but I think that's the earliest it can happen.

Robust.Shared/Serialization/YamlFlagSerializer.cs
ComicIronic: Weirdly, I do that below.

Robust.Shared/Serialization/YamlObjectSerializer.cs
PJB3005: I missed this but yes there is still the issue of, because it uses a static, this could theoretically be shared between two separate game instances.

ComicIronic: So what should I do? Add a `ICustomFormatManager`, and register an instance with every IoC instance?

PJB3005: Probably, yeah.

ComicIronic: Alright, done.

PR #1055, opened 2020-05-13
Replaces noSprite.png with Gmod Error Sprite

SweptWasTaken:
It's just a nonanimated png right now because I couldn't figure out how to load an rsi with 
`ResourcePath`. 

https://github.com/space-wizards/RobustToolbox/blob/b7f3627ef179c272152b89189002f1b34cc46355/Robust.Client/ResourceManagement/ResourceTypes/TextureResource.cs#L14

I included the animated .rsi if anyone can add to this pr and make it work.

Looks cool doe


PJB3005:
You can't make it an animated fallback because the whole point is to return a single texture.

SweptWasTaken:
Ok then use the static png then?

PJB3005:
Yes.

I do think I could write some code to make sprite component use this as a animated fallback explicitly (probably a good idea tbh) but making everything use the animated fallback is (sanely) impossible.

PJB3005:
The best part about this is that this made me actually look at failure paths in case of missing RSI/states so the game is significantly less prone to blowing up from these things now.

PR #1058, opened 2020-05-16
Implement text macros

Hugal31:
This is some draft work on text macros.

- It currently only have English pronouns.
- Also, to make it work with NGettext, ~~I had to subclass `CultureInfo` to override its implementation of `IFormatProvider`. I am not convinced this is a good design.~~ I created a subclass of `Catalog`.
- I am not really happy with the implementation of `IGenderable.GetGenderOrEpicene`, which have a "special case" for IEntity.

This is also missing "Name" and "The" macros.


Robust.Shared/Localization/Macros/Attributes.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        public Gender Gender => Gender.Epicene;
+
+        /// <summary>
+        /// Helper function to get the gender of an object, or Epicene if the object is not IGenderable
+        /// </summary>
+        public static Gender GetGenderOrEpicene(object argument)
+        {
+            // FIXME The Entity special case is not really good
+            if (argument is IEntity entity)
```
ComicIronic, line 27: Could implement `IGenderable` as an interface on `IEntity` - though I don't know how others feel about that as a design decision.

Hugal31: Yeah, I thought about that, but I think it would quickly become bloated with tons of unrelated features.

Another idea is to systematically wrap the entity with something implementing IGenderable + other "grammar attributes" getters. This could be done in `Loc` or at the call.

ComicIronic: I don't understand - where would the bloat come from?

Hugal31: If adding implementations to IEntity be the default solution for everything, it will become another god-interface.
Just with grammar-related features, let's say we want a firstname, surname, occupation name macros, or whatever, that's more interface to implement.

Robust.Shared/Localization/Macros/Attributes.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Robust.Shared.Interfaces.GameObjects;
+
+namespace Robust.Shared.Localization.Macros
+{
+    /// <summary>
+    /// "Possible" gender for grammatical usage only.
+    /// Arranged in alphabetical order so as not to offend anyone.
+    /// </summary>
+    public enum Gender
```
PJB3005, line 9: Does having a plural gender make sense here? I know BYOND has one but not sure if it's useful.

Robust.Shared/Localization/Macros/Attributes.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Robust.Shared.Interfaces.GameObjects;
+
+namespace Robust.Shared.Localization.Macros
+{
+    /// <summary>
+    /// "Possible" gender for grammatical usage only.
+    /// Arranged in alphabetical order so as not to offend anyone.
```
PJB3005: I think maybe not bringing it up in the first place is probably best.

PJB3005 (commented):
> This is some draft work on text macros. It currently only have English pronouns. Also, to make it work with NGettext

Could maybe consider shooting a PR to NGetText for this.

Robust.Shared/Localization/Macros/Attributes.cs
Hugal31: Fair point

Robust.Shared/Localization/Macros/Attributes.cs
Hugal31: I was thinking about another attribute. For instance, let's say you have a steel sheet. The ItemComponent (or a separated component) could be giving `Gender == Neutral`, but the StackComponent could be saying "Number = 3".
In addition, I think I remember some languages (Russian ?) have different plurals for different numbers, so an enum could not be enough.

Hugal31: Also some languages might have different plurals for the different genders.
And I am not talking about Japanese, which have like 20 ways to count things depending of the type (flat, large, small, clothes, small living, big living, etc...)

PJB3005: Yeah plurals kinda suck in various languages. Gettext (and by extension NGettext) has a system for this including like an arbitrary expression parser to select plural forms.

Hugal31:
> Could maybe consider shooting a PR to NGetText for this.

I opened an issue. Meanwhile, I could simply subclass `NGettext.Catalog`, there are only 4 functions to change.

Hugal31:
I added a register macro for text macros, its much cleaner now.

PR #1066, opened 2020-05-23
Hot reloading for YAML files

### LINKED ISSUE #1064: Hot reload YAML prototypes
Instead of having to restart the application every single time you make a change to the YAML, what if the game automatically updated whenever you made a change?

### Some ideas on how to do this
- Make this a config option in `server_config.toml`, so that actual servers won't waste time checking for YAML updates
- Every 2 seconds or so (120 ticks if we're doing 60fps), check if any files in the Resources folder have been updated (the simplest way to do this that I can think of would be to hash all the files and compare it with the last hash that was taken)
- Another way to reload files could be through a dedicated console command
- If the files changed, perform a reload of all the prototypes. Changes would be applied by comparing the new and old versions of the prototypes and performing updates on all existing entities in the world (removing components, adding new default components, updating sprites/bounding boxes, etc.)

I plan to work on this once the physics rewrite is finished.

snowsignal:
Closes #1064.

This will allow you to reload prototype files while in game, which will update the game accordingly. It will be possible to do this both automatically and manually via the `reload` command.

## To do
- [X] Implement reloading for entity prototypes
- [X] Preserve the state of entity components if they have been modified since their creation
- [ ] Implement reloading for other prototype variants
- [ ] Make reloading automatic, which can be enabled/disabled in `server_config.toml`
- [ ] Send messages so that things get updated client-side too

### DIFF
--- a/Robust.Server/Console/Commands/ReloadCommand.cs
+++ b/Robust.Server/Console/Commands/ReloadCommand.cs
@@ -1,0 +1,41 @@
+using JetBrains.Annotations;
+using Robust.Server.Interfaces.Console;
+using Robust.Server.Interfaces.Player;
+using Robust.Shared.GameObjects;
+using Robust.Shared.Interfaces.GameObjects;
+using Robust.Shared.IoC;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Utility;
+
+namespace Robust.Server.Console.Commands
+{
+    [UsedImplicitly]
+    public class ReloadCommand: IClientCommand
+    {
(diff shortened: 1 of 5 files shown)


Tyler-IN:
If this isn't _debug only_ it probably needs to set some public static property identifying when content has been reloaded at least once so e.g. shared string mapping can turn itself off or avoid trying to reinit and just stick to the initial bit if it uses strings from YAML in the future.

snowsignal:
@Tyler-IN this is debug-only.

PR #1068, opened 2020-05-23
Implement audio occlusion and upgrade to OpenTK 4

Clyybber:
also fix a bug with IntersectRayWithPredicate, where the predicate would be called on everything, ignoring the CollisionMask

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
@@ -13,8 +13,11 @@ using Robust.Shared.Interfaces.Map;
 using Robust.Shared.Map;
 using Robust.Shared.Utility;
 using System.Collections.Generic;
+using Content.Shared.Physics;
 using JetBrains.Annotations;
 using Robust.Client.Interfaces.Graphics.ClientEye;
+using Robust.Shared.Interfaces.Physics;
+using Robust.Shared.Maths;
 
 namespace Robust.Client.GameObjects.EntitySystems
 {
@@ -103,7 +106,20 @@ namespace Robust.Client.GameObjects.EntitySystems
                     }
                     else
                     {
+                        var occlusion = 0;
+                        if ((_eyeManager.CurrentEye.Position.Position - pos.Position).Length > 0)
+                        {
+                            IoCManager.Resolve<IPhysicsManager>().IntersectRayWithPredicate(
+                                pos.MapId,
+                                new CollisionRay(
+                                    pos.Position,
+                                    (_eyeManager.CurrentEye.Position.Position - pos.Position).Normalized,
+                                    (int) (CollisionGroup.Impassable)),
+                                (pos.Position - _eyeManager.CurrentEye.Position.Position).Length,
+                                (e) => { occlusion++; return false; }, false);
+                        }
                         stream.Source.SetVolume(stream.Volume);
+                        stream.Source.SetOcclusion(occlusion);
                     }
 
                     if (!stream.Source.SetPosition(pos.Position))
--- a/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
@@ -105,15 +105,16 @@ namespace Robust.Client.GameObjects.EntitySystems
                     else
                     {
                         var occlusion = 0;
-                        if ((_eyeManager.CurrentEye.Position.Position - pos.Position).Length > 0)
+                        var sourceRelative = _eyeManager.CurrentEye.Position.Position - pos.Position;
+                        if (sourceRelative.Length > 0)
                         {
                             IoCManager.Resolve<IPhysicsManager>().IntersectRayWithPredicate(
                                 pos.MapId,
                                 new CollisionRay(
                                     pos.Position,
-                                    (_eyeManager.CurrentEye.Position.Position - pos.Position).Normalized,
+                                    sourceRelative.Normalized,
                                     (int) (CollisionGroup.Impassable)),
-                                (pos.Position - _eyeManager.CurrentEye.Position.Position).Length,
+                                sourceRelative.Length,
                                 (e) => { occlusion++; return false; }, false);
                         }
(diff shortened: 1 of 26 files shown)


PJB3005:
I don't like that this is all a massive single PR but oh well.

Robust.Client/Robust.Client.csproj
```diff
@@ -26,7 +26,7 @@
     <PackageReference Include="SixLabors.ImageSharp.Drawing" Version="1.0.0-beta0006" />
     <PackageReference Condition="'$(TargetFramework)' == 'net472'" Include="System.Memory" Version="4.5.3" />
     <PackageReference Include="YamlDotNet" Version="8.1.0" />
-    <PackageReference Include="OpenTK" Version="3.1.0" />
+    <PackageReference Include="OpenTK" Version="4.0.0-pre9.1" />
```
PJB3005: Please add direct dependencies on `OpenToolkit.Graphics` and `OpenToolkit.OpenAL` instead of relying on the meta package.

Robust.Client/Robust.Client.csproj
```diff
@@ -37,6 +37,7 @@
     <ProjectReference Include="..\Robust.Shared.Scripting\Robust.Shared.Scripting.csproj" />
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\Content.Shared\Content.Shared.csproj" />
```
PJB3005: This is definitely an accident.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
```diff
@@ -171,31 +172,36 @@ public IClydeAudioSource CreateAudioSource(AudioStream stream)
             // ReSharper disable once PossibleInvalidOperationException
             // TODO: This really shouldn't be indexing based on the ClydeHandle...
             AL.Source(source, ALSourcei.Buffer, _audioSampleBuffers[(int) stream.ClydeHandle.Value.Value].BufferHandle);
-            var audioSource = new AudioSource(this, source, stream);
+
+            var filter = EFX.GenFilter();
+            EFX.Filter(filter, FilterInteger.FilterType, (int) FilterType.Lowpass);
+            AL.Source(source, ALSourcei.EfxDirectFilter, filter);
```
PJB3005: Should we really instantiate a filter here even if it might not ever be used?

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
```diff
@@ -16,25 +15,25 @@ namespace Robust.Client.Graphics.Clyde
...
 
         private readonly HashSet<string> _alcExtensions = new HashSet<string>();
         private readonly HashSet<string> _alContextExtensions = new HashSet<string>();
 
         // Used to track audio sources that were disposed in the finalizer thread,
         // so we need to properly send them off in the main thread.
-        private readonly ConcurrentQueue<int> _sourceDisposeQueue = new ConcurrentQueue<int>();
-        private readonly ConcurrentQueue<uint> _bufferedSourceDisposeQueue = new ConcurrentQueue<uint>();
-        private readonly ConcurrentQueue<uint> _bufferDisposeQueue = new ConcurrentQueue<uint>();
+        private readonly ConcurrentQueue<(int, int)> _sourceDisposeQueue = new ConcurrentQueue<(int, int)>();
```
PJB3005: These tuple members should be named for clarity.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -97,13 +99,27 @@ public override void FrameUpdate(float frameTime)
...
                     {
                         stream.Source.SetVolume(-10000000);
                     }
                     else
                     {
+                        var occlusion = 0;
+                        var sourceRelative = _eyeManager.CurrentEye.Position.Position - pos.Position;
+                        if (sourceRelative.Length > 0)
+                        {
+                            IoCManager.Resolve<IPhysicsManager>().IntersectRayWithPredicate(
```
PJB3005: Thanks to the physics rewrite you can actually get a proper list of results now I believe, so you don't need to do this hack of mutating inside the predicate anymore.

PJB3005 (changes requested):

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
Clyybber: Yeah, but this is faster I think.

Clyybber: I had it count the results before, but figured this might be faster.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
Clyybber: When would it not be used?

Robust.Client/Robust.Client.csproj
Clyybber: Also had to add `OpenToolkit.Windowing.GraphicsLibraryFramework`

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
PJB3005: Mutating inside a predicate like this is *generally* considered an anti-pattern so I would say it's better to get a proper list and check the count.

If it performs a performance issue later we can optimize it.

Robust.Client/Graphics/Clyde/Audio/Clyde.Audio.cs
PJB3005: Global sounds, sounds that aren't occluded (short sounds like keyboard sounds).

Robust.Client/Robust.Client.csproj
PJB3005: > OpenToolkit.Windowing.GraphicsLibraryFramework

You shouldn't need that since we already include it in-repo. You might have to implement the `IBindingsContext` yourself though but that's trivial.

PJB3005 (approved):

PR #1078, opened 2020-05-27
Add `Hard` property to collidables and fix airlock issues

### LINKED ISSUE #1069: Re-add Hard Collidables in some form
My plan is to have a boolean `Hard` in `CollidableComponent`. If this is true, a manifold will be generated for the collision: otherwise, only events will be triggered. It will be set to true by default.

snowsignal:
Closes #1069.

Hard collidables have been added back in with the `Hard` property. Non-hard collidables will still trigger collision events but won't cause impulses. Also, airlocks won't close on players anymore!


snowsignal:
Oh, I just realized that I included changes from a separate issue I was working on.

Robust.Shared/GameObjects/Components/Collidable/CollidableComponent.cs
```diff
@@ -122,6 +124,12 @@ public bool CanCollide
             set => _canCollide = value;
         }
 
+        public bool Hard
```
ComicIronic, line 135: Definitely needs a doc comment to explain what `Hard`-ness does, in my view.

Robust.Shared/Interfaces/Physics/IPhysicsManager.cs
```diff
@@ -149,22 +149,24 @@ public Vector2 RelativeVelocity
         public readonly Vector2 Normal;
         public readonly ICollidableComponent A;
         public readonly ICollidableComponent B;
+        public readonly bool Soft;
```
ComicIronic: Is there a reason the other variable is `Hard` and this is `Soft`? Could they not both be `Hard`, to make the connection between them clearer?

ComicIronic:
Also - are the changes to the optional `approximate` argument related to `Hard`? It seemed to me that, from the code, they didn't do anything.

snowsignal:
@ComicIronic They actually do. I realized this when I was fixing airlocks, but approximate needed to be set to false when finding collisions around it, or else it would collide with the walls. There's a separate content PR for airlocks that I forgot to make, thanks for reminding me.

PJB3005:
Could maybe fix the "airlocks collide with walls and themselves" by just changing their collision layers so they don't overlap.

Robust.Shared/Interfaces/Physics/IPhysicsManager.cs
snowsignal: You're right, I should change it to `Hard` for consistency.

snowsignal:
That could also work. I still want to expose `approximate` to the `IsColliding` and `GetCollidingEntities` API in case it needs to be used in the future.

PR #1079, opened 2020-05-27
Add distance to AudioSystem

gradientvera:
This seemed to spawn a lot of discussion on Discord, for some reason.


AJCM-git:
Does this fix https://github.com/space-wizards/space-station-14/issues/897 ?

gradientvera:
Yes, this fixes https://github.com/space-wizards/space-station-14/issues/897 for the most part

gradientvera:
requesting review from @FL-OZ since he refactored the AudioSystem and @Clyybber since he worked on audio occlusion

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -95,7 +131,8 @@ public AudioSourceServer PlayFromEntity(string filename, IEntity entity, AudioPa
         /// <param name="filename">The resource path to the OGG Vorbis file to play.</param>
         /// <param name="coordinates">The coordinates at which to play the audio.</param>
         /// <param name="audioParams"></param>
-        public AudioSourceServer PlayAtCoords(string filename, GridCoordinates coordinates, AudioParams? audioParams = null)
+        /// <param name="range">The max range at which the audio will be heard. Less than or equal to 0 to send to every player.</param>
+        public AudioSourceServer PlayAtCoords(string filename, GridCoordinates coordinates, AudioParams? audioParams = null, int range = AudioDistanceRange)
```
Clyybber, line 135: I think the range should be supplied via `audioParams.MaxDistance`

gradientvera: I disagree for these reasons:
- This would take away flexibility from coders to decide the range of the network message and the range of the sound itself: You could send a sound with `range = 0` (sent to everyone) but with `audioParams.MaxDistance` set to some other value.
- I feel like mixing the range of the network message and the max distance for the audio itself isn't a good idea. While it's true that the two should work together to ensure sounds are seamless and 'realistic', I'm unsure we should do this at the cost of good design.

My proposal for resolving this is the following:
- Keep the range parameter but set the default `AudioParams.MaxDistance` to be the same as range as `AudioDistanceRange` and set `AudioParams.MaxDistance` to `range` when `audioParams` is null. Since most of the time positional sounds will play with the default range parameter value, this will ensure sounds are seamless most of the time. If coders want to expand the range of the message, they can. And the audio MaxDistance will be the same as the range *if* the `audioParams` is not null. If the coder decides to set an `audioParam` *and* modify the message range, they should also decide whether to match the audio max distance to the message range.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -1,59 +1,70 @@
...
             private readonly uint _id;
             private readonly AudioSystem _audioSystem;
+            private readonly IEnumerable<IPlayerSession> _sessions;
 
-            internal AudioSourceServer(AudioSystem parent, uint identifier)
+            internal AudioSourceServer(AudioSystem parent, uint identifier, IEnumerable<IPlayerSession> sessions = null)
             {
                 _audioSystem = parent;
                 _id = identifier;
+                _sessions = null;
```
ComicIronic: `_sessions = sessions`?

gradientvera: Ah, fugg. I missed that one--

PR #1080, opened 2020-05-27
Adds theyre text macro

lzimann:


ComicIronic:
There are some tests for the various text macros in `MacroFormatProvider_tests.cs`. It might be a good idea to add tests for this macro as well.

Hugal31:
At some point, if it is interesting for more macros and other languages, I would like to refactor those macros in YAML, so it is easy to add one.

Brain fart of the schema:
```yaml
- name: theyre
  switch:
    on: number
    1:
      switch:
        on: gender
        male: he's
        female: she's
        neuter: it's
        default: they're
    default: they're
```

It might become a whole scripting language though.

PJB3005:
> It might become a whole scripting language though.

My understanding from reading about localization is that it would devolve into that evemtually so I would prefer them staying as C#.

RemieRichards:
why not adopt something like fluent?

PR #1081, opened 2020-05-28
Add server side sound event stopping.

ghost:
https://github.com/space-wizards/space-station-14/issues/1012#issue-625775642

This adds caching of sound event identifiers for retrieving them later and stopping them if you want.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
@@ -137,19 +137,6 @@ namespace Robust.Client.GameObjects.EntitySystems
                 }
             }
 
-            foreach (var kvp in _playingClydeStreams)
-            {
-                if (!kvp.Value.Done)
-                {
-                    continue;
-                }
-
-                var msgToServer = new StopAudioMessageServer
-                {
-                    Identifier = kvp.Key
-                };
-                RaiseNetworkEvent(msgToServer);
-            }
             _playingClydeStreams.Values.ToList().RemoveAll(p => p.Done);
         }
 
--- a/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
@@ -15,6 +15,7 @@ using Robust.Shared.Utility;
 using System.Collections.Generic;
 using JetBrains.Annotations;
 using Robust.Client.Interfaces.Graphics.ClientEye;
+using System.Linq;
 
 namespace Robust.Client.GameObjects.EntitySystems
 {
@@ -28,7 +29,7 @@ namespace Robust.Client.GameObjects.EntitySystems
         [Dependency] private readonly IEyeManager _eyeManager;
 #pragma warning restore 649
 
-        private static Dictionary<byte, PlayingStream> _playingClydeStreams = new Dictionary<byte, PlayingStream>();
+        private static List<PlayingStream> _playingClydeStreams = new List<PlayingStream>();
 
         /// <inheritdoc />
         public override void Initialize()
@@ -41,10 +42,7 @@ namespace Robust.Client.GameObjects.EntitySystems
 
         private void StopAudioMessageHandler(StopAudioMessageClient ev)
         {
-            if (_playingClydeStreams.TryGetValue(ev.Identifier, out var stream))
-            {
-                StreamDone(ev.Identifier, stream);
-            }
+            StreamDone(_playingClydeStreams.ElementAt(ev.Identifier));
         }
 
         private void PlayAudioPositionalHandler(PlayAudioPositionalMessage ev)
@@ -58,7 +56,7 @@ namespace Robust.Client.GameObjects.EntitySystems
             var stream = Play(ev.FileName, ev.Coordinates, ev.AudioParams);
             if (stream != null)
             {
-                _playingClydeStreams.TryAdd(ev.Identifier, (PlayingStream)stream);
+                _playingClydeStreams.Add((PlayingStream)stream);
             }
         }
 
@@ -67,7 +65,7 @@ namespace Robust.Client.GameObjects.EntitySystems
             var stream = Play(ev.FileName, ev.AudioParams);
             if (stream != null)
             {
-                _playingClydeStreams.TryAdd(ev.Identifier, (PlayingStream)stream);
+                _playingClydeStreams.Add((PlayingStream)stream);
             }
         }
 
@@ -82,7 +80,7 @@ namespace Robust.Client.GameObjects.EntitySystems
             var stream = Play(ev.FileName, entity, ev.AudioParams);
             if (stream != null)
             {
-                _playingClydeStreams.TryAdd(ev.Identifier, (PlayingStream)stream);
+                _playingClydeStreams.Add((PlayingStream)stream);
             }
         }
 
@@ -91,11 +89,11 @@ namespace Robust.Client.GameObjects.EntitySystems
             var currentMap = _eyeManager.CurrentMap;
 
             // Update positions of streams every frame.
-            foreach (var (key, stream) in _playingClydeStreams)
+            foreach (var stream in _playingClydeStreams)
             {
                 if (!stream.Source.IsPlaying)
                 {
-                    StreamDone(key,stream);
+                    StreamDone(stream);
                     continue;
                 }
 
@@ -108,7 +106,7 @@ namespace Robust.Client.GameObjects.EntitySystems
                 {
                     if (stream.TrackingEntity.Deleted)
                     {
-                        StreamDone(key,stream);
+                        StreamDone(stream);
                         continue;
                     }
 
@@ -135,12 +133,11 @@ namespace Robust.Client.GameObjects.EntitySystems
                 }
             }
 
-            _playingClydeStreams.Values.ToList().RemoveAll(p => p.Done);
+            _playingClydeStreams.RemoveAll(p => p.Done);
         }
 
-        private static void StreamDone(byte key, PlayingStream stream)
+        private static void StreamDone(PlayingStream stream)
         {
-            _playingClydeStreams.Remove(key);
             stream.Source.Dispose();
             stream.Done = true;
             stream.DoPlaybackDone();
--- a/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
@@ -294,8 +294,6 @@ namespace Robust.Client.GameObjects.EntitySystems
             public GridCoordinates? TrackingCoordinates;
             public bool Done;
             public float Volume;
-            public PlayingStream InnerStream;
-
             public void Stop()
             {
                 Source.StopPlaying();
--- a/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
@@ -28,7 +28,7 @@ namespace Robust.Client.GameObjects.EntitySystems
         [Dependency] private readonly IEyeManager _eyeManager;
 #pragma warning restore 649
 
-        private static readonly Dictionary<ushort, PlayingStream> _playingClydeStreams = new Dictionary<ushort, PlayingStream>();
+        private static Dictionary<byte, PlayingStream> _playingClydeStreams = new Dictionary<byte, PlayingStream>();
 
         /// <inheritdoc />
         public override void Initialize()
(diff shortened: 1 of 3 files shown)


Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -1,18 +1,126 @@
...
 using Robust.Shared.Interfaces.GameObjects;
 using Robust.Shared.Map;
+using Robust.Shared.Random;
 
 namespace Robust.Server.GameObjects.EntitySystems
 {
     public class AudioSystem : EntitySystem
     {
+
+        private static List<ushort> _streams = new List<ushort>(ushort.MaxValue);
```
gradientvera: Setting this to max ushort value seems wasteful imo. I doubt there'll be THAT many streams playing at once.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -1,18 +1,126 @@
...
+            var streamID = (ushort)_random.Next(ushort.MaxValue);
+            while (streamID == _previousRandom)
+            {
+                streamID = (ushort)_random.Next(ushort.MaxValue);
+                break;
+            }
+            _streams.Add(streamID);
+            _previousRandom = streamID;
+            return streamID;
+        }
```
gradientvera: Why not just use an int and increment that everytime a new stream plays? I **hate** how this uses random.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -1,18 +1,126 @@
...
+
         /// <summary>
         ///     Play an audio file globally, without position.
         /// </summary>
         /// <param name="filename">The resource path to the OGG Vorbis file to play.</param>
         /// <param name="audioParams"></param>
+        public ushort PlayGlobal(string filename, AudioParams? audioParams = null)
```
gradientvera: Instead of making all these return the ID, I'd return an object (that contains the ID, of course) which you can then use to stop the playback.

gradientvera (changes requested):

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ComicIronic: Also, if you're going to do this, you may as well use an array. There's no benefit to `List<T>`'s dynamic resizing if you pre-allocate all the memory you could possibly use. Though @Zumorica this is only 128KB total.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ghost: I did exactly that first. But what about when we hit the limit?

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ghost: I was going to use byte, but **someone** always tells me "just use int"

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ghost: Why?

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ComicIronic: Modulo. This is going to hit duplicate identifiers faster than modulo arithmetic ever will (thanks, birthday problem).

ghost:
how do i fix it

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -8,11 +11,122 @@ namespace Robust.Server.GameObjects.EntitySystems
...
+            {
+                Identifier = id
+            };
+
+            RaiseNetworkEvent(msg);
+        }
+
+        private byte CacheIdentifier()
+        {
+            var newId = _streams[_streamIndex] = _streamIndex++;
```
ComicIronic: Should be `%`ed, right? I think overflow throws an exception, rather than wrapping.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -8,11 +11,122 @@ namespace Robust.Server.GameObjects.EntitySystems
...
+                _audioSystem = parent;
+                Id = identifier;
+            }
+            public void Stop()
+            {
+                _audioSystem.InternalStop(Id);
+            }
+        }
+
+        private static readonly byte[] _streams = new byte[int.MaxValue];
```
ComicIronic: Is this not too big, if you only need as many indices as there are bytes?

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ghost: What do you mean?

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ghost: I was told to just use int. But I will... shrink it I guess.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -67,17 +84,23 @@ private void PlayAudioEntityHandler(PlayAudioEntityMessage ev)
...
+            {
+                _playingClydeStreams.TryAdd(ev.Identifier, (PlayingStream)stream);
+            }
         }
 
         public override void FrameUpdate(float frameTime)
         {
             // Update positions of streams every frame.
-            foreach (var stream in _playingClydeStreams)
+            foreach (var kvp in _playingClydeStreams)
```
ShadowCommander: You can change this to `foreach (var (key, stream) in _playingClydeStreams)`

ghost:
[dnm] until a client crash is fixed.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ComicIronic: The difference is that this would be `2^24` times as big as it needs to be. That's a lot of memory. Did you not notice a massive spike in memory usage when testing this? I think the result is a 4GB array.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ComicIronic: Seems like you resolved this on the Discord, but say so if you still want me to explain what I meant.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
gradientvera: Convenience

ghost:
It doesn't crash the client anymore, I think? I tested it for even longer than past the point where it would crash before. /shrug

gradientvera (approved):

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -8,11 +11,112 @@ namespace Robust.Server.GameObjects.EntitySystems
...
     public class AudioSystem : EntitySystem
     {
+        private static byte _streamIndex = byte.MaxValue - 1;
+
+        public class AudioSourceServer
+        {
+            private readonly byte _id;
+            private readonly AudioSystem _audioSystem;
+
+            public AudioSourceServer(AudioSystem parent, byte identifier)
```
PJB3005: This constructor should be made at the very least internal so that content couldn't create one of these, then call `Stop()` on them with arbitrary values.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -8,11 +11,112 @@ namespace Robust.Server.GameObjects.EntitySystems
...
+            {
+                Identifier = id
+            };
+
+            RaiseNetworkEvent(msg);
+        }
+
+        private byte CacheIdentifier()
+        {
+            if (_streamIndex >= byte.MaxValue)
```
PJB3005: 255 identifiers is not nearly enough to avoid wraparound issues. This should be at the *very* least a ushort. Preferably an `uint`.

Robust.Shared/GameObjects/EntitySystemMessages/AudioMessages.cs
```diff
@@ -11,10 +11,17 @@ namespace Robust.Shared.GameObjects
     [Serializable, NetSerializable]
     public abstract class AudioMessage : EntitySystemMessage
     {
+        public byte Identifier { get; set; }
         public string FileName { get; set; }
         public AudioParams AudioParams { get; set; }
     }
 
+    [Serializable, NetSerializable]
+    public class StopAudioMessageClient : AudioMessage
```
PJB3005: This should not inherit from `AudioMessage` since it doesn't make use of `FileName` or `AudioParams`

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -8,11 +11,112 @@ namespace Robust.Server.GameObjects.EntitySystems
 {
     public class AudioSystem : EntitySystem
     {
+        private static byte _streamIndex = byte.MaxValue - 1;
```
PJB3005: This should not be static.

PJB3005: Also why initialize this to `byte.MaxValue-1` instead of just zero?

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -114,11 +135,12 @@ public override void FrameUpdate(float frameTime)
                 }
             }
 
-            _playingClydeStreams.RemoveAll(p => p.Done);
+            _playingClydeStreams.Values.ToList().RemoveAll(p => p.Done);
```
PJB3005: ... this straight up does not work. `.ToList()` creates a copy. If you mutate that copy, it won't change the original collection.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -157,7 +179,6 @@ public IPlayingAudioStream Play(AudioStream stream, AudioParams? audioParams = n
                 Source = source,
                 Volume = audioParams?.Volume ?? 0
             };
-            _playingClydeStreams.Add(playing);
```
PJB3005, line 160: When are  these streams getting added now?

PJB3005 (changes requested):

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -62,19 +79,23 @@ private void PlayAudioEntityHandler(PlayAudioEntityMessage ev)
                 return;
             }
 
-            Play(ev.FileName, entity, ev.AudioParams);
+            var stream = Play(ev.FileName, entity, ev.AudioParams);
+            if (stream != null)
+            {
+                _playingClydeStreams.TryAdd(ev.Identifier, (PlayingStream)stream);
```
PJB3005: This can only ever fail to add to the identifier wrapped around in the server. In that case it *probably* makes sense to use the *new* stream instead so this should probably just be using the indexer.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ghost: > Also why initialize this to `byte.MaxValue-1` instead of just zero?

Hah, I completely forgot about that. @ShadowCommander and I were testing overflow and I forgot to discard that change.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
FL-OZ: What do you recommend I do instead?

Robust.Shared/GameObjects/EntitySystemMessages/AudioMessages.cs
```diff
@@ -11,15 +11,15 @@ namespace Robust.Shared.GameObjects
...
         public string FileName { get; set; }
         public AudioParams AudioParams { get; set; }
     }
 
     [Serializable, NetSerializable]
-    public class StopAudioMessageClient : AudioMessage
+    public class StopAudioMessageClient : EntitySystemMessage
     {
-
+        public int Identifier {get; set;}
```
PJB3005, line 22: This should be a `uint`.

PR #1085, opened 2020-05-28
Fixed buttons not exiting pressed visual state

### LINKED ISSUE #960: Spam clicking microwave buttons makes you press them at the same time
If you spam click the time buttons, you cal press multiple of them at the same time 
image

FarewellFire:
Fixes https://github.com/space-wizards/space-station-14/issues/960
The problem was not in spam clicking, but rather in pressing down on a button and pressing up outside of it.

### DIFF
--- a/Robust.Client/UserInterface/Controls/BaseButton.cs
+++ b/Robust.Client/UserInterface/Controls/BaseButton.cs
@@ -242,6 +242,18 @@ namespace Robust.Client.UserInterface.Controls
         {
             base.KeyBindUp(args);
 
+            var drawMode = DrawMode;
+            var hasPoint = HasPoint((args.PointerLocation.Position - GlobalPixelPosition) / UIScale);
+
+            if (!hasPoint)
+            {
+                _attemptingPress = false;
+                if (drawMode != DrawMode)
+                {
+                    DrawModeChanged();
+                }
+            }
+
             if (Disabled || (!_enableAllKeybinds && args.Function != EngineKeyFunctions.UIClick))
             {
                 return;
@@ -250,8 +262,7 @@ namespace Robust.Client.UserInterface.Controls
             var buttonEventArgs = new ButtonEventArgs(this, args);
             OnButtonUp?.Invoke(buttonEventArgs);
 
-            var drawMode = DrawMode;
-            if (Mode == ActionMode.Release && _attemptingPress && HasPoint((args.PointerLocation.Position - GlobalPixelPosition) / UIScale))
+            if (Mode == ActionMode.Release && _attemptingPress && hasPoint)
             {
                 // Can't un press a radio button directly.
                 if (Group == null || !Pressed)


Robust.Client/UserInterface/Controls/BaseButton.cs
```diff
@@ -242,6 +242,18 @@ protected internal override void KeyBindUp(GUIBoundKeyEventArgs args)
...
+            var drawMode = DrawMode;
+            var hasPoint = HasPoint((args.PointerLocation.Position - GlobalPixelPosition) / UIScale);
+
+            if (!hasPoint)
+            {
+                _attemptingPress = false;
+                if (drawMode != DrawMode)
+                {
+                    DrawModeChanged();
+                }
```
ComicIronic, line 254: Under what circumstances does this need to happen here, where it won't be called below?

FarewellFire: > Under what circumstances does this need to happen here, where it won't be called below?

It won't be called below if you press the mouse button down on a UI button, then drag the pointer out of the UI window and press it up. It'll just return in the line 257.

ComicIronic: I'm not a fan of this fix, personally. It seems like something unexpected is happening in the click code that means that the action seen here is not always a `UIClick` - and either that shouldn't be happening, or it shouldn't be checked for at line 257.

From testing, it looks like drag-drops off the button trigger the `Use` event instead, but drag drops onto another button still trigger the `UIClick`. I don't know what the actual behaviour should be in either case, maybe @PJB3005 can shed some light on it.

FarewellFire: > I'm not a fan of this fix, personally. It seems like something unexpected is happening in the click code that means that the action seen here is not always a `UIClick` - and either that shouldn't be happening, or it shouldn't be checked for at line 257.
> 
> From testing, it looks like drag-drops off the button trigger the `Use` event instead, but drag drops onto another button still trigger the `UIClick`. I don't know what the actual behaviour should be in either case, maybe @PJB3005 can shed some light on it.

But even in this case, if I understand correctly, this fix is needed to process up clicks outside of the UI window, no?

PJB3005: Yeah the bug here is that there's no `KeyUp` for `UIClick` being raised because mouse focus of the button is lost as soon as `Use` is released, so further keyups (`UIClick` is later for whatever reason..?) aren't hitting it.

What I would do here is make it so mouse focus gets released after all keybinds on a key combo have had their key combo, by making another function analogous to `UserInterfaceManager.HandleCanFocusDown` for the key up, and then firing it at the end of `InputManager.KeyUp` if any of the released keybinds were `CanFocus`.

`_controlFocused = null;` is the line to release mouse focus.

ShadowCommander:
Fixed by #1097

PR #1087, opened 2020-05-29
Adds SpriteKey value to SpriteComponent

CakeQ:
Made for https://github.com/space-wizards/space-station-14/pull/1013

While I'm aware it might seem strange to assign an engine variable for my changes specifically in the above PR I believe it's a good idea to allow for sprite components to contain keys for use with other interacting components.

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -115,7 +115,7 @@ namespace Robust.Client.GameObjects
         /// <summary>
         ///     We will allow for unique interactions with other objects with the same key.
         /// </summary>
-        [ViewVariables(VVAccess.ReadWrite)]
+        [ViewVariables(VVAccess.ReadOnly)]
         public string SpriteKey => _spriteKey;
 
         private RSI _baseRsi;
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -30,6 +30,7 @@ namespace Robust.Client.GameObjects
         IComponentDebug
     {
         private bool _visible = true;
+        private string _spriteKey;
 
         [ViewVariables(VVAccess.ReadWrite)]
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -111,6 +112,12 @@ public bool Directional
 
         private bool _directional = true;
 
+        /// <summary>
+        ///     We will allow for unique interactions with other objects with the same key.
+        /// </summary>
+        [ViewVariables(VVAccess.ReadWrite)]
+        public string SpriteKey => _spriteKey;
```
ComicIronic: Why is this marked `ReadWrite` if it's currently read-only?

ComicIronic:
What's the use case that this fits and `SmoothKey` doesn't?

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
CakeQ: Whoops

CakeQ:
> 
> 
> What's the use case that this fits and `SmoothKey` doesn't?

This allows for IconSmoothingComponents to interact with SpriteComponents

Some smoothing sprites (like walls) will want to be able to smooth with other sprites (like doors and other miscallaneous structures)

Aside from that it's also usable in other ways for identifying sprites via keys when needed

ComicIronic:
I feel like this a bit fishy. Smoothing with doors sounds like a nice feature, but for me it would suggest that doors themselves should have an `IconSmoothingComponent`, even if it has no effect on their final appearance.

My main concern would be that the current description of what `SpriteKey` does is very non-specific, and it applies to a lot of components that will never need it. And generic fields, globally present, tend to get abused in ugly ways. `SpriteKey` might end up having 10 different conditional use cases, depending on what code actually uses it.

CakeQ:
> 
> 
> I feel like this a bit fishy. Smoothing with doors sounds like a nice feature, but for me it would suggest that doors themselves should have an `IconSmoothingComponent`, even if it has no effect on their final appearance.

You're right and I'll be doing this

PR #1089, opened 2020-05-30
Command Binding System supporting multiple bindings

chairbender:
needed to do this for https://github.com/space-wizards/space-station-14/issues/200

Replaces the previous command binding system with a new approach which adds the following capabilities:

- Ability to register multiple handlers to a given key function
- Ability to specify the order in which bindings should fire when multiple classes register bindings to the same key function (basically lets you build a dependency graph)...not currently used except in the tests, but will be needed for the above mentioned ticket
- idiot-proof de-registering of bindings (one liner rather than having to ensure you de-register every handler that was registered and risk missing something)
- Bindings are now registered under an "owner Type", typically the manager / system that creates the bindings (this is what allows for the bulk unregistering of bindings and binding dependency 
- Test coverage

Would suggest starting with the CommandBindRegistry_Test to see how this new logic generally works, and check a few of the various systems that register bindings to see how it's used in an actual system (like InteractionSystem).

Some other things worth looking at:
- Note the test setup logic, I had to inject an instance of the logger in order for the tests to work because the command bind registry has some logging code. Is there a better way? IMO we should allow logging to work without requiring DI stuff to be done so this isn't needed, it might discourage people from adding useful logging messages to their code if it's making their tests fail.
- Note the `CommandBinds.Register<TOwner, TInputSystem>` - I would rather just do EntitySystem.Get<SharedEntitySystem> so the input system type doesn't need to be passed in, but that won't work because Get only works on the concrete type, and that type doesn't exist in Shared. Is this a shortcoming we should address? If we allowed getting entity systems by their base type or by an interface, it could create confusion if there are multiple entity systems running which are super / sub types of each other or sharing a common interface (which instance would you return?)


Robust.Shared/Input/Binding/CommandBindRegistry.cs
```diff
@@ -0,0 +1,228 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Robust.Shared.IoC;
+using Robust.Shared.Log;
+
+namespace Robust.Shared.Input.Binding
+{
+    /// <inheritdoc />
```
ComicIronic: You need to specify the interface this should inherit the doc from. Otherwise, it will inherit the doc of its base class i.e. `Object`.

Robust.Shared/Input/Binding/CommandBindRegistry.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// <inheritdoc />
+        public void Register<TOwner>(CommandBinds commandBinds)
+        {
+            Register(commandBinds, typeof(TOwner));
+        }
+
+        /// <inheritdoc />
+        public void Register(CommandBinds commandBinds, Type owner)
+        {
+            foreach (var existingBinding in _bindings)
```
ComicIronic: You could write this loop with `Any`, which might make it more readable. That's a style preference, though.

Robust.Shared/Input/Binding/CommandBindRegistry.cs
```diff
@@ -0,0 +1,228 @@
...
+            {
+                return handlers;
+            }
+            return Enumerable.Empty<InputCmdHandler>();
+        }
+
+        /// <inheritdoc />
+        public void Unregister(Type owner)
+        {
+            _bindings.RemoveAll(binding => binding.ForType != owner);
```
ComicIronic: `!=`?

Robust.Shared/Input/Binding/CommandBinds.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <summary>
+        /// Unregisters from the current InputSystem's BindRegistry all bindings currently registered under
+        /// indicated owner type so they will no longer receive / handle inputs. No effect if input system
+        /// no longer exists.
+        /// </summary>
+        /// <typeparam name="TOwner">owner type whose bindings should be unregistered, typically a system / manager,
+        /// should usually be typeof(this) - same type as the calling class.</typeparam>
+        public static void Unregister<TOwner>()
+        {
+            if (EntitySystem.TryGet<SharedInputSystem>(out var inputSystem))
```
ComicIronic, line 41: Was there case in testing where this safety was necessary? I can't imagine `Unregister` ever being called without an input system.

ComicIronic:
Looks like a finely crafted API. The dependency graph code might be worth abstracting out into the datatypes part of the engine, but that might not be possible, and isn't necessary for the PR in my eyes.

Robust.Client/GameObjects/EntitySystems/InputSystem.cs
```diff
@@ -87,15 +82,18 @@ public void PredictInputCommand(FullInputCmdMessage inputCmd)
         {
             var keyFunc = _inputManager.NetworkBindMap.KeyFunctionName(inputCmd.InputFunctionId);
 
-            if (!_bindMap.TryGetHandler(keyFunc, out var handler))
-                return;
+            //TODO: If there are multiple handlers how should prediction work? I think this code isn't even used
```
PJB3005: This code is used. The handlers should get re-fired in the same order as the initial keydown and with the same blocking rules (first handler can block later handlers).

chairbender:
alright, gonna make EntitySystem.Get on SharedInputSystem work and some other improvements in addition to the suggestions that I'm realizing after a decent rest. Should've had tests for Unregistering, adding those too.

Robust.Shared/Input/Binding/CommandBinds.cs
chairbender: the check existed in some of the components that were previously manually unregistering each binding, so I kept it because I wasn't completely sure it was safe to do without. Some systems are created / destroyed during the normal course of the game as far as I know so I wouldn't be surprised there could be some situation where everything is shutting down and this system temporarily no longer exists

chairbender:
thanks for the feedback, applied all the suggestions, I also added capability to EntitySystemManager to get a system type by its parent type (SharedInputSystem), and added EntitySystemManager tests for that new behavior. Also figured out I can have my test extend RobustUnitTest to avoid needing to setup IoC stuff manually.

Ready for review again @ComicIronic and @PJB3005

PJB3005 (approved):

ComicIronic:
LGTM.

PR #1091, opened 2020-05-30
Add ghosts layer between mobs and overlays

Hugal31:
Add a specific layer for the ghost.

For reference, SS13 ghosts layer is 6, vs 4 for the mob layer.


PJB3005:
This will basically be superseded by #1090 so...

PR #1092, opened 2020-05-31
Fix up midi stuff, support mild black midis

### LINKED ISSUE #1035: MIDI becomes garbled on other clients when the instrument moves outside the PVS
1. play an instrument on client A and walk away
2. instrument audio becomes extremely garbled on client B, timing is completely screwed when client A enters PVS range

We need to sanely destroy/create the renderer depending on whether the instrument is inside the client's PVS or not.

https://cdn.discordapp.com/attachments/310555209753690112/716599259272577024/midi-2020-05-31_12.28.57.webm

Tyler-IN:
Fixes https://github.com/space-wizards/space-station-14/issues/1035

Links to https://github.com/space-wizards/space-station-14/pull/1056

### DIFF
--- a/Robust.Client/Audio/Midi/MidiManager.cs
+++ b/Robust.Client/Audio/Midi/MidiManager.cs
@@ -310,11 +310,22 @@ namespace Robust.Client.Audio.Midi
                             renderer.Source.SetOcclusion(occlusion);
                         }
 
-                        if (!renderer.Source.SetPosition(pos.Position))
+                        if (renderer.Source.SetPosition(pos.Position))
                         {
-                            _midiSawmill?.Warning("Interrupting positional audio, can't set position.");
-                            renderer.Source.StopPlaying();
+                            continue;
                         }
+
+                        if (float.IsNaN(pos.Position.X) || float.IsNaN(pos.Position.Y))
+                        {
+                            // just move the source far away instead of to NaN
+                            if (renderer.Source.SetPosition(new Vector2(float.MaxValue, float.MaxValue)))
+                            {
+                                continue;
+                            }
+                        }
+
+                        _midiSawmill?.Warning("Interrupting positional audio, can't set position.");
+                        renderer.Source.StopPlaying();
                     }
                 }
         }
--- a/Robust.Client/Audio/Midi/MidiManager.cs
+++ b/Robust.Client/Audio/Midi/MidiManager.cs
@@ -135,7 +135,12 @@ namespace Robust.Client.Audio.Midi
                 _settings["synth.lock-memory"].IntValue = 0;
                 _settings["synth.threadsafe-api"].IntValue = 1;
                 _settings["synth.gain"].DoubleValue = 1.0d;
+                _settings["synth.polyphony"].IntValue = 1024;
+                _settings["synth.cpu-cores"].IntValue = 2;
+                _settings["synth.overflow.age"].DoubleValue = 3000;
                 _settings["audio.driver"].StringValue = "file";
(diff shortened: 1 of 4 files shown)


Tyler-IN:
(Almost) unrelated changes;

Removed setter from TimeScale because it can't work this way.

BlockingCollection is an extra layer over ConcurrentQueue for the sync ctx, reduced to just ConcurrentQueue. Not significant.

Timestamp -> Tick to reduce implication that Timestamp is somehow seconds or milliseconds.

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -79,7 +79,7 @@ public interface IMidiRenderer : IDisposable
         /// <summary>
         ///     Gets or sets the Time Scale of the sequencer in ticks per second. Default is 1000 for 1 tick per millisecond.
         /// </summary>
-        double SequencerTimeScale { get; set; }
+        double SequencerTimeScale { get; }
```
gradientvera: Modify the doc to reflect this change.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -310,11 +310,22 @@ public void FrameUpdate(float frameTime)
...
+                        {
+                            // just move the source far away instead of to NaN
+                            if (renderer.Source.SetPosition(new Vector2(float.MaxValue, float.MaxValue)))
+                            {
+                                continue;
+                            }
+                        }
+
+                        _midiSawmill?.Warning("Interrupting positional audio, can't set position.");
+                        renderer.Source.StopPlaying();
```
gradientvera: Even if you handle NaN cases manually, the actual issue still persists: There's no way to effectively stop or pause a renderer's audio source playback. The renderer will just keep playing the audio source for as long as it renders audio.

gradientvera (changes requested):

Robust.Client/Audio/Midi/MidiManager.cs
Tyler-IN: Yeah, that's intentional here... fades in nicely from a distance and thru walls.

Robust.Client/Audio/Midi/MidiRenderer.cs
Tyler-IN: Ok

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -79,7 +79,7 @@ public interface IMidiRenderer : IDisposable
         /// <summary>
         ///     Gets or sets the Time Scale of the sequencer in ticks per second. Default is 1000 for 1 tick per millisecond.
```
Tyler-IN:
```suggestion
        ///     Gets the Time Scale of the sequencer in ticks per second. Default is 1000 for 1 tick per millisecond.
```
@Zumorica like this?

Robust.Client/Audio/Midi/MidiRenderer.cs
Tyler-IN: See https://github.com/space-wizards/RobustToolbox/pull/1092#discussion_r432974248

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Yeah, exactly like that.

gradientvera (approved):
Renderers could still use a way to get paused, but other than that this seems okay.

PR #1096, opened 2020-06-01
Add check to solve UIClick upbind not getting sent to Controls

ShadowCommander:
Specifically button highlighting not getting cleared and updated due to `EngineKeyFunctions.UIClick` not getting triggered because `EngineKeyFunctions.Use` set `_controlFocused` to `null` before it processed the `UIClick`.

### DIFF
--- a/Robust.Client/UserInterface/UserInterfaceManager.cs
+++ b/Robust.Client/UserInterface/UserInterfaceManager.cs
@@ -300,7 +300,11 @@ namespace Robust.Client.UserInterface
                 args.PointerLocation.Position - control.GlobalPixelPosition);
 
             _doGuiInput(control, guiArgs, (c, ev) => c.KeyBindUp(ev));
-            _controlFocused = null;
+
+            if (guiArgs.Handled || args.Function == EngineKeyFunctions.UIClick)
+            {
+                _controlFocused = null;
+            }
 
             // Always mark this as handled.
             // The only case it should not be is if we do not have a control to click on,


ComicIronic:
I feel like this needs some explanation in the code, because I do not understand what it does.

ShadowCommander:
So, this is currently a visual bug, but it could make problems down the line. If you UIClick Down on a button, then drag your mouse off then UIClick Up with it off of the button it stays green because at that point `_controlFocused` is `null`.

The timeline is thus:
**Mouse down**
Send `UIClick` to `UIManager`
Set `_controlFocused` to control (button in this case) under cursor
UIManager.Send `UIClick` to button. Button sets color to green.

Send `Use` to control under cursor

**Mouse move**

**Mouse up**
Send `Use` to control under cursor
Get `_controlFocused` and pass `Use` to that control
Set `_controlFocused` to `null`

Send `UIClick` to control under cursor
`_controlFocused` is `null` so get control under cursor

^ The part where `Use` sets `_controlFocused` to `null` is the problem, so I added an if statement so that it only gets cleared when a control handles a keybind such as `Use` or the keybind is `UIClick`.

PJB3005 (changes requested):
This is not a correct fix for the problem: `_controlFocused` gets set by ALL keybinds that are `CanFocus`, but now it's only getting reset by specifically `UIClick`. Other keybinds would have the mouse focused control get stuck which would be very bad.

ComicIronic:
So how does this relate to PJB's proposed fix in https://github.com/space-wizards/RobustToolbox/pull/1085#discussion_r432154408? It's at a lower level?

I don't know much about the input system - my main reaction is just against the fact that this has a special case for a specific click interation. Is there any way to classify the kind of `KeyUp`s that should have this effect, or is it really only `UIClick`?

PJB3005:
>  Is there any way to classify the kind of KeyUps that should have this effect,

That classification is `CanFocus` and it being hardcoded to `UIClick` means all other `CanFocus` keybinds like examine will break, hard.

PR #1097, opened 2020-06-02
Fix UI Control focusing

ShadowCommander:
This fixes buttons not visually deselecting when dragging the mouse off of a button.

### DIFF
--- a/Robust.Client/UserInterface/UserInterfaceManager.cs
+++ b/Robust.Client/UserInterface/UserInterfaceManager.cs
@@ -168,7 +168,6 @@ namespace Robust.Client.UserInterface
 
         public void Dispose()
         {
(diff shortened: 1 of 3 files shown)


Robust.Client/UserInterface/UserInterfaceManager.cs
```diff
@@ -168,6 +168,7 @@ public void InitializeTesting()
 
         public void Dispose()
         {
+            _controlFocused?.Dispose();
```
PJB3005: Why is this necessary?

ShadowCommander: It's a Visual Studio suggestion, I can remove it.

PR #1102, opened 2020-06-04
Better sound occlusion

Clyybber:


Robust.Shared/Physics/PhysicsManager.cs
```diff
@@ -276,6 +276,26 @@ public void RemoveBody(IPhysBody physBody)
         public IEnumerable<RayCastResults> IntersectRay(MapId mapId, CollisionRay ray, float maxLength = 50, IEntity ignoredEnt = null, bool returnOnFirstHit = true)
             => IntersectRayWithPredicate(mapId, ray, maxLength, entity => entity == ignoredEnt, returnOnFirstHit);
 
+        /// <inheritdoc />
+        public float IntersectRayPenetration(MapId mapId, CollisionRay ray, float maxLength, IEntity ignoredEnt = null)
+        {
+            var penetration = 0f;
+
+            var sourceToDest = (List<RayCastResults>) IntersectRay(mapId, ray, maxLength, ignoredEnt, false);
```
ComicIronic: I don't think you should cast these. If you need list-y features, you should either create a list or change the return interface of `IntersectRay`.

ComicIronic: Or alternatively use a private function with return type `List` that the public function `IntersectRay` wraps.

RemieRichards: `IEnumerable<T>` actually *does* have `.Count()` available as an extension method, and `.Count()` seems to be all @Clyybber wants.

(Yes it's a method, `.Count()` not the same as the property `.Count`)

Clyybber: I also need to index elements. I simply used `.ToArray` now

RemieRichards: Ah fair enough, missed that.

Robust.Shared/Physics/PhysicsManager.cs
```diff
@@ -276,6 +276,26 @@ public void RemoveBody(IPhysBody physBody)
...
+        public float IntersectRayPenetration(MapId mapId, CollisionRay ray, float maxLength, IEntity ignoredEnt = null)
+        {
+            var penetration = 0f;
+
+            var sourceToDest = IntersectRay(mapId, ray, maxLength, ignoredEnt, false).ToArray();
+            var destToSource = IntersectRay(mapId,
+                new CollisionRay(ray.Position + ray.Direction * maxLength, -ray.Direction, ray.CollisionMask),
+                maxLength, ignoredEnt, false).ToArray();
+
+            if (sourceToDest.Length != destToSource.Length) Logger.Error("This should never occur.");
```
gradientvera: If that should never occur, why not throw an exception instead?

Clyybber: Uses `Debug.Assert` now

gradientvera (approved):

PR #1103, opened 2020-06-04
Fix lasers changing DrawingHandleWorld transform

### LINKED ISSUE #1065: DrawingHandleWorld goes skewiff when firing lasers
Whenever firing a laser anything that uses DrawingHandleWorld will move:

image

On master branch if you have the spawn entity icon up and fire a laser it will distort.

image

(It's easier to do when I full-screen on my 4k monitor)

I assume the problem is in EffectSystem somewhere as from memory it only ever happens with the content hitscan weapon.

ComicIronic:
Lasers were manipulating the global drawing transform instead of just rendering in the correct direction. They have been changed to use a rotated box.

Fixed an issue with Clyde not respecting the way Box2Rotated rotates, which was not allowing for the behaviour needed to draw lasers. This has added some overhead to rendering, but hopefully not too much. It can always be optimized later.

Fixes #1065.

## Why did I touch Clyde code?

It assumed that all rotations were about the origin. To make things worse, it was bypassing `Box2Rotated`'s coordinate system and doing manual rotations itself. Both of these things were incorrect, and they meant that lasers (as well as general offset rotated effects) couldn't be implemented.

## Why did I change `Box2Rotated`?

I don't think the old API made much sense, and nothing was using it. It assumed the `Box2` was graphed from the rotation origin. That might be faster, but it was also more confusing.

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
@@ -458,8 +458,7 @@ namespace Robust.Client.Graphics.Clyde
             AllocRenderCommand(RenderCommandType.ResetViewMatrix);
         }
 
-        private void DrawTexture(ClydeHandle texture, Vector2 a, Vector2 b, Color modulate, UIBox2? subRegion,
-            Angle angle)
+        private void DrawTexture(ClydeHandle texture, in Box2Rotated drawRegion, Color modulate, UIBox2? subRegion)
         {
             EnsureBatchState(texture, modulate, true, BatchPrimitiveType.TriangleFan, _queuedShader);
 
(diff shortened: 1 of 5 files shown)


Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
```diff
@@ -489,24 +488,10 @@ private void DrawResetViewTransform()
...
-                tl = _currentModelMatrix.Transform(new Vector2(a.X, b.Y));
-            }
-            else
-            {
-                bl = _currentModelMatrix.Transform(angle.RotateVec(a));
-                br = _currentModelMatrix.Transform(angle.RotateVec(new Vector2(b.X, a.Y)));
-                tr = _currentModelMatrix.Transform(angle.RotateVec(b));
-                tl = _currentModelMatrix.Transform(angle.RotateVec(new Vector2(a.X, b.Y)));
-            }
+            Vector2 bl = _currentModelMatrix.Transform(drawRegion.BottomLeft);
```
PJB3005: Doing these rotation operations every time now (because it's only ever taking in a `Box2Rotated`) is very costly and definitely not preferable since it's something that's actually used very little.

PJB3005 (commented):
The real bug that should probably be fixed is that changing transform shouldn't carry across from overlays in a rendering operation.

Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
ComicIronic: I'll change the method to take 4 `Vector2` arguments for the corners. The issue is that the other way was faster but wrong. I would use some kind of interface so it wasn't necessary to wrap everything in `Box2Rotated`, but I believe that would add loads of overhead.

ComicIronic: There is a possibility to optimize by making a separate method for rotated rectangles that would avoid the two extra `Vector2`s on the stack, but I think it would mean a lot of code duplication and it's not worthwhile until there's some performance metrics saying the speedup is necessary. I'm not even sure that it would be faster, since it will add a branch, while this is now branchless.

ComicIronic:
This still has the value of letting you draw rotated textures that don't all have the same transform, so I don't think it would be superceded by your suggested fix.

There's also some legacy code that seems to reset the `DrawingHandleWorld` transform after changing it to draw stuff, so I think your fix really alters the API as used. It's less of a bug and more legacy feature.

PJB3005:
> This still has the value of letting you draw rotated textures that don't all have the same transform, so I don't think it would be superceded by your suggested fix.

Oh, of course

> There's also some legacy code that seems to reset the `DrawingHandleWorld` transform after changing it to draw stuff, so I think your fix really alters the API as used. It's less of a bug and more legacy feature.

Pretty sure those are workarounds for this same issue.

That said though, this PR is fine as it is so I'll just merge it.

PR #1105, opened 2020-06-05
Placement improvements

Clyybber:
Needed for https://github.com/space-wizards/space-station-14/pull/1078


Robust.Client/Interfaces/Placement/IPlacementManager.cs
```diff
@@ -16,7 +16,7 @@ public interface IPlacementManager
         PlacementMode CurrentMode { get; set; }
         PlacementInformation CurrentPermission { get; set; }
 
-        event EventHandler PlacementCanceled;
+        event EventHandler PlacementChanged;
```
ComicIronic, line 22: I think this is worth documenting, since `Changed` is a bit generic and people might be tempted to think this fires every time the placement moves (which it doesn't, right?).

PR #1106, opened 2020-06-05
Fix TickRate causing CurTime jumps

### LINKED ISSUE #1104: CurTime jumps around if you modify server tick rate on a live server.
This is because it's calculated purely as `ticks passed / tickrate`. Changing tickrate changes total time.

This is problematic because `CurTime` is used as an absolute time measure server side. If it were to change many things like cooldowns could get stuck or act incorrectly.

To reproduce:

1. Spawn a bike horn
2. `cvar net.tickrate 10` in the server.
3. Honk the horn
4. `cvar net.tickrate 60` in the server.
5. Your bike horn cooldown is now stuck on a huge value.

ComicIronic:
Fixes #1104 .

Getting the current time, or changing the tick rate, now caches the last tick and the timestamp at that tick. This means that new ticks are only cumulative on top of older ticks already used for time calculations, and tick rate no longer retrospectively changes the effective time period of ticks from a different tick rate.

Confirmed fixed with a bike horn.

### DIFF
--- a/Robust.Client/BaseClient.cs
+++ b/Robust.Client/BaseClient.cs
@@ -195,6 +195,7 @@ namespace Robust.Client
         private void HandleSetTickRate(MsgSetTickRate message)
         {
             _timing.TickRate = message.NewTickRate;
+            Logger.InfoS("client", $"Tickrate changed to: {message.NewTickRate}");
         }
 
         private void OnLocalStatusChanged(object obj, StatusEventArgs eventArgs)
--- a/Robust.Shared/Timing/GameTiming.cs
+++ b/Robust.Shared/Timing/GameTiming.cs
@@ -47,7 +47,10 @@ namespace Robust.Shared.Timing
         // tick that the value was calculated for. The next time CurTime is
         // calculated, it should only try to monotonically increase both of
         // these values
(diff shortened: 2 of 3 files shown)


ComicIronic:
Test failures are because the default tickrate is set to the frame count, and the tick number starts at `1`, not `0`, so the first tick is cached as `1/ frame count` seconds, when the tests expect it to not be.

Fixes include: making ticks start at `0`, and just changing the test values to reflect the new behaviour.

PJB3005:
I'm not sure making 0 the first tick is a good idea. That means an uninitialized value is seen as valid meaning errors may be harder to find.

Couldn't you just initialize `_cachedCurTimeInfo` to `(TimeSpan.Zero, new GameTick(1))` instead?

ComicIronic:
Tried that, and it actually made the difference in timing from the test values worse - which makes sense if you think about it, because you're then declaring that the first tick took no time at all, rather than 1/60th of a second.

I think probably either way the test values will need to change - we just need to decide if the first tick should count for 0 or not.

PJB3005:
> we just need to decide if the first tick should count for 0 or not.

It probably shouldn't.

PR #1109, opened 2020-06-06
Enable nullable reference types on Robust.Shared and fix all warnings.

PJB3005:


Robust.Shared/Containers/ContainerHelpers.cs
```diff
@@ -34,9 +35,9 @@ public static bool IsInContainer(IEntity entity)
         /// <param name="entity">Entity that might be inside a container.</param>
         /// <param name="manager">The container manager that this entity is inside of.</param>
         /// <returns>If a container manager was found.</returns>
-        public static bool TryGetContainerMan(IEntity entity, out IContainerManager manager)
+        public static bool TryGetContainerMan(IEntity entity, [NotNullWhen(true)] out IContainerManager? manager)
```
ComicIronic, line 38: Is it not possible to just use `NotNull`?

PJB3005: No, because the value *is* null when no container manager is found.

PR #1112, opened 2020-06-07
Make AddComponent load default DataField values

### LINKED ISSUE #1082: AddComponent should run ExposeData with a dummy empty serializer to load defaults.
Right now if you just do plain `AddComponent`, `ExposeData` does not get called at all. This is a problem because for the sake of keeping the code clean we want component defaults to be only in `ExposeData`. See https://github.com/space-wizards/space-station-14/commit/d2a38bdb00d95faecce1108b216b38e4d085efce as an example of why this is necessary.

ComicIronic:
Attempts to fix #1082 .

This adds a new `ObjectSerializer` which only tries to write in the default `DataField` values for unset fields.

This **breaks the content!** `BodyManagerComponent` tries to load some data into string vars, but those vars are set to `""`, not `null`, so the default value isn't loaded (because the var is "set"). This can be fixed by changing `string templateName = "";` to `string templateName = null;` and the same for `presetName`.

### Why only overwrite unset fields?

Because the prototype loader loads the component data before it calls `AddComponent`, so if the default value serializer overwrite all fields it would end up overriding the prototype values.

### Other changes

Moved a method from `YamlObjectSerializer` to the base class `ObjectSerializer`, because it was YAML-agnostic and the new serializer benefits from sharing the implementation.

### DIFF
--- a/Robust.Shared/GameObjects/ComponentManager.cs
+++ b/Robust.Shared/GameObjects/ComponentManager.cs
@@ -129,7 +129,7 @@ namespace Robust.Shared.GameObjects
                 ComponentAdded?.Invoke(this, new ComponentEventArgs(component));
             }
 
-            // component.ExposeData(DefaultValueSerializer.Reader());
+            component.ExposeData(DefaultValueSerializer.Reader());
 
             component.OnAdd();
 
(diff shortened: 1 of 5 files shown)


ComicIronic:
This was tested by reverting https://github.com/space-wizards/space-station-14/commit/d2a38bdb00d95faecce1108b216b38e4d085efce and checking that the integration tests passed (they did).

PR #1117, opened 2020-06-08
Have RobustSerializer use a shared string dictionary

Tyler-IN:
implements shared string dictionary and handshake from net-code-2

makes handshake async

### DIFF
--- a/Robust.Server/Player/PlayerManager.cs
+++ b/Robust.Server/Player/PlayerManager.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using Prometheus;
 using Robust.Server.Interfaces;
 using Robust.Server.Interfaces.Player;
@@ -359,18 +360,25 @@ namespace Robust.Server.Player
             Dirty();
         }
 
-        private void HandleWelcomeMessageReq(MsgServerInfoReq message)
+        private async void HandleWelcomeMessageReq(MsgServerInfoReq message)
         {
-            var session = GetSessionByChannel(message.MsgChannel);
-
-            var netMsg = message.MsgChannel.CreateNetMessage<MsgServerInfo>();
+            var channel = message.MsgChannel;
+            var netMsg = channel.CreateNetMessage<MsgServerInfo>();
 
             netMsg.ServerName = _baseServer.ServerName;
             netMsg.ServerMaxPlayers = _baseServer.MaxPlayers;
             netMsg.TickRate = _timing.TickRate;
+
+            IPlayerSession session;
+            while (!TryGetSessionByChannel(channel, out session))
+            {
+                await Task.Delay(10);
+                if (!channel.IsConnected) return;
+            }
+
             netMsg.PlayerSessionId = session.SessionId;
 
-            message.MsgChannel.SendMessage(netMsg);
+            channel.SendMessage(netMsg);
         }
 
         private void HandlePlayerListReq(MsgPlayerListReq message)
--- a/Robust.Shared/ContentPack/DirLoader.cs
+++ b/Robust.Shared/ContentPack/DirLoader.cs
@@ -126,35 +126,9 @@ namespace Robust.Shared.ContentPack
                     }
                 });
             }
-
             public IEnumerable<string> GetRelativeFilePaths()
             {
-                foreach (var file in _directory.EnumerateFiles())
-                {
-                    if ((file.Attributes & FileAttributes.Hidden) != 0 || file.Name[0] == '.')
-                        continue;
-
-                    var filePath = file.FullName;
-                    var relPath = filePath.Substring(_directory.FullName.Length);
-                    yield return ResourcePath.FromRelativeSystemPath(relPath).ToRootedPath().ToString();
-                }
-
-                foreach (var subDir in _directory.EnumerateDirectories())
-                {
-                    if ((subDir.Attributes & FileAttributes.Hidden) != 0 || subDir.Name[0] == '.')
-                        continue;
-
-                    switch (subDir.Name)
-                    {
-                        case "logs":
-                            continue;
-                    }
-
-                    foreach (var relPath in GetRelativeFilePaths(subDir))
-                    {
-                        yield return relPath;
-                    }
-                }
+                return GetRelativeFilePaths(_directory);
             }
 
             private IEnumerable<string> GetRelativeFilePaths(DirectoryInfo dir)
@@ -184,8 +158,6 @@ namespace Robust.Shared.ContentPack
                     }
                 }
             }
-
-
         }
     }
 }
--- a/Robust.Shared/ContentPack/DirLoader.cs
+++ b/Robust.Shared/ContentPack/DirLoader.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Reflection;
 using System.Threading.Tasks;
 using Robust.Shared.Interfaces.Log;
 using Robust.Shared.Utility;
@@ -125,6 +126,66 @@ namespace Robust.Shared.ContentPack
                     }
                 });
             }
+
+            public IEnumerable<string> GetRelativeFilePaths()
+            {
+                foreach (var file in _directory.EnumerateFiles())
+                {
+                    if ((file.Attributes & FileAttributes.Hidden) != 0 || file.Name[0] == '.')
+                        continue;
+
+                    var filePath = file.FullName;
+                    var relPath = filePath.Substring(_directory.FullName.Length);
+                    yield return ResourcePath.FromRelativeSystemPath(relPath).ToRootedPath().ToString();
+                }
+
+                foreach (var subDir in _directory.EnumerateDirectories())
+                {
+                    if ((subDir.Attributes & FileAttributes.Hidden) != 0 || subDir.Name[0] == '.')
+                        continue;
+
+                    switch (subDir.Name)
+                    {
+                        case "logs":
+                            continue;
+                    }
+
+                    foreach (var relPath in GetRelativeFilePaths(subDir))
+                    {
+                        yield return relPath;
+                    }
+                }
+            }
+
+            private IEnumerable<string> GetRelativeFilePaths(DirectoryInfo dir)
+            {
+                foreach (var file in dir.EnumerateFiles())
+                {
+                    if ((file.Attributes & FileAttributes.Hidden) != 0 || file.Name[0] == '.')
+                    {
+                        continue;
+                    }
+
+                    var filePath = file.FullName;
+                    var relPath = filePath.Substring(_directory.FullName.Length);
+                    yield return ResourcePath.FromRelativeSystemPath(relPath).ToRootedPath().ToString();
+                }
+
+                foreach (var subDir in dir.EnumerateDirectories())
+                {
+                    if (((subDir.Attributes & FileAttributes.Hidden) != 0) || subDir.Name[0] == '.')
+                    {
+                        continue;
+                    }
+
+                    foreach (var relPath in GetRelativeFilePaths(subDir))
+                    {
+                        yield return relPath;
+                    }
+                }
+            }
+
+
         }
     }
 }
--- a/Robust.Shared/ContentPack/ResourceManager.cs
+++ b/Robust.Shared/ContentPack/ResourceManager.cs
@@ -107,12 +107,6 @@ namespace Robust.Shared.ContentPack
             {
                 _contentRootsLock.ExitWriteLock();
             }
-
-            Logger.InfoS("res", $"Mapping file path strings from {loader}");
-            if (IoCManager.Resolve<INetManager>().IsServer)
-            {
-                RobustSerializer.MappedStringSerializer.AddStrings(loader.GetRelativeFilePaths());
-            }
         }
 
         private static ResourcePath SanitizePrefix(ResourcePath? prefix)
@@ -143,12 +137,6 @@ namespace Robust.Shared.ContentPack
 
             var loader = new DirLoader(pathInfo, Logger.GetSawmill("res"));
             AddRoot(prefix, loader);
-
-            Logger.InfoS("res", $"Mapping file path strings from {path}");
-            if (IoCManager.Resolve<INetManager>().IsServer)
-            {
-                RobustSerializer.MappedStringSerializer.AddStrings(loader.GetRelativeFilePaths());
-            }
         }
 
         /// <inheritdoc />
@@ -323,12 +311,6 @@ namespace Robust.Shared.ContentPack
             finally
             {
                 _contentRootsLock.ExitWriteLock();
-            }
-
-            Logger.InfoS("res", $"Mapping file path strings from {loader}");
-            if (IoCManager.Resolve<INetManager>().IsServer)
-            {
-                RobustSerializer.MappedStringSerializer.AddStrings(loader.GetRelativeFilePaths());
             }
         }
 
--- a/Robust.Shared/ContentPack/ResourceManager.cs
+++ b/Robust.Shared/ContentPack/ResourceManager.cs
@@ -108,7 +108,7 @@ namespace Robust.Shared.ContentPack
                 _contentRootsLock.ExitWriteLock();
             }
 
-            Logger.InfoS("res", $"Mapping file path strings from {pack}");
+            Logger.InfoS("res", $"Mapping file path strings from {loader}");
             if (IoCManager.Resolve<INetManager>().IsServer)
             {
                 RobustSerializer.MappedStringSerializer.AddStrings(loader.GetRelativeFilePaths());
--- a/Robust.Shared/ContentPack/ResourceManager.cs
+++ b/Robust.Shared/ContentPack/ResourceManager.cs
@@ -5,9 +5,11 @@ using System.IO;
 using System.Text.RegularExpressions;
 using System.Threading;
 using Robust.Shared.Interfaces.Configuration;
+using Robust.Shared.Interfaces.Network;
 using Robust.Shared.Interfaces.Resources;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
+using Robust.Shared.Serialization;
 using Robust.Shared.Utility;
 
 namespace Robust.Shared.ContentPack
@@ -105,6 +107,12 @@ namespace Robust.Shared.ContentPack
             {
                 _contentRootsLock.ExitWriteLock();
             }
+
+            Logger.InfoS("res", $"Mapping file path strings from {pack}");
+            if (IoCManager.Resolve<INetManager>().IsServer)
+            {
+                RobustSerializer.MappedStringSerializer.AddStrings(loader.GetRelativeFilePaths());
+            }
         }
 
         private static ResourcePath SanitizePrefix(ResourcePath? prefix)
@@ -135,6 +143,12 @@ namespace Robust.Shared.ContentPack
 
             var loader = new DirLoader(pathInfo, Logger.GetSawmill("res"));
             AddRoot(prefix, loader);
+
+            Logger.InfoS("res", $"Mapping file path strings from {path}");
+            if (IoCManager.Resolve<INetManager>().IsServer)
+            {
+                RobustSerializer.MappedStringSerializer.AddStrings(loader.GetRelativeFilePaths());
+            }
         }
 
         /// <inheritdoc />
@@ -309,6 +323,12 @@ namespace Robust.Shared.ContentPack
             finally
             {
                 _contentRootsLock.ExitWriteLock();
+            }
+
+            Logger.InfoS("res", $"Mapping file path strings from {loader}");
+            if (IoCManager.Resolve<INetManager>().IsServer)
+            {
+                RobustSerializer.MappedStringSerializer.AddStrings(loader.GetRelativeFilePaths());
             }
         }
 
--- a/Robust.Shared/Interfaces/GameObjects/IComponentFactory.cs
+++ b/Robust.Shared/Interfaces/GameObjects/IComponentFactory.cs
@@ -105,7 +105,7 @@ namespace Robust.Shared.Interfaces.GameObjects
         IComponent GetComponent(string componentName);
 
         /// <summary>
-        /// Gets a new component instantiated of the specified <see cref="IComponent.Name"/>.
+        /// Gets a new component instantiated of the specified network ID.
         /// </summary>
         /// <param name="netId">net id of component to make</param>
         /// <returns>A Component</returns>
--- a/Robust.Shared/Interfaces/GameObjects/IComponentFactory.cs
+++ b/Robust.Shared/Interfaces/GameObjects/IComponentFactory.cs
@@ -105,6 +105,13 @@ namespace Robust.Shared.Interfaces.GameObjects
         IComponent GetComponent(string componentName);
 
         /// <summary>
+        /// Gets a new component instantiated of the specified <see cref="IComponent.Name"/>.
+        /// </summary>
+        /// <param name="netId">net id of component to make</param>
+        /// <returns>A Component</returns>
+        IComponent GetComponent(uint netId);
+
+        /// <summary>
         /// Gets the registration belonging to a component.
         /// </summary>
         /// <param name="componentName">The name of the component.</param>
--- a/Robust.Shared/Network/NetManager.cs
+++ b/Robust.Shared/Network/NetManager.cs
@@ -14,13 +14,14 @@ using Prometheus;
 using Robust.Shared.Configuration;
 using Robust.Shared.Interfaces.Configuration;
 using Robust.Shared.Interfaces.Network;
+using Robust.Shared.Interfaces.Serialization;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
 using Robust.Shared.Utility;
 
 namespace Robust.Shared.Network
 {
-    /// <summary>
+   /// <summary>
     ///     Callback for registered NetMessages.
     /// </summary>
     /// <param name="message">The message received.</param>
@@ -37,6 +38,9 @@ namespace Robust.Shared.Network
     /// </summary>
     public partial class NetManager : IClientNetManager, IServerNetManager, IDisposable
     {
+
+        [Dependency] private readonly IRobustSerializer _serializer = default!;
+
         private static readonly Counter SentPacketsMetrics = Metrics.CreateCounter(
             "robust_net_sent_packets",
             "Number of packets sent since server startup.");
@@ -226,7 +230,16 @@ namespace Robust.Shared.Network
             _config.RegisterCVar("net.fakelagrand", 0.0f, CVar.CHEAT, _fakeLagRandomChanged);
 #endif
 
-            _strings.Initialize(this, () => { OnConnected(ServerChannel!); });
+            //_strings.Initialize(this, () => { OnConnected(ServerChannel!); });
+            _strings.Initialize(this, () =>
+            {
+                Logger.InfoS("net","Message string table loaded.");
+            });
+            _serializer.ClientHandshakeComplete += () =>
+            {
+                Logger.InfoS("net","Client completed serializer handshake.");
+                OnConnected(ServerChannel!);
+            };
 
             _initialized = true;
         }
@@ -621,7 +634,7 @@ namespace Robust.Shared.Network
             HandleInitialHandshakeComplete(connection);
         }
 
-        private void HandleInitialHandshakeComplete(NetConnection sender)
+        private async void HandleInitialHandshakeComplete(NetConnection sender)
         {
             var session = _assignedSessions[sender];
 
@@ -629,6 +642,8 @@ namespace Robust.Shared.Network
             _channels.Add(sender, channel);
 
             _strings.SendFullTable(channel);
+
+            await _serializer.Handshake(channel);
 
             Logger.InfoS("net", $"{channel.RemoteEndPoint}: Connected");
 
@@ -666,13 +681,22 @@ namespace Robust.Shared.Network
         {
             var peer = msg.SenderConnection.Peer;
             if (peer.Status == NetPeerStatus.ShutdownRequested)
+            {
+                Logger.WarningS("net", $"{msg.SenderConnection.RemoteEndPoint}: Received data message, but shutdown is requested.");
                 return true;
+            }
 
             if (peer.Status == NetPeerStatus.NotRunning)
+            {
+                Logger.WarningS("net", $"{msg.SenderConnection.RemoteEndPoint}: Received data message, peer is not running.");
                 return true;
+            }
 
             if (!IsConnected)
+            {
+                Logger.WarningS("net", $"{msg.SenderConnection.RemoteEndPoint}: Received data message, but not connected.");
                 return true;
+            }
 
             if (_awaitingData.TryGetValue(msg.SenderConnection, out var info))
             {
@@ -680,6 +704,7 @@ namespace Robust.Shared.Network
                 _awaitingData.Remove(msg.SenderConnection);
                 cancel.Dispose();
                 tcs.TrySetResult(msg);
+                Logger.WarningS("net", $"{msg.SenderConnection.RemoteEndPoint}: Received data message - had awaiting data.");
                 return false;
             }
 
--- a/Robust.Shared/Network/NetManager.cs
+++ b/Robust.Shared/Network/NetManager.cs
@@ -704,7 +704,6 @@ namespace Robust.Shared.Network
                 _awaitingData.Remove(msg.SenderConnection);
                 cancel.Dispose();
                 tcs.TrySetResult(msg);
-                Logger.WarningS("net", $"{msg.SenderConnection.RemoteEndPoint}: Received data message - had awaiting data.");
                 return false;
             }
 
--- a/Robust.Shared/Network/NetManager.cs
+++ b/Robust.Shared/Network/NetManager.cs
@@ -230,7 +230,6 @@ namespace Robust.Shared.Network
             _config.RegisterCVar("net.fakelagrand", 0.0f, CVar.CHEAT, _fakeLagRandomChanged);
 #endif
 
-            //_strings.Initialize(this, () => { OnConnected(ServerChannel!); });
             _strings.Initialize(this, () =>
             {
                 Logger.InfoS("net","Message string table loaded.");
--- a/Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgServerHandshake.cs
+++ b/Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgServerHandshake.cs
@@ -13,6 +13,12 @@ namespace Robust.Shared.Serialization
         public partial class MappedStringSerializer
         {
 
+            /// <summary>
+            /// The server part of the string-exchange handshake. Sent as the
+            /// first message in the handshake. Tells the client the hash of
+            /// the current string mapping, so the client can check if it has
+            /// a local copy.
+            /// </summary>
             [UsedImplicitly]
             private class MsgServerHandshake : NetMessage
             {
@@ -22,6 +28,9 @@ namespace Robust.Shared.Serialization
                 {
                 }
 
+                /// <value>
+                /// The hash of the current string mapping held by the server.
+                /// </value>
                 public byte[]? Hash { get; set; }
 
                 public override void ReadFromBuffer(NetIncomingMessage buffer)
--- a/Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgServerHandshake.cs
+++ b/Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgServerHandshake.cs
@@ -1,0 +1,55 @@
+using System;
+using JetBrains.Annotations;
+using Lidgren.Network;
+using Robust.Shared.Interfaces.Network;
+using Robust.Shared.Network;
+
+namespace Robust.Shared.Serialization
+{
+
+    public partial class RobustSerializer
+    {
+
+        public partial class MappedStringSerializer
+        {
+
+            [UsedImplicitly]
+            private class MsgServerHandshake : NetMessage
+            {
+
+                public MsgServerHandshake(INetChannel ch)
+                    : base($"{nameof(RobustSerializer)}.{nameof(MappedStringSerializer)}.{nameof(MsgServerHandshake)}", MsgGroups.Core)
+                {
+                }
+
+                public byte[]? Hash { get; set; }
+
+                public override void ReadFromBuffer(NetIncomingMessage buffer)
+                {
+                    var len = buffer.ReadVariableInt32();
+                    if (len > 64)
+                    {
+                        throw new InvalidOperationException("Hash too long.");
+                    }
+
+                    Hash = buffer.ReadBytes(len);
+                }
+
+                public override void WriteToBuffer(NetOutgoingMessage buffer)
+                {
+                    if (Hash == null)
+                    {
+                        throw new InvalidOperationException("Package has not been specified.");
+                    }
+
+                    buffer.WriteVariableInt32(Hash.Length);
+                    buffer.Write(Hash);
+                }
+
+            }
+
+        }
+
+    }
+
+}
--- a/Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgServerHandshake.cs
+++ b/Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgServerHandshake.cs
@@ -19,6 +19,7 @@ namespace Robust.Shared.Serialization
             /// the current string mapping, so the client can check if it has
             /// a local copy.
             /// </summary>
+            /// <seealso cref="NetworkInitialize"/>
             [UsedImplicitly]
             private class MsgServerHandshake : NetMessage
             {
--- a/Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgStrings.cs
+++ b/Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgStrings.cs
@@ -1,0 +1,63 @@
+using System;
+using System.IO;
+using JetBrains.Annotations;
+using Lidgren.Network;
+using Robust.Shared.Interfaces.Network;
+using Robust.Shared.Network;
+
+namespace Robust.Shared.Serialization
+{
+
+    public partial class RobustSerializer
+    {
+
+        public partial class MappedStringSerializer
+        {
+
+            [UsedImplicitly]
+            private class MsgStrings : NetMessage
+            {
+
+                public MsgStrings(INetChannel ch)
+                    : base($"{nameof(RobustSerializer)}.{nameof(MappedStringSerializer)}.{nameof(MsgStrings)}", MsgGroups.Core)
+                {
+                }
+
+                public byte[]? Package { get; set; }
+
+                public override void ReadFromBuffer(NetIncomingMessage buffer)
+                {
+                    var l = buffer.ReadVariableInt32();
+                    var success = buffer.ReadBytes(l, out var buf);
+                    if (!success)
+                    {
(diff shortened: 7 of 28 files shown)


Robust.Shared/ContentPack/DirLoader.cs
```diff
@@ -125,6 +126,66 @@ private void CheckPathCasing(ResourcePath path)
                     }
                 });
             }
+
+            public IEnumerable<string> GetRelativeFilePaths()
```
ComicIronic, line 129: Is this not just `GetRelativeFilePaths(_directory)`? Except for the `logs` exclusion which should probably not be hardcoded.

Tyler-IN: It does recurse into GetRelativeFilePaths(...).

The hard-coded exclusion for logs is intended to provide a mechanism to avoid content being stored in / retrieved from logs, and as a benefit also means log file paths are not added to the shared string dictionary.

In the future it'd probably be a good idea to add a blacklist of sorts to avoid the loader being able to access any other non-content paths.

Robust.Shared/Interfaces/GameObjects/IComponentFactory.cs
```diff
@@ -104,6 +104,13 @@ public interface IComponentFactory
         /// <returns>A Component</returns>
         IComponent GetComponent(string componentName);
 
+        /// <summary>
+        /// Gets a new component instantiated of the specified <see cref="IComponent.Name"/>.
```
ComicIronic: Doc comment needs updating.

Robust.Shared/ContentPack/DirLoader.cs
Tyler-IN: (So it could do with a TODO: make formal non-content blacklist for DirLoader)

Robust.Shared/Interfaces/GameObjects/IComponentFactory.cs
Tyler-IN:
```suggestion
        /// Gets a new component instantiated of the specified network ID.
```

Robust.Shared/ContentPack/DirLoader.cs
Tyler-IN: See https://github.com/space-wizards/RobustToolbox/pull/1117#discussion_r438262689

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,688 @@
...
+            private static byte[] WriteStringPackage()
+            {
+                using var ms = new MemoryStream();
+                WriteStringPackage(ms);
+                return ms.ToArray();
+            }
+
+            private static void WriteStringPackage(Stream stream)
+            {
+                var buf = new byte[65536];
```
ComicIronic: I would make this value a constant and document it as a limitation somewhere.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,688 @@
...
+                var hash = CalculateHash(MappedStringsPackage);
+
+                _sawmill.Info($"Hashing {MappedStrings.Count} strings took {sw.ElapsedMilliseconds}ms.");
+                _sawmill.Info($"Size: {MappedStringsPackage.Length} bytes, Hash: {ConvertToBase64Url(hash)}");
+                return hash;
+            }
+
+            private static byte[] CalculateHash(byte[] data)
+            {
+                if (data is null)
```
ComicIronic, line 1056: Is this necessary with null checks?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,688 @@
...
+                    foreach (var str in MappedStrings)
+                    {
+                        if (str.Length >= 65536)
+                        {
+                            throw new NotImplementedException("Overly long string in strings package.");
+                        }
+
+                        var l = enc.GetBytes(str, buf, true);
+
+                        if (l >= 65536)
```
Tyler-IN:
```suggestion
            /// <summary>
            ///     Strings longer than this will throw an exception and a better strategy will need to be employed to deal with large strings.
            /// <summary>
            public static int StringPackageMaximumBufferSize = 65536;

            public static void WriteStringPackage(Stream stream)
            {
                var buf = new byte[StringPackageMaximumBufferSize];
                var sw = Stopwatch.StartNew();
                var enc = Encoding.UTF8.GetEncoder();

                using (var zs = new DeflateStream(stream, CompressionLevel.Optimal, true))
                {
                    var bytesWritten = WriteCompressedUnsignedInt(zs, (uint) MappedStrings.Count);
                    foreach (var str in MappedStrings)
                    {
                        if (str.Length >= StringPackageMaximumBufferSize)
                        {
                            throw new NotImplementedException("Overly long string in strings package.");
                        }

                        var l = enc.GetBytes(str, buf, true);

                        if (l >= StringPackageMaximumBufferSize)
```

@ComicIronic does this fit?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,688 @@
...
+            {
+                if (value == null)
+                {
+                    WriteCompressedUnsignedInt(stream, 0);
+                    return;
+                }
+
+                if (_StringMapping.TryGetValue(value, out var mapping))
+                {
+                    WriteCompressedUnsignedInt(stream, (uint) mapping + 2);
```
ComicIronic: Why `+ 2`?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,688 @@
...
+                }
+
+                var mapIndex = ReadCompressedUnsignedInt(stream);
+                if (mapIndex == 0)
+                {
+                    value = null;
+                    return;
+                }
+
+                if (mapIndex == 1)
```
ComicIronic: These magic numbers could use being constants.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,688 @@
...
+                if (_StringMapping.TryGetValue(value, out var mapping))
+                {
+                    WriteCompressedUnsignedInt(stream, (uint) mapping + 2);
+                    return;
+                }
+
+                // indicate not mapped
+                WriteCompressedUnsignedInt(stream, 1);
+                var buf = Encoding.UTF8.GetBytes(value);
+                WriteCompressedUnsignedInt(stream, (uint) buf.Length + 1);
```
ComicIronic: This looks like it's off-by-one on each end. Why?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Had scenarios where due to bad code or network errors;
`var checkHash = CalculateHash(msg.Package!);` and `var hash = CalculateHash(MappedStringsPackage);` have passed null.
If JIT compiler determines a non-null contract is satisfied, it'll remove this check when optimizing.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,688 @@
...
+                    var buf = new byte[length - 1];
+                    stream.Read(buf);
+                    value = Encoding.UTF8.GetString(buf);
+                    return;
+                }
+
+                value = _MappedStrings[(int) mapIndex - 2];
+            }
+
+            public static int WriteCompressedUnsignedInt(Stream stream, uint value)
```
ComicIronic, line 1230: Could use documentation of how this works.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Given compressed ints;
[ 0 ] => null
[ 1, 0 ] => empty string
[ 1, n>0, string ] => uncompressed string
[ n > 1 ] => mapped string

mapping starts at index 0, so first index > 1 is 2.

Robust.Shared/Serialization/RobustSerializer.cs
```diff
@@ -6,95 +6,133 @@
...
+            {
+                MappedStringSerializer.LockMappedStrings = true;
+            }
+            else
+            {
+                var defaultAssemblies = AssemblyLoadContext.Default.Assemblies;
+                var gameAssemblies = _reflectionManager.Assemblies;
+                var robustShared = defaultAssemblies.First(a => a.GetName().Name == "Robust.Shared");
+                MappedStringSerializer.AddStrings(robustShared);
+                var contentShared = gameAssemblies.FirstOrDefault(a => a.GetName().Name == "Content.Shared");
```
ComicIronic: This is not good. Could `IReflectionManager` not somehow handle the assemblies to cache strings from?

Robust.Shared/Serialization/RobustSerializer.cs
```diff
@@ -6,95 +6,133 @@
...
+            var settings = new Settings
+            {
+                CustomTypeSerializers = new ITypeSerializer[] {mappedStringSerializer}
+            };
+            _serializer = new Serializer(types, settings);
+            _serializableTypes = new HashSet<Type>(_serializer.GetTypeMap().Keys);
+
+            if (_netManager.IsClient)
+            {
+                MappedStringSerializer.LockMappedStrings = true;
```
ComicIronic, line 75: Why is this locked here? Won't this be before the strings are sent?

Robust.Shared/ContentPack/DirLoader.cs
ComicIronic: It *does* recurse, but other than the `logs` exclusion it is pretty much exactly the same. Could they not be combined in some way to cut the code duplication?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
ComicIronic: Yeah that looks good.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,688 @@
...
+                    // not mapped
+                    var length = ReadCompressedUnsignedInt(stream);
+                    var buf = new byte[length - 1];
+                    stream.Read(buf);
+                    value = Encoding.UTF8.GetString(buf);
+                    return;
+                }
+
+                value = _MappedStrings[(int) mapIndex - 2];
+            }
```
Tyler-IN:
```suggestion
            private const int MappedNull = 1;
            private const int UnmappedString = 1;
            private const int FirstMappedIndexStart = 2;


            public static void WriteMappedString(Stream stream, string? value)
            {
                if (value == null)
                {
                    WriteCompressedUnsignedInt(stream, MappedNull);
                    return;
                }

                if (_StringMapping.TryGetValue(value, out var mapping))
                {
                    WriteCompressedUnsignedInt(stream, (uint) mapping + FirstMappedIndexStart);
                    return;
                }

                // indicate not mapped
                WriteCompressedUnsignedInt(stream, UnmappedString);
                var buf = Encoding.UTF8.GetBytes(value);
                WriteCompressedUnsignedInt(stream, (uint) buf.Length + 1);
                stream.Write(buf);
            }

            public static void ReadMappedString(Stream stream, out string? value)
            {
                if (!LockMappedStrings)
                {
                    throw new InvalidOperationException("Not performing unlocked string mapping.");
                }

                var mapIndex = ReadCompressedUnsignedInt(stream);
                if (mapIndex == MappedNull)
                {
                    value = null;
                    return;
                }

                if (mapIndex == UnmappedString)
                {
                    // not mapped
                    var length = ReadCompressedUnsignedInt(stream);
                    var buf = new byte[length - 1];
                    stream.Read(buf);
                    value = Encoding.UTF8.GetString(buf);
                    return;
                }

                value = _MappedStrings[(int) mapIndex - FirstMappedIndexStart];
            }
```

ComicIronic:
General comment: looks good, but the whole protocol needs lots of documentation to be maintainable.

Compression itself would also benefit from some documentation, since bit arithmetic is awful to try to reverse-engineer.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: @ComicIronic ^ better?

ComicIronic: Yeah that's immediately much more grokkable. :+1:

ComicIronic:
Another thing that would be great is some tests for the various handshake components, some tests for mapped/unmapped/null strings, and some tests for how strings get into the mapping.

What's the goal for sourcing mapped strings? Are they just aggressively loaded from all files that show up in the search? Will you be able to add to the mapping at runtime?

Robust.Shared/ContentPack/DirLoader.cs
Tyler-IN: I'd rather move the exclusion mechanism into the other and have `GetRelativeFilePaths()` call `GetRelativeFilePaths("/" or "")`, but we'd need to introduce a formal exclusion mechanism first, not just a quick hard-coded exclusion for logs, and that's out of the scope of this PR.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Probably + 1 / - 1 can be remved, TBH I'm not sure why I did that - maybe in earlier testing I had other representations of null/empty/mapped/unmapped.

I don't feel comfortable removing this after having confirmed everything is working correctly as is though - can be done in a future PR.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Yeah, but I'm not 100% sure how it works. It looks like [VLQ-ZZ](https://en.wikipedia.org/wiki/Variable-length_quantity#Zigzag_encoding) but I dunno.
Other WriteCompressedUnsignedInts are all private or work on arrays instead of Stream.

Robust.Shared/Serialization/RobustSerializer.cs
Tyler-IN: Prevents strings from being added from subsequently loaded content. After the dictionary is finalized / transmitted, it's not valid to add more strings / re-hash / re-transmit, etc..

Robust.Shared/Serialization/RobustSerializer.cs
Tyler-IN: I don't remember for sure, but I think the IReflectionManager didn't contain (or didn't return) the assembly it was defined in?

Tyler-IN:
The current integration tests _I think_ may already do the extended handshake.

They're loaded from paths and symbols - they could be loaded from Yaml too I think.

The strategy should be to load any string that is valid as shared content and not contain any protected information like passwords or any generated strings.

There's room to expand this to have a tool to generate / curate the shared string dictionary offline.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
ComicIronic: I would add that as a comment then, because otherwise it looks like a necessary part of the protocol.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgStrings.cs
```diff
@@ -0,0 +1,63 @@
...
+            [UsedImplicitly]
+            private class MsgStrings : NetMessage
+            {
+
+                public MsgStrings(INetChannel ch)
+                    : base($"{nameof(RobustSerializer)}.{nameof(MappedStringSerializer)}.{nameof(MsgStrings)}", MsgGroups.Core)
+                {
+                }
+
+                public byte[]? Package { get; set; }
```
ComicIronic, line 35: I don't know if you're limited by `NetMessage` - but is there any chance of using a smart constructor so these aren't nullable? (Similar question for the other nullable array message).

Tyler-IN: The NetMessage must already be created before the instance methods for reading/writing can be invoked;
`ReadFromBuffer(NetIncomingMessage buffer)` `WriteToBuffer(NetOutgoingMessage buffer)`

`CacheNetMsgFunction` also expects a constructor that takes INetChannel, meaning even if a convenience constructor were defined to set these to definitely not-null, that constructor would have to set them to some not-null state artificially (null being preferred over byte[0]).

Robust.Shared/ContentPack/ResourceManager.cs
```diff
@@ -135,6 +143,12 @@ public void MountContentDirectory(string path, ResourcePath? prefix = null)
 
             var loader = new DirLoader(pathInfo, Logger.GetSawmill("res"));
             AddRoot(prefix, loader);
+
+            Logger.InfoS("res", $"Mapping file path strings from {path}");
+            if (IoCManager.Resolve<INetManager>().IsServer)
+            {
+                RobustSerializer.MappedStringSerializer.AddStrings(loader.GetRelativeFilePaths());
+            }
```
Tyler-IN:
```suggestion
```

Robust.Shared/ContentPack/ResourceManager.cs
```diff
@@ -310,6 +324,12 @@ public void MountStreamAt(MemoryStream stream, ResourcePath path)
             {
                 _contentRootsLock.ExitWriteLock();
             }
+
+            Logger.InfoS("res", $"Mapping file path strings from {loader}");
+            if (IoCManager.Resolve<INetManager>().IsServer)
+            {
+                RobustSerializer.MappedStringSerializer.AddStrings(loader.GetRelativeFilePaths());
+            }
```
Tyler-IN:
```suggestion
```

Robust.Shared/ContentPack/ResourceManager.cs
```diff
@@ -105,6 +107,12 @@ private void AddRoot(ResourcePath prefix, IContentRoot loader)
...
```
Tyler-IN:
```suggestion
```

Robust.Shared/ContentPack/DirLoader.cs
```diff
@@ -125,6 +126,66 @@ private void CheckPathCasing(ResourcePath path)
...
+                    }
+
+                    foreach (var relPath in GetRelativeFilePaths(subDir))
+                    {
+                        yield return relPath;
+                    }
+                }
+            }
+
+
```
Tyler-IN:
```suggestion
            public IEnumerable<string> GetRelativeFilePaths()
            {
                return GetRelativeFilePaths(_directory);
            }

            private IEnumerable<string> GetRelativeFilePaths(DirectoryInfo dir)
            {
                foreach (var file in dir.EnumerateFiles())
                {
                    if ((file.Attributes & FileAttributes.Hidden) != 0 || file.Name[0] == '.')
                    {
                        continue;
                    }

                    var filePath = file.FullName;
                    var relPath = filePath.Substring(_directory.FullName.Length);
                    yield return ResourcePath.FromRelativeSystemPath(relPath).ToRootedPath().ToString();
                }

                foreach (var subDir in dir.EnumerateDirectories())
                {
                    if (((subDir.Attributes & FileAttributes.Hidden) != 0) || subDir.Name[0] == '.')
                    {
                        continue;
                    }

                    foreach (var relPath in GetRelativeFilePaths(subDir))
                    {
                        yield return relPath;
                    }
                }
            }
```

Tyler-IN:
> General comment: looks good, but the whole protocol needs lots of documentation to be maintainable.

Yeah it needs documentation, but I'm not adding that to the scope of this PR.

> Compression itself would also benefit from some documentation, since bit arithmetic is awful to try to reverse-engineer.

The shared string dictionary mechanism does act like compression, but it's just mapping strings to ints. There's not a significant amount of 'bit arithmetic' here though - not even math.

As far as compressed ints, it's borrowed magic. I know how it works, but I'm not gonna be able to document it in any meaningful way.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,698 @@
...
+                if (_StringMapping.TryGetValue(value, out var mapping))
+                {
+                    WriteCompressedUnsignedInt(stream, (uint) mapping + FirstMappedIndexStart);
+                    return;
+                }
+
+                // indicate not mapped
+                WriteCompressedUnsignedInt(stream, UnmappedString);
+                var buf = Encoding.UTF8.GetBytes(value);
+                WriteCompressedUnsignedInt(stream, (uint) buf.Length + 1);
```
Tyler-IN:
```suggestion
                // TODO: is +1 here and -1 in decode actually needed?
                WriteCompressedUnsignedInt(stream, (uint) buf.Length + 1);
```

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,698 @@
...
+                {
+                    value = null;
+                    return;
+                }
+
+                if (mapIndex == UnmappedString)
+                {
+                    // not mapped
+                    var length = ReadCompressedUnsignedInt(stream);
+                    var buf = new byte[length - 1];
```
Tyler-IN:
```suggestion
                    // TODO: is -1 here and +1 in encode actually needed?
                    var buf = new byte[length - 1];
```

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Added as TODOs.

Robust.Shared/Serialization/RobustSerializer.cs
Tyler-IN:
```suggestion
                // TODO: Need to add a GetSharedAssemblies method to the reflection manager
                var contentShared = gameAssemblies.FirstOrDefault(a => a.GetName().Name == "Content.Shared");
```

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,877 @@
...
+namespace Robust.Shared.Serialization
+{
+
+    public partial class RobustSerializer
+    {
+
+        public partial class MappedStringSerializer : IStaticTypeSerializer
+        {
+
+            private static readonly INetManager NetManager = IoCManager.Resolve<INetManager>();
```
ComicIronic: Reading through the code again, I spotted this - this is a brittle call, since it depends on the static initializer for this class running after IoC is initialized. @PJB3005 has made the point in the past that we shouldn't have that in the code.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,877 @@
...
+
+            private static readonly IDictionary<string, int> _StringMapping = new Dictionary<string, int>();
+
+            public static IReadOnlyList<String> MappedStrings => new ReadOnlyCollection<string>(_MappedStrings);
+
+            public static bool LockMappedStrings { get; set; }
+
+            private static readonly Regex RxSymbolSplitter
+                = new Regex(
+                    @"(?<=[^\s\W])(?=[A-Z])|(?<=[^0-9\s\W])(?=[0-9])|(?<=[A-Za-z0-9])(?=_)|(?=[.\\\/,#$?!@|&*()^`""'`~[\]{}:;\-])|(?<=[.\\\/,#$?!@|&*()^`""'`~[\]{}:;\-])",
```
ComicIronic: This definitely needs an explanation.

Robust.Shared/Network/NetManager.cs
```diff
@@ -226,7 +230,16 @@ public void Initialize(bool isServer)
             _config.RegisterCVar("net.fakelagrand", 0.0f, CVar.CHEAT, _fakeLagRandomChanged);
 #endif
 
-            _strings.Initialize(this, () => { OnConnected(ServerChannel!); });
+            //_strings.Initialize(this, () => { OnConnected(ServerChannel!); });
```
ComicIronic: Should be deleted?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
ComicIronic: Also, does this invalidate the `INetManager` argument to `NetworkInitialize`?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,877 @@
...
+                    throw new InvalidOperationException("Mapped strings are locked, will not clear.");
+                }
+
+                _MappedStrings.Clear();
+                _StringMapping.Clear();
+                _stringMapHash = null;
+            }
+
+            [MethodImpl(MethodImplOptions.Synchronized)]
+            public static void AddStrings(IEnumerable<string> strings)
```
ComicIronic: Should this not checked `LockMappedStrings` for consistency?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,877 @@
...
+
+            private const int MaxMappedStringSize = 420;
+
+            private const int MappedNull = 0;
+
+            private const int UnmappedString = 1;
+
+            private const int FirstMappedIndexStart = 2;
+
+            public static void WriteMappedString(Stream stream, string? value)
```
ComicIronic, line 1146: This should check for a mapping lock as well, right?

ComicIronic:
The documentation is now good enough to make this maintainable, in my eye (especially with the handshake diagrams included).

I might starting writing some tests for the string mapping later, but I don't think the PR should be blocked waiting for those.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Fixed.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: _Fine_ lol

Robust.Shared/Network/NetManager.cs
Tyler-IN: You know, you can do suggestions.
```suggestion
```

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: No, this can be invoked on an arbitrary Stream.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Yep.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
ComicIronic: What? I meant like the check in `ReadMappedString` that the mapping is locked.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1290 @@
...
+            /// </remarks>
+            private const int FirstMappedIndexStart = 2;
+
+            /// <summary>
+            /// Write the encoding of the given string to the stream.
+            /// </summary>
+            /// <param name="stream">The stream to write to.</param>
+            /// <param name="value"> The (possibly null) string to write.</param>
+            public static void WriteMappedString(Stream stream, string? value)
+            {
```
ComicIronic, line 1147:
```suggestion
            {
                if (!LockMappedStrings)
                {
                    throw new InvalidOperationException("Not performing unlocked string mapping.");
                }
                
```

Tyler-IN:
Dammit. I had a suspicion.

Robust.Server/Player/PlayerManager.cs
```diff
@@ -359,18 +360,25 @@ private void EndSession(object? sender, NetChannelArgs args)
...
 
             netMsg.ServerName = _baseServer.ServerName;
             netMsg.ServerMaxPlayers = _baseServer.MaxPlayers;
             netMsg.TickRate = _timing.TickRate;
+
+            IPlayerSession session;
+            while (!TryGetSessionByChannel(channel, out session))
+            {
+                await Task.Delay(10);
+                if (!channel.IsConnected) return;
```
PJB3005, line 376: 1. Why is `MsgServerInfoReq` even being sent to the server before the client is properly connected?
2. Using `Task.Delay(10)` in a loop here is an *extremely* crappy way to do this.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgServerHandshake.cs
```diff
@@ -0,0 +1,65 @@
...
+                }
+
+                public override void WriteToBuffer(NetOutgoingMessage buffer)
+                {
+                    if (Hash == null)
+                    {
+                        throw new InvalidOperationException("Package has not been specified.");
+                    }
+
+                    buffer.WriteVariableInt32(Hash.Length);
```
PJB3005, line 55: Why even bother sending this with `WriteVariableInt32` when that still has to write at *least* a single byte and even a single byte can fit way more than the size of the SHA512 hash.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgServerHandshake.cs
```diff
@@ -0,0 +1,65 @@
...
+            /// the current string mapping, so the client can check if it has
+            /// a local copy.
+            /// </summary>
+            /// <seealso cref="NetworkInitialize"/>
+            [UsedImplicitly]
+            private class MsgServerHandshake : NetMessage
+            {
+
+                public MsgServerHandshake(INetChannel ch)
+                    : base($"{nameof(RobustSerializer)}.{nameof(MappedStringSerializer)}.{nameof(MsgServerHandshake)}", MsgGroups.Core)
```
PJB3005, line 28: Not keen on these huge names being duplicated instead of using a constant like all the other net messages do.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+        /// serializer can use the same mapping to recover the original string.
+        /// </remarks>
+        public partial class MappedStringSerializer : IStaticTypeSerializer
+        {
+
+            private static INetManager? _net;
+
+            private static readonly ISawmill LogSzr = Logger.GetSawmill("szr");
+
+            private static readonly HashSet<INetChannel> IncompleteHandshakes = new HashSet<INetChannel>();
```
PJB3005, line 60: Basically nothing in this class should be static.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+
+                IncompleteHandshakes.Add(channel);
+
+                var message = net.CreateNetMessage<MsgServerHandshake>();
+                message.Hash = MappedStringsHash;
+                net.ServerSendMessage(message, channel);
+
+                while (IncompleteHandshakes.Contains(channel))
+                {
+                    await Task.Delay(1);
```
PJB3005, line 96: Oh dear god doing `Task.Delay(1)` here is even worse. This should *absolutely* be explicitly waiting on an event/callback based on the actual messages coming in.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+                    LogSzr.Error("Received strings from client.");
+                    return;
+                }
+
+                LockMappedStrings = false;
+                ClearStrings();
+                DebugTools.Assert(msg.Package != null, "msg.Package != null");
+                LoadStrings(new MemoryStream(msg.Package!, false));
+                var checkHash = CalculateHash(msg.Package!);
+                if (!checkHash.SequenceEqual(ServerHash))
```
PJB3005, line 208: Doing LINQ sequence equal is pretty ugly here. Doing `checkHash.AsSpan().SequenceEqual(ServerHash)` is much better.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+                }
+
+                ClearStrings();
+
+                var hashStr = ConvertToBase64Url(Convert.ToBase64String(msg.Hash!));
+
+                LogSzr.Info($"Received server handshake with hash {hashStr}.");
+
+                var fileName = CacheForHash(hashStr);
+                if (!File.Exists(fileName))
```
PJB3005, line 345: This should not be using the general file system API, it should use the `UserData` provider in `ResourceManager`.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+                LockMappedStrings = false;
+
+                if (LockMappedStrings)
+                {
+                    throw new InvalidOperationException("Mapped strings are locked.");
+                }
+
+                ClearStrings();
+
+                var hashStr = ConvertToBase64Url(Convert.ToBase64String(msg.Hash!));
```
PJB3005, line 340: Not keen on this using Base64 instead of a hex string but I guess it is more compact filename wise.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+
+                LogSzr.Info($"Received server handshake with hash {hashStr}.");
+
+                var fileName = CacheForHash(hashStr);
+                if (!File.Exists(fileName))
+                {
+                    LogSzr.Info($"No string cache for {hashStr}.");
```
PJB3005, line 347: Most of these should be debug logs, not info logs.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+            /// </returns>
+            private static string CacheForHash(string hashStr)
+                => PathHelpers.ExecutableRelativeFile($"strings-{hashStr}");
+
+            /// <summary>
+            ///  Saves the string cache to a file based on it's hash.
+            /// </summary>
+            private static void WriteStringCache()
+            {
+                var hashStr = Convert.ToBase64String(MappedStringsHash);
```
PJB3005, line 407: I don't like the duplication of the hash -> file-safe base64 string conversion code between this and `HandleServerHandshake`.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+            {
+                using var ms = new MemoryStream();
+                WriteStringPackage(ms);
+                return ms.ToArray();
+            }
+
+            /// <summary>
+            ///     Strings longer than this will throw an exception and a better strategy will need to be employed to deal with large strings.
+            /// <summary>
+            public static int StringPackageMaximumBufferSize = 65536;
```
PJB3005, line 439: Any reason this isn't a const?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+            /// <para>
+            /// While <c>false</c>, strings can be added to the mapping, but
+            /// it cannot be saved to a cache.
+            /// </para>
+            /// <para>
+            /// While <c>true</c>, the mapping cannot be modified, but can be
+            /// shared between the server and client and saved to a cache.
+            /// </para>
+            /// </value>
+            public static bool LockMappedStrings { get; set; }
```
PJB3005, line 627: It feels like this *really* should not be public set.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+            /// </remarks>
+            /// <returns>
+            /// <c>true</c> if the string was added to the mapping for the first
+            /// time, <c>false</c> otherwise.
+            /// </returns>
+            /// <exception cref="InvalidOperationException">
+            /// Thrown if the mapping is locked, and strings cannot be added, or
+            /// if the string is not normalized (<see cref="String.IsNormalized()"/>).
+            /// </exception>
+            public static bool AddString(string str)
```
PJB3005, line 659: I am skeptical of how much the massive splitting of things in this method actually helps. It sounds to me like it would create a massive amount of extra strings that are *never* actually used on their own.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+            /// <summary>
+            /// Add the constant strings from an <see cref="Assembly"/> to the
+            /// mapping.
+            /// </summary>
+            /// <param name="asm">The assembly from which to collect constant strings.</param>
+            /// <exception cref="InvalidOperationException">
+            /// Thrown if the mapping is locked.
+            /// </exception>
+            [MethodImpl(MethodImplOptions.Synchronized)]
+            public static unsafe void AddStrings(Assembly asm)
```
PJB3005, line 797: It feels to me like dumping the entire string table of the assembly is a terrible idea both for security (leaking secrets) and bloat-of-dict reasons. You'll be sending *method, class and field names* that have not a snowball's chance in hell of ever touching the wire.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+                using var hasher = SHA512.Create();
+                var hash = hasher.ComputeHash(data);
+                return hash;
+            }
+
+            /// <summary>
+            /// Implements <see cref="ITypeSerializer.Handles"/>.
+            /// Specifies that this implementation handles strings.
+            /// </summary>
+            public bool Handles(Type type) => type == typeof(string);
```
PJB3005, line 1070: Make these explicit implementations?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
```diff
@@ -0,0 +1,1295 @@
...
+                {
+                    value = null;
+                    return;
+                }
+
+                if (mapIndex == UnmappedString)
+                {
+                    // not mapped
+                    var length = ReadCompressedUnsignedInt(stream, out _);
+                    var buf = new byte[length];
```
PJB3005, line 1206: This feels like a costly allocation to make.

Robust.Shared/Serialization/RobustSerializer.cs
```diff
@@ -6,95 +6,147 @@
...
+
+        #endregion
 
         public void Initialize()
         {
-            var types = reflectionManager.FindTypesWithAttribute<NetSerializableAttribute>().ToList();
-#if DEBUG
+            var mappedStringSerializer = new MappedStringSerializer();
+            var types = _reflectionManager.FindTypesWithAttribute<NetSerializableAttribute>().ToList();
+#if !FULL_RELEASE
```
PJB3005, line 50: Why should this check only run on full release?

Robust.Server/Player/PlayerManager.cs
TYoung86: 1. `MsgServerInfoReq` is handled by you guys. Feel free to delay it to after if it's possible.

2. Task.Delay hands control back to the task manager, Thread.Sleep would hang.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgServerHandshake.cs
TYoung86: Something between paranoia, flexibility, existing patterns and it only happens at handshake. Also, yeah it only writes `64` so just 1 byte.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.MsgServerHandshake.cs
TYoung86: image

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
TYoung86: It's an `IStaticTypeSerializer` - the `NetSerializer` implementation demands it.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
TYoung86:
```suggestion
                if (!checkHash.AsSpan().SequenceEqual(ServerHash))
```

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Yeah it does. Unfortunately many of the strings that are sent are fractional. A large portion of the generated strings are not sent.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Yeah.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Styling wise?

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: Could be a stackalloc probably.

Robust.Shared/Serialization/RobustSerializer.cs
Tyler-IN: Should show up for developers that have the code checked out, manipulating C#.

It isn't costly, it can be changed to show up for all.

Robust.Server/Player/PlayerManager.cs
PJB3005: 1. What kind of ridiculous argument is that? If you're gonna change the networking handshake protocol then you should go the whole way instead of this halfway "not my problem" garbage.
2. The problem is that you are using an async spin loop as a substitute for proper waiting on events in the first place, not which method you used.

Robust.Shared/Serialization/RobustSerializer.MappedStringSerializer.cs
Tyler-IN: It probably needs to be a delay of at least a game tick, yeah.

Robust.Shared/Serialization/RobustSerializer.cs
PJB3005: I could've sworn I didn't post this comment after thinking about it again.

But yeah my initial thought was that it shouldn't run on *plain* release either due to performance.

PR #1123, opened 2020-06-10
Implement corner collision resolution

Tyler-IN:
change CalculateNormal to apply rounded collision resolution to corners

reduce concurrency on concurrent dictionaries

add TODO for container mass

minor refactor of GetCollidingEntities, and don't re-enumerate modifiers in an inner loop

### DIFF
--- a/Robust.Client/Console/Commands/Debug.cs
+++ b/Robust.Client/Console/Commands/Debug.cs
@@ -758,7 +758,7 @@ namespace Robust.Client.Console.Commands
 
         public static Dictionary<string, FileSystemWatcher> _watchers;
 
-        public static ConcurrentDictionary<string, bool> _reloadShadersQueued = new ConcurrentDictionary<string, bool>();
+        public static ConcurrentDictionary<string, bool> _reloadShadersQueued = new ConcurrentDictionary<string, bool>(1,4);
 
         public bool Execute(IDebugConsole console, params string[] args)
         {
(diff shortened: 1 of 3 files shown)


Robust.Shared/Physics/PhysicsManager.cs
```diff
@@ -52,22 +55,29 @@ public bool IsWeightless(GridCoordinates gridPosition)
...
+            var rightDist = targetBox.Right - sourceBox.Left;
+            var x = leftDist - rightDist;
+            // Y is the axis of separation
+            var bottomDist = sourceBox.Top - targetBox.Bottom;
+            var topDist = targetBox.Top - sourceBox.Bottom;
+            var y = bottomDist - topDist;
+
+            // octagonal surface normals; create 8 regions, a third of each of the 4 faces is flat
+            // corners are treated as curved during resolution, making players feel somewhat pill shaped
+            const float faceToCornerBias = 1/3f; // higher is more square, higher is more pill shaped
```
RemieRichards: comment seems wrong

Robust.Client/Console/Commands/Debug.cs
```diff
@@ -758,7 +758,7 @@ internal class ReloadShadersCommand : IConsoleCommand
 
         public static Dictionary<string, FileSystemWatcher> _watchers;
 
-        public static ConcurrentDictionary<string, bool> _reloadShadersQueued = new ConcurrentDictionary<string, bool>();
+        public static ConcurrentDictionary<string, bool> _reloadShadersQueued = new ConcurrentDictionary<string, bool>(1,4);
```
ComicIronic, line 761: Is there any particular reason for choosing 4 as the initial capacity?

Robust.Shared/Physics/PhysicsManager.cs
Tyler-IN: how so?

Robust.Client/Console/Commands/Debug.cs
Tyler-IN: It's a small power of 2

Robust.Shared/Physics/PhysicsManager.cs
Tyler-IN: oh - higher is more square, lower is more pill shaped

Tyler-IN:
```suggestion
            const float faceToCornerBias = 1/3f; // higher is more square, lower is more pill shaped
```

Robust.Client/Console/Commands/Debug.cs
Tyler-IN: (Smaller than the default size, still scales acceptably)

PJB3005 (commented):
It feels like a terrible idea to make this enabled for every single bounding box unconditionally.

PR #1126, opened 2020-06-12
Shared String Dictionary ctd.

Tyler-IN:
the great refactoring

also fixes integration tests

also fixes integration tests logging on windows

also makes integration tests just log to the integration test logs instead of a file

but it's a hack - can be refactored out of scope of this PR


Tyler-IN:
Preferaby, Func\<ILogHandler\> can be replaced w/ IoC in a later PR.

Tyler-IN:
should be no cases of 'new byte[]' in the robust serializer & associated classes, and string alloc in unpacking is limited to max string length for dict (420) so no more 64k buffer alloc'd there

Robust.Server/Maps/MapLoader.cs
```diff
@@ -99,6 +101,8 @@ public void SaveBlueprint(GridId gridId, string yamlPath)
 
                 var data = new MapData(reader);
 
+                LoadedMapData?.Invoke(data.Stream, resPath.ToRelativeSystemPath());
```
PJB3005: System path should not be exposed like this.

PJB3005 (commented):
So this is interesting. Apparently `server_config.toml` already has a definition for `log.enabled = false` and this is causing it to be disabled. This isn't an issue on its own but I did notice that when logging is disabled like this the log file still gets *created*, but it stays empty.

Robust.Server/Maps/MapLoader.cs
Tyler-IN: Honestly I have no idea why it's got System in the name.
"/\Maps\saltern.yml" is relative to the resource root.

Tyler-IN:
So should I remove `log.enabled` handling?

PJB3005:
Since you provided a suitable alternative for the integration tests I guess that could work, yeah.

PR #1128, opened 2020-06-12
Upgrades SixLabors.ImageSharp

Tyler-IN:
Also adds InlineIl.Fody for fast access to image backing buffers.


gradientvera (approved):

TYoung86:
Looks like builder fails cause content out of sync.

Tyler-IN:
Builder fails 'cause content still refs old lib versions. See https://github.com/space-wizards/space-station-14/pull/1120

PR #1130, opened 2020-06-13
Use CannyFastMath & Update Some Packages

Tyler-IN:
Use CannyFastMath in various places (but not everywhere).

Even in some places that don't really matter.

Fix some float/double miscasts.

Make Lidgren use netcoreapp3.1 to use CannyFastMath and reduce overhead of netstandard aliasing.

Probably doesn't benefit that much, can revert parts.

Physics and entity queries probably benefit the most.

### DIFF
--- a/Robust.Shared.Maths/Angle.cs
+++ b/Robust.Shared.Maths/Angle.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using JetBrains.Annotations;
+using Math = CannyFastMath.Math;
+using MathF = CannyFastMath.MathF;
 
 namespace Robust.Shared.Maths
 {
--- a/Robust.Client/GameStates/ClientGameStateManager.cs
+++ b/Robust.Client/GameStates/ClientGameStateManager.cs
@@ -20,6 +20,8 @@ using Robust.Shared.Interfaces.Map;
 using Robust.Shared.Interfaces.Timing;
 using Robust.Shared.Timing;
 using Robust.Shared.Utility;
+using Math = CannyFastMath.Math;
+using MathF = CannyFastMath.MathF;
 
 namespace Robust.Client.GameStates
 {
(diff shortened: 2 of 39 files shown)


gradientvera (approved):
Seems fine.

Lidgren.Network/NetUtility.cs
```diff
@@ -352,9 +355,9 @@ public static string ToHumanReadable(long bytes)
 		{
 			if (bytes < 4000) // 1-4 kb is printed in bytes
 				return bytes + " bytes";
-			if (bytes < 1000 * 1000) // 4-999 kb is printed in kb
-				return Math.Round(((double)bytes / 1000.0), 2) + " kilobytes";
-			return Math.Round(((double)bytes / (1000.0 * 1000.0)), 2) + " megabytes"; // else megabytes
+			if (bytes < 1000000) // 4-999 kb is printed in kb
```
ComicIronic: C# now has underscores as a numeric separator, so I would appreciate this being `1_000_000` for readability.

Robust.Shared.Maths/Angle.cs
```diff
@@ -52,28 +55,28 @@ public Vector2 ToVec()
             return new Vector2((float) x, (float) y);
         }
 
-        private const double Segment = 2 * Math.PI / 8.0; // Cut the circle into 8 pieces
+        private const double Segment = 2 * Math.π / 8.0; // Cut the circle into 8 pieces
```
ComicIronic: Don't think Unicode is a good idea in code files.

Lidgren.Network/NetUtility.cs
Tyler-IN:
```suggestion
			if (bytes < 1_000_000) // 4-999 kb is printed in kb
```

Tyler-IN:
Tests were failing due to some bad find-and-replaces.

Beefed up tests to report multiple assertions in some places too.

Tyler-IN:
See https://github.com/space-wizards/space-station-14/pull/1124 for test results

PR #1131, opened 2020-06-13
Added KeyValuePair IsNull extension method

DamianX:
Makes it easier to use with FirstOrDefault.


DamianX:
As it turns out you can just use `dictionary.Key == default`

PR #1132, opened 2020-06-13
Input Handling + other cleanups to support click drag functionality

chairbender:
Add some various capabilities related to clickdrag functionality, also a few small cleanups here and there as I noticed them.

New stuff:

- copy spritecomponent from another (used for the drag drop shadow)
- "replay" in InputHandler, allows clientside to "store" an input to be later re-triggered, this is used so that drag and drop can block the mousedown event when you mousedown on a draggable item, then re-fire that event if the mouseup happens before the interaction turns into a drag
- input handlers ignore mouseup by default, now you can tell them to not do that


Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -178,6 +178,27 @@ public bool Directional
         [ViewVariables(VVAccess.ReadWrite)]
         public bool IsInert { get; private set; }
 
+        /// <summary>
+        /// Update this sprite component to visibly match the other
+        /// </summary>
+        public void CopyFrom(SpriteComponent other)
```
PJB3005, line 192: Is this supposed to only do a copy at the time it's called or is it supposed to keep the visual data in sync? Because this doesn't do keep copies of various things like the `Layers` list, it's gonna entangle those parts of the sprite while other stuff like `color` will stay separate.

If these are supposed to stay perfectly in sync, I feel a different API is necessary (maybe a way to hook into the renderer so the other entity can be visually copied immediately?).  If it's not then this should do deeper copies of things like `Layers` and such to prevent entanglement.

PJB3005 (commented):
I'm not 100% sure on those input system changes but I'll have to check the content-side PR to see.

Robust.Shared/Input/InputCmdHandler.cs
```diff
@@ -2,7 +2,7 @@
 using Robust.Shared.Map;
 using Robust.Shared.Players;
 
-namespace Robust.Shared.Input
+namespace Robust.Shared.Input.Binding
```
ComicIronic, line 5: You should move this file.

chairbender:
@ComicIronic @PJB3005 thanks, fixed

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -176,6 +177,31 @@ public bool Directional
...
 
+        /// <summary>
+        /// Update this sprite component to visibly match the current state of other at the time
+        /// this is called. Does not keep them perpetually in sync.
+        /// This does some deep copying thus exerts some gc pressure, so avoid this for hot code paths.
+        /// </summary>
+        public void CopyFrom(SpriteComponent other)
+        {
+            //deep copying things to avoid entanglement
+            this._baseRsi = other._baseRsi == null ? null : new RSI(other._baseRsi);
```
PJB3005: RSIs aren't mutable so this deep copy isn't necessary.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -176,6 +177,31 @@ public bool Directional
...
+            this._directional = other._directional;
+            this._visible = other._visible;
+            this._layerMapShared = other._layerMapShared;
+            this.color = other.color;
+            this.offset = other.offset;
+            this.rotation = other.rotation;
+            this.scale = other.scale;
+            this.drawDepth = other.drawDepth;
+            this.Layers = new List<Layer>(other.Layers);
```
PJB3005: The actual objects *inside* the layer list also need to be cloned.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -176,6 +177,31 @@ public bool Directional
...
+            this.IsInert = other.IsInert;
+            this.LayerMap = other.LayerMap.ToDictionary(entry => entry.Key,
+                entry => entry.Value);
+            this.PostShader = other.PostShader?.Duplicate();
```
PJB3005: You can check if `PostShared.Mutable` is false here and if it is you don't have to duplicate it.

PJB3005 (changes requested):

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -183,6 +184,44 @@ public bool Directional
...
+            this._layerMapShared = other._layerMapShared;
+            this.color = other.color;
+            this.offset = other.offset;
+            this.rotation = other.rotation;
+            this.scale = other.scale;
+            this.drawDepth = other.drawDepth;
+            this.Layers = new List<Layer>(other.Layers.Count);
+            foreach (var otherLayer in other.Layers)
+            {
+                this.Layers.Add(new Layer(otherLayer));
```
PJB3005, line 207: This `Layer` constructor doesn't actually duplicate the shader in it (even if necessary) so still gotta do that.

PJB3005 (approved):

PR #1139, opened 2020-06-19
Dynamically Manage Collision Solver Iterations

Tyler-IN:
increase solver iterations by 1 at 60tps

modulate solver iterations dynamically depending on tps


Tyler-IN:
dramatically reduces ability to clip into / thru things at low tick rates

PR #1142, opened 2020-06-22
Add toggle outline cvar

dch-GH:


PJB3005:
There is no reason for this cvar to be part of the engine, when content is the only thing that cares about it.

dch-GH:
@PJB3005 Where should I register it in Client? I was having trouble finding the right place to register client cvars.

PJB3005:
`EntryPoint.cs` is a good place to.

dch-GH:
I figured so. Just not seeing any other cvars being registered there put me off. Will do.

dch-GH:
M'kay, https://github.com/space-wizards/space-station-14/pull/1185 should be ready now.

PR #1143, opened 2020-06-22
ComponentManager Rework

Tyler-IN:
create indexing solution, UniqueIndex<TKey,TValue>

rework ComponentManager with flatter dictionaries and indexes

split out ComponentEventArgs, add sealed variants

fix up documentation

fix terrible crap

make benchmark work

needs change on content

### DIFF
--- a/Robust.Server/GameObjects/ServerEntityManager.cs
+++ b/Robust.Server/GameObjects/ServerEntityManager.cs
@@ -335,7 +335,7 @@ namespace Robust.Server.GameObjects
             = new ThreadLocal<PlayerSeenEntityStatesResources>(() => new PlayerSeenEntityStatesResources());
 
         /// <inheritdoc />
-        public List<EntityState>? UpdatePlayerSeenEntityStates(GameTick fromTick, IPlayerSession player, float range, bool inParallel = false)
+        public List<EntityState>? UpdatePlayerSeenEntityStates(GameTick fromTick, IPlayerSession player, float range)
         {
             var playerEnt = player.AttachedEntity;
             if (playerEnt == null)
--- a/Robust.Server/GameObjects/ServerEntityManager.cs
+++ b/Robust.Server/GameObjects/ServerEntityManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Prometheus;
 using Robust.Server.GameObjects.Components;
 using Robust.Server.GameObjects.Components.Container;
@@ -172,13 +173,17 @@ namespace Robust.Server.GameObjects
 
         private ISet<EntityUid> GetSeenMovers(IPlayerSession player)
         {
-            if (!_seenMovers.TryGetValue(player, out var movers))
+            lock (_seenMovers)
             {
-                movers = new SortedSet<EntityUid>();
-                _seenMovers.Add(player, movers);
-            }
+                if (!_seenMovers.TryGetValue(player, out var movers))
+                {
+                    movers = new SortedSet<EntityUid>();
+                    _seenMovers.Add(player, movers);
+                }
 
-            return movers;
+
+                return movers;
+            }
         }
 
         private readonly Dictionary<IPlayerSession, Dictionary<EntityUid, GameTick>> _playerLastSeen
@@ -265,7 +270,10 @@ namespace Robust.Server.GameObjects
 
         public void DropPlayerState(IPlayerSession player)
         {
-            _playerLastSeen.Remove(player);
+            lock (_playerLastSeen)
+            {
+                _playerLastSeen.Remove(player);
+            }
         }
 
         private void IncludeRelatives(IEnumerable<IEntity> children, HashSet<IEntity> set)
@@ -309,32 +317,25 @@ namespace Robust.Server.GameObjects
             }
         }
 
-        private readonly struct PlayerSeenEntityStatesResources
+        private class PlayerSeenEntityStatesResources
         {
 
-            public readonly HashSet<EntityUid> IncludedEnts;
+            public readonly HashSet<EntityUid> IncludedEnts = new HashSet<EntityUid>();
 
-            public readonly List<EntityState> EntityStates;
+            public readonly List<EntityState> EntityStates = new List<EntityState>();
 
-            public readonly HashSet<EntityUid> NeededEnts;
+            public readonly HashSet<EntityUid> NeededEnts = new HashSet<EntityUid>();
 
-            public readonly HashSet<IEntity> Relatives;
+            public readonly HashSet<IEntity> Relatives = new HashSet<IEntity>();
 
-            public PlayerSeenEntityStatesResources(bool memes = false)
-            {
-                IncludedEnts = new HashSet<EntityUid>();
-                EntityStates = new List<EntityState>();
-                NeededEnts = new HashSet<EntityUid>();
-                Relatives = new HashSet<IEntity>();
-            }
 
         }
 
-        private readonly PlayerSeenEntityStatesResources _playerSeenEntityStatesResources
-            = new PlayerSeenEntityStatesResources(false);
+        private readonly ThreadLocal<PlayerSeenEntityStatesResources> _playerSeenEntityStatesResources
+            = new ThreadLocal<PlayerSeenEntityStatesResources>(() => new PlayerSeenEntityStatesResources());
 
         /// <inheritdoc />
-        public List<EntityState>? UpdatePlayerSeenEntityStates(GameTick fromTick, IPlayerSession player, float range)
+        public List<EntityState>? UpdatePlayerSeenEntityStates(GameTick fromTick, IPlayerSession player, float range, bool inParallel = false)
         {
             var playerEnt = player.AttachedEntity;
             if (playerEnt == null)
@@ -352,10 +353,11 @@ namespace Robust.Server.GameObjects
 
             var seenMovers = GetSeenMovers(player);
 
-            var checkedEnts = _playerSeenEntityStatesResources.IncludedEnts;
-            var entityStates = _playerSeenEntityStatesResources.EntityStates;
-            var neededEnts = _playerSeenEntityStatesResources.NeededEnts;
-            var relatives = _playerSeenEntityStatesResources.Relatives;
+            var pseStateRes = _playerSeenEntityStatesResources.Value;
+            var checkedEnts = pseStateRes.IncludedEnts;
+            var entityStates = pseStateRes.EntityStates;
+            var neededEnts = pseStateRes.NeededEnts;
+            var relatives = pseStateRes.Relatives;
             checkedEnts.Clear();
             entityStates.Clear();
             neededEnts.Clear();
--- a/Robust.Server/GameObjects/ServerEntityManager.cs
+++ b/Robust.Server/GameObjects/ServerEntityManager.cs
@@ -353,7 +353,7 @@ namespace Robust.Server.GameObjects
 
             var seenMovers = GetSeenMovers(player);
 
-            var pseStateRes = _playerSeenEntityStatesResources.Value;
+            var pseStateRes = _playerSeenEntityStatesResources.Value!;
             var checkedEnts = pseStateRes.IncludedEnts;
             var entityStates = pseStateRes.EntityStates;
             var neededEnts = pseStateRes.NeededEnts;
--- a/Robust.Server/GameStates/ServerGameStateManager.cs
+++ b/Robust.Server/GameStates/ServerGameStateManager.cs
@@ -175,6 +175,7 @@ namespace Robust.Server.GameStates
 
             foreach (var (msg, chan) in mailBag)
             {
+                // see session.Status != SessionStatus.InGame above
                 if (chan == null) continue;
                 _networkManager.ServerSendMessage(msg, chan);
             }
--- a/Robust.Server/GameStates/ServerGameStateManager.cs
+++ b/Robust.Server/GameStates/ServerGameStateManager.cs
@@ -1,5 +1,8 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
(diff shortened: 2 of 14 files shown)


Robust.Shared/Physics/PhysicsManager.cs
```diff
@@ -220,7 +220,7 @@ public void RemoveBody(IPhysBody physBody)
             }
 
             if (!removed)
-                Logger.WarningS("phys", $"Trying to remove unregistered PhysicsBody! {physBody.Owner}");
+                Logger.WarningS("phys", $"Trying to remove unregistered PhysicsBody! {physBody.Owner.Uid}");
```
Tyler-IN, line 223: `physBody.Owner` is an `Entity`, this invokes `Entity.ToString()` which gets `Name` which gets `MetaDataComponent` which is already deleted at this juncture.

Robust.Shared/GameObjects/ComponentEventArgs.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <summary>
+    /// Arguments for an event related to a component being added.
+    /// </summary>
+    public sealed class AddedComponentEventArgs : ComponentEventArgs
+    {
+        /// <summary>
+        /// Constructs a new instance of <see cref="AddedComponentEventArgs"/>.
+        /// </summary>
+        /// <param name="component">The relevant component</param>
+        public AddedComponentEventArgs([JetBrains.Annotations.NotNull] IComponent component) : base(component) { }
```
ComicIronic: This isn't the right kind of `NotNull` to use, I don't think.

RemieRichards: it isn't.

Robust.Shared/Utility/UniqueIndex.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using JetBrains.Annotations;
+
+namespace Robust.Shared.Utility
+{
+
+    [PublicAPI]
+    public struct UniqueIndex<TKey, TValue> : IEnumerable<KeyValuePair<TKey, ISet<TValue>>>
```
ComicIronic: Could use documentation for use cases/performance/etc.

Robust.Shared/GameObjects/ComponentEventArgs.cs
Tyler-IN: It's an artifact of moving the class by refactoring from a file with `#nullable enable` to one that doesn't have it.

Robust.Shared/Physics/PhysicsManager.cs
Tyler-IN: Any idea why this is a warning and not just an early exit?

Robust.Shared/GameObjects/ComponentManager.cs
```diff
@@ -415,106 +402,75 @@ public bool TryGetComponent(EntityUid uid, uint netId, [NotNullWhen(true)] out I
...
         }
 
         /// <inheritdoc />
         public IEnumerable<IComponent> GetNetComponents(EntityUid uid)
         {
-            return _netComponents[uid].Values;
+            var comps = _entCompIndex[uid];
+            foreach (var comp in comps)
+            {
+                if (comp.Deleted || comp.NetID == null) continue;
```
PJB3005, line 451: `comp.NetID == null` should never be possible, should it?

Robust.Shared/GameObjects/ComponentManager.cs
```diff
@@ -415,106 +402,75 @@ public bool TryGetComponent(EntityUid uid, uint netId, [NotNullWhen(true)] out I
...
         {
-            foreach (var (type, kvTypeDict) in _dictComponents)
+            var comps = _entCompIndex[uid];
+            foreach (var comp in comps)
             {
-                if (kvTypeDict.TryGetValue(uid, out var comp) && comp is T t && type == comp.GetType() && !comp.Deleted)
-                {
-                    yield return t;
-                }
+                if (comp.Deleted || !(comp is T tComp)) continue;
```
PJB3005, line 439: `comp is T tComp` should never be capable of failing, should it?

Robust.Shared/GameObjects/Entity.cs
```diff
@@ -140,27 +143,31 @@ public void InitializeComponents()
         {
             Initializing = true;
             // Initialize() can modify the collection of components.
-            var components = EntityManager.ComponentManager.GetComponents(Uid);
-            foreach (var t in components)
-            {
-                var comp = (Component) t;
-                if (comp != null && !comp.Initialized)
+            var components = EntityManager.ComponentManager.GetComponents(Uid)
+                .OrderBy(x => x switch
```
PJB3005, line 147: Why do these components have to be initialized in this order? The fact that they're not order independent feels like a bug.

Robust.Shared/Physics/PhysicsManager.cs
PJB3005: No idea.
image

Robust.Shared/Utility/UniqueIndex.cs
```diff
@@ -0,0 +1,209 @@
...
+    /// </remarks>
+    /// <typeparam name="TKey">The type of key.</typeparam>
+    /// <typeparam name="TValue">The type of value.</typeparam>
+    /// <seealso cref="UniqueIndexExtensions"/>
+    /// <seealso cref="IUniqueIndex{TKey,TValue}" />
+    [PublicAPI]
+    public struct UniqueIndex<TKey, TValue> : IUniqueIndex<TKey, TValue> where TKey : notnull
+    {
+
+        private ImmutableDictionary<TKey, ISet<TValue>> _index;
```
PJB3005, line 32: Storing `ISet<>` values in this dictionary instead of `HashSet<>`s directly feels like it's just leaving performance on the table for no good reason.

Robust.Shared/Physics/PhysicsManager.cs
Acruid: Calling RemoveBody() multiple times with the same PhysicsBody is usually a symptom of a bug, so that is why there is a logged warning.

Tyler-IN: Is it a bug here? Pretty sure this only shows up during disposal now. Could be ifdef'd out, made to check for the scenario, or something?

Robust.Shared/GameObjects/ComponentManager.cs
Tyler-IN:
```cs
        /// <inheritdoc />
        [ViewVariables]
        public virtual uint? NetID => null;
```
Change it to `abstract` and remove the `?` maybe?

Robust.Shared/GameObjects/Entity.cs
Tyler-IN: Yeah it is. Out of scope tho.

Robust.Shared/GameObjects/ComponentManager.cs
Tyler-IN: `_entCompIndex` doesn't have components of all the same type.

Robust.Shared/Utility/UniqueIndex.cs
Tyler-IN: Haven't measured to check.
Might have scenarios where sets will be mixed to incl. sortedset / immutable.
Call overhead to retrieve from set might not be significant compared to rest of operations.

Robust.Shared/GameObjects/ComponentManager.cs
PJB3005: ...no?

If a component doesn't have a net ID it won't end up in the netID index in the first place.

Robust.Shared/GameObjects/ComponentManager.cs
PJB3005: Oh, alright.

Robust.Shared/Utility/UniqueIndex.cs
PJB3005: > Might have scenarios where sets will be mixed to incl. sortedset / immutable.

Then the performance overhead from making it "flexible" like that makes it unsuitable for the highly specific purpose we're using it for. I don't think this is something worry about.

Robust.Shared/GameObjects/ComponentManager.cs
PJB3005: Oh wait it's the same deal as the other case, I thought this was iterating an index specifically for net components like we had in the previous component manager. My bad.

Robust.Shared/Utility/UniqueIndex.cs
Tyler-IN: An example would be if a set of components on a control that will never have any more components added should be converted to frozen immutable to reap the benefits.

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -160,6 +166,17 @@ public void SendGameStateUpdate()
...
+
+                return (stateUpdateMessage, channel);
+            }
+
+            var mailBag = _playerManager.GetAllPlayers()
+                .AsParallel().Select(GenerateMail);
+
+            foreach (var (msg, chan) in mailBag)
+            {
+                if (chan == null) continue;
```
PJB3005, line 179: Why would this ever be null?

Robust.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -309,32 +317,25 @@ private static void AddContainedRecursive(IEntity ent, HashSet<IEntity> set)
...
         }
 
-        private readonly PlayerSeenEntityStatesResources _playerSeenEntityStatesResources
-            = new PlayerSeenEntityStatesResources(false);
+        private readonly ThreadLocal<PlayerSeenEntityStatesResources> _playerSeenEntityStatesResources
+            = new ThreadLocal<PlayerSeenEntityStatesResources>(() => new PlayerSeenEntityStatesResources());
 
         /// <inheritdoc />
-        public List<EntityState>? UpdatePlayerSeenEntityStates(GameTick fromTick, IPlayerSession player, float range)
+        public List<EntityState>? UpdatePlayerSeenEntityStates(GameTick fromTick, IPlayerSession player, float range, bool inParallel = false)
```
PJB3005: This parameter is unused?

Robust.Server/GameStates/ServerGameStateManager.cs
Tyler-IN: Artifact from quick development, was initially a collection that could return a default tuple which would have default/null chan before moving to select. I think it's safe to remove.

Robust.Server/GameObjects/ServerEntityManager.cs
Tyler-IN: Artifact from quick development too. Was a quick pull local as parameter refactor before adding locks and ThreadLocal.

Robust.Server/GameStates/ServerGameStateManager.cs
Tyler-IN: Nevermind, can be null. Added comment pointing to reason.

Robust.Shared/Physics/PhysicsManager.cs
Tyler-IN: image

Tyler-IN: This was fixed btw.

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -160,6 +166,18 @@ public void SendGameStateUpdate()
                 {
                     _ackedStates[channel.ConnectionId] = _gameTiming.CurTick;
                 }
+
+                return (stateUpdateMessage, channel);
+            }
+
+            var mailBag = _playerManager.GetAllPlayers()
+                .AsParallel().Select(GenerateMail);
```
Tyler-IN, line 174: Can probably filter the player session list to exclude session.Status != SessionStatus.InGame, maybe introduce IPlayerManager.GetPlayersInGame.

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -113,13 +118,17 @@ public void SendGameStateUpdate()
 
             var oldestAck = GameTick.MaxValue;
 
+            var deps = new DependencyCollection();
+            deps.RegisterInstance<ILogManager>(IoCManager.Resolve<ILogManager>());
+            deps.BuildGraph();
```
Tyler-IN, line 123: Could maybe cache this and move it out.

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -113,13 +118,17 @@ public void SendGameStateUpdate()
...
-            foreach (var session in _playerManager.GetAllPlayers())
+            (MsgState, INetChannel) GenerateMail(IPlayerSession session)
             {
+                IoCManager.InitThread(deps, true);
+
                 if (session.Status != SessionStatus.InGame)
                 {
-                    continue;
+                    return default;
                 }
```
Tyler-IN, line 132: If this was filtered out from player list as mentioned below, wouldn't need this check and wouldn't return nulls.

Robust.Server/GameStates/ServerGameStateManager.cs
Tyler-IN: Could convert this to use dedicated worker threads that have dedicated IoC setups that use OS sync primitives to wait between ticks - could kick them off into low-thread-priority spin waits at the beginning of a tick, before issuing work, to reduce thread context activation latency. Kick them up to near realtime when legit working and then back them off to low when spinning between work.
When the tick ends or this method ends then send them back to sleeping on a OS primitive.
This is quite a bit less efficient and only _slightly_ less latent than what plinq is gonna do.
It might be worth it anyway depending on criticality/guarantees of timing.

PR #1146, opened 2020-06-24
Implement MultipleTypeEntityQuery

ShadowCommander:
Allows selecting entities that match multiple components


PJB3005 (commented):
This would be *much* better suited as doing a query for the first type (much faster since it doesn't look at every entity) and then doing the manual `HasComponent` calls later.

PR #1148, opened 2020-06-24
Add BeforeDraw method to Overlay.cs

thedraccx:
Add BeforeDraw method to overlays to allow for necessary setup

### DIFF
--- a/Robust.Client/Graphics/Overlays/Overlay.cs
+++ b/Robust.Client/Graphics/Overlays/Overlay.cs
@@ -69,6 +69,10 @@ namespace Robust.Client.Graphics.Overlays
         {
         }
 
+        public virtual void BeforeDraw()
+        {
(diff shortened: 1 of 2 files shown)


PJB3005:
I'm not seeing the use for this. Couldn't this be done by the code that adds the overlay?

thedraccx:
Well since it's so decoupled at the moment you can't actually get a reference to the overlay you're adding. It's just like ChangeOverlay(enum). So I needed a way to know when the overlay was about to start being drawn so I could properly set the start time, and take a screenshot of the game world.

PJB3005:
That's better solved by fixing the crappy screen effects component content-side, probably?

thedraccx:
Yeah you're right. For some reason I thought that one was also under RobustToolbox. I'll look into that instead

PR #1153, opened 2020-06-25
Protect IgnoresAccessChecksTo, Add Task Scheduler Context, Lock on Console Log Flushing

Tyler-IN:
Move Impl of IgnoreAccessChecksToAttribute to Robust.Shared, make it internal

launch server into task scheduler context with custom task scheduler that does not run things inline / on the main thread

### DIFF
--- a/Robust.Client/GameController.cs
+++ b/Robust.Client/GameController.cs
@@ -331,6 +331,10 @@ namespace Robust.Client
                 args.SetObserved(); // don't crash
 #endif
             };
+
+#if !DEBUG
+            ConsoleLogHandler.TryDetachFromConsoleWindow();
+#endif
         }
 
         private string GetUserDataDir()
--- a/Robust.Server/Program.cs
+++ b/Robust.Server/Program.cs
@@ -26,8 +26,6 @@ namespace Robust.Server
             Start(args);
         }
 
-        private static Thread? _offMainThread;
-
         internal static void Start(string[] args, bool contentStart = false)
         {
             if (_hasStarted)
--- a/Robust.Server/Program.cs
+++ b/Robust.Server/Program.cs
@@ -42,6 +42,8 @@ namespace Robust.Server
 
             ThreadPool.SetMinThreads(Environment.ProcessorCount * 2, Environment.ProcessorCount);
 
+            // this sets up TaskScheduler.Current for new tasks to be scheduled off the main thread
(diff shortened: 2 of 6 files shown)


PJB3005:
Please fix the compiler warning for `_offThread` being unused now.

Robust.Client/GameController.cs
```diff
@@ -331,6 +331,10 @@ internal static void SetupLogging(ILogManager logManager, Func<ILogHandler> logH
                 args.SetObserved(); // don't crash
 #endif
             };
+
+#if !DEBUG
+            ConsoleLogHandler.TryDetachFromConsoleWindow();
```
PJB3005, line 336: Wait hang on I'm re-reading this. What's this supposed to accomplish?

Robust.Shared/Asynchronous/RobustTaskScheduler.cs
```diff
@@ -0,0 +1,86 @@
...
+            if ((options & TaskCreationOptions.LongRunning) == 0)
+            {
+                ThreadPool.UnsafeQueueUserWorkItem((IThreadPoolWorkItem) task, false);
+            }
+            else
+            {
+                var wantsToBeFair = (task.CreationOptions & TaskCreationOptions.PreferFairness) != 0;
+                new Thread(Execute)
+                {
+                    Name = "Long Running Task",
```
PJB3005, line 60: Also: I would prefer if the main game thread actually had a unique name so debugging is easier.

Robust.Server/Program.cs
```diff
@@ -41,17 +42,15 @@ internal static void Start(string[] args, bool contentStart = false)
...
 
-            _offMainThread = new Thread(() => ParsedMain(parsed, contentStart))
-            {
-                IsBackground = false,
-                Name = "Server",
-                Priority = ThreadPriority.Highest,
-                CurrentCulture = CultureInfo.InvariantCulture,
-                CurrentUICulture = CultureInfo.InvariantCulture
-            };
+            ThreadPool.SetMinThreads(Environment.ProcessorCount * 2, Environment.ProcessorCount);
```
Tyler-IN, line 44:
```suggestion

            // this sets up TaskScheduler.Current for new tasks to be scheduled off the main thread
            // the LongRunning option causes it to not be scheduled on the thread pool, so it's just a plain thread with a task context
```

PR #1154, opened 2020-06-25
Mutual Types Hash Verification

Tyler-IN:
have clients and servers verify their serializer types hashes with each other


Robust.Shared/Serialization/RobustMappedStringSerializer.cs
```diff
@@ -254,35 +256,43 @@ private void HandleStringsMessage(INetManager net, MsgRobustMappedStringsSeriali
...
+            if (!msg.NeedsStrings)
             {
-                LogSzr.Debug($"Completing handshake with {msgRobustMappedStringsSerializer.MsgChannel.RemoteEndPoint.Address}.");
-                _incompleteHandshakes.Remove(msgRobustMappedStringsSerializer.MsgChannel);
+                LogSzr.Debug($"Client reports serializer types hash {NetUtility.ToHexString(msg.TypesHash)}");
+                if (!msg.TypesHash.AsSpan().SequenceEqual(TypesHash))
+                {
+                    LogSzr.Error($"Client's serializer types hash does not match!\nClient: {NetUtility.ToHexString(msg.TypesHash)}\nServer: {NetUtility.ToHexString(TypesHash)}");
+                    _incompleteHandshakes.Remove(msg.MsgChannel);
+                    msg.MsgChannel.Disconnect("Incompatible serializer: types hash must match.");
```
PJB3005, line 276: Since this will get displayed to users, it would be a good idea to put something descriptive like `"This usually indicates a version mismatch.\nClose the game and reconnect to allow the launcher to download an update."`

Robust.Shared/Serialization/Messages/MsgRobustMappedStringsSerializerServerHandshake.cs
```diff
@@ -37,17 +42,32 @@ public override void ReadFromBuffer(NetIncomingMessage buffer)
             }
 
             buffer.ReadBytes(Hash = new byte[len]);
+
+            len = buffer.ReadVariableInt32();
```
PJB3005, line 46: Does it make much sense to send a length parameter here? We know the length of the hashes ahead of time and if they ever change in the future (already doubtful) the result will just be a slightly different exception.

PJB3005 (commented):
It may make sense to move serializer hash verification to earlier in the handshake, ideally during the initial connection handshake.

PR #1156, opened 2020-06-25
Logging Tweaks

Tyler-IN:
reduce size of backing buffer, alloc conversion buffer on stack

add ability for windows to output utf8 for testing

use utf16 output on windows

### DIFF
--- a/Robust.Shared/Log/ConsoleLogHandler.cs
+++ b/Robust.Shared/Log/ConsoleLogHandler.cs
@@ -4,7 +4,9 @@ using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Text.Unicode;
 using System.Timers;
+using JetBrains.Annotations;
 
 namespace Robust.Shared.Log
 {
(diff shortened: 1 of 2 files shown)


Robust.Shared/Log/ConsoleLogHandler.cs
```diff
@@ -96,16 +101,43 @@ public void Log(in LogMessage message)
...
+                    foreach (var chunk in _line.GetChunks())
+                    {
+                        var chunkSize = chunk.Length;
+                        var totalRead = 0;
+                        var span = chunk.Span;
+                        for (;;)
+                        {
+                            var finalChunk = totalRead + chunkSize >= totalChars;
+                            Utf8.FromUtf16(span, buf, out var read, out var wrote, isFinalBlock: finalChunk);
+                            _stream.Write(MemoryMarshal.AsBytes(buf.Slice(0, wrote)));
```
PJB3005: You shouldn't need that `AsBytes` anymore should you?

Tyler-IN:
```suggestion
                            _stream.Write(buf.Slice(0, wrote));
```

PR #1166, opened 2020-07-06
Implement ListContainer

ShadowCommander:
This is a list that hides controls that are out of view by removing them as children. When a child gets added it gets added to the internal list. The `LayoutUpdateOverride` adds and removes children of the `ListContainer` based on the visible indices of the list.

There is a little bit of bouncing when resizing the `ListContainer` when at the top or at the bottom, but that seems to be due to the `ScrollBar` lerp.

Things to decide
- Should there be code for a horizontal version?
- Should `SizeFlags.FillExpand` do anything?


ShadowCommander:
This is ready to be reviewed

ShadowCommander:
Should I add a button index callback like ItemList or should that be done as a class that extends this one?

RemieRichards:
Any reason this isn't a virtual list?

ShadowCommander:
What do you mean by virtual list?

RemieRichards:
Say you have a list of 3000 items.
Given the current dimensions of the list control, you can show, say, 12 of them.

Which is the best idea:
1. having 3000 controls ready and waiting to be placed as the list scrolls
2. having 12 controls who's data is cycled out as the list scrolls

Virtual lists are a bit more complicated to write, but they're obviously cheaper memory wise.
We don't have any MVC or anything, so this would really only work for simple string lists, unless you provide some kind of virtual implementable 'fillRow' method or such.

ShadowCommander:
I want to implement a control like that next, but for now this should work as a better alternative to adding ScrollContainers with a VBoxContainer.

This one is implemented in a way that allows the contents to be different sizes, which would probably be different from the virtual list one.

RemieRichards:
You can certainly handle virtual lists with differing-sized rows (I know, I've done it) but it prevents you using some optimising tricks to make it faster, so yeah, definitely.

PJB3005:
How is it much better than a `ScrollContainer` combined with a `VBoxContainer`? Most of the cost from this combination is the *instantiation* of the controls, not the actual scrolling. (which is almost free, actually. Just look at the licenses section of the credits window, there's about 2000 lines of text so 2000-ish controls in that `VBoxContainer` and it handles fine after the initial construction).

This control doesn't seem to actually solve that problem, because as Remie said there's no virtualization.

PR #1167, opened 2020-07-06
Added User Input Validation For Most Debug Commands

SoulSloth:
- Added checks to most debug commands so non-exceptional input won't crash the client, mainly for my own exploratory purposes.
- Refactored some debug commands for clarity/consistency
- Made `getclipboard` put contents of clipboard into debug window


SoulSloth:
What's the preference on determining if a string represents a valid whole number?

```csharp
!new Regex(@"^[0-9]+$").IsMatch(args[0])
//V.S.
!int.TryParse(args[0], out var id)
```

RemieRichards:
Go for readability, `int.TryParse()` is more readable than a regex.
(Also performance, but for something like this, readability is better)
(`int.TryParse()` wins anyway)

SoulSloth:
```
TryParse ticks: 18 TryParse Milliseconds: 0
Regex ticks: 224 Regex Milliseconds: 0
```

Seems so.

RemieRichards:
Don't forget the memory costs from allocating the Regex. (granted, you could store it statically but it's still more memory than try parse would use)

PR #1174, opened 2020-07-08
Adds logout command, tweaks login

ike709:
Adds a logout command, which demotes host/admin to player. Not as intricate as a proper deadmin command should be (booting aghost back to body, removing admin chat filter button, etc aren't dealt with) but it is useful for testing things.

Also adds feedback to successful login and allows you to use the host password for admin login, since I mix them up sometimes.

Required by: space-wizards/space-station-14#1320


ike709:
Build errors appear to be unrelated to this PR.

PR #1180, opened 2020-07-14
Add component message for an entity moving

DrSmugleaf:
For entities to independently clang in https://github.com/space-wizards/space-station-14/pull/1367, instead of needing to do it through the parent (tube).


gradientvera (approved):
LGTM.

PJB3005:
I fail to see why `RelayMove` isn't sufficient for disposals here.

DrSmugleaf:
> I fail to see why `RelayMove` isn't sufficient for disposals here.

Keeping track of when an entity last clanged requires a dictionary of entities to times in the tube or a property in the DisposableComponent, outside of the component that is listening to the `RelayMove` message, which is a little strange when each entity could do both by itself in its DisposableComponent instead. A locker has one global delay for all of its contents for instance, just as a disposal tube does right now, but when keeping track of both things from the tube's component it leads to shorter delays than intended due to the entity travelling through new tubes or longer delays from the entity travelling through a loop of tubes as the delays clear behind it just as it passes.

It is already implemented with RelayMove and the delays can be kept track of separately though, so this pr isn't holding disposals back anyway if you don't deem it necessary.

Acruid (approved):

PR #1181, opened 2020-07-14
Remember Window Size & Pos when switching to/from fullscreen

exp111:


Robust.Client/Graphics/Clyde/Clyde.Windowing.cs
```diff
@@ -183,7 +185,10 @@ private void InitWindow()
             _windowScale = (scaleX, scaleY);
 
             GLFW.GetWindowSize(_glfwWindow, out var w, out var h);
-            _windowSize = (w, h);
+            _prevWindowPos = _windowSize = (w, h);
```
gradientvera: Why are you setting `_prevWindowPos` to this here? Shouldn't you set `_prevWindowSize` instead?

exp111: oh damn.

PR #1188, opened 2020-07-21
Refactor DetachParent

ShadowCommander:
Rename `DetachParent` to `AttachToGridOrMap`.

Implement `ContainerHelpers.AttachParentToContainerOrGrid` which inserts the entity into the next container up, or if there isn't a container it runs `AttachToGridOrMap`.

Added a test for `AttachParentToContainerOrGrid`.

### DIFF
--- a/Robust.Shared/Containers/ContainerHelpers.cs
+++ b/Robust.Shared/Containers/ContainerHelpers.cs
@@ -65,6 +65,18 @@ namespace Robust.Shared.Containers
             return false;
         }
 
+        public static void AttachParentToContainerOrGrid(ITransformComponent transform)
+        {
+            if (transform.Parent != null && TryGetContainer(transform.Parent.Owner, out var container))
+            {
+                container.Insert(transform.Owner);
+            }
+            else
+            {
+                transform.AttachToGridOrMap();
+            }
+        }
+
         private static bool TryGetManagerComp(IEntity entity, [NotNullWhen(true)] out IContainerManager? manager)
         {
             DebugTools.AssertNotNull(entity);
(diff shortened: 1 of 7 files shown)


ProfanedBane:
Will be dumping issues as separate comments as I find them.

ProfanedBane:
If you buckle to a chair in space you're moved to grid 1, unbuckling moves you to grid 0.

ProfanedBane:
Buckling in space moves you to grid 1 as expected, unbuckling however sends you to grid 0 which I imagine is unintended.
The same applies to items, inside a locker in space it's grid 1, opened it's outputted to grid 0.

I should note that buckling into a different grid and unbuckling correctly adds you to that grid.

ShadowCommander:
The Entity Spawn panel spawns entities on Grid 1 when spawning in space, which seems to be a mistake and is a separate issue from this. The grid 0 setting is the correct and expected behavior.

Robust.Shared/Containers/ContainerHelpers.cs
```diff
@@ -65,6 +65,18 @@ public static bool TryGetContainer(IEntity entity, [NotNullWhen(true)] out ICont
             return false;
         }
 
+        public static void AttachParentToContainerOrGrid(ITransformComponent transform)
+        {
+            if (transform.Parent != null && TryGetContainer(transform.Parent.Owner, out var container))
+            {
+                container.Insert(transform.Owner);
```
PJB3005: This needs a (possibly tail recursive) fallback path in case the new container refuses insertion.

gradientvera (approved):

Issue #1191, opened 2020-07-26
Melee Hit Audio Fails To Play On Hit That Destroys Entity


SoulSloth:
Reproduction steps:
1. Destroy an entity with punches or any melee attack
Expected Result: The hit sound plays as normal
Actual: The client fails to find the entity that has been destroyed and logs an error: Server tried to play audio file /Audio/Weapons/genhit1.ogg on entity ... which does not exist.

Acruid:
Cannot reproduce, seems resolved.

PR #1192, opened 2020-07-26
Fix melee audio on entity destruction

SoulSloth:
Fix for #1191.

Adds a coordinate to `PlayAudioEntityMessage` and changes the `play` function on the client audio system to no longer rely on an entity that may be deleted. 

Can't be tested until https://github.com/space-wizards/space-station-14/pull/1494 is complete as the exception it fixes supersedes this issue.


ShadowCommander:
I think it should probably use the coordinates as a fallback if getting the entity failed. If you don't it's possible when chasing someone that the sound of their shooting will come from behind you instead of in front.

SoulSloth:
> I think it should probably use the coordinates as a fallback if getting the entity failed. If you don't it's possible when chasing someone that the sound of their shooting will come from behind you instead of in front.

Yeah sounds right. Forgot that `PlayFromEntity` supports guns too.

PR #1199, opened 2020-07-29
1464 Ignore text functions if no keyboard is focused

### LINKED ISSUE #1464: You cannot hold Ctrl and move left at the same time
Due to Ctrl A being "select all text" that takes priority.

StrawberryMoses:
Fixes https://github.com/space-wizards/space-station-14/issues/1464

If there's no keyboard focused, ignore all commands that have "Text". IE: TextSelectAll which makes control A not moving left.

### DIFF
--- a/Robust.Client/Input/InputManager.cs
+++ b/Robust.Client/Input/InputManager.cs
@@ -131,8 +131,7 @@ namespace Robust.Client.Input
             foreach (var binding in _bindings)
             {
                 // check if our binding is even in the active context
-                if (!Contexts.ActiveContext.FunctionExistsHierarchy(binding.Function)
-                    || (_userInterfaceManagerInternal.KeyboardFocused == null && binding.Function.FunctionName.Contains("Text")))
+                if (!Contexts.ActiveContext.FunctionExistsHierarchy(binding.Function))
                 {
                     continue;
                 }
@@ -153,6 +152,10 @@ namespace Robust.Client.Input
                     {
                         // kill any lower level matches
                         UpBind(binding);
+                    }
+                    else
+                    {
+                        bindsDown.Add(binding);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


Robust.Client/Input/InputManager.cs
```diff
@@ -131,8 +131,11 @@ public void KeyDown(KeyEventArgs args)
             foreach (var binding in _bindings)
             {
                 // check if our binding is even in the active context
-                if (!Contexts.ActiveContext.FunctionExistsHierarchy(binding.Function))
+                if (!Contexts.ActiveContext.FunctionExistsHierarchy(binding.Function)
+                    || (_userInterfaceManagerInternal.KeyboardFocused == null && binding.Function.FunctionName.Contains("Text")))
```
gradientvera: I feel like basing this on the name of the function is not a good idea exactly...

StrawberryMoses: It was mostly just because I'm lazy. All of the functions that appeared to need blocking in normal movement mode (in KeyFunctions.cs) are all prepended with "Text". The other option that I considered was to add specific ones to a list and then check if the list contained the function name.

PJB3005 (changes requested):
Yeah this really isn't a good solution. The proper solution I can think of would be to consider keycombos with subset of the best match (instead of ONLY the best match) if the key handlers aren't getting handled.

StrawberryMoses:
Ok- so I really dunno if this is the correct solution, but it seems to work. It adds to the bindsDown list if it matches (but isn't the combo or a sub pattern), then the foreach on 169 tests if things were handled. Oddly enough it makes it seem like this used to work.

SweptWasTaken:
Someone review this.

metalgearsloth:
Superseded by #1473

PR #1203, opened 2020-07-30
Add EntityUid YAML type serializer

DrSmugleaf:


DrSmugleaf:
Not necessary, already done.

PR #1205, opened 2020-08-01
VV numeric field display attribute

Acruid:
Adds the ability to mark numeric fields an properties with info about how to display the number in the UI (dec, hex, bin). These are the backend changes to make displaying an int as a set of bit flags possible.

collision groups UI


PJB3005:
While this is enough for hex/flags display, this doesn't seem remotely flexible enough to work for collision layers?

Acruid:
> 
> 
> While this is enough for hex/flags display, this doesn't seem remotely flexible enough to work for collision layers?

There is nothing special about collision layers, they are an int like any other one. VV does not have the knowledge that the CollisionGroups enum is even related to the layer/mask int on PhysBodies. This just tells VV that you want the int to be displayed like flags to help the user read it better.

How is this not flexible enough to work in this case?

PJB3005:
Treating the collision groups like bitflags is significantly less useful if you don't have the groups to name them correctly, IMO. Adding support for that would require replacing this API entirely which is why I'm not very fond of it.

PR #1209, opened 2020-08-03
More null entity fixes

ghost:
Additional checks for PVS & containers to fix exceptions from null entities.

### DIFF
--- a/Robust.Client/GameObjects/Components/Containers/ContainerManagerComponent.cs
+++ b/Robust.Client/GameObjects/Components/Containers/ContainerManagerComponent.cs
@@ -144,6 +144,11 @@ namespace Robust.Client.GameObjects.Components.Containers
                 // Add new entities.
                 foreach (var uid in data.ContainedEntities)
                 {
+                    if (!Owner.EntityManager.TryGetEntity(uid, out var ent))
+                    {
+                        continue;
+                    }
+
                     var entity = Owner.EntityManager.GetEntity(uid);
 
                     if (!container.Entities.Contains(entity))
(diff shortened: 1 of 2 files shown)


Robust.Client/GameObjects/Components/Containers/ContainerManagerComponent.cs
```diff
@@ -144,6 +144,11 @@ public override void HandleComponentState(ComponentState? curState, ComponentSta
                 // Add new entities.
                 foreach (var uid in data.ContainedEntities)
                 {
+                    if (!Owner.EntityManager.TryGetEntity(uid, out var ent))
+                    {
+                        continue;
+                    }
+
                     var entity = Owner.EntityManager.GetEntity(uid);
```
RemieRichards:
```suggestion
                    if (!Owner.EntityManager.TryGetEntity(uid, out var entity))
                    {
                        continue;
                    }
```

Robust.Client/GameObjects/Components/Containers/ContainerManagerComponent.cs
```diff
@@ -144,7 +144,10 @@ public override void HandleComponentState(ComponentState? curState, ComponentSta
                 // Add new entities.
                 foreach (var uid in data.ContainedEntities)
                 {
-                    var entity = Owner.EntityManager.GetEntity(uid);
+                    if (!Owner.EntityManager.TryGetEntity(uid, out var entity))
```
PJB3005, line 147: This really should never be possible. I would rather a test case for this get made and the root bug fixed than ignoring the error here.

PJB3005 (changes requested):
None of the cases where you added checks *should* ever be capable of having deleted entities in them. You should fix the reason these deleted entities are in here instead of making PVS resilient against them.

ghost:
This shouldn't be necessary anymore actually since I made singulo not delete entities that are within a container, but that's what was causing it. I think the container was getting deleted and then the items within it were getting deleted afterwards, somehow.

PR #1212, opened 2020-08-06
Added Console Tab Completion

exp111:
Tab for next cmd
Shift-Tab for previous cmd

### DIFF
--- a/Robust.Client/UserInterface/CustomControls/DebugConsole.cs
+++ b/Robust.Client/UserInterface/CustomControls/DebugConsole.cs
@@ -40,7 +40,7 @@ namespace Robust.Client.UserInterface.CustomControls
         private bool _targetVisible;
 
         private bool commandChanged = true;
-        private List<string> searchResults;
+        private readonly List<string> searchResults;
         private int searchIndex = 0;
 
         public DebugConsole(IClientConsole console, IResourceManager resMan)
@@ -253,7 +253,7 @@ namespace Robust.Client.UserInterface.CustomControls
                 if (searchResults.Count == 0)
                     return;
 
-                searchIndex = ++searchIndex % searchResults.Count;
+                searchIndex = (searchIndex + 1) % searchResults.Count;
                 SetInput(searchResults[searchIndex]);
                 return;
             }
@@ -272,9 +272,7 @@ namespace Robust.Client.UserInterface.CustomControls
                 if (searchResults.Count == 0)
                     return;
 
-                searchIndex--;
-                if (searchIndex < 0)
-                    searchIndex += searchResults.Count;
+searchIndex = MathHelper.Mod(searchIndex - 1, searchResults.Count);
                 SetInput(searchResults[searchIndex]);
                 return;
             }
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/CustomControls/DebugConsole.cs
```diff
@@ -204,6 +212,81 @@ private void CommandBarOnOnKeyBindDown(GUIBoundKeyEventArgs args)
...
             }
+            else if (args.Function == EngineKeyFunctions.GuiTabNavigateNext)
+            {
+                NextCommand();
+                args.Handle();
+                return;
+            }
+            else if(args.Function == EngineKeyFunctions.GuiTabNavigatePrev)
+            {
+                PrevCommand();
```
PJB3005, line 223: Why does this one not get `args.Handle()`?

exp111: Because I forgot

Robust.Client/UserInterface/CustomControls/DebugConsole.cs
```diff
@@ -39,6 +39,10 @@ public class DebugConsole : Control, IDebugConsole
 
         private bool _targetVisible;
 
+        private bool commandChanged = true;
+        private List<string> searchResults;
```
PJB3005:
```suggestion
        private readonly List<string> searchResults;
```

Robust.Client/UserInterface/CustomControls/DebugConsole.cs
```diff
@@ -204,6 +212,83 @@ private void CommandBarOnOnKeyBindDown(GUIBoundKeyEventArgs args)
...
+                    searchIndex += searchResults.Count;
+                SetInput(searchResults[searchIndex]);
+                return;
+            }
+
+            FindCommands();
+            if (searchResults.Count == 0)
+                return;
+
+            SetInput(searchResults[searchResults.Count - 1]);
```
PJB3005: You can use C# 8 `Index`es.

```suggestion
            SetInput(searchResults[^1]);
```

Robust.Client/UserInterface/CustomControls/DebugConsole.cs
```diff
@@ -204,6 +212,83 @@ private void CommandBarOnOnKeyBindDown(GUIBoundKeyEventArgs args)
...
+        private void PrevCommand()
+        {
+            if (!commandChanged)
+            {
+                if (searchResults.Count == 0)
+                    return;
+
+                searchIndex--;
+                if (searchIndex < 0)
+                    searchIndex += searchResults.Count;
```
PJB3005: You can use `MathHelper.Mod` here:

```suggestion
searchIndex = MathHelper.Mod(searchIndex - 1, searchResults.Count);
```

Robust.Client/UserInterface/CustomControls/DebugConsole.cs
```diff
@@ -204,6 +212,83 @@ private void CommandBarOnOnKeyBindDown(GUIBoundKeyEventArgs args)
...
+        }
+
+        private void NextCommand()
+        {
+            if (!commandChanged)
+            {
+                if (searchResults.Count == 0)
+                    return;
+
+                searchIndex = ++searchIndex % searchResults.Count;
```
PJB3005: Don't use `++` here since you're not making use of the actual mutate it does.
```suggestion
                searchIndex = (searchIndex + 1) % searchResults.Count;
```

PJB3005 (commented):
Minor tweaks I'll just commit right now.

It would be cool if we could extend this to other commands in the future.

PR #1220, opened 2020-08-11
Bug/placement not blocking mouseclicks

### LINKED ISSUE #1369: While placing construction ghosts your clicks are still being registered by whatever object they hit
So if you try to place a ghost over a chair you will buckle to it etc.

Repro
- Get a chair/locker or other interactable
- Stand in range and place a ghost over it
- You will sit in the chair/open the locker etc

### LINKED ISSUE #1562: In Erase/Place Mode you will still do actions with the item in your hand
1. Equip a crowbar/weapon
2. Place/Erase a thing
3. You will crowbar the floor tile/attack with your weapon (given you're in combat mode)

juliangiebel:
fixes space-wizards/space-station-14#1562
fixes space-wizards/space-station-14#1369

Solved the problem by blocking all key bindings for `MouseLeft` except for `EngineKeyFunctions.EditorPlaceObject` when placing or erasing

### DIFF
--- a/Robust.Client/Input/InputManager.cs
+++ b/Robust.Client/Input/InputManager.cs
@@ -113,6 +113,15 @@ namespace Robust.Client.Input
         }
 
         /// <inheritdoc />
+        public void ToggleBindingsForKey(Key key, bool enabled)
+        {
+            foreach (var binding in GetKeyBindingsForKey(key))
+            {
+                binding.Enabled = enabled;
+            }
+        }
+
+        /// <inheritdoc />
         public void KeyDown(KeyEventArgs args)
         {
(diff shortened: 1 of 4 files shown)


PJB3005 (changes requested):
I'm actually 99% sure that the reason this is even a bug is because the placement keybinds have lower priority than `Use` so world interaction is always done earlier. Fixing that would is better. (this has to be done in content `keybinds.yml`)

The disabling system in this PR is already the job of the input context system, I believe, so I think it's not necessary.

juliangiebel:
Moving the editor inputs above use doesn't seem to fix the issue.
The issue get's fixed by this PR #1222 from ShadowCommander, so you can close this.

PR #1221, opened 2020-08-12
Reliable Coordinates

Acruid:
A new ReliableCoordinates struct that tries as hard as it can to resolve it's location. It provides sanity checks on it's fields when resolving the map or grid position, so no more NaN/infinity propagation, and verifies the grid/map actually exists. This is really useful for sending over the network, because locations that cannot be resolved on the server when a grid is deleted are gracefully dropped. TransformComponent has a property to generate ReliableCoordinates for an entity.

As part of the sanitization, this adds an IsFinite() float helper, and an IsFinite() function on Vector2. Of course we don't check these on incoming network messages, so hopefully these new functions will be useful in the future.

### DIFF
--- a/Robust.Shared.Maths/FloatHelpers.cs
+++ b/Robust.Shared.Maths/FloatHelpers.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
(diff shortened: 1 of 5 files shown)


Acruid:
Superseded by #1273

PR #1248, opened 2020-08-21
Dispose of MIDI renderers on the MIDI thread

gradientvera:
Seems to work fine.


Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -558,6 +563,12 @@ public void Dispose()
                     CloseMidi();
                     break;
             }
+        }
+
+        /// <inheritdoc />
+        void IMidiRenderer.InternalDispose()
+        {
+            Disposed = true;
```
PJB3005: This is already true, isn't it?

gradientvera: Yeah, I guess we don't need to set it to true again

PR #1258, opened 2020-08-25
Refactor Server Entity Manager

Visne:
Mostly `UpdatePlayerSeenEntityStates()`.
Some stuff definitely still needs to be changed but a preliminary review would be nice if possible.

- [ ] Are `_seenMovers` and `_seenMovers2` somehow related? If so, can they be merged somehow? If not, what should the names be changed to?
- [ ] Names for new methods (currently named ProcessX)?
- [ ] Removed some unused methods, keep them removed or add back?
- [ ] Lots of null-forgiving. Okay or change to check at start of method? In my opinion it's fine as the methods won't be used for anything else and are mainly for readability.
- [ ] Duplicate code? Not sure if its duplicate but lots of code in the ProcessX methods look very similar. Not sure if it can be avoided.
- [ ] Some comments by Q can be made more clear, but I'm not sure about them as I'm not very familiar with PVS.
- [ ] Some variable names by Q can be made more clear, same reasoning as above.

Allow edits is turned on, so feel free to add commits.

### DIFF
--- a/Robust.Server/GameObjects/ServerEntityManager.cs
+++ b/Robust.Server/GameObjects/ServerEntityManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using Prometheus;
 using Robust.Server.GameObjects.Components;
@@ -18,6 +19,7 @@ using Robust.Shared.Map;
 using Robust.Shared.Maths;
 using Robust.Shared.Timing;
 using Robust.Shared.Utility;
+using static Robust.Shared.GameObjects.Components.Transform.TransformComponent;
 
 namespace Robust.Server.GameObjects
 {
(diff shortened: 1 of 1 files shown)


Nukeuler123:
Helping out with this

metalgearsloth:
FWIW when I profiled CullDeletedEntities and UpdatePlayerSeenStates where the 2 most expensive functions.

metalgearsloth:
I may end up rewriting this because I'm trying to increase UpdateEntityTree performance to make moving shuttables decently performant and its job is primarily to feed into this.

(Not to deliberately step on your toes, sorry).

Visne:
> 
> 
> I may end up rewriting this because I'm trying to increase UpdateEntityTree performance to make moving shuttables decently performant and its job is primarily to feed into this.
> 
> (Not to deliberately step on your toes, sorry).

Go wild, this refactor is just in the way for stuff I actually want to work on, so the sooner this is done the better, and I don't really care how

Visne:
Closing in favor of #1478

PR #1263, opened 2020-08-27
Fix Button Labels not updating styles when Button is disabled/enabled

ShadowCommander:


Robust.Client/UserInterface/Controls/Button.cs
```diff
@@ -20,6 +20,13 @@ public Button() : base()
             AddChild(Label);
         }
 
+        protected override void StylePropertiesChanged()
+        {
+            base.StylePropertiesChanged();
+
+            Label.Restyle();
```
PJB3005, line 30:
```suggestion
			// Temporary workaround to fix button styles in content thanks to a lack of correct style updating for child selectors.
			// The changing of the style class on the parent control (button in this case) doesn't cause a style update in the label correctly currently.
			// This works around that.
            Label.Restyle();
```

PR #1271, opened 2020-09-01
Make HighestMapID and HighestGridID public for content to read

DrSmugleaf:
Needed for https://github.com/space-wizards/space-station-14/pull/1998


PJB3005 (changes requested):
I would rather if this was allocated similar to OpenGL where there is simply a method to return an unused ID and bump the number internally.

DrSmugleaf:
Reviews addressed.

gradientvera (approved):

PJB3005 (approved):

PR #1274, opened 2020-09-02
More configurable lighting (also, "Rounded" mode for Range)

20kdc:
This makes lighting resolution more configurable, switching the backend config option to being based on the amount the screen size is divided by, switching the frontend to using a slider with 3 places (/8, /2, and /1), and adding a "Soft Shadows" option.

In practice the Soft Shadows option does a lot by itself as PJB predicted, but disabling it seems to sometimes have worse effects on quality than lowering lighting resolution to /8, which gives a similar speed benefit.

Compare these 3 screenshots to get what I mean by this (options visible in the screenshot, taken under GLES2):
image
image
image

### DIFF
--- a/Robust.Client/UserInterface/CustomControls/OptionsMenu.cs
+++ b/Robust.Client/UserInterface/CustomControls/OptionsMenu.cs
@@ -10,8 +10,7 @@ namespace Robust.Client.UserInterface.CustomControls
         private readonly Button ApplyButton;
         private readonly CheckBox VSyncCheckBox;
         private readonly CheckBox FullscreenCheckBox;
-        private readonly Slider LightingQualitySlider;
-        private readonly CheckBox SoftShadowsCheckBox;
+        private readonly OptionButton LightingPresetOption;
         private readonly IConfigurationManager configManager;
 
         protected override Vector2? CustomSize => (180, 160);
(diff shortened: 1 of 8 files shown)


20kdc:
hilariously the big lighting differences between soft shadows and non-soft-shadows are actually caused by derivatives-related hilarity

ike709:
Maybe we should just have Low/Medium/High presets for simplicity?

RemieRichards:
While I love the concept of a lighting resolution slider, it's such a coder-ism (which is why I love it)
We should advocate for better UX :P, so +1 to Low/Med/High presets.

20kdc:
Until the exact performance implications of soft shadows + lowres vs. non-soft-shadows + lowres are known and profiled I'm not 100% sure that's a good idea

20kdc:
state of the presets:
Very Low: /8 + no soft shadows
Low: /2 + no soft shadows
Medium: /2 + soft shadows
High: /1 + soft shadows
that should cover every reasonable case (soft shadows don't really work at /8 anyway)

Robust.Client/UserInterface/CustomControls/OptionsMenu.cs
```diff
@@ -65,15 +71,69 @@ private void OnCheckBoxToggled(BaseButton.ButtonToggledEventArgs args)
...
             var isFullscreenSame = FullscreenCheckBox.Pressed == ConfigIsFullscreen;
-            ApplyButton.Disabled = isVSyncSame && isHighResLightsSame && isFullscreenSame;
+            var isLightingQualitySame = LightingPresetOption.SelectedId == ConfigLightingQuality;
+            ApplyButton.Disabled = isVSyncSame && isFullscreenSame && isLightingQualitySame;
         }
 
         private bool ConfigIsFullscreen =>
             configManager.GetCVar<int>("display.windowmode") == (int) WindowMode.Fullscreen;
+
+        private int ConfigLightingQuality
```
PJB3005: Don't make this a property and just make it two separate methods.

20kdc:
so, just to be clear, I dealt with the configlightingquality property thing

Robust.Client/UserInterface/CustomControls/OptionsMenu.cs
```diff
@@ -65,15 +71,66 @@ private void OnCheckBoxToggled(BaseButton.ButtonToggledEventArgs args)
...
+                return 1;
+            }
+            else if (val >= 2)
+            {
+                return 2;
+            }
+            else
+            {
+                return 3;
+            }
```
gradientvera, line 110: You could probably replace this with a switch with `when` keywords

Robust.Client/UserInterface/CustomControls/OptionsMenu.cs
```diff
@@ -65,15 +71,66 @@ private void OnCheckBoxToggled(BaseButton.ButtonToggledEventArgs args)
...
+            else if (value == 2)
+            {
+                configManager.SetCVar("display.lightmapdivider", 2);
+                configManager.SetCVar("display.softshadows", true);
+            }
+            else
+            {
+                configManager.SetCVar("display.lightmapdivider", 1);
+                configManager.SetCVar("display.softshadows", true);
+            }
```
gradientvera: This could also be replaced with a switch

gradientvera (approved):
Seems fine to me. I only have two minor nitpicks.

Robust.Client/UserInterface/CustomControls/OptionsMenu.cs
20kdc: Not sure about the 'when' here, but the other side can indeed be converted to a switch

Robust.Client/UserInterface/CustomControls/OptionsMenu.cs
20kdc: Done.

PR #1277, opened 2020-09-03
EntityCoordinates

gradientvera:
Supersedes #1273
Polishes `EntityCoordinates` a bit, adds some useful methods to it, unit tests and makes Transform have `EntityCoordinates`
I'd like to get this merged as soon as possible so we can start moving stuff over.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -294,16 +294,6 @@ namespace Robust.Shared.GameObjects.Components.Transform
                     var newEntity = _entityManager.GetEntity(value.EntityId);
                     AttachParent(newEntity);
                 }
-                else
-                {
-                    var worldCoords = value.ToMapPos(_entityManager);
-
-                    var newPos = Parent!.InvWorldMatrix.Transform(worldCoords);
-
-                    // float rounding error guard, if the offset is less than 1mm ignore it
-                    if ((newPos - _localPosition).LengthSquared < 10.0E-3)
-                        return;
-                }
 
                 _localPosition = value.Position;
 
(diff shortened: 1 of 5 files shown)


Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -281,6 +281,46 @@ public Vector2 WorldPosition
             }
         }
 
+        [ViewVariables(VVAccess.ReadWrite)]
+        public EntityCoordinates Coordinates
+        {
+            get => new EntityCoordinates(_parent.IsValid() ? _parent : Owner.Uid, LocalPosition);
```
PJB3005: This is not correct. If we have to return a set of coordinates for entities without parents then it should be either:

1. Our own entity with 0 offset
2. A null entity.

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -281,6 +281,46 @@ public Vector2 WorldPosition
...
+                else
+                {
+                    var worldCoords = value.ToMapPos(_entityManager);
+
+                    var newPos = Parent!.InvWorldMatrix.Transform(worldCoords);
+
+                    // float rounding error guard, if the offset is less than 1mm ignore it
+                    if ((newPos - _localPosition).LengthSquared < 10.0E-3)
+                        return;
+                }
```
PJB3005: This entire `else` block:
1. isn't even doing anything
2. doesn't have a purpose now that entity coordinates correctly bakes in an exact position.

PJB3005 (changes requested):

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
gradientvera: 1 is what I was going for with this, I thought this was the correct way to go about it. What should I do?

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
gradientvera: Removed now

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
gradientvera: Nevermind, I figured out what you meant. Made it so it returns its own entity with 0 offset now

PR #1281, opened 2020-09-05
Grid tile lookup system

metalgearsloth:
To be used for tile-based systems such as atmos and pathfinding (once I refactor it).

Entity intersecting tiles are cached as they move so the lookups are faster than creating and querying dynamic trees.

The current IEntityManager uses the entity's MapCoordinates for GetEntitiesIntersecting but I used the bounds here as it seemed more appropriate.


gradientvera (approved):
:pray:

PR #1283, opened 2020-09-08
Toggle function in BoundUserInterface

luiderek:
Something about doing OOP properly and recycling code. If spamming opens in the base game doesn't break things, then this shouldn't either? Worst case scenario the list of _subscribedsessions is longer than it oughta, not entirely how to debug and check this isn't what's happening.

### DIFF
--- a/Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
+++ b/Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
@@ -173,39 +173,14 @@ namespace Robust.Server.GameObjects.Components.UserInterface
         /// <exception cref="ArgumentNullException">Thrown if <see cref="session"/> is null.</exception>
         public void Toggle(IPlayerSession session)
         {
-            if (session == null)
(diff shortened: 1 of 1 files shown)


Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
```diff
@@ -162,6 +162,54 @@ public void SetState(BoundUserInterfaceState state, IPlayerSession? session = nu
...
+
+            else
+            {
+                var msg = new CloseBoundInterfaceMessage();
+                SendMessage(msg, session);
+                CloseShared(session);
+            }
+
+            session.PlayerStatusChanged += OnSessionOnPlayerStatusChanged;
+        }
```
PJB3005: To reduce duplication, why not just do:

```suggestion
        public void Toggle(IPlayerSession session)
        {
            if (_subscribedSessions.Contains(session))
            {
                Close(session);
            }
            else
            {
                Open(session);
            }
        }
```

luiderek:
@PJB3005 because that would've been the smart thing to do.

PR #1284, opened 2020-09-09
Make more fields VV and editable

DTanxxx:
Added some more VV and VVAccess.ReadWrite. However some may be inappropriate and may need to be changed back.

### DIFF
--- a/Robust.Client/GameObjects/Components/Appearance/AppearanceComponent.cs
+++ b/Robust.Client/GameObjects/Components/Appearance/AppearanceComponent.cs
@@ -24,7 +24,6 @@ namespace Robust.Client.GameObjects
 
         [Dependency] private readonly IReflectionManager _reflectionManager = default!;
 
-        [ViewVariables]
         private static bool _didRegisterSerializer;
 
         [ViewVariables]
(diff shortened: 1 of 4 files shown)


DTanxxx:
This is for issue: [https://github.com/space-wizards/space-station-14/issues/2020]

Robust.Client/GameObjects/Components/Appearance/AppearanceComponent.cs
```diff
@@ -24,8 +24,10 @@ public sealed class AppearanceComponent : SharedAppearanceComponent
 
         [Dependency] private readonly IReflectionManager _reflectionManager = default!;
 
+        [ViewVariables]
```
exp111: This doesn't appear ingame, probably because it's static
grafik

PJB3005: Yes this should probably be removed.

Robust.Client/Graphics/RSI/RSI.cs
```diff
@@ -18,7 +18,7 @@ public sealed partial class RSI : IEnumerable<RSI.State>
         /// <summary>
         ///     The size of this RSI, width x height.
         /// </summary>
-        [ViewVariables]
+        [ViewVariables(VVAccess.ReadWrite)]
```
PJB3005: You really should not make this writable with VV. You'll only break things.

If you *really* need to do advanced hacks like this I recommend you use CSI instead.

DTanxxx:
Hi PJB3005,
Does my 'Applied feedback' commit look like what you were looking for? I have removed the VV and ReadWrite for the two variables. Please let me know if there are any more variables that should be made VV or ReadWrite.

PR #1285, opened 2020-09-10
Purge localization manager

DTanxxx:
Removed localization manager dependency from EntitySpawnWindow

### DIFF
--- a/Robust.Client/UserInterface/CustomControls/EntitySpawnWindow.cs
+++ b/Robust.Client/UserInterface/CustomControls/EntitySpawnWindow.cs
@@ -20,7 +20,6 @@ namespace Robust.Client.UserInterface.CustomControls
         private readonly IPlacementManager placementManager;
         private readonly IPrototypeManager prototypeManager;
         private readonly IResourceCache resourceCache;
-        private readonly ILocalizationManager _loc;
 
         private VBoxContainer MainVBox;
         private PrototypeListContainer PrototypeList;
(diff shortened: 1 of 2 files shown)


exp111:
Seems like the commits got mixed up.
Are there more _loc usages in the engine you can purge now?

DTanxxx:
I think in the engine there is only EntitySpawnWindow that uses _loc. Would you like me to add LoadCulture method to Loc so that I can use Loc.LoadCulture in the main project solution?

exp111:
Yeah you could do that

DTanxxx:
Just a quick question, in order for the build test to pass, I don’t think I can just remove the _loc parameter from method since wouldn’t that break the main project solution where _loc is actually passed in as an argument? Is there a way to simultaneously commit changes in both projects in one go when they are separate projects?

PJB3005:
Make a new overload of the constructor that doesn't take in a localization manager, flag the old one with `[Obsolete]`, remove it some time later. Way easier than doing simultaneous updates like that.

DTanxxx:
Hi PJB3005,
Does my 'Applied feedback' commit look like what you were looking for? I added [Obsolete] to the old constructor and also added a new one without _loc parameter. However I couldn't find a way to remove the old constructor yet so I might need to wait until this commit becomes merged.

PJB3005 (changes requested):
Don't duplicate the code like that. Have the obsolete constructor just call the new constructor instead.

PJB3005 (approved):

PR #1292, opened 2020-09-19
Improve fallback soundfont and remove aliased patterns

20kdc:
...This is one of those PRs where you're basically going to have to hear the difference for yourself.
NOTE: This PR is *NOT* tested in SS14 directly. Testing was performed on Fluidsynth outside of SS14.
My defense is I don't know a sane way to force the fallback soundfont to load.
To be clear, this does NOT increase the fallback soundfont's size. It actually decreases it because I removed a bunch of presets which will be auto-aliased to Piano anyway. (This is something of a reliance on Fluidsynth, but that's fine.)


gradientvera (approved):

20kdc:
Don't merge, just noticed I somehow managed to detune everything by 1 semitone or so during final adjustments. Apparently my reference pitch was wrong.

20kdc:
Should be fine now, I think.

gradientvera (approved):
Tested locally, is gud.

PR #1295, opened 2020-09-27
Removed obsolete EntitySpawnWindow constructor

DTanxxx:
To wrap up the 'purge localization manager dependency' issue

### DIFF
--- a/Robust.Client/UserInterface/CustomControls/EntitySpawnWindow.cs
+++ b/Robust.Client/UserInterface/CustomControls/EntitySpawnWindow.cs
@@ -60,14 +60,6 @@ namespace Robust.Client.UserInterface.CustomControls
 
         protected override Vector2? CustomSize => (250, 300);
 
-        [Obsolete]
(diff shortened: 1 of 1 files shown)


DTanxxx:
Hi, this PR is a wrap up to the issue on removing localization manager dependencies (specifically, removing the obsolete constructor for EntitySpawnWindow). However during the build test, there seems to be some issue with WatchdogApi and StatusHost's `IConfigurationManager.RegisterCVar` usage. 

The following log message was shown:
`ServerStatus/WatchdogApi.cs(45,13): error CS8714: The type 'string?' cannot be used as type parameter 'T' in the generic type or method 'IConfigurationManager.RegisterCVar<T>(string, T, CVar, Action<T>?)'. Nullability of type argument 'string?' doesn't match 'notnull' constraint.`
A similar message was shown for StatusHost.

I have updated my local repo with upstream repo, as well as the NetSerializer, and I haven't touched WatchdogApi or StatusHost files. Does anyone know how to resolve this issue? Any help is appreciated!

PR #1299, opened 2020-09-30
Add YAML serialization for nullable primitive types

DrSmugleaf:
Otherwise it believes its a struct and fails the YamlMappingNode cast.

Needed for https://github.com/space-wizards/space-station-14/pull/2074

### DIFF
--- a/Robust.Shared/Serialization/YamlObjectSerializer.cs
+++ b/Robust.Shared/Serialization/YamlObjectSerializer.cs
@@ -441,24 +441,16 @@ namespace Robust.Shared.Serialization
 
         public object NodeToType(Type type, YamlNode node)
         {
+            var underlyingType = Nullable.GetUnderlyingType(type) ?? type;
+
             // special snowflake string
             if (type == typeof(String))
                 return node.ToString();
 
             // val primitives
-            if (type.IsPrimitive || type == typeof(decimal))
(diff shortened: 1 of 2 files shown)


Robust.Shared/Serialization/YamlObjectSerializer.cs
```diff
@@ -838,6 +841,13 @@ internal static bool IsSerializedEqual(object? a, object? b)
 
         private static object StringToType(Type type, string str)
         {
+            // TypeConverter.ConvertFromInvariantString uses the underlying type of a nullable
+            // type for conversion, causing an exception when nullable primitives are used
+            if (str.Equals("null", StringComparison.InvariantCultureIgnoreCase))
```
PJB3005: So if I have a string with the exact contents `"null"`...

Honestly I would use type tags to encode "really really null" for this instead.

PJB3005 (changes requested):

Robust.Shared/Serialization/YamlObjectSerializer.cs
DrSmugleaf: As it turns out I didn't even need that check at all, it already handles nulls just fine. Oops.

DrSmugleaf:
Reviews addressed.

PJB3005 (approved):

PR #1303, opened 2020-10-03
Add component references for SpriteComponent

DrSmugleaf:
Required for https://github.com/space-wizards/space-station-14/pull/2074

### DIFF
--- a/Robust.Client/GameObjects/ClientComponentFactory.cs
+++ b/Robust.Client/GameObjects/ClientComponentFactory.cs
@@ -39,9 +39,6 @@ namespace Robust.Client.GameObjects
 
             Register<InputComponent>();
 
-            Register<SpriteComponent>();
-            RegisterReference<SpriteComponent, ISpriteComponent>();
-
             Register<ClientOccluderComponent>();
             RegisterReference<ClientOccluderComponent, OccluderComponent>();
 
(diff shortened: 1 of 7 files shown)


PJB3005 (commented):
This doesn't actually work. `[ComponentReference]` is only evaluated for components that also have `[RegisterComponent]`, which these don't (they're still registered via the component factory constructor).

PJB3005 (commented):
The engine currently does not actually run component auto-registrations itself so I would prefer if we kept engine components manually registered for now. That or make the engine properly use autoreg.

DrSmugleaf:
> The engine currently does not actually run component auto-registrations itself so I would prefer if we kept engine components manually registered for now. That or make the engine properly use autoreg.

I fixed it for the failing test with https://github.com/space-wizards/RobustToolbox/pull/1303/commits/6d6b9dedd06db4aa91caefd06ec27821846813af, should it be fixed in another way? Or should I just add the server ISpriteRenderableComponent component reference through EntryPoint as it was done for the client and ISpriteComponent before?

PJB3005:
Yes just go back to the manual `Register<>()` calls in the constructor of the component factory like before.

DrSmugleaf:
> Yes just go back to the manual `Register<>()` calls in the constructor of the component factory like before.

Done.

PR #1305, opened 2020-10-04
Scale frame debug into control boundaries

janekvap-havok:
Before:
Robust Client_8gScD5igGi
After:
image


PJB3005 (commented):
Instead of storing the calculated heights and colors in an array like that, wouldn't it be much easier to just have a simple for loop that ONLY calculates the max height, then do all the rest of the work including scaling and color checking in another one?

Calculating height is a single multiplication so it's not like this is particularly expensive.

PJB3005 (approved):

PR #1306, opened 2020-10-05
Things needed for singulo

PaulRitter:


Robust.Client/UserInterface/Controls/SpinBox.cs
```diff
@@ -29,9 +29,26 @@ public int Value
...
+        public void OverrideValue(int value)
+        {
+            if (IsValid != null && !IsValid(value))
+            {
+                return;
+            }
+            _lineEdit.Text = value.ToString();
+        }
+
+        public event EventHandler<ValueChangedEventArgs> ValueChanged = null!;
```
PJB3005: Don't `= null!` this. Define the event as type `EventHandler<ValueChangedEventArgs>?` instead, because that's what it actually is. (events are null if nothing is registered to them).

PJB3005 (changes requested):

Robust.Client/UserInterface/Controls/SpinBox.cs
PaulRitter: you're right. i've worked with events before, just not when using the nullable annotation.

Robust.Client/UserInterface/Controls/SpinBox.cs
```diff
@@ -97,6 +114,32 @@ public void SetButtons(List<int> leftButtons, List<int> rightButtons)
             }
         }
 
+        /// <summary>
+        /// Changes the editability of the lineedit-field
+        /// </summary>
+        /// <param name="disabled"></param>
+        public void SetLineEditDisabled(bool disabled)
```
PJB3005: Make this a property instead?

Robust.Server/Placement/PlacementManager.cs
```diff
@@ -152,7 +152,13 @@ private void PlaceNewTile(ushort tileType, MapId mapId, Vector2 position)
             if (closest != null) // stick to existing grid
             {
                 // round to nearest cardinal dir
-                var normal = new Angle(position - intersect.Center).GetCardinalDir().ToVec();
+                var deltaVec = position - intersect.Center;
+                var normal = new Vector2(0,0);
+                if (deltaVec != Vector2.Zero)
+                {
+                    normal = new Angle(deltaVec).GetCardinalDir().ToVec();
```
PJB3005, line 159: What's this do?

PaulRitter: a fix to a runtime introduced by deltaVec being zero and then getting the cardinal dir `-123124124124`. you get it, some arbitrary high number, which then throws due to the tovec() loopup I introduced

it will literally fail every time without this fix. as far as i can tell it works, my assumption is that deltaVec was always 0,0 and it just went unnoticed.

PJB3005 (approved):

PR #1307, opened 2020-10-07
Physics caching go brrt

metalgearsloth:
Hope I didn't leave any grid 0 bugs in here...

Benefits from:
1. No more ToList() being called for SimulateWorld which is like 10% perf (a little bit of that was GC pressure).
2. Awake bodies being cached. This was like 25%.
3. There was like 5 iterations over all collidables which was reduced to have less duplication (ProcessQueue).
4. Deferred some Transform stuff which was another 25% or so.

Didn't notice any bugs when playtesting in game... there's room for more improvement though idle physics is under power system now.

Old client allocations:
old_client_allocs

New client allocations:
new_client_allocs

Old client perf:
old_client_physics

New client perf:
new_client_physics

~double framerate on my laptop conservatively

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -73,17 +73,14 @@ namespace Robust.Shared.GameObjects.Systems
         private void ProcessQueue()
         {
             // At this stage only the dynamictree cares about asleep bodies
-            
+            // Implicitly awake bodies so don't need to check .Awake again
+            // Controllers should wake their body up (inside)
             foreach (var collidable in _queuedUpdates)
             {
-                if (collidable.Awake)
-                {
-                    if (collidable.Predict)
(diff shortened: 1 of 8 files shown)


Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -24,28 +23,95 @@ public abstract class SharedPhysicsSystem : EntitySystem
...
         {
-            _awakeBodies.Clear();
+            base.Initialize();
+            SubscribeLocalEvent<CollidableUpdateMessage>(HandleCollidableUpdateMessage);
+        }
 
-            foreach (var body in physicsComponents)
+        private void HandleCollidableUpdateMessage(CollidableUpdateMessage message)
+        {
+            if (message.Component.Deleted || !message.Component.Awake)
```
janekvap-havok, line 60: It seems that multiple messages can come for single component before ProcessQueue is called? This would cause inconsistency so there should be single List<(ICollidableComponent, bool deleted)>

PR #1311, opened 2020-10-09
Componentdependency

PaulRitter:
ToDo:
- [x] write unit tests
- [x] fix retrieverdelegate
- [X] ~~restrict attribute to be only allowed on fields who are subclasses of component so the linter complains (?)~~ just dont be a dumbdumb
- [x] consider componentreference attribute for dependencies (?)


Robust.Shared/GameObjects/ComponentDependencies/ComponentDependencyManager.cs
```diff
@@ -0,0 +1,249 @@
...
+        /// Filling the dependencies of newComp by iterating over entity's components
+        /// </summary>
+        /// <param name="entity"></param>
+        /// <param name="newComp"></param>
+        public void InjectIntoComponent(IEntity entity, IComponent newComp)
+        {
+            var compType = newComp.GetType();
+
+            //get all present components in entity
+            Dictionary<Type, object> entityComponents = new Dictionary<Type, object>();
```
PJB3005: Making this copy dictionary is *extremely* wasteful and expensive, especially because this method gets called any time a component gets added to an entity.

Why not just `TryGetComponent` in the loop on line 69 where you're iterating the types that need to be injected?

Robust.Shared/GameObjects/ComponentDependencies/ComponentDependencyManager.cs
```diff
@@ -0,0 +1,249 @@
...
+
+        /// <inheritdoc />
+        public void OnComponentRemove(IEntity entity, IComponent removedComp)
+        {
+            ClearRemovedComponentDependencies(removedComp);
+            SetDependencyForEntityComponents(entity, removedComp.GetType(), null);
+        }
+
+        /// <summary>
+        /// Sets all dependencies onto compType's references, which are present in entity, to the value of comp
```
PJB3005:
```suggestion
        /// Updates all dependencies to type compType on the entity (in fields on components), to the value of comp
```
Slightly more clear what this does since this took me 30 seconds of reading to figure out.

Robust.Shared/Utility/NullableHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+            var nullableAttribute = field.GetCustomAttribute(assemblyNullableEntry.Item1);
+            if (nullableAttribute == null)
+            {
+                return new byte[]{};
+            }
+
+            return assemblyNullableEntry.Item2!(nullableAttribute);
+        }
+
+        private static void CacheGetNullableFlag(Assembly assembly)
```
PJB3005: This seems really overengineered when you can just use `FieldInfo.GetValue()` for this instead. It's not a critical code path and reflection isn't THAT slow.

PJB3005 (changes requested):
Definitely gonna want to look into ref abuse stuff to make these assignments less painful.

Robust.Shared/GameObjects/ComponentDependencies/ComponentDependencyManager.cs
```diff
@@ -0,0 +1,197 @@
...
+namespace Robust.Shared.GameObjects.ComponentDependencies
+{
+    public class ComponentDependencyManager : IComponentDependencyManager
+    {
+        [IoC.Dependency] private readonly IComponentFactory _componentFactory = null!;
+
+        /// <summary>
+        /// Cache of queries and their corresponding field offsets
+        /// </summary>
+        private readonly Dictionary<Type, (Type, int)[]> _componentDependencyQueries = new Dictionary<Type, (Type, int)[]>();
```
RemieRichards: Please name your tuple fields. None of this `.Item1` `.Item2` nonsense.

```cs
(SomeType someName, SomeType2 someName2, ...)
```

Robust.Shared/Utility/NullableHelper.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// Checks if the field has a nullable annotation [?]
+        /// </summary>
+        /// <param name="field"></param>
+        /// <returns></returns>
+        public static bool IsMarkedAsNullable(FieldInfo field)
+        {
+            if (Nullable.GetUnderlyingType(field.FieldType) != null) return true;
+
+            var flags = GetNullableFlags(field);
+            if (flags.Length != 0 && flags[0] == 2) return true;
```
RemieRichards: What is this magical `2` you speak of? Make a named constant please.

Robust.Shared/GameObjects/ComponentDependencies/ComponentDependencyManager.cs
```diff
@@ -0,0 +1,197 @@
...
+            //create the return array
+            generator.Emit(OpCodes.Ldc_I4, attributeFieldsLength);
+            generator.Emit(OpCodes.Newarr, typeof(int));
+
+            int i = 0;
+            foreach (var field in attributeFields)
+            {
+                if (!NullableHelper.IsMarkedAsNullable(field))
+                {
+                    throw new Exception($"Field {field} of Type {objType} is marked as ComponentDependecy, but does not have ?(Nullable)-Flag!");
```
RemieRichards: Dependecy isn't a word :P

Robust.Shared/Utility/NullableHelper.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Reflection.Emit;
+using System.Runtime.CompilerServices;
+using Mono.CompilerServices.SymbolWriter;
+using Robust.Shared.Log;
+
+namespace Robust.Shared.Utility
+{
+    public static class NullableHelper
+    {
+        private static Dictionary<Assembly, (Type, FieldInfo)> _nullableAttributeTypeCache = new Dictionary<Assembly, (Type, FieldInfo)>();
```
RemieRichards: Same as before, name your tuple fields.

Robust.Shared/GameObjects/ComponentDependencies/ComponentDependencyManager.cs
PaulRitter: I CAN NAME TUPLE FIELDS?!?!?!?!

Robust.Shared/Utility/NullableHelper.cs
PaulRitter: https://github.com/dotnet/roslyn/blob/master/docs/features/nullable-metadata.md

Robust.Shared/GameObjects/ComponentDependencies/ComponentDependencyManager.cs
PaulRitter: 😬

Robust.Shared/Utility/NullableHelper.cs
PaulRitter: `Each type reference in metadata may have an associated NullableAttribute with a byte[] where each byte represents nullability: 0 for oblivious, 1 for not annotated, and 2 for annotated.`
where would i create constants for this?

RemieRichards: Yes, *I'm* aware of what it means, it's just not clear to an outside reader.
if you had a const int called say, `AnnotatedNullableFlag`:

```cs
if (flags.Length != 0 && flags[0] == 2) return true;

//vs

if (flags.Length != 0 && flags[0] == AnnotatedNullableFlag) return true;

```

One of these is much more readable!

RemieRichards: @PaulRitter you're just using 2, so realistically you'd only have to define 2 (the others, 0 and 1 would complain about never being used, which is lame)

Since they're only used in this file (as far as I can see?) just stick it at the top of the class.

`public static int AnnotatedNullableFlag = 2;`

Robust.Shared/GameObjects/ComponentDependencies/ComponentDependencyManager.cs
RemieRichards: yep. very useful feature.

Robust.Shared/GameObjects/ComponentDependencies/ComponentDependencyManager.cs
```diff
@@ -0,0 +1,178 @@
...
+                generator.Emit(OpCodes.Ldflda, fieldOffsetField);
+
+                //calculating offset
+                generator.Emit(OpCodes.Sub);
+
+                //return offset
+                generator.Emit(OpCodes.Ret);
+
+                var @delegate = (Func<object, int>)dynamicMethod.CreateDelegate(typeof(Func<object, int>));
+                var offset = @delegate(obj);
```
PJB3005: I would prefer if you moved this entire bit into its own method.

Robust.Shared/GameObjects/ComponentDependencies/ComponentDependencyManager.cs
```diff
@@ -0,0 +1,178 @@
...
+
+            var attributeFields = objType.GetAllFields()
+                .Where(f => Attribute.IsDefined(f, typeof(ComponentDependencyAttribute))).ToArray();
+            var attributeFieldsLength = attributeFields.Length;
+
+
+            var queries = new (Type, int)[attributeFieldsLength];
+
+            int i = 0;
+            foreach (var field in attributeFields)
```
PJB3005: If you are going to keep track of an `i` counter like that then make the loop over `attributeFields` a normal for loop instead.

Robust.Shared/GameObjects/ComponentDependencies/ComponentDependencyManager.cs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                if (!NullableHelper.IsMarkedAsNullable(field))
```
PJB3005: You should probably explicitly check that the fields here are reference types so that the attribute on e.g. `Vector2 position` gets rejected early.

Robust.Shared/Utility/NullableHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+
+namespace Robust.Shared.Utility
+{
+    public static class NullableHelper
+    {
+        private const int AnnotatedNullableFlag = 2;
+
+        private static Dictionary<Assembly, (Type AttributeType, FieldInfo NullableFlagsField)> _nullableAttributeTypeCache = new Dictionary<Assembly, (Type AttributeType, FieldInfo NullableFlagsField)>();
+
+        private static Dictionary<Assembly, (Type AttributeType, FieldInfo FlagsField)> _nullableContextAttributeTypeCache = new Dictionary<Assembly, (Type AttributeType, FieldInfo FlagsField)>();
```
PJB3005: You have two problems here with this cache:
1. No locking so this is completely non-threadsafe. And yes, threads will definitely touch this because of integration testing.
2. This can cause issues with assembly unloading if we ever explore that.

You can solve (2) by using a [`ConditionalWeakTable<TKey, TValue>`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2?view=netcore-3.1) instead of a normal dictionary.

Robust.Shared/Utility/NullableHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            return (byte) (assemblyNullableEntry.FlagsField.GetValue(nullableAttribute) ?? 0);
+        }
+
+        private static void CacheNullableFieldInfo(Assembly assembly)
+        {
+            var nullableAttributeType = assembly.GetType("System.Runtime.CompilerServices.NullableAttribute");
+            if (nullableAttributeType == null)
+            {
+                throw new Exception($"No System.Runtime.CompilerServices.NullableAttribute found in Assembly {assembly}");
```
PJB3005: This is something that can totally reasonably happen from the assembly just being compiled without any nullable annotations. You should not log an error here and treat the entire assembly as being oblivious (0).

PJB3005 (changes requested):

Robust.Shared/Utility/NullableHelper.cs
```diff
@@ -0,0 +1,135 @@
...
+                    return 1;
+                }
+
+                return (byte) (assemblyNullableEntry.Value.FlagsField.GetValue(nullableAttribute) ?? 1);
+            }
+        }
+
+        private static void CacheNullableFieldInfo(Assembly assembly)
+        {
+            lock (_nullableAttributeTypeCache)
```
PJB3005: Do not lock it again inside these cache methods. The lock is and MUST already held by the calling Get* method and locking twice like this is generally frowned upon.

PaulRitter: got it

PJB3005 (approved):

PR #1314, opened 2020-10-10
Pause refactor

metalgearsloth:
Marginally faster for the systems that were checking for it. ComponentManager could be more aware of it rather than using properties but future problem.

I used pause = true until systems have been changed to use it.

Old:
old_pauses

New:
new_pauses


PJB3005:
I don't like this in principle because it *really* should just be cheaper to have these kinds of flags buttttt I'll go along with it for now until something better comes along.

PJB3005:
~~On second thought, I would prefer if this remained an implementation detail in the engine. I'll just quickly...~~

Nevermind

PR #1318, opened 2020-10-11
Replace MapIndices with Vector2i

DrSmugleaf:
Very high speed refactors


PJB3005:
This will break stuff but I don't care.

PR #1324, opened 2020-10-12
Polar coordinate lighting implementation

20kdc:
This was actually ready for ages but I never PR'd it.
Or if I did I forgot or it was closed or something. IDK.
Following screenshots were taken with `--cvar net.predict=false` for more accurate performance information.

BEFORE COMMIT:
image

AFTER COMMIT:
image

PR credits:
+ PJB brought up the idea of using polar coordinates
+ https://www.gamasutra.com/blogs/RobWare/20180226/313491/Fast_2D_shadows_in_Unity_using_1D_shadow_mapping.php was referenced but not used in full

Important notes:
+ Visible (if small) difference in performance on this hardware
+ Shadows/fov "bend" with circular outlines. This is not a resolution problem, it's a lack of proper correction - interpolation of depth over angles is being done linearly, which is incorrect. For lighting I doubt this matters, especially given wall blend blur. For FOV a solution would be good.
+ Discord user "Abraham" appears to have performance problems with lighting that take them down from 25FPS to 8FPS. It's possible that the depthmap rendering is having more of an effect than it should on their hardware, and this may help mitigate that.
+ The approach by RobWare used to solve the aliasing problem (making the shadowmap larger and requiring two shadow samples) was *not used* due to differing performance priorities - instead a 'double-strike' rendering of the depthmap was used.
+ Ultimately this set of tradeoffs should guarantee that polar coordinates will have superior performance unless atan is incredibly worse-performing - the depthmap draw calls are halved (rather than quartered), but as a result, occludeDepth (called once per pixel on every single sprite) is simpler.

### DIFF
--- a/Resources/Shaders/Internal/shadow_cast_shared.swsl
+++ b/Resources/Shaders/Internal/shadow_cast_shared.swsl
@@ -2,56 +2,13 @@
 
 const highp float PI = 3.1415926535897932384626433; // That enough digits?
 
-// Because the shadow map is projected as a SQUARE, not a circle.
-// We can't just use some atan magic to undo the projection.
-// This matrix is the projection matrix used to create the projection
-// so we can use that to "reverse engineer" the position.
-uniform highp mat4 shadowMatrix;
-
 // This returns a vec2 because of VSM moments.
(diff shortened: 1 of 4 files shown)


20kdc:
Ok, so, you know that thing about the shadows bending? Not a thing anymore.

20kdc:
image

PR #1325, opened 2020-10-13
Fix UI keybinds when not a printable character

### LINKED ISSUE #2143: Moving text cursor to the Left/Right is broken. Using Up/Down for the console history is broken too.
I'm gonna point my blame finger at the tetris arcade PR

ShadowCommander:
Tentative fix for UI keybinds
Fixes space-wizards/space-station-14#2143

There's three ways I can think of doing this. Static like I have it here, non-static and include Clyde in InputManager, or pass Printable from Clyde all the way down to InputManager.

The things that should work are LineEdit hotkeys, NT Block Game hotkeys, and typing keys (e.g. WASD) in LineEdit without triggering hotkeys like moving the character.


PJB3005:
I really don't think is a correct way to fix this. Let me try some stuff myself.

PR #1327, opened 2020-10-14
Fix (some) physics stuttering

metalgearsloth:
This fixes the visual one where you walk into a wall, i.e. the occlusion + character stuttering when walking into particular walls. I'm not 100% on the best way to fix it given it's floating-point imprecision in the physics itself but this is a way to fix it.

Running into walls should also have less of a performance impact as the deferred updates are no longer being run as well.

I think the object stuttering when being pushed is due to using velocities and not forces but I need to test further.


PJB3005:
This seems to cause a failure in the save load tests in content, though admittedly it seems float related.

gradientvera:
Conflicts. Also, do we want to merge this? The save load tests can probably be fixed by saving the maps again.

PJB3005:
The maps *do* get saved again during the test. That's why it's called `LoadSaveTicksSaveStationStation`, it makes a control save *right* after loading, then waits 5 ticks, then makes another save. To make sure the map does not get modified while paused & uninitialized.

metalgearsloth:
Should be fixed now, I did some less dumb things. Makes physics a bit faster too (like 20% when I profiled?) because LocalRotation events aren't getting spammed out every tick anymore as well.

PR #1335, opened 2020-10-15
(Attempt to) draw world overlays inbetween entities

metalgearsloth:
I don't know enough about renderers to make it good. At this stage it'd probably be better to have a separate space for above grids buuuttt this code could definitely be made better.

Main reason I flushed the render queue is it seems like the sprites touching it affected it.

Ingame:
image

Old Profile:
image

New Profile:
image


Robust.Client/Graphics/Clyde/Clyde.HLR.cs
```diff
@@ -153,6 +166,28 @@ private void DrawEntities(Viewport viewport, Box2 worldBounds)
...
+                    if (overlay.ZIndex <= entry.sprite.DrawDepth)
+                    {
+                        if (!flushed)
+                        {
+                            FlushRenderQueue();
+                            flushed = true;
+                        }
+
+                        overlay.ClydeRender(_renderHandle, OverlaySpace.WorldSpace);
+                        worldOverlays.RemoveAt(0);
```
PJB3005: Instead of doing this, just keep track of the `j` index into the list and don't reset it. Then you don't have to move stuff around in the array.

PJB3005 (commented):
Was that profile made in release mode? I see stuff like ``RefList`1.get_Item(Int32)`` in there which I would think would be inlined away.

metalgearsloth:
image

new one

PR #1338, opened 2020-10-16
Fix even more physics jank

metalgearsloth:
tl;dr because I accidentally deleted my last description: Looked at Randy Gaul's code more and used InvMass more here in the SolveCollisionImpulse step. This means less FixClipping needed so less jitter. Stuff still occasionally phases through walls but I'm a lot happier with it in game.

Could also adjust CanMove so it only checks Anchored probably as the other checks are redundant.

I also had mass serialize as well because it wasn't being done. It also has my eye fix (mainly because it was making testing ingame easier with it applied).

[fizix.zip](https://github.com/space-wizards/RobustToolbox/files/5391066/fizix.zip)

zip with old and new physics applied (hopefully it's obvious which is which).

### DIFF
--- a/Robust.Shared/GameObjects/Components/Collidable/PhysicsComponent.Collision.cs
+++ b/Robust.Shared/GameObjects/Components/Collidable/PhysicsComponent.Collision.cs
@@ -76,7 +76,7 @@ namespace Robust.Shared.GameObjects.Components
             {
                 if (_sleepAccumulator == value)
                     return;
-                
+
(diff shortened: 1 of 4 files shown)


metalgearsloth:
I THINK if you increase the solve iterations it'll stop stuff yeeting through walls although I'm not sure on why this line is here (lag stopper?)

`if (_timing.InSimulation) divisions = 1;`

gradientvera (approved):

PR #1340, opened 2020-10-18
Add debug assert for transform self parenting

DrSmugleaf:
It stack overflows in that case.


Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -538,6 +538,9 @@ public virtual void AttachParent(ITransformComponent newParent)
             if (newParent == null)
                 return;
 
+            DebugTools.Assert(newParent.Owner != Owner,
```
PJB3005: So... just `newParent != this`?

DrSmugleaf: Fixed.

PR #1347, opened 2020-10-23
Fix grid parenting

metalgearsloth:
ROAD TO SHUTTLES.
A bunch of transform stuff probably needs refactoring because it's from the pre-EntityCoordinates times.

Anything that's parented to another non-map/grid entity shouldn't be touching its WorldPosition.


metalgearsloth:
I'm dumb and will fix the tests.

PJB3005 (commented):
I really don't feel like this is a good idea. Why is this necessary?

metalgearsloth:
If a mob's going from invalidgrid -> grid and back it needs to be handled some way or another.

PR #1351, opened 2020-10-26
Minor physics optimisation

metalgearsloth:
It's the shuffle for every single collision which seemed less than ideal when you can just use a random index. Rest is just cleanup.

Mainly spotted after I connected to USW and was wondering why my FPS was 0.01 as below:

image

Still a WIP overall.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Buffers;
 using System.Collections.Generic;
 using Robust.Shared.Containers;
 using Robust.Shared.GameObjects.Components;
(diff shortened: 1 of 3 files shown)


Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -241,21 +242,25 @@ private void ProcessCollisions(IEnumerable<IPhysicsComponent> awakeBodies)
...
             var counter = 0;
-            while(GetNextCollision(_collisionCache, counter, out var collision))
-            {
-                collision.A.WakeBody();
-                collision.B.WakeBody();
 
-                counter++;
-                var impulse = _physicsManager.SolveCollisionImpulse(collision);
-                if (collision.A.CanMove())
+            if (_collisionCache.Count > 0)
```
metalgearsloth, line 249: GetNextCollision will throw if these are 0 now so I just added this up front.

Robust.Shared/Interfaces/Physics/IPhysicsManager.cs
```diff
@@ -128,22 +128,7 @@ public Ray Ray
...
-                        return B.LinearVelocity - A.LinearVelocity;
-                    return -A.LinearVelocity;
-                }
-
-                if (B != null)
-                    return B.LinearVelocity;
-                return Vector2.Zero;
-            }
-        }
+        public Vector2 RelativeVelocity => B.LinearVelocity - A.LinearVelocity;
```
metalgearsloth, line 131: A and B can never be null

Robust.Shared/Physics/PhysicsManager.cs
```diff
@@ -106,8 +106,8 @@ public Vector2 SolveCollisionImpulse(Manifold manifold)
             if (!aP.CanMove() && !bP.CanMove()) return Vector2.Zero;
 
             var restitution = 0.01f;
-            var normal = CalculateNormal(manifold.A, manifold.B);
-            var rV = bP.LinearVelocity - aP.LinearVelocity;
+            var normal = manifold.Normal;
```
metalgearsloth, line 109: Just reducing duplication on these, will be especially when we have manifolds between disparate shapes (e.g. circle against AABB).

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -309,19 +314,17 @@ private bool GetNextCollision(IReadOnlyList<Manifold> collisions, int counter, o
...
+            var offset = _random.Next(collisions.Count - 1);
+            for (var i = 0; i < collisions.Count; i++)
             {
+                var index = (i + offset) % collisions.Count;
                 if (collisions[index].Unresolved)
                 {
                     collision = collisions[index];
                     return true;
                 }
+
```
metalgearsloth, line 329: The actual meat of it.

gradientvera (approved):

PR #1357, opened 2020-10-28
Fix map deserialized entities not being included in OccluderSystem

DrSmugleaf:
Because when OccluderSystem gets the event for saltern, entities are parented but their transforms are not initialized which means that the occluder system cant loop through Transform.Children.

This is only an issue on the server.

Needed for https://github.com/space-wizards/space-station-14/pull/2253 and for unoccluded checks to work in general.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Light/OccluderComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Light/OccluderComponent.cs
@@ -45,6 +45,13 @@ namespace Robust.Shared.GameObjects
             }
         }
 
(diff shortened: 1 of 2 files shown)


PJB3005 (changes requested):
Do this from `Startup()` on `OccluderComponent` instead.

DrSmugleaf:
> Do this from `Startup()` on `OccluderComponent` instead.

Done.

metalgearsloth (approved):
Sounds PJB approved to me

PR #1359, opened 2020-10-28
Fix a bunch of typos

DrSmugleaf:
Needs a content PR.


metalgearsloth (approved):

PJB3005:
Please make this compatible with content again with `[Obsolete]` attributes so it isn't such a royal PITA to synchronize with content.

DrSmugleaf:
> Please make this compatible with content again with `[Obsolete]` attributes so it isn't such a royal PITA to synchronize with content.

That seems like too much effort for what it's worth, I'll reopen the PR later if I get around to it.

PR #1360, opened 2020-10-29
Rotatable bounding boxes

PaulRitter:
wha- is it actually? might zumo actually be able to start using edge-firelocks?!?!?

no stackoverflow copypasta, i offered my sleep for this, its 2am

### Now for the technical part
This fixes `PhysShapeRect` to actually rotate the bounding box relative to the **entitys** center point, not the **bounding boxes** center point.
**This also supports non-cardinal dirs, if it ever is needed.**
```
  - type: Physics
    shapes:
    - !type:PhysShapeAabb
      bounds: "-0.45, 0.3, 0.45, 0.5"
      layer: [ Passable ]
```

pics:
image


Robust.Shared.Maths/Box2Rotated.cs
```diff
@@ -46,24 +46,46 @@ public Box2Rotated(Box2 box, Angle rotation, Vector2 origin)
...
-            var (X1, Y1) = Box.TopRight;
-
-            var Fi = Rotation.Theta;
-
-            var CX = (X0 + X1) / 2;  //Center point
-            var CY = (Y0 + Y1) / 2;
-            var WX = (X1 - X0) / 2;  //Half-width
-            var WY = (Y1 - Y0) / 2;
+            float[] allX = new float[4];
+            float[] allY = new float[4];
```
PaulRitter: this seems to alloc alot of ram... any idea of optimizing this? @PJB3005

PaulRitter:
ready for review

Robust.Shared.Maths/Box2Rotated.cs
metalgearsloth: Spans?

Robust.Shared/Physics/PhysShapeRect.cs
```diff
@@ -62,15 +92,45 @@ public void ExposeData(ObjectSerializer serializer)
             serializer.DataField(ref _collisionLayer, "layer", 0, WithFormat.Flags<CollisionLayer>());
             serializer.DataField(ref _collisionMask, "mask", 0, WithFormat.Flags<CollisionMask>());
             serializer.DataField(ref _rectangle, "bounds", Box2.UnitCentered);
+
+            List<BoundingBoxOverridePrototype> overrides = new List<BoundingBoxOverridePrototype>();
```
metalgearsloth: var

Robust.Shared/Physics/PhysShapeRect.cs
```diff
@@ -1,17 +1,36 @@
...
+using System.Collections.Generic;
+using Robust.Shared.Interfaces.Serialization;
 using Robust.Shared.Maths;
 using Robust.Shared.Serialization;
 using Robust.Shared.ViewVariables;
 
 namespace Robust.Shared.Physics
 {
+    /// <summary>
+    /// A physics shape that represents an Axis-Aligned Bounding Box.
```
metalgearsloth: (No longer axis-aligned). Represents an OBB more now.

metalgearsloth (changes requested):
Looks fine to me, would need to see how it profiles. I'd also need to see how it works ingame as all of the Box2 methods were built under the assumption it's an AABB (I don't know how much PhysShapeRect was used in the past).

PaulRitter:
>  I'd also need to see how it works ingame as all of the Box2 methods were built under the assumption it's an AABB (I don't know how much PhysShapeRect was used in the past).

if i understand what an AABB is correctly, what im doing is converting the OBB (PhysShapeRect) into an AABB in `CalcBoundingBox`

at least it works ingame :D

Robust.Client/ViewVariables/Editors/VVPropEditorUIBox2.cs
```diff
@@ -102,10 +102,10 @@ protected override Control MakeUI(object? value)
             {
                 case BoxType.Box2:
                 case BoxType.Box2i:
-                    hBoxContainer.AddChild(left);
                     hBoxContainer.AddChild(bottom);
-                    hBoxContainer.AddChild(right);
+                    hBoxContainer.AddChild(left);
                     hBoxContainer.AddChild(top);
+                    hBoxContainer.AddChild(right);
```
PJB3005: This should most definitely stay `LBTR` order. I know prototypes have stupid ordering here, those should be fixed instead. X coordinate naturally comes before Y.

PJB3005 (commented):
The bounding box override stuff would probably be better served as something explicitly supported by collidable component so certain phys shapes are only selected at certain angles.

That said, it seems like a *really* bad idea so I would like to know what problem it's actually solving here.

PaulRitter:
> That said, it seems like a _really_ bad idea so I would like to know what problem it's actually solving here.

imo you will have cases where the the sprite wont rotate according to origin (example: poweredlight)
altough maybe making another physshape and defining all cardinal dirs would probably be a better solution here(?)

PJB3005:
For those kinds of cases the hitbox literally does not matter and can *never* be correct anyways because the sprite is always dependent on camera rotation, so it's impossible to ever match a hitbox to.

Robust.Shared/Physics/PhysShapeRect.cs
```diff
@@ -50,10 +58,14 @@ public int CollisionMask
         public void ApplyState() { }
 
         public void DebugDraw(DebugDrawingHandle handle, in Matrix3 modelMatrix, in Box2 worldViewport,
-            float sleepPercent)
+            float sleepPercent, Angle angle)
         {
-            handle.SetTransform(modelMatrix);
-            handle.DrawRect(Rectangle, handle.CalcWakeColor(handle.RectFillColor, sleepPercent));
+            var m = Matrix3.Identity;
```
PJB3005: Get rid of all of this and get rid of the `angle` parameter aswell. That's what `modelMatrix` is for. (Yes I know this was here because of the now-gone override code)

Robust.Shared.Maths/Box2Rotated.cs
```diff
@@ -46,24 +46,46 @@ public Box2Rotated(Box2 box, Angle rotation, Vector2 origin)
...
-            var (X0, Y0) = Box.BottomLeft;
-            var (X1, Y1) = Box.TopRight;
-
-            var Fi = Rotation.Theta;
-
-            var CX = (X0 + X1) / 2;  //Center point
-            var CY = (Y0 + Y1) / 2;
-            var WX = (X1 - X0) / 2;  //Half-width
-            var WY = (Y1 - Y0) / 2;
+            Span<float> allX = stackalloc float[4];
```
PJB3005, line 110: Don't these modifications make the method SIGNIFICANTLY more expensive? You're doing sin/cos 4 times now by fetching corners like that.

Why did you change this? Was the previous method broken?

PaulRitter: the previous method would rotate along the bb center, instead of the entity center.

PJB3005: I would prefer if the previous implementation was fixed then to fix the problem. If not, at the very least some unit tests please.

gradientvera (approved):
The SSE math looks good to me

PR #1365, opened 2020-10-29
Fix transform state throws?

metalgearsloth:
On server it said line 757 which isn't a valid line so I was guessing it's this one as it was from an EntityManager.GetEntity


PJB3005:
The question is, why is the server sending entity states with invalid transform parents? That shouldn't be happening in the first place.

metalgearsloth:
Good question, maybe I'll need to debug more.

I only seem to get it on live where I have like 300 ping which probably exposes it.

I have a feeling it's PVS because there is some spaghetti in UpdatePlayerSeenStates.

PR #1368, opened 2020-10-29
Add Force and Impulse to VirtualController and implement in PhysicsSystem

ShadowCommander:
Adds Force and Impulse to VirtualController.

Add HasFriction to VirtualController and change physics friction to use VirtualControllers.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -123,22 +123,33 @@ namespace Robust.Shared.GameObjects.Systems
                 if(!body.CanMove())
                     continue;
 
+                var mt = body.InvMass * deltaTime;
+
                 var linearVelocity = Vector2.Zero;
 
                 foreach (var controller in body.Controllers.Values)
                 {
                     controller.UpdateBeforeProcessing();
+                    controller.LinearVelocity += controller.Force * mt;
+
+                    controller.LinearVelocity += controller.Impulse * body.InvMass;
+
                     linearVelocity += controller.LinearVelocity;
+                    
+                    // forces are instantaneous, so these properties are cleared
+                    // once integrated. If you want to apply a continuous force,
+                    // it has to be re-applied every tick.
+                    controller.Force = Vector2.Zero;
+                    controller.Impulse = Vector2.Zero;
                 }
 
                 // i'm not sure if this is the proper way to solve this, but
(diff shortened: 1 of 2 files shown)


gradientvera:
CC @DrSmugleaf @metalgearsloth

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -123,22 +123,33 @@ protected void SimulateWorld(float deltaTime, bool prediction)
                 if(!body.CanMove())
                     continue;
 
+                var mt = body.InvMass * deltaTime;
+
                 var linearVelocity = Vector2.Zero;
```
gradientvera, line 128: I wonder if we could stop resetting the linear velocity every tick, hmm

ShadowCommander: So, to stop resetting the LinearVelocity every tick would need a rework of the whole idea of VirtualControllers, since it seems the original idea of using them in the first place was to allow separation of LinearVelocities so that stopping one of them is easier.

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -346,18 +355,24 @@ private void ProcessFriction(IPhysicsComponent body, float deltaTime)
...
 
-            // Clamp friction because friction can't make you accelerate backwards
-            friction = Math.Min(fVelocity, body.LinearVelocity.Length);
-
-            if (friction == 0.0f)
+            foreach (var controller in body.Controllers.Values)
             {
-                return;
-            }
+                if (controller.LinearVelocity == Vector2.Zero || !controller.HasFriction)
```
metalgearsloth, line 360: Which controllers would be setting their friction?

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
metalgearsloth: I wonder if we could have controllers reset every tick but then have the player mover be able to be started and stopped instantly?

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
ShadowCommander: Any of the VirtualControllers that were using `LinearVelocity * 0.85f`, so SlipController, HighPressureMovementController, FrictionController, and ThrowController at least.

PR #1370, opened 2020-10-30
Fix collision crash

metalgearsloth:


gradientvera (approved):

metalgearsloth:
Okay so 95% sure that it's because the static timer method is adding the component to the entity during collision so the ComponentManager has a cry so ToArray is probably the best fix for it to prevent this happening with other stuff in future.

PR #1378, opened 2020-11-02
Adds helper to get the next closest multiple of a number given a value.

gradientvera:


metalgearsloth (approved):

moonheart08:
in my personal opinion, there should be a little warning sticker on the integer functions noting they use floats internally, due to precision.

moonheart08:
OR, change them to use `(( x - 1 ) | (m - 1)) + 1` where x is the target number and m is what it needs to be a multiple of. This only works if it needs to be at or above x though, and may be better as a separate, no-precision-loss helper.
@Zumorica

gradientvera:
> OR, change them to use `(( x - 1 ) | (m - 1)) + 1` where x is the target number and m is what it needs to be a multiple of. This only works if it needs to be at or above x though, and may be better as a separate, no-precision-loss helper.
> @Zumorica

done

PR #1384, opened 2020-11-04
EyeComponent improvements

gradientvera:


Robust.Server/GameObjects/Components/Eye/EyeComponent.cs
```diff
@@ -1,32 +1,81 @@
...
                 Dirty();
             }
         }
 
+        public override Vector2 Zoom
+        {
+            get => _zoom;
+            set
+            {
+                if (_zoom == value)
```
metalgearsloth: My experience with vectors is that it should probably use approx because floats are annoying.

gradientvera: Done. I made it use `EqualsApprox` where appropiate.

Robust.Client/GameObjects/Components/Eye/EyeComponent.cs
```diff
@@ -72,47 +71,42 @@ public Vector2 Zoom
...
-        [ViewVariables(VVAccess.ReadWrite)]
-        public Vector2 Offset
+        public override Vector2 Offset
         {
-            get => offset;
+            get => _offset;
             set
             {
-                if(offset == value)
+                if(_offset == value)
```
metalgearsloth:
```suggestion
                if(_offset.EqualsApprox(value))
```

PR #1386, opened 2020-11-05
AlignWithClosestGridTile extension for EntityCoordinates

gradientvera:
Used for aligning an EntityCoordinates to the closest grid. Does the same as the `AlignTileAny` placement mode.
Also fixes component exception when adding a timer component to a deleted entity.

### DIFF
--- a/Robust.Shared/Map/CoordinatesExtensions.cs
+++ b/Robust.Shared/Map/CoordinatesExtensions.cs
@@ -25,9 +25,7 @@ namespace Robust.Shared.Map
 
             var gridId = coords.GetGridId(entityManager);
 
-            IMapGrid? mapGrid = null;
-
-            if (!gridId.IsValid() || !mapManager.TryGetGrid(gridId, out mapGrid))
+            if (!gridId.IsValid() || !mapManager.TryGetGrid(gridId, out _))
             {
                 // create a box around the cursor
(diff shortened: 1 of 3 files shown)


Robust.Shared/Map/CoordinatesExtensions.cs
```diff
@@ -12,8 +13,64 @@ public static EntityCoordinates ToEntityCoordinates(this Vector2i vector, GridId
...
+        {
+            var coords = coordinates;
+            entityManager ??= IoCManager.Resolve<IEntityManager>();
+            mapManager ??= IoCManager.Resolve<IMapManager>();
+
+            var gridId = coords.GetGridId(entityManager);
+
+            IMapGrid? mapGrid = null;
+
+            if (!gridId.IsValid() || !mapManager.TryGetGrid(gridId, out mapGrid))
```
metalgearsloth: Couldn't you just use out var mapGrid?

metalgearsloth (approved):
LGTM considering it's a port.

Robust.Shared/Map/CoordinatesExtensions.cs
gradientvera: Done. It was also unused so I renamed it to `_`

gradientvera: Actually, given how it's unused I just changed this to `GridExists()`

PR #1387, opened 2020-11-06
Large number of changes to Shaders and Overlays

ghost:
Requires space-wizards/space-station-14#2517
This is currently all things that this PR changes:

- Add SCREEN_TEXTURE as an opt-in for shaders.
- Adds priority option to define overlap for shaders in the same OverlaySpace.
- Reworks IOverlayManager to be less shitty
- Uses the stencil buffer to capture the FOV and adds a new OverlaySpace.WorldSpaceFOVStencil that applies it, allowing for shaders to be rendered in world space but below FOV.

And other various changes.


metalgearsloth:
FYI worldspace overlays get rendered at the same time as entities now.

I needed to render an overlay under entities (and above grids) and PJB said we could draw the overlays in-between entity drawdepths so that's what I did.

ghost:
I might need to change that and add a different OverlaySpace for "below entities, above grids".
As it stands in my branch WorldSpace is rendered after FOV buffer is applied, plus there's an additional one that is "below FOV but right above normal worldspace"

ghost:
Hey metalgear what exactly did you need worldspace overlays to render under entities for?

metalgearsloth:
Decals given they won't be full entities.

ghost:
I just added a handful of new WorldSpaces to choose from - 
WorldSpaceFOVStencil,
WorldSpace,
WorldSpaceBelowFOV,
WorldSpaceBelowEntities
Not exactly sure what you need it for so I didn't change the overlay you had in mind but that will be trivial to fix later.

ghost:
Oh, you replied right as I was typing. Are those in game yet? (I can change the space to WorldSpaceBelowEntities)

metalgearsloth:
Currently they're still entities. I have a branch with them working but haven't PR'd as it looked like you started this after the old PR.

ghost:
Well just change the worldspace of the overlay to WorldSpaceBelowEntities once this is merged and you should be good to go.

metalgearsloth:
Originally I had overlays rendered that were above grids but below entities but PJB mentioned it instead as someone else suggested it could be useful.

ghost:
That's what the new overlayspaces are for.

Robust.Client/Debugging/DebugDrawing.cs
```diff
@@ -94,8 +95,7 @@ private class PhysicsOverlay : Overlay
             private Vector2 _hoverStartScreen = Vector2.Zero;
             private List<IPhysBody> _hoverBodies = new List<IPhysBody>();
 
-            public PhysicsOverlay(IComponentManager compMan, IEyeManager eyeMan, IPrototypeManager protoMan, IInputManager inputManager)
-                : base(nameof(PhysicsOverlay))
+            public PhysicsOverlay(IComponentManager compMan, IEyeManager eyeMan, IPrototypeManager protoMan, IInputManager inputManager) : base()
```
metalgearsloth: Remove empty base calls (see below first).

Robust.Client/Graphics/Clyde/Clyde.HLR.cs
```diff
@@ -238,11 +274,7 @@ private void DrawEntitiesAndWorldOverlay(Viewport viewport, Box2 worldBounds)
             _drawingSpriteList.Clear();
             FlushRenderQueue();
 
-            // Cleanup remainders
-            foreach (var overlay in worldOverlays)
-            {
-                overlay.ClydeRender(_renderHandle, OverlaySpace.WorldSpace);
-            }
+            RenderOverlays(OverlaySpace.WorldSpaceBelowEntities);
```
metalgearsloth: This is being done above entities

Robust.Client/Graphics/Clyde/Clyde.HLR.cs
```diff
@@ -113,19 +116,52 @@ private void RenderOverlays(OverlaySpace space)
...
+                                FlushRenderQueue();
+                                UpdateOverlayScreenTexture(space, _mainViewport.RenderTarget);
+                            }
+                            if (overlay.OverwriteTargetFrameBuffer) {
+                                ClearFramebuffer(default);
+                            }
+                            overlay.ClydeRender(_renderHandle, space);
+                            FlushRenderQueue();
+                        }
+                    }
```
metalgearsloth: If these are sorted by priority you can store the offset so you don't iterate through each overlay for every priority. IMO you're better off just sorting once by priority rather than going through each tier.

Robust.Client/Graphics/Overlays/OverlayManager.cs
```diff
@@ -22,48 +23,100 @@ public void FrameUpdate(FrameEventArgs args)
...
+        public void RemoveOverlaysOfClass(string className) {
+            var overlaysCopy = new Dictionary<Guid, Overlay>(_overlays);
+            foreach (var (id, overlay) in overlaysCopy) {
+                if (overlay.GetType().ToString() == className) {
+                    overlay.Dispose();
+                    _overlays.Remove(id);
+                }
+            }
+
+        }
```
metalgearsloth: Do nameof(T) and T.GetType().ToString() get the same thing? Just seems odd to use one internally and the other externally.

Also would there ever be duplicate overlays of the same type? Seems wasteful to iterate over the dictionary every time when you could just use the types as keys and call it a day.

Robust.Client/Graphics/Overlays/OverlayManager.cs
```diff
@@ -22,48 +23,100 @@ public void FrameUpdate(FrameEventArgs args)
...
+            return false;
+        }
+
+        public Overlay GetOverlay(Guid id)
         {
             return _overlays[id];
         }
 
-        public T GetOverlay<T>(string id) where T : Overlay
+        public bool GetOverlaysOfClass<T>(out List<T> overlays) where T : Overlay
```
metalgearsloth: If this stays should be get TryGet

Robust.Shared/Enums/OverlaySpaces.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        /// <summary>
+        ///     This overlay will be drawn beneath entities, lighting, and FOV; above grids.
+        /// </summary>
+        WorldSpaceBelowEntities = 0b010000,
+
+        /// <summary>
+        ///     This overlay will be drawn in screen coordinates behind the world. 
+        /// </summary>
+        ScreenSpaceBelowWorld = 0b100000,
```
metalgearsloth, line 45: Make it a ushort, also bitshifting would be more readable IMO.

Robust.Shared/Enums/OverlayPriorities.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Robust.Shared.Enums {
+    /// <summary>
+    ///     Determines in which order overlays are drawn in the same <see cref="OverlaySpace"/>. Lower numbers get drawn first.
+    /// </summary>
+    public enum OverlayPriority { P1, P2, P3, P4, P5, P6, P7, P8, P9 }
```
metalgearsloth: I guess it helps to coordinate overlays...

Robust.Client/Utility/SpriteSpecifierExt.cs
```diff
@@ -35,6 +35,40 @@ public static Texture Frame0(this SpriteSpecifier specifier)
...
+                            return state.Icons[0];
+                        }
+                    }
+                    Logger.Error("Failed to load RSI {0}", rsi.RsiPath);
+                    return new Texture[0];
+
+                default:
+                    throw new NotImplementedException();
+            }
+        }
```
metalgearsloth, line 76: What's this for? We already have like 3 helpers to get sprite icons.

metalgearsloth (changes requested):
Also needs that PJB seal of approval obviously.

Robust.Shared/Enums/OverlayPriorities.cs
ghost: We needed a way to organize overlays within the same Space and I can't think of any better way.

Robust.Client/Utility/SpriteSpecifierExt.cs
ghost: This gets all frames of an RSI, and as far as I'm aware this didnt exist before (only for frame 0)

Robust.Shared/Enums/OverlaySpaces.cs
ghost: why is it even a bit when we never use the bit part anyway

Robust.Client/Graphics/Clyde/Clyde.HLR.cs
ghost: can I leave this as a TODO

ghost: also performance drop here is very minimal anyway since there wont ever really be more than 10 overlays, a year 2005 computer could handle it with no drop

Robust.Client/Graphics/Clyde/Clyde.HLR.cs
```diff
@@ -113,25 +116,60 @@ private void RenderOverlays(OverlaySpace space)
...
+                if (overlay.RequestScreenTexture && overlay.Space == space) {
+                    oTargets.Add(overlay);
+                }
             }
+            if (oTargets.Count > 0 && ScreenBufferTexture != null) {
+
+                GL.BindTexture(TextureTarget.Texture2D, screenBufferHandle.Handle);
+                GL.TexImage2D(TextureTarget.Texture2D, 0, PixelInternalFormat.Srgb8Alpha8, _framebufferSize.X, _framebufferSize.Y, 0,
+                    PixelFormat.Rgba, PixelType.UnsignedByte, IntPtr.Zero); //Need to do this every frame in case the screen/framebuffer size changes. Small optimization: just do a vec2 check instead? 
+                                                                            //I don't think TexImage2D is very costly but I could be wrong.
```
20kdc: I would recommend verifying that the framebuffer size actually changed before doing this.

If possible, it'd probably also be best if your overlay buffers were per-Viewport, as this would guarantee the framebuffer size won't mysteriously change (a viewport's size is fixed, the viewport will be recreated if necessary)
(note that this would, however, mean that ScreenSpace/ScreenSpaceBelowWorld overlays would not have access to this functionality as these are not drawn within a viewport)

Robust.Client/Graphics/Clyde/Clyde.HLR.cs
```diff
@@ -338,12 +368,20 @@ private void RenderViewport(Viewport viewport)
...
+
                     if (_lightManager.Enabled && _lightManager.DrawHardFov && eye.DrawFov)
                     {
+                        GL.Clear(ClearBufferMask.StencilBufferBit);
+                        GL.Enable(EnableCap.StencilTest);
+                        GL.StencilOp(StencilOp.Keep, StencilOp.Keep, StencilOp.Replace);
+                        GL.StencilFunc(StencilFunction.Always, 1, 0xFF);
+                        GL.StencilMask(0xFF);
                         ApplyFovToBuffer(viewport, eye);
+                        GL.StencilMask(0x00);
```
20kdc, line 419: This isn't disabled and the state will leak until it's disabled later. Not sure if this is a problem. Also, this whole thing will bring the stencil test flag out of sync with `_isStencilling` in `Clyde.Rendering.cs`, which could produce an issue under certain (probably rare) conditions.

Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
```diff
@@ -418,6 +420,28 @@ private void ClearFramebuffer(Color color)
...
                         break;
+                    case ClydeTexture clydeTexture:
+                        //This is important it starts at Texture6 since DrawCommandBatch uses Texture0 and Texture1 immediately after calling this
+                        //function! If passing in textures as uniforms ever stops working it might be since someone made it use all the way up to Texture6 too.
+                        //Hey what can ya do?
+                        TextureUnit cTarget = TextureUnit.Texture6+textureUnitVal;
+                        SetTexture(cTarget, ((ClydeTexture)clydeTexture).TextureId);
+                        program.SetUniformTexture(name, cTarget);
+                        textureUnitVal++;
+                        break;
```
20kdc, line 425: This looks like it would be better served by reducing the reliance on fixed texture unit numbers in the first place.

Don't actually consider that as something that needs to be changed though, just... Something to put in an issue.

That said, if doing so ends up worsening performance, nevermind.

Robust.Client/Graphics/Clyde/Clyde.HLR.cs
20kdc: If they're already sorted by priority, then there's no reason whatsoever to even do any of this.
If they're not sorted by priority, *why*?

Robust.Client/Graphics/Clyde/Clyde.Helpers.cs
```diff
@@ -31,6 +32,60 @@ private void SetTexture(TextureUnit unit, ClydeHandle textureId)
...
+
+        private ClydeTexture AtlasToClydeTextureExpensive(AtlasTexture atlas) {
+            ClydeTexture parentTex = (ClydeTexture)atlas.SourceTexture;
+            int width = (int)atlas.SubRegion.Width;
+            int height = (int)atlas.SubRegion.Height;
+
+            Byte4[] pixels = new Byte4[parentTex.Width * parentTex.Height];
+            GL.BindTexture(TextureTarget.Texture2D, _loadedTextures[parentTex.TextureId].OpenGLObject.Handle);
+            CheckGlError();
+            GL.GetTexImage(TextureTarget.Texture2D, 0, PixelFormat.Rgba, PixelType.UnsignedByte, pixels);
```
20kdc: This (`GetTexImage`) isn't available on OpeNGL ES 2.
Consider acquiring the image during atlas texture creation, getting the atlas texture coordinates rectangle and using that in the shader, or outright removing the functionality.
This is critical and requires changes.

Robust.Client/UserInterface/CustomControls/EntitySpawnWindow.cs
```diff
@@ -70,7 +70,7 @@ public sealed class EntitySpawnWindow : SS14Window
             this.resourceCache = resourceCache;
 
             Title = Loc.GetString("Entity Spawn Panel");
-
+
```
20kdc: ????

20kdc (changes requested):
The actual screentexture parts show promise, but this is going to need work, definitely.

Acruid:
Need those changes and conflicts resolved.

Robust.Client/Graphics/Clyde/Clyde.HLR.cs
20kdc: Please be advised that this does not appear to have been resolved at time of merge according to GitHub diff.

PR #1397, opened 2020-11-09
Fix occluder tree crash

metalgearsloth:


Robust.Shared/GameObjects/Systems/OccluderSystem.cs
```diff
@@ -25,9 +25,15 @@ public abstract class OccluderSystem : EntitySystem
...
         {
-            return _gridTrees[mapId][gridId];
+            if (!_gridTrees.TryGetValue(mapId, out var grids))
+                return null;
+
+            if (!grids.TryGetValue(gridId, out var gridTree))
+                return null;
+
+            return gridTree;
         }
```
gradientvera: At this point, just make this a `TryGetOccluderTreeForGrid` method similar to `TryGetComponent`.

gradientvera (changes requested):

PR #1399, opened 2020-11-11
spritecomp will ignore state/texture attribute when layers are defined

PaulRitter:
what it says on the tin. imo this is the only logical behaviour


metalgearsloth:
Probably out of scope but having a test that indicates this would be great (if it is out of scope make an issue for it).

PaulRitter:
i am planning to write an analyser that should indicate ANY unconsumed (by exposedata) attribute. but yeah, out of scope

PR #1401, opened 2020-11-12
EntityPrototypeIcon SpriteSpecifier

PaulRitter:
what it says on the tin, really useful for building ui


PaulRitter:
This PR now also features a slight refactor of Texturerect as well as a new method for the containermanager for the objectives pr

Robust.Server/GameObjects/Components/Container/ContainerManagerComponent.cs
```diff
@@ -258,6 +260,22 @@ public override void Initialize()
             _entitiesWaitingResolve = null;
         }
 
+        public int CountPrototypeOccurencesRecursive(string prototypeId)
```
PJB3005: This method really should not be in the engine, IMO.

PJB3005 (changes requested):

Robust.Server/GameObjects/Components/Container/ContainerManagerComponent.cs
PaulRitter: moved to content extension-func

PJB3005 (approved):

PR #1417, opened 2020-11-20
Add searching by component name to View Variables

DrSmugleaf:
Requires https://github.com/space-wizards/RobustToolbox/pull/1414

Should probably separate the search bar from the component list first before merging.

EWJ2cG0lbB


metalgearsloth (approved):
Events should be unsubscribed but rest LGTM

PR #1419, opened 2020-11-21
Implemented erasing rectangular areas

### LINKED ISSUE #666: Add area delete for entities

ManelNavola:
- Ctrl+Click can now be used (the same way as for grid placing) while erasing to delete entities in a free rectangular area (ignores entities with either IMapComponent or IActorComponent)
- Disabled the placing options drop-down menu while erasing to avoid confusion (?)

Fixes space-wizards/space-station-14#666

### DIFF
--- a/Robust.Client/Placement/PlacementManager.cs
+++ b/Robust.Client/Placement/PlacementManager.cs
@@ -494,22 +494,24 @@ namespace Robust.Client.Placement
 
         private bool CurrentEraserMouseCoordinates(out EntityCoordinates coordinates)
         {
-            var map = MapId.Nullspace;
             var ent = PlayerManager.LocalPlayer?.ControlledEntity;
-            if (ent != null)
-            {
-                map = ent.Transform.MapID;
-            }
-
-            if (map == MapId.Nullspace || !Eraser)
+            if (ent == null)
             {
                 coordinates = new EntityCoordinates();
                 return false;
             }
-
-            coordinates = EntityCoordinates.FromMap(ent.EntityManager, MapManager,
-                eyeManager.ScreenToMap(new ScreenCoordinates(_inputManager.MouseScreenPosition)));
-            return true;
+            else
+            {
+                var map = ent.Transform.MapID;
+                if (map == MapId.Nullspace || !Eraser)
+                {
+                    coordinates = new EntityCoordinates();
+                    return false;
+                }
+                coordinates = EntityCoordinates.FromMap(ent.EntityManager, MapManager,
+                    eyeManager.ScreenToMap(new ScreenCoordinates(_inputManager.MouseScreenPosition)));
+                return true;
+            }
         }
 
         /// <inheritdoc />
--- a/Robust.Client/Placement/PlacementManager.cs
+++ b/Robust.Client/Placement/PlacementManager.cs
@@ -21,6 +21,7 @@ using Robust.Shared.Reflection;
 using Robust.Shared.Utility;
 using Robust.Shared.Serialization;
 using Robust.Shared.Timing;
+using Robust.Shared.Log;
 
 namespace Robust.Client.Placement
 {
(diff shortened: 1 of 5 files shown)


Robust.Client/Placement/PlacementManager.cs
```diff
@@ -470,11 +514,62 @@ private bool CurrentMousePosition(out ScreenCoordinates coordinates)
             return true;
         }
 
+        private bool CurrentEraserMouseCoordinates(out EntityCoordinates coordinates)
+        {
+            var map = MapId.Nullspace;
+            var ent = PlayerManager.LocalPlayer!.ControlledEntity;
```
metalgearsloth:
```suggestion
            var ent = PlayerManager.LocalPlayer?.ControlledEntity;
```

Robust.Client/Placement/PlacementManager.cs
```diff
@@ -470,11 +514,62 @@ private bool CurrentMousePosition(out ScreenCoordinates coordinates)
...
+                map = ent.Transform.MapID;
+            }
+
+            if (map == MapId.Nullspace || !Eraser)
+            {
+                coordinates = new EntityCoordinates();
+                return false;
+            }
+
+            coordinates = EntityCoordinates.FromMap(ent!.EntityManager, MapManager,
```
metalgearsloth:
```suggestion
            coordinates = EntityCoordinates.FromMap(ent.EntityManager, MapManager,
```

Robust.Server/Placement/PlacementManager.cs
```diff
@@ -198,6 +202,60 @@ private void HandleEntRemoveReq(EntityUid entityUid)
                 _entityManager.DeleteEntity(entity);
         }
 
+        private void HandleRectRemoveReq(MsgPlacement msg)
+        {
+            EntityCoordinates start = msg.EntityCoordinates;
+            EntityCoordinates end = msg.EndEntityCoordinates;
+            EntityCoordinates center = new EntityCoordinates(start.EntityId, (start.Position + end.Position) / 2.0f);
```
metalgearsloth: Position is relative to their parent so if their parents are different then this won't work as intended.

Robust.Server/Placement/PlacementManager.cs
```diff
@@ -198,6 +202,60 @@ private void HandleEntRemoveReq(EntityUid entityUid)
...
+                else
+                {
+                    low = start.Y;
+                    high = end.Y;
+                }
+                foreach (IEntity entity in _entityManager.GetEntitiesInRange(center, xSize))
+                {
+                    if (entity.Deleted || entity.HasComponent<IMapGridComponent>() || entity.HasComponent<IActorComponent>())
+                        continue;
+                    if (entity.Transform.Coordinates.Y < low || entity.Transform.Coordinates.Y > high)
```
metalgearsloth: This as well for example

Robust.Client/Placement/PlacementManager.cs
```diff
@@ -512,6 +607,14 @@ private void ActivateGridMode()
...
         }
 
+        private void EraseRectMode()
+        {
+            if (!CurrentEraserMouseCoordinates(out EntityCoordinates coordinates))
+                return;
+
+            EraserStartPoint = coordinates;
+        }
+
```
metalgearsloth: Not sure if it needs changing to screenspace or shader needs changing but the rectangle is affected by lighting.

metalgearsloth (changes requested):

ManelNavola:
Thanks for the review! I'll fix the code asap

Robust.Server/Placement/PlacementManager.cs
ManelNavola: This should be fixed in the latest commit, let me know if I misunderstood the issue!

PJB3005:
While I appreciate the work here, the problem is that `PlacementManager` will have to be torn out sooner or later for legal reasons.

ManelNavola:
> While I appreciate the work here, the problem is that `PlacementManager` will have to be torn out sooner or later for legal reasons.

Yeah that is true, but maybe it can be of help until then? I mean if the PlacementManager it's going to be rewritten very soon then there's no real reason to merge this.

Robust.Client/Placement/PlacementManager.cs
PJB3005: To make it not affected by lighting you add `light_mode unshaded;` to the top of the shader.

ManelNavola: Added the "unshaded" shader which should use `light_mode unshaded;`, now it isn't affected by lighting, but it is affected by shadows nonetheless.

metalgearsloth:
> 
> 
> > While I appreciate the work here, the problem is that `PlacementManager` will have to be torn out sooner or later for legal reasons.
> 
> Yeah that is true, but maybe it can be of help until then? I mean if the PlacementManager it's going to be rewritten very soon then there's no real reason to merge this.

The main people who'd need it would be moony or swept I think, and moony's just started rewriting the placementmanager now (though who knows whether or not they finish).

Acruid:
Since it won't be merged, should this stay open?

Acruid:
Who knows when the placement system will be rewritten, so in the meantime I'm merging this so people can enjoy it.

PR #1421, opened 2020-11-25
Cleanup: Fix grid lookups performed just to get map IDs

20kdc:
NOTE: This commit assumes that `mapManager.GetGrid(effect.Coordinates.GetGridId(_entityManager)).ParentMapId` is equivalent to `effect.Coordinates.GetMapId(_entityManager)` but the latter won't crash when no grid is available.
Please do not merge if this is incorrect.
This is sort of a sequel to last time, https://github.com/space-wizards/space-station-14/pull/2624


gradientvera:
Your assumption is correct.

Clyybber (approved):

PR #1427, opened 2020-11-29
Added master volume configuration and set master volume method

ManelNavola:
- Added CVar for master volume
- Added method in ClydeAudio for setting the master volume
- Both method and master volume configurations treat 1 as 100% and 0 as 0%

### DIFF
--- a/Robust.Client/Interfaces/Graphics/IClydeAudio.cs
+++ b/Robust.Client/Interfaces/Graphics/IClydeAudio.cs
@@ -1,4 +1,4 @@
-using System.IO;
+﻿using System.IO;
 using Robust.Client.Audio;
 
 namespace Robust.Client.Interfaces.Graphics
@@ -8,6 +8,8 @@ namespace Robust.Client.Interfaces.Graphics
         // AUDIO SYSTEM DOWN BELOW.
         AudioStream LoadAudioOggVorbis(Stream stream, string? name = null);
         AudioStream LoadAudioWav(Stream stream, string? name = null);
+
+        void SetMasterVolume(float newVolume);
 
         IClydeAudioSource CreateAudioSource(AudioStream stream);
         IClydeBufferedAudioSource CreateBufferedAudioSource(int buffers, bool floatAudio=false);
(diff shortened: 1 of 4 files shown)


Robust.Shared/CVars.cs
```diff
@@ -220,6 +220,9 @@ protected CVars()
         public static readonly CVarDef<string> AudioDevice =
             CVarDef.Create("audio.device", string.Empty);
 
+        public static readonly CVarDef<float> AudioMasterVolume =
+            CVarDef.Create("audio.mastervolume", 1.0f);
+
         public static readonly CVarDef<string> PlayerName =
```
metalgearsloth: Should be on client I think? Server definitely doesn't care.

ManelNavola: Agreed, fixed

metalgearsloth (approved):

Robust.Client/Interfaces/Graphics/IClydeAudio.cs
```diff
@@ -9,6 +9,8 @@ public interface IClydeAudio
         AudioStream LoadAudioOggVorbis(Stream stream, string? name = null);
         AudioStream LoadAudioWav(Stream stream, string? name = null);
 
+        void SetMasterVolume(float newVolume);
```
PJB3005, line 12: Not sure this should be a public method since using it would always go out of sync with the actual cvar.

ManelNavola: The method is public as when the volume slider is moved during the game, the audio volume changes without saving to the CVar each time. In the audio options menu, the CVar is only saved after the apply button is clicked, like the graphics menu does. Should I do what I'm doing in a different way?

PJB3005: Fair point. Not constantly re-saving the config file is a good thing yeah.

One minor difference with the audio options is that those are only applied to the game when you hit the apply button (and they're actually saved). Of course I realize that would be awful for an audio slider so we're good here.

PR #1430, opened 2020-12-01
Add bytes or sbytes to enum where available 

### LINKED ISSUE #973: The great Enum Space Waste of 2020  (aka why are enums not byte sized)
This is probably of importance to networking in particular:
Enums default to `int` sized. It is highly likely that *every single enum in the codebase* should be `byte` sized for network savings.
Using up a byte of space is better than using 4 and depending on the compressor to handle the 3 extra zero bytes.

Ygg01:
Fixes #973 

Sets all enums to be `byte` sized (or `sbyte` where possible).

There were a few exceptions that didn't allow YamlSerializer to pass.

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.Constants.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.Constants.cs
@@ -35,7 +35,7 @@ namespace Robust.Client.Graphics.Clyde
         // To be clear: You shouldn't change this. This just helps with understanding where Primitive Restart is being used.
         private const ushort PrimitiveRestartIndex = ushort.MaxValue;
 
-        private enum Renderer
+        private enum Renderer : short
         {
             // Default: Try all supported renderers (not necessarily the renderers shown here)
             Default = default,
--- a/OpenToolkit.GraphicsLibraryFramework/Enums/InputAction.cs
+++ b/OpenToolkit.GraphicsLibraryFramework/Enums/InputAction.cs
@@ -13,7 +13,7 @@ namespace OpenToolkit.GraphicsLibraryFramework
     /// Defines event information for <see cref="GLFWCallbacks.KeyCallback"/>
     /// or <see cref="GLFWCallbacks.MouseButtonCallback"/>.
     /// </summary>
-    public enum InputAction
+    public enum InputAction : byte
     {
         /// <summary>
         /// The key or mouse button was released.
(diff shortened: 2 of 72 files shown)


gradientvera (approved):
This *seems* fine but maybe someone else should take a look at it first.

PaulRitter:
inb4 packet size drops by 500%

Ygg01:
Yeah. I wanted to mark his as draft. I suspect space-station-14 needs to be changed too, right?

gradientvera:
> Yeah. I wanted to mark his as draft. I suspect space-station-14 needs to be changed too, right?

You mean content? The tests seem to past on content so I'd say it's fine.

Robust.Client/Graphics/Clyde/Clyde.Constants.cs
```diff
@@ -35,7 +35,7 @@ internal sealed partial class Clyde
         // To be clear: You shouldn't change this. This just helps with understanding where Primitive Restart is being used.
         private const ushort PrimitiveRestartIndex = ushort.MaxValue;
 
-        private enum Renderer
+        private enum Renderer : short
```
exp111, line 38: Can't you make this sbyte?

Ygg01: Think I can. Really should have set this PR to draft

Ygg01:
> inb4 packet size drops by 500%

If the server doesn't work, then the packet size is 0. Infinite% reduction!

Ygg01:
@Zumorica can we test the Robust-Toolbox local version against SS14? I want to see that Serializer isn't crapping on my parade.

gradientvera:
> @Zumorica can we test the Robust-Toolbox local version against SS14? I want to see that Serializer isn't crapping on my parade.

That's what the "Test content master against engine / build" test does

Ygg01:
I should have worded that more precisely.  I want to start SS14 project using local RobustToolbox version. Even if temporarily. Tests are ok, but I don't have too great of confidence in them,

gradientvera:
Just check your branch in the RobustToolbox submodule locally then.

SaphireLattice:
If this is done for serialisation purposes, then shouldn't that happen within serializer and setting it to use bytes for enums? This is unlikely to help the actual processing as the memory words are still fair bit larger than a byte.

metalgearsloth:
> 
> 
> If this is done for serialisation purposes, then shouldn't that happen within serializer and setting it to use bytes for enums? This is unlikely to help the actual processing as the memory words are still fair bit larger than a byte.

Isn't it to help with what's being sent over the wire?

PJB3005:
> If this is done for serialisation purposes, then shouldn't that happen within serializer and setting it to use bytes for enums? This is unlikely to help the actual processing as the memory words are still fair bit larger than a byte.

This mostly changes the size of the enum on the heap. NetSerializer uses variable length int writing (that I do not know a formal name of) so a value like `(int) 5` still only gets written as one byte. The only place this really helps is `byte` enums that actually manage to use the last bit (which isn't any from what I can tell).

PJB3005 (approved):

PR #1431, opened 2020-12-02
Add EnsureComponentWarn methods

DrSmugleaf:
Ensures a component, if it doesn't exist it logs a warning. A lot of components in content were repeating this pattern manually in different places.

### DIFF
--- a/Robust.Shared/GameObjects/ComponentExt.cs
+++ b/Robust.Shared/GameObjects/ComponentExt.cs
@@ -62,11 +62,7 @@ namespace Robust.Shared.GameObjects
                 return true;
             }
 
-            warning ??= $"Entity {entity} " +
-                        (entity.Prototype == null
-                            ? ""
-                            : $"with prototype id {entity.Prototype.ID} ")
-                        + $"at {entity.Transform.MapPosition} did not have a {nameof(T)}";
+            warning ??= $"Entity {entity} at {entity.Transform.MapPosition} did not have a {typeof(T)}";
 
             Logger.Warning(warning);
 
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/ComponentExt.cs
```diff
@@ -41,5 +42,65 @@ public static T EnsureComponent<T>(this IEntity entity) where T : Component, new
...
+            if (entity.TryGetComponent(out component!))
+            {
+                return true;
+            }
+
+            warning ??= $"Entity {entity} " +
+                        (entity.Prototype == null
+                            ? ""
+                            : $"with prototype id {entity.Prototype.ID} ")
+                        + $"at {entity.Transform.MapPosition} did not have a {nameof(T)}";
```
PJB3005: 1. Entities already have their prototype ID in their `ToString()` so you can at least cut that part out.
2. `nameof(T)` just returns `"T"`. You need `typeof(T)` here.

Robust.Shared/GameObjects/ComponentExt.cs
```diff
@@ -41,5 +42,65 @@ public static T EnsureComponent<T>(this IEntity entity) where T : Component, new
...
+            if (entity.TryGetComponent(out T? component))
+            {
+                return component;
+            }
+
+            warning ??= $"Entity {entity} " +
+                        (entity.Prototype == null
+                            ? ""
+                            : $"with prototype id {entity.Prototype.ID} ")
+                        + $"at {entity.Transform.MapPosition} did not have a {nameof(T)}";
```
PJB3005: Same problems apply here.

PJB3005 (changes requested):

Robust.Shared/GameObjects/ComponentExt.cs
DrSmugleaf: Fixed.

Robust.Shared/GameObjects/ComponentExt.cs
DrSmugleaf: Fixed.

DrSmugleaf:
Reviews addressed.

PJB3005 (approved):

PR #1433, opened 2020-12-03
Fixed handling keybinds.yml with no binds

JohnGinnane:
If the user's keybinds.yml file existed but does not contain any "binds" of "leaveEmpty" then "**KeyNotFoundException**" was being thrown in InputManager.cs

I have tested the following permutations:

1. No user keybinds.yml exists
2. User keybinds.yml exists and contains "binds" and "leaveEmpty"
3. User keybinds.yml exists and is a valid YML file but does not contain "binds" or "leaveEmpty" - This would cause exceptions for missing keys
4. User keybinds.yml exists but only contains "binds" - This originally caused exception on a missing "leaveEmpty" key
5. User keybinds.yml exists but only contains "leaveEmpty" - This originally caused exception on a missing "binds" key

### DIFF
--- a/Robust.Client/Input/InputManager.cs
+++ b/Robust.Client/Input/InputManager.cs
@@ -406,38 +406,61 @@ namespace Robust.Client.Input
             var mapping = (YamlMappingNode) yamlStream.Documents[0].RootNode;
 
             var baseSerializer = YamlObjectSerializer.NewReader(mapping);
-            var baseKeyRegs = baseSerializer.ReadDataField<KeyBindingRegistration[]>("binds");
 
-            foreach (var reg in baseKeyRegs)
+            try
             {
-                if (!NetworkBindMap.FunctionExists(reg.Function.FunctionName))
-                {
-                    Logger.ErrorS("input", "Key function in {0} does not exist: '{1}'", file,
-                        reg.Function.FunctionName);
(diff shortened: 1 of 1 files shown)


Robust.Client/Input/InputManager.cs
```diff
@@ -406,38 +406,61 @@ private void LoadKeyFile(ResourcePath file, bool userData)
             var mapping = (YamlMappingNode) yamlStream.Documents[0].RootNode;
 
             var baseSerializer = YamlObjectSerializer.NewReader(mapping);
-            var baseKeyRegs = baseSerializer.ReadDataField<KeyBindingRegistration[]>("binds");
 
-            foreach (var reg in baseKeyRegs)
+            try
```
PJB3005: Using try catch on a `KeyNotFoundException` is *never* the correct choice. You should use a function like `TryReadDataField` instead of `ReadDataField` if you want to make it optional.

PJB3005 (changes requested):

Robust.Client/Input/InputManager.cs
JohnGinnane: Right, will do that tomorrow, thanks

PJB3005 (approved):

PR #1435, opened 2020-12-05
Multiselect option button, tooltip delay (Action Hotbar Support)

chairbender:
Adds a multiselect option button which allows selecting multiple elements instead of just one, used in action menu.

Tooltip delays can be customized per-control. The default of 1s was feeling a bit sluggish for some cases.

Added some clarification to how layout container margin setting works because it tripped me up, even though it should've been obvious to me.


Robust.Client/UserInterface/Controls/LayoutContainer.cs
```diff
@@ -126,6 +126,12 @@ public static void SetSize(Control control, Vector2 size)
 
         /// <summary>
         ///     Sets an anchor AND a margin preset. This is most likely the method you want.
+        ///
+        ///     Note that the current size and minimum size of the control affects how
+        ///     each of the margins will be set, so if your control needs to shrink beyond its
+        ///     current size / min size, you should either not call this method or only call it when your
+        ///     control has a size of (0, 0). Otherwise your control's size will never be able
+        ///     to go below the size implied by the margins set in this method.
```
metalgearsloth: There's also a remarks tag you can use but LGTM.

Robust.Client/UserInterface/Controls/MultiselectOptionButton.cs
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        public void SetItemText(int idx, string text)
+        {
+            var data = _buttonData[idx];
+            data.Text = text;
+            data.Button.Text = text;
+        }
+
+        private void _onPressed(ButtonEventArgs args)
```
metalgearsloth: Private methods are normally still PascalCased but eh.

Robust.Client/UserInterface/Controls/MultiselectOptionButton.cs
```diff
@@ -0,0 +1,301 @@
...
+            {
+                throw new ArgumentException("An item with the same key already exists.");
+            }
+
+            var button = new Button
+            {
+                Text = label,
+                ToggleMode = true
+            };
+            button.OnPressed += ButtonOnPressed;
```
metalgearsloth, line 103: Should be unsubscribed when removed. (via Clear / Remove)

Robust.Client/UserInterface/Controls/MultiselectOptionButton.cs
```diff
@@ -0,0 +1,301 @@
...
+
+        public void Clear()
+        {
+            _keyMap.Clear();
+            _buttonData.Clear();
+            _popupVBox.DisposeAllChildren();
+            _selectedKeys = new List<TKey>();
+        }
+
+        public int ItemCount => _buttonData.Count;
```
metalgearsloth: Move up with the others IMO

Robust.Client/UserInterface/Controls/MultiselectOptionButton.cs
```diff
@@ -0,0 +1,301 @@
...
+        public void SetItemDisabled(int idx, bool disabled)
+        {
+            var data = _buttonData[idx];
+            data.Disabled = disabled;
+            data.Button.Disabled = disabled;
+        }
+
+        public void SetItemIcon(int idx, Texture texture)
+        {
+        }
```
metalgearsloth: Should something be here?

Robust.Client/UserInterface/Controls/MultiselectOptionButton.cs
```diff
@@ -0,0 +1,301 @@
...
+                        _selectedKeys.Add(buttonData.Key);
+                    }
+                    else
+                    {
+                        _selectedKeys.Remove(buttonData.Key);
+                    }
+                    OnItemSelected?.Invoke(new ItemPressedEventArgs(buttonData.Key, obj.Button.Pressed, this));
+                    return;
+                }
+            }
```
metalgearsloth: If speed is ever a problem you could store the buttons in a dictionary / hashset but it'll be fine for now.

metalgearsloth (approved):
NGL I suck at UI so the comments aren't about functionality.

chairbender:
all the issues you are pointing out with multiselect option button were from copy-pasting the existing OptionButton, so I can fix those there as well.

chairbender:
Good to go. I also just discovered I had already merged in changes from robust master but hadn't pushed that merge yet. It's pushed now.

EDIT: Also note one small change to ReflectionManager so it can access nested public classes.

Robust.Client/UserInterface/Controls/MultiselectOptionButton.cs
```diff
@@ -0,0 +1,302 @@
...
+
+        private void OnPressedInternal(ButtonEventArgs args)
+        {
+            var globalPos = GlobalPosition;
+            var (minX, minY) = _popupVBox.CombinedMinimumSize;
+            var box = UIBox2.FromDimensions(globalPos, (Math.Max(minX, Width), minY));
+            _popup.Open(box);
+        }
+
+        protected override void Dispose(bool disposing)
```
PJB3005: `Dispose()` on controls is kind of an old wart API that I need to properly deprecate. It dates back to Godot where it was necessary to directly dispose the controls.

I would prefer if you didn't use dispose and instead added/removed the pop to/from the popup root when necessary (so only when showing it) and also closed the popup explicitly in `ExitedTree()`.

Robust.Shared/Reflection/ReflectionManager.cs
```diff
@@ -204,7 +204,7 @@ public bool TryParseEnumReference(string reference, [NotNullWhen(true)] out Enum
             Type? found = null;
             foreach (var derivedType in GetAllChildren(baseType))
             {
-                if (derivedType.Name == typeName && derivedType.IsPublic)
+                if (derivedType.Name == typeName && (derivedType.IsPublic || derivedType.IsNestedPublic))
```
PJB3005: Wouldn't this allow instantiating nested types on internal engine classes?

Robust.Client/UserInterface/Controls/MultiselectOptionButton.cs
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        public void RemoveItem(int idx)
+        {
+            var data = _buttonData[idx];
+            data.Button.OnPressed -= ButtonOnPressed;
+            _keyMap.Remove(data.Key);
+            _popupVBox.RemoveChild(data.Button);
+            _buttonData.RemoveAt(idx);
+        }
```
PJB3005: You're not updating the indices of the other entries of `_keyMap` here so if you remove something everything will shift up incorrectly.

Robust.Shared/Reflection/ReflectionManager.cs
chairbender: I only added it because I thought this was working before on one of my classes, then when I merged with the latest master (which included .net 5) it stopped working. But I'm pretty sure I remembered wrong.

Why we wouldn't want to allow this? It still checks if it's public, why is it a big deal whether it's nested or not? In my case it was nice to have IExposeData nested class inside one of my components, for defining the toggle action behavior. But I'd rather get this PR done so I'll just revert this for now and move my class to non-nested.

The unfortunate side effect is now I need to make one of the methods on the component public, even though it is only used by this class, whereas before the class was nested so I could access the private method.

chairbender:
I also applied these fixes to OptionButton.

Robust.Shared/Reflection/ReflectionManager.cs
PJB3005: If you have a definition like this:
```cs
internal class Foo
{
	public class Bar
	{

	}
}
```
We can't allow content to instantiate the `Bar` type. AFAIK this change would allow that?

chairbender: Yeah.

chairbender:
I'm not even remotely done with the content changes yet, but before I forget - I added something to the engine to make it easier to implement custom tooltips with a lot less code - Control.TooltipSupplier. It's allowed me to cut down a LOT on the tooltip-related code in content.

PJB3005 (approved):

PR #1436, opened 2020-12-06
Add new "TargetEntity" property to Popup control

JohnGinnane:
~~This is the engine change required to allow me to render the examine popup partially offscreen. It looks at the default behaviour for popups is to always remain on screen, for example the right click menu. However in some cases we might want to override this behaviour so I have added a property which allows us to do that (default false)~~

I have added a new attached property to **PopupContainer** which allows us to set the target entity of a popup. If the target entity is set then **Popup.Draw()** will draw a line from the popup to the target entity. Linked SS14 PR will manage setting/removing the target entity

SS14 PR: https://github.com/space-wizards/space-station-14/pull/2716


Robust.Client/UserInterface/Controls/Popup.cs
```diff
@@ -30,6 +30,7 @@ public void Open(UIBox2? box = null, Vector2? altPos = null)
                 MinimumSizeChanged();
             }
 
+            AllowDrawOffScreen = allowDrawOffScreen;
```
PJB3005: You should use an attached property for this instead, just lie the other two origin parameters.

Robust.Client/UserInterface/Controls/PopupContainer.cs
```diff
@@ -59,42 +59,51 @@ protected override void LayoutUpdateOverride()
                 var (r, b) = size + offset; // bottom right corner.
 
                 var isAltPos = false;
+                bool keepOnScreen = true;
 
-                // Clamp the right edge.
-                if (r > Width)
+                if (child.GetType() == typeof(Popup))
```
PJB3005: While this is not a block of code you should keep (since I'd prefer you made it an attached property), doing `.GetType() == typeof()` is a *very* bad way to do type checks. It's slow and only works if the type isn't inherited.

A better way to do type checks is to do something like `child is Popup p`.

Though again, you shouldn't need a type check in the first place because attached property.

PJB3005 (changes requested):

RemieRichards:
Popups that go off-screen are a sin and shouldn't be allowed in the first place.

Robust.Client/UserInterface/Controls/Popup.cs
JohnGinnane: I will do that

Robust.Client/UserInterface/Controls/PopupContainer.cs
JohnGinnane: With the above change, if I try to call a new `PopupContainer.GetAllowDrawOffScreen` on a control that doesn't have that property will it throw any errors?

JohnGinnane:
> 
> 
> Popups that go off-screen are a sin and shouldn't be allowed in the first place.

I think for this case it looks better than just sticking to the edge of the screen. And by default the popups don't go offscreen

RemieRichards:
It's going to lead to information going off screen, which is entirely worthless.

JohnGinnane:
> 
> 
> It's going to lead to information going off screen, which is entirely worthless.

Yep, but if the popup goes off the screen completely then it should disappear as there's nothing useful on there. Here's a video of how it used to behave when you couldn't draw offscreen: https://streamable.com/ig3nb4

RemieRichards:
That behaviour is good though.

JohnGinnane:
> 
> 
> That behaviour is good though.

Maybe? I posted it on discord and got different feedback so I changed it to stop it "sticking" to the screen

RemieRichards:
We can discuss it a bit more in discord later if you want (5 hours or so), I'll be having words with anyone who dislikes the sticking :P

JohnGinnane:
> 
> 
> We can discuss it a bit more in discord later if you want (5 hours or so), I'll be having words with anyone who dislikes the sticking :P

Sure let's do that. I might hold off on making changes as if we decide to keep it to sticking then we won't need to property at all I think

Robust.Client/UserInterface/Controls/PopupContainer.cs
PJB3005: No. You specify a default value that will be returned instead when defining the attached property.

metalgearsloth:
Is this supposed to be examines tracking or the line? When I checked it out with the SS14 pr it didn't track and also the line didn't handle UI scaling (for the record I prefer the tracking):

image

JohnGinnane:
> 
> 
> Is this supposed to be examines tracking or the line? When I checked it out with the SS14 pr it didn't track and also the line didn't handle UI scaling

This should draw a green line to centre of target entity if you or they move around. As for UI scaling, I need to look into that. To test that I just change the UI scaling in the options and see how it behaves?

>
>
>(for the record I prefer the tracking):

I could add a user setting to toggle between no behaviour, line, or track?

metalgearsloth:
> 
> 
> > Is this supposed to be examines tracking or the line? When I checked it out with the SS14 pr it didn't track and also the line didn't handle UI scaling
> 
> This should draw a green line to centre of target entity if you or they move around. As for UI scaling, I need to look into that. To test that I just change the UI scaling in the options and see how it behaves?
> 
> > (for the record I prefer the tracking):
> 
> I could add a user setting to toggle between no behaviour, line, or track?

1. yeah that should show it. My game defaults to 150% scaling now so any issues are immediately obvious lol.

2. Eh it'd probably be better to be consistent even if it is the line or make it a server-side config as to what it should be. My main gripe is now you have autoaim to an item which has actual balance implications that need to be coded around, e.g.
* If someone goes "invisible" (e.g. if they go into darkness) you need to break examine otherwise you have a line showing you the entity. Another example is if you're flashed and the control is drawn in screenspace you can still have a line to the entity.
* If an item is picked up you need to break examine or else (like if someone picks up nuke disk in a group of people you know exactly who did it).

With that being said I don't like particularly like the tracking entirely either for balance reasons currently but in my opinion it's slightly better.

JohnGinnane:
>
>
>If someone goes "invisible" (e.g. if they go into darkness) you need to break examine otherwise you have a line showing you the entity. Another example is if you're flashed and the control is drawn in screenspace you can still have a line to the entity.

This is a great point and I'm hoping to handle all of this inside the original `CanExamine()`. Someone mentioned that it is going to be difficult to check for brightness. Do you know if there is a function that returns how lit an entity is?

>
>
>If an item is picked up you need to break examine or else (like if someone picks up nuke disk in a group of people you know exactly who did it).

I should have shown this in the video but I will test this to be sure, but I'm not sure if `CanExamine()` would return true if the entity is in the hands of another player?

metalgearsloth:
> 
> 
> > If someone goes "invisible" (e.g. if they go into darkness) you need to break examine otherwise you have a line showing you the entity. Another example is if you're flashed and the control is drawn in screenspace you can still have a line to the entity.
> 
> This is a great point and I'm hoping to handle all of this inside the original `CanExamine()`. Someone mentioned that it is going to be difficult to check for brightness. Do you know if there is a function that returns how lit an entity is?
> 
> > If an item is picked up you need to break examine or else (like if someone picks up nuke disk in a group of people you know exactly who did it).
> 
> I should have shown this in the video but I will test this to be sure, but I'm not sure if `CanExamine()` would return true if the entity is in the hands of another player?

I don't think there is a way to check brightness currently given all of the lighting calcs are exclusively handled by the gpu (plus I'm very doubtful you'd get this working well server-side) but don't quote me on that.

Isn't CanExamine only called when the entity is initially examined?

I'm not gonna have a sook about it if another maintainer wants to merge it but I'd prefer at least to not have it enabled by default if it is (even if the new version is much fancier).

JohnGinnane:
Currently every `Update()` (soon to be `FrameUpdate()`) I check `CanExamine` so that if line of sight is broken then the line disappears, but the popup remains

Regarding not having it enabled by default, how should I handle this in code? Have it check for a non-existent setting?

metalgearsloth:
> 
> 
> Currently every `Update()` (soon to be `FrameUpdate()`) I check `CanExamine` so that if line of sight is broken then the line disappears, but the popup remains
> 
> Regarding not having it enabled by default, how should I handle this in code? Have it check for a non-existent setting?

Yeah I missed that the first time, I just profiled and it's not hugely expensive (like 10% of the client-side systems so YMMV) though it only handles the LOS and not visibility.

I just read through the conversations on discord and I agree with remie on putting it on a consistent position like the bottom-left, but like I said if you make it a cvar there's no dramas as long as another maintainer merges it.

JohnGinnane:
>
> 
> I just read through the conversations on discord and I agree with remie on putting it on a consistent position like the bottom-left, but like I said if you make it a cvar there's no dramas as long as another maintainer merges it.

If I put the popup on the bottom left of the UI should we still draw a line? By "cvar", do you mean using `IConfigurationManager.GetCVar()`?

PJB3005:
This really feels like something that can and should be implemented purely in content.

Acruid:
This is such a one-off feature that putting it in the base Control and PopupControl classes does not really make sense. Move all this code into content. Ask in discord if you need help.

PR #1441, opened 2020-12-10
FixClipping() now check if entity is deleted (bug fix)

komunre:
I need this fix because when flashbangs broke the window into space it cause error because decompression destroys grenade.

FixClipping was trying to move null parent at 0, 0 position.


PJB3005 (commented):
Shouldn't the code *calling* `CanMove()` in this instance be checking if it's deletes?

komunre:
@PJB3005 `CanMove()` Called by `SharedPhysicsSystem` and `PhysicsManager` code in context of checking collided items. I don't think Entity will check itself for moving possibility (anyways, what's wrong?). If you asking what `Owner.Deleted` do - then no. It get owner of `PhysicsComponent` so it's entity with attached component.

Also i tested it on my clusterbang item.

metalgearsloth:
> 
> 
> Shouldn't the code _calling_ `CanMove()` in this instance be checking if it's deletes?

Yes. CanMove actually used to check for Deleted and I removed it for this reason as CanMove is called in a lot of places and not all of them care.

komunre:
So should i add Deleted check in FixCollide instead? By the way why we need to move deleted objects?

metalgearsloth:
If they're deleted as part of collision behaviors it can happen.

It should probably be checked under the part of SimulateWorld that's calling FixClipping. This is me speculating without debugging it myself where I'd be able to give you a concrete answer.

komunre:
nice, i accidentally deleted branch. Now i need to create master branch PR?

komunre:
I removed Delete check from CanMove and added it to the FixClipping (I always was saying it wrong).

PJB3005 (approved):
Seems good. @metalgearsloth ?

metalgearsloth (approved):

PR #1448, opened 2020-12-15
Cache TryLooseGetType

metalgearsloth:
Roughly 10x speedup on AppearanceComponent when joining saltern.

Before:
image

After:
image


gradientvera (approved):
LGTM. CC @PJB3005 for review & approval though

PR #1452, opened 2020-12-17
Dirty call injector

### LINKED ISSUE #1450: Have an attribute for properties to automatically call dirty when set

PaulRitter:
fixes #1450

checks for a new [Dirty] attribute and injects a Dirty()-Call
image

You can also disable it making an equals check by doing [Dirty(false)]
image


RemieRichards:
Uhhh, shouldn't the value setting come first?

PaulRitter:
wait thats actually a good point

Robust.Build.Injections/DirtyCallInjectionTask.cs
```diff
@@ -0,0 +1,170 @@
...
+                            ilProcessor.Append(ilProcessor.Create(OpCodes.Call, equalsMethod));
+                            ilProcessor.Append(ilProcessor.Create(OpCodes.Brfalse, fourth));
+                        }
+                    }
+                    ilProcessor.Append(ilProcessor.Create(OpCodes.Ldarg_0));
+                    ilProcessor.Append(ilProcessor.Create(OpCodes.Call, internalDirtyMethod));
+                    ilProcessor.Append(first);
+                    ilProcessor.Append(second);
+                    ilProcessor.Append(third);
+                    ilProcessor.Append(fourth);
```
RemieRichards: You probably want this

```csharp
        const storeInField = instr[2];
        ...
        ilProcessor.Append(ilProcessor.Create(OpCodes.Ldarg_0));                   //load 'this'
        ilProcessor.Append(ilProcessor.Create(OpCodes.Ldarg_1));                   //load 'value'
        ilProcessor.Append(storeInField);                                          //store value in field
        ilProcessor.Append(ilProcessor.Create(OpCodes.Ldarg_0));                   //load 'this'
        ilProcessor.Append(ilProcessor.Create(OpCodes.Call, internalDirtyMethod)); //call Dirty()
        ilProcessor.Append(setterReturn);                                          //return
```

* Don't think there's any point in saving the instructions without operands (only the field set is worth saving)
    * There's a readability gain by recreating the other instructions
* Put the Dirty() after the set (else there's no point)
* Better name for the (now-only) instruction we're keeping

PaulRitter: yep, thats what i changed it too, oops :P

i just bundled it with the if check originally and didn't think any further. will update pics of the il in the pr now if you wanna take a look

PaulRitter:
i updated this to use the same workaround as xamlil. this is ready to merge, it has tests & has been tested otherwise

PaulRitter:
as discussed with pjb this needs to be changed from being a buildtask to being an exec that gets called on build

Acruid:
Any update on this?

PaulRitter:
Shouldn't be too hard to convert this to be run as a seperate process (see XamlIL). I just haven't come around to doing it yet due to SerV3

PR #1458, opened 2020-12-20
Sends the AttachedEntity of the PlayerState

daniel-cr:
This should fix the lack of names in the admin menu Player tab and the admin teleport function. Related to [#2438](https://github.com/space-wizards/space-station-14/issues/2438)


PJB3005:
This data should absolutely not be sent to clients. The player panel should be rewritten instead.

PR #1461, opened 2020-12-21
Replace MaybeNullWhen(false) with NotNullWhen(true) in AppearanceComponent

DrSmugleaf:
Because Rider yelled at me today


metalgearsloth (approved):

metalgearsloth:
We can still merge on RT with versioning right.

gradientvera (approved):

PR #1467, opened 2020-12-24
New ControlFocusExited hook on Control

### LINKED ISSUE #1449: Controls don't receive KeyBindUp when using multiple mouse buttons
If you LMB down on a control, then drag to a different control and RMB down and then LMB up, the original control doesn't receive the LMB up, nor is there an event it can listen for to know it has "lost focus" (the "FocusExited" event doesn't fire since that's only for keyboard fous).

This is most noticeable when trying to drag an action and pressing RMB while dragging over another control, the drag shadow becomes "stuck" until you try dragging something else. This is because the action slot never receives the keyup event.

chairbender:
Fixes #1449 , but not quite in the way the ticket suggests.

I initially tried an approach where you could have multiple controls with "focus" simultaneously. This would ensure that the control always received the eventual KeyBindUp when it got a KeyBindDown. Also all the currently focused controls would receive mouse hover events (which could be multiple controls).

I ended up not really liking this approach, not only was it complicated, but it also allowed for weird UI behavior like letting you drag / drop something with LMB while simultaneously right clicking other UI elements. I couldn't see a reason this capability would ever be necessary and it didn't feel like it was worth the complexity, though I did get it working (see https://github.com/chairbender/RobustToolbox/tree/1449-keybindup if interested).

Instead, I opted to simply add a hook to let controls know when they've lost focus, which can happen when, for example, you press LMB down on one control, then drag, then RMB down on a different control while still holding LMB down. Controls which care about this (generally controls which have some drag / drop logic or special keybindup logic) can implement that method. This fi

The downside of this approach is, if a dev isn't aware of this quirk and they only implement KeyBindUp but not ControlFocusExited, their drag/drop can become "stuck" (because the control will never get the KeyBindUp if it lost focus). My alternate approach did not have that downside - KeyBindUp was guaranteed to be fired so there was no need for this hook (but suffered all the other issues I mentioned above). Since it seems like there are only a few controls that would actually care about ControlFocusExited I don't think that's a big problem, and overall I think this approach is simpler and better and fits more naturally into our existing UI system's concept of "control focus", rather than complicating it by allowing multiple things to be "focused".

EDIT: see this follow up ticket where we can leverage this capability to fix other controls affected by this problem https://github.com/space-wizards/RobustToolbox/issues/1468

### DIFF
--- a/Robust.Client/Interfaces/UserInterface/IUserInterfaceManager.cs
+++ b/Robust.Client/Interfaces/UserInterface/IUserInterfaceManager.cs
@@ -21,7 +21,21 @@ namespace Robust.Client.Interfaces.UserInterface
         /// </summary>
         Stylesheet? Stylesheet { get; set; }
 
+        /// <summary>
+        /// A control can have "keyboard focus" separate from ControlFocused, obtained when calling
+        /// Control.GrabKeyboardFocus. Corresponding events in Control are KeyboardFocusEntered/Exited
+        /// </summary>
         Control? KeyboardFocused { get; }
+
+        /// <summary>
+        /// A control gets "ControlFocused" when a mouse button (or any KeyBinding which has CanFocus = true) is
+        /// pressed down on the control. While it is focused, it will receive mouse hover events and the corresponding
+        /// keyup event if it still has focus when that occurs (it will NOT receive the keyup if focus has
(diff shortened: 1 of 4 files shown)


metalgearsloth (approved):

metalgearsloth:
Gonna merge in a week if no objections

PR #1469, opened 2020-12-24
Fix component dependency tests when checking for thrown exceptions

DrSmugleaf:


PJB3005 (commented):
The problem is that this causes "no exception thrown" to get thrown if an exception *did* get thrown of the wrong type.

DrSmugleaf:
This was fixed at some point.

PR #1471, opened 2020-12-25
Serialization v3

### LINKED ISSUE #134: Tool for verifying YML templates against the assemblies.
Considerations:

* Tool needs to be able to access some form of class or type dump of the Services, GameObjects and Content assemblies. No idea how this'll be done? Can Mono and Roslyn emit this metadata?

### LINKED ISSUE #392: No way to get initial values from prototypes
We need a more generalized method to get properties that a prototype has, possibly by initializing one of the prototypes as an entity and taking it out of the process lists.

This became a problem for me recently as the entityspawnpanel displays the sprite just by the one hardcoded way we have to get the spritename, and then using getsprite for itself. It this particular example would be far easier if I just used the entityprototype's render method and passed it the X and Y it needs to render to. 

But yeah the primary problem is that components may load nondefault parameters for their components which we have no way currently to read.

### LINKED ISSUE #891: Serialization v3
The current serialization system, `ExposeData`, has a lot of shortcomings:

1. Serialization happens *on* an instance of a component (or somehow nested) in most cases. This makes it basically impossible to sanely validate prototypes without instantiating them, or doing any sort of automatic operation on prototypes at all for that matter.
2. Data is continuously re-parsed from YAML. This is incredibly slow and also means we're forced to keep the YAML data round, which is also very memory intensive (Most of the server managed heap memory while I am writing is in fact the parsed YAML tree). There are some mechanisms in the current exposedata system to do caching. It's all ugly.

I hereby propose the following for serialization v3:

Data is (de) serialized into an "intermediate" parsed class. This class can then be cheaply copied (deep copies perhaps necessary for certain components) into the component when the component is instantiated. Being a separate class with little logic, this data can then also be acted upon more easily by tooling to automatically very whether it is indeed sane.

To avoid drowning in boilerplate, it is probably best if there were attributes that could be applied to various things like component properties to automate all of this.
Various things like dynamic code generation can also be used to improve performance of these automatic attributes.

### LINKED ISSUE #2920: Fix remaining DataReadWriteFunction
Looks like the pr got most of them but there's still some floating around, e.g.

`serializer.DataField(this, mod => SpeedModifier, "speed", 1);`

Should be
`serializer.DataField(this, mod => mod.SpeedModifier, "speed", 1);`

PaulRitter:
requires https://github.com/space-wizards/space-station-14/pull/2975

TODO:
- nuke deepclone, expand dataclass system to EVERYTHING. NO MORE EXPOSEDATA 
- write guide/xmldocs
- required option
- fix case: inheriting comps define same tag
- inheriting generic component's dataclasses (BaseNetConnectorComponent) OH GOD OH FUCK I GOTTA GENERATE GENERIC DATACLASSES HOW THE FUCK  !!! THROW ERROR ON GENERIC YAMLFIELD FOR NOW !!!

things to be optimized
- customfield caching
- field setting/getting
- deepclone fallback
- generate pushinheritance & write snowflake custom dataclass handling
- everything basically

### Analyzers (some could be done at runtime tbh)
- [ERR] inheriting comps define same tag
- [WARN] Error if CustomYamlFields in CustomDataClass are not initialized with null (or not even nullable)
- [WARN] [CustomDataClass] with no [CustomYamlField]
- [WARN] Every Class which has IDeepClone in its inheritance should implement it
- [WARN] fields being copied in DeepClone() without call to IDeepClone.CloneValue()
- [ERR] [CustomYamlField] should always have a counterpart in the custom dataclass
- [ERR] Component with [CustomYamlField] needs to have a [CustomDataClass]

### How to handle generic components
only create a generic dataclass if a yamlvalue is of the generic type, take all generic types for now and only use the one, its just easier than picking them, especially for inheritance logic
[helpful link for roslyn](https://johnkoerner.com/csharp/creating-an-analyzer-based-on-generic-type-parameters/)

fixes #134
fixes #392
fixes #891
fixes https://github.com/space-wizards/space-station-14/issues/2920

might include #1064 while im at it


PaulRitter:
closing due to the new draft rule

PR #1472, opened 2020-12-26
Fixes key combinations blocking single keys.

### LINKED ISSUE #2782: Left control key issue
When I hold my left control key, I can't move to my left. However, I can move to my right, up, and downwards. The link I'll attach shows me trying to move left when I'm not holding left control and then when I am holding left control down.

https://gyazo.com/272cabbfc6dc8e4e6d549c0476bc3b8e

Radrark:
Fixes space-wizards/space-station-14#2782

This changes the InputManager so that, when a key combination is pressed, the manager adds its subcombinations bindings to the list of binds to be triggered. For example, when you press Ctrl+A, the input manager adds TextSelectAll and MoveLeft to the list, instead of just TextSelectAll. So, if TextSelectAll doesn't get handled, MoveLeft is then triggered.
Also, I fixed the PackedIsSubPattern method but I noticed it isn't useful now. Anyways, I left it there.
I'm fairly new to C# so maybe this isn't the right way to fix the issue idk.

### DIFF
--- a/Robust.Client/Input/InputManager.cs
+++ b/Robust.Client/Input/InputManager.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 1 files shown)


ShadowCommander:
Check if something like this PR works before changing stuff inside InputManager, since I'm pretty sure those checks are needed.
space-wizards/space-station-14#2534

Radrark:
Ok so this causes another issue. Should I close this PR and open another when I actually fix the issue?

PR #1473, opened 2020-12-26
Adds new keybind property: AllowSubCombs

### LINKED ISSUE #2782: Left control key issue

Radrark:
This new keybind property makes a keybind, bound to a key combination, trigger any subcombination of it. For example, TextSelectAll is bound to Ctrl+A, if you add "allowSubCombs = true" to it, then, when the combination gets pressed, the input manager will add any bind bound with A (In this case MoveLeft) to the trigger list, so, if TextSelectAll doesn't get handled then MoveLeft will get triggered. This fixes space-wizards/space-station-14#2782 and similar issues, for example, pressing Ctrl+C when you are not writing in the chat doesn't open the character menu.

### DIFF
--- a/Robust.Client/Input/InputManager.cs
+++ b/Robust.Client/Input/InputManager.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
@@ -134,7 +134,8 @@ namespace Robust.Client.Input
                     Priority = p.Priority,
                     Type = p.BindingType,
                     CanFocus = p.CanFocus,
-                    CanRepeat = p.CanRepeat
+                    CanRepeat = p.CanRepeat,
+                    AllowSubCombs = p.AllowSubCombs
                 }).ToArray();
 
             var leaveEmpty = _modifiedKeyFunctions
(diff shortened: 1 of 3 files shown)


ShadowCommander:
Can you check if space-wizards/space-station-14#2782 can be solved by having the TextSelectAll keybind handler return false if KeyboardFocus is false? That's how it's supposed to work and it'd use the existing fall through instead of having to implemente this new one on all the keybinds.

Radrark:
> Can you check if [space-wizards/space-station-14#2782](https://github.com/space-wizards/space-station-14/issues/2782) can be solved by having the TextSelectAll keybind handler return false if KeyboardFocus is false? That's how it's supposed to work and it'd use the existing fall through instead of having to implemente this new one on all the keybinds.

Yeah I checked that. Even if TextSelectAll returns false MoveLeft doesn't get triggered. The cause of the bug is the input manager mapping Ctrl+A to TextSelectAll, instead of, Ctrl+A to TextSelectAll and MoveLeft. So, initially I tried changing the input manager such that every combination also maps to all its subcombinations bindings, but that caused both TryPullObject and Use to get triggered because they are server binds. Therefore, I figured this solution where only some key combinations also allow subcombinations to get triggered. I think adding the property only to client binds should not break anything.

PR #1475, opened 2020-12-28
Autogenerates NetIDs

PaulRitter:
requires https://github.com/space-wizards/space-station-14/pull/2856


PJB3005 (changes requested):
Same problem as with your ignored components generator. This does **way** too much stuff in the source generator.

PaulRitter:
new iteration sees to just hardcode a prefixnumber for each project, fine? @PJB3005 that way, we can easily avoid the big overhead & still generate the netids

Issue #1255, opened 2020-08-23
Dirty() for owner player only


metalgearsloth:
Or an equivalent.

Some content-side components such as statuseffects and overlays only need to send their data to the player attached to the entity (to reduce bandwidth and reduce data exposed for cheating) rather than every single player in PVS range. Currently this is boilerplated in where needed.

PJB3005:
It would probably be best to have an overload of `GetComponentState` that takes in the `IPlayerSession` it's being generated for.

metalgearsloth:
How about a bool on a component that the alternative GetComponentState method should be called and pvs then chooses which one to use?

metalgearsloth:
Done in https://github.com/space-wizards/RobustToolbox/pull/1569

PR #1478, opened 2020-12-31
Pvs refactor

metalgearsloth:
I combined the DynamicTrees under the EntityManager with gridtilelookups (mainly using the existing gridtilelookup system) to make a new generic SharedEntityLookupSystem that also handles invalid grids. I also removed much of the storing of entity locations out of EntityManager for obvious reasons.

Benchmarks (me running across the station back and forth opening doors):
Old perf:
old_perf

New perf:
new_perf

Old allocs:
old_allocs

New allocs:
new_allocs

Performance:
* Physics is faster as UpdateEntityTree was removed (the updates to DynamicTree were expensive)
* Spawning is faster as the 2 calls to UpdateEntityTree were removed
* PVS is faster

Should be easy to add multi-viewport support now.

Allocations are lower though not as low as I'd like them to be. Also has the groundwork for shuttles to work.

Overall memory is about the same, at least for 1 player.

I added a command `chunkupdates` that shows whenever a chunk is made dirty. I also added `lookupnodes` to see what particular entities are on a tile client-side.

Here's what the node one looks like:

image

Still todo in future PRs:
* Potentially do some level of visibility culling for PVS
* Override PVS for lights (shouldn't be too hard to do now)
* Suss out not sending every entity upon connection (currently this is way too slow for the client spawning entities so I haven't changed it). If you make a new map it will do this (so moving around for the first time will be expensive) though the current setup does this regardless so I didn't consider it a regression.


metalgearsloth:
Should be good for a review though I'd like to be back at my desktop first so I'll leave it as draft for the time being.

metalgearsloth:
Should be good for review now, "should" be relatively bug-free. I tried to listen to entity parenting events for the lookup updates but I was getting some bugs (e.g. EntParentChanged isn't called if the parent is deserialized but then when I fixed that another bug popped up etc.) so I was listening to container events instead.

It might also fix the transform parenting bug given PVS should be a lot easier to read. I've outlined in comments how to do the TODOs (multi-viewports, only sending state to specific player).

Could do with more optimisation but should be good for now I think.

metalgearsloth:
Requires https://github.com/space-wizards/space-station-14/pull/2906

metalgearsloth:
Just remembered I need to add 2 things so making draft for now.

metalgearsloth:
Should be good for review again. At some stage I'll probably look at using DynamicTree again though using a naive implementation meant that tile lookups were way too slow.

metalgearsloth:
Fixes https://github.com/space-wizards/RobustToolbox/issues/1255 as of latest commit

ghost:
I don't know much about how PVS works right now but I will still review with what I know.
Still should have someone more experienced (cough pjb) review the real tough pvs stuff.

metalgearsloth:
Should also mention here as well: I hadn't done any culling for the client for entities out of range. There were some suggestions made (e.g. having the server send the MapId as nullspace or having the client do it itself) but nothing was settled on.

Currently only moving entities get culled out of pvs range due to NaNs being sent (so stuff like walls etc. still exist client-side).

Ideally if culling were to be added in this PR then you'd get fixed viewport in first because at 4k I can see double pvs range so it'll look even jankier.

Robust.Client/GameObjects/EntitySystems/EntityLookupSystem.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            foreach (var entity in _nodeEntities)
+            {
+                _control.AddChild(new Label
+                {
+                    Text = $"uid: {entity.Uid}, name: {entity.Name}"
+                });
+            }
+
+            LayoutContainer.SetPosition(_control, mousePos + new Vector2(0, 20));
```
PJB3005, line 86: You need to run `mousePos` through `IUserInterfaceManager.ScreenToUIPosition` to account for UI scaling.

Robust.Server/GameObjects/EntitySystems/EntityLookupSystem.cs
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        internal PlayerLookupChunks? GetPlayerLastSeen(IPlayerSession session)
+        {
+            return !_lastSeen.TryGetValue(session, out var chunks) ? null : chunks;
+        }
+    }
+}
+
+internal sealed class PlayerLookupChunks
```
PJB3005, line 167: This isn't in a namespace?

Robust.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -146,558 +145,155 @@ public override IEntity SpawnEntityNoMapInit(string? protoName, EntityCoordinate
...
-        }
-
-        private readonly Dictionary<IPlayerSession, SortedSet<EntityUid>> _seenMovers
-            = new();
-
-        // Is thread safe.
-        private SortedSet<EntityUid> GetSeenMovers(IPlayerSession player)
-        {
-            lock (_seenMovers)
+            if (session != null)
```
PJB3005: To make sure I'm understanding the code correctly: This is to make sure switching to/from `net.pvs` mid round is possible, right? 

If so, might want to elaborate in a comment.

PJB3005 (commented):
I couldn't get this reviewed 100% and I'm getting really sleepy but these are some things I noticed already. More to come tomorrow probably.

Robust.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -146,558 +145,155 @@ public override IEntity SpawnEntityNoMapInit(string? protoName, EntityCoordinate
...
+        ///     AKA PVS (potentially visible set). Get all relevant ComponentStates in range of us.
+        /// </summary>
+        /// <param name="fromTick"></param>
+        /// <param name="currentTick"></param>
+        /// <param name="player"></param>
+        /// <param name="range"></param>
+        /// <returns></returns>
+        public List<EntityState>? GetEntityStates(GameTick fromTick, GameTick currentTick, IPlayerSession player, float range)
+        {
+            // Old PVS used to just get all for no session...
```
PJB3005, line 189: I think you mean entity.

Robust.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -146,558 +145,156 @@ public override IEntity SpawnEntityNoMapInit(string? protoName, EntityCoordinate
...
+            // TODO: I think there may still be the issue of parents not being sent
+            // If there is then to solve this just have each entity retrieved by the lookup
+            // also check if its parent has been retrieved and if not then also retrieve it.
 
-                var oldState = (TransformComponent.TransformComponentState) state.ComponentStates[idx];
-                var newState =
-                    new TransformComponent.TransformComponentState(Vector2NaN, oldState.Rotation, oldState.ParentID);
-                state.ComponentStates[idx] = newState;
+            // Sort for the client.
+            entityStates.Sort((a, b) => a.Uid.CompareTo(b.Uid));
```
PJB3005, line 258: Why does the client need them sorted?

Robust.Server/GameObjects/ServerEntityManager.cs
```diff
@@ -879,83 +356,82 @@ public override void Startup()
...
+            {
+                entitySession = entity.PlayerSession();
+            }
+
             foreach (var comp in compMan.GetNetComponents(entityUid))
             {
                 DebugTools.Assert(comp.Initialized);
 
+                // If only the relevant person gets the state then we'll filter it out.
+                if (comp.PlayerOnlyState && entitySession != session) continue;
```
PJB3005: Please add a check here if `session` is null to always send (PVS disabled scenario) so that when PVS is disabled all data is sent to all clients.

This could be important for debug tools or replays in the future.

Robust.Shared/Interfaces/GameObjects/IComponent.cs
```diff
@@ -34,7 +34,12 @@ public interface IComponent : IEntityEventSubscriber
         ///     Whether the Owner has been paused.
         /// </summary>
         bool Paused { get; }
-        
+
+        /// <summary>
+        ///     If set to true only an attached player will receive states.
+        /// </summary>
+        bool PlayerOnlyState { get; set; }
```
PJB3005: I understand the incentive but I feel like this might just be a little too crude to really be all that useful currently?

metalgearsloth: I'm happy to remove though I wasn't sure how to do stuff like "Flashable / Alert states only go to the specific player" for example. Overall it's probably not a big bandwidth saving anyway given overlays were removed as well so I'll just dump it.

metalgearsloth:
Gonna mark as draft till I have the server sending nullspace maps.

metalgearsloth:
Closing in favour of acruid's who is smarter and wiser.

PR #1480, opened 2021-01-02
Cache enum references

metalgearsloth:
It's like 5-10% of entity spawning client-side (at least in debug).


PJB3005:
Thank you GitHub for your well-designed UI that doesn't jump around the place.

PR #1482, opened 2021-01-02
Replaces AnchoredChanged C# event with a ComponentMessage

collinlunn:
Required by space-wizards/space-station-14#2905
space-wizards/space-station-14#2905 also requires this PR to still build


metalgearsloth (approved):
Waiting on content before merging.

PJB3005:
I brought the event back as obsolete so that I don't have to do an annoying dance to keep content compatible.

PR #1484, opened 2021-01-05
Add RadioOptions.cs

Rockdtben:
I'm magic.

rt0nhZEV1K

### DIFF
--- a/Robust.Client/UserInterface/Controls/RadioOptions.cs
+++ b/Robust.Client/UserInterface/Controls/RadioOptions.cs
@@ -1,16 +1,19 @@
 using Robust.Client.Graphics;
 using System;
+using System.Linq;
 using System.Collections.Generic;
 using static Robust.Client.UserInterface.Controls.BaseButton;
 
 namespace Robust.Client.UserInterface.Controls
 {
-    public class RadioOptions : Control
-    {
-        public enum RadioLayout { Horizontal, Vertical }
+    public enum RadioOptionsLayout { Horizontal, Vertical }
 
-        private readonly List<ButtonData> _buttonDataList = new();
-        private readonly Dictionary<int, int> _idMap = new();
+    public class RadioOptions<T> : Control
+    {
+        private int internalIdCount = 0;
+
+        private readonly List<RadioOptionButtonData<T>> _buttonDataList = new();
+        //private readonly Dictionary<int, int> _idMap = new();
         private ButtonGroup _buttonGroup = new();
         private Container _container;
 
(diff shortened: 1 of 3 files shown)


Robust.Client/UserInterface/Controls/RadioOptions.cs
```diff
@@ -0,0 +1,253 @@
...
+        public enum RadioLayout { Horizontal, Vertical }
+
+        private readonly List<ButtonData> _buttonDataList = new();
+        private readonly Dictionary<int, int> _idMap = new();
+        private ButtonGroup _buttonGroup = new();
+        private Container _container;
+
+        public string ButtonStyle = "";
+        public string FirstButtonStyle = "";
+        public string LastButtonStyle = "";
```
gradientvera: Small style nitpick, but these could use `string.Empty` instead.

gradientvera:
```suggestion
        public string ButtonStyle = string.Empty;
        public string FirstButtonStyle = string.Empty;
        public string LastButtonStyle = string.Empty;
```

Robust.Client/UserInterface/Controls/RadioOptions.cs
```diff
@@ -0,0 +1,253 @@
...
+                case RadioLayout.Horizontal:
+                default:
+                    _container = new HBoxContainer();
+                    break;
+            }
+
+            this.AddChild(_container);
+        }
+
+        public void AddItem(Texture icon, string label, int? id = null)
```
ghost: "icon" field unused here.

Robust.Client/UserInterface/Controls/RadioOptions.cs
```diff
@@ -0,0 +1,253 @@
...
+        {
+            AddItem(label, id);
+        }
+
+        public void AddItem(string label, int? id = null, Action<ItemSelectedEventArgs>? itemSelectedAction = null)
+        {
+            if (id == null)
+            {
+                id = _buttonDataList.Count;
```
ghost: I don't like how the ID's work here in general, might be my preference but I would much rather keep ID's separate entirely from array size and have it be a randomly generated int or something, then change "void AddItem()" to "int AddItem()", where it returns the randomly generated int ID. Something along those lines.

Also, actual issue here too: Say that you don't care about ID's and just add 4 different radio options. They get the ID's 0 1 2 and 3. You remove ID number 2, so you have 0 1 3. You add another, and it fails since it tries to add 3 to 0 1 3.

Robust.Client/UserInterface/Controls/RadioOptions.cs
```diff
@@ -0,0 +1,253 @@
...
+        public void SetItemText(int idx, string text)
+        {
+            var data = _buttonDataList[idx];
+            data.Text = text;
+            data.Button.Text = text;
+        }
+
+        public class ItemSelectedEventArgs : EventArgs
+        {
+            public RadioOptions Button { get; }
```
ghost: Why is this called "Button"?

Robust.Client/UserInterface/Controls/RadioOptions.cs
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        public void SetItemDisabled(int idx, bool disabled)
+        {
+            var data = _buttonDataList[idx];
+            data.Disabled = disabled;
+            data.Button.Disabled = disabled;
+        }
+
+        public void SetItemId(int idx, int id)
```
ghost: Being able to set an button index's ID is kind of weird. I would much rather prefer SetItemId(int oldID, int newID), but that's up to you.

ghost:
Other than the issues I found it seems good. One personal preference is that would organize the file in a more connected way, like placing SetItemMetadata() and GetItemMetadata() next to each other vertically in the file, but that's just my own way of doing it.

PR #1496, opened 2021-01-10
Fixes sprite scale not being taken into account.

gradientvera:
Only applies scale if length of scale vector not close or equal to 1.
Requesting review from PJB just in case.

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -992,8 +992,7 @@ namespace Robust.Client.GameObjects
             Matrix3.Multiply(ref mRotation, ref mOffset, out var transform);
 
             // Only apply scale if needed.
-            if(!MathHelper.CloseTo(Scale.Length, 1f))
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -991,6 +991,10 @@ public RSI.StateId LayerGetState(int layer)
             var mRotation = Matrix3.CreateRotation(angle);
             Matrix3.Multiply(ref mRotation, ref mOffset, out var transform);
 
+            // Only apply scale if needed.
+            if(!MathHelper.CloseTo(Scale.Length, 1f))
```
PJB3005: Comparing with length makes no sense here? You can have plenty of vectors that are length 1 but aren't an identity scale.

PJB3005 (changes requested):

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
gradientvera: Fixed, turns out I suck at maths

Issue #685, opened 2018-09-09
View Variables should support editing enums.


PJB3005:
Probably with some kinda drop down.

Acruid:
Basic support for editing VV Enums was added in https://github.com/space-wizards/RobustToolbox/commit/19d7cd445541cf12c5b9ebe388a91f5bfde9a722. This currently uses a textbox, it would be nice to improve it with a combo-box UI widget.

PR #1503, opened 2021-01-13
VV Enum

kira-er:
grafik
grafik


PaulRitter (approved):
good first pr :+1:

Robust.Client/ViewVariables/Editors/VVPropEditorEnum.cs
```diff
@@ -1,60 +1,52 @@
...
 
-            var hBox = new HBoxContainer
+            var optionButton = new OptionButton();
+            foreach (var val in enumList)
             {
-                CustomMinimumSize = new Vector2(200, 0)
-            };
+                if(val == null)
+                    continue;
+                optionButton.AddItem(val.ToString()!, (int)val);
```
metalgearsloth: Suppressing null here seems bad?

Robust.Client/ViewVariables/Editors/VVPropEditorEnum.cs
```diff
@@ -1,60 +1,52 @@
...
     {
         protected override Control MakeUI(object? value)
         {
+            if (value is ViewVariablesBlobMembers.ServerValueTypeToken typeToken)
+            {
+                return new Label
+                {
+                    Text = typeToken.ToString()
+                };
+            }
```
PJB3005: Shouldn't this be impossible since `VVPropEditorEnum` only gets picked if the type is an enum, which `ServerValueTypeToken` isn't?

PJB3005 (commented):
Please fix nullability errors

Robust.Client/ViewVariables/Editors/VVPropEditorEnum.cs
kira-er: during testing the method received a servervaluetypetoken when trying out server-only enums

PJB3005: I tried locally and couldn't seem to reproduce it? Do you have a test case?

PaulRitter: managed to reproduce locally
image
image
(enum not annotated)

PaulRitter:
fixes #685

PR #1504, opened 2021-01-14
Polar Coordinates: simplify maths, any-angle occluders are no longer evil

20kdc:
Someone please performance-test this, but I think the fragment shader code simplifications speak for themselves.

@Radrark found the "line in polar coordinates" equation I needed and a diagram that explained the variables.

That equation was essentially the missing piece to the whole thing.

Expected visual effect: None, outside of situations that SS14 doesn't currently support or that are arguably bugs themselves (rotated occluders)

### DIFF
--- a/Robust.Client/Graphics/Clyde/Shaders/shadow-depth.frag
+++ b/Robust.Client/Graphics/Clyde/Shaders/shadow-depth.frag
@@ -1,24 +1,16 @@
 
-// xy: A, zw: B
-varying highp vec4 fragPos;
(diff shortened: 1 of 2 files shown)


metalgearsloth:
Does this mean occluders work for rotated eye now?

20kdc:
@metalgearsloth They are not likely to be more broken than they were before polar coordinates was introduced. I did try and it appears the lighting buffer is not of the correct size, and there appears to be some sort of angular snapping below 8-9 degrees eye rotation (NOT lighting related)

PR #1506, opened 2021-01-15
Make the lighting manager much more configurable, including a console lock

20kdc:
+ Ambient light can be set
+ Added documentation comments to lighting manager properties
+ Added console access lock (makes togglelight/etc. NOP)
+ Added way to turn on/off lighting without interfering with hard FOV

This allows for lighting to be used in various ways depending on what kind of game content is and what it needs.

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
@@ -30,7 +30,6 @@ namespace Robust.Client.Graphics.Clyde
         // Horizontal width, in pixels, of the shadow maps used to render FOV.
         // I figured this was more accuracy sensitive than lights so resolution is significantly higher.
         private const int FovMapSize = 2048;
-        private const int MaxLightsPerScene = 128;
 
         private ClydeShaderInstance _fovDebugShaderInstance = default!;
 
@@ -85,11 +84,14 @@ namespace Robust.Client.Graphics.Clyde
         private ClydeTexture FovTexture => _fovRenderTarget.Texture;
         private ClydeTexture ShadowTexture => _shadowRenderTarget.Texture;
 
-        private readonly (PointLightComponent light, Vector2 pos)[] _lightsToRenderList
-            = new (PointLightComponent light, Vector2 pos)[MaxLightsPerScene];
+        private (PointLightComponent light, Vector2 pos, float distanceSquared)[] _lightsToRenderList = default!;
 
         private unsafe void InitLighting()
         {
+            // Ensure _lightsToRenderList is ready.
+            MaxLightsPerSceneChanged(_maxLightsPerScene);
+
+            // Other...
             LoadLightingShaders();
 
             {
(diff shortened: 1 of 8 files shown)


moonheart08:
Could you possibly also allow content to adjust the light cap?

20kdc:
I'm not sure that's a great idea, for performance control reasons. I could do it, but, hm.

moonheart08:
@20kdc the reason _is_ performance control, to allow the player to adjust the limit in settings.

20kdc:
Hmm, so that's not really a lighting manager thing so much as a setting thing. Will do.

20kdc:
Just to be clear, while I left the default at 128 because that's what it's at presently, 64 lights works just fine with 1080p with the new algorithm.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -501,26 +504,69 @@ private void DrawLightsAndFov(Viewport viewport, Box2 worldBounds, IEye eye)
...
+                    // If that fails, start stealing slots from faraway lights.
+                    if (distanceSquared >= state.maximumDistanceSquared)
+                    {
+                        // It's not possible to find such a light.
+                        return true;
+                    }
+                    // It is possible: Which is it?
+                    var success = false;
+                    float successTarget = distanceSquared;
+                    for (var newId = 0; newId < _maxLightsPerScene; newId++)
```
PJB3005: Isn't this O(n^2)?

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -959,6 +1005,12 @@ protected override void LightmapDividerChanged(int newValue)
             RegenAllLightRts();
         }
 
+        protected override void MaxLightsPerSceneChanged(int newValue)
+        {
+            _lightsToRenderList = new (PointLightComponent light, Vector2 pos, float distanceSquared)[newValue];
```
PJB3005: This gets called every frame so this alone is 3 KiB of allocs per frame. Not good.

PJB3005 (commented):
Not sure I like the sorting thing. It's totally possible to fix the MaxLightsPerScene limit by just doing multiple passes if we go over. I'd much prefer if that was done instead.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
20kdc: Wait, *how* does this get called every frame? It's a cvar onchanged callback.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
20kdc: Yup, it is. Honestly my concern with multiple passes is that the current mechanism acts as a hard limiter on the amount of performance damage that can be done by point lights, though it would fit in really nicely with the work for stencil shadows (that effectively has a maxlightsperscene of 8 and performs the same multi-pass work you''re suggesting)

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
PJB3005: Wait, I misread. Nevermind.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
20kdc: Ok, so I found another problem. You know how you really *don't* want allocs-per-frame? Well, in order to do any pre-sorting to prevent this whole O(n^2) thing, additional memory is required equal to the maximum number of lights that could ever be seen, ever.

PJB3005: Yes, this is one of the reasons I don't like the "MaxLightsPerScene is a perf thing" stuff.

Just over allocate a massive buffer and if you run out of EVEN THAT then whatever and take the allocs. Shouldn't ever happen in a real scenario I would think.

20kdc: Done, though without the "take the allocs". Instead I made the new fixed constant light storage limit 2048 lights, which is hopefully sufficiently stupidly high.

PR #1514, opened 2021-01-27
Reorganizes NumericsHelpers into different files using partial classes.

gradientvera:
Functionality stays the same, but now NumericsHelpers is a partial class divided in 8 different files, which should make maintaining it easier.


PaulRitter (approved):
merging in a week if noone objects

PR #1519, opened 2021-01-31
Optimise showbb

metalgearsloth:
GetCollidingEntities is faster than using EntityQuery on every component.


PaulRitter (approved):
real bigbrain hours

PR #1520, opened 2021-01-31
Formatted Text

Acruid:
- [ ] Adds a Fluent API for constructing formatted text strings.
- [ ] Adds a parser for parsing formatted text strings.
- [ ] Write a hackmd spec for formatted text strings.


PJB3005:
If we are going through with this then this should probably replace the existing `FormattedMessage`, no? Or does this just map onto that.

Acruid:
I decided not to do this.

PR #1528, opened 2021-02-03
Viewport Improvements

### LINKED ISSUE #1170: Clyde will straight up die when rendering more than 65535 vertices in a single render pass.
Making multiple vertex buffers is hard, ok?

gradientvera:
@20kdc coded most of this.

**DO NOT MERGE WITHOUT @PJB3005's  EXPLICIT REVIEW AND APPROVAL.**

Current issues with this:
- Hiding UI hides the main viewport as well...
- Debug overlays are screwed? I think?
- Skybox is not drawn on the viewport
- ???? probably more

---
- Adds ViewportContainer control.
- ViewportContainer can set viewport resolution.
This means you could set the viewport resolution to half, and double the eye's zoom for **FREE** performance gains!
- Main viewport is now contained within a control.
Content can access this control and reparent it as needed, therefore allowing things like this: image _BYOND layout, anyone?_

Fixes #1528
Fixes #1170


PJB3005 (commented):
The basic structure needs good but it just needs more work to make it, well, workable for production:

1. For rendering viewports when UI is disabled, I'd say it should be possible to mark viewports as "important" and if the UI system is disabled Clyde will try to shortcut render those where they "would be" from the UI system.
2. Overlays should be able to be specified on a per-viewport basis and actually renderered as part of the, well, viewport render. Also make parallax a proper viewport overlay instead of the current "render below everything" and just have UIs like lobby render parallax manually. Debug overlays should probably be rendered after UI or something (maybe make them a UI control, tbh??)

Robust.Client/Graphics/Clyde/Clyde.HLR.cs
```diff
@@ -73,15 +73,10 @@ public void Render()
...
+            foreach (var weak in _viewports.Values)
             {
-                var handle = _renderHandle.DrawingHandleScreen;
-                var tex = _mainViewport.RenderTarget.Texture;
-
-                handle.DrawTexture(tex, (0, 0));
-                FlushRenderQueue();
+                if(weak.TryGetTarget(out var viewport))
+                    RenderViewport(viewport);
             }
```
PJB3005: Maybe ignore viewports that are invisible in the UI system?

Issue #1225, opened 2020-08-14
Refactor IClientCommand to use a separate "output" parameter.


PJB3005:
Right now the player session passed is `null` to indicate "it's actually the server console". On top of generally being really awful (using nulls like that...), it means that commands like `sudo` can't show output to the admin running the command.

Example of the proposed API:

```cs
    internal sealed class FooCommand : IClientCommand
    {
        public string Command => "foo";
        public string Description => "Does things";
        public string Help => "foo";

        public void Execute(IConsoleShell shell, IConsoleOutput output, IPlayerSession? player, string[] args)
        {
			var name = player?.Name ?? "server console";

			output.Send($"Hi, {name}");
        }
    }
```

DTanxxx:
Hi Project Managers,

I am keen to work on this issue. Regarding the issue itself, I have a few questions to help clarify some things:
- when commands like `sudo` are used and they can't show output, is that because they only show output if IPlayerSession is not null (otherwise the method returns without outputting anything)?
- since IConsoleShell can already show text output with SendText(), why do we need another IConsoleOutput to show the output?

Thank you for your time.

PJB3005:
Sorry for the late reply.

> when commands like `sudo` are used and they can't show output, is that because they only show output if IPlayerSession is not null (otherwise the method returns without outputting anything)?

> since IConsoleShell can already show text output with SendText(), why do we need another IConsoleOutput to show the output?

`sudo` specifically executes a command as the server console, from your client. The problem is that to actually be "the server console" you HAVE to, well, not be a player. And the command system is currently set up such that the output target and "who is running this" are one and the same. This is done by assuming that if the player is null, it's the server console, and the output methods work like this aswell. It is not possible to say "this should be executed as the server console" AND "send the output to this player" at the same time. THAT needs to be fixed.

Acruid:
#1513 redesigned the API around the execute function, and #1530 resolved the sudo output problem. This issue is either obsolete or resolved.

PR #1530, opened 2021-02-04
Makes sudo command show output to user.

PJB3005:
For all intents and purposes means #1225 is fixed.


Acruid (approved):
This is almost exactly the same implementation I came up with. Good to merge.

PR #1531, opened 2021-02-04
Fix loadbp on paused maps

metalgearsloth:


metalgearsloth:
If the grid was being loaded onto a paused map the entities weren't being paused (I regret making paused a settable property slightly more).

DrSmugleaf (approved):

PR #1536, opened 2021-02-05
Post shader will use real sprite bounding box

Macoron:
- Sprite can calculate its world-space bounding box
- Works with any scale, offset or rotation
- Works with multiple layers sprites. BB will be based on layers union size
- PostShader will use render texture based on screen-space size of sprite BB. Prior this PR render texture size was hardcoded.

~~Also fixed bug when sprite offset was scaled by its Scale
**Breaking:** offset will now move sprite in world-space cords, not rotated by 90 degrees like it was before. I believe this is more consistent way to do it.~~
Not anymore, see #1551


PJB3005 (commented):
Changing `SpriteComponent.Offset` to ignore world rotation is definitely incorrect. While the 90 degree bug should be fixed, it should still consider rotation. It's e.g. used for melee lunge animations and those should still rotate correctly when the world rotates.

Otherwise, looks good.

Macoron:
> While the 90 degree bug should be fixed, it should still consider rotation

Okay, I can ignore rotation bug and leave for another PR. Need to rewrite sprite BB function a bit. 
I also found out that after my PR scale also ignores world rotation. Looks like changing multiplication order broke something.

Macoron:
After #1551 it all just start working. I only added support for layers offset with new BB logic.
Remove all code related to -90 rotations hacks. Ready for review.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1607,38 +1607,53 @@ public string GetDebugString()
...
+                    lowestBottomLeft.X = bottomLeft.X;
+                if (lowestBottomLeft.Y > bottomLeft.Y)
+                    lowestBottomLeft.Y = bottomLeft.Y;
+
+                // and hightest top right
+                var topRight = layerBB.TopRight;
+                if (highestTopRight.X < topRight.X)
+                    highestTopRight.X = topRight.X;
+                if (highestTopRight.Y < topRight.Y)
+                    highestTopRight.Y = topRight.Y;
```
PJB3005: You can use `Box2.Union` for this instead, much shorter.

PJB3005 (changes requested):
I'll just fix this real quick myself but I'll still leave the review comment so you see it.

PJB3005 (approved):

PR #1538, opened 2021-02-06
Add KeyValuePair YAML serialization

DrSmugleaf:
For when you really want a dictionary with just one entry.

### DIFF
--- a/Robust.Shared/Serialization/YamlObjectSerializer.cs
+++ b/Robust.Shared/Serialization/YamlObjectSerializer.cs
@@ -5,7 +5,6 @@ using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Reflection;
-using Robust.Shared.ContentPack;
 using Robust.Shared.Interfaces.Reflection;
 using Robust.Shared.Interfaces.Serialization;
 using Robust.Shared.IoC;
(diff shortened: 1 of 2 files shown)


Robust.Shared/Serialization/YamlObjectSerializer.cs
```diff
@@ -546,6 +545,33 @@ public object NodeToType(Type type, YamlNode node)
...
+            // KeyValuePair<K, V>
+            if (TryGenericKeyValuePairType(type, out var kvpKeyType, out var kvpValType))
+            {
+                var pairType = typeof(KeyValuePair<,>).MakeGenericType(kvpKeyType, kvpValType);
+                var pairNode = (YamlMappingNode) node;
+
+                switch (pairNode.Children.Count)
+                {
+                    case 0:
+                        return Activator.CreateInstance(pairType, new object?[] {default, default})!;
```
PJB3005: Does this actually work for e.g. `KeyValuePair<int, int>`? Surely not because `default` here is `null` and that doesn't fit into an `int`?

DrSmugleaf: I have added two tests for a similar case now, I will check it again in a few hours to make sure.

DrSmugleaf: Test done in https://github.com/space-wizards/RobustToolbox/pull/1538/commits/45b17dbb8e60f2edf28123e672dc22e5a6ca2a8a

It works, beats me as well 
image

DrSmugleaf: Changed to just do `Activator.CreateInstance(pairType)!` instead.

PR #1541, opened 2021-02-09
DynamicTypeFactory PreInject Dependencies

Acruid:
DynamicTypeFactory now injects the dependencies into the fields before the constructor is called.

There might be a better way of doing this with the IL codegen in the IoCManager to cache the constructor, but someone that knows how that works would have to change it. Maybe a new `IoCManager.CreateInstance()`?


Acruid:
I don't enjoy having to fight the nullable system like this, I don't think this PR is going to be an improvement.

PR #1545, opened 2021-02-10
Add SerializedType attribute to specify the id used in !type tags

DrSmugleaf:
Because Acruid didn't like the system that he made to serialize types, also in an attempt to fix container namespaces being serialized so https://github.com/space-wizards/space-station-14/pull/3120 can be merged.
The previous usage of !type still works with this PR.

The name of the type is read from its SerializedType attribute instead of the class name itself.
There are four tests added with this PR.
This PR also makes types serializable without being a mapping type, so you don't need to do `!type:TestType {}` for empty mappings.

Start of the forbidden design docs: https://discord.com/channels/310555209753690112/770682801607278632/809030201547685898

image
image

@Acruid


gradientvera:
Good change but, it breaks so much stuff in content... :suncry:

DrSmugleaf:
> Good change but, it breaks so much stuff in content... :suncry:

> The previous usage of !type still works with this PR.

DrSmugleaf:
Ready for review.

Acruid (approved):

PR #1549, opened 2021-02-12
Add (de)serialization for immutable lists

DrSmugleaf:
Also tests for it.
The reflection calls are a bit arse but I don't know of another way to cast it.


Robust.Shared/Serialization/YamlObjectSerializer.cs
```diff
@@ -510,6 +511,22 @@ public object NodeToType(Type type, YamlNode node)
...
+            {
+                var listNode = (YamlSequenceNode) node;
+                var list = typeof(ImmutableList<>).MakeGenericType(immutableListType);
+                var immutableList = list.GetField("Empty")!.GetValue(null);
+                var add = list.GetMethod("Add")!;
+
+                foreach (var entryNode in listNode.Children)
+                {
+                    var value = NodeToType(immutableListType, entryNode);
+                    immutableList = add.Invoke(immutableList, new[] {value});
```
PJB3005: Isn't this literally allocating a new list for every single element, resulting in O(n!) behavior?

You should probably copy the elements into an intermediate array or list and then create the immutablelist to finish it.

DrSmugleaf: Good point it is, I'll fix it tomorrow then.

DrSmugleaf: Fixed.

DrSmugleaf:
Reviews addressed.

PR #1560, opened 2021-02-15
Physics refactor

metalgearsloth:
Okay woo:

Still TODO before I undraft this shit:
- [x] Merge IPhysicsComponent and IPhysBody
- [x] Make PhysicsComponent not partial anymore
- [ ] Some random 11th hour bug I find

So full disclaimer: stuff will need tweaking, not everything's done, etc etc.
Stuff that's in now:
* Circle + edge shapes (circle human colliders)
* Some level of polygon support (centre of mass not in yet)
* Actual physics so you don't need to use a controller to move stuff every tick
* Optimisations to subfloor and items (every item can have a mask now whaaaaaatttt)
* grid-based broadphase
* Initial joint support
* Throwing is much easier to do
* Box2D's block solver (pushing a line of objects into a wall should be more stable)
* VirtualControllers replaced with system-y controllers that just use components
* LinearDamping so objects will slowly come to a rest
* Weightlessness rubberbanding should be fixed (I kept it as is for now)

Stuff that's not in but sometime this year:
* TOI solver (CCD)
* Raycasts for the different shape types (currently it still just does the broadphase AABB)
* Good shuttles (PVS refactor and renderingtree grid-based still need doing)
* Rest of the joints
* Maybe multi-threading some internals (based on what Velcro and Aether2D do)
* Start and end contact events

The idle performance will inherently be slower because it's doing a lot more but under heavy load it should be better.

Notes:
* Velcro and Aether are like 80-90% Farseer code. Aether mainly just has multithreading slapped in
* I rewrote mover code to be a bit cleaner IMO, it's under SharedMoverController. e.g. I moved footsteps out of IMoverController and made an IMobMoverController
* Some of the controller code is still a bit disgusting
* Broadphase system is mostly custom
* Physics body is somewhat custom to fit our networking

At a high level how it works:
1. SharedPhysicsSystem runs controllers globally
2. Runs each map in turn via the Step method which generally just does Find Contacts (via ContactManager) -> Solve
3. Solve will go through all awake bodies and build islands (essentially graphs of connecting bodies).
4. For each island it will go through and more or less do our current physics pipeline (i.e. calculate velocity, solve velocity, solve position, updates). There were some slight changes to this to defer moveevents until the end of the physics step as well as grid changes.

Just expand that out to 15k lines of code.

Some internal stuff like Evalute(ref Manifold) blah blah are all farseer code so don't even ask me about it and just glaze your eyes over. If it looks maths heavy don't even ask me.


metalgearsloth:
Stale

PR #1562, opened 2021-02-15
Fix file not found exceptions when starting up the game with a debugger

DrSmugleaf:
Having to mute breakpoints/convince Rider to ignore these is harder than making this PR.


PJB3005 (commented):
You can get rid of the try catch around that code then.

DrSmugleaf:
> You can get rid of the try catch around that code then.

Done, I wasn't sure if it could still *technically* fail.

PR #1564, opened 2021-02-16
Ensuring that all deleted BasicActorComponents are actually deleted in PlayerSession

io-june:
needed for https://github.com/space-wizards/space-station-14/pull/3233

### DIFF
--- a/Robust.Server/Player/PlayerSession.cs
+++ b/Robust.Server/Player/PlayerSession.cs
@@ -88,7 +88,7 @@ namespace Robust.Server.Player
             {
                 return;
             }
-
+            a.EntityManager.ComponentManager.CullRemovedComponents();
             var actorComponent = a.AddComponent<BasicActorComponent>();
             actorComponent.playerSession = this;
             AttachedEntity = a;


Acruid:
What bug does this PR attempt to fix? Why do you need to call this function? Anything that this function call removes has already been marked as deleted. You should not need to do this.

Acruid:
https://discord.com/channels/310555209753690112/310555209753690112/811914223252144148
playercomp_bug

Someone needs to figure out why deleted components are needing to be cleaned up before assigning a new BasicActorComponent. I'm pretty sure this PR is a symptom of a deeper bug in EntityManager.

io-june:
> https://discord.com/channels/310555209753690112/310555209753690112/811914223252144148
> playercomp_bug
> 
> Someone needs to figure out why deleted components are needing to be cleaned up before assigning a new BasicActorComponent. I'm pretty sure this PR is a symptom of a deeper bug in EntityManager.

I suspect that because when we add component via AddComponent, it checks for duplicates and calls this to immediately remove duplicates:
https://github.com/space-wizards/RobustToolbox/blob/master/Robust.Shared/GameObjects/ComponentManager.cs#L108

and this function especially ignores components, marked for deletion:
https://github.com/space-wizards/RobustToolbox/blob/e0cdcd228e4773b9c737cc9c09737c08d13d248d/Robust.Shared/GameObjects/ComponentManager.cs#L265

Acruid:
> 
> 
> > https://discord.com/channels/310555209753690112/310555209753690112/811914223252144148
> > playercomp_bug
> > Someone needs to figure out why deleted components are needing to be cleaned up before assigning a new BasicActorComponent. I'm pretty sure this PR is a symptom of a deeper bug in EntityManager.
> 
> I suspect that because when we add component via AddComponent, it checks for duplicates and calls this to immediately remove duplicates:
> https://github.com/space-wizards/RobustToolbox/blob/master/Robust.Shared/GameObjects/ComponentManager.cs#L108
> 
> and this function especially ignores components, marked for deletion:
> 
> https://github.com/space-wizards/RobustToolbox/blob/e0cdcd228e4773b9c737cc9c09737c08d13d248d/Robust.Shared/GameObjects/ComponentManager.cs#L265

Yep, that is the problem. Pretty sure `RemoveComponentImmediate` needs to call `DeleteComponent` regardless if it runs the shutdown code or not. Clearly we need MORE UNIT TESTING. Thanks for looking into this.

io-june:
Anyway, i'm closing this and creating an issue for discovered problem. Better I write a function to reassign and revisit Mind instances inside entities without reassigning PlayerSession back and forth

PR #1573, opened 2021-02-19
Add window focused callback to Clyde

DrSmugleaf:
We have the technology.


Robust.Client/Graphics/Clyde/Clyde.Windowing.cs
```diff
@@ -118,6 +119,7 @@ public int GetKeyScanCode(Keyboard.Key key)
 
         private List<Exception>? _glfwExceptionList;
         private bool _isMinimized;
+        private bool _isFocused;
```
PJB3005: Maybe expose this with a property?

DrSmugleaf: Done.

PR #1584, opened 2021-02-20
Fluent Localization

RemieRichards:
Localization Improvement: Replace NGettext with Fluent.NET implementation of [Project Fluent](https://www.projectfluent.org/).

> Asymmetric Localization
Natural-sounding translations with genders and grammatical cases only when necessary. Expressiveness is not limited by the grammar of the source language.

> Progressive Enhancement
Translations are isolated; locale-specific logic doesn't leak to other locales. Authors can iteratively improve translations without impact on other languages.

Have a play on the [playground](https://www.projectfluent.org/play/).


PJB3005 (approved):

PaulRitter (commented):
prob wouldn't be a bad idea to add some tests to this, seeing how the old one had some too

RemieRichards:
@PaulRitter those were purely for the evil macro system, which I have slain in this PR.

I'll do a follow up PR at some point with some tests and an analyser for messageId<->.ftl definition comparisons to ensure we've always got a definition for each messageId

PR #1590, opened 2021-02-21
Add a scroll speed property to ScrollContainer

lzimann:
Small change to allow the customization of the scroll speed for ScrollContainers


Robust.Client/UserInterface/Controls/ScrollContainer.cs
```diff
@@ -17,6 +17,9 @@ public class ScrollContainer : Container
 
         private bool _suppressScrollValueChanged;
 
+        public int ScrollSpeedX { get; set; } = 50;
```
lzimann, line 20: I am not sure if this is the best name of the property, but couldn't think of anything else.

PJB3005 (approved):

PR #1597, opened 2021-02-25
Merged CameraRecoilComponent into EyeComponent

Acruid:
Merged CameraRecoilComponent from SS14 into EyeComponent. Now every Eye gets viewkick goodness!
Made the client/server EyeComponent internal, client did not need access anyways.

### DIFF
--- a/Robust.Client/GameObjects/ClientComponentFactory.cs
+++ b/Robust.Client/GameObjects/ClientComponentFactory.cs
@@ -35,9 +35,6 @@ namespace Robust.Client.GameObjects
             Register<ClientOccluderComponent>();
             RegisterReference<ClientOccluderComponent, OccluderComponent>();
 
-            Register<EyeComponent>();
-            RegisterReference<EyeComponent, SharedEyeComponent>();
-
             Register<AppearanceComponent>();
             RegisterReference<AppearanceComponent, SharedAppearanceComponent>();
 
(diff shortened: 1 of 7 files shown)


metalgearsloth (approved):

PJB3005 (changes requested):
What's the point of merging these? This really isn't the kind of feature the engine should support itself.

Acruid:
> 
> 
> What's the point of merging these? This really isn't the kind of feature the engine should support itself.

Being able to bump/kick/shake the camera is not a SS14-content specific gameplay feature, any RobustToolbox game can benefit from it. Any general game feature like this one should be supported by the engine.

The actual reason why I made this PR was because I was looking at why projectile code was serverside, then realized every human mob has a Recoil and Eye in SS14, and Recoil had a `[RequiresComponent]` on Eye (it can't exist as a lone component without Eye), and this is a game agnostic feature, so it was a prime candidate for being pulled up into the EyeComponent. One less component on every mob, one less ECS System with a Query, an entire set of shared/server/client components removed from the codebase, no functionality loss.

PJB3005:
The problem with moving this into the engine is that it is a feature that is often very specifically tweaked by specific games, and hardcoding it into the engine like this massively reduces flexibility. What you end up getting is that content ignores the feature because it's too limited and now we just have cruft in the engine.

Acruid:
Bad idea.

PR #1604, opened 2021-03-01
SoundSystem

Acruid:
Added a new static proxy class named `SoundSystem` for playing sounds from anywhere. This is the debut of the new `Filter` class for selecting which players to send the sounds to.

Requires https://github.com/space-wizards/space-station-14/pull/3467.

### DIFF
--- a/Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
+++ b/Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Robust.Server.Player;
 using Robust.Shared.Audio;
 using Robust.Shared.GameObjects;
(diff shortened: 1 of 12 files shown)


Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -107,7 +115,7 @@ public AudioSourceServer PlayGlobal(string filename, AudioParams? audioParams =
         /// <param name="audioParams"></param>
         /// <param name="range">The max range at which the audio will be heard. Less than or equal to 0 to send to every player.</param>
         /// <param name="excludedSession">Sessions that won't receive the audio message.</param>
-        public AudioSourceServer PlayFromEntity(string filename, IEntity entity, AudioParams? audioParams = null, int range = AudioDistanceRange, IPlayerSession? excludedSession = null)
+        public IPlayingAudioStream PlayFromEntity(string filename, IEntity entity, AudioParams? audioParams = null, int range = AudioDistanceRange, IPlayerSession? excludedSession = null)
```
metalgearsloth: Should these be obsoleted? Ideally it'd just be Play() with overloads like the client one right

Acruid: Good idea.

PR #1624, opened 2021-03-13
Create integration prototype manager to reduce integration test run time

DrSmugleaf:
PR draft until I make the code way less suspect as it is extremely jank right now.

Reduced by 60% when I tried in https://github.com/DrSmugleaf/space-station-14/pull/9
Down to 4 minutes from 9 minutes 30 seconds (30 seconds less with no preloading which was merged).
It loads default prototypes once.

Local profiles TBD


DrSmugleaf:
Closed in favor of https://github.com/space-wizards/RobustToolbox/pull/1634

PR #1630, opened 2021-03-14
Fix static sleeping crash

metalgearsloth:
Removing the assert should be gucci given a body swapping types may or may not already be awake.

Also this is a change from Farseer where static bodies can be awake even though they can't propagate islands which is stupid so I just don't allow them to ever be awake.


mirrorcult:
fixes https://github.com/space-wizards/space-station-14/issues/3646? or is that different

metalgearsloth:
Das da fix

PR #1631, opened 2021-03-14
Fixed view distance

### LINKED ISSUE #1001: You can see different distances at different resolutions
At higher resolutions you can see further, which means that you'll get slightly different feel / view at different resolutions. At 2K this is annoying compared to 1920x1080 since everything is smaller. Haven't tested at 4K but I'd imagine it's pretty bad at that res. I think a good fix for this would be to have the camera dynamically zoom based on window resolution. Could be an option if not everyone wants this.

metalgearsloth:
This code is kinda stinky but I wasn't sure how to do it better so please advise. I'm also not sure what the multi-viewport code will look like so this will need changes to support that.

Settable server-side.

Fixes https://github.com/space-wizards/RobustToolbox/issues/1001

Pic of 1 half of my 4k monitor:
image

- [ ] Fix do_after
- [ ] Black bars for anything past the view distance

### DIFF
--- a/Robust.Client/GameObjects/Components/Eye/EyeComponent.cs
+++ b/Robust.Client/GameObjects/Components/Eye/EyeComponent.cs
@@ -19,8 +19,6 @@ namespace Robust.Client.GameObjects
         [ViewVariables]
         private Eye? _eye = default!;
 
-        // Horrible hack to get around ordering issues.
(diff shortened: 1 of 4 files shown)


metalgearsloth:
I underestimated the level of effort involved in making sure only part of the screen is drawn to so closing til I get the motivation to work on it again / someone else works on it.

PR #1639, opened 2021-03-15
Fix raising local events not taking into account the argument type

DrSmugleaf:
It took in the generic type, which when using a ternary can be EntityEventArgs or some other base type. Now it uses the actual argument type to lookup the table.


Acruid:
Is there really a case where you don't know the event type when raising it? And if this is changed, the other Raise*Event functions need to be changed as well.

DrSmugleaf:
> Is there really a case where you don't know the event type when raising it? And if this is changed, the other Raise*Event functions need to be changed as well.

```csharp
EntityEventArgs message = value == null
                    ? new PullableStopMovingMessage()
                    : new PullableMoveMessage();

                Owner.EntityManager.EventBus.RaiseLocalEvent(Owner.Uid, message);
```

That's the only use-case I could find.

Acruid:
Both of the event types are known at compile time in that example. Call RaiseEvent with proper generic type args so we don't need runtime reflection to deal with *creative* code.

PR #1649, opened 2021-03-18
SoundSystem Improvements

Acruid:
SoundSystem has now been battle tested.

Requires https://github.com/space-wizards/space-station-14/pull/3697.

### DIFF
--- a/Robust.Shared/Player/Filter.cs
+++ b/Robust.Shared/Player/Filter.cs
@@ -1,8 +1,10 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using JetBrains.Annotations;
+using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
-using Robust.Shared.Maths;
+using Robust.Shared.Map;
 using Robust.Shared.Players;
 using Robust.Shared.Utility;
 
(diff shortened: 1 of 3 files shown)


Robust.Shared/Player/Filter.cs
```diff
@@ -59,14 +61,51 @@ public Filter AddPlayer(ICommonSession player)
...
+        /// Adds all players inside an entity's PVS.
+        /// </summary>
+        protected Filter AddPlayersByPvs(MapCoordinates origin)
+        {
+            //TODO: Calculate this from the PVS system that does not exist.
+            var playerMan = IoCManager.Resolve<ISharedPlayerManager>();
+
+            const int range = 25;
+            var players = playerMan.NetworkedSessions.Where(x =>
+                x.AttachedEntity != null && origin.InRange(x.AttachedEntity.Transform.MapPosition, range));
```
metalgearsloth, line 96: Isn't the default PVS range 12.5? Probably doesn't matter given the refactor coming anyway as it's just a bit of extra bandwidth.

metalgearsloth (approved):

gradientvera (approved):

Robust.Shared/Player/Filter.cs
gradientvera: Yes, but you still want people to hear sounds that are not in their view imo, so it makes sense that the sound range is double the view PVS range

PR #1651, opened 2021-03-22
Fix CopyWithTypeSerializer not copying when null

ShadowCommander:
Changed `CopyWithTypeSerializer` to match with `CopyToTarget`
https://github.com/space-wizards/RobustToolbox/blob/cedfa0ee2f6d41efd07657c747de5de7ef61a2c1/Robust.Shared/Serialization/Manager/SerializationManager.cs#L473-L476

https://discord.com/channels/310555209753690112/770682801607278632/823457120544227378


PaulRitter:
yeeting this into the codebase with no survivors

PR #1652, opened 2021-03-24
Re-implement inertia

metalgearsloth:
Requires https://github.com/space-wizards/space-station-14/pull/3709 for namespace changes.

So box2d has MassData as its own struct on shapes.
Few problems with this in Robust:
1. Making it play nice with serv3 is a pain.
2. You can't as easily share shapes (which is something I want to do eventually to reduce memory usage like shaders can be shared).
3. You may as well just put the data onto the Fixture directly.

This allows us to have rotatable bodies again. Mass was made read-only on bodies and now goes on the fixture instead. For whatever reason FPE / Aether2D lets you set mass on a body directly but then never sets the inertia so you would normally need to manually re-calculate it.


Robust.Shared/Physics/Collision/Shapes/PhysShapeAabb.cs
```diff
@@ -70,6 +71,16 @@ public PhysShapeAabb()
...
 
+        public float CalculateArea()
+        {
+            return _localBounds.Width * _localBounds.Height;
+        }
+
+        public void ComputeProperties()
+        {
+            var area = CalculateArea();
+        }
```
gradientvera: This looks unfinished.

Robust.Shared/Physics/Collision/Shapes/PhysShapeGrid.cs
```diff
@@ -49,6 +50,11 @@ void ISerializationHooks.AfterDeserialization()
             _mapGrid = (IMapGridInternal)mapMan.GetGrid(_gridId);
         }
 
+        public void ComputeProperties()
+        {
+            throw new NotImplementedException();
+        }
```
gradientvera: Also looks unfinished, and throws if called. Is that intended?

gradientvera (approved):
I have no clue what I just reviewed.

PR #1659, opened 2021-03-27
Fix net.rate cvar warning

metalgearsloth:
This also fixes a bunch of tests for CI that were logging warnings.


ShadowCommander:
It looks like it's unused now so I'm just going to merge this.

PR #1662, opened 2021-03-27
Adds prototype "Variant" helper methods to IPrototypeManager

gradientvera:
Needed for a future VV improvements PR.
Example: `"entity"` -> `EntityPrototype`

Also adds `class` constrains to some existing `IPrototype` methods, and adds a new `TryIndex` overload.

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.cs
@@ -49,6 +49,14 @@ namespace Robust.Shared.Prototypes
         IEnumerable<IPrototype> EnumeratePrototypes(Type type);
 
         /// <summary>
+        /// Return an IEnumerable to iterate all prototypes of a certain variant.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


Robust.Shared/Prototypes/PrototypeManager.cs
```diff
@@ -64,8 +64,61 @@ public interface IPrototypeManager
         /// </exception>
         IPrototype Index(Type type, string id);
 
-        bool HasIndex<T>(string id) where T : IPrototype;
-        bool TryIndex<T>(string id, [NotNullWhen(true)] out T? prototype) where T : IPrototype;
+        /// <summary>
+        ///     Returns whether a prototype of type <see cref="T"/> with the specified <see cref="id"/> exists.
```
PJB3005: You can't use `<see>` with (type) parameters. You need `<paramref>` and `<typeparamref>`.

PJB3005 (changes requested):

PR #1665, opened 2021-03-30
Split entity management from entity queries

metalgearsloth:
Content requires https://github.com/space-wizards/space-station-14/pull/3747

Start of slowly porting my pvs refactor changes over.

Removed all of the spatial query stuff from EntityManager. I also removed a lot of UpdateEntityTree calls given there were many duplicates and these are hella expensive; nothing seems to have blown up and the game seems to still work.

I haven't properly profiled but my client went from 30 fps to 35fps while walking around.

UpdateEntityTree is deferred but IMO it's worth it given the performance gains we can get out of it e.g. most of the time you don't even need to do the UpdateEntityTree for RotateEvent given when an entity's rotated they typically also issue a MoveEvent update.


Acruid (approved):
Make sure unit tests are passing before merging this. Right now 225 are failing.

metalgearsloth:
Yeah will do.

metalgearsloth:
Unit tests don't seem to like it being an EntitySystem so might make it an IoCManager instead.

PJB3005:
> UpdateEntityTree is deferred but IMO it's worth it given the performance gains we can get out of it

This is how it should have been from the start, so.

metalgearsloth:
Tests run locally + acruid seems gucci with it after the new-and-approved DI was added.

PR #1666, opened 2021-03-30
Grid-trees for rendering

metalgearsloth:
I added MapId to GridChangedEventHandler (You can't use GetGrid to get the parentmapid when destroyed is invoked so it seemed easier) so this will also need a content update.


Acruid (approved):

metalgearsloth:
I've run tests locally and the only 2 failing are the 2 compdependencies tests that are already failing on master so I'ma merge it.

PR #1668, opened 2021-03-30
Fix crash on checking if grid 0 is paused

DrSmugleaf:
AISteeringSystem in content uses this to know if steering should continue.
If the steerer is on grid 0 then an exception occurs since its not found.
It makes the most sense to me for grid 0 to never be paused, otherwise the check would have to be moved to AISteeringSystem's Update method.

### DIFF
--- a/Robust.Shared/Timing/PauseManager.cs
+++ b/Robust.Shared/Timing/PauseManager.cs
@@ -92,6 +92,11 @@ namespace Robust.Shared.Timing
 
         public bool IsGridPaused(GridId gridId)
         {
+            if (!gridId.IsValid())
+            {
+                return false;
+            }
+
             if (_mapManager.TryGetGrid(gridId, out var grid))
             {
                 return IsGridPaused(grid);


Acruid:
`If the steerer is on grid 0`
Acruid: no, it isn't
Acruid: because GridId.Invalid isn't a real grid
Acruid: whatever content is doing with this, it's dumb
`It makes the most sense to me for grid 0 to never be paused, otherwise the check would have to be moved to AISteeringSystem's Update method.`
Acruid: no, we are not adding a special case hack to the engine so that content code can do whatever dumb stuff it's doing
Acruid: the question you should be asking is "where is AISteeringSystem getting a GridId.Invalid from, and why is it checking if the invalid grid is paused`

This function should already be returning true if passed an invalid grid, and logging a warning that content is doing something dumb. IF it is throwing an exception, that is a completely different bug.

Acruid (changes requested):
So anyways, unless there is an actual design reason why you would ever want Invalid Grids (which should always be unknown since they can't be created) to be Unpaused (and keep unknown grids paused), this isn't getting merged.

PR #1669, opened 2021-03-30
Fix the client trying to pivot the entity of already pivoted maps

DrSmugleaf:
Otherwise it acknowledges a tick which state creates a map, then gets the created map sent again on the next tick and causes an exception.

### DIFF
--- a/Robust.Shared/Map/MapManager.Network.cs
+++ b/Robust.Shared/Map/MapManager.Network.cs
@@ -209,8 +209,13 @@ namespace Robust.Shared.Map
                     var mapComps = _entityManager.ComponentManager.EntityQuery<IMapComponent>(true);
                     foreach (var mapComp in mapComps)
                     {
-                        if (!mapComp.Owner.Uid.IsClientSide() && mapComp.WorldMap == mapId)
+                        if (mapComp.WorldMap == mapId)
                         {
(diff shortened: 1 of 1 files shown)


Robust.Shared/Map/MapManager.Network.cs
```diff
@@ -209,8 +209,13 @@ public void ApplyGameStatePost(GameStateMapData? data)
                     var mapComps = _entityManager.ComponentManager.EntityQuery<IMapComponent>(true);
                     foreach (var mapComp in mapComps)
                     {
-                        if (!mapComp.Owner.Uid.IsClientSide() && mapComp.WorldMap == mapId)
+                        if (mapComp.WorldMap == mapId)
```
Acruid, line 212: Your new code is basically `for the mapId mapComp; if mapComp entity is clientside; return`

`mapComp.Owner.Uid` should always equal `_mapEntities[mapId]`, (line 220) so why is line 200 not catching this?

Also shouldn't it be breaking, not returning? The entire inner loop is a linear search for a mapId->mapComp where mapcomp entities are clientside. Yes, this entire section of code could (and should) prob be a 4 line linq statement. If this is the first time *another* map id is being pivoted, this will return before that happens.

I'm just trying to figure out of the entire pivot system is fubar, because this fix alone seems sketchy.

PR #1670, opened 2021-03-31
Make content able to choose which log level leads to test failures

DrSmugleaf:


metalgearsloth (approved):

Acruid (approved):
As long as this isn't set on every test "to get it to pass" it should be fine. There are a few tests where it is OK for warnings to show up, but I would be really reluctant for tests to allow error/critical log levels to pass.

DrSmugleaf:
> As long as this isn't set on every test "to get it to pass" it should be fine. There are a few tests where it is OK for warnings to show up, but I would be really reluctant for tests to allow error/critical log levels to pass.

On ShadowCommander's PR I made 3 tests allow warnings, but not errors or critical. This is for those.

PR #1674, opened 2021-04-01
broadphase cleanup

metalgearsloth:
Was I high when I wrote this idk

### DIFF
--- a/Robust.Shared/Physics/BroadPhase/SharedBroadPhaseSystem.cs
+++ b/Robust.Shared/Physics/BroadPhase/SharedBroadPhaseSystem.cs
@@ -38,8 +38,8 @@ namespace Robust.Shared.Physics.Broadphase
         /// </summary>
         private HashSet<IEntity> _handledThisTick = new();
 
-        private Queue<MoveEvent> _queuedMoveEvents = new();
-        private Queue<RotateEvent> _queuedRotateEvent = new();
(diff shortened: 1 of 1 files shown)


PJB3005 (changes requested):
Why change those queues to stacks? Do they actually have to be processed in reverse order like that?

metalgearsloth:
Technically no but there is a hashset checking if an update is already processed (this is mainly useful when changing directions so you don't process both RotateEvent and MoveEvent) so if multiple MoveEvents are queued we only care about the latest.

PJB3005:
Yeah but you're not actively removing those duplicated events from the stack in any way, so you still have to go through the entire thing either way, no?

metalgearsloth:
That is true but it's more about making sure the latest position is used. (I also just realised I'm getting WorldPosition again in SynchronizeFixtures which is really stupid so I also need to fix that).

PJB3005:
As far as I can tell you're using both the new and old position in the move event to calculate displacement (this is for CCD, right?). But wouldn't using the newest event mean you only catch part of the displacement?

it seems like going from oldest, and then just pulling the position off the transform instead of the event, would be better. No?

metalgearsloth:
You're right; I'll come back to this when I port the TOI solver which will be a while so I'll close this for now.

PR #1675, opened 2021-04-02
Remove NetId requirement for local event subscriptions

ShadowCommander:
Moves AddedComponentEventArgs so that EntitySystem.SubscribeLocalEvent works with components without NetIds.


Acruid (approved):
Yep, that was a bug.

PR #1678, opened 2021-04-03
GridId caching

### LINKED ISSUE #960: Cache GridId and MapId inside TransformComponent

metalgearsloth:
Closes https://github.com/space-wizards/RobustToolbox/issues/960

I wasn't sure if this should be networked given the client should be able to calculate these on its own?


metalgearsloth:
Should be okay for review now I think

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -65,25 +66,20 @@ internal class TransformComponent : Component, ITransformComponent, IComponentDe
         [ViewVariables]
         public GridId GridID
         {
-            get
+            get => _gridId;
+            set
```
PJB3005: Make this `internal set` to avoid content doing something dumb?

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -424,6 +420,40 @@ static MapId FindMapIdAndSet(TransformComponent p)
...
+        }
+
+        public bool HasGridIndex()
+        {
+            return !Owner.HasComponent<IMapComponent>() &&
+                   !Owner.HasComponent<IMapGridComponent>() &&
+                   (!Owner.Transform.ParentUid.IsValid() ||
+                   Owner.Transform.Parent!.Owner.HasComponent<IMapComponent>() ||
+                   Owner.Transform.Parent!.Owner.HasComponent<IMapGridComponent>());
+        }
```
PJB3005: Wasn't grid ID lookup recursive before so that parenting things kept the map/grid ID?

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
metalgearsloth: I ended up having to cast it back to TransformComponent but I guess it's not bad considering it's rarely set?

PR #1683, opened 2021-04-06
Don't use linked-list for contact pooling

metalgearsloth:
Always seemed icky to me. Aether uses a linked-list (like world contacts) and Farseer just uses a Queue.

### DIFF
--- a/Robust.Shared/Physics/Dynamics/ContactManager.cs
+++ b/Robust.Shared/Physics/Dynamics/ContactManager.cs
@@ -50,10 +50,11 @@ namespace Robust.Shared.Physics.Dynamics
         /// </summary>
         public BroadPhaseDelegate OnBroadPhaseCollision;
 
-
         public readonly ContactHead ContactList;
         public int ContactCount { get; private set; }
-        internal readonly ContactHead ContactPoolList;
+        private const int ContactPoolInitialSize = 64;
+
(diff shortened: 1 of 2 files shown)


Robust.Shared/Physics/Dynamics/ContactManager.cs
```diff
@@ -71,14 +72,22 @@ public ContactManager()
...
             ContactCount = 0;
-            ContactPoolList = new ContactHead();
             OnBroadPhaseCollision = AddPair;
         }
 
         public void Initialize()
         {
             IoCManager.InjectDependencies(this);
             _broadPhaseSystem = EntitySystem.Get<SharedBroadPhaseSystem>();
+            InitializePool();
```
lordpidey, line 82: Why was the initialization of ContactPoolList moved from the constructor to the initializer?

metalgearsloth: Contact has an IEntitySystemManager dependency.

gradientvera (approved):

gradientvera:
Do you return the contacts to the stack at any point, or?

metalgearsloth:
> 
> 
> Do you return the contacts to the stack at any point, or?

Under Destroy they do

gradientvera:
good for merge then

PR #1687, opened 2021-04-11
Immediate broadphase updates

### LINKED ISSUE #1686: Broadphase needs to be immediate
For the updates after physics we just need to process it in a specific order so we can probably get away with avoiding duplicates from MoveEvent / RotateEvent but generally it should probably be immediate.

metalgearsloth:
Fixes https://github.com/space-wizards/RobustToolbox/issues/1686

Made some PhysicsComponents into IPhysBody so also needs content PR

Requires https://github.com/space-wizards/space-station-14/pull/3879


Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -684,12 +721,14 @@ public override void HandleComponentState(ComponentState? curState, ComponentSta
...
                     SetPosition(newState.LocalPosition);
 
-                    var ev = new MoveEvent(Owner, oldPos, Coordinates);
+                    // UpdatePhysicsTree();
+                    var ev = new MoveEvent(Owner, _oldCoords.Value, Coordinates);
                     EntitySystem.Get<SharedTransformSystem>().DeferMoveEvent(ev);
 
+                    _oldCoords = null;
                     rebuildMatrices = true;
                 }
```
metalgearsloth: If I try and update the broadphase here it lags the shit out of the client so uhhhh not sure what to do about it now.

metalgearsloth:
Broadphase still needs work (updates on the client during FrameUpdate that don't lag to fix the bounceback...) but should be good enough until I come back to it.

Acruid:
Without a Content PR, this needs to pass the Content CI test.

metalgearsloth:
My kingdom for a review.

PJB3005 (approved):

PR #1688, opened 2021-04-11
Adds abstract class for entity events that can be cancelled.

gradientvera:


Robust.Shared/GameObjects/EntityEvents.cs
```diff
@@ -17,11 +17,30 @@ public abstract class EntityEventArgs { }
...
+        /// </summary>
+        public bool Cancelled { get; private set; }
+
+        /// <summary>
+        ///     Cancels the event.
+        /// </summary>
+        public void Cancel() => Cancelled = true;
+
+        /// <summary>
+        ///     Uncancels the event. Don't call this if you don't know what you're doing.
```
gradientvera:
```suggestion
        ///     Uncancels the event. Don't call this unless you know what you're doing.
```

PR #1696, opened 2021-04-14
Refactors MapManager to not do any networking.

gradientvera:
Now, ServerMapManager and ClientMapManager handle all map networking.

### DIFF
--- a/Robust.Client/Map/ClientMapManager.cs
+++ b/Robust.Client/Map/ClientMapManager.cs
@@ -12,7 +12,7 @@ using Robust.Shared.Utility;
 
 namespace Robust.Client.Map
 {
-    internal sealed class ClientMapManager : MapManager, IClientMapManager
+    internal class ClientMapManager : MapManager, IClientMapManager
     {
         [Dependency] private readonly INetManager _netManager = default!;
 
--- a/Robust.Client/Map/ClientMapManager.cs
+++ b/Robust.Client/Map/ClientMapManager.cs
@@ -16,8 +16,6 @@ namespace Robust.Client.Map
     {
         [Dependency] private readonly INetManager _netManager = default!;
 
-        public override event EventHandler<GridChangedEventArgs>? GridChanged;
-
         public void ApplyGameStatePre(GameStateMapData? data)
         {
             // There was no map data this tick, so nothing to do.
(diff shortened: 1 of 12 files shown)


Robust.UnitTesting/Server/RobustServerSimulation.cs
```diff
@@ -191,7 +191,7 @@ public ISimulation InitializeInstance()
             //Tier 3: Networking
             //TODO: Try to remove these
             container.RegisterInstance<IEntityNetworkManager>(new Mock<IEntityNetworkManager>().Object);
-            container.RegisterInstance<INetManager>(new Mock<INetManager>().Object);
+            //container.RegisterInstance<INetManager>(new Mock<INetManager>().Object);
```
Acruid: No commented code in PRs.

Robust.Shared/Map/MapManager.cs
```diff
@@ -38,7 +36,7 @@ internal partial class MapManager : IMapManagerInternal
         ///     so that you don't spam an event for each of the million station tiles.
         /// </summary>
         /// <inheritdoc />
-        public event EventHandler<GridChangedEventArgs>? GridChanged;
+        public virtual event EventHandler<GridChangedEventArgs>? GridChanged;
```
Acruid: Why is this virtual?

Robust.Client/Map/ClientMapManager.cs
```diff
@@ -5,88 +5,21 @@
...
 using Robust.Shared.Maths;
 using Robust.Shared.Network;
-using Robust.Shared.Timing;
 using Robust.Shared.Utility;
 
-namespace Robust.Shared.Map
+namespace Robust.Client.Map
 {
-    internal partial class MapManager
+    internal sealed class ClientMapManager : MapManager, IClientMapManager
```
Acruid: There is no reason to seal this?

Acruid (changes requested):
Also fix whatever that semicolon bug is in MapManager:148 that github is complaining about.

Robust.Shared/Map/MapManager.cs
gradientvera: Because ClientMapManager can't invoke it otherwise

Robust.UnitTesting/Server/RobustServerSimulation.cs
gradientvera: Fixed now

Robust.Client/Map/ClientMapManager.cs
gradientvera: Not really, you're right.

Robust.Shared/Map/MapManager.cs
gradientvera: fixed now

Acruid (approved):

PR #1698, opened 2021-04-15
ConsoleHost no longer depends on INetManager

gradientvera:
- I decided against removing `IsServer` and `IsClient` since they're still used by `IPauseManager`


gradientvera:
I'm not sure about this one anymore, closing for now.

PR #1699, opened 2021-04-15
UserInterfaceManager no longer depends on INetManager (and others)

gradientvera:
- Creates IDebugControlsFactory, which returns debug controls!


gradientvera:
This is probably not needed anymore, closing for now

PR #1711, opened 2021-04-18
Adds GameController options for games using RobustToolbox as a library.

gradientvera:
Required by https://github.com/space-wizards/space-station-14/pull/3874
- Many options, ranging from sandbox disabling to user data directory name.
- Content.Shared.dll is no longer needed for mounting stuff.
- Content.Client may not need to be called Content.Client anymore, in games using robust as a library.


ZoldorfTheWizard (approved):
looks good to me.

Issue #1719, opened 2021-04-23
Networked Grid Chunk Updates


Acruid:
Right now the PVS system implements a bandwidth optimization where it sends the entire set of server entities over all maps, as well as all grid chunks, to the client when they leave the lobby. It then uses this as a "full state" LastTickUpdated baseline to compare server entity dirty timestamps. Most entities (walls) are never updated after the initial state, so the server never has to send info about them for the rest of the round.

There are multiple downsides to this approach. First, there is a multi-second lag spike when the round starts because the server has to serialize and send a full state to every client. This will get worse as more clients join the server. Second, every client is receiving a state update for *every* shared entity in the server, regardless of which map it is on or where it is. Besides the potential for cheating, if the client stays in one section of one map, he will never see any of the entities on another map, so sending them is a complete waste. Ex. an engie spending the entire round building a supermatter engine on the station does not care about the entities in lavaland.

The first part of the idea is to switch to a grid chunk system similar to Minecraft. This will only send grid tile chunk updates to the client if the player is near them. Basically a player is "subscribed" to the chunks around them, and chunk updates will be sent to the client. Each client has it's own per-chunk subscribed GameTick time that can be used for delta states during the subscription lifetime. The client can cache these chunks and display them as needed.

Using a naive implementation, when a chunk enters the client view, the server is required to send a full chunk state, since it would not know the LastTickUpdated value in the client's chunk cache. This can be greatly improved using prediction, by having the client send the LastTickUpdated of the chunks that entered the view along with the input commands that moved the player. The server can use that value to delta compress the updates against instead of sending a full state. This does have the same downside as Minecraft, where if the server is lagging the moving client will not get chunk updates in time before they become visible.

The second part is rewriting the SnapGrid system into a new AnchoredEntity system. An Anchored entity will be the equivalent of a Minecraft TileEntity, which is actually considered a part of the grid chunk, as opposed to unanchored "loose" moving entities. Anchored entities will be updated through the new grid chunk update system, instead of the existing enter/leave PVS system. Because of this, all anchored entities can use the chunk's LastTickUpdated baseline delta states instead of the global joined round time. The server will not have to send all Anchored entities to the client, because they are now a part of chunk updates.

[...]

PR #1720, opened 2021-04-23
SnapGridComponent Removal

Acruid:
* Removes all references to `SnapGridComponent` outside of `TransformComponent` and `SnapGridSystem`. Various functionality has been moved to `TransformComponent` and/or `IMapGrid`.
* Brings remaining SnapGrid ECS code up to 2021 standards.

This is in preparation for SnapGrid rework in https://github.com/space-wizards/RobustToolbox/issues/1719. The SnapGrid implementation code is not optimal, because it is about to be completely reworked.

Requires https://github.com/space-wizards/space-station-14/pull/3884.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -321,6 +321,10 @@ namespace Robust.Shared.GameObjects
             }
         }
 
+        /// <inheritdoc />
+        [ViewVariables]
+        public bool Anchored => Owner.EntityManager.ComponentManager.HasComponent<SnapGridComponent>(Owner.Uid);
+
         [ViewVariables]
         public IEnumerable<ITransformComponent> Children =>
             _children.Select(u => Owner.EntityManager.GetEntity(u).Transform);
--- a/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -323,7 +323,18 @@ namespace Robust.Shared.GameObjects
 
         /// <inheritdoc />
         [ViewVariables]
-        public bool Anchored => Owner.EntityManager.ComponentManager.HasComponent<SnapGridComponent>(Owner.Uid);
+        public bool Anchored
+        {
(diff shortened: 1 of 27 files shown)


metalgearsloth:
Based

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -321,6 +321,21 @@ public Vector2 LocalPosition
...
 
+        /// <inheritdoc />
+        [ViewVariables]
+        public bool Anchored
+        {
+            get => Owner.HasComponent<SnapGridComponent>();
+            set
+            {
+                if(value && !Owner.HasComponent<SnapGridComponent>())
+                    Owner.AddComponent<SnapGridComponent>();
```
metalgearsloth, line 332: Could probably just do an EnsureComponent<T> instead

Robust.Shared/GameObjects/Systems/SnapGridSystem.cs
```diff
@@ -1,27 +1,119 @@
...
+                }
+
+                lastGrid.RemoveFromSnapGridCell(snapComp.LastTileIndices, euid);
+            }
+
+            if (!_mapManager.TryGetGrid(transform.GridID, out var grid))
+            {
+                // Either a race condition, or we're not on any grids.
+                return;
+            }
```
metalgearsloth, line 80: Should LastGrid be set to Invalid in this case? If it does have a GridID that isn't invalid?

metalgearsloth (approved):

Robust.Shared/GameObjects/Systems/SnapGridSystem.cs
Acruid: The PR was basically a refactor, so this Godot legacy code is untouched from the existing SnapGridComponent that I moved to a System. I have no idea what race condition they are talking about, and I don't want to touch it until I have some tests covering it.

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
Acruid: I could, but the next PR is going to rework all of this, so it's only temporary code to keep everything working.

PR #1724, opened 2021-04-25
Fix placement of grid tiles on the edge of a grid

ShadowCommander:
At some point tile snapping when placing new tiles on the edge of the grid broke. Removing all of the conversions fixed it.

It snaps if the cursor is closer than around 2/3rds of a tile.

### DIFF
--- a/Robust.Shared/Map/CoordinatesExtensions.cs
+++ b/Robust.Shared/Map/CoordinatesExtensions.cs
@@ -1,4 +1,4 @@
-﻿using Robust.Shared.GameObjects;
+using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
 using Robust.Shared.Maths;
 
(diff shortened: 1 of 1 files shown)


Robust.Shared/Map/CoordinatesExtensions.cs
```diff
@@ -53,7 +53,7 @@ public static EntityCoordinates AlignWithClosestGridTile(this EntityCoordinates
                 if (closest != null) // stick to existing grid
                 {
                     // round to nearest cardinal dir
-                    var normal = new Angle(coords.Position - intersect.Center).GetCardinalDir().ToVec();
+                    var normal = coords.Position - intersect.Center;
```
metalgearsloth: Should this be converted to mapcoordinates instead? Given EntityCoordinates are relative and all. Same with some of the code above.

metalgearsloth (approved):

PR #1727, opened 2021-04-25
Make serialization work with backing fields automatically

DrSmugleaf:
This removes the need to write [field: DataField(...)] on get-only properties with a backing field, for which the only warning was a line in the serialization doc and an error after starting up the game.

The backing field is found by looking for a field with the name \<PropertyName\>k__BackingField and an attribute of CompilerGenerated, as far as I know these are an implementation detail but the same across C# and a bunch of other compilers. The method that finds the backing field is in TypeHelpers in case its implementation ever needs to be changed. The other way to do it would be to analyze the instructions of the method body and finding the Ldfld op code (second instruction) and going from there, but that is above my paygrade.

### DIFF
--- a/Robust.Shared/Serialization/Manager/SerializationDataDefinition.cs
+++ b/Robust.Shared/Serialization/Manager/SerializationDataDefinition.cs
@@ -142,8 +142,7 @@ namespace Robust.Shared.Serialization.Manager
 
             _duplicates = fieldDefs
                 .Where(f =>
-                    fieldDefs
-                        .Count(df => df.Attribute.Tag == f.Attribute.Tag) > 1)
+                    fieldDefs.Count(df => df.Attribute.Tag == f.Attribute.Tag) > 1)
                 .Select(f => f.Attribute.Tag)
                 .Distinct()
                 .ToArray();
--- a/Robust.Shared/Serialization/Manager/SerializationDataDefinition.cs
+++ b/Robust.Shared/Serialization/Manager/SerializationDataDefinition.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
 using Robust.Shared.Network;
@@ -81,13 +82,18 @@ namespace Robust.Shared.Serialization.Manager
             Type = type;
             var dummyObj = Activator.CreateInstance(type)!;
 
-            var fieldDefs = new List<FieldDefinition>();
+            var fieldDefs = new Dictionary<AbstractFieldInfo, FieldDefinition>();
 
             foreach (var abstractFieldInfo in type.GetAllPropertiesAndFields())
             {
                 var attr = abstractFieldInfo.GetCustomAttribute<DataFieldAttribute>();
 
-                if (attr == null) continue;
+                if (attr == null || abstractFieldInfo.IsBackingField())
+                {
+                    continue;
+                }
+
+                var backingField = abstractFieldInfo;
 
                 if (abstractFieldInfo is SpecificPropertyInfo propertyInfo)
                 {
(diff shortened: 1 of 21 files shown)


Robust.Shared/Serialization/Manager/SerializationDataDefinition.cs
```diff
@@ -87,7 +88,12 @@ public SerializationDataDefinition(Type type)
             {
                 var attr = abstractFieldInfo.GetCustomAttribute<DataFieldAttribute>();
 
-                if (attr == null) continue;
+                if (attr == null || abstractFieldInfo.IsBackingField())
```
PaulRitter, line 90: this disables the [field: ]-variant. why?

Robust.Shared/Utility/TypeHelpers.cs
```diff
@@ -281,10 +282,56 @@ public static IEnumerable<AbstractFieldInfo> GetAllPropertiesAndFields(this Type
 
             return commonType;
         }
+
+        public static SpecificFieldInfo? GetBackingField(this Type type, string propertyName)
+        {
+            foreach (var parent in type.GetClassHierarchy())
+            {
+                var field = parent.GetField($"<{propertyName}>k__BackingField",
+                    BindingFlags.NonPublic | BindingFlags.Instance);
```
PaulRitter, line 291: i am 99% sure there is a property on the propertyinfo object for this

Robust.Shared/Serialization/Manager/SerializationDataDefinition.cs
```diff
@@ -140,7 +160,27 @@ public SerializationDataDefinition(Type type)
...
+                if (!Equals(definition.Attribute, attribute))
+                {
+                    continue;
+                }
+
+                return definition.InheritanceBehaviour;
+            }
+
+            return null;
+        }
```
PaulRitter: please yeet this out of the class and into the type, make it an extension or smth. exposing basefielddefinitions as a readonly list is cleaner than having this in the class imo

PaulRitter (changes requested):
i only have limited braincells due to exams but looks ok aside from the few notes i had.
i also saw you using generics and nullability and got a bit scared but as i said my mental capacity is nearing a 5 year low point so i just skipped that part and trusted you. pls dont disappoint

Robust.Shared/Serialization/Manager/SerializationDataDefinition.cs
DrSmugleaf: That is to skip registering both the property and the backing field, though you are right if field: is used then it should not be skipped.

Robust.Shared/Serialization/Manager/SerializationDataDefinition.cs
DrSmugleaf: It is only for the test but will do.

Robust.Shared/Utility/TypeHelpers.cs
PaulRitter: i was 99% wrong disregard

Robust.Shared/Serialization/Manager/SerializationDataDefinition.cs
DrSmugleaf: On second thought this is already a test case. I changed it so looking up attributes on the property object also returns the ones on its backing field.

Robust.Shared/Serialization/Manager/SerializationDataDefinition.cs
DrSmugleaf: Done in https://github.com/space-wizards/RobustToolbox/pull/1727/commits/5aa025da6611792b0a83a7190f27c670f3a30538

Robust.UnitTesting/Shared/Serialization/PropertyAndFieldDefinitionTest.cs
```diff
@@ -0,0 +1,129 @@
...
+            mapping.AddNode(GetAndSetPropertyName, new ValueDataNode("15"));
+            mapping.AddNode(FieldName, new ValueDataNode("20"));
+            mapping.AddNode(GetOnlyPropertyWithOtherAttributeFieldTargetedName, new ValueDataNode("25"));
+            mapping.AddNode(GetOnlyPropertyFieldTargetedAndOtherAttributeName, new ValueDataNode("30"));
+
+            var definition = Serialization.ReadValue<PropertyAndFieldDefinitionTestDefinition>(mapping);
+
+            Assert.NotNull(definition);
+
+            // Set only property with backing field, property targeted
```
RemieRichards, line 41: This comment, and the one below, refer to a "Set only" property.
Yet the two asserts are against a get only property.

So 1 of 3 things it at play.

1. The comments are wrong (i.e. "Set" vs "Get")
2. The code is wrong (Using wrong field, field defined wrong, etc.)
3. The comments are misleading and should be reworded (Halp brain melt)

DrSmugleaf: 1 was correct, the comments were wrong. I have fixed up some of the other comments as well to make them consistent with the rest.
Fixed in https://github.com/space-wizards/RobustToolbox/commit/61113d24349c6e2a10b4b908e17c3e87d0a9a34c

PR #1746, opened 2021-05-08
Change maxlinvelocity and maxangvelocity to take tickrate

ShadowCommander:
Changes maxlinvelocity and maxangvelocity so that they scale based on tickrate.

### DIFF
--- a/Robust.Shared/CVars.cs
+++ b/Robust.Shared/CVars.cs
@@ -418,6 +418,7 @@ namespace Robust.Shared
         /// <summary>
         /// Maximum linear velocity per second.
         /// Make sure that MaxLinVelocity / <see cref="NetTickrate"/> is around 0.5 or higher so that moving objects don't go through walls.
+        /// MaxLinVelocity is compared to the dot product of linearVelocity * frameTime.
         /// </summary>
         /// <remarks>
         /// Default is 35 m/s. Around half a tile per tick at 60 ticks per second.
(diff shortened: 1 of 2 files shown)


gradientvera (approved):

metalgearsloth:
If something exceeds won't it get mispredicts because the client's physics time is very small (being run during rendering) and not exceeding but the server's is larger and exceeding the cap?

ShadowCommander:
> 
> 
> If something exceeds won't it get mispredicts because the client's physics time is very small (being run during rendering) and not exceeding but the server's is larger and exceeding the cap?

From what I can tell it won't because `frameTime` here is always constant since the provided value is `GameTiming.TickPeriod.TotalSeconds`, which doesn't change unless `TickRate` changes.
https://github.com/space-wizards/RobustToolbox/blob/2b589215aa07fd3b26d427ba35a04a795547deb3/Robust.Shared/Physics/Dynamics/PhysicsIsland.cs#L407-L412
https://github.com/space-wizards/RobustToolbox/blob/2b589215aa07fd3b26d427ba35a04a795547deb3/Robust.Shared/Timing/GameTiming.cs#L147
Client
https://github.com/space-wizards/RobustToolbox/blob/2b589215aa07fd3b26d427ba35a04a795547deb3/Robust.Client/GameStates/ClientGameStateManager.cs#L308
Server
https://github.com/space-wizards/RobustToolbox/blob/2b589215aa07fd3b26d427ba35a04a795547deb3/Robust.Shared/Timing/GameLoop.cs#L182

metalgearsloth (approved):
That makes sense

PR #1755, opened 2021-05-12
Deferred Input Context Switches (fixes spawn entities menu while moving)

20kdc:
Fixes spawn entities menu crashing the debug client if you're moving while you select an entry, or any other case where an input event causes a context change.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/InputSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/InputSystem.cs
@@ -152,6 +152,12 @@ namespace Robust.Client.GameObjects
 
             SetEntityContextActive(_inputManager, _playerManager.LocalPlayer.ControlledEntity);
         }
+
+        /// <inheritdoc />
+        public virtual void FrameUpdate(float frameTime) {
+            base.FrameUpdate(frameTime);
+            _inputManager.Contexts.RunDeferredContextSwitch();
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


PJB3005:
I don't like this because it means that input ordering isn't handled correctly.
If you were to press two keys close enough to each other that they're processed in the same frame, and the first changes the input context, with this change the *second* will still be in the old input context as if the input context hasn't changed yet. Something like a stutter is always a possibility so I'd rather not misinterpret inputs as a result.

Maybe defer it to exit from the key event handler instead?

20kdc:
I might have to restructure the PR then, as it sounds like InputContextContainer might not be the right place for this code. Then again the other relevant code is kinda between InputManager and InputSystem...

20kdc:
Worked out a possibility: Adding `Contexts.RunDeferredContextSwitch();` in the `finally` section of `InputManager.SetBindState`. Thoughts?

PJB3005:
`InputSystem` is actually never the correct place to put this because `InputSystem` is concerned with simulation inputs, which things like UI inputs don't hit.

> Worked out a possibility: Adding `Contexts.RunDeferredContextSwitch();` in the `finally` section of `InputManager.SetBindState`. Thoughts?

Presumably, although that'd still raise an interesting edge case if something changes the input context *without* handling the input event, and the event falls through to the next thing. Probably not a big deal?

PJB3005 (approved):

PR #1756, opened 2021-05-14
MIDI improvements

gradientvera:
Just some minor midi improvements that I coded a while back and forgot to push.


Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -426,27 +424,13 @@ public override unsafe int Read(IntPtr buf, long count, IntPtr sfHandle)
...
-
-                        buffer.CopyTo(span);
-                    }
-                    else
-                    {
-                        var buffer = stream.ReadExact(length);
-
-                        buffer.CopyTo(span);
-                    }
+                    stream.ReadExact(span);
```
PJB3005: This violates Fluidsynth's API which states that, on failure, the buffer should be unmodified. That's why that code did what it did.

PJB3005 (changes requested):

Robust.Client/Audio/Midi/MidiManager.cs
PJB3005: Add a comment instead for clarity, I guess?

gradientvera: done, my bad

PaulRitter (approved):

PR #1758, opened 2021-05-17
Automatically unsubscribe event bus registrations in entity system shutdown.

PJB3005:


PaulRitter (changes requested):
imo the unsubscribe methods should remove subscriptions from the `_subscriptions` list, else you'll have weird behaviour if you unsubscribe manually for whatever reason

gradientvera:
> imo the unsubscribe methods should remove subscriptions from the `_subscriptions` list, else you'll have weird behaviour if you unsubscribe manually for whatever reason

imo, they should remove the subscriptions from the list AND unsubscribe from the event bus.

PJB3005:
I mostly did it like this because the code doesn't explode from a double unregistration, and I don't want to bother synchronizing a submodule update.

We already decided that stuff shouldn't be doing eventbus registations mid-game anyways so I don't think it's a big deal.

PaulRitter:
> > imo the unsubscribe methods should remove subscriptions from the `_subscriptions` list, else you'll have weird behaviour if you unsubscribe manually for whatever reason
> 
> imo, they should remove the subscriptions from the list AND unsubscribe from the event bus.

yes, i meant he should *add* the removal to the unsubscribe methods

PaulRitter (approved):

PR #1759, opened 2021-05-17
RenderingTreeSystem cleanup

metalgearsloth:
10% less bad but 80% still boilerplate.

I'd advise looking at split when reviewing as unified is unreadable.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/RenderingTreeSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/RenderingTreeSystem.cs
@@ -1,4 +1,6 @@
 using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
 using JetBrains.Annotations;
 using Robust.Client.Physics;
 using Robust.Shared.GameObjects;
@@ -6,6 +8,7 @@ using Robust.Shared.IoC;
 using Robust.Shared.Map;
 using Robust.Shared.Maths;
 using Robust.Shared.Physics;
+using Robust.Shared.Utility;
 
 namespace Robust.Client.GameObjects
 {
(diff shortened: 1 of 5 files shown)


Robust.Client/GameObjects/EntitySystems/RenderingTreeSystem.cs
```diff
@@ -215,47 +249,126 @@ private void MapManagerOnGridCreated(MapId mapId, GridId gridId)
...
+                    if (_gridTrees.TryGetValue(sprite.IntersectingMapId, out var oldMapTree))
+                    {
+                        foreach (var gridId in sprite.IntersectingGrids)
+                        {
+                            if (!oldMapTree.TryGetValue(gridId, out var gridTree)) continue;
+                            gridTree.SpriteTree.Remove(sprite);
+                        }
+                    }
+
+                    sprite.IntersectingGrids.Clear();
```
PJB3005: Isn't this code duplicated with `ClearSprite`?

Robust.Client/GameObjects/EntitySystems/RenderingTreeSystem.cs
```diff
@@ -47,151 +50,182 @@ public override void Initialize()
...
-            UnsubscribeLocalEvent<RenderTreeRemoveLightMessage>();
+        private void SpriteParentChanged(EntityUid uid, SpriteComponent component, EntParentChangedMessage args)
+        {
+            QueueSpriteUpdate(component);
         }
 
-        // For these next 2 methods (the Remove* ones):
-        // If the Transform is removed BEFORE the Sprite/Light,
-        // then the MapIdChanged code will handle and remove it (because MapId gets set to nullspace).
-        // Otherwise these will still have their past MapId and that's all we need..
```
PJB3005: As far as I can tell the code still relies on this behavior, so this comment should probably be kept.

PR #1760, opened 2021-05-17
Add linguini loc

Ygg01:
Basically replaces Fluent.Net with my implementation of fluent.rs.

Please be merciless and review it and the https://github.com/Ygg01/Linguini repository.

~~I'm new to C#, and making this thread safe is next to impossible.~~
Wrapped in Concurrent Dictionary should be fine, in theory.

It depends on file changes in:
https://github.com/space-wizards/space-station-14/pull/4018/
(the .FTL fixes)

Image of Blue Rat


Ygg01:
I have an idea to solve this issue. I'll log error instead of breaking.

Robust.Shared/Localization/ILocalizationManager.cs
```diff
@@ -46,13 +46,13 @@ public interface ILocalizationManager
         string GetString(string messageId, params (string, object)[] args);
 
         /// <summary>
-        ///     Try- version of <see cref="GetString(string, ValueTuple{string, object}[])"/>
+        ///     Try- version of <see cref="GetString(string, (string, object)[])"/>
```
PJB3005: This isn't valid syntax and it causes the doc reference to break.

Robust.Shared/Localization/LocFunction.cs
```diff
@@ -62,17 +63,17 @@ public LocArgs(IReadOnlyList<ILocValue> args, IReadOnlyDictionary<string, ILocVa
...
         ///     Format this value to a string.
         /// </summary>
         /// <remarks>
         ///     Used when this value is interpolated directly in localizations.
         /// </remarks>
-        /// <param name="ctx">Context containing data like culture used.</param>
+        /// <param name="bundle">Context containing data like culture used.</param>
         /// <returns>The formatted string.</returns>
-        string Format(LocContext ctx);
+        string Format(LocContext bundle);
```
PJB3005: These parameters shouldn't be renamed since the type is still called a `LocContext` in engine. Holds true for the rest of the file aswell.

.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "ManagedHttpListener"]
 	path = ManagedHttpListener
 	url = https://github.com/space-wizards/ManagedHttpListener.git
+[submodule "Linguini"]
+	path = Linguini
+	url = https://github.com/Ygg01/Linguini.git
```
PJB3005: I would prefer this submodule reference be on a fork of the repo under our org instead of to your repo. 

That or transfer linguini to our org completely but that's your choice since you made it.

Robust.Shared/Localization/LocFunction.cs
```diff
@@ -116,70 +117,123 @@ protected LocValue(T val)
...
+        }
+
+        public override string AsString()
+        {
+            return ((FluentNumber) Value).AsString();
+        }
+
+        public override IFluentType Copy()
+        {
+            return (FluentNumber) Value;
```
PJB3005: 1. These types are immutable, you shouldn't need to instantiate a new one.
2. You're returning a different object, here.

Robust.Shared/Localization/LocFunction.cs
```diff
@@ -62,17 +63,17 @@ public LocArgs(IReadOnlyList<ILocValue> args, IReadOnlyDictionary<string, ILocVa
     ///     A value passed around in the localization system.
     /// </summary>
     /// <seealso cref="LocValue{T}"/>
-    public interface ILocValue
+    public interface ILocValue : IFluentType
```
PJB3005: With this change, content won't be able to implement custom `ILocValue` types anymore because it doesn't have access to Linguini. That's not desirable.

You can either bring the `FluentLocWrapperType` back, and make `ILocValue` NOT implement `IFluentType`, OR you can cut out the middleman and expose Linguini to Robust directly. I'd prefer the former if only just to avoid having a breaking change with content and all that, since the alternative doesn't offer any meaningful advantages I can think of.

This problem is also evident in the confusion surrounding `FluentNumber` and `LocValueNumber` in the code below. `LocValue<T>` and derivatives exist to facilitate the separation layer on top of Fluent. Merging the type systems of fluent and Robust makes spaghetti out of it.

Robust.Shared/Localization/LocHelper.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Text;
+using Linguini.Bundle.Errors;
+using Linguini.Syntax.Parser.Error;
+
+namespace Robust.Shared.Localization
+{
+    public static class LocHelper
```
PJB3005:
```suggestion
    internal static class LocHelper
```

I'd say this is a pretty internal helper method that content shouldn't concern itself with.

Robust.Shared/Localization/LocalizationManager.Entity.cs
```diff
@@ -41,7 +42,7 @@ private bool TryGetEntityLocAttrib(IEntity entity, string attribute, [NotNullWhe
         // Flush caches conservatively on prototype/localization changes.
         private void OnPrototypesReloaded(PrototypesReloadedEventArgs args)
         {
-            if (!args.ByType.TryGetValue(typeof(EntityPrototype), out var changeSet))
+            if (!args.ByType.TryGetValue(typeof(EntityPrototype), out _))
```
PJB3005:
```suggestion
            if (!args.ByType.ContainsKey(typeof(EntityPrototype)))
```

I know you're cleaning up after my mess here but I can't help but notice.

Robust.Shared/Localization/LocalizationManager.Entity.cs
```diff
@@ -52,46 +53,57 @@ private EntityLocData CalcEntityLoc(string prototypeId)
...
                     {
                         // Only set name if the value isn't empty.
                         // So that you can override *only* a desc/suffix.
-                        name = ctx.Format(msg.Value);
+                        name = bundle.FormatPattern(msg.Value, null, out var fmtErr);
+                        WriteWarningForErrs(fmtErr, locId);
                     }
 
-                    if (mAttribs != null && mAttribs.Count != 0)
+                    IList<FluentError> errs = new List<FluentError>();
```
PJB3005: This list allocation is never used since you exclusively use it with `out` parameters.

Robust.Shared/Localization/LocalizationManager.Entity.cs
```diff
@@ -52,46 +53,57 @@ private EntityLocData CalcEntityLoc(string prototypeId)
...
+                    if (msgAttrs.Count != 0)
                     {
-                        if (desc == null && mAttribs.TryGetValue("desc", out var mDesc))
+                        if (desc == null
+                            && bundle.TryGetMsg(locId, "desc", null, out errs, out desc))
                         {
-                            desc = ctx.Format(mDesc);
                         }
```
PJB3005: 1. Generally, `TryGet` methods leave their out parameter in an undefined state when they return false, not necessarily a null state. It's kind of unclean to assume "out parameter is null" as a success check here.
2. This looks *really* weird with the empty if statement.

Robust.Shared/Localization/LocalizationManager.Entity.cs
```diff
@@ -52,46 +53,57 @@ private EntityLocData CalcEntityLoc(string prototypeId)
...
 
-                        if (suffix == null && mAttribs.TryGetValue("suffix", out var mSuffix))
+                        WriteWarningForErrs(errs, locId);
```
PJB3005: You can move all these checks to the end of the loop, tiny improvement.

Robust.Shared/Localization/LocalizationManager.Functions.cs
```diff
@@ -180,39 +183,53 @@ private ILocValue FuncProper(LocArgs args)
...
-            foreach (var (k, v) in fluentOptions)
+            var options = new Dictionary<string, ILocValue>(namedArgs.Count);
+            foreach (var (k, v) in namedArgs)
             {
-                options.Add(k, ValFromFluent(v));
+                options.Add(k, v.ToLocValue());
             }
 
             var argStruct = new LocArgs(args, options);
+            return function.Invoke(argStruct);
```
PJB3005: I touched on this earlier but returning the value directly here creates a complete spaghetti where *some* numbers in fluent are `FluentNumber` and *some* are `LocValueNumber`.

Robust.Shared/Localization/LocalizationManager.cs
```diff
@@ -73,158 +68,100 @@ public string GetString(string messageId, params (string, object)[] args0)
...
-                node = null;
+                value = null;
                 return false;
             }
 
-            if (attribName != null)
+            try
             {
-                if (message.Attributes == null || !message.Attributes.TryGetValue(attribName, out var attrib))
+                bundle.TryGetAttrMsg(messageId, args, out var errs, out value);
```
PJB3005: You should probably really return the return of this call here instead of `errs.Count == 0`.

Robust.Shared/Localization/ILocalizationManager.cs
Ygg01: Strange, I got offered this refactor by IntelliJ, maybe I set the version of C# too high

PR #1761, opened 2021-05-21
Directed event improvements

### LINKED ISSUE #1730: Directed eventbus should also work with ComponentReference

gradientvera:
~~- Subscriptions for <TComp, TEvent> pairs can be duplicated~~ Bad idea.
- Component references are taken into account and added/removed from entity event tables.
- Component references are taken into account when dispatching component events.

**Do NOT merge without @Acruid's explicit review and approval**

Fixes #1730


Acruid (approved):
Just leaving a note that DispatchComponent (might still be) an optimized version of Dispatch, so it requiring the GetReferences call is still a little questionable. If ~~Acruid~~ somebody cares they can look into it at a later date. This PR only has a perf impact on Component Lifetime events, and general directed events should stay the same. This is acceptable.

PR #1766, opened 2021-05-22
Adds overloads for entity system methods with out parameters.

gradientvera:
Okay, I know this seems silly but bear with me here for a second.
This will (imo) improve QOL, as you can use `new WhateverEvent()` in the method calls to raise them, and have it as the return value to do funny pattern matching with.


PJB3005:
Yeah I really don't like this...

RemieRichards:
Smh she didnt even update the title.

gradientvera:
:pensive:

ZoldorfTheWizard:
instamerging

PR #1769, opened 2021-05-23
Add loglevel command line arg

ShadowCommander:
Adds a command line arg to change the log level of a sawmill.

It'd be nice to have this so that we can get better logs from users who are having problems.


PJB3005 (commented):
Maybe add this on the server too?

PaulRitter (approved):
gamin. and yes, please add it on server too

ShadowCommander:
Added it to the server too, but I'm not sure what to do about overwriting log levels that get added by Content, since they get loaded after this gets run.

PR #1770, opened 2021-05-23
Make NetMessage name and group virtual to remove required boilerplate

DrSmugleaf:
Before the convention was to place this in every class that inherited NetMessage:
```csharp
#region REQUIRED
public static readonly MsgGroups GROUP = MsgGroups.Command;
public static readonly string NAME = nameof(YourNetMessage);
public YourNetMessage(INetChannel channel) : base(NAME, GROUP) { }
#endregion
```

That is no longer necessary.
The above example would be replaced by overriding the name and group properties where necessary, with the name property defaulting to the type's name.
The IL code that creates net messages has been changed to account for this. The old way still works, but the constructor with two arguments in NetMessage for it is marked as obsolete.

A method overload has also been added to INetManager.RegisterNetMessage that automatically resolves the name for you as that was done manually in the same way everywhere that I could see. The old method still works, but is also marked as obsolete.
Before you would need to do this:
```csharp
_netManager.RegisterNetMessage<MsgViewVariablesReqSession>(MsgViewVariablesReqSession.NAME,
                _msgReqSession);
 _netManager.RegisterNetMessage<MsgViewVariablesReqData>(MsgViewVariablesReqData.NAME, _msgReqData);
 _netManager.RegisterNetMessage<MsgViewVariablesModifyRemote>(MsgViewVariablesModifyRemote.NAME,
                _msgModifyRemote);
_netManager.RegisterNetMessage<MsgViewVariablesCloseSession>(MsgViewVariablesCloseSession.NAME,
                _msgCloseSession);
_netManager.RegisterNetMessage<MsgViewVariablesDenySession>(MsgViewVariablesDenySession.NAME);
_netManager.RegisterNetMessage<MsgViewVariablesOpenSession>(MsgViewVariablesOpenSession.NAME);
_netManager.RegisterNetMessage<MsgViewVariablesRemoteData>(MsgViewVariablesRemoteData.NAME);
```

The following now does the same:
```csharp
_netManager.RegisterNetMessage<MsgViewVariablesReqSession>(_msgReqSession);
_netManager.RegisterNetMessage<MsgViewVariablesReqData>(_msgReqData);
_netManager.RegisterNetMessage<MsgViewVariablesModifyRemote>(_msgModifyRemote);
_netManager.RegisterNetMessage<MsgViewVariablesCloseSession>(_msgCloseSession);
_netManager.RegisterNetMessage<MsgViewVariablesDenySession>();
_netManager.RegisterNetMessage<MsgViewVariablesOpenSession>();
_netManager.RegisterNetMessage<MsgViewVariablesRemoteData>();
```


PJB3005:
Is this really necessary to do with attributes? It feels like there'd be easier ways to clean up this boilerplate, like a virtual property.

DrSmugleaf:
> Is this really necessary to do with attributes? It feels like there'd be easier ways to clean up this boilerplate, like a virtual property.

I can do that too, I wasn't entirely sure which one would better.

DrSmugleaf:
> Is this really necessary to do with attributes? It feels like there'd be easier ways to clean up this boilerplate, like a virtual property.

Done in https://github.com/space-wizards/RobustToolbox/pull/1770/commits/7ed566378cbc817d16d1fbe5ff96523092e2729d
I imagine MsgGroup should probably be abstract as it defaults to MsgGroup.Error, but that would require a change to all net messages.

Robust.UnitTesting/RobustIntegrationTest.NetManager.cs
```diff
@@ -282,6 +282,11 @@ public void ServerSendToMany(NetMessage message, List<INetChannel> recipients)
                     _callbacks.Add(typeof(T), msg => rxCallback((T) msg));
             }
 
+            public void RegisterNetMessage<T>(ProcessMessage<T>? rxCallback = null, NetMessageAccept accept = NetMessageAccept.Both) where T : NetMessage
+            {
+                RegisterNetMessage(typeof(T).Name, rxCallback, accept);
```
PJB3005: This is incorrect. You should be using `MsgName` here.

Robust.Shared/Network/NetManager.cs
```diff
@@ -928,16 +953,35 @@ public T CreateNetMessage<T>()
...
+
+                gen.Emit(OpCodes.Newobj, constructor);
+                gen.Emit(OpCodes.Stloc_0);
+
+                gen.Emit(OpCodes.Ldloc_0);
+                gen.Emit(OpCodes.Ldnull);
+                gen.Emit(OpCodes.Stfld, channelField);
+
+                gen.Emit(OpCodes.Ldloc_0);
+                gen.Emit(OpCodes.Ret);
```
PJB3005: You don't have to do this whole dance to set channel since it'll already be null.

Robust.Shared/Network/NetManager.cs
```diff
@@ -868,22 +867,40 @@ private void CacheNetMsgFunction(byte id)
...
+                gen.Emit(OpCodes.Newobj, constructor);
+                gen.Emit(OpCodes.Stloc_0);
+
+                gen.Emit(OpCodes.Ldloc_0);
+                gen.Emit(OpCodes.Ldarg_0);
+                gen.Emit(OpCodes.Stfld, channelField);
+
+                gen.Emit(OpCodes.Ldloc_0);
+                gen.Emit(OpCodes.Ret);
+            }
```
PJB3005: This whole code predates the change I made where the constructor of the net message isn't used to set the channel anymore and can honestly be removed.

PJB3005 (changes requested):

Robust.Shared/Network/NetManager.cs
DrSmugleaf: If I remove that branch the new way of doing net messages (with no constructor) introduced with this PR won't work, neither will the old one be supported for content if I remove the branch above it. Should I only create the object here or am I understanding you incorrectly?

Robust.UnitTesting/RobustIntegrationTest.NetManager.cs
DrSmugleaf: Done in https://github.com/space-wizards/RobustToolbox/pull/1770/commits/475f027bd55ced7516e301bda643717f0e257af1

Robust.Shared/Network/NetManager.cs
DrSmugleaf: Done in https://github.com/space-wizards/RobustToolbox/pull/1770/commits/475f027bd55ced7516e301bda643717f0e257af1

Robust.Shared/Network/NetManager.cs
DrSmugleaf: Nevermind I see what you mean. The first branch still needs to exist in order for content to function though. Changed in https://github.com/space-wizards/RobustToolbox/pull/1770/commits/dc135d0fb48197036d5c6523ffd65220a07ae377

PJB3005 (approved):

PR #1742, opened 2021-05-07
Prototype References

PaulRitter:
fixes #1650

untested, smug will provide tests, dnm

also tbd on the struct variant, generics not mix well with nullability

PR #1773, opened 2021-05-26
Add test for YAML hot reloading

DrSmugleaf:
In the interest of https://github.com/space-wizards/RobustToolbox/pull/1742


ZoldorfTheWizard:
so does this need to be merged before or after #1742 and can I gamermerge the two branches

DrSmugleaf:
> so does this need to be merged before or after #1742 and can I gamermerge the two branches

This goes first, then #1742 is vibe checked to make sure it actually works with hot reloading by updating this test there.

Issue #1751, opened 2021-05-10
Actor component cleanup.


gradientvera:
My proposal for cleaning up BasicActorComponent and IActorComponent is the following,
- [x] Remove IActorComponent. (#1752)
- [x] Remove PlayerAttachedMsg and PlayerDetachedMsg component messages. (#1832)
- [x] Rename PlayerAttachSystemMessage to PlayerAttachedEvent. (#1774)
- [x] Rename PlayerDetachedSystemMessage to PlayerDetachedEvent. (#1774)
- [x] Rename BasicActorComponent to ActorComponent. (#1752)
- [x] Create ActorSystem to handle players being attached to entities, and detached from them in an ECS manner. (#1774)

There's an awful antipattern lingering in our codebase, and that is having useless interfaces that only a single class will *ever* inherit. Content code will be split between using the interface and the actual class implementing it, and spaghetti will ensure.

PR #1774, opened 2021-05-27
Adds ActorSystem to handle Attaching/Detaching players to/from entities sanely.

gradientvera:
ActorSystem, and not PlayerSession, is now in charge of adding/removing ActorComponent in entities.

### DIFF
--- a/Robust.Server/GameObjects/EntitySystems/ActorSystem.cs
+++ b/Robust.Server/GameObjects/EntitySystems/ActorSystem.cs
@@ -39,6 +39,7 @@ namespace Robust.Server.GameObjects
 
             actor = ComponentManager.AddComponent<ActorComponent>(args.Entity);
             actor.PlayerSession = args.Player;
+            args.Player.SetAttachedEntity(args.Entity);
             args.Result = true;
 
             args.Entity.SendMessage(actor, new PlayerAttachedMsg(args.Player));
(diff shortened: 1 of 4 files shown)


Robust.Server/GameObjects/EntitySystems/ActorSystem.cs
```diff
@@ -0,0 +1,141 @@
...
+                RaiseLocalEvent(uid, new DetachPlayerEvent());
+                args.ForceKicked = actor.PlayerSession;
+            }
+
+            actor = ComponentManager.AddComponent<ActorComponent>(args.Entity);
+            actor.PlayerSession = args.Player;
+            args.Player.SetAttachedEntity(args.Entity);
+            args.Result = true;
+
+            args.Entity.SendMessage(actor, new PlayerAttachedMsg(args.Player));
```
mirrorcult, line 59: is there an alternative to using this since it's marked as deprecated? or is it even deprecated?

mirrorcult: ah i see you have removing these msgs listed in #1751 so i'll assume thats for another pr

gradientvera: Yup, sadly some things in content still use these so I'll leave it for a future PR

Robust.Server/GameObjects/EntitySystems/ActorSystem.cs
```diff
@@ -0,0 +1,171 @@
...
+{
+    /// <summary>
+    ///     System that handles players being attached/detached from entities.
+    /// </summary>
+    [UsedImplicitly]
+    public class ActorSystem : EntitySystem
+    {
+        public override void Initialize()
+        {
+            SubscribeLocalEvent<AttachPlayerEvent>(OnActorPlayerAttach);
```
metalgearsloth: Should call base.Initialize();?

metalgearsloth (approved):

Robust.Server/GameObjects/EntitySystems/ActorSystem.cs
gradientvera: I can add it, but rn base.Initialize is empty

PR #1775, opened 2021-05-28
Add new fluent functions for words that change based on gender

mirrorcult:
Needed for and requires space-wizards/space-station-14#4071

### DIFF
--- a/Robust.Shared/Localization/LocalizationManager.Functions.cs
+++ b/Robust.Shared/Localization/LocalizationManager.Functions.cs
@@ -49,7 +49,6 @@ namespace Robust.Shared.Localization
         private ILocValue FuncCapitalize(LocArgs args)
         {
             var input = args.Args[0].Format(new LocContext());
-            System.Console.WriteLine(input);
             if (!String.IsNullOrEmpty(input))
                 return new LocValueString(input.First().ToString().ToUpper() + input.Substring(1));
             else return new LocValueString("");
--- a/Robust.Shared/Localization/LocalizationManager.Functions.cs
+++ b/Robust.Shared/Localization/LocalizationManager.Functions.cs
@@ -1,9 +1,12 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using Fluent.Net;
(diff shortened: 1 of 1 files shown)


RemieRichards:
A lot of these smell remarkably... "englishy"

mirrorcult:
what exactly is the alternative? other localizations will use different .ftl files and so won't need these functions and will write their own for the language they're localizing for

RemieRichards:
Exactly, they'll use their own ftl files, which ideally wouldn't have access to such locale-specific functions.

I wonder if when registering functions we can annotate what locales its intended for, and have it simply not work/return a warning when used in "invalid" locales. (For example, not all languages will need or want to implement a `PROPER()`)

It would've been nice to implement `THE()` and `PROPER()` and so on in ftl directly, as terms, but sadly you can't pass entities around as arguments inside of fluent itself (only from C#->ftl)

mirrorcult:
locale-specific functions would work for this, yeah, but its out of scope as far as im concerned (unless its very easy in which case i will do it)

RemieRichards:
I think it's both. (out of scope of this PR, and easy), so up to you.

It'd just be nice to curtail it before we have like 50 bajillion functions (this pr *does* increase the number of functions by a lot, which is what got me thinking, is all)

Robust.Shared/Localization/LocalizationManager.Functions.cs
```diff
@@ -1,33 +1,63 @@
...
             return new LocValueString(GetString("zzzz-the", ("ent", args.Args[0])));
         }
 
+        /// <summary>
+        /// Returns the string passed in, with the first letter capitalized.
+        /// </summary>
+        private ILocValue FuncCapitalize(LocArgs args)
+        {
+            var input = args.Args[0].Format(new LocContext());
+            System.Console.WriteLine(input);
```
PaulRitter: grafik

PaulRitter (dismissed):

Robust.Shared/Localization/LocalizationManager.Functions.cs
mirrorcult: oops

ShadowCommander (approved):

PR #1778, opened 2021-05-30
ComponentHandleState event

gradientvera:
This is a directed component event. What this means is that this will only be dispatched directed to the component and its references. (I changed the enumerator method so it'll be dispatched to the component AFTER the references)
This allows us to handle component states from entity systems, therefore removing *even more* logic from component, and moving us closer to struct components, maybe.

The reason I haven't coded an event for getting the state is because that change wouldn't be backwards-compatible, we'd have to change EVERY SINGLE COMPONENT'S GET COMPSTATE to use the event instead.

Anyway, **do NOT merge without @Acruid and @PJB3005's explicit approvals.**


Acruid:
Is this a good idea? The component states are technically part of networking, they are not game logic. Component structs *can* have methods and properties on them to help access the data they contain.

gradientvera:
Mmmm, how would component structs handling states look? Would they use an interface or something?

Robust.Shared/GameObjects/Component.cs
```diff
@@ -228,6 +228,7 @@ public virtual ComponentState GetComponentState(ICommonSession player)
         }
 
         /// <inheritdoc />
+        [Obsolete("Use the ComponentHandleState event instead.")]
```
PJB3005: Probably a good idea to curb your enthusiasm on this and *not* mark something like this obsolete yet.

Robust.Shared/GameObjects/EntityEventBus.Directed.cs
```diff
@@ -294,9 +294,17 @@ public void Dispose()
...
+            /// </summary>
             private IEnumerable<Type> GetReferences(Type type)
             {
-                return _comFac.GetRegistration(type).References;
+                var list = _comFac.GetRegistration(type).References;
+
+                for (var i = list.Count-1; i >= 0; i--)
+                {
+                    yield return list[i];
+                }
```
PJB3005: `IComponentRegistration` makes no guarantee that the first element is the type itself. Filter to ignore the type in the inner loop and then do a final `yield return` at the end instead.

PJB3005 (changes requested):

Robust.Shared/GameObjects/Component.cs
gradientvera: My YEAST is SLOW.
Done.

Robust.Shared/GameObjects/EntityEventBus.Directed.cs
gradientvera: Done, that's indeed better

PJB3005 (approved):

PR #1783, opened 2021-05-30
Fix log assert breaking tests

ShadowCommander:
For some reason there's an assert here in logs. If the assert gets tripped while in a test then it'll fail the test.


DrSmugleaf:
This is done to make tests fail if a certain log level is reached during their execution.

DrSmugleaf:
See https://github.com/space-wizards/RobustToolbox/commit/ebf53248cfc2581914f6c57a69e165349adb84f6 and https://github.com/space-wizards/RobustToolbox/commit/de2e505a12389e8f04692ff4fb44739f14526e76

PR #1784, opened 2021-05-30
Pre-reqs for multithreaded physics

metalgearsloth:
Loosely based on bullet's implementation.

Islands are now stored on IslandManager which should also save memory costs as we don't need a massive island that can potentially store every body on it.

I also made it so there's a special island for all bodies that don't have any contacts so these all get put into the same island and have their own sleep behavior applied (the alternative is to have them all in separate islands, but if you had 100 players running around that's kinda wasteful).

I was getting a race condition leading to bodies flickering with the solver so I'll port the multi-threading later once I figure out what was causing it.


Robust.Shared/Physics/IslandManager.cs
```diff
@@ -0,0 +1,190 @@
+/*
```
PJB3005, line 1: Since you're adding Bullet code here you should add the license to `Resources/EngineCredits/Libraries.yml`. Minor thing.

Robust.Shared/Physics/IslandManager.cs
```diff
@@ -0,0 +1,190 @@
...
+        PhysicsIsland AllocateIsland(int numBodies, int numContacts, int numJoints);
+
+        IReadOnlyList<PhysicsIsland> GetActive { get; }
+    }
+
+    /// <summary>
+    /// Contains physics islands for use by PhysicsMaps.
+    /// Key difference from bullet3 is we won't merge small islands together due to the way box2d sleeping works.
+    /// </summary>
+    public class IslandManager : IIslandManager
```
PJB3005: Make this internal?

Robust.Shared/Physics/IslandManager.cs
```diff
@@ -0,0 +1,190 @@
...
+    {
+        private IslandBodyCapacitySort _capacitySort = new();
+        private IslandBodyCountSort _countSort = new();
```
PJB3005: Use `private static readonly` for these instead maybe?

PJB3005 (commented):
A few tiny nits

metalgearsloth:
Reviews done but I'd prefer to wait for the WorldAABB update before merging

PR #1785, opened 2021-05-30
Make awake dispatch local events

### LINKED ISSUE #4113: Spears do contact damage when dropped
How to reproduce:

- Make a spear.
- Go into combat mode.
- Repeatedly drop the spear onto something while mashing clicks to pick it up as soon as it's dropped.
- The spear will rapidly chew through pretty much anything it touches, up to and including reinforced walls.

image

It seems like this bug is kind of inconsistent and lag-dependent, but it does work for some people.

ShadowCommander:
Fixes space-wizards/space-station-14#4113


gradientvera (approved):

gradientvera:
This is minor enough that you could've just commited it directly, tbh

PR #1789, opened 2021-05-31
Bandaid GridTileLookup containers

metalgearsloth:
I'm hoping tile entities deprecates this system but this should fix our problems for now.


Robust.Shared/Containers/ContainerManagerComponent.cs
```diff
@@ -32,6 +32,17 @@ public class ContainerManagerComponent : Component, IContainerManager
...
+        public IEnumerable<IContainer> Containers
+        {
+            get
+            {
+                foreach (var (_, container) in _containers)
+                {
+                    yield return container;
+                }
+            }
+        }
```
gradientvera: Why not use the `GetAllContainers` method instead of adding this?

metalgearsloth: Because I was retarded and didn't see it until after

gradientvera (approved):

metalgearsloth:
quick on the draw

PR #1794, opened 2021-05-31
Add joint support for CollisionWakeComponent

metalgearsloth:
If the component has any joints then it can't disable itself until all of them are removed.

### DIFF
--- a/Robust.Shared/GameObjects/Components/CollisionWakeComponent.cs
+++ b/Robust.Shared/GameObjects/Components/CollisionWakeComponent.cs
@@ -33,12 +33,6 @@ namespace Robust.Shared.GameObjects
             }
         }
 
-        public override void Initialize()
-        {
-            base.Initialize();
-            RaiseStateChange();
-        }
-
         internal void RaiseStateChange()
         {
             Owner.EntityManager.EventBus.RaiseLocalEvent(Owner.Uid, new CollisionWakeStateMessage(), false);
--- a/Robust.Shared/GameObjects/Components/CollisionWakeComponent.cs
+++ b/Robust.Shared/GameObjects/Components/CollisionWakeComponent.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using Robust.Shared.Physics;
 using Robust.Shared.Players;
 using Robust.Shared.Serialization;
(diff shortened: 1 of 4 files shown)


ShadowCommander (approved):

Robust.Shared/GameObjects/Components/CollisionWakeComponent.cs
```diff
@@ -32,7 +33,13 @@ public bool Enabled
             }
         }
 
-        private void RaiseStateChange()
+        public override void Initialize()
+        {
+            base.Initialize();
+            RaiseStateChange();
+        }
```
gradientvera: Logic in component detected.
HERESY! HERESY!

Robust.Shared/Physics/JointMessages.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public sealed class JointRemovedEvent : EntityEventArgs
+    {
+        public Joint Joint { get; }
+
+        public JointRemovedEvent(Joint joint)
+        {
+            Joint = joint;
+        }
+    }
```
gradientvera: Despite `Joint` having both bodies, these events don't and they should.
This will make it confusing when you're using the directed events and want to, for example, get the other body to perform logic on it. I'd rather not have to check each body in the joint to check it's not "mine"...

Robust.Shared/Physics/Dynamics/PhysicsMap.cs
```diff
@@ -220,6 +220,11 @@ public void AddJoint(Joint joint)
             bodyA.Dirty();
             bodyB.Dirty();
             // Note: creating a joint doesn't wake the bodies.
+
+            // Raise broadcast last so we can do both sides of directed first.
+            _entityManager.EventBus.RaiseLocalEvent(bodyA.Owner.Uid, new JointAddedEvent(joint), false);
+            _entityManager.EventBus.RaiseLocalEvent(bodyB.Owner.Uid, new JointAddedEvent(joint), false);
+            _entityManager.EventBus.RaiseEvent(EventSource.Local, new JointAddedEvent(joint));
```
gradientvera: allocating three times? sad
Apply the review ~~below~~ above and then reduce the allocations here to two events only, imo.
**(THIS REVIEW APPLIES TO THE JOINTREMOVEDEVENT BELOW AS WELL)**

gradientvera (changes requested):

Robust.Shared/Physics/JointMessages.cs
metalgearsloth: In that instance that means 2 events getting allocated if that's okay?

Robust.Shared/GameObjects/Components/CollisionWakeComponent.cs
metalgearsloth: image

PR #1795, opened 2021-05-31
Dedicated AABB-AABB and AABB-Circle solvers

metalgearsloth:
Draft so I can push through the motivation to finish it.

Main problem is I can't just straight use Randy Gaul's code as Box2d requires more data (LocalPoint for each point) so I'm adding a bunch of test cases around it to ensure parity.

Currently it's not parity, I just need to add more tests.


metalgearsloth:
Making this work properly with grid rotation is gonna be a pain so closing for now.

PR #1797, opened 2021-06-02
Fix light pop-in

### LINKED ISSUE #1754: Really bad lighting pop-in
## Description

happens basically everywhere but bridge is most egregious

https://user-images.githubusercontent.com/19853115/116831003-d180f900-ab61-11eb-8fa0-959cf088f05b.mp4

**Additional context**

metalgearsloth:
D'oh.

Fixes https://github.com/space-wizards/RobustToolbox/issues/1754

Not sure if it's the best way to go about it but the problem is that lights / sprites that intersect 2 grids get drawn twice.

I also added a debug command to show every light in the viewport.

### DIFF
--- a/Robust.Client/Console/Commands/LightBBCommand.cs
+++ b/Robust.Client/Console/Commands/LightBBCommand.cs
@@ -1,0 +1,17 @@
+using Robust.Client.GameObjects;
+using Robust.Shared.Console;
+using Robust.Shared.GameObjects;
+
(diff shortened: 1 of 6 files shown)


metalgearsloth:
Gonna re-do this so lights are only ever on their grid.

20kdc:
Two grids don't usually exist on Saltern. I'm curious as to how this affects the existing bug.

PR #1798, opened 2021-06-02
Fix light pop-in 2

metalgearsloth:
Problem was that lights intersecting 2 grids would show up twice if you're iterating through both grids. Storing them only on 1 grid and then enlarging to the maximum range seemed better probably? This is more of a vibe check as I'll prolly clean it up some more.


Robust.Client/GameObjects/Components/Light/PointLightComponent.cs
```diff
@@ -138,6 +139,9 @@ public bool VisibleNested
             set => _visibleNested = value;
         }
 
+        // TODO: Test to validate all yamls meet this value
+        public const float MaxRadius = 10.0f;
+
```
metalgearsloth: I'm pretty suss on this; it's mainly used to know how much we need to enlarge the viewport by to find any possibly intersecting render trees (given each comp is only stored on 1 tree).

Open to suggestions.

metalgearsloth:
Should be gucci for review. I wanted to write a test but it'd be significantly easier as an integration one (given it requires the map entity and grid entities) but that needs doing content-side.

Fixes https://github.com/space-wizards/space-station-14/issues/3922

metalgearsloth:
Getting opinion on my comment and merging in a week if no objections

Robust.Client/GameObjects/Components/Light/PointLightComponent.cs
metalgearsloth: Might make it a cvar instead and just warn if anything is higher or something...

mirrorcult: definitely should make it a cvar and have it just warn on a higher value imo, it's not a fatal error to have it higher, it would just look ugly because of pop-in, i assume?

metalgearsloth: Ye

PR #1802, opened 2021-06-06
Fix GetWorldAABB again

metalgearsloth:
No weird-ass bug around 0,0 this time. There was also an issue where occasionally the AABB on the client would disappear if you had a bunch of objects and used showbb but I think it's fixed now.


metalgearsloth:
I'm gonna merge this when I'm free as the last pr was approved and this code is a copy-paste of the working bits from it.

PJB3005:
@metalgearsloth `GridTileLookupTest` integration test fails. Any idea?

metalgearsloth:
I changed the scaling on AABBs for gridtilelookup so the content test returns different amounts which needs the content PR.

PR #1813, opened 2021-06-07
MapInit ECS event

mirrorcult:
it is what it is

Required for space-wizards/space-station-14#4154


gradientvera (approved):
Good ECS PR

gradientvera:
Given the event instance doesn't have any data in it, maybe you can hold an static instance to it somewhere btw, that would reduce allocations and make your code go BRRR

mirrorcult:
addressed review, idk if thats correct or not

gradientvera (approved):

gradientvera:
aye, that works, good PR

PR #1814, opened 2021-06-07
Removed old Loc.GetString() use instances

GalacticChimp:
Requires https://github.com/space-wizards/space-station-14/pull/4155 - client & shared side changes

**WARNING!** They need to be merged at the same time (this one preferably second)!


GalacticChimp:
Server build fails because it needs some stuff from Client's PR Shared parts

Robust.Client/Console/Commands/Scripting.cs
```diff
@@ -8,7 +8,7 @@ namespace Robust.Client.Console.Commands
     internal sealed class ScriptCommand : IConsoleCommand
     {
         public string Command => "csi";
-        public string Description => "Opens a C# interactive console.";
+        public string Description => Loc.GetString("console-script-command-description");
```
PJB3005: I'd name these something like `concmd-Script-desc`.

Though I'm also not sure we should really bother localizing console commands?

PJB3005 (commented):
Move localization strings that are used from engine into the engine, please. 

(I'm assuming you have them in content right now)

Robust.Client/Console/Commands/Scripting.cs
GalacticChimp: I can revert console command changes, I simply wasn't sure if we needed them

GalacticChimp: ok, removed Loc.GetString() from console commands, they didn't make sense

GalacticChimp:
ok, I think everything is fixed now

PR #1815, opened 2021-06-08
Add directed events for player attach/detach

DrSmugleaf:


gradientvera (approved):

Robust.Client/GameObjects/EntitySystems/InputSystem.cs
```diff
@@ -173,4 +173,24 @@ public PlayerAttachSysMessage(IEntity? attachedEntity)
...
+    }
+
+    public class PlayerDetachedEvent : EntityEventArgs
+    {
+        public PlayerDetachedEvent(IEntity old)
+        {
+            Old = old;
+        }
+
+        public IEntity Old { get; }
```
gradientvera: Could you call this `Entity` so it's consistent with the server-side equivalent?

gradientvera (approved):

Robust.Client/GameObjects/EntitySystems/InputSystem.cs
DrSmugleaf: Done in https://github.com/space-wizards/RobustToolbox/pull/1815/commits/cd660dbdb3833451b0cd0b29638f47ef6af0f22e

PR #1816, opened 2021-06-08
Add SetAndDirtyIfChanged extension method

DrSmugleaf:
In the interest of replacing this

With this 

It works like [RaiseAndSetIfChanged in ReactiveUI](https://www.reactiveui.net/api/reactiveui/ireactiveobjectextensions/554ad0a8) but it returns the component instead of the value being set for method chaining.

### DIFF
--- a/Robust.Shared/GameObjects/ComponentExt.cs
+++ b/Robust.Shared/GameObjects/ComponentExt.cs
@@ -95,7 +95,7 @@ namespace Robust.Shared.GameObjects
             return entity.AddComponent<T>();
         }
 
-        public static TValue SetAndDirtyIfChanged<TComp, TValue>(
+        public static TComp SetAndDirtyIfChanged<TComp, TValue>(
             this TComp comp,
             ref TValue backingField,
             TValue value)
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/ComponentExt.cs
```diff
@@ -93,5 +94,22 @@ public static T EnsureComponentWarn<T>(this IEntity entity, string? warning = nu
 
             return entity.AddComponent<T>();
         }
+
+        public static TValue SetAndDirtyIfChanged<TComp, TValue>(
```
PJB3005: I don't think you need a generic for `TComp` here?

PJB3005 (changes requested):

Robust.Shared/GameObjects/ComponentExt.cs
DrSmugleaf: Correct my bad, I'll fix it in a sec.

DrSmugleaf: Fixed in https://github.com/space-wizards/RobustToolbox/pull/1816/commits/30073ffd260826d17711514dad578c6d1a6fcf2a

PaulRitter (approved):
attaboy

i gotta ask one thing though: would this still work on struct components?

PJB3005:
> i gotta ask one thing though: would this still work on struct components?

There are *way* more problems to solve with struct components than this minor one. The mere presence of the dirty flag is problematic for struct components. Not really worth considering at the moment.

PJB3005 (approved):

PR #1818, opened 2021-06-08
Add more tolerance for if player sessions are connected to dead entities

20kdc:
One may hope this stops the servers getting stuck throwing exceptions on round restart and never actually restarting the round.
This can be caused by various sorts of Bad Stuff (tm).

### DIFF
--- a/Robust.Server/Player/PlayerSession.cs
+++ b/Robust.Server/Player/PlayerSession.cs
@@ -133,6 +133,17 @@ namespace Robust.Server.Player
             if (AttachedEntity == null)
                 return;
 
+#if EXCEPTION_TOLERANCE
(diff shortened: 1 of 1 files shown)


PJB3005:
This fails to compile, apparently.

20kdc:
...wait what
hold on, will fix

20kdc:
So apparently the reference to an interface method via the direct name `SetAttachedEntity` was the reason for the error, but further, I wasn't building in Release mode, so my code wasn't actually in my build test...

Re-built and re-ran on release with build fix. Doesn't seem to break anything and logically shouldn't under the context.

gradientvera (approved):

Issue #1772, opened 2021-05-26
Making everything ECS would render ViewVariables useless.


gradientvera:
This is easier to explain with an example.
Before StackComponent in SS14 was stripped of all logic, it had a `Count` property with a setter that would `Dirty()` the component, change appearance, raise an event, etc.
After I refactored it, `Count` is a simple auto-property with no special logic in the setter. You need to raise a `StackChangeCount` event in order to change this property, and an EntitySystem will take care of all the logic `Count`'s setter did before.
However, now editing `Count` in VV is not ideal. It doesn't perform any of the logic it should, and there's no current way to fix this.
I am not sure how this should be handled, or what should be done in order to ensure VV doesn't become useless after we transition to a more ECS approach of coding content.

RemieRichards:
It'd be *half* useless, given *VIEW* Variables is also for viewing.

Might be fine to just add appropriate properties specifically for VV, which do have logic in them (this technically violates the ECS nature, but given that the actual main flow of the code would still *not* be going through them, it'd be fine, no?)

gradientvera:
That is a solution indeed, but I don't particularly like it due to its hacky nature, hmm... If nobody can come up with a better solution, I think we'll just have to do that.

Acruid:
Another solution is to use CSI instead of VV.

gradientvera:
> Another solution is to use CSI instead of VV.

Sadly this is not an option, as we can't give every admin SCSI perms

RemieRichards:
CSI is way too dangerous to give to the average admin

PaulRitter:
i guess the easiest way would be to make the entitysystem somehow subscribe to changes to the autoproperty. am thinking emitting some funny il into the component, then having a generic method on the entitysystem that can just subscribe to changes to a specific field. maybe even just emit the funny il if a function-call for the field in question exists?

imo this would also be a function specific to VV, so you could just go "every field that has the system subscribe to changes can be edited, the others are readonly"?

gradientvera:
Having to subscribe to every single property that can be changed by VV on entity systems seems far from ideal, but I can see why.
One problem is that the EntitySystems themselves change the property when they get the event, so you have to somehow distinguish between something like VV/animatable changing it and the entity system changing it when handling the event itself.

gradientvera:
So I had a discussion with Paul on discord today and he proposed a [great idea](https://discord.com/channels/310555209753690112/770682801607278632/852446307981328384).
[...]

PR #1821, opened 2021-06-10
Adds ReflectRaiseEvent attribute, add SetValue method to IReflectionManager

gradientvera:
When you set members (fields and properties only) through ReflectionManager, it will check if they have the ReflectRaiseEvent attribute and if so raise an event. If you pass the entity uid, the event will be raised directed. If the attribute on your member has `broadcast` set to true, it will raise the event broadcast regardless of uid.
When an event is raised, the actual data in the member isn't modified, as we assume the event does so.
This makes VV able to raise events when it changes properties and fields that require it.

*VV seriously needs a rewrite, I swear...*

TODO:
- Change Animations to use this
- Tests

### DIFF
--- a/Robust.Client/ViewVariables/IViewVariablesManager.cs
+++ b/Robust.Client/ViewVariables/IViewVariablesManager.cs
@@ -1,3 +1,4 @@
+using Robust.Shared.GameObjects;
 using Robust.Shared.ViewVariables;
 
 namespace Robust.Client.ViewVariables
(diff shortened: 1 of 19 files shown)


PJB3005:
What's the use of this? Why not use properties?

gradientvera:
> What's the use of this? Why not use properties?

https://github.com/space-wizards/RobustToolbox/issues/1772#issuecomment-858390943

What do you mean by using properties?

PJB3005:
So the idea is that you can set the `Count` field on a stack, and then VV goes through this channel to raise a message to the entitysystem for the change?

gradientvera:
Exactly, you're meant to raise the event to change the `Count` property instead of changing it directly, as the entity system dirties the component, changes the appearance, etc.

This would be like the equivalent of doing something like:
```csharp
[DataField("count")]
public int Count { get; set; } = 30;

[ViewVariables(ReadWrite)]
public int CountVV
{
     get => Count;
     set => Owner.EntityManager.EventBus.RaiseLocalEvent(Owner.Uid, new StackCountChangeEvent(value));
}
```

PJB3005:
> This would be like the equivalent of doing something like:

Then do that. This is an absolutely atrocious solution.

PR #1826, opened 2021-06-13
Upgrade Linguini to latest (prevents IDE issue with SG)

Ygg01:
Basically, Source Generators with `netstandard 2.1` aren't guaranteed to work with VS or IDEs or `dotnet msbuild` just `dotnet build` :( 

See https://github.com/dotnet/roslyn/issues/47087

So I downgraded SG to `netstandard 2.0`  but there are some nullability warnings that break the pipeline.


PaulRitter:
yep...

PR #1829, opened 2021-06-16
Engine Entity Anchoring

Acruid:
## Summary
This is basically full engine support for the content `AnchorableComponent`. It completely supersedes the existing SnapGrid system in engine. Check out the AnchoredSystem unit tests for more info and examples on how to use it.

The content `AnchorableComponent` now just changes the `TransformComponent.Anchored` flag. This flag should generally be managed by the construction system, random content code should not be setting it. Setting the physics body type to static DOES NOT anchor an entity, and there are only a few special cases where you want a non-anchored static entity (like the immovable station grid, or markers in space).

## Why This Should Be Merged
This is the foundation for #1719 and Tile Entities. Because of the key design requirement of "anchored entities cannot move locally", PVS, physics broadphase, and the snapgrid/GridTileLookup (used by atmos, power) acceleration data structures can be improved. This PR clearly defines the separation of "movable" and "level geometry"/Tile entities.

## Prototype Migration
Any prototype that previously wanted to spawn "anchored" would include the `SnapGridComponent`. Now you just need to set the `anchored` flag to true on TransformComponent.

Content Requires https://github.com/space-wizards/space-station-14/pull/4195.


Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -328,6 +347,9 @@ public Vector2 LocalPosition
                 if (_localPosition.EqualsApprox(value, 0.00001))
                     return;
 
+                if(Anchored)
```
ShadowCommander: Maybe move this above the EqualsApprox check?

ShadowCommander (approved):

PR #1830, opened 2021-06-17
Remove NetMessage deprecated boilerplate entirely

mirrorcult:
Removes all boilerplate/obsoleted uses of netmessage constructors/nameof stuff deprecated in #1770

Requires space-wizards/space-station-14#4196

### DIFF
--- a/Robust.Shared/Network/NetManager.cs
+++ b/Robust.Shared/Network/NetManager.cs
@@ -902,10 +902,11 @@ namespace Robust.Shared.Network
         #region NetMessages
 
         /// <inheritdoc />
-        public void RegisterNetMessage<T>(string name, ProcessMessage<T>? rxCallback = null,
+        public void RegisterNetMessage<T>(ProcessMessage<T>? rxCallback = null,
             NetMessageAccept accept = NetMessageAccept.Both)
-            where T : NetMessage
+            where T : NetMessage, new()
         {
+            var name = new T().MsgName;
             var id = _strings.AddString(name);
 
             _messages.Add(name, typeof(T));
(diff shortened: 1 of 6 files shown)


Robust.UnitTesting/RobustIntegrationTest.NetManager.cs
```diff
@@ -272,21 +272,16 @@ public void ServerSendToMany(NetMessage message, List<INetChannel> recipients)
             public event EventHandler<NetChannelArgs>? Connected;
             public event EventHandler<NetDisconnectedArgs>? Disconnect;
 
-            public void RegisterNetMessage<T>(string name, ProcessMessage<T>? rxCallback = null,
-                NetMessageAccept accept = NetMessageAccept.Both) where T : NetMessage
+            public void RegisterNetMessage<T>(ProcessMessage<T>? rxCallback = null, NetMessageAccept accept = NetMessageAccept.Both) where T : NetMessage, new()
             {
+                var name = new T().MsgName;
```
metalgearsloth, line 277: Is this used?

Robust.Shared/Network/NetManager.cs
```diff
@@ -902,10 +902,11 @@ private void CacheNetMsgFunction(byte id)
         #region NetMessages
 
         /// <inheritdoc />
-        public void RegisterNetMessage<T>(string name, ProcessMessage<T>? rxCallback = null,
+        public void RegisterNetMessage<T>(ProcessMessage<T>? rxCallback = null,
             NetMessageAccept accept = NetMessageAccept.Both)
-            where T : NetMessage
+            where T : NetMessage, new()
         {
+            var name = new T().MsgName;
```
metalgearsloth, line 909: Won't this allocate a new message?

Robust.UnitTesting/RobustIntegrationTest.NetManager.cs
mirrorcult: guess not, might have to update tests to use it?

Robust.Shared/Network/NetManager.cs
mirrorcult: idk, this is how it was done before, just moved into this function

Robust.UnitTesting/RobustIntegrationTest.NetManager.cs
metalgearsloth: Maybe just comment it out and leave a note explaining that it isn't used here but is used in the live one or something?

PR #1837, opened 2021-06-19
#1607 deprecate IEntityQuery

### LINKED ISSUE #1607: Deprecate IEntityQuery
There are other, better ways of doing it IMO:
image
* AllEntityQuery -> EntityManager.AllEntities
* IntersectingEntityQuery -> Collision lookup (SharedBroadPhaseSystem on new physics, IPhysicsManager on old)
* MultipleTypeEntityQuery -> ComponentManager
* PredicateEntityQuery -> ComponentManager
* TypeEntityQuery -> ComponentManager
* TypeEntityQuery<T> -> ComponentManager

GalacticChimp:
Fixes #1607


Robust.Shared/GameObjects/IEntityManager.cs
```diff
@@ -77,15 +76,19 @@ public interface IEntityManager
...
         IEnumerable<IEntity> GetEntities();
 
+
+        /// <summary>
+        /// Returns all entities which pass the provided filter
+        /// </summary>
+        /// <param name="filter">Filter query</param>
+        /// <returns></returns>
+        IEnumerable<IEntity> GetEntities(Func<IEntity, bool> filter);
+
```
metalgearsloth: This shouldn't exist IMO and should just use components.

This is because you need to iterate every entity which is atrociously expensive.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -2160,6 +2160,16 @@ public bool HasComponent(Type type)
...
+            public bool HasAllComponents(IEnumerable<IComponent> components)
+            {
+                return components.All(c => HasComponent(c.GetType()));
+            }
+
+            public bool HasAllComponents(IEnumerable<Type> components)
+            {
+                return components.All(c => HasComponent(c));
+            }
+
```
metalgearsloth: Delete? SpriteComp shouldn't have these in an ideal world.

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -251,6 +243,21 @@ public IEnumerable<IEntity> GetEntities()
...
+
+            Parallel.ForEach(
+                AllEntities.Where(filter),
+                e =>
+                    {
+                        entities.Add(e);
+                    });
+
+            return entities;
+        }
```
GalacticChimp: the standard GetEntities iterates over AllEntities List and uses indexes, to fetch entitities, I could not do that easily with Where() so I added this little code snippet to simplify things

Robust.Shared/GameObjects/ComponentManager.cs
```diff
@@ -370,6 +370,20 @@ public bool HasComponent(EntityUid uid, uint netId)
...
+        {
+            return HasAllComponents(uid, components.Select(c => c.GetType()));
+        }
+
+        /// <inheritdoc />
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool HasAllComponents(EntityUid uid, IEnumerable<Type> components)
+        {
+            return components.All(c => HasComponent(uid, c));
+        }
```
metalgearsloth: This is silly when it has to iterate every component when it can just directly look at the dictionary instead. What code's even using this still? As ideally they'd probably just use a helper with generics IMO (also why aggressively inline it?)

Robust.UnitTesting/Server/GameObjects/ThrowingEntityDeletion_Test.cs
```diff
@@ -55,14 +55,15 @@ public void Setup()
...
+        [TestCase("throwInAdd")]
+        [TestCase("throwsInInitialize")]
+        [TestCase("throwsInStartup")]
+        public void Test(string prototypeName)
         {
             Assert.That(() => EntityManager.SpawnEntity(prototypeName, MapCoordinates.Nullspace),
                 Throws.TypeOf<EntityCreationException>());
 
-            Assert.That(EntityManager.GetEntities().Where(p => p.Prototype?.ID == prototypeName), Is.Empty);
+            Assert.That(EntityManager.GetEntities(p => p.Prototype?.ID == prototypeName), Is.Empty);
```
GalacticChimp: turns out this is broken and the test hid it, the thrownInAdd etc entities do get spawned in, and they were spawned even without my code changes.

Robust.Shared/GameObjects/Entity.cs
```diff
@@ -244,6 +244,18 @@ public bool HasComponent(Type type)
...
+        {
+            return EntityManager.ComponentManager.HasAllComponents(Uid, components);
+        }
+
+        /// <inheritdoc />
+        public bool HasAllComponents(IEnumerable<Type> components)
+        {
+            return EntityManager.ComponentManager.HasAllComponents(Uid, components);
+        }
+
```
metalgearsloth: Anyone needing this should just be calling ComponentManager directly IMO.

Robust.Shared/GameObjects/EntityManager.cs
metalgearsloth: This is also needlessly expensive and should be deleted.

metalgearsloth (changes requested):

metalgearsloth:
I also should've mentioned anything that *really* needs EntityManager.AllEntities should be managing it itself because iterating it is hella expensive.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
GalacticChimp: this is in a SpriteComp's private class: private class DummyIconEntity : IEntity 
where most of the IEntity stuff isn't actually used

Robust.Shared/GameObjects/ComponentManager.cs
```diff
@@ -583,6 +583,16 @@ public IEnumerable<IComponent> GetAllComponents(Type type, bool includePaused =
...
 
+        /// <inheritdoc />
+        public IEnumerable<IEntity> GetAllEntitiesWithAllComponents(IEnumerable<Type> types)
+        {
+            var filteredTypes = types.Select(t => _entTraitDict[t]);
+            if (!filteredTypes.Any()) return Enumerable.Empty<IEntity>();
+            var fetchedEntities = filteredTypes.Select(x => x.Values.Select(y => y.Owner).ToList());
+            var intersectedEntities = fetchedEntities.Skip(1).Aggregate(new HashSet<IEntity>(fetchedEntities.First()), (h, e) => { h.IntersectWith(e); return h; });
+            return intersectedEntities.ToList();
+        }
```
GalacticChimp: decided to go with returning IEntities instead of Components, because we want the Entities which have all the inputted components, not the components themselves (see FindEntitiesWithComponents and DeleteEntitiesWithComponent for use cases)

Robust.Shared/GameObjects/ComponentManager.cs
```diff
@@ -256,26 +256,26 @@ private void RemoveComponentDeferred(Component component, EntityUid uid, bool re
...
-                if (!_deleteSet.Add(component))
-                {
-                    // already deferred deletion
-                    return;
-                }
+            if (!_deleteSet.Add(component))
+            {
+                // already deferred deletion
+                return;
+            }
```
metalgearsloth, line 270: Try to avoid just making tab or whitespace changes coz it can bloat reviews, this was done in a few spots.

Robust.Shared/GameObjects/ComponentManager.cs
metalgearsloth: Won't this allocate out the ass because of multiple ToList calls + a HashSet call? Also this is just duplicating what EntityQuery already does but in a less-robust way (e.g. doesn't check for deleted components).

IMO just delete.

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -251,6 +243,21 @@ public IEnumerable<IEntity> GetEntities()
             }
         }
 
+        /// <inheritdoc />
+        public IEnumerable<IEntity> GetEntities(Func<IEntity, bool> filter)
+        {
+            var entities = new ConcurrentBag<IEntity>();
+
+            Parallel.ForEach(
```
metalgearsloth: IMO just have the regular GetEntities. I'm suss of having it parallel because there's a bunch of non-thread-safe APIs someone may inadvertantly use (e.g. most of the public physics methods).

metalgearsloth (changes requested):

Robust.Shared/GameObjects/EntityManager.cs
GalacticChimp: this one isn't used actually, forgot to delete it

metalgearsloth (approved):

Robust.Shared/GameObjects/ComponentManager.cs
```diff
@@ -256,26 +256,26 @@ private void RemoveComponentDeferred(Component component, EntityUid uid, bool re
...
 
                 if(component.Running)
                     component.LifeShutdown();
 
                 if (component.LifeStage != ComponentLifeStage.PreAdd)
                     component.LifeRemoveFromEntity();
-                _componentDependencyManager.OnComponentRemove(uid, component);
-                ComponentRemoved?.Invoke(this, new RemovedComponentEventArgs(component, uid));
+            _componentDependencyManager.OnComponentRemove(uid, component);
+            ComponentRemoved?.Invoke(this, new RemovedComponentEventArgs(component, uid));
```
gradientvera: You forgot to unintend these lines, it's inconsistent right now.

metalgearsloth (approved):

PR #1838, opened 2021-06-25
MapGridSystem for grid initialization

metalgearsloth:
Doesn't seem to be a clean way to hook into grids being initialized (can't duplicate directed bus events and a few systems need to EnsureComponent) hence the event.


gradientvera (approved):

PJB3005 (changes requested):
I already added a `MapSystem` for a simialr purpose. Maybe add to that instead.

metalgearsloth:
Okey dokey

PR #1842, opened 2021-06-30
Remove Static Component NetIds

Acruid:
## What This Does
Removes the giant class of static component NetIds. Now they are dynamically generated at runtime.

## If This is Merged
* You don't have to rely on humans assigning static NetIds to components anymore.
* Trades the NetId property on components for an attribute, and reduces ComponentState boilerplate.
* Adds a test to make sure the sets of networked components on client and server are equivalent (spoiler: they were not, that is one Dirty() call away from an exception).
* NetIds are now ushort, instead of uint, so you only get 65K of them.

## Component Code Migration
Remove the now missing `Component.NetId` property, remove the old `ComponentState.base()` constructor call, add a `NetworkedComponentAttribute` to the component class.

Content Requires https://github.com/space-wizards/space-station-14/pull/4247


gradientvera (approved):
LGTM

PR #1629, opened 2021-03-14
Network View Bubble

Acruid:
Remake of the entire network bubble system.

> just simply cull entities outside of a view box, really, how hard could that be? *queue screams of the damned*

- [x] basic entity culling
- [x] mover support
- [x] allow per-client entity visibility (Resolves https://github.com/space-wizards/RobustToolbox/issues/1255)
- [ ] obscure container contents (per-entity vis) (https://github.com/space-wizards/RobustToolbox/issues/1432)
- [x] multiple viewports (Resolves https://github.com/space-wizards/RobustToolbox/issues/1257)
- [ ] needs to handle large AABBs, like lights (https://github.com/space-wizards/RobustToolbox/issues/1198)
- [ ] client PVS entity cache using mapids (https://github.com/space-wizards/RobustToolbox/issues/1592)
- [ ] no NANs (https://github.com/space-wizards/RobustToolbox/issues/1210)

Network usage is currently higher, ~10KB/s up from ~3KB/s while moving.

*30TPS server with 1 Client running around the station for 60 seconds.*
*Napkin Math: 2.8s/60s = 4.6% of cpu time, US-WEST has 24 cores, ~5x24 players for 25% total time in Grafana*
image

Misc Changes:
* Moved visibility mask from client session to eye.
* Network debug tool improvements.

PR #1845, opened 2021-07-02
PVS is now aware of entity anchoring.

Acruid:
PVS is now aware of entity anchoring. This allows large PVS perf improvements.

Sick Gains Brah
*30TPS server with 1 Client running around the station for 60 seconds, opening hallway doors.*

Here is the old profiler performance for comparison. https://github.com/space-wizards/RobustToolbox/pull/1629
the old code could fit (15,000 / 1,902 = ~7) clients per thread, new one can fit (15,000 / 344 = ~44) clients per thread with each thread taking 25% of CPU time.
Roughly 44 clients per thread, a decent 8 core CPU in can easily process 300 players. I'm benchmarking this on an old I5-4670, so perf is going to be better on newer server hardware.


Robust.Server/GameStates/EntityViewCulling.cs
```diff
@@ -159,22 +158,89 @@ private HashSet<EntityUid> GetSessionViewers(ICommonSession session)
...
+                        // for each grid, calculate the chunks that are inside the box
+                        foreach (var chunk in grid.GetMapChunks().Values)
+                        {
+                            // for each chunk, check dirty
+                            if (chunk.LastAnchoredModifiedTick < fromTick)
+                                continue;
+
+                            // for each chunk, check actually intersects view
+                            if (!chunk.CalcWorldBounds().Intersects(in viewBox))
+                                continue;
```
metalgearsloth: Would be good to potentially just use their origins instead in the future as this has to recalculate the grid's worldposition every time.

Robust.Server/GameStates/EntityViewCulling.cs
```diff
@@ -235,7 +310,7 @@ private List<EntityState> GenerateEntityStates(ICommonSession session, GameTick
 
                     // skip sending anchored entities (walls)
                     var xform = _compMan.GetComponent<ITransformComponent>(entityUid);
-                    if (xform.Anchored && _entMan.GetEntity(entityUid).LastModifiedTick <= lastMapUpdate)
+                    if (xform.Anchored && xform.Owner.LastModifiedTick <= lastMapUpdate)
```
metalgearsloth: Should this be Anchored || lastmodified now? As the anchored state is already added in the chunk part right?

PJB3005 (approved):

Acruid:
This is the network update portion of https://github.com/space-wizards/RobustToolbox/issues/1719.

PJB3005:
Any response on the review comments by metalgearsloth @Acruid?

Robust.Server/GameStates/EntityViewCulling.cs
Acruid: After looking at this further, there is actually a nasty edge case in this whole section of code, where a mover in PVS is parented to an anchored ent outside PVS on a non-visible chunk. All of this code needs to be redesigned.

metalgearsloth:
You still working on this @Acruid ?

PJB3005 (approved):

PR #1849, opened 2021-07-06
Make entitylookup grid-relative

metalgearsloth:
As the scope of https://github.com/space-wizards/RobustToolbox/pull/1848 is now a cruise missile and includes shuttles this is the last blocker to having grids move performantly that I was stalling until I was ready.

Client perf has dropped marginally but we can look at fixing it when entity anchoring is stable.

### DIFF
--- a/Robust.Client/GameObjects/Components/RenderingTreeComponent.cs
+++ b/Robust.Client/GameObjects/Components/RenderingTreeComponent.cs
@@ -15,29 +15,43 @@ namespace Robust.Client.GameObjects
         private static Box2 SpriteAabbFunc(in SpriteComponent value)
         {
             var worldPos = value.Owner.Transform.WorldPosition;
-            return new Box2(worldPos, worldPos);
+            var tree = RenderingTreeSystem.GetRenderTree(value.Owner);
+
+            var pos = worldPos - tree?.Owner.Transform.WorldPosition ?? Vector2.Zero;
+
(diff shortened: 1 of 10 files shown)


metalgearsloth:
Will push my latest changes that work with strafing shuttles when I get a chance.

metalgearsloth:
Good for review. Probably doesn't work for rotated shuttles (grid fixtures need more work first before playtesting it is easy) but works fine for strafing ones.

Robust.Client/GameObjects/Components/RenderingTreeComponent.cs
```diff
@@ -15,29 +15,43 @@ public sealed class RenderingTreeComponent : Component
         private static Box2 SpriteAabbFunc(in SpriteComponent value)
         {
             var worldPos = value.Owner.Transform.WorldPosition;
-            return new Box2(worldPos, worldPos);
+            var tree = RenderingTreeSystem.GetRenderTree(value.Owner);
+
+            var pos = worldPos - tree?.Owner.Transform.WorldPosition ?? Vector2.Zero;
+
+            return new Box2(pos, pos);
         }
```
metalgearsloth, line 23: I'm not a huge fan of these and I might just end up making the query approx instead so it doesn't do it, depending how it profiles.

gradientvera (approved):
looks good to me (<- clueless)

PR #1861, opened 2021-07-10
some misc serv3 fixes

PaulRitter:
adds a typeserializer
adds null parser to read
adds another overload for deserializationresult.value


gradientvera (approved):
LGTM

PR #1862, opened 2021-07-11
Deprecate PhysShapeGrid

metalgearsloth:
This is a part of the shuttle PR and fixes grid collisions not working.

In the interest of cleanliness it seems more reasonable that grids just use the existing poly shapes rather than having their own shape. This will also make the eventual "grid matches tiles" code much easier to do IMO.

Content PR: https://github.com/space-wizards/space-station-14/pull/4264


Robust.Shared/Physics/Collision/Shapes/PolygonShape.cs
```diff
@@ -157,6 +199,15 @@ public bool Equals(IPhysShape? other)
...
         }
 
+        public bool Intersects(Box2 worldAABB, Vector2 worldPos, Angle worldRot)
+        {
+            var aabb = CalculateLocalBounds(worldRot).Translated(worldPos);
+            if (!worldAABB.Intersects(aabb)) return false;
+
+            // TODO
+            return true;
+        }
```
metalgearsloth, line 209: This doesn't actually change any behavior for the moment but will need doing for accurate hitboxes in the future.

Robust.Server/Physics/GridFixtureSystem.cs
```diff
@@ -0,0 +1,126 @@
...
+    internal sealed class GridFixtureSystem : EntitySystem
+    {
+        [Dependency] private readonly IMapManager _mapManager = default!;
+
+        private SharedBroadPhaseSystem _broadphase = default!;
+
+        // Is delaying fixture updates a good idea? IDEK. We definitely can't do them on every tile changed
+        // because if someone changes 50 tiles that will kill perf. We could probably just run it every Update
+        // (and at specific times due to race condition stuff).
+        private float _cooldown;
```
metalgearsloth: I'm not sure on this tbh and want a 2nd opinion.

Currently re-generating the collision isn't that expensive but eventually we'll need to work out exteriors etc etc.

metalgearsloth:
Tests passing locally so should be good to review.

gradientvera (approved):
Looks good to me, although I'm not the most knowledgeable person on physics

PR #1866, opened 2021-07-15
Remove unused IEntityManager parameter from EntityCoordinates.FromMap()

Visne:


gradientvera:
Did you mean EntityCoordinates?

Visne:
> 
> 
> Did you mean EntityCoordinates?

Yes, I scrolled to the top of the file without looking.

PJB3005 (changes requested):
Make a new overload instead and obsolete the old one so we don't have to do a submodule update dance.

Visne:
> 
> 
> Make a new overload instead and obsolete the old one so we don't have to do a submodule update dance.

Done.

PJB3005 (approved):

PR #1867, opened 2021-07-16
Replace most VBox/HBoxContainers with BoxContainers

Visne:


Robust.Client/Console/Commands/Debug.cs
```diff
@@ -536,7 +537,10 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
             var scroll = new ScrollContainer();
             tabContainer.AddChild(scroll);
             //scroll.SetAnchorAndMarginPreset(Control.LayoutPreset.Wide);
-            var vBox = new VBoxContainer();
+            var vBox = new BoxContainer
+            {
+            	Orientation = LayoutOrientation.Vertical
```
PJB3005: Your IDE is misconfigured and inserting tabs here instead of spaces. Please fix it.

This goes for many other lines in the PR diff too.

PJB3005 (changes requested):

Robust.Client/Console/Commands/Debug.cs
Visne: Fixed. I replaced with `\t` instead of four spaces.

Visne:
I'm not sure what to do with that failing testt, apart from that this can be reviewed again.

PJB3005:
Failing test isn't your fault.

PR #1871, opened 2021-07-18
Fix IMapGrid WorldToLocal + LocalToWorld

metalgearsloth:
I noticed this from the F3 menu on my rotated grids branch as it was the reason context menus don't work on rotated grids (amongst other things). I talked to Vera and we agreed that a -10,0 EntityCoordinate should remain -10,0 even if the grid is flipped 180 degrees.

Hence we need to account for the grid's rotation when returning WorldToLocal + LocalToWorld points.


gradientvera (approved):
LGTM, and the tests failing doesn't seem to be your fault

PR #1872, opened 2021-07-19
Change KinematicControllerCollision to use WorldNormal instead

metalgearsloth:
Content cares about the WorldNormal and not the LocalNormal so this fixes pushing bugs.


Robust.Shared/Physics/Dynamics/ContactManager.cs
```diff
@@ -434,26 +435,29 @@ internal void Collide()
 
         public void PreSolve(float frameTime)
         {
+            Span<Vector2> points = stackalloc Vector2[2];
+
```
metalgearsloth, line 439: Only done because InitializeWorldManifold needs it even though we don't care about it.

PR #1892, opened 2021-07-25
Fix containers with entities that do not exist yet

ShadowCommander:
Sometimes an entity contained in a container does not exist on the client yet, which causes a crash. This fixes that by adding an EntityUid list to the container for expected entities.

Related:
space-wizards/space-station-14#4354
space-wizards/space-station-14#1659
This is a partial fix for those two errors. It'll prevent the error from crashing the game, but it doesn't get the missing entity from the server, so I'm not sure if it'll actually fix these issues. Hopefully the entity is just arriving late, but it could be that the entity is outside PVS and thus not sent at all.

Reworking containers to only use EntityUids is going to take quite a bit more work, so I'm putting this fix out since there have been multiple issues reported about this crashing the game.

Test: space-wizards/space-station-14#4371


ShadowCommander:
Ready for review

mirrorcult (approved):

Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
```diff
@@ -0,0 +1,236 @@
...
+                        unexpected ??= new List<EntityUid>();
+                        unexpected.Add(entityUid);
+                    }
+                }
+
+                if (unexpected != null)
+                {
+                    foreach (var entityUid in unexpected)
+                        RemoveExpectedEntity(entityUid);
+                }
```
metalgearsloth: Should probably log unexpected entities tbh.

ShadowCommander: Haha, the variable `unexpected` was for entities that were part of the `ExpectedEntities` list but were removed on the server before the client ever got the entity. I'll rename the variable so it's clearer.

PR #1896, opened 2021-07-25
Don't use DeferMoveEvent if the position is NaN

### LINKED ISSUE #1893: Anchoring doesn't raise MoveEvents
Currently this means the broadphase between server and client desyncs. @Acruid do you have a recommendation on this? Should stuff just subscribe to the anchoring event or should anchoring also raise the MoveEvent?

metalgearsloth:
The issue is that currently moving around means any non-anchored entities have their positions updated to / from NaN. As you can imagine this is a performance nightmare for anything subscribing to it, especially considering it leads to the broadphase getting desynced for physics when anchoring (currently) OR broadphase performance tanking the client due to contacts never stopping their overlap (at least without adding specific NaN handlers everywhere for this which is FUN).

Realistically we need one of the alternatives Acruid has laid out because I can't foresee a sane solution to keep the broadphase up to date without adding hacks for it.

Fixes: https://github.com/space-wizards/RobustToolbox/issues/1893

I'd suggest not merging without acruid / PJB's approval.

@Acruid

### DIFF
--- a/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -734,7 +734,7 @@ namespace Robust.Shared.GameObjects
 
                 var newParentId = newState.ParentID;
                 var rebuildMatrices = false;
-                if (Parent?.Owner?.Uid != newParentId)
+                if (Parent?.Owner.Uid != newParentId)
                 {
                     if (newParentId != _parent)
                     {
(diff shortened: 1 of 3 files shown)


metalgearsloth:
image

See if Acruid comes back with anything.

dekakenedi:
Pada tanggal Min, 25 Jul 2021 22.40, metalgearsloth <
***@***.***> menulis:

dekakenedi:
Pada tanggal Min, 25 Jul 2021 22.52, dekakenedi ***@***.***>
menulis:

PR #1897, opened 2021-07-26
Multi-thread physics contacts

metalgearsloth:
I'd also recommend any reviewer suss out ICollisionManager but it should be thread-safe (assuming GetCvar is).

I also removed the start + end collision lists as I can just use the contact array now instead as the update returns a byte instead of updating the list directly.

I haven't done proper benchmarking but it was roughly a 2x speedup on the pyramid testbed (it drops my framerate from 300 to only 150 now and I profiled with DotTrace). With the transform caching PR + if we could SIMD the Transform.Mul / Transform.MulT methods Collide() would be BTFO.

Fixes https://github.com/space-wizards/RobustToolbox/pull/1897

### DIFF
--- a/Robust.Shared/CVars.cs
+++ b/Robust.Shared/CVars.cs
@@ -353,7 +353,7 @@ namespace Robust.Shared
 
         // - Contacts
         public static readonly CVarDef<int> ContactMultithreadThreshold =
-            CVarDef.Create("physics.contact_multithread_threshold", 16);
+            CVarDef.Create("physics.contact_multithread_threshold", 32);
 
         public static readonly CVarDef<int> ContactMinimumThreads =
             CVarDef.Create("physics.contact_minimum_threads", 2);
--- a/Robust.Shared/CVars.cs
+++ b/Robust.Shared/CVars.cs
@@ -351,6 +351,13 @@ namespace Robust.Shared
          * PHYSICS
          */
 
+        // - Contacts
+        public static readonly CVarDef<int> ContactMultithreadThreshold =
+            CVarDef.Create("physics.contact_multithread_threshold", 16);
+
+        public static readonly CVarDef<int> ContactMinimumThreads =
+            CVarDef.Create("physics.contact_minimum_threads", 2);
+
         // - Sleep
         public static readonly CVarDef<float> AngularSleepTolerance =
             CVarDef.Create("physics.angsleeptol", 2.0f / 180.0f * MathF.PI);
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -366,10 +366,10 @@ namespace Robust.Shared.GameObjects
             _physicsManager.ClearTransforms();
         }
 
-        internal static (int Batches, int BatchSize) GetBatch(int count, int minimumThreads)
+        internal static (int Batches, int BatchSize) GetBatch(int count, int minimumBatchSize)
         {
             // Deduct 1 for networking thread I guess?
-            var batches = Math.Min((int) MathF.Floor((float) count / minimumThreads), Math.Max(1, Environment.ProcessorCount - 1));
+            var batches = Math.Min((int) MathF.Floor((float) count / minimumBatchSize), Math.Max(1, Environment.ProcessorCount - 1));
             var batchSize = (int) MathF.Ceiling((float) count / batches);
 
             return (batches, batchSize);
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -210,6 +210,7 @@ namespace Robust.Shared.GameObjects
             var map = _maps[eventArgs.Map];
             map.ContactManager.KinematicControllerCollision -= KinematicControllerCollision;
 
+            map.Shutdown();
             _maps.Remove(eventArgs.Map);
             Logger.DebugS("physics", $"Destroyed physics map for {eventArgs.Map}");
         }
@@ -355,6 +356,15 @@ namespace Robust.Shared.GameObjects
                 if (mapId == MapId.Nullspace) continue;
                 map.ProcessQueue();
             }
+        }
+
+        internal static (int Batches, int BatchSize) GetBatch(int count, int minimumThreads)
+        {
+            // Deduct 1 for networking thread I guess?
+            var batches = Math.Min((int) MathF.Floor((float) count / minimumThreads), Math.Max(1, Environment.ProcessorCount - 1));
+            var batchSize = (int) MathF.Ceiling((float) count / batches);
+
+            return (batches, batchSize);
         }
     }
 }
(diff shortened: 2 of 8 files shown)


Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -359,5 +360,14 @@ protected void SimulateWorld(float deltaTime, bool prediction)
 
             _physicsManager.ClearTransforms();
         }
+
+        internal static (int Batches, int BatchSize) GetBatch(int count, int minimumThreads)
+        {
+            // Deduct 1 for networking thread I guess?
```
PJB3005: Network thread should be idle most of the time, not sure this is necessary.

If we wanted to we could make a centralized system for handling parallelization like this to consider thread counts properly, I guess.

Robust.Shared/Physics/Dynamics/ContactManager.cs
```diff
@@ -60,10 +60,10 @@ internal sealed class ContactManager
         /// </summary>
         internal event Action<Fixture, Fixture, float, Vector2>? KinematicControllerCollision;
 
+        private int _contactMultithreadThreshold = 16;
+        private int _contactMinimumThreads = 2;
```
PJB3005: These probably shouldn't be initialized to a default value if they're always gonna get replaced with the config default anyways?

Robust.Shared/CVars.cs
```diff
@@ -351,6 +351,13 @@ protected CVars()
          * PHYSICS
          */
 
+        // - Contacts
+        public static readonly CVarDef<int> ContactMultithreadThreshold =
+            CVarDef.Create("physics.contact_multithread_threshold", 32);
+
+        public static readonly CVarDef<int> ContactMinimumThreads =
+            CVarDef.Create("physics.contact_minimum_threads", 2);
```
PaulRitter, line 359: minimum threads seems to be a bit misleading of a name, give how it will use this amount of threads everytime, not scaling up.
if you want this to be a proper minimum, you'd need a max thread count, as well as an ideal contactperthread count it would shoot for afaict

Robust.Shared/Physics/Dynamics/ContactManager.cs
```diff
@@ -348,43 +375,103 @@ internal void Collide()
...
 
-            _startCollisions.Clear();
-            _endCollisions.Clear();
+            ArrayPool<Contact>.Shared.Return(contacts);
+            ArrayPool<ContactStatus>.Shared.Return(status);
+        }
+
+        private void BuildManifolds(Contact[] contacts, int count, ContactStatus[] status)
+        {
+            if (count > _contactMultithreadThreshold * _contactMinimumThreads)
```
PaulRitter, line 451: why `_contactMultithreadThreshold * _contactMinimumThreads`?
if i understand the meaning of `_contactMultithreadThreshold` correctly, it would be the count threshold of contacts on where multithreading would kick in. shouldn't it be `count > _contactMultithreadThreshold` then?

Robust.Shared/Physics/PhysicsManager.cs
```diff
@@ -26,18 +30,24 @@ public void ClearTransforms()
...
+            if (_transforms.ContainsKey(body)) return false;
 
-            transform = body.GetTransform();
-            _transforms[body] = transform;
+            _transforms[body] = body.GetTransform();
+            return true;
+        }
 
-            return transform;
+        public Transform GetOrCreateTransform(PhysicsComponent body)
```
PaulRitter: nitpick: wouldn't `EnsureTransform` be better?

PaulRitter (approved):
lgtm, just had a few nitpicks/questions

Robust.Shared/Physics/Dynamics/ContactManager.cs
metalgearsloth: fuk I saw it and forgot to update

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -364,5 +365,14 @@ protected void SimulateWorld(float deltaTime, bool prediction)
 
             _physicsManager.ClearTransforms();
         }
+
+        internal static (int Batches, int BatchSize) GetBatch(int count, int minimumBatchSize)
+        {
+            // Deduct 1 for networking thread I guess?
+            var batches = Math.Min((int) MathF.Floor((float) count / minimumBatchSize), Math.Max(1, Environment.ProcessorCount - 1));
```
PaulRitter: just realized: does environment.processorcount return the processor or threadcount, bc threads != processorcount in the case of intel hyperthreading for example

Robust.Shared/CVars.cs
metalgearsloth: Yeah I was just dumb in the initial code, it should be the actual minimum now and add another thread every time we hit another contacts per thread x thread count breakpoint (and once it's past all available threads it just splits it).

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
PaulRitter: a quick google search reveals this returns logical processors, so you should be fine(?)

Robust.Shared/Physics/Dynamics/ContactManager.cs
metalgearsloth: The idea was someone who only has 2 threads can have it start at 64 contacts and someone who has 12 threads won't just be throwing 5 contacts at each thread which would be a waste of overhead (probably) so they'll also only use 2 threads.

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
metalgearsloth: How the heckin do I get thread-count then

PJB3005: A logical processor *is* a thread under hyperthreading. Key here is *logical*.

Robust.Shared/Physics/Dynamics/ContactManager.cs
PaulRitter: but this results in it not jumping to multithreading if you have say, 4 threads selected, while having <128 contacts. imo it should then just try to figure out how many threads to use for the amount of contacts, as this will ensure that multihreading will kick in *very* late when designating alot of threads, leading to confusing behaviour

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -364,5 +365,13 @@ protected void SimulateWorld(float deltaTime, bool prediction)
 
             _physicsManager.ClearTransforms();
         }
+
+        internal static (int Batches, int BatchSize) GetBatch(int count, int minimumBatchSize)
+        {
+            var batches = Math.Min((int) MathF.Floor((float) count / minimumBatchSize), Math.Max(1, Environment.ProcessorCount));
```
PaulRitter, line 371: assuming you change my other criticism, you should take minimumthreads into account here, since there is no consideration of it in this logic

Robust.Shared/Physics/Dynamics/ContactManager.cs
PaulRitter: > someone who has 12 threads won't just be throwing 5 contacts at each thread which would be a waste of overhead (probably) so they'll also only use 2 threads.

wont the logic in getbatches literally curb that? we cant stop people from setting the *minimum* to 12, but oh well

PaulRitter: rethinking this again, while this is a somewhat weird way of ensuring the minimum, i suppose its fine

PR #1901, opened 2021-07-27
Adds SetButtonDisabledRecursive() method to Control

Visne:
Will break content without space-wizards/space-station-14#4322


gradientvera:
Why

PR #1907, opened 2021-07-29
RaiseLocalEvent non-generic overload that checks for GetType()

gradientvera:
Needed for stuff like do after, which stores any arbitrary event to be raised directed to an entity at a later time, without the use of generics.

### DIFF
--- a/Robust.Shared/GameObjects/EntityEventBus.Directed.cs
+++ b/Robust.Shared/GameObjects/EntityEventBus.Directed.cs
@@ -12,7 +12,7 @@ namespace Robust.Shared.GameObjects
         void RaiseLocalEvent<TEvent>(EntityUid uid, TEvent args, bool broadcast = true)
             where TEvent:EntityEventArgs;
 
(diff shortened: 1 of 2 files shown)


Robust.Shared/GameObjects/EntityEventBus.Directed.cs
```diff
@@ -12,6 +12,8 @@ public interface IDirectedEventBus
         void RaiseLocalEvent<TEvent>(EntityUid uid, TEvent args, bool broadcast = true)
             where TEvent:EntityEventArgs;
 
+        public void RaiseLocalEvent(EntityUid uid, EntityEventArgs args, bool broadcast = true);
```
PJB3005:
```suggestion
        void RaiseLocalEvent(EntityUid uid, EntityEventArgs args, bool broadcast = true);
```

Unnecessary access modifiers in interfaces annoy me.

PJB3005 (approved):

PR #1913, opened 2021-07-30
Physics timescale cvar

metalgearsloth:
This is mainly for funsies and will lead to a different simulation because the iteration count isn't adjusted for it (not sure if that's something content should handle manually?)


PJB3005:
Would it make more sense to have a global timescale value that affects *everything* in the simulation?

metalgearsloth:
Yeah you got a point

PR #1917, opened 2021-08-01
Clyde can now render multiple viewports with different maps.

gradientvera:
It defaulted to using the main (current) eye's map, instead of passing the actual eye being rendered around.

### DIFF
--- a/Robust.Server/GameStates/EntityViewCulling.cs
+++ b/Robust.Server/GameStates/EntityViewCulling.cs
@@ -281,7 +281,6 @@ namespace Robust.Server.GameStates
                     visMask = eyeComp.VisibilityMask;
 
                 //Always include the map entity of the eye
-                //TODO: Add Map entity here
                 visibleEnts.Add(_mapManager.GetMapEntityId(eyeComp.Owner.Transform.MapID));
 
                 //Always include viewable ent itself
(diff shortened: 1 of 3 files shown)


Robust.Server/GameStates/EntityViewCulling.cs
```diff
@@ -282,6 +282,7 @@ private HashSet<EntityUid> CalcCurrentViewSet(ICommonSession session)
 
                 //Always include the map entity of the eye
                 //TODO: Add Map entity here
```
RemieRichards: vera pls

gradientvera: my yeast is slow

Robust.Server/GameStates/EntityViewCulling.cs
```diff
@@ -281,7 +281,7 @@ private HashSet<EntityUid> CalcCurrentViewSet(ICommonSession session)
                     visMask = eyeComp.VisibilityMask;
 
                 //Always include the map entity of the eye
-                //TODO: Add Map entity here
+                visibleEnts.Add(_mapManager.GetMapEntityId(eyeComp.Owner.Transform.MapID));
```
Visne: Looks good to me assuming that this is fast enough

Visne (approved):

PR #1927, opened 2021-08-04
Make no rotation work when eye is rotated

Visne:
I'm not sure if there is a better way to do this.


ShadowCommander (approved):

PaulRitter (approved):

PaulRitter:
Would like @PJB3005 to look at this

PR #1928, opened 2021-08-05
Add Friend classes to C# with the help of Analyzers and Attributes.

gradientvera:
So basically, because of ECS every member in components is public and anything can write/invoke them even if they shouldn't, yes? The solution we came up with was to code better comments/xmldocs, and review things properly.

But of course, the dumb coder will still write to/invoke them directly instead of going through the intended entity system... The drunk maintainer will absolutely miss these awful crimes against humanity, and the codebase will burn and suffer as a result. All because there's no encapsulation. Smh.

Because of these awful predictions about the future, I decided to sacrifice myself and learn roslyn analyzers with the goal of avoiding that timeline and save the project from the shitcoders: 
I implemented "friend classes" in C#.

The way this works is very simple. You add a `[Friend(typeof(MyFriend), typeof(MyBestFriend))]` attribute to your component, and now only those two classes will be able to write/invoke members in it. Any class can *read*, but only friends can write/invoke.
This also works with inheritance: If your base class is friends with Foo, and the inheriting class is friends with Bar, the inheriting class is friends with Foo as well, but the base class is not friends with Bar. In short, this should probably force you to think about your relationships between components/things using components..

Shoutouts to Y for giving me the motivation required for coding this.

### DIFF
--- a/Robust.Analyzers/FriendAnalyzer.cs
+++ b/Robust.Analyzers/FriendAnalyzer.cs
@@ -1,0 +1,108 @@
+using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Robust.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class FriendAnalyzer : DiagnosticAnalyzer
+    {
+        const string FriendAttribute = "Robust.Shared.Analyzers.FriendAttribute";
+
+        public const string DiagnosticId = "RA0002";
+
+        private const string Title = "Tried to access friend-only member";
+        private const string MessageFormat = "Tried to access member \"{0}\" in class \"{1}\" which can only be accessed by friend classes";
+        private const string Description = "Make sure to specify the accessing class in the friends attribute.";
+        private const string Category = "Usage";
+
+        [SuppressMessage("ReSharper", "RS2008")]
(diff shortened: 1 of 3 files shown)


Ygg01:
Motivation? I believe it was anguish.

Robust.Analyzers/FriendAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+                    break;
+
+                // Otherwise, do nothing.
+                default:
+                    return;
+            }
+
+            // Get the friend attribute, if it can't be found do nothing...
+            if (context.Compilation.GetTypeByMetadataName(FriendAttribute) is not {} friendAttr)
+                return;
```
PaulRitter: this is somewhat sus, i'd at least throw a warning here?

PaulRitter (approved):
very non-invasive, meaning easily reworkable if we find a better solution down the line.

i like it

Robust.Analyzers/Robust.Analyzers.csproj
```diff
@@ -1,7 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
PaulRitter: are you sure this is supported?

gradientvera: works on my machine, but I can revert it easy

gradientvera: Reverted, just in case!

Robust.Analyzers/FriendAnalyzer.cs
gradientvera: Mmm, technically this should always be fine, I looked at your analyzer and you don't even check if it's null so I'll just remove the check and get it always

PaulRitter (approved):

PR #1929, opened 2021-08-05
Fix outline when eye is rotated

Visne:
The PostShader bounding box calculation does not account for eye rotation, and thus clips the sprite. This change essentially rotates the calculated bounding box by the eye rotation, making it correct again.

### DIFF
--- a/Robust.Shared.Maths/Angle.cs
+++ b/Robust.Shared.Maths/Angle.cs
@@ -103,6 +103,9 @@ namespace Robust.Shared.Maths
         [Pure]
         public Vector2 RotateVec(in Vector2 vec)
         {
+            // No calculation necessery when theta is zero
+            if (Theta == 0) return vec;
+
             var (x, y) = vec;
             var cos = Math.Cos(Theta);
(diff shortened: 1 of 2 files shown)


ShadowCommander (approved):

Robust.Shared.Maths/Angle.cs
```diff
@@ -103,6 +103,9 @@ public Direction GetCardinalDir()
         [Pure]
         public Vector2 RotateVec(in Vector2 vec)
         {
+            // No calculation necessery when theta is zero
+            if (Theta == 0) return vec;
+
```
PaulRitter, line 108: breaking: codebase load drops by 50%

PaulRitter (approved):

Robust.Shared.Maths/Angle.cs
Visne: Yeah tbh I'm not sure how much effect it has but I thought since we're gonna run this a lot of frames and eyes are almost always at 0 degrees, might as well

PR #1931, opened 2021-08-06
Add support for struct directed events, add by-ref directed/broadcast events.

gradientvera:
Requires https://github.com/space-wizards/space-station-14/pull/4478
Also turns transform events into readonly structs.
You can now pass events by ref! And you can now use structs as directed events!
The combination of these two features is bound to bring many good things to RobustToolbox.


Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -918,7 +918,7 @@ public void SetAnchored(bool value)
     ///     Raised whenever an entity moves.
     ///     There is no guarantee it will be raised if they move in worldspace, only when moved relative to their parent.
     /// </summary>
-    public class MoveEvent : HandledEntityEventArgs
+    public readonly struct MoveEvent
```
PaulRitter, line 936: This used to be a handled Event. Intended?

PaulRitter (changes requested):
I'm gonna be honest i dont really like the `object args` signature.

What speaks against having an `IEntityEventArgs` interface that all args have to implement, would work with classes as well as structs? Could then use that interface in the generic methods, as well as the non-generic ones.

This might just be me, but i dont like when signatures are being typed as object since you can pass virtually anything (as far as c# and the linter knows).

PaulRitter:
Besides that, i obviously think struct args rock. Good job :)

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
gradientvera: Yes, intended. we discussed it in discord and concluded that it was accidentally made handled.

gradientvera:
> I'm gonna be honest i dont really like the `object args` signature.

Same signature as in the broadcast eventbus, it's fine.

> What speaks against having an `IEntityEventArgs` interface that all args have to implement, would work with classes as well as structs? Could then use that interface in the generic methods, as well as the non-generic ones.

Interfaces, and boxing are really slow. We don't need an interface for this, at all.

> This might just be me, but i dont like when signatures are being typed as object since you can pass virtually anything (as far as c# and the linter knows).

It's always been this way with broadcast events and we've never had a problem

metalgearsloth:
I said this on discord but putting here for brevity: would like a benchmark of the physics MoveEvents with this change sooooooooo tumbler when

PaulRitter:
> Interfaces, and boxing are really slow. We don't need an interface for this, at all.

While i dont think "we've always done it like this" is a proper argument, i think this is. Consider it sorted

PaulRitter (approved):

Robust.Shared/GameObjects/EntityEventBus.Directed.cs
```diff
@@ -69,7 +69,7 @@ internal void RaiseComponentEvent<TEvent>(IComponent component, TEvent args)
 
         /// <inheritdoc />
         public void RaiseLocalEvent<TEvent>(EntityUid uid, TEvent args, bool broadcast = true)
-            where TEvent : EntityEventArgs
```
PJB3005, line 72: You *only* changed this generic bound, which measn that the `_eventTables.Dispatch` call, which is not generic, will do a boxing allocation for your struct. This makes struct directed events rather useless with this implementation since at best their performance is just worse than ref types.

PJB3005 (changes requested):

Robust.Shared/GameObjects/EntityEventBus.Directed.cs
gradientvera: I think I addressed this now. Would be nice if you could give this another review

gradientvera:
Requesting another review from @PJB3005 as I increased the scope of this PR significantly

PJB3005 (approved):
I am now complicit in the impending `Unsafe` disaster this PR has become so...

gradientvera:
if this ever actually becomes an issue, feel free to throw rotten tomatoes at me

PR #1933, opened 2021-08-07
Fix map deletion not working because of deleted entity referencing map

Visne:


Robust.Shared/GameObjects/Components/CollisionWakeComponent.cs
```diff
@@ -42,7 +42,7 @@ internal void RaiseStateChange()
         protected override void OnRemove()
         {
             base.OnRemove();
-            if (Owner.TryGetComponent(out IPhysBody? body))
+            if (Owner.TryGetComponent(out IPhysBody? body) && Owner.LifeStage != EntityLifeStage.Terminating)
```
metalgearsloth:
```suggestion
            if (Owner.TryGetComponent(out IPhysBody? body) && Owner.LifeStage < EntityLifeStage.Terminating)
```

metalgearsloth (approved):

PR #1935, opened 2021-08-08
Add vertices simplifier

metalgearsloth:
This takes in a set of vector2s and simplifies it (within tolerance). I only ported the RamerDouglasPeucker one from farseer as it was the only that was recursive.

(Farseer's license is already in the repo).

Intend to use it eventually for accurate grid bounds.

Source at https://github.com/Zumorica/FarseerPhysicsEngine/blob/master/FarseerPhysics/Common/PolygonManipulation/SimplifyTools.cs


PJB3005 (approved):
These implementations could probably be optimized with modern SS13 shit like span but that's minor and it otherwise looks good.

PR #1940, opened 2021-08-09
Fix FlagSerializer not working with 1 << 31 and add test

DrSmugleaf:


metalgearsloth (approved):
ship it ship it ship it ship it ship it

PR #1941, opened 2021-08-10
Add support for grid chunk removals

metalgearsloth:
Also allows grids to be removed when they have no more chunks remaining.

Content PR https://github.com/space-wizards/space-station-14/pull/4452


Robust.Server/GameObjects/EntitySystems/MapSystem.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        private void HandleGridEmpty(EntityUid uid, MapGridComponent component, EmptyGridEvent args)
+        {
+            if (!_deleteEmptyGrids ||
+                !EntityManager.TryGetEntity(uid, out var gridEnt) ||
+                gridEnt.LifeStage >= EntityLifeStage.Terminating) return;
+
+            MapManager.DeleteGrid(args.GridId);
+        }
```
PaulRitter: do we just accept that there might be empty grids lingering around when you switch the cvar from false to true? or should we cache empty grids if the cvar is false & work through them in the setgriddeletion function?

metalgearsloth: At least for networking purposes they're technically maintained indefinitely (until the server has 0 players), but I see your point on checking existing grids.

PR #1945, opened 2021-08-12
Optimise TryFindGridAt

metalgearsloth:
Should be O(1) to number of fixtures on a map instead of O(n).


PJB3005:
Why does this cause integration tests to fail?

metalgearsloth:
> 
> 
> Why does this cause integration tests to fail?

The EntitySystem resolve. I've just fixed MapManager's startup ordering so it can have dependencies on it (though there is a bit of a suss resolve on RobustUnitTest).

PR #1954, opened 2021-08-18
Joints refactor

metalgearsloth:
Dusting off this PR. Will need updating after #1953 so do that first.

* Joints moved to their own component reducing physics state + code clutter
* Joints now have their own state classes rather than serializing the whole thing over the network
* PhysicsMap no longer stores joints so JointComponent + JointSystem can work with prediction on its own.

There's still some mess internally but I don't really know a way to make it "clean" so it is what it is for now I guess.

Content PR: https://github.com/space-wizards/space-station-14/pull/4487


metalgearsloth:
Good for review now I guess.

Robust.Client/Debugging/DebugPhysicsSystem.cs
```diff
@@ -346,5 +372,20 @@ private void DrawShape(DrawingHandleWorld worldHandle, Fixture fixture, Transfor
                     return;
             }
         }
+
+        private void DrawJoint(DrawingHandleWorld worldHandle, Joint joint)
+        {
+            if (!_entityManager.TryGetComponent(joint.BodyAUid, out TransformComponent? xf1) ||
+                !_entityManager.TryGetComponent(joint.BodyBUid, out TransformComponent? xf2)) return;
+
+            var color = new Color(0.5f, 0.8f, 0.8f);
```
PaulRitter: could just make this static readonly, no?

Robust.Client/Debugging/DebugPhysicsSystem.cs
```diff
@@ -346,5 +372,20 @@ private void DrawShape(DrawingHandleWorld worldHandle, Fixture fixture, Transfor
...
+
+            switch (joint)
+            {
+                case DistanceJoint distance:
+                    worldHandle.DrawLine(xf1.WorldPosition, xf2.WorldPosition, color);
+                    break;
+            }
```
PaulRitter: why arent we drawing other types of joints?

Robust.Shared/Physics/Dynamics/Joints/DistanceJoint.cs
```diff
@@ -28,14 +28,49 @@
...
+                MinLength = MinLength,
+                MaxLength = MaxLength,
+                Stiffness = Stiffness,
+                LocalAnchorA = LocalAnchorA,
+                LocalAnchorB = LocalAnchorB
+            };
+
+            var configManager = IoCManager.Resolve<IConfigurationManager>();
+            joint.LinearSlop = configManager.GetCVar(CVars.LinearSlop);
+            joint.WarmStarting = configManager.GetCVar(CVars.WarmStarting);
```
PaulRitter, line 67: these wont get updated when the cvar gets updated unless i've missed something. do we care? i'd say yes

Robust.Shared/Physics/Dynamics/Joints/FrictionJoint.cs
```diff
@@ -298,11 +311,7 @@ public bool Equals(FrictionJoint? other)
         {
             if (ReferenceEquals(null, other)) return false;
             if (ReferenceEquals(this, other)) return true;
-            return base.Equals(other) &&
-                   MathHelper.CloseToPercent(MaxForce, other.MaxForce) &&
-                   MathHelper.CloseToPercent(MaxTorque, other.MaxTorque) &&
-                   LocalAnchorA.EqualsApprox(other.LocalAnchorA) &&
-                   LocalAnchorB.EqualsApprox(other.LocalAnchorB);
+            return base.Equals(other);
```
PaulRitter: afaict you dont compare force & torque anymore, as well as the anchor points. is this intentional, and why?

Robust.Shared/Physics/Dynamics/Joints/DistanceJoint.cs
```diff
@@ -574,20 +523,14 @@ internal override bool SolvePositionConstraints(SolverData data)
...
             if (ReferenceEquals(null, other)) return false;
             if (ReferenceEquals(this, other)) return true;
-            return LocalAnchorA.EqualsApprox(other.LocalAnchorA) &&
-                   LocalAnchorB.EqualsApprox(other.LocalAnchorB) &&
-                   MathHelper.CloseToPercent(Length, other.Length) &&
-                   MathHelper.CloseToPercent(Stiffness, other.Stiffness) &&
-                   MathHelper.CloseToPercent(Damping, other.Damping) &&
-                   MathHelper.CloseToPercent(MaxLength, other.MaxLength) &&
-                   MathHelper.CloseToPercent(MinLength, other.MinLength);
+            return base.Equals(other);
```
PaulRitter: afaict this only compares both uids now. why not compare the rest of the fields?

Robust.Shared/Physics/Dynamics/Joints/RevoluteJoint.cs
```diff
@@ -22,35 +22,69 @@
...
+                LocalAnchorA = LocalAnchorA,
+                LocalAnchorB = LocalAnchorB
+            };
+
+            return joint;
+        }
 
+    }
+
+    public class RevoluteJoint : Joint
```
PaulRitter: how come no iequatable on this?

Robust.Shared/Physics/Dynamics/SharedPhysicsMapComponent.cs
```diff
@@ -91,9 +91,9 @@ public Vector2 Gravity
         private List<PhysicsComponent> _awakeBodyList = new();
 
         /// <summary>
-        ///     Get all the joints on this map
+        /// Temporary joint storage during solving
```
PaulRitter, line 93: cant comment on non-diff entries so: remove `_entityManager` in L39 of this file, its unused now

Robust.Shared/Physics/JointComponent.cs
```diff
@@ -0,0 +1,39 @@
...
+using Robust.Shared.Physics.Dynamics.Joints;
+using Robust.Shared.Players;
+using Robust.Shared.Serialization;
+using Robust.Shared.ViewVariables;
+
+namespace Robust.Shared.Physics
+{
+    [RegisterComponent]
+    [NetworkedComponent]
+    public sealed class JointComponent : Component
```
PaulRitter, line 18: would p be good to restrict access to jointsystem with a friend attribute

PaulRitter (changes requested):
nice work proud of ya

Robust.Client/Debugging/DebugPhysicsSystem.cs
metalgearsloth: None of the other types I've actually implemented have drawing stuff on box2d https://github.com/erincatto/box2d/blob/95f74a46c077825478f989df6a81003ca538c2c2/src/dynamics/b2_joint.cpp#L255

Just distance, mouse, and pulley. In the future I'll think of a good way to visualize it.  I mean it has the default but ehh.

Robust.Shared/Physics/Dynamics/Joints/DistanceJoint.cs
metalgearsloth: SharedJointSystem subscribes and updates all joints; realistically these would only ever be touched in debugging (and even then I've never touched it).

Robust.Shared/Physics/Dynamics/Joints/RevoluteJoint.cs
metalgearsloth: Simple image

Robust.Shared/Physics/Dynamics/Joints/DistanceJoint.cs
metalgearsloth: (Realistically LinearSlop needs to be a constant as well but baby steps)

Also IMO it should be passed in via the solverdata which box2d doesn't do for reasons.

PR #1958, opened 2021-08-19
Use arrays for polygon vertices

### LINKED ISSUE #1955: Make PolygonShape use arrays internally
This was a farseer holdover.

I think also make SetVertices a method and add an optional parameter so you can specify if the vertices are already in winding order so you can avoid doing the GiftWrap calc, rather than just setting via the property.

metalgearsloth:
Fixes https://github.com/space-wizards/RobustToolbox/issues/1955

Also pointed out windoors had incorrect bounds which is nice.


PJB3005:
@metalgearsloth tests fail.

metalgearsloth:
Forgot to mention it needed a content PR, oops

PJB3005:
@metalgearsloth tests still failing.

PR #1959, opened 2021-08-19
PVS Anchoring take 2

metalgearsloth:
Okay so a spiel on this:
Previously, the last PR assumed that all chunks were sent to the player whenever they were created / on roundstart. In actuality they get dumped on the client when they initially join the server and never again. The problem is when the round restarts they never received any chunks that weren't already in range of them (because the code assumed they had already received it).

My solution was to just track whenever a client saw a chunk. Given there's not that many chunks (rn there's <25 for Saltern) I just made it so we track when each was sent for each player so we don't need to continue dumping entities onto the player (like PVS currently does) that they've already seen.

The downside right now is that there's a noticeable framedrop whenever a new chunk comes in range on the client (from spawning all of the new entities in a 16 x 16 area). There's minor framedrops from moving around on master anyway as the client needs to spawn all the new entities. Given this is likely only going to be an issue for the next 6-12 months (until entity spawning is optimised at some point) and doesn't really affect development I'm personally fine with it.

Also, 90% of the code is the same as the original PR, it's just caching what chunks we've seen before is added (and also I had to fix the anchoring crash which was a direct push).


metalgearsloth:
Good for review

Robust.Server/GameStates/EntityViewCulling.cs
```diff
@@ -169,23 +171,115 @@ private HashSet<EntityUid> GetSessionViewers(ICommonSession session)
...
+                foreach (var eyeEuid in viewers)
+                {
+                    var (viewBox, mapId) = CalcViewBounds(in eyeEuid);
+
+                    uint visMask = 0;
+                    if (_compMan.TryGetComponent<EyeComponent>(eyeEuid, out var eyeComp))
+                        visMask = eyeComp.VisibilityMask;
+
+                    //Always include the map entity of the eye
+                    //TODO: Add Map entity here
```
ShadowCommander: Does this need to be implemented?

Robust.Server/GameStates/EntityViewCulling.cs
```diff
@@ -347,20 +405,49 @@ private bool RecursiveAdd(TransformComponent xform, HashSet<EntityUid> visSet, u
...
 
+        /// <summary>
+        /// If we recursively get an anchored entity need to ensure the entire chunk is included (as it may be out of view).
+        /// </summary>
+        private void EnsureAnchoredChunk(TransformComponent xform, Dictionary<GridId, HashSet<IMapChunkInternal>> includedChunks)
+        {
+            // If we recursively get an anchored entity need to ensure the entire chunk is included (as it may be out of view).
+            if (!xform.Anchored) return;
+
+            // This is slow as but entities being parented to anchored ones is hopefully rare so shouldn't be hit too much.
```
ShadowCommander:
```suggestion
            // This is slow but entities being parented to anchored ones is hopefully rare so shouldn't be hit too much.
```

Robust.Shared/GameObjects/Entity.cs
```diff
@@ -322,6 +323,9 @@ public IEnumerable<T> GetAllComponents<T>()
         public void Dirty()
         {
             LastModifiedTick = EntityManager.CurrentTick;
+
+            if (LifeStage >= EntityLifeStage.Initialized && Transform.Anchored)
+                EntityManager.ComponentManager.GetComponent<IMapGridComponent>(Transform.ParentUid).AnchoredEntityDirty(Transform);
```
ShadowCommander, line 328: Are anchored entities always anchored to a Grid?

Robust.Shared/Map/MapChunk.cs
```diff
@@ -250,6 +254,21 @@ public void RemoveFromSnapGridCell(ushort xCell, ushort yCell, EntityUid euid)
...
             ref var cell = ref _snapGrid[xCell, yCell];
             cell.Center?.Remove(euid);
+            LastAnchoredModifiedTick = _grid.CurTick;
+        }
+
+        public IEnumerable<EntityUid> GetAllAnchoredEnts()
+        {
+            foreach (var cell in _snapGrid)
+            {
+                if(cell.Center is null)
```
ShadowCommander:
```suggestion
                if (cell.Center is null)
```

ShadowCommander (approved):

Robust.Shared/GameObjects/Entity.cs
gradientvera: yup, can't have anchored entities not on a grid, they'll immediately unanchor themselves

Robust.Server/GameStates/EntityViewCulling.cs
metalgearsloth: My culture

PR #1961, opened 2021-08-19
Fix invoking VisualizerComponent twice

Ygg01:
After investigating the Visualizer component being double called, I've came to conclusion this part is causing problems.

What happens is the following. For example assume AppearanceVisualizer is changed once on `(LastModifiedTick= 100)`.

When [`CalculateEntityStates.cs`](https://github.com/space-wizards/RobustToolbox/blob/5594bd7203ac3dc08aa23f9cbeaf72c19437f0ea/Robust.Server/GameStates/EntityViewCulling.cs#L158) will eventually call [`ServerGameStateManager.cs:L284`](https://github.com/space-wizards/RobustToolbox/blob/5594bd7203ac3dc08aa23f9cbeaf72c19437f0ea/Robust.Server/GameStates/ServerGameStateManager.cs#L284).

It will be invoked twice one from `(fromTick = 99, toTick =100)` and `(fromTick = 100, toTick = 101)`.
Because  `(LastModifiedTick= 100)` will both be greater than or equal to 99 and 100 the message for handling will be twice invoked. 

I changed all `>=` to `>` for consistency and I tried it in-game and it did fix the visualizer errors I got. I didn't notice any issues, but I could be breaking some underlying assumptions.

PLEASE REVIEW extremely carefully.

### DIFF
--- a/Robust.Server/GameStates/ServerGameStateManager.cs
+++ b/Robust.Server/GameStates/ServerGameStateManager.cs
@@ -271,21 +271,21 @@ namespace Robust.Server.GameStates
 
                 DebugTools.Assert(component.LastModifiedTick >= component.CreationTick);
 
-                if (component.CreationTick != GameTick.Zero && component.CreationTick >= fromTick && !component.Deleted)
+                if (component.CreationTick != GameTick.Zero && component.CreationTick > fromTick && !component.Deleted)
                 {
                     ComponentState? state = null;
-                    if (component.NetSyncEnabled && component.LastModifiedTick != GameTick.Zero && component.LastModifiedTick >= fromTick)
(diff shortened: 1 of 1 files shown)


Ygg01:
@Acruid gonna need your opinion on this one. I'm not sure why this is happening.

metalgearsloth:
Like I said on discord I think this is messing with game states for something that is an appearancecomponent problem.

AppearanceComponent always assumes new data is dirty and hence always queues an update which can break things so why don't you just diff the state in HandleComponentState?

PJB3005 (changes requested):
A single component state can always be applied on multiple ticks, this is by design.

This is not the correct way to fix this.

PR #1962, opened 2021-08-21
Adds new helpful filter methods.

gradientvera:
- AddWhereAttachedEntity and RemoveWhereAttachedEntity methods
- AddInGrid and AddInMap methods
- BroadcastGrid and BroadcastMap static methods

### DIFF
--- a/Robust.Shared/Player/Filter.cs
+++ b/Robust.Shared/Player/Filter.cs
@@ -1,13 +1,11 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using JetBrains.Annotations;
 using Robust.Shared.Configuration;
 using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
 using Robust.Shared.Map;
 using Robust.Shared.Players;
-using Robust.Shared.Utility;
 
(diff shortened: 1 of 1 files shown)


Robust.Shared/Player/Filter.cs
```diff
@@ -123,6 +121,31 @@ public Filter AddWhere(Predicate<ICommonSession> predicate)
             return this;
         }
 
+        /// <summary>
+        ///     Add all players whose attached entity match a predicate.
+        ///     Doesn't consider players without an attached entity.
+        /// </summary>
+        public Filter AddWhereAttachedEntity(Predicate<IEntity> predicate)
+        {
+            return AddWhere(session => session.AttachedEntity is { } entity && predicate(entity));
```
PaulRitter, line 130: `session.AttachedEntity is { } entity`
a pr from vera isnt a pr from vera without this pattern.

prove me wrong, you cant

PaulRitter (approved):

Robust.Shared/Player/Filter.cs
gradientvera: it's a good pattern, of course I'm gonna use it

PR #1978, opened 2021-08-25
Change Button Pressed to work with non-focus keybinds

ShadowCommander:
This PR solves two somewhat interconnected things.
First is that when `EnableAllKeybinds` is enabled, `UIClick` fails to trigger `OnPressed` because `_attemptingPress` gets set to false by `Use`, which stops all KeyBinds triggered afterward.
Second is for non-CanFocus keybinds not grabbing button focus when `EnableAllKeybinds` is set. This is a problem when KeyUps don't trigger when the button is not focused, leaving `_attemptingPress` set to true or after changing `_attemptingPress` to an int, not decrimenting the counter.

So I've changed `_attemptingPress` to an int that counts how many KeyBinds are currently KeyDowned. It resets to 0 when `ControlFocused` is lost.
I've also made `ControlFocused` public settable to allow controls to set which control is focused, which allows the non-CanFocus keybinds to have hover highlighting and receive KeyUp events.

I'm not sure if this is the best way to go about this, but I got it to work this way. One of alternative solutions for the first problem would be to make BaseButton take in a list of BoundKeyFunctions that it handles instead of having `EnableAllKeybinds`.


ShadowCommander:
There are some problems with this solution like all hotkeys now make the `EnableAllKeybinds` buttons show the selection color, even if they don't do anything, but to fix that would require providing the button the list of KeyBinds it can handle or something similar.

Issue #889, opened 2019-11-14
Network Level Gamestate Compression


PJB3005:
We currently use a *ton* of bandwidth to do netcode.

NetSerializer is not exactly bandwidth efficient as far as I can tell. It also does not provide any way to do delta updates efficiently (for example, in a component state, using bitflags and stuff to only send the new position).

It would be best if we didn't simply run the game state delta through netserializer, and instead used lidgren's NetOutgoingMessage/NetIncomingMessage API to reduce state delta size.

Acruid:
Looking closer at lidgren's API, you can directly modify the backing array behind the bit reader/writer. This allows us to do things like compress the entire packet payload after crafting the packet with the Read/Write API. The Reader/Writer itself is good enough (I think), no need to create our own BitStream class.

Another major optimization would be a string table, so that we can intern prototype names, and only have to send the string ID over the network. This can be implemented independent of anything else, any time.

PJB3005:
Does running regular compression algorithms over a bitstream like that even help much? Ideally we would be using Lidgren's bit packing to the fullest extent.

Acruid:
`The delta compressed bit packed data is not a stream of fixed length words. The delta compression introduces single bits and the bit packed variables can use an arbitrary number of bits. As a result regular entropy encoding is difficult because entropy encoders typically assume a word length of a fixed number of bits.`

"Regular" compression algorithms (Deflate) won't work well here.

`The delta compression writes out a single zero bit for any variable that did not change. Because only few variables change over time the delta compression introduces a lot of sequences of zero bits. For this reason 3-bit word length zero based run-length compression is used. The compressor reads 3 bits at a time and if the bits are not all zero the same bits are written out without changes. If the 3 bits are all zero the compressor keeps reading words of 3 bits at a time until they are no longer all zero. The compressor then writes out 3 zero bits followed by 3 bits that represent the number of times 3 zero bits were read in succession. Using a 3-bit word length results in a maximum compression ratio of 4:1. A different number of bits than 3 could be used for the word length but for the DOOM III entities using 3 bits results in the best compression ratios in practice.`
http://mrelusive.com/publications/papers/The-DOOM-III-Network-Architecture.pdf, Section 4.3

So the theoretical ratio is 4:1, they do not mention what their actual average ratio is. After delta compression is implemented, someone should manually analyze our packet payloads and see if there are large groups of zeros like they mention.

moonheart08:
Related: #973

[...]

PR #1981, opened 2021-08-25
Adds deflate compression to MsgState.

gradientvera:
Related #889
GameStates above a certain size threshold (256 bytes right now) will be compressed using the deflate algorithm.
I've tested this extensively, and it seems to produce very good results. Here is an example:
image
Initial gamestate being compressed from 1.38 MB down to 73.22 KB

Paging @PJB3005 for review

### DIFF
--- a/Robust.Shared/Network/Messages/MsgState.cs
+++ b/Robust.Shared/Network/Messages/MsgState.cs
@@ -63,7 +63,6 @@ namespace Robust.Shared.Network.Messages
             var serializer = IoCManager.Resolve<IRobustSerializer>();
             MemoryStream finalStream;
             var stateStream = new MemoryStream();
-            DebugTools.Assert(stateStream.Length <= Int32.MaxValue);
             serializer.SerializeDirect(stateStream, State);
             buffer.WriteVariableInt32((int) stateStream.Length);
 
--- a/Robust.Shared/Network/Messages/MsgState.cs
+++ b/Robust.Shared/Network/Messages/MsgState.cs
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.IO;
+using System.IO.Compression;
 using Lidgren.Network;
 using Robust.Shared.GameStates;
 using Robust.Shared.IoC;
+using Robust.Shared.Log;
 using Robust.Shared.Serialization;
 using Robust.Shared.Utility;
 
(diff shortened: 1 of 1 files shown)


Robust.Shared/Network/Messages/MsgState.cs
```diff
@@ -26,27 +31,66 @@ public class MsgState : NetMessage
...
+            var uncompressedLength = buffer.ReadVariableInt32();
+            var compressedLength = buffer.ReadVariableInt32();
+            MemoryStream finalStream;
+
+            // State is compressed.
+            if (compressedLength > 0)
+            {
+                var stream = buffer.ReadAlignedMemory(compressedLength);
+                using var decompressStream = new DeflateStream(stream, CompressionMode.Decompress);
+                var decompressedStream = new MemoryStream(decompressStream.CopyToArray(), false);
```
PJB3005: When you do `.CopyToArray()`, the implementation makes a `MemoryStream` which has to gradually grow its internal buffer. You should probably have `decompressedStream` be initialized to an empty stream (of the correct size) and then decompress into it with `decompressStream`. Then just `.Position = 0` to reset it and you can pass it down to the deserializer.

Robust.Shared/Network/Messages/MsgState.cs
```diff
@@ -26,27 +31,66 @@ public class MsgState : NetMessage
...
+            State.PayloadSize = uncompressedLength;
         }
 
         public override void WriteToBuffer(NetOutgoingMessage buffer)
         {
             var serializer = IoCManager.Resolve<IRobustSerializer>();
-            using (var stateStream = new MemoryStream())
+            MemoryStream finalStream;
+            var stateStream = new MemoryStream();
+            DebugTools.Assert(stateStream.Length <= Int32.MaxValue);
```
PJB3005: I would be pretty concerned aswell if a stream I just created and should thus be empty had a length that big, yeah.

Robust.Shared/Network/Messages/MsgState.cs
gradientvera: Should be fixed now, I think.

Robust.Shared/Network/Messages/MsgState.cs
gradientvera: Oops

PR #1983, opened 2021-08-26
EntityLookup flags for PVS optimisations

metalgearsloth:
Given anchored entities are handled by chunk for PVS we no longer need to store them in EntityLookup's DynamicTrees leading to a nice memory and performance savings.

I also added approximate to the flag as it was a bit weird to have a bool and then flags for a bunch of other stuff. Eventually I'll add grids to it too (given most of the time callers don't want it) and (potentially) containermanagers.

Content PR: https://github.com/space-wizards/space-station-14/pull/4531

### DIFF
--- a/Robust.Shared/GameObjects/Systems/EntityLookup.cs
+++ b/Robust.Shared/GameObjects/Systems/EntityLookup.cs
@@ -391,9 +391,15 @@ namespace Robust.Shared.GameObjects
                 }, offsetBox, (flags & LookupFlags.Approximate) != 0x0);
             }
 
-            if ((flags & LookupFlags.IncludeAnchored) != 0x0)
+            if ((flags & LookupFlags.IncludeAnchored) != 0x0 &&
+                _mapManager.TryFindGridAt(mapId, position, out var grid) &&
+                grid.TryGetTileRef(position, out var tile))
             {
-
+                foreach (var ent in grid.GetAnchoredEntities(tile.GridIndices))
+                {
+                    if (!_entityManager.TryGetEntity(ent, out var entity)) continue;
+                    state.list.Add(entity);
+                }
             }
 
             return list;
--- a/Robust.Shared/GameObjects/Systems/EntityLookup.cs
+++ b/Robust.Shared/GameObjects/Systems/EntityLookup.cs
@@ -296,13 +296,10 @@ namespace Robust.Shared.GameObjects
 
                 lookup.Tree.QueryAabb(ref found, (ref bool found, in IEntity ent) =>
                 {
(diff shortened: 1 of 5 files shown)


Robust.Shared/GameObjects/Systems/EntityLookup.cs
```diff
@@ -262,26 +303,46 @@ public bool AnyEntitiesIntersecting(MapId mapId, Box2 box, bool approximate = fa
...
-                }, offsetBox, approximate);
+                }, offsetBox, (flags & LookupFlags.Approximate) != 0x0);
+            }
+
+            if (!found)
+            {
+                foreach (var _ in GetAnchored(mapId, box, flags))
+                {
+                    return true;
+                }
```
gradientvera: Needs to check lookup flags for the anchored flag

Robust.Shared/GameObjects/Systems/EntityLookup.cs
```diff
@@ -298,14 +359,19 @@ public IEnumerable<IEntity> GetEntitiesIntersecting(MapId mapId, Box2 position,
                         list.Add(ent);
                     }
                     return true;
-                }, offsetBox, approximate);
+                }, offsetBox, (flags & LookupFlags.Approximate) != 0x0);
+            }
+
+            foreach (var ent in GetAnchored(mapId, position, flags))
+            {
+                list.Add(ent);
```
gradientvera, line 364: Probably needs to check the flags for the anchored flag

Robust.Shared/GameObjects/Systems/EntityLookup.cs
```diff
@@ -433,21 +503,37 @@ public IEnumerable<IEntity> GetEntitiesInRange(IEntity entity, float range, bool
...
+            {
+                foreach (var tile in grid.GetAllTiles())
+                {
+                    foreach (var ent in grid.GetAnchoredEntities(tile.GridIndices))
+                    {
+                        if (!_entityManager.TryGetEntity(ent, out var entity)) continue;
+                        yield return entity;
+                    }
+                }
+            }
```
gradientvera, line 531: Probably needs to check the flags for the anchored flag

gradientvera (changes requested):

Robust.Shared/GameObjects/Systems/EntityLookup.cs
metalgearsloth: I double-checked and GetAnchored already checks for the flag and will yield break. I mainly added it as a helper for entitylookup internally to avoid duplicating it (especially when more methods get added in future).

PR #1985, opened 2021-08-27
Chunk streaming

### LINKED ISSUE #1984: chunk pop-up needs fixing
Currently when the client moves around it can get up to 3 chunks / up to 768 anchored entities dumped on it at once from pvs which is bad.

We can probably just enlarge pvs to get a lower priority chunk range (for any chunks the client has never seen before and isn't already included in pvs) and send these entities over multiple ticks instead to reduce this impact.

If sending a whole chunk is still bad (It probably will be) you could piecemeal it further and send say 1/4 of a chunk every tick (when we aren't already sending a bunch of chunks).

metalgearsloth:
Fixes https://github.com/space-wizards/RobustToolbox/issues/1984
Requires https://github.com/space-wizards/RobustToolbox/pull/1983

There's still a framedrop in debug but it won't be a full-on game freeze, and on release it shouldn't be noticeable.


metalgearsloth:
Could do with more tweaks on range + how fast it goes but should be okay for first version.

PR #1988, opened 2021-08-27
Adds ComponentGetState ECS event

gradientvera:
Requires #1987
Adds a way to get component states from an ECS event!


PJB3005 (approved):
@Zumorica compile fails

gradientvera:
@PJB3005 My bad, should be good now

PR #1993, opened 2021-08-30
Add component for higher PVS range

metalgearsloth:
Thought about using an interface but using a component seemed easier where multiple components wished to increase the PVS range.

Currently used for lights. Needs another PR to actually fix lights because currently the client and server data is not the same at all so the client will still get pop-in as the server thinks all the lights are off.


metalgearsloth:
The code's a bit messy and most of it needs rewriting now that rotation is in so I'll close it for now.

PR #1995, opened 2021-08-30
Sync PointLight data between client and server.

### LINKED ISSUE #1994: PointLight data isn't synchronised
I think it's visualizers overriding it but if you vv PointLightComponent they aren't remotely similar on client and server.

metalgearsloth:
Also did a bit of ECS at the time. Main goal was just sharing more between client and server e.g. their defaults were different.
Fixes https://github.com/space-wizards/RobustToolbox/issues/1994


Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
```diff
@@ -0,0 +1,86 @@
...
+                if (_offset.EqualsApprox(value)) return;
+                _offset = value;
+                Dirty();
+            }
+        }
+
+        public override ComponentState GetComponentState(ICommonSession player)
+        {
+            return new PointLightComponentState(_enabled, _color, _radius, _offset);
+        }
```
gradientvera: This can be ECS'd easily now

PR #2003, opened 2021-09-05
Fix 90% of grid rotation bugs

metalgearsloth:
Still need some tests


Robust.Client/GameObjects/EntitySystems/EyeUpdateSystem.cs
```diff
@@ -74,17 +75,16 @@ public override void FrameUpdate(float frameTime)
...
-
-                var dot = Vector2.Dot(closestDir, currentDir);
-                if (MathHelper.CloseTo(dot, 1, CameraSnapTolerance))
-                {
-                    currentEye.Rotation = closestDir.ToAngle();
-                }
+                // TODO: Detect parent change and start lerping
+
+                var parentRotation = parent.WorldRotation;
+                currentEye.Rotation = -parentRotation;
```
metalgearsloth, line 87: This kinda got jumbled in when I was debugging but TBH it needs a more robust system so content can override it so it's probably fine for now I guess.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1217,8 +1217,7 @@ internal void Render(DrawingHandleWorld drawingHandle, Angle eyeRotation, Angle
             RenderInternal(drawingHandle, eyeRotation, worldRotation, Vector2.Zero, overrideDirection);
         }
 
-        [DataField("noRot")]
-        private bool _screenLock = true;
+        [DataField("noRot")] private bool _screenLock = false;
```
PaulRitter, line 1220: why change the default? imo having true as a default makes a bit more sense (?)

PaulRitter (approved):
looks good reading over it. will check it out in game/rider next week when i got time again. (if its not merged by then)

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
metalgearsloth: I imagine most things would rotate with the station by default. Without this it's locked to your screen so if the station is rotated 45 degrees stuff like walls look terrible because they're half-on half-off the tile.

PR #2004, opened 2021-09-05
Fix Robust tests

metalgearsloth:
Not the greatest solutions but at least it's fixed for now.


gradientvera (approved):
Ugly.
But it works.

PR #2005, opened 2021-09-06
Significantly reduce debug allocations

metalgearsloth:
I know it's only debug but it's number 1 on entity spawning

image


PJB3005:
.NET 6 when https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/

PR #2006, opened 2021-09-06
Optimise animations a bit

metalgearsloth:
Also did some ECS at the time.


Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
```diff
@@ -1,15 +1,64 @@
...
-                animationPlayerComponent.Update(frameTime);
+                var anim = _activeAnimations[i];
+                if (!Update(anim, frameTime)) continue;
+                _activeAnimations.RemoveAt(i);
             }
         }
+
+        internal void AddComponent(AnimationPlayerComponent component)
+        {
+            if (_activeAnimations.Contains(component)) return;
```
PJB3005: This being a list could be quite problematic perf wise if many animation players start playing at once. Maybe keep a bool flag on the component "I am currently playing" instead?

Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
```diff
@@ -1,15 +1,64 @@
...
+
         public override void FrameUpdate(float frameTime)
         {
-            foreach (var animationPlayerComponent in EntityManager.ComponentManager.EntityQuery<AnimationPlayerComponent>(true))
+            for (var i = _activeAnimations.Count - 1; i >= 0; i--)
             {
-                animationPlayerComponent.Update(frameTime);
+                var anim = _activeAnimations[i];
+                if (!Update(anim, frameTime)) continue;
+                _activeAnimations.RemoveAt(i);
```
PJB3005: This is again really bad perf wise. Use `RemoveSwap()` instead?

Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
```diff
@@ -1,15 +1,64 @@
...
+            if (_activeAnimations.Contains(component)) return;
+            _activeAnimations.Add(component);
+        }
+
+        private bool Update(AnimationPlayerComponent component, float frameTime)
+        {
+            if (component.PlayingAnimationCount == 0)
+                return true;
+
+            List<string>? toRemove = null;
```
PJB3005: You can use `RemQueue<T>` for this now.

PJB3005:
This causes a very gnarly test failure related to dummy icons by the way. Not sure how to go about handling that.

Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
metalgearsloth: good idea

Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
metalgearsloth: Unfortunately I did consider this but the greatness that is RemQueue is on content (for now)

metalgearsloth: Alright made a new PR for https://github.com/space-wizards/RobustToolbox/pull/2066 so will update after that's merged

PR #2007, opened 2021-09-06
Avoid unnecessary TransformComponent cast in PVS

metalgearsloth:
I profiled for a minute running in release with 20 clients:
master:
image
PR:
image

Although it doesn't use the transform property but this supports removing IEntity better + avoids all of the unnecessary casts when going up transforms (which I imagine happens very frequently).


gradientvera:
Unrelated, but deleting `ITransformComponent` in favor of just `TransformComponent` whence

Clyybber (approved):

PR #2027, opened 2021-09-12
Accurate grid bounds

metalgearsloth:
Currently how broadphase is setup driving the station is gonna be relatively slow (Saltern would lead to a >100 fixture broadphase update) but I'll deal with that in future to special-case grids and make this PR smaller. I might just store them in their own tree by their whole AABB or something.

I'm not sure how to get the fixture count much smaller without making chunks bigger. Tiles frequently lead to concave shapes so currently it just does the largest rectangles it can in each chunk. If we did diagonals then we could significantly cut down on fixture size but it means you wouldn't be able to slot in ships like tetris.

Content PR: https://github.com/space-wizards/space-station-14/pull/4632


Robust.Shared/GameObjects/ComponentManager.cs
```diff
@@ -377,6 +373,17 @@ public bool HasComponent(EntityUid uid, ushort netId)
                    && netSet.ContainsKey(netId);
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public T EnsureComponent<T>(IEntity entity) where T : Component, new()
```
mirrorcult, line 378: the rest of the componentmanager methods take EntityUid, why does this take IEntity?

metalgearsloth: AddComponent takes IEntity actually which is the sole reason I did that.

gradientvera: Bad, make this take EntityUid

metalgearsloth:
this

PR #2033, opened 2021-09-13
Makes BoundUserInterfaces partially ECS.

gradientvera:
- All received BoundUserInterfaceMessages will be raised as directed events.
- Base BoundUserInterfaceMessage now has fields for the entity UID, UI key and relevant player session.
    - Please note that these are *only* set before raising the directed event, for ECS purposes.
- Made UserInterfaceSystem public, added a bunch of *proxy* methods to it.
    - These proxy methods call into the actual BoundUserInterface methods. In the future the logic will be moved into the systems though.
    - This means you can start using these methods to ECS your bound user interfaces!
- Added BoundUserInterfaceClosedEvent for when a bound user interface is closed. Yay, ECS!

I think this is done now. Please review and give me feedback on these changes!
For an example of the changes in this PR being used, refer to this content PR: https://github.com/space-wizards/space-station-14/pull/4608

### DIFF
--- a/Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
+++ b/Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
@@ -175,7 +175,7 @@ namespace Robust.Server.GameObjects
         ///     Thrown if the session's status is <c>Connecting</c> or <c>Disconnected</c>
         /// </exception>
         /// <exception cref="ArgumentNullException">Thrown if <see cref="session"/> is null.</exception>
-        public void Open(IPlayerSession session)
+        public bool Open(IPlayerSession session)
         {
             if (session == null)
             {
@@ -189,7 +189,7 @@ namespace Robust.Server.GameObjects
 
             if (_subscribedSessions.Contains(session))
             {
-                return;
+                return false;
             }
 
             _subscribedSessions.Add(session);
@@ -208,6 +208,7 @@ namespace Robust.Server.GameObjects
             }
 
             session.PlayerStatusChanged += OnSessionOnPlayerStatusChanged;
+            return true;
         }
 
         private void OnSessionOnPlayerStatusChanged(object? sender, SessionStatusEventArgs args)
@@ -223,7 +224,7 @@ namespace Robust.Server.GameObjects
         /// </summary>
         /// <param name="session">The session to close the UI on.</param>
         /// <exception cref="ArgumentNullException">Thrown if <paramref name="session"/> is null.</exception>
-        public void Close(IPlayerSession session)
+        public bool Close(IPlayerSession session)
         {
             if (session == null)
             {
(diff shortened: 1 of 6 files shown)


Robust.Shared/GameObjects/Components/UserInterface/SharedUserInterfaceComponent.cs
```diff
@@ -89,4 +107,18 @@ public override string ToString()
             return $"{nameof(BoundUIWrapMessage)}: {Message}";
         }
     }
+
+    public class BoundUserInterfaceClosedEvent : EntityEventArgs
```
DrSmugleaf: How much to rename it to something like BUIClosedEvent instead

Robust.Shared/GameObjects/Components/UserInterface/SharedUserInterfaceComponent.cs
```diff
@@ -89,4 +107,18 @@ public override string ToString()
...
+        public readonly EntityUid Entity;
+        public readonly object UiKey;
+
+        public BoundUserInterfaceClosedEvent(EntityUid entity, object uiKey, ICommonSession session)
+        {
+            Entity = entity;
+            UiKey = uiKey;
+            Session = session;
+        }
+    }
```
mirrorcult: why not have this inherit BoundUserInterfaceMessage, has all the same fields

Robust.Shared/GameObjects/Components/UserInterface/SharedUserInterfaceComponent.cs
gradientvera: BUI sounds bad tho :pensive:

DrSmugleaf: I'll sell you BoundUIClosedEvent for 3.50

gradientvera: fair, that doesn't sound bad

Robust.Shared/GameObjects/Components/UserInterface/SharedUserInterfaceComponent.cs
gradientvera: done

Robust.Shared/GameObjects/Components/UserInterface/SharedUserInterfaceComponent.cs
```diff
@@ -89,4 +107,8 @@ public override string ToString()
             return $"{nameof(BoundUIWrapMessage)}: {Message}";
         }
     }
+
+    public class BoundUIClosedEvent : BoundUserInterfaceMessage
+    {
+    }
```
PaulRitter: shouldn't this still have a ctor and fill all of the properties inherited from `BoundUserInterfaceMessage`?

Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
```diff
@@ -232,16 +233,19 @@ public void Close(IPlayerSession session)
...
             var msg = new CloseBoundInterfaceMessage();
             SendMessage(msg, session);
             CloseShared(session);
+            return true;
         }
 
         private void CloseShared(IPlayerSession session)
         {
+            var owner = Owner.Owner;
+            owner.EntityManager.EventBus.RaiseLocalEvent(owner.Uid, new BoundUIClosedEvent());
```
PaulRitter: see other comment, this should be given the appropriate values via ctor

PaulRitter (changes requested):

Robust.Shared/GameObjects/Components/UserInterface/SharedUserInterfaceComponent.cs
gradientvera: when you address reviews so fast you forget something this vital

PaulRitter (approved):

PR #2038, opened 2021-09-14
new xaml error for invalid root node type

### LINKED ISSUE #2037: Inheriting a different control than the one at the root of a xaml file does not cause an appropriate error
Instead it causes a sandbox error.

```
[ERRO] res.typecheck: Content.Client: ILVerify: Unexpected type on the stack., method: void Content.Client.Body.Surgery.UI.PartButton.!XamlIlPopulateTrampoline(Content.Client.Body.Surgery.UI.PartButton)
```

To reproduce:
- Make a .xaml file for a control.
- Make its corresponding .xaml.cs file.
- Make the root control in the .xaml file different from the one that the .xaml.cs class is inheriting.

PaulRitter:
requires https://github.com/space-wizards/space-station-14/pull/4618
fixes #2037


PJB3005:
github is a well designed ui with no problems

PR #2046, opened 2021-09-18
Re-parent unanchored entities

metalgearsloth:
Requires https://github.com/space-wizards/RobustToolbox/pull/2027

Essentially when something is unanchored currently it just assumes it's still attached to the grid even though the bounds may no longer be valid.

This method only gets called when a tile gets set to empty so whenever that happens it will attach everything that was unanchored to the map instead.

Reason it requires the other PR is because that actually has accurate bounds.


gradientvera:
Needs updating for `ComponentManager` -> `EntityManager`

PaulRitter (approved):

PJB3005:
Conflicts.

PR #2047, opened 2021-09-19
EntitySystem Resolve helper methods.

gradientvera:
Part of the [ESM Resolves Proposal](https://docs.spacestation14.io/en/proposals/esm-resolves).
~~DO NOT MERGE BEFORE THIS IS DISCUSSED AND AGREED UPON BY MAINTAINERS.~~ This was agreed to, yay

Also paging @PJB3005 for review, does aggressive inlining make sense here?

### DIFF
--- a/Robust.Shared/GameObjects/EntitySystem.Resolve.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.Resolve.cs
@@ -1,5 +1,6 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
+using Robust.Shared.Utility;
 
 namespace Robust.Shared.GameObjects
 {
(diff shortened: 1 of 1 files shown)


gradientvera:
Since this was agreed upon by maintainers, this is good for merge now

Robust.Shared/GameObjects/EntitySystem.Resolve.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Robust.Shared.GameObjects
+{
+    public abstract partial class EntitySystem
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        protected bool Resolve<TComp>(EntityUid uid, [NotNullWhen(true)] ref TComp? component)
+            where TComp : IComponent
+        {
+            DebugTools.Assert(uid == component?.Owner.Uid, "Specified Entity is not the component's Owner!");
```
PJB3005: Won't this throw if the component is null?

gradientvera: Oh, you're right. I fixed and properly tested the assert now, should be good to go

PR #2049, opened 2021-09-19
Remove some unused obsolete stuff

Visne:
Requires space-wizards/space-station-14#4657.


Robust.Client/UserInterface/Controls/GridContainer.cs
```diff
@@ -84,55 +84,12 @@ public int Rows
...
-        {
-            set => MaxGridWidth = value;
-        }
 
         public float MaxGridWidth
         {
             set => SetMaxSize(Dimension.Column, value);
         }
 
-        /// <summary>
```
mirrorcult: these xmldocs seem valuable, put them on the non-obsolete property instead? (same goes for maxwidth)

Robust.Shared/GameObjects/Components/ITransformComponent.cs
```diff
@@ -89,11 +89,6 @@ public interface ITransformComponent : IComponent
         /// </summary>
         public EntityUid ParentUid { get; set; }
 
-        /// <summary>
-        /// Whether or not this entity is on the map, AKA it has no parent.
-        /// </summary>
-        bool IsMapTransform { get; }
```
mirrorcult, line 95: why is this removed?

Robust.Client/UserInterface/Controls/GridContainer.cs
Visne: Done.

PR #2054, opened 2021-09-20
Fix occluder edges for grid rotation

metalgearsloth:
(is eye.Scale the right thing to use for the view matrix?)

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
@@ -335,8 +335,9 @@ namespace Robust.Client.Graphics.Clyde
             }
 
             var (lights, count, expandedBounds) = GetLightsToRender(mapId, worldBounds, worldAABB);
+            eye.GetViewMatrix(out var eyeTransform, eye.Scale);
 
-            UpdateOcclusionGeometry(mapId, expandedBounds, eye.Position.Position);
+            UpdateOcclusionGeometry(mapId, expandedBounds, eyeTransform);
 
             DrawFov(viewport, eye);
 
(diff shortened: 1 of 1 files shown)


Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -908,10 +909,10 @@ private void UpdateOcclusionGeometry(MapId map, Box2 expandedBounds, Vector2 eye
...
 
                         // Calculate delta positions from camera.
-                        var (dTlX, dTlY) = (tlX, tlY) - eyePosition;
-                        var (dTrX, dTrY) = (trX, trY) - eyePosition;
-                        var (dBlX, dBlY) = (blX, blY) - eyePosition;
-                        var (dBrX, dBrY) = (brX, brY) - eyePosition;
+                        var (dTlX, dTlY) = eyeTransform.Transform(tl);
+                        var (dTrX, dTrY) = eyeTransform.Transform(tr);
+                        var (dBlX, dBlY) = eyeTransform.Transform(bl);
+                        var (dBrX, dBrY) = eyeTransform.Transform(br);
```
metalgearsloth, line 915: The actual important bit

metalgearsloth, line 915: The important bit

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -335,8 +335,9 @@ private void DrawLightsAndFov(Viewport viewport, Box2Rotated worldBounds, Box2 w
             }
 
             var (lights, count, expandedBounds) = GetLightsToRender(mapId, worldBounds, worldAABB);
+            eye.GetViewMatrix(out var eyeTransform, eye.Scale);
```
PaulRitter, line 338: without having looked at the code: why tf is this using an out var? is this a tryget? if so use it like that

PaulRitter (approved):
lgtm besides the one nitpick

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
metalgearsloth: It's void for reasons.

PaulRitter: weird af

PR #2057, opened 2021-09-22
Option to blur FOV

mirrorcult:
This PR adds an option to blur the FOV applied to the final framebuffer, the same way that lighting is blurred. There are a couple issues with it right now:

- [ ] For whatever reason, it stopped working inexplicably. It worked perfectly on a commit, I tried some new changes that didn't work, and then reverted back to that same commit and it stopped working entirely. I have no fucking explanation for this
- [ ] FOV blur framebuffers should use R8 instead of Rgba8 to reduce VRAM usage since FOV is just an alpha channel. Basically had this working but again, stopped working for literally no reason
- [ ] On anything lower than high lighting settings the viewport needs to be scaled down by half
- [ ] Need to profile it. Cursory testing it doesn't look that bad for perf and it can always be disabled. R8 channel should help with this too

This is what it looks like at default settings:

image
image


Acruid:
Closed for inactivity. Feel free to open this again if someone picks it up.

ike709:
Sad, this was a good PR.

mirrorcult:
i will do it eventually but the heisenbug killed my motivation a bit

PR #2060, opened 2021-09-24
Adds everything else OpenDream needs to the sandbox

ike709:
https://github.com/wixoaGit/OpenDream/issues/374


Robust.Shared/ContentPack/Sandbox.yml
```diff
@@ -13,6 +13,7 @@ AllowedVerifierErrors:
 WhitelistedNamespaces:
 - Robust
 - Content
+- OpenDreamShared
```
ike709, line 16: This isn't ideal in the long run but PJB caved so... /shrug

Robust.Shared/ContentPack/Sandbox.yml
```diff
@@ -467,6 +470,8 @@ Types:
     UnmanagedType: { } # `unmanaged` constraint in C# modreqs with this.
   System.Runtime.Versioning:
     TargetFrameworkAttribute: { All: True }
+  System.Text.Json.Serialization:
+    JsonIgnoreAttribute: { All: True }
```
ike709, line 479: This is actually just for the ctor, should I change it or does it not really matter?

Robust.Shared/ContentPack/Sandbox.yml
```diff
@@ -299,6 +300,7 @@ Types:
     IEnumerator: { All: True }
     IReadOnlyList`1: { All: True }
   System.ComponentModel:
+    CancelEventArgs: { All: True }
```
ike709, line 303: This is actually just for the setter, should I change it or does it not really matter?

Robust.Shared/ContentPack/Sandbox.yml
PJB3005: doesn't really matter, though you won't be able to actually make use of these attributes ever anyways as it's reflection-heavy deserialization we can't allow.

ike709:
oops I goofed the whitespace

Robust.Shared/ContentPack/Sandbox.yml
```diff
@@ -332,6 +334,7 @@ Types:
     CompareOptions: { }
     CultureInfo: { All: True }
     DateTimeStyles: { All: True } # Enum
+    NumberStyles: { All: True } # Enum
```
ike709:
```suggestion
    NumberStyles: { } # Enum
```

ike709:
OpenDream now has everything whitelisted that it needs whitelisted.

PaulRitter (approved):

Robust.Shared/ContentPack/Sandbox.yml
```diff
@@ -342,14 +345,20 @@ Types:
...
   System.IO:
+    BinaryReader: { All: True }
     FileAccess: { } # Enum
     FileMode: { } # Enum
     FileShare: { } # Enum
     FileNotFoundException: { All: True }
     InvalidDataException: { All: True }
     IOException: { All: True }
     MemoryStream: { All: True }
+    Path: { All: True }
```
PJB3005: Sorry for looking at this so late but `Path` can be used to probe file system layout (Short path expansion in `Path.GetFullPath()`) and OS setup (is the user on Windows or Linux?).

image

This API should not be exposed to content.

PJB3005 (changes requested):

Robust.Shared/ContentPack/Sandbox.yml
ike709:
```suggestion
```
Path is no longer needed as I moved the entirety of the compiler out of shared.

PJB3005 (approved):

PR #2065, opened 2021-09-25
Adds support for "1x2" Vector2i serializing

ike709:
Opendream uses `480x640` rather than `480,640`, for example.


PaulRitter (approved):
mental note to eventually have all vector serializers share logic to parse args. i hate the copypaste they are rn

PR #2067, opened 2021-09-26
Optimise anchored entity states for PVS

metalgearsloth:
I hate IEnumerable I hate IEnumerable I hate IEnumerable.

I split out parts of PVS into their own methods just to make profiling easier.

### DIFF
--- a/Robust.Shared/Map/MapChunk.cs
+++ b/Robust.Shared/Map/MapChunk.cs
@@ -271,6 +271,20 @@ namespace Robust.Shared.Map
             }
         }
 
+        public void FastGetAllAnchoredEnts(EntityUidQueryCallback callback)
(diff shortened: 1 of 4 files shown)


Robust.Shared/Map/MapChunk.cs
```diff
@@ -271,6 +271,20 @@ public IEnumerable<EntityUid> GetAllAnchoredEnts()
...
+            {
+                if (cell.Center is null)
+                    continue;
+
+                foreach (var euid in cell.Center)
+                {
+                    callback(euid);
+                }
+            }
+        }
```
metalgearsloth, line 286: The actual change

gradientvera (approved):

PR #2068, opened 2021-09-26
Move LifeStage and Paused data to MetaDataComponent.

gradientvera:
SS14 goes in-game with this PR, and no tests seem to fail whatsoever.

- Removes `IMetaDataComponent`
- Moves `LifeStage` and `Paused` data out of `Entity` and into `MetaDataComponent`.
- `MetaDataComponent` is deleted after `TransformComponent` now.
- Added `AddComponent<T>` and `EnsureComponent<T>` overloads in `EntityManager` that take in `EntityUid`s.

@PJB3005 please review!


Robust.Shared/GameObjects/Components/MetaDataComponent.cs
```diff
@@ -140,14 +125,37 @@ public string EntityDescription
...
+        [ViewVariables]
+        public EntityLifeStage EntityLifeStage { get; internal set; }
+
+        [ViewVariables]
+        public bool EntityPaused
+        {
+            get => _entityPaused;
+            set
+            {
+                if (_entityPaused == value || value && Owner.HasComponent<IgnorePauseComponent>())
```
PaulRitter, line 141: at what point are we making ignorepause a field on this component?

Robust.Shared/GameObjects/Entity.cs
```diff
@@ -76,32 +67,23 @@ public string Name
...
-        private IMetaDataComponent? _metaData;
+        private MetaDataComponent? _metaData;
 
         /// <inheritdoc />
         [ViewVariables]
-        public IMetaDataComponent MetaData => _metaData ??= GetComponent<IMetaDataComponent>();
+        public MetaDataComponent MetaData
+        {
+            get => _metaData ??= GetComponent<MetaDataComponent>();
+            internal set => _metaData = value;
```
PaulRitter, line 85: can the metadata component be changed? should it be?

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -359,8 +360,12 @@ private protected Entity AllocEntity(EntityUid? uid = null)
             // We do this after the event, so if the event throws we have not committed
             Entities[entity.Uid] = entity;
 
-            // allocate the required MetaDataComponent
-            AddComponent<MetaDataComponent>(entity);
+            // Create the MetaDataComponent and set it directly on the Entity to avoid a stack overflow in DEBUG.
+            var metadata = new MetaDataComponent() { Owner = entity };
+            entity.MetaData = metadata;
```
PaulRitter, line 367: dont know how i feel about making metadata settable, even internally, for this one piece of code. is there maybe another way to do this?

Robust.Shared/Timing/PauseManager.cs
```diff
@@ -115,9 +115,6 @@ public void PostInject()
                 _unInitializedMaps.Add(args.Map);
             };
 
-            if(_conhost.IsClient)
-                return;
-
```
PaulRitter, line 120: why remove this?

gradientvera: Because it didn't even work on the server, and honestly there's some value to having these on the client too.
https://discord.com/channels/310555209753690112/770682801607278632/891636371817652224

Robust.Shared/GameObjects/Components/MetaDataComponent.cs
gradientvera: Good question, it should probably be. I'll open an issue and do it in the future once this is merged

Robust.Shared/GameObjects/Entity.cs
gradientvera: See review below

Robust.Shared/GameObjects/EntityManager.cs
gradientvera: Not at all. When you add the MetaDataComponent to the entity, it will `Dirty()` it, which in turns `Dirty()`s the `Entity`, and on `DEBUG` that checks `entity.Deleted`, which gets `MetaData`, causing in a very hilarious and very funny stack overflow.
I seriously doubt there's a better way to do this...

PaulRitter: dangit

PaulRitter (approved):

PR #2072, opened 2021-09-26
fix placement grid

### LINKED ISSUE #1891: SnapgridCenter placement doesn't account for viewport size.
image

### LINKED ISSUE #2112: Constructiongrid renders relative to map, not grid
should be relative to grid

PaulRitter:
fixes #2112

placement grid is now relative to the grid.

still breaks when you resize the window, no clue why. @PJB3005 
if we ever get around to fixing that, it will also fix #1891


metalgearsloth:
> still breaks when you resize the window, no clue why. @PJB3005

Because it doesn't consider viewport size and just does the lines every 32 pixels.

PaulRitter:
> > still breaks when you resize the window, no clue why. @PJB3005
> 
> Because it doesn't consider viewport size and just does the lines every 32 pixels.

b-but i used `EyeManager.PixelsPerMeter` 👉👈

metalgearsloth (approved):

PaulRitter:
suppose maybe eyemanager.pixelspermeter is just broken then? paging @PJB3005

20kdc:
I think there's a misunderstanding here of what pixelspermeter is.
pixelspermeter is the "normal", unaltered, amount of pixels per meter (read: tile), assuming that SS14 ran at 1x resolution like a BYOND game would.
So it's a constant 32.
But practically, there are effects like the zoom of an eye that change the relation between this and actual "physical" pixels as used by ScreenToWorld.

PJB3005:
Pixels per meter is the relation of sprite size -> world units. Nothing more.

PR #2074, opened 2021-09-27
Adds utility "PrototypeFlags" collection.

gradientvera:
This PR adds a new utility collection class, `PrototypeFlags`.
This collection acts as a set of unordered strings which resolve to prototypes.
This class can be (de)serialized as a sequence of strings, with full linter support.

This is useful for things such as `TagComponent` in SS14, or tool quality prototypes in the future.
I feel like this helper class should be in engine, as it's a fairly simple collection that's unlikely to see many changes.


PaulRitter (approved):
lgtm

PR #2078, opened 2021-09-29
Aggressively sleep physics bodies

metalgearsloth:
image

Actual changes:
* CloseTo renamed to CloseToPercent so physics can do minute body changes
* Sleep tolerance threshold significantly increased
* Damping significantly increased (20%)

I was mainly concerned with how weightless items and shuttles performed, i.e. weightless items still look floaty but sleep as fast as possible. This is why I had to do a 5x multiplier when initially moving the shuttle on content to make sure it exceeded the sleep threshold from the initial impulse.


gradientvera (approved):

gradientvera:
`TestCircleStack` and maybe more tests are failing

PR #2085, opened 2021-10-02
Opt out of shadow casting for some lights

mirrorcult:
@PJB3005 

not a CVar since setting it sufficiently high essentially disables shadowcasting. impetus of this PR is to make sure that https://github.com/space-wizards/space-station-14/pull/4709 doesn't lag too much and we can have a lot of pretty lights. every light added in that PR is <= 2.0 radius, and really small lights -essentially- don't cast shadows at all so it's a waste to do the occlusion depth calcs for this. Maybe this doesn't optimize that much, I can't really tell (seems to save a bit of FPS on my machine, 3-5 or so, might be better on igpu)

Things we might want to do
- Decrease the distance?
- Make sure walllights/emergency lights in SS14 have AlwaysCastShadows set
- Remove the minimum radius stuff and just force every light that doesn't want to cast shadows to specify that (cbt)


mirrorcult:
also might want to increase max lights but 128 seems okay especially for saltern

mirrorcult:
an example of the difference visually (literally the only place on saltern where I could find a visual difference)

no shadow casting for this small light:
image

yes shadow casting:
image

mirrorcult:
nvm this makes them visible through walls lol

mirrorcult:
actually this is salvageable i just made the minimum radius smaller

PJB3005:
1.5 feels a little high to me because that feels like it could still show through walls in some edge cases. Conservatively I'd say it should be no more than 1 (because that's how thick walls are).

PJB3005:
My understanding is that stuff like the computer lights are too large to fit under 1 radius, but they can't clip because they're always fixed in the center of the tile, right? Maybe make it a flag on the light itself to disable shadow casting?

mirrorcult:
opted to remove the culling-by-radius thing and i'll just make every light opt in to not cast shadows, will make a content PR for this later

PaulRitter:
> 1.5 feels a little high to me because that feels like it could still show through walls in some edge cases. Conservatively I'd say it should be no more than 1 (because that's how thick walls are).

terrible idea imo, since we might have occluders that are < 1 tile in all directions

suppose this is a tricky one

PJB3005:
> terrible idea imo, since we might have occluders that are < 1 tile in all directions

Cross that bridge when we actually *have* such occluders, because this is completely non-trivial to actually add due to how the occlusion system actually works.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -361,6 +361,8 @@ private void DrawLightsAndFov(Viewport viewport, Box2Rotated worldBounds, Box2 w
                     {
                         var (light, lightPos, _) = lights[i];
 
+                        if (!light.CastShadows) continue;
```
PJB3005, line 364: I just realized that because you're only doing a continue here, you're still taking up a slot in the shadow atlas and still trying to *draw* a shadow aswell (just you end up drawing no shadow) so this doesn't perform as well as it could either.

PJB3005 (changes requested):

PJB3005 (approved):

PR #2089, opened 2021-10-03
Adds Write/Copy TypeSerializer methods for base SpriteSpecifier.

gradientvera:
Allows us to serialize/copy the base `SpriteSpecifier` as its correct type.
@PaulRitter pls review


PaulRitter (approved):
lgtm

PR #2091, opened 2021-10-03
Fix stuck lathes properly.

### LINKED ISSUE #4518: Protolathe freezes if items are added to the queue too quickly.
This bug was originally reported by Raven#1990 on Discord. I reproduced it in a local environment.

How to reproduce:

- Fill the protolathe with enough materials to make more than one of a given item. I was able to make this happen with around half a stack of steel.

- Rapidly click on an item. 

- The protolathe will continue queuing the item you're mashing clicks on until you stop clicking, and will then break. The sprite will go into its "off" state and while it will continue to add the items into the queue, it won't print anything else out. 

You can fix the problem by unscrewing the protolathe and crowbarring it back into a machine frame, but this will eat all the materials that were in the lathe.

Ideally, the protolathe should probably subtract the materials when the button is clicked, and deny the ability to queue more items than it physically has resources for. Alternatively, put a cooldown on how fast items can be added to the queue and how long the queue can be.

**Screenshots**

image

image

image

image

image

20kdc:
**Big warning: This PR needs severe testing. It messes around with the component registry. Feel free to close it and put in a PR that does something else, but I would not recommend simply disabling TimerComponent auto-removal as that leaves the engine open to future problems of this kind.**

Ought to fix https://github.com/space-wizards/space-station-14/issues/4518

# An explaination

So, you may know that autolathes and protolathes on SS14 get stuck right now.
What's happening?

Well, here's how a lathe works in relation to the relevant systems:
image

And here's what actually happens when a lathe freezes:
image

In further detail, when the old TimerComponent is finally removed from the deferred parts of the component dictionaries, it is not actually checked that the targetted parts of the component dictionaries refer to the old TimerComponent.

This PR performs that check and avoids doing any "anonymous" component actions in that event, on the basis that a replacement component exists.

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.Components.cs
+++ b/Robust.Shared/GameObjects/EntityManager.Components.cs
@@ -330,24 +330,29 @@ namespace Robust.Shared.GameObjects
 
             var entityUid = component.Owner.Uid;
 
-            foreach (var refType in reg.References)
+            // If another component of the same type was added, then we can't actually remove this properly.
+            // To test this case, have a component that gets removed, and then is EnsureComponent'd.
+            // If this is broken, the removal of the old component will get rid of the new component.
+            if (_entCompIndex.Remove(entityUid, component))
             {
-                _entTraitDict[refType].Remove(entityUid);
+                foreach (var refType in reg.References)
+                {
+                    _entTraitDict[refType].Remove(entityUid);
+                }
+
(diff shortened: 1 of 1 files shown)


gradientvera (approved):
Looks good to me, in-game it seems to work fine, and tests pass.
I'll keep an eye out for any possible side effects and/or exceptions on the live servers as well.

PR #2094, opened 2021-10-05
Don't log missing resolve on containers

metalgearsloth:
The reason it's a TryGet is because it may not exist


gradientvera:
What did he mean by this
76ab68dc3d6fbf07c392080edf1da8c90d8b5d87

metalgearsloth:
thanks fork

PR #2098, opened 2021-10-06
Update the comments related to the CEF binary

ike709:
Updates the version and makes it less misleading.


gradientvera:
rather than misleading, I'd say "outdated" lol

ike709:
> rather than misleading, I'd say "outdated" lol


> // Supposedly, you should just need libcef.so in Release and icudtl.dat in Resources...

Rather than misleading, I'd say "wrong"

gradientvera:
No shit, the comments are from when I couldn't even get CEF to work on my machine and I pushed the branch so PJB could take a look at it

PR #2102, opened 2021-10-07
Bandaid entity deletion entitylookup

### LINKED ISSUE #4787: Creating off-station grids causes localized lag on servers.
image

This bug was originally reported by Raven#1990 on Discord, who replicated the bug on a live server on 10/07/2021 around 11:45 PM EDT. I am unable to reproduce in a local environment, but this appears to be reproducible on the servers themselves.

How to reproduce:

- Create a bunch of grids by placing lattice (using rods inhand, not the spawn menu).
- Cover the lattice in steel to make floor.
- Place low walls on the floors.
- After enough low walls are placed, anything within the PVS view limit of the grids will cause huge amounts of lag that will stop when the grids are no longer in PVS. This happens for all players on a server.

metalgearsloth:
I still blame Q for the absolute state of EntityLookup don't @ me son

"Fixes https://github.com/space-wizards/space-station-14/issues/4787" and I use that term loosely because entitylookup needs a cleanup due to all the ordering bullshit going on.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/EntityLookup.cs
+++ b/Robust.Shared/GameObjects/Systems/EntityLookup.cs
@@ -669,9 +669,10 @@ namespace Robust.Shared.GameObjects
         {
             // TODO: Need to fix ordering issues and then we can just directly remove it from the tree
             // rather than this O(n) legacy garbage.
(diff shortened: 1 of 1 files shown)


PaulRitter (approved):
angery

PR #2103, opened 2021-10-07
Fix FOV shader to allow for s p i n

20kdc:
Theoretically this could've done what BlurOntoWalls does and messed with the projection and view matrices... but it's a rectangle that covers the whole screen. There's no need to calculate matrices for that part.
There is however a need to calculate inverse matrices to get back to world coordinates again, and that is done here through synthesizing the matrix via LocalToWorld for lack of better options.

### DIFF
--- a/Resources/Shaders/Internal/fov.swsl
+++ b/Resources/Shaders/Internal/fov.swsl
@@ -4,30 +4,27 @@ preset raw;
 
 const highp float g_MinVariance = 0.0;
 
-varying highp vec2 worldPosition;
+// World-space position offset from centre to pixel.
+varying highp vec2 worldSpaceDiff;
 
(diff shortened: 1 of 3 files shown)


20kdc:
had to rebase to fix tests, and encountered some weirdness with the cefglue stuff doing so

20kdc:
Tests were not in fact fixed, but judging by the error it's not my fault, so...

mirrorcult:
afaik you have to manually change the submodule url for cefglue to `https://github.com/space-wizards/cefglue`

metalgearsloth (approved):
merge it merge it merge it merge it merge it merge it merge it merge it merge it merge it merge it merge it

PR #2104, opened 2021-10-07
Adds ICefManager

ike709:
I'm getting OpenDream's unit testing working and we need to be able to not load CEF binaries.

### DIFF
--- a/Robust.Client.CEF/ICefManager.cs
+++ b/Robust.Client.CEF/ICefManager.cs
@@ -1,0 +1,13 @@
+namespace Robust.Client.CEF
+{
+    public interface ICefManager
+    {
+        void Initialize();
+
+        void CheckInitialized();
+
+        void Update();
+
+        void Shutdown();
+    }
+}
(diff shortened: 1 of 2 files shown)


Robust.Client.CEF/ICefManager.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    public interface ICefManager
+    {
+        void Initialize();
+
+        void CheckInitialized();
+
+        void Update();
+
+        void Shutdown();
```
ike709, line 11:
```suggestion
        void Shutdown();
        
        IBrowserWindow CreateBrowserWindow(BrowserWindowCreateParameters createParams);
```

ike709:
Checks will pass when someone updates content's RT version.

PJB3005:
I feel like all APIs outside of the `CreateBrowserWindow` should be made internal. That does raise the question for how Robust.Client.CEF should be loaded though and I'm not sure how to go about that.

PJB3005:
I guess the simplest approach would be that engine modules are declared in a resource manifest and loaded by the engine automatically. This does create some pains for loading it in dev environments though

ike709:
...What do you actually want me to do here?

ike709:
someone re-run checks

ike709:
Semi-annual reminder that this is a thing.

PJB3005:
Superseded by ba2f46424957d321db5504256824746bda0a3d9a

PR #2107, opened 2021-10-08
Add option to make `ScrollContainer` return size.

ElectroJr:
Currently a `ScrollContainer` always reports its desired size as zero. Apparently there's a reason for that but it causes issues when you want a control whose size is determined by the size of a `ScrollContainer`. This just adds a `bool` that toggles whether or not it reports its size as zero.

This feature would be nice to have for [space-wizards/space-station-14#4768](https://github.com/space-wizards/space-station-14/pull/4768), which currently just has some janky code to infer the actual desired size of the scroll container.

### DIFF
--- a/Robust.Client/UserInterface/Controls/ScrollContainer.cs
+++ b/Robust.Client/UserInterface/Controls/ScrollContainer.cs
@@ -20,6 +20,8 @@ namespace Robust.Client.UserInterface.Controls
         public int ScrollSpeedX { get; set; } = 50;
         public int ScrollSpeedY { get; set; } = 50;
 
+        public bool ReturnMeasure { get; set; } = false;
+
         public ScrollContainer()
         {
             MouseFilter = MouseFilterMode.Pass;
@@ -89,11 +91,9 @@ namespace Robust.Client.UserInterface.Controls
                 size = Vector2.ComponentMax(size, child.DesiredSize);
             }
 
-            // Unlike WPF/Avalonia we report ZERO here instead of available size.
-            // This is to fix a bunch of jank with e.g. BoxContainer.
-            // Tbh this might be a mistake.
-            // DockPanel when.
-            return Vector2.Zero;
+            // Unlike WPF/Avalonia we default to reporting ZERO here instead of available size. This is to fix a bunch
+            // of jank with e.g. BoxContainer.
+            return ReturnMeasure ? size : Vector2.Zero;
         }
 
         protected override Vector2 ArrangeOverride(Vector2 finalSize)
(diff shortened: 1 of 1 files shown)


ShadowCommander:
The reason why ScrollContainer returns size 0 is so that the parent control can resized without thinking that scroll container is super tall. (e.g. BoxContainer)
@PJB3005 should it be done like this or should it be something like an ActualSize property added somwhere?

PJB3005:
Is this supposed to only be useful when combined with disabling scrolling on one of the axis (so it gets automatic size and no scrolling horizontally, but scrolling vertically)?

ElectroJr:
It should still work if scrolling is enabled on both axes. I just tested by enabling Horizontal scrolling & giving the context menu a MaxWidth, and the scrollbar appears to work fine.


image

Though there are some unrelated issues with text line wrapping, which I never ran  into due to not giving the menu or buttons a MaxWidth, which I probably should do.

PJB3005:
In that case, something else is limiting the available size of the control with a `MaxWidth` or something, so it sounds better to fix the UI layout to simply make the scroll container take up full space top down?

ElectroJr:
So the actual controls involved here are
`Popup` -> `PanelContainer` -> `ScrollContainer` -> `BoxContainer` -> buttons / menu elements

The popup window is supposed to automatically resize in order to be as small as possible, with some upper limit. Without having `ScrollContainer` return its size, I don't know of a neat way of making that work. How do you determine the required size of the popup? I'm still new to the UI stuff, maybe there is an easy solution.

The current work around I've been using in the Popup's `MeasureOverride` is:
1. Get the desired size of the box container
2. If it's larger than the max height, limit it and add the width of the `ScrollContainer`'s vertical scroll bar (which is not publicly accessible, I've just been assuming it's the default width)
3. Get the PanelContainer's `StyleBox` and add the syle size to that to that measurement 
4. Return that as the desired size for the pop-up and open it

I hate that solution and It would be much simpler if the `ScrollContainer` were just able to just report the desired size directly to the `PanelContainer`, which in turn would report to its parent (the popup).

ShadowCommander:
Does the `ScrollContainer` check in your code need `DesiredSize` to be the property checked against or can it be something like a new property `ScrollContainer.ActualSize`?

ElectroJr:
If the `ScrollContainer` stored its desired size in some other property that would help with figuring out the scrollbar width, but AFAIK it would still require me to manually account for the the `PanelContainer`'s stylebox size.

Doing that only takes only a couple of lines of code, so I guess that's an OK solution but It still feels a little bit janky when the scroll container could just optionally report it's size directly to PanelContainer.

PJB3005:
Yeah you're right, this is necessary to be able to handle some scenarios. Looks good then.

PR #2124, opened 2021-10-11
Use spans for chunk mesh

metalgearsloth:
Using a stopwatch it was mainly an increase for the first mesh and the rest were more-or-less the same when I re-ran it a bunch (me lazy). try-catch was also removed given all it did was return the buffers.

Pool:
image

Spans:
image


PJB3005:
This also reduces memory overhead from the pool retaining the arrays used, so this is just a nice saving in general.

PR #2125, opened 2021-10-11
Add support for entity system timed updates

mirrorcult:
This is to avoid the whole `_accumulatedFrametime += frameTime` boilerplate that content does a lot everywhere. Introduces a new virtual method, TimedUpdate so you can still do stuff every tick and timed in the same entity system. Should still respect ordering. EntitySystems that don't override `TimedInterval` shouldn't ever call it and only incurs a single if check.

note: unsure if this is really the best way to do it or if this bloats entitysystems too heavily

### DIFF
--- a/Robust.Shared/GameObjects/EntitySystem.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.cs
@@ -26,6 +26,10 @@ namespace Robust.Shared.GameObjects
         protected internal List<Type> UpdatesAfter { get; } = new();
         protected internal List<Type> UpdatesBefore { get; } = new();
 
+        public virtual float TimedInterval { get; set; }
+
+        public float AccumulatedFrametime { get; set; }
+
         IEnumerable<Type> IEntitySystem.UpdatesAfter => UpdatesAfter;
         IEnumerable<Type> IEntitySystem.UpdatesBefore => UpdatesBefore;
 
@@ -42,6 +46,9 @@ namespace Robust.Shared.GameObjects
 
         /// <inheritdoc />
         public virtual void FrameUpdate(float frameTime) { }
+
+        /// <inheritdoc />
+        public virtual void TimedUpdate() { }
 
         /// <inheritdoc />
         public virtual void Shutdown()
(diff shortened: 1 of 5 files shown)


gradientvera:
This is gonna require some discussion

PaulRitter:
i feel like annotating a method with an attribute specifying the frequence would be better, since it would allow for more than one frequency/method per system (as is often seen in systems)

what i mean by this is:
```csharp
[TimedUpdate(0.5)]
private void GetsCalledEveryHalfSecond(){
    //stuff
}

[TimedUpdate(2)]
private void GetsCalledEveryTwoSeconds(){
    //stuff
}
```

PJB3005:
@PaulRitter I don't like that because of all the reflection magic.

20kdc:
API seems ok, though this is being added to every single EntitySystem which I find odd.
Paul's suggestion does look nice at least. I'm not sure how well/badly the reflection will go.

I do find the proliferation of this pattern to be a bit interesting though, everything's working at different tickrates ~~inb4 48tps EntitySystem~~, but it does makes sense given that different EntitySystems have different tick requirements.

Something key I think about all of this proliferation of differing lower tickrates is that above all else a key aspect of preventing jankiness might end up being to start staggering these to prevent overlap.

All that in mind, may I suggest something like SubscribeLocalEvent and such?
I'd suggest using a struct instead, but a central manager could time these for ideal jank prevention.

PJB3005:
Centralized staggering to avoid synchronization is a compelling case for supporting this in-engine yeah.

mirrorcult:
seems like consensus is to avoid a solution like this in favor of a more featured master controller?

20kdc:
I wouldn't say a more featured master controller would be necessary right this very moment. Just that the API design has to allow one to be drop-in. Just my opinion though.

PJB3005:
It's hard to design an API ahead of time for a future use case like that, so it's not very actionable right now.

metalgearsloth:
2 systems I can think of that could use it rn are gastileoverlays and radiation, though gastileoverlays is stupidly parallel.

PaulRitter:
> 
> 
> @PaulRitter I don't like that because of all the reflection magic.

would literally only add overhead at the system start. that being said having something like
`SubscribeUpdateMethod(float seconds, Action Method)` would be fine too imo.

PJB3005:
In that case we can do the same with all the calls we set up in `Initialize()` too, and stuff like the base `Update()` method doesn't use it which is inconsistent.

PR #2128, opened 2021-10-14
Fixes component deletion/removal not being synced with clients.

gradientvera:
Fixes networked components deletions not being synced to clients. Before, game state code assumed that `GetNetComponents` would return deleted components, which is no longer the case. This is also a step towards removing deferred component deletion. 

I made `GetDeletedComponents` return a `List<short>` for consistency with `EntityViewCulling`'s `GetDeletedEntities`, which returns a `List<EntityUid>`. I can code a struct enumerator instead if needed, though.

Please review @PJB3005

### DIFF
--- a/Robust.Server/GameObjects/IServerEntityNetworkManager.cs
+++ b/Robust.Server/GameObjects/IServerEntityNetworkManager.cs
@@ -1,10 +1,14 @@
+using System.Collections.Generic;
 using Robust.Server.Player;
 using Robust.Shared.GameObjects;
+using Robust.Shared.Timing;
 
 namespace Robust.Server.GameObjects
 {
     public interface IServerEntityNetworkManager : IEntityNetworkManager
     {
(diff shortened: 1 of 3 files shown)


metalgearsloth:
YES

metalgearsloth:
Thinking about it we shouldn't *technically* need to keep them around right? Because anytime an entity re-enters PVS view the server (at least) pretends the client's never seen it before. I imagine the client just keeps it around and doesn't recreate it which causes the problem.

Though I imagine this would save on bandwidth for sending deleted components?

mirrorcult (approved):

gradientvera:
Yeah, I don't see the point to keeping deleted components around. My solution in this PR is to just keep the NetUID of the component being deleted and the tick where this happened, and just send that to clients as needed. Given there are some entities which would never go out of PVS (looking at mobs...) I think this is gud

Issue #2130, opened 2021-10-14
IL Verification occasionally throws exceptions on Linux


Efruit:
[Most involve exceptions saying things like `Unknown PE Magic Value`](https://gist.github.com/Efruit/6e6508d31822a1e7930cf8ef528dd941) or mentioning the `COFF` header. Some mention file structure sizes are wrong. As far as I'm aware, it only impacts game clients; I've neither heard of, nor seen any instances on game severs.

On a rare occasion, [it will mention memory mapping issues](https://gist.github.com/Efruit/66d5ac7331946d6ccc8758b2fd220da8).

Attempted fixes:
- ~~[Limiting parallelism when using `ILVerify`](https://github.com/Efruit/RobustToolbox/tree/linux-ilverify-workaround)~~
- [Setting a flag on `PEReader` that may change its memory mapping behavior](https://github.com/Efruit/RobustToolbox/tree/linux-ilverify-fix-take-2) #2131

"""""Reproduction""""" steps:
- Start and stop a game's `Content.Client` and `Content.Server` repeatedly.
- Sometimes rebuilding the game causes it to happen.
- Sometimes it happens a lot of times in a row, sometimes it doesn't happen for hours.

Ygg01:
Strange I've seen some issues restarting the client fixed. But I didn't take note of exceptions.

Problem is, this happened on Windows.

PJB3005:
Logged bug on dotnet/runtime by the way: https://github.com/dotnet/runtime/issues/60545

PR #2131, opened 2021-10-14
Shared/ContentPack: Work around a possible bug in PEReader (and/or ILVerify)

Efruit:
`PEReader` (or `ILVerify`) was doing some weird `mmap`ing shit on Linux that caused evil heisenbuggy race condition exceptions during content pack verification. Apparently setting `PEStreamOptions.PrefetchEntireImage` may skip that step. If it isn't actually fixing the issue, at least it slows loading down enough to avoid the race condition.

Disclosure: Since there's no known way to deterministically reproduce the issue, I don't actually know if this fix really works. Consequently, if it does work, I have no clue why.

<details><summary>Spoiler:</summary>
The last attempt at fixing this worked perfectly until I started writing up the PR comment, so maybe this is a good way to repro?
<br/><br/>

**Update:** No luck with reproducing it, maybe the fix actually works?
</details>

If this doesn't work, I'm gating the `Parallel` IL verification parts behind `#if !LINUX` and writing sequential versions. If we get to that point, only the Lord will know if that will work either.

### DIFF
--- a/Robust.Shared/ContentPack/AssemblyTypeChecker.cs
+++ b/Robust.Shared/ContentPack/AssemblyTypeChecker.cs
@@ -809,7 +809,9 @@ namespace Robust.Shared.ContentPack
                         continue;
                     }
 
-                    return new PEReader(File.OpenRead(path));
+                    return new PEReader(File.OpenRead(path),
+                        OperatingSystem.IsLinux() ? PEStreamOptions.PrefetchEntireImage : PEStreamOptions.Default
+                    );
                 }
 
                 var extraStream = _parent.ExtraRobustLoader?.Invoke(dllName);
@@ -823,7 +825,9 @@ namespace Robust.Shared.ContentPack
                     try
                     {
                         var path = resLoadPath / dllName;
-                        return new PEReader(_parent._res.ContentFileRead(path));
+                        return new PEReader(_parent._res.ContentFileRead(path),
+                            OperatingSystem.IsLinux() ? PEStreamOptions.PrefetchEntireImage : PEStreamOptions.Default
+                        );
                     }
                     catch (FileNotFoundException)
                     {


N3X15:
[`mmap`](https://en.wikipedia.org/wiki/Mmap) is a way of accessing the file without opening the entire thing.  I believe prefetch just loads the entire damn thing into memory so that paging isn't necessary and would probably get around the problem.

Efruit:
> [`mmap`](https://en.wikipedia.org/wiki/Mmap) is a way of accessing the file without opening the entire thing. I believe prefetch just loads the entire damn thing into memory so that paging isn't necessary and would probably get around the problem.

Could be. My current hypothesis on the *how* is that there are [dotnet internals that behave slightly differently with the flag](https://github.com/dotnet/runtime/blob/d1a6945c4ab23b2d621b6e0738d9411409ab8b56/src/libraries/System.Reflection.Metadata/src/System/Reflection/PortableExecutable/PEReader.cs#L28-L33), but yours sounds equally plausible.

There are more details (and a stack trace) on #2130.

PJB3005:
PEReader *does* claim that it is thread safe in the documentation. I'm wondering if we might want to try reporting this on the .NET runtime?

Otherwise if this seems to fix it then no problem merging it for now.

CI failure is unrelated AFAICT.

Efruit:
@PJB3005

> I'm wondering if we might want to try reporting this on the .NET runtime?

Maybe? I don't know how kindly they take to "Our code does something weird, but we don't know why, and we can't make it happen on demand. Good luck!"

>Otherwise if this seems to fix it then no problem merging it for now.

Maybe it's just me, but I know I'd feel better about this if someone else could try to repro with and without the patch.

PJB3005:
I'll try a brute force stress test on my server tonight and then file a bug report on dotnet/runtime.

gradientvera:
> I'll try a brute force stress test on my server tonight and then file a bug report on dotnet/runtime.

Did you end up doing this in the end btw? @PJB3005

Efruit:
After additional testing, I was still able to repro with the patch.

In light of that, I'm closing the PR.

PR #2136, opened 2021-10-16
Added set English language checkbox

lvvova1:
Engine part

For this issue
https://github.com/space-wizards/space-station-14/issues/4712
For this PR
https://github.com/space-wizards/space-station-14/pull/4898

### DIFF
--- a/Robust.Client/Graphics/Clyde/Windowing/Glfw.Keys.cs
+++ b/Robust.Client/Graphics/Clyde/Windowing/Glfw.Keys.cs
@@ -21,7 +21,6 @@ namespace Robust.Client.Graphics.Clyde
             // TODO: to avoid having to ask the windowing thread, key names are cached.
             // This means they don't update correctly if the user switches keyboard mode. RIP.
 
-            [Dependency] private readonly IConfigurationManager _cfgg = default!;
             private readonly Dictionary<Key, string> _printableKeyNameMap = new();
 
             private void InitKeyMap()
@@ -46,9 +45,9 @@ namespace Robust.Client.Graphics.Clyde
                     if (rKey == Key.Unknown)
                         return;
 
-                    string name = "";
+                    string name;
 
-                    if (!_cfgg.GetCVar(CVars.DisplayEnglishHotkeys))
+                    if (!_clyde._cfg.GetCVar(CVars.DisplayUSQWERTYHotkeys))
                     {
                         name = GLFW.GetKeyName(key, 0);
                     }
--- a/Robust.Client/Graphics/Clyde/Windowing/Glfw.Keys.cs
+++ b/Robust.Client/Graphics/Clyde/Windowing/Glfw.Keys.cs
@@ -1,5 +1,6 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Runtime.InteropServices;
+using Robust.Shared;
 using System.Threading;
 using OpenToolkit.GraphicsLibraryFramework;
 using Robust.Client.Input;
@@ -8,6 +9,8 @@ using GlfwKey = OpenToolkit.GraphicsLibraryFramework.Keys;
 using GlfwButton = OpenToolkit.GraphicsLibraryFramework.MouseButton;
 using static Robust.Client.Input.Mouse;
 using static Robust.Client.Input.Keyboard;
+using Robust.Shared.IoC;
+using Robust.Shared.Configuration;
 
 namespace Robust.Client.Graphics.Clyde
 {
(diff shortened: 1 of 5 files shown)


mirrorcult:
@PJB3005

Robust.Shared/CVars.cs
```diff
@@ -430,6 +430,12 @@ protected CVars()
         public static readonly CVarDef<string> DisplaySplashLogo =
             CVarDef.Create("display.splash_logo", "", CVar.CLIENTONLY);
 
+        /// <summary>
+        /// Use english hotkeys.
+        /// </summary>
+        public static readonly CVarDef<bool> DisplayEnglishHotkeys =
+            CVarDef.Create("display.use_english_hotkeys", false, CVar.CLIENTONLY | CVar.ARCHIVE);
```
PJB3005: This should be referred to as "use US QWERTY hotkeys" everywhere, not "English"

Robust.Client/Input/InputManager.cs
```diff
@@ -676,6 +677,11 @@ public void SetInputCommand(BoundKeyFunction function, InputCmdHandler? cmdHandl
             }
         }
 
+        public void UpdateKeys()
+        {
+            OnKeyUpdated?.Invoke();
+        }
+
```
PJB3005: This event should be fired automatically when the input mode/CVar changes, not be raised manually by content.

Robust.Client/Graphics/Clyde/Windowing/Glfw.Keys.cs
```diff
@@ -18,10 +21,12 @@ private sealed partial class GlfwWindowingImpl
             // TODO: to avoid having to ask the windowing thread, key names are cached.
             // This means they don't update correctly if the user switches keyboard mode. RIP.
 
+            [Dependency] private readonly IConfigurationManager _cfgg = default!;
```
PJB3005: You can use `_clyde._cfg` instead so you don't have to cache it here.

Robust.Client/Graphics/Clyde/Windowing/Glfw.Keys.cs
```diff
@@ -41,8 +46,18 @@ void CacheKey(GlfwKey key)
                     if (rKey == Key.Unknown)
                         return;
 
-                    var name = GLFW.GetKeyName(key, 0);
-                    if (name != null)
+                    string name = "";
```
PJB3005:
```suggestion
                    string name;
```

Don't initialize this variable to anything since you're assigning it in both code paths.

PJB3005 (changes requested):

Robust.Client/Input/IInputManager.cs
```diff
@@ -111,6 +111,9 @@ public interface IInputManager
 
         event Action<IKeyBinding> OnKeyBindingAdded;
         event Action<IKeyBinding> OnKeyBindingRemoved;
+        event Action OnKeyUpdated;
```
PJB3005: One last thing: Change this to something like `OnInputModeChanged`

lvvova1: Done

Robust.Client/Graphics/Clyde/Windowing/Glfw.Keys.cs
lvvova1: Done

Robust.Client/Graphics/Clyde/Windowing/Glfw.Keys.cs
lvvova1: Done

Robust.Shared/CVars.cs
lvvova1: I'm made DisplayUSQWERTYHotkeys name

Robust.Client/Graphics/Clyde/Windowing/Glfw.cs
```diff
@@ -60,6 +61,7 @@ public void Shutdown()
                 if (_glfwInitialized)
                 {
                     _sawmill.Debug("Terminating GLFW.");
+                    _cfg.UnsubValueChanged(CVars.DisplayUSQWERTYHotkeys, x => InitKeyMap());
```
PJB3005: This doesn't actually work, the lambda you are unsubscribing will be different from the one you're subscribing with. You should make a separate function that receives the event if you want to unsubscribe it.

Robust.Client/Input/InputManager.cs
```diff
@@ -569,6 +574,8 @@ public void RemoveBinding(IKeyBinding binding, bool markModified = true)
             OnKeyBindingRemoved?.Invoke(binding);
         }
 
+        private void InputModeChanged(bool onValueChanged) => OnInputModeChanged?.Invoke();
```
PJB3005: This even should be invoked by Clyde instead.

lvvova1: You mean put 
`_cfg.OnValueChanged(CVars.DisplayUSQWERTYHotkeys,_inputManager.InputModeChanged);`
Into `Clyde.InitializePreWindowing`?

PJB3005: Clyde should be invoking the event on input manager from whatever code hooks the `DisplayUSQWERTYHotkeys` yes. You should not be registering a second `OnValueChanged` handler for it.

lvvova1:
So, I got problem, when I moved `OnValueChanged` into "Clyde" 

Because I had 2 `OnValueChanged` they called in wrong order, it's worked like:
> "first called "reDraw letters" after that reInit letters whos will used for reDrawing"

And I move `OnValueChanged` into Glfw where i'm reinit all letters by `OnValueChanged` also, here it's works as expected.

PJB3005:
Yep, that is exactly why you should only have a single `OnValueChanged`.

PJB3005 (approved):

PR #2140, opened 2021-10-17
nests test outputs

PaulRitter:
making the pr to see if i didn't fuck up the enviroment variable check

### DIFF
--- a/Robust.UnitTesting/Helpers.cs
+++ b/Robust.UnitTesting/Helpers.cs
@@ -20,5 +20,10 @@ namespace Robust.UnitTesting
             stream.Position = 0;
             resourceManager.MountStreamAt(stream, new ResourcePath(path));
         }
+
+        public static bool IsRunByGithubActions()
+        {
+            return Environment.GetEnvironmentVariable("GITHUB_ACTIONS") == "true";
+        }
     }
 }
--- a/Robust.UnitTesting/RobustIntegrationTest.cs
+++ b/Robust.UnitTesting/RobustIntegrationTest.cs
@@ -38,6 +38,22 @@ namespace Robust.UnitTesting
     {
         private readonly List<IntegrationInstance> _integrationInstances = new();
 
+        #region Github Actions
+        [SetUp]
+        public void GhActionsGroupStart()
+        {
+            if(Helpers.IsRunByGithubActions())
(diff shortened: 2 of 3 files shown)


PaulRitter:
why tf are tests failing man? prob need to merge master dont i?

PJB3005:
Does this even work correctly with parallelized threads? Does NUnit make sure to avoid mixing their output?

PaulRitter:
> Does this even work correctly with parallelized threads? Does NUnit make sure to avoid mixing their output?

it does not, its very janky. but parallelized threads also shit out their outputs in parallel anyways, so i just went fuck it. that being said, the main usecase for this are integrationtests that dont use parallelization anyways, r-r-right?

PJB3005:
> the main usecase for this are integrationtests that dont use parallelization anyways, r-r-right?

Integration tests are always threaded.

PaulRitter:
> > the main usecase for this are integrationtests that dont use parallelization anyways, r-r-right?
> 
> Integration tests are always threaded.

welp, i ran integration tests locally and it looked fine. only looked weird with the unittests, but those dont produce any output anyways afaict.

PaulRitter:
@PJB3005 see: https://github.com/space-wizards/RobustToolbox/runs/3918868201?check_suite_focus=true

PJB3005:
image

I don't know what to tell you but this looks extremely broken.

PaulRitter:
> image
> 
> I don't know what to tell you but this looks extremely broken.

expand the things and see that there is literally no output there. if those tests were to ever even output text, it would look just as broken.

check the integration tests, thats what this is for. you are looking at the unittests

DrSmugleaf:
Hello I'm here to ask does this work with pooling because if not I'm definitely not fixing it after the fact

PJB3005:
> check the integration tests, thats what this is for. you are looking at the unittests

Then why are empty groups being emitted for unit tests? Why doesn't the group name show the test name? 

*Why* shouldn't this be a thing with unit tests? They can also have output.

PaulRitter:
> Then why are empty groups being emitted for unit tests? Why doesn't the group name show the test name?

because that makes it infinitely hard / i havent found a way to easily add that. wyci

Acruid:
Closed for inactivity.

PR #2141, opened 2021-10-17
MIDI VolumeBoost property

20kdc:
VolumeBoost is for a potential upcoming Content PR.
~~midi.soundfont is in case you have a better soundfont.~~ EDIT: REMOVED


gradientvera:
Might want to get @PJB3005 to look at this for the soundfont part, *just in case*. Specially since content can change that at will without the user's consent.

20kdc:
Oh, wait, it can?
Oh. Right. It can.
Cherry-picking elsewhere and rebasing that commit out of existence.

20kdc:
The commit with `midi.soundfont` has been moved to `gen10-episode-29-microcosm-x-rt-backup`.

PR #2142, opened 2021-10-17
Int slider with input field

Macoron:
Add a new slider combined with input field for integer numbers
Fix minor bug with SpinBox event invoke

### DIFF
--- a/Robust.Client/UserInterface/Controls/SliderIntInput.cs
+++ b/Robust.Client/UserInterface/Controls/SliderIntInput.cs
@@ -1,0 +1,108 @@
+using System;
+
+namespace Robust.Client.UserInterface.Controls
+{
+    public class SliderIntInput : Control
+    {
+        private Slider _slider;
+        private SpinBox _spinBox;
+
+        private int _minValue = 0;
+        private int _maxValue = 100;
+        private int _value = 0;
+        private float _divisionRatio = 0.3f;
+
+        public int MinValue
+        {
+            get => _minValue;
+            set
+            {
+                _slider.MinValue = value;
+                _minValue = value;
+            }
+        }
+
+        public int MaxValue
+        {
+            get => _maxValue;
+            set
+            {
+                _slider.MaxValue = value;
+                _maxValue = value;
+            }
+        }
+
+        public int Value
+        {
+            get => _value;
+            set
+            {
+                if (_value == value)
+                    return;
+                _value = value;
+
+                _slider.Value = value;
+                _spinBox.Value = value;
+
+                OnValueChanged?.Invoke(_value);
+            }
+        }
+
+        public float DivisionRatio
+        {
+            get => _divisionRatio;
+            set => _slider.SizeFlagsStretchRatio = value;
+        }
+
+        public event Action<int>? OnValueChanged;
+
+        public SliderIntInput()
+        {
+            var hBox = new BoxContainer()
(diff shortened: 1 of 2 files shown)


Robust.Client/UserInterface/Controls/SpinBox.cs
```diff
@@ -67,6 +69,11 @@ public SpinBox()
             Value = 0;
 
             _lineEdit.IsValid = (str) => int.TryParse(str, out var i) && (IsValid == null || IsValid(i));
+            _lineEdit.OnTextChanged += (args) =>
+            {
+                if (int.TryParse(args.Text, out int i))
+                    Value = i;
+            };
```
PaulRitter, line 76: have you tested that this doesnt do le funny recursion because

change Value -> change _lineEdit.Text -> this event fires & changes Value -> start from the top

Robust.Client/UserInterface/Controls/SliderIntInput.cs
```diff
@@ -0,0 +1,108 @@
...
+                _spinBox.Value = value;
+
+                OnValueChanged?.Invoke(_value);
+            }
+        }
+
+        public float DivisionRatio
+        {
+            get => _divisionRatio;
+            set => _slider.SizeFlagsStretchRatio = value;
```
PaulRitter: gets a different value than it sets? please explain this to me

Robust.Client/UserInterface/Controls/SliderIntInput.cs
```diff
@@ -0,0 +1,108 @@
...
+namespace Robust.Client.UserInterface.Controls
+{
+    public class SliderIntInput : Control
+    {
+        private Slider _slider;
+        private SpinBox _spinBox;
+
+        private int _minValue = 0;
+        private int _maxValue = 100;
+        private int _value = 0;
```
PaulRitter: why have these when you are mirroring it to the slider anyways? just have `MinValue`, `MaxValue` and `Value` get&set the respective values of `_slider`.

PaulRitter (changes requested):

Robust.Client/UserInterface/Controls/SpinBox.cs
Macoron: Yes, it does. But Value setter checks it to be changed only once:
```
        public int Value
        {
            get => _value;
            set
            {
                if (_value == value)
                    return;
                _value = value;
```

Robust.Client/UserInterface/Controls/SliderIntInput.cs
Macoron: Yeah, it's just an error. I'll fix it

Robust.Client/UserInterface/Controls/SpinBox.cs
PaulRitter: correction: unless i've missed something, the lineedit only invokes ontextchanged on *keyevents*, but not when the actual text-value gets changed. your good, although im not sure how i feel about that behaviour, but thats not your issue :P

Robust.Client/UserInterface/Controls/SliderIntInput.cs
Macoron: I had problem with Value, because of recursion in setter. Max and Min moved to slider.

PR #2146, opened 2021-10-20
Add test pooling

DrSmugleaf:
57% time decrease to run content integration tests on GitHub CI, 210 seconds to 90 seconds on average.

- Engine integration tests are not pooled.
- Instances created depend on how many tests you can run at once, one per thread, plus the any not pooled ones and some variation depending on what tests do, and if any fail.
- How to cleanup instances is left to content, except for updating FailureLogLevel which is done in the engine.
- Moves FailureLogLevel to a cvar in RTCVars and adds cvar override enum parsing.
- There is also a nullable Pool property in instance options that can be set for each test. When null or false, tests are not pooled.
- Changes test instance threads to be background threads.
- Disposes of all the threads in a global one time teardown fixture. **Content must also implement this.**
- Adds a PreviousOptions property to test instances.
- Also changes SpriteComponent to wipe the LayerMap in AfterDeserialization, to fix client prototype manager resync.
- Server test instances now instantly remove the kicked client from their channels instead of waiting for the disconnect message to come in. This may be an issue in regular server instances as well but that's out of scope.

When tests finish running, info is printed about which clients and servers were pooled or not, like this (here truncated to reduce the wall of text):
```
Clients created (16/24):
Content.IntegrationTests.Tests.Commands.RestartRoundNowTest.RestartRoundAfterStart(True)
Content.IntegrationTests.Tests.ContainerOcclusionTest.TestAb
...


Clients pooled (22/24):
Content.IntegrationTests.Tests.Commands.RestartRoundNowTest.RestartRoundAfterStart(True)
Content.IntegrationTests.Tests.ContainerOcclusionTest.TestAb
...


Clients not pooled (2/24):
Content.IntegrationTests.Tests.Networking.SimplePredictReconcileTest.Test
Content.IntegrationTests.Tests.StartTest.TestClientStart


Servers created (48/98):
Content.IntegrationTests.Tests.Power.PowerTest
Content.IntegrationTests.Tests.Atmos.GasMixtureTest.RemoveRatio(0.25f)
...


Servers pooled (72/98):
Content.IntegrationTests.Tests.Power.PowerTest
Content.IntegrationTests.Tests.Atmos.GasMixtureTest.RemoveRatio(0.25f)
...


Servers not pooled (26/98):
Content.IntegrationTests.Tests.Commands.RestartRoundNowTest.RestartRoundAfterStart(True)
Content.IntegrationTests.Tests.ContainerOcclusionTest.TestAb
...
```

For this to be applicable to content, it too must implement a global setup fixture. Two methods are exposed on the engine's global setup fixture, Shutdown and PrintTestPoolingInfo, so that content does not need to reimplement both behaviors.


Robust.UnitTesting/RobustIntegrationTest.cs
```diff
@@ -416,51 +571,57 @@ private BaseServer Init()
...
-                            .MountString("/Prototypes/__integration_extra.yml", _options.ExtraPrototypes);
+                            .MountString("/Prototypes/__integration_extra.yml", Options.ExtraPrototypes);
                     }
                 }
 
-                cfg.OverrideConVars(new[] {("log.runtimelog", "false"), (CVars.SysWinTickPeriod.Name, "-1")});
+                cfg.OverrideConVars(new[]
+                {
+                    ("log.runtimelog", "false"),
+                    (CVars.SysWinTickPeriod.Name, "-1"),
```
PaulRitter, line 630: why this cvar? just curious

PaulRitter (approved):
lgtm

Robust.UnitTesting/RobustIntegrationTest.cs
DrSmugleaf: That was added by PJB because Windows system time apparently doesn't play well with Lidgren.

DrSmugleaf: If you mean the one below, RTCVars.FailureLogLevel, it's to be able to reset it properly and to move it out of IntegrationOptions.

PR #2148, opened 2021-10-22
Make autogenerated XAML UI fields access level settable

Visne:
Access level of XAML UI controls can now be set. If no level is set, the old behaviour is used (`protected` by default, or `private` for `sealed`).


mirrorcult (approved):

PaulRitter:
I'm sorry, but no. Encapsulation of nested XAML controls has nothing to do with OOP bad, its important for custom controls. you dont want anyone dicking around in your nested controls.

If anything this should be implemented as a field on controls in xaml, indicating if their field should be public.

Visne:
Access level can now be set manually, ready for rereview.

Robust.Client/UserInterface/Control.cs
```diff
@@ -40,6 +40,8 @@ public partial class Control : IDisposable
         [ViewVariables]
         public string? Name { get; set; }
 
+        public AccessLevel? Access { get; set; }
```
PJB3005:
```suggestion
        public AccessLevel? Access { set {  } }
```

So that no field gets allocated and this does nothing except trick Rider.

Robust.Client.NameGenerator/XamlUiPartialClassGenerator.cs
```diff
@@ -56,27 +58,42 @@ public IXamlAstNode Visit(IXamlAstNode node)
...
                             }
                         }
                     }
+
+                    if (nameText == null)
+                        return node;
+
+                    var reg = (nameText.Text,
+                        $@"{clrtype.Namespace}.{clrtype.Name}",
+                        accessText != null ? accessText.Text : "Default");
```
PJB3005:
```suggestion
                        accessText?.Text ?? "Default");
```

PJB3005 (changes requested):

Visne:
Can be reviewed again.

PJB3005 (approved):

ShadowCommander (approved):

Robust.Client/UserInterface/Control.cs
```diff
@@ -995,4 +998,11 @@ public ControlChildMovedEventArgs(Control control, int oldIndex, int newIndex)
         public readonly int OldIndex;
         public readonly int NewIndex;
     }
+
+    public enum AccessLevel
+    {
+        Private,
+        Protected,
+        Public,
+    }
```
PaulRitter: missing:
- internal
- protected internal
- private protected

Robust.Client.NameGenerator/XamlUiPartialClassGenerator.cs
```diff
@@ -118,11 +135,21 @@ public void Pop()
...
+
+                // Overwrite default
+                if (access == "Default")
+                {
+                    access = classSymbol.IsSealed ? "private" : "protected";
+                }
+
+                access = access.ToLower(CultureInfo.InvariantCulture);
+
+                return $"        {access} global::{type} {name} => this.FindControl<global::{type}>(\"{name}\");";
```
PaulRitter: two things:
- when the class is sealed, `protected` should be converted to `private`, `protected internal` should be converted to `internal` and `private protected` should be converted to `private`.
- please parse the string into the enum to validate the value & throw an exception if it fails. just throw a descriptive exception in this function, the calling function will catch it and log it as an msbuild error.

PaulRitter (changes requested):
otherwise lgtm

ShadowCommander (approved):

PaulRitter (approved):
lgtm

PR #2160, opened 2021-10-24
Fix resizing

SaphireLattice:


Robust.Client/UserInterface/CustomControls/BaseWindow.cs
```diff
@@ -130,22 +130,29 @@ protected internal override void MouseMove(GUIMouseMoveEventArgs args)
                 var (left, top) = Position;
                 var (right, bottom) = Position + SetSize;
 
+                if (Double.IsNaN(SetSize.X)) {
+                    right = Position.X + Size.X;
+                }
+                if (Double.IsNaN(SetSize.Y)) {
+                    bottom = Position.Y + Size.Y;
+                }
```
PJB3005:
```suggestion
                if (float.IsNaN(SetSize.X)) {
                    right = Position.X + Size.X;
                }
                if (float.IsNaN(SetSize.Y)) {
                    bottom = Position.Y + Size.Y;
                }
```

Use float variants instead to avoid an implicit widening cast.

PR #2161, opened 2021-10-24
Use enum for access level in XamlUiPartialClassGenerator

Visne:


PaulRitter (changes requested):
duplicating the enum is not the way to go. if anything, check if the projectreference from robust.client to the namegenerator also imports types. in short: check if you can use types&namespaces from the namegenerator project in the client. if not, strings is actually the cleanest thing to do here imo

PR #2165, opened 2021-10-25
Improves PlacementManager (construction/etc.) handling of parent world rotation

20kdc:
Note: This *does not* fully handle every possible rotation situation.

In practice, the specific thing that breaks is noRot sprites in space while the eye has been rotated (standing on rotated grid), but those were already broken in one way or another as far as I can tell.

This still fixes the rotation of the common cases, which are:
+ tile placement indicators
+ any noRot sprite on the same grid as you

It doesn't fix the rotation of rotatable sprites with no sprite directions (read: rotated item sprites), but it shouldn't introduce a bug with it either.

The hypothetical ideal of course would be that more data would be provided to the placement overlay renderer to handle all of this, but in practice this is more than enough right now, and it's a basis for further work.

### DIFF
--- a/Robust.Client/Placement/PlacementMode.cs
+++ b/Robust.Client/Placement/PlacementMode.cs
@@ -114,13 +114,19 @@ namespace Robust.Client.Placement
             var size = TexturesToDraw[0].Size;
             foreach (var coordinate in locationcollection)
             {
+                if (!coordinate.IsValid(pManager.EntityManager))
+                    return; // Just some paranoia just in case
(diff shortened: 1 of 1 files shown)


20kdc:
Rebased and force-pushed just to check compatibility. AOK.

20kdc:
Rebasing in preparation for *possible* further work that may not actually occur or actually work.

20kdc:
oh
um
wow, that was terrible timing

PR #2168, opened 2021-10-26
Fix placement manager on rotated grid

metalgearsloth:
Specifically the ghost entity should always match what gets placed and it should align with cursor properly.

### DIFF
--- a/Robust.Client/Placement/PlacementManager.cs
+++ b/Robust.Client/Placement/PlacementManager.cs
@@ -709,8 +709,7 @@ namespace Robust.Client.Placement
 
             // world x and y
             message.EntityCoordinates = coordinates;
-
-            message.DirRcv = Direction;
+            message.DirRcv = Direction.ToAngle();
 
             NetworkManager.ClientSendMessage(message);
         }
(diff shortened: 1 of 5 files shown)


20kdc:
uuuuuh
conceptual duplicate of https://github.com/space-wizards/RobustToolbox/pull/2165 ???
also, the correct rotation actually may not actually match the ghost entity (cross-grid placement)

20kdc:
ok, so I did some poking, and...
The following image is from this PR:
image
The problem appears to be that the direction angle is being applied, which only works for placements that actually rotate that way, i.e. not noRot.
As I mentioned in my PR this is a matter of there not being enough information to determine how to treat the sprites (i.e. no way to determine if they are/are not noRot).
I defaulted to not respecting rotation this way because sprites that are noRot generally care particularly about their rotation, while sprites that aren't look generally okay in any orientation, and the bug with the rotation of those has been around since forever anyway and has never gotten in the way of anything.

metalgearsloth:
Woops, I'll suss it out.

Also are you talking about noRot sprites or noRot transforms because noRot sprite should be irrelevant.

metalgearsloth:
Why wouldn't we just fix noRot? I don't see how because it was bad before it should keep being bad.

Not being able to actually see rotation when mapping sounds bad ngl.

Like I'm fine with a bandaid but I'd still consider it broken in either PR.

20kdc:
Because actually fixing things *properly* requires using SpriteComponent rather than having an entirely separate rendering scheme specifically for this purpose.

PR #2171, opened 2021-10-27
Fix all warnings

Visne:
Did some cleanup while I was there.


PaulRitter:
brave touching the dynamictree

Visne:
If I did everything correctly, which might not be true, that stuff is all unused

PJB3005:
I'm still gonna wait on a review from @metalgearsloth just in case.

metalgearsloth:
Okay I double-checked and it was used for the old movebuffer like I thought (UpdatePairs). This got nuked because I had to rewrite it to make shuttles work so it's okay to remove, I just didn't cull it at the time.

PR #2173, opened 2021-10-28
Fix PVS crash

### LINKED ISSUE #2135: Transform exception from AdminObserver
There's some sequence of events in SS14 where (presumably) an admin holding an item leads to an exception because a client receives the entity state but has no visibility of the parent (AdminObserver).

### LINKED ISSUE #4995: Cursed locker doesn't always correctly sync
## Description
Sometimes it fails to inform the client the entity is gone, resulting in a desync and immense lag.

metalgearsloth:
Fixes https://github.com/space-wizards/RobustToolbox/issues/2135
Might fix https://github.com/space-wizards/space-station-14/issues/4995 but no stacktrace

The issue is that when the client gets sent a NaN state it assumes the client knows about the new parent, but because it never knows about an aghost this throws.

This will lead to additional re-parent events on the client but that's what you get for using NaN states Q.

### DIFF
--- a/Robust.Server/GameStates/EntityViewCulling.cs
+++ b/Robust.Server/GameStates/EntityViewCulling.cs
@@ -460,6 +460,7 @@ namespace Robust.Server.GameStates
                 // PVS leave message
                 //TODO: Remove NaN as the signal to leave PVS
                 var oldState = (TransformComponent.TransformComponentState) xform.GetComponentState(session);
+                var mapEnt = _mapManager.GetMapEntityId(xform.MapID);
 
                 entityStates.Add(new EntityState(entityUid,
                     new[]
@@ -467,7 +468,7 @@ namespace Robust.Server.GameStates
                         ComponentChange.Changed(_transformNetId,
                             new TransformComponent.TransformComponentState(Vector2NaN,
                                 oldState.Rotation,
-                                oldState.ParentID,
+                                mapEnt,
                                 oldState.NoLocalRotation,
                                 oldState.Anchored)),
                     }));


gradientvera:
NaNs and their consequences have been a disaster for the codebase

gradientvera (approved):

PaulRitter:
> NaNs and their consequences have been a disaster for the codebase

cant wait for the eventual day i get pissed off too much an rip it out of pvs

Tyler-IN:
Having the client miss information it needs definitely should throw. Gotta either get that info to the client or have the client deal without it. Having NaNs throw helped detect problems during the initial development of PVS.

It was never intended to be a permanent solution, it was a hack that needed to be done in a short timeframe and as a proof of concept; telling the client "you can't know this position anymore" needs it's own state, like it or hate it, that's what the Vector2NaN is. I'd have expected that additional state and messaging added by now, idk why you guys haven't. Good luck and here's hoping you figure it out. It's not super complicated.


Running into the NaNs just means the PVS implementation is incomplete though, so even if you use a different mechanism to convey 'unknowable position' you need to handle it anyway.

metalgearsloth:
> I'd have expected that additional state and messaging added by now, idk why you guys haven't.

Q I think the only piece of your code left these days is the NaN.

Tyler-IN:
Yeah looks like 99% of my code has someone else's name on it _if you use blame_, been shuffled around the codebase into different files.
https://github.com/space-wizards/RobustToolbox/commits?author=Tyler-IN


Running into the NaNs just means the PVS implementation is incomplete though, so even if you use a different mechanism to convey 'unknowable position' you need to handle it anyway. I think the next step was to have the clients handle NaN updates specially and leave the previous position or whatever instead of assigning it.

Tyler-IN:
The catch was that the events that happen between some _in-PVS item exiting and getting NaN'd_ and that item _re-entering PVS on the server and being conveyed to the client_ needed a mechanism to not 'ghost' on the client when the PVS moves over where the item _no longer is_ on the client only. It technically didn't move but the client can't assume it is actually in the same place unless the server says so - needs to either hide the entity (which is what the NaN hack does) or give it said special state flag that can do the same thing while preserving its assumed position. Have to account for audio (and other stuff) being emitted from entity instead of from a specific position too, as they can't be positioned correctly.

Tyler-IN:
> Sometimes it fails to inform the client the entity is gone, resulting in a desync and immense lag.

Deletes should be sent to clients outside of PVS, no?

PR #2174, opened 2021-10-28
Dirty changes

metalgearsloth:
2 changes:
1. IEntity.Dirty() moved to EntityManager. This required LastModifiedTick setter to be public which is a bit suss and also means a GetEntity needs to be done on dirty.
2. Rather than doing the anchoring thing directly in Dirty this is done via the eventbus. This is mainly setup for future PVS optimisations I want to try.

Needs https://github.com/space-wizards/space-station-14/pull/5069 merged first


PaulRitter (approved):
pr might be dirty but code lgtm

20kdc:
Thought up an idea about the set accessibility: What about `internal set`?

PaulRitter:
> Thought up an idea about the set accessibility: What about `internal set`?

i dig it. ten times better than having it be public

PR #2176, opened 2021-10-29
Add tab "completion" for the server-side C# console

Efruit:
Screenshot from 2021-10-29 08-48-21

Note that it's not *exactly* like normal completion, as it suggests everything the current token _could be_.
i.e. `Robust.<Tab>` and `Robust.Ser<Tab>` will both produce `Shared` and `Server`. I assume that normal editors take the raw suggestion lists and filter them to things that only have some common subset with the current token, but this does not. That being said, it's unlikely that it would be difficult to add if so desired.

You can click on the entries to insert them at the cursor position. The above caveat still applies.

The functionality has not yet been ported to the client-side C# console.

### DIFF
--- a/Robust.Client/Console/Completions.cs
+++ b/Robust.Client/Console/Completions.cs
@@ -37,10 +37,15 @@ namespace Robust.Client.Console
             ContentsContainer.AddChild(_suggestPanel);
         }
 
+        private bool _firstopen = true;
         public void OpenAt(Vector2 position, Vector2 size)
         {
-            SetSize = size;
-            LayoutContainer.SetPosition(this, position);
+            if (_firstopen)
+            {
+                SetSize = size;
+                LayoutContainer.SetPosition(this, position);
+                _firstopen = false;
+            }
             Open();
         }
 
--- a/Robust.Client/Console/Completions.cs
+++ b/Robust.Client/Console/Completions.cs
@@ -1,0 +1,124 @@
+using System;
+using System.Collections.Immutable;
+using Robust.Client.UserInterface;
+using Robust.Client.UserInterface.Controls;
+using Robust.Client.UserInterface.CustomControls;
+using Robust.Shared.Input;
+using Robust.Shared.Maths;
+using Robust.Shared.Network.Messages;
+using Robust.Shared.Utility;
+using Robust.Shared.Scripting;
+using static Robust.Client.UserInterface.Controls.BoxContainer;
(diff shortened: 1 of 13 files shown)


Robust.Shared/Robust.Shared.csproj
```diff
@@ -19,6 +19,9 @@
     <PackageReference Include="Serilog" Version="2.10.0" />
     <PackageReference Include="YamlDotNet" Version="9.1.4" />
     <PackageReference Include="System.Management" Version="5.0.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Features" Version="3.8.0" />
```
PJB3005: Absolutely do not add Roslyn to shared as a dependency. This is what `Robust.Shared.Scripting` is for.

Robust.Client/Console/Completions.cs
```diff
@@ -0,0 +1,124 @@
...
+using Robust.Shared.Maths;
+using Robust.Shared.Network.Messages;
+using Robust.Shared.Utility;
+using Robust.Shared.Scripting;
+using static Robust.Client.UserInterface.Controls.BoxContainer;
+using static Robust.Shared.Network.Messages.MsgScriptCompletionResponse;
+
+namespace Robust.Client.Console
+{
+    public class Completions : SS14Window
```
PJB3005, line 16: This could probably do with using XAML instead.

PJB3005: Also use a `Popup` for this instead so you can sanely place it.

Efruit: > This could probably do with using XAML instead.

I'll do that once there's documentation on how to do it.

> Also use a `Popup` for this instead so you can sanely place it.

Tried it, it was a pain. The mock-docked window gets it 95% right with 75% of the code.

Efruit: Frankly I prefer it being resizable and movable, so I've changed it to only mock-dock on the first open. After that, it will remember its size/position like a normal window.

Efruit:
There is (at least) one bug that I'm aware of, but haven't been able to fix: Script globals aren't represented properly in completions. It's not a showstopper, but it is inconvenient.

I've tried adding variations of `scriptClassName: nameof(ScriptGlobalsImpl)` to the `CSharpCompileOptions` to no avail.

PJB3005:
Better than nothing, I guess.

PR #2181, opened 2021-10-30
Bandaid sprite animation in SpriteView

Macoron:
See [this issue](https://github.com/space-wizards/space-station-14/issues/5104)
If anyone has better idea how to fix it, would glad to get feedback

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
@@ -17,6 +17,7 @@ namespace Robust.Client.GameObjects
         [Dependency] private readonly RenderingTreeSystem _treeSystem = default!;
 
         private readonly Queue<SpriteComponent> _inertUpdateQueue = new();
+        private HashSet<ISpriteComponent> _manualUpdate = new();
 
         public override void Initialize()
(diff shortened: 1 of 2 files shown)


Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
```diff
@@ -57,10 +64,21 @@ public override void FrameUpdate(float frameTime)
                         return true;
                     }
 
-                    value.FrameUpdate(state);
+                    if (!_manualUpdate.Contains(value))
+                        value.FrameUpdate(state);
```
metalgearsloth, line 68: Is this ever true?

metalgearsloth (commented):
Not having to check ContainerOccluded in ProcessSpriteEntities is pretty useful but now we get the Contains check for every visualised sprite which kinda negates it, though we also don't need to worry about SpriteView being in the world viewport or not.

Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
Macoron: Yeah. Examine window also use SpriteView, so you get update function called twice

PR #2184, opened 2021-10-31
Placement Manager: Spin Cycle 2

20kdc:
*It works with almost everything.*

image

EDIT:
Things tested:

+ Placement of tiles? Anywhere anytime.
+ Placement of noRot objects? Yup!
+ Placement of rotatable objects? You betcha!
+ On other grids? Indeed!


20kdc:
Ok, so, I might need to put more of an explaination as to what this commit actually does

20kdc:
Explaination:

So as I said previously to @metalgearsloth , the only way to make the rendering completely accurate would be if it were to use SpriteComponent
Which of course begs the question: Why not use SpriteComponent?
So I did, and this happened.
By using SpriteComponent I managed to get rid of every single bug with the placement overlay and rotation, I believe.
The catch is *how to get the data*, that's where the true horrors lie.

metalgearsloth (approved):

PR #2188, opened 2021-11-01
Make client console check perms locally

metalgearsloth:
I nuked my admin perms locally and tested chat works and togglefov / togglelight don't work.

Content PR is https://github.com/space-wizards/space-station-14/pull/5124


PJB3005 (changes requested):
Checking this means that you can't run any client-side commands (because those aren't understood by the con group controller)

metalgearsloth:
Do you mean engine-wise without SS14? I checked with and without admin perms and it *seemed* to work as intended (the content test failure is because it doesn't have "disconnect" allowed which has a PR up).

PJB3005:
I ran the SS14 client without connecting to a server and I couldn't use any command.

moonheart08:
> I ran the SS14 client without connecting to a server and I couldn't use any command.

It wouldn't be a very effective fix if you could simply `togglelight` before connecting

PJB3005:
I should, however, be capable of running `devwindow` offline to test UI stuff.

If running `togglelight` is a problem then we can have content reset it on connect. Or the engine.

metalgearsloth:
Content needs a way to reset commands anyway because people leave stuff on after sandbox rounds are over so if the client isn't InGame we should probably just skip the check?

metalgearsloth:
gucci 100 for check

Robust.Client/Console/ClientConsoleHost.cs
```diff
@@ -103,6 +108,14 @@ public override void ExecuteCommand(ICommonSession? session, string command)
 
             if (AvailableCommands.ContainsKey(commandName))
             {
+                var playerManager = IoCManager.Resolve<IPlayerManager>();
+
+                if (!_conGroup.CanCommand(commandName) && playerManager.LocalPlayer?.Session.Status > SessionStatus.Connecting)
```
PJB3005, line 113: `CanCommand()` defaults to returning false so this will break console commands on repos that doesn't implement `IClientConGroupImplementation` still.

PaulRitter: wouldnt true be a more sensible default?

PJB3005: Probably, yes.

PaulRitter (approved):

PR #2193, opened 2021-11-02
Make transform delete entity if parent invalid

metalgearsloth:
These days this pretty much only happens if PVS didn't send the parent (and we know of a circumstance in which it doesn't where the new parent for the NaN state has never been seen). Right now this just floods the log with errors every tick and hammers client performance so this hopefully handle those situations well enough that the game is playable.

I playtested the 2 situations I knew about that cause the PVS issue: Admin picking up item and using the cursed locker.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -764,11 +764,10 @@ namespace Robust.Shared.GameObjects
                         {
                             if (!Owner.EntityManager.TryGetEntity(newParentId, out var newParent))
                             {
-                                var msg = $"Unable to find new parent {newParentId}! Deleting {Owner}";
 #if !EXCEPTION_TOLERANCE
-                                throw new InvalidOperationException(msg);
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -761,7 +762,17 @@ public override void HandleComponentState(ComponentState? curState, ComponentSta
...
                         {
-                            var newParent = Owner.EntityManager.GetEntity(newParentId);
+                            if (!Owner.EntityManager.TryGetEntity(newParentId, out var newParent))
+                            {
+#if !EXCEPTION_TOLERANCE
+                                throw new InvalidOperationException($"Unable to find new parent {newParentId}! This probably means the server never sent it.");
+#else
+                                Logger.ErrorS("transform", $"Unable to find new parent {newParentId}! Deleting {Owner}");
+#endif
+                                Owner.QueueDelete();
```
gradientvera, line 771: honestly, just put this under the else so the compiler doesn't complain

PR #2199, opened 2021-11-05
Adds collapsible widgets

vulppine:
Adds collapsible widgets. These are a framework widget that require that the children only consist of two elements when constructed in XAML or through directly specifying children while constructing it.

The two elements must be something that derives from BaseButton in toggle mode for the first one, and any type of control for the second one. When the first element is toggled, the second element will be hidden.

Also includes a widget that is practically a clickable label with a chevron in front of it (`CollapsibleHeader`), and a widget that is a container that is autohidden (`CollapsibleBody`)


Robust.Client/UserInterface/Controls/Collapsible.cs
```diff
@@ -0,0 +1,175 @@
...
+            foreach (var control in current)
+            {
+                if (control == _box) continue;
+                ReparentToBox(control);
+            }
+
+            _initialized = true;
+        }
+
+        private void ReparentToBox<T>(T control) where T : Control
```
PJB3005: You don't have to make this a generic, it doesn't do anything.

PJB3005 (approved):
I don't personally like the rigid approach with the initialization but it's still something I need to come up with a solution to so it's fine for now.

vulppine:
yeah, the initialize bit in the Collapsible itself is just a dirty solution (but it does *work*), but i still need time to work on CollapsibleHeader so there's at least something that resembles that UI pattern without it looking so terrible (it currently reparents during draw because i haven't figured out a better way to draw the chevron without just pulling code from BoxContainer), but once i get it working i'll make another comment here

vulppine:
alright, should be good to merge (if the tests pass as of time of writing)

PR #2200, opened 2021-11-06
Add function to directly perform lookups using an `EntityLookupComponent`

ElectroJr:
This adds a variant of `EntityLookup.FastEntitiesIntersecting` that directly uses a given `EntityLookupComponent` instead of calling `GetLookupsIntersecting()`. Useful if you need to do a large number of lookups on a known grid w/o having to do the whole grid-intersections calculation in `GetLookupsIntersecting` for each tile.

Alternatively, if `EntityLookupComponent.Tree` were public instead of internal I could just query that directly, but I assume setting that to internal was intentional to stop content monkeys like myself from somehow messing up the tree by directly modifying it.

Opening this PR as it would be nice to have for a content PR that I'll open a draft of sometime today or tomorrow. Otherwise, I don't really know if there is actually a need for this, especially if QueurySystem or whatever comes along.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/EntityLookup.cs
+++ b/Robust.Shared/GameObjects/Systems/EntityLookup.cs
@@ -398,7 +398,6 @@ namespace Robust.Shared.GameObjects
         }
 
         /// <inheritdoc />
-        [MethodImpl(MethodImplOptions.AggressiveOptimization)]
         public void FastEntitiesIntersecting(EntityLookupComponent lookup, ref Box2 localAABB, EntityQueryCallback callback)
         {
             lookup.Tree._b2Tree.FastQuery(ref localAABB, (ref IEntity data) => callback(data));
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Systems/EntityLookup.cs
```diff
@@ -395,6 +397,13 @@ public void FastEntitiesIntersecting(in MapId mapId, ref Box2 worldAABB, EntityQ
             }
         }
 
+        /// <inheritdoc />
+        [MethodImpl(MethodImplOptions.AggressiveOptimization)]
```
ElectroJr: I have no idea how `MethodImpl` actually works and whether this even has any impact on a function this simple. I just added it cause the other `FastEntitiesIntersecting` has it.

metalgearsloth: It's specifically designed for simple functions. I don't know the specifics but I know it will try harder to inline if possible.

It can make debugging more painful and also lead to slower code if used carelessly so I don't use it unless I've profiled

ElectroJr: Well I have no idea how to properly do profiling, so I guess I'll just remove it for now.

PR #2203, opened 2021-11-07
Shared/GameObjects: Apply grid's rotational velocity when leaving

Efruit:
Caveat: Does not work for controlled mobs.

Many thanks to @ElectroJr and @Tomeno for correcting my atrocious math
...repeatedly.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedGridTraversalSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedGridTraversalSystem.cs
@@ -1,4 +1,6 @@
 using System.Collections.Generic;
+using Robust.Shared.Utility;
+using Robust.Shared.Maths;
 using Robust.Shared.Containers;
 using Robust.Shared.IoC;
 using Robust.Shared.Map;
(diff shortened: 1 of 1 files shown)


Efruit:
(100% fine if you guys don't want to upstream this, but it's still good to have it posted somewhere in case someone wants to refer to it later)

metalgearsloth:
Is there any reason this can't be on parent change? This also seems to ignore the linearvelocity being set on parent change.

Also only applying it when leaving a grid but not entering seems too game specific to put on robust.

Efruit:
> Is there any reason this can't be on parent change?

Not that I know of.

> This also seems to ignore the linearvelocity being set on parent change.

I don't follow.

> Also only applying it when leaving a grid but not entering seems too game specific to put on robust.

```
   EFR | quick q: if we were preserving (angular) momentum when switching grids, would 
       | that go in content or engine?                                                 
   EFR | I can imagine downstreams where the current behavior is more or less preferred
 MoMMI | <m​etalgearsloth> engine                                                       
```
<p align="center"></p>

metalgearsloth:
Don't braindent me m8, I just didn't write out "preserving velocities on parent change should be global and an engine thing but if you want to do the funny on grid change it should be content".

SharedPhysicsSystem subscribes to parent changes and re-applies its velocity in terms of the new parent which is probably interacting with this in some regard.

Efruit:
<b>Nothing to see here. Move along. >>/#2208/</b>

PR #2208, opened 2021-11-08
SharedPhysicsSystem: convert parents' angular velocity to linear when leaving

Efruit:
As with last time, thanks to @ElectroJr and @Tomeno for whacking some math in to me.

<p align="center"><b>Caveat: does not perform the reverse.</b></p>

rotation fun

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -148,28 +148,35 @@ namespace Robust.Shared.GameObjects
 
             var oldParent = args.OldParent;
             var linearVelocityDiff = Vector2.Zero;
-            var angularVelocityDiff = 0f;
 
             if (oldParent != null && oldParent.TryGetComponent(out PhysicsComponent? oldBody))
             {
                 var (linear, angular) = oldBody.MapVelocities;
 
-                linearVelocityDiff += linear;
-                angularVelocityDiff += angular;
+                // Our rotation system is backwards; invert the angular velocity
+                var o = -angular;
+
+                // Get the vector between the parent and the entity leaving
+                var r = oldParent.Transform.WorldPosition - entity.Transform.WorldPosition;
+
+                // Get the tangent of r by rotating it π/2 rad (90°)
+                var v = new Angle(MathHelper.PiOver2).RotateVec(r);
+
+                // Scale the new vector by the angular velocity
+                v *= o;
+
+                linearVelocityDiff += linear + v;
             }
 
             var newParent = entity.Transform.Parent?.Owner;
 
             if (newParent != null && newParent.TryGetComponent(out PhysicsComponent? newBody))
             {
-                var (linear, angular) = newBody.MapVelocities;
-
-                linearVelocityDiff -= linear;
-                angularVelocityDiff -= angular;
+                // TODO: Apply child's linear as angular on the parent?
+                linearVelocityDiff -= newBody.MapLinearVelocity;
             }
 
             body.LinearVelocity += linearVelocityDiff;
-            body.AngularVelocity += angularVelocityDiff;
         }
 
         private void HandleGridInit(GridInitializeEvent ev)
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -177,8 +177,8 @@ namespace Robust.Shared.GameObjects
                 //
                 // https://cwzx.wordpress.com/2014/03/25/the-dynamics-of-a-transform-hierarchy/
                 var w = new Matrix3(
-                    0, oldBody.MapAngularVelocity, 0,
-                    -oldBody.MapAngularVelocity, 0, 0,
+                    0, angular, 0,
+                    -angular, 0, 0,
                     0, 0, 0
                 );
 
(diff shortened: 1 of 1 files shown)


PJB3005 (approved):
Looks good to me but @metalgearsloth please double check.

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -148,28 +148,35 @@ private void HandleParentChange(ref EntParentChangedMessage args)
...
-                linearVelocityDiff += linear;
-                angularVelocityDiff += angular;
+                // Our rotation system is backwards; invert the angular velocity
+                var o = -angular;
+
+                // Get the vector between the parent and the entity leaving
+                var r = oldParent.Transform.WorldPosition - entity.Transform.WorldPosition;
+
+                // Get the tangent of r by rotating it π/2 rad (90°)
+                var v = new Angle(MathHelper.PiOver2).RotateVec(r);
```
metalgearsloth: I'm not a maths guy but:
- You probably want centre of mass not origin (LocalCenter is the property and needs to be converted to a world position)
- This is just a world angle?

Efruit: - Yeah, COM would be preferred. Can't test it yet (t. no .NET 6), but it should be a quick fix.
- I'm not a maths guy either; your guess is as good as mine (probably significantly better).

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -148,28 +148,35 @@ private void HandleParentChange(ref EntParentChangedMessage args)
...
             {
                 var (linear, angular) = oldBody.MapVelocities;
 
-                linearVelocityDiff += linear;
-                angularVelocityDiff += angular;
+                // Our rotation system is backwards; invert the angular velocity
+                var o = -angular;
+
+                // Get the vector between the parent and the entity leaving
+                var r = (oldParent.Transform.WorldPosition + oldBody.LocalCenter) - entity.Transform.WorldPosition;
```
metalgearsloth: Will need to consider parent's rotation, i.e. if the parent is rotated 180 degrees the actual spot is different and you can't just add the vectors together. You can probably just use the worldmatrix of oldParent to transform the localcenter. 

Also might need to consider the entity's centre of mass?

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -147,25 +147,68 @@ private void HandleParentChange(ref EntParentChangedMessage args)
...
 
             var newParent = entity.Transform.Parent?.Owner;
 
             if (newParent != null && newParent.TryGetComponent(out PhysicsComponent? newBody))
             {
-                var (linear, angular) = newBody.MapVelocities;
-
-                linearVelocityDiff -= linear;
-                angularVelocityDiff -= angular;
+                // TODO: Apply child's linear as angular on the parent?
```
metalgearsloth, line 193: nope

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -147,25 +147,68 @@ private void HandleParentChange(ref EntParentChangedMessage args)
...
                 var (linear, angular) = oldBody.MapVelocities;
 
-                linearVelocityDiff += linear;
-                angularVelocityDiff += angular;
+                // Our rotation system is backwards; invert the angular velocity
+                var o = -angular;
+
+                // Get the vector between the parent and the entity leaving
+                var r = oldParent.Transform.WorldMatrix.Transform(oldBody.LocalCenter) -
+                    entity.Transform.WorldPosition; // TODO: Use entity's LocalCenter/center of mass somehow
```
metalgearsloth: Cache these into local variables where possible because they're expensive please.

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -147,25 +147,68 @@ private void HandleParentChange(ref EntParentChangedMessage args)
...
+                v *= o;
+
+                var R = Matrix3.CreateRotation(entity.Transform.WorldRotation);
+                var w = new Matrix3(
+                    0, newBody.MapAngularVelocity, 0,
+                    -newBody.MapAngularVelocity, 0, 0,
+                    0, 0, 0
+                );
+                R.Transpose(out var nRT);
+                nRT.Multiply(-1f);
```
metalgearsloth: same as above, please cache because this method will be very expensive.

metalgearsloth (changes requested):

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
Efruit: well it was a nice thought

metalgearsloth (changes requested):
Needs more caching, also use the new GetWorldPositionRotation method.

I tried to just push the changes but it was blocked so

Efruit:
> Needs more caching,

That's it. There's nothing left to cache.

> also use the new GetWorldPositionRotation method.

The one that this PR predates? Guess that gives me a rebase pass.

metalgearsloth (approved):
image

FYI it's more the fact it halves the GetComponent calls being done which are the most expensive part of calculating world stuff.

PR #2210, opened 2021-11-09
Add overlay for physics COM

metalgearsloth:
(Center of mass)

Will be useful whenever I get around to fixing it (tried in the past but was getting weird collision resolution).


gradientvera:
Nice, screenshots?

metalgearsloth:
image

(Saltern COM is 0,0 because COM is still broken)

metalgearsloth:
(Also drawing circles is slow as so this hammered my framerate)

PJB3005:
Damn I didn't know we have physics Component Object Model.

Issue #2101, opened 2021-10-07
Multiline text input UI control


gradientvera:
Having an UI control like `LineEdit` but for an arbitrary amount of lines would be very convenient for many things.

PaulRitter:
have some limit tho, because of that "crash by massive unicode char spam" issue

Efruit:
**I REFUSE TO BE BLOCKED BY THIS ANY LONGER**

![1499217568353](https://user-images.githubusercontent.com/602406/140941803-2843a3b6-9450-48fe-9a88-8fcc97bfb3a9.jpg)

gradientvera:
Implemented in #3436

PR #2213, opened 2021-11-09
Rich text redux

Efruit:
The preparation for Books and #2101 begins.


Efruit:
The content side is on [my fork's branch of the same name](https://github.com/Efruit/SS14/tree/rich-text-redux). Not gonna bother PRing it (yet) since this one still needs to be torn to shreds first.

Efruit:
Content side at https://github.com/space-wizards/space-station-14/pull/5625

Efruit:
Future Improvements:
- `Section.Meta` flag for color not being set
- `ISectionable` that runs `Loc.GetString` on sections with `Section.Meta`'s `Localized` flag (requires extra metadata for fluent args)
- Clickable links (also requires extra metadata)
- Reduce the number of `FormattedMessage.Builder` call sites (constructing `Section`s really is NOT hard)

Robust.Shared/Utility/CollectionExtensions.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Robust.Shared.Utility
 {
-    public static class Extensions
+    public static partial class Extensions
```
Acruid, line 9: Why was this turned into a partial class?

Acruid (changes requested):
All of the obsolete 'FormattedMessage.Builder' warnings need to be fixed.

Robust.Shared/Utility/CollectionExtensions.cs
Efruit: [strictly for personal reasons](https://github.com/space-wizards/RobustToolbox/pull/2213/commits/5b70986c5cb8d226cfcfbe8e76bf8aa0bc9683e9)

(Because having `Robust.Shared.Utility.Extensions` and `Robust.Shared.Utility.TextExtensions` seemed messy, and I didn't want to rename the existing `Extensions` to `CollectionExtensions`.)

Efruit:
> All of the obsolete 'FormattedMessage.Builder' warnings need to be fixed.

It'll have to be an ongoing process. `Builder` was only added as a stopgap to make the transition easier. It's a warning because it should NOT be used in new code. After `Content` has caught up, and no longer needs it, we can make it `internal` and nix the warning.

Efruit:
To expand a bit on that, `Builder` is *not*, in fact, obsolete. It has sanctioned (engine-internal) uses within markup parsers, since they need something that does exactly what `Builder` does.

Content, on the other hand, should either be putting `Section`s together manually (it's not hard), or using a markup parser.

But I haven't been able to find the `[PleaseDontUseThisExceptInThesePlaces]` attribute, so I went with `[Obsolete]` instead.

Acruid (approved):

PR #2218, opened 2021-11-11
Arbitrary occluder rotation

20kdc:
Stuff the PR does as of now:

+ Occluders may be locally rotated without visual glitching, `fixoccluders` is now only necessary because inconsistency is bad.

Things To Discuss:

+ Arbitrary polygonal occluder meshes?
+ If so, how far to push the API for such?
+ Face elimination in vertex shader?


20kdc:
I forgot to mark as draft, but that's "ok" - this does work as-is

20kdc:
Consider this ready - the API changes and so on and so forth can happen later, I have a few too many PRs that keep needing rebasing to actually do further work and then somehow get that further work through review here

Acruid (approved):

PR #2220, opened 2021-11-11
Move physics fixtures to its own component

metalgearsloth:
Downsides:
- Makes physics more of a pain to work with in terms of API
- Makes tracking the component alongside physics comp a pain
- I needed to touch engine tests again

Upsides:
- We can move fixtures out of the physics compstate and significantly cut down on bandwidth when grids move
- Decent performance increase from caching some properties and having a significantly smaller state
- I inadvertantly found and fixed a bug with fixture mass on clients so pushing locker chains should be much more stable.

Content PR:
https://github.com/space-wizards/space-station-14/pull/5524


Robust.Shared/Robust.Shared.csproj
```diff
@@ -47,6 +47,7 @@
     <Compile Update="Serialization\RobustMappedStringSerializer.MappedStringDict.cs">
       <DependentUpon>RobustMappedStringSerializer.cs</DependentUpon>
     </Compile>
+    <Compile Remove="Physics\ShapeManager.cs" />
```
metalgearsloth: FFS Rider
```suggestion
```

PR #2222, opened 2021-11-12
Add JsonConverter for EntityUids

DrSmugleaf:
For admin logging.

### DIFF
--- a/Robust.Shared/GameObjects/EntityUid.cs
+++ b/Robust.Shared/GameObjects/EntityUid.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Text.Json.Serialization;
 using JetBrains.Annotations;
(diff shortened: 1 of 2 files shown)


PJB3005:
> For admin logging.

This isn't really useful for admin logging though? Just dumping an entity UID will be completely useless in any sort of logs. You'll need at *least* a dump of data including entity name, type, character name (if any) and controlling client (if any).

DrSmugleaf:
> > For admin logging.
> 
> This isn't really useful for admin logging though? Just dumping an entity UID will be completely useless in any sort of logs. You'll need at _least_ a dump of data including entity name, type, character name (if any) and controlling client (if any).

That's what I'm doing now, this was for before when I was storing players separately for each log entry.

PR #2224, opened 2021-11-13
Optimise physics broadphase cache

metalgearsloth:
Roughly 10% gains on packed by not having to iterate every fixture to get the WorldAABB as LocalBounds is already stored.


Robust.Shared/Physics/SharedBroadphaseSystem.cs
```diff
@@ -170,16 +172,26 @@ private void UpdateBroadphaseCache(BroadphaseComponent broadphase)
         {
             var uid = broadphase.OwnerUid;
 
-            var transform = _physicsManager.UpdateTransform(uid);
+            var xformComp = EntityManager.GetComponent<TransformComponent>(uid);
+
+            var matrix = xformComp.WorldMatrix;
+            var worldPosition = new Vector2(matrix.R0C2, matrix.R1C2);
+            var transform = new Transform(worldPosition, xformComp.WorldRotation);
```
metalgearsloth, line 179: When the combined worldpos + worldrot method is merged I'll update this in that PR

PR #2225, opened 2021-11-13
Automatic Client Zipping

20kdc:
This is designed to make it a lot easier to host an SS14 fork.
As of the engine versioning stuff, you don't need to fill out any fields anymore.
At all.

Build with `-c Release`, make sure the config doesn't get overwritten, and your station can be connected to from the launcher!

### DIFF
--- a/Robust.Server/ServerStatus/IStatusHandlerContext.cs
+++ b/Robust.Server/ServerStatus/IStatusHandlerContext.cs
@@ -28,6 +28,16 @@ namespace Robust.Server.ServerStatus
             int code = 200,
             string contentType = "text/plain");
 
+        void Respond(
+            byte[] data,
+            HttpStatusCode code = HttpStatusCode.OK,
+            string contentType = "text/plain");
+
+        void Respond(
+            byte[] data,
+            int code = 200,
+            string contentType = "text/plain");
+
         void RespondError(HttpStatusCode code);
 
         void RespondJson(object jsonData, HttpStatusCode code = HttpStatusCode.OK);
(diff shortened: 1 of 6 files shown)


20kdc:
Just to be absolutely clear, this should only trigger if the launcher's poking the server *and* there is no `download_url`.

20kdc:
I'm going to ask this PR not be merged at this time (but feel free to review it) because I have some further changes I want to do to make even more CVars optional.

The ability for the launcher to infer `status.connectaddress` is a good start but right now it conflicts with the way this PR uses it to infer `status.selfaddress`. 

With that in mind I think it'd be best if the launcher is responsible for locating a status-host-hosted `acz.zip` (maybe under the name `client.zip`), and I drop the `selfaddress` stuff entirely.

20kdc:
With that last commit I have made the changes I wanted to make, but it now requires https://github.com/space-wizards/SS14.Launcher/pull/27 to be of use.

20kdc:
Well, with that merged, here's an update on the current status of the PR:
Assuming a compatible launcher version, you only need to supply the engine version. The automatic client zip generator will do the rest assuming a compatible build environment.

20kdc:
Unable to run an absolute full test because we're between RT versions right now, but obviously the live tests went well enough, and I've checked the info endpoint to make sure engine version is propagating nicely

20kdc:
It looks like actual client zip generation won't work with `FULL_RELEASE`, but that's because `FULL_RELEASE` has a different filesystem layout - and with that layout it's not actually possible to determine where the client and shared assemblies are supposed to be (maybe stuff them in Resources/Assemblies or something).
Note that loading an existing pre-built client zip into the status host server will remain working on `FULL_RELEASE`.

PJB3005:
Performance wise this implementation isn't all that great:
1. Using `lock()` and no `Task.Run` will lock up the HTTP server while the ACZ is genrated for the first time, async would be better here.
2. Using `byte[]` buffers will cause quite massive memory abuse.

But that's probably not the end of the world.

PR #2226, opened 2021-11-13
Add method to get worldpos and worldrot at the same time

metalgearsloth:
Somewhat of a gain because it halves the number of GetComponent<TransformComponent> + _parent.IsValid() calls.

~5% speedup for collision code (3/4 of it is still just GetComponent<TransformComponent>)

Needs a test + benchmark added before merging.

Edit: Okay I realise there's no easy way to setup a sim and benchmark this in isolation (just the physics benchmark where the reduced GetComponent<T> calls were showing) but I doubt getting the 3 properties at once would be slower.


metalgearsloth:
Gonna try benchmarking with dummy transformcomponents then good for review after I attempt that.

metalgearsloth:
Okay benchmarking "well" was proving difficult so I ended up just spawning 100k entities on content and then comparing each multiple times and the new method takes roughly 60% of the time of doing each one separately.

Acruid (approved):
This being twice as fast makes sense, nice!

PR #2235, opened 2021-11-15
Async status host handlers to reduce ACZ stalling

20kdc:
This addresses that PR comment about use of `Task.Run`/etc.
However it doesn't address the byte[] situation - I'm not sure of a good fix for now, and it may become semi-irrelevant if some sort of resource manifest system crops up later.

### DIFF
--- a/Robust.Server/ServerStatus/StatusHost.ClientZip.cs
+++ b/Robust.Server/ServerStatus/StatusHost.ClientZip.cs
@@ -47,21 +47,36 @@ namespace Robust.Server.ServerStatus
             return true;
         }
 
+        // Only call this if the download URL is not available!
         private async Task<AutomaticClientZipInfo?> PrepareACZ()
         {
-            // Fast-path: Attempt to acquire lock on the calling thread, grab fields, release and return synchronously
-            if (_aczLock.Wait(0))
-            {
-                var p = _aczPrepareAttempted;
-                var d = _aczPrepared;
-                _aczLock.Release();
-                if (p) return d;
-            }
-            // Slow-path: Take the ACZ lock asynchronously, then let PrepareACZTask do everything (synchronously, but it's everything-bound)
+            // Take the ACZ lock asynchronously
             await _aczLock.WaitAsync();
             try
             {
-                return await Task.Run(PrepareACZTask);
+                // Setting this now ensures that it won't fail repeatedly on exceptions/etc.
+                if (_aczPrepareAttempted) return _aczPrepared;
+                _aczPrepareAttempted = true;
+                // ACZ hasn't been prepared, prepare it
+                byte[] data;
+                try
+                {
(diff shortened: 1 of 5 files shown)


Robust.Server/ServerStatus/StatusHost.ClientZip.cs
```diff
@@ -35,22 +36,38 @@ private bool HandleAutomaticClientZip(IStatusHandlerContext context)
...
+            Monitor.TryEnter(_aczLock, ref lockTaken);
+            if (lockTaken)
+            {
+                var p = _aczPrepareAttempted;
+                var d = _aczPrepared;
+                Monitor.Exit(_aczLock);
+                if (p) return d;
+            }
+            // Slow-path: Start a task to run ACZ (prevents stalling the HTTP server during packaging)
+            return await Task.Run(PrepareACZTask);
```
PJB3005: You can use a `SemaphoreSlim` instead of all this instead and then `.WaitAsync()` it instead. Much easier and won't clog any thread pool threads.

PJB3005 (changes requested):
You can use

Robust.Server/ServerStatus/StatusHost.ClientZip.cs
20kdc: the `TryEnter` is meant as a fast-path, preventing any async operations from having to occur if the lock isn't contended, so I'll use `.Wait(0)` here and `WaitAsync` in the async part.

20kdc: Done

Robust.Server/ServerStatus/StatusHost.ClientZip.cs
```diff
@@ -35,43 +36,61 @@ private bool HandleAutomaticClientZip(IStatusHandlerContext context)
...
-                        return null;
-                    }
-                    data = maybeData;
-                }
-                catch (Exception e)
+                var p = _aczPrepareAttempted;
+                var d = _aczPrepared;
+                _aczLock.Release();
+                if (p) return d;
+            }
```
PJB3005: You don't need this fast path anymore with usage of `WaitAsync`

PJB3005: *that's if you put the actual ACZ generation in a `Task.Run`, which you probably should since it's much clearer than advanced:tm: semaphore usage.

20kdc: see line 64 - that's the `Task.Run`
Then PrepareACZTask etc. runs non-async

20kdc: Moved more stuff to being in the async portion so only ACZ generation is synchronous, but that won't stop the necessity of semaphore usage to stop two requests from conflicting

20kdc:
more code asyncified

PR #2237, opened 2021-11-16
Status host uses net.port by default, add UPnP port forwarding option

20kdc:
*CHECK/FIX PRODUCTION SERVER CONFIGS BEFORE THIS REACHES PRODUCTION!!!*
should be compatible but check anyway

UPnP is not enabled by default because it could mess with stuff that people don't want messed with


PJB3005:
image Good tests, github

20kdc:
ok, so, if you look at the test failure details, the answer is that the criteria for a test failure is *any* warning
and *any* warning includes, say, a UPnP failure warning

Robust.Server/Network/UPnP.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            // We DON'T want to hold up the main server on this!
+            new Thread(() =>
+            {
+                try
+                {
+                    // The way the NetUPnP code is written, we're doing guesswork anyway
+                    // It seems to be that the assumption in regards to IPv6 is "what?" w/ UPnP????
+                    // ATTENTION FUTURE 20KDC (or anyone else who comes by):
+                    // IF YOU GET IPv6 FOR REALSIES, WORK OUT HOW TO DEAL W/ THIS!
```
PJB3005: IPv6 support was retrofitted into Lidgren (by me) so take that as you will.

20kdc: supposedly the IPv6 Plan (r) is "no NAT!" so we'll have to see how it goes

20kdc:
disabled UPnP by default for... well, various reasons really.

Robust.Server/Network/UPnP.cs
```diff
@@ -0,0 +1,62 @@
...
+using Robust.Shared.Log;
+
+namespace Robust.Server.Network
+{
+    internal static class UPnP
+    {
+        public static void Start(int port, ISawmill sawmill)
+        {
+            // We DON'T want to hold up the main server on this!
+            new Thread(() =>
```
PJB3005: Using `Task.Run()` here is probably easier?

20kdc: Does the thread-pool automatically expand if one of the threads blocks for a long period of time?

PJB3005: It does, yeah. But I figure the amount of time this takes shouldn't be too long to cause a problem.

The even higher effort approach would be to make the Lidgren APIs for this async but ah well. You can at least make the `NetUtility.Sleep` used to poll the UPnP status a `Task.Delay` instead (obviously ideally it'd be an event or smth but ah well...)

20kdc: ah, I was under the assumption that using NetUtility.Sleep was critically necessary to the functioning of Lidgren

PJB3005: It just calls `Thread.Sleep()` directly, so...

mirrorcult:
should update the hosting docs once this is merged to indicate that you can use UPnP

PJB3005:
Lidgren needs some work on its side to fix the UPnP stuff, working on it.

PR #2239, opened 2021-11-17
Miscellaneous fixes for thrusters

metalgearsloth:
Most important one is the sprite layer offset fix.

Note that this kinda breaks the spawn entity menu for anything that actually uses it so not sure on the best bandaid (I mean ideally it'd have a dedicated icon); could just not draw layers that have offsets or something?

Known issue: InteractionOutline (only on anything that uses layer offset), because of course, but I have NFI how to fix.

### DIFF
--- a/Robust.Shared/Physics/SharedBroadphaseSystem.cs
+++ b/Robust.Shared/Physics/SharedBroadphaseSystem.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Security.Cryptography;
+using Robust.Shared.Configuration;
 using Robust.Shared.Containers;
 using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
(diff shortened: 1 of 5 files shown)


Robust.Shared/Physics/SharedBroadphaseSystem.cs
```diff
@@ -847,8 +864,8 @@ private void OnMapDestroyed(object? sender, MapEventArgs e)
         public override void Shutdown()
         {
             base.Shutdown();
-            _mapManager.MapCreated -= OnMapCreated;
-            _mapManager.MapDestroyed -= OnMapDestroyed;
```
PaulRitter: a what now

PaulRitter (changes requested):

Robust.Shared/Physics/SharedBroadphaseSystem.cs
metalgearsloth: uhh the merge might've nuked it, that's weird it didn't show up

PR #2243, opened 2021-11-18
Optimise broadphase a LOT for grids moving

metalgearsloth:
Previously it went through the entire movebuffer (O(n) for moved bodies). Now it just uses the localbounds (O(1)).


gradientvera:
conflicts

gradientvera (approved):

Issue #2198, opened 2021-11-05
Overlapping PVS view subscribers cause crash on client if anchored entities change


PJB3005:
Go to the bar, add saltern as view subscriber to your client, place a new wall somewhere. Client receives entity creation states multiple times and dies.

PR #2247, opened 2021-11-20
pvs refactor

PaulRitter:
requires https://github.com/space-wizards/space-station-14/pull/5484

entitystates are be generated o(n)
everything is now cached in grid/mapchunks
new entities get streamed one by one


PaulRitter:
ready for review

Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -1,680 +1,632 @@
...
-        #region Player Status
+        if(e.Map == MapId.Nullspace) return;
+        var uid = _mapManager.GetMapEntityId(e.Map);
+        _entityPvsCollection.UpdateIndex(uid);
+    }
+
+    #endregion
+
+    //todo make this actually use toTick in non-all-sending
+    public (List<EntityState>? updates, List<EntityUid>? deletions) CalculateEntityStates(ICommonSession session,
```
mirrorcult, line 261: it kinda seems like a lot of comments were removed in this func + a lot of other funcs were just merged into this

Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -1,680 +1,632 @@
...
+        if (entityStates.Count == 0) entityStates = default;
+        return (entityStates, deletions);
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveOptimization)]
+    private bool RecursiveAdd(EntityUid uid, HashSet<EntityUid> visSet, uint visMask)
+    {
+        // we are done, this ent has already been checked and is visible
+        if (visSet.Contains(uid))
+            return true;
```
mirrorcult: does this fix #2198?

Robust.Shared/GameObjects/EntityManager.Components.cs
```diff
@@ -695,12 +695,19 @@ public IEnumerable<IComponent> GetAllComponents(Type type, bool includePaused =
...
+            var getState = new ComponentGetState();
             eventBus.RaiseComponentEvent(component, ref getState);
 
-            return getState.State ?? component.GetComponentState(player);
+            return getState.State ?? component.GetComponentState();
+        }
+
+        public bool CanGetComponentState(IEventBus eventBus, IComponent component, ICommonSession player)
+        {
+            var attempt = new ComponentGetStateAttemptEvent(player);
```
mirrorcult, line 708: feels like this could be wasteful considering 99.9% of things won't ever care, buuuut its not like the eventbus is slow and this is way better than the alternative..

Robust.Shared/GameStates/ComponentStateEvents.cs
```diff
@@ -21,19 +21,26 @@ public ComponentHandleState(ComponentState? current, ComponentState? next)
...
-        public readonly ICommonSession Player;
+        public ComponentState? State { get; set; }
 
+        public ComponentGetState()
+        {
+            State = null;
+        }
+    }
+
+    public class ComponentGetStateAttemptEvent : CancellableEntityEventArgs
```
mirrorcult, line 34: i feel like this should really be ref struct, and just implement cancellation yourself

Robust.UnitTesting/Shared/GameObjects/ContainerTests.cs
```diff
@@ -99,14 +99,13 @@ public async Task TestContainerNonexistantItems()
                 }
 
                 var container = containerManagerComp.GetContainer("dummy");
-                Assert.That(container.ContainedEntities.Count, Is.EqualTo(1));
-                Assert.That(container.ExpectedEntities.Count, Is.EqualTo(0));
+                Assert.That(container.ContainedEntities.Count, Is.EqualTo(0));
+                Assert.That(container.ExpectedEntities.Count, Is.EqualTo(1));
```
mirrorcult, line 103: why the changes here?

Robust.UnitTesting/Shared/GameObjects/ContainerTests.cs
```diff
@@ -226,14 +225,12 @@ public async Task TestContainerExpectedEntityDeleted()
...
```
mirrorcult, line 229: same here

mirrorcult (commented):
some misc things
- does sloth's 'pvs turned off' optimization still work?
- have you tested for any regressions for previous PVS bugs? (this kinda makes me wish we had testmerges)
- was all the PVS entity anchoring stuff removed? i dont really see any mentions of anchoring but I could just be blind? if so was it removed because it was forgotten/not viable with the current architecture/unneeded optimization/overcomplicated and will be done in a future PR?

Robust.Server/GameStates/PVSSystem.cs
mirrorcult: be nice to have more docs overall--not that there aren't any, but

mirrorcult:
another misc question: benchmarks?

Robust.Shared/GameObjects/EntityManager.Components.cs
PaulRitter: yeah... i cant come up with a better way to make generating entitystates be o(n)

Robust.UnitTesting/Shared/GameObjects/ContainerTests.cs
PaulRitter: sloth fumbled it to incorrect values when fixing tests. these were the initial values, i just reverted back to them

Robust.UnitTesting/Shared/GameObjects/ContainerTests.cs
PaulRitter: sloth fumbled it to incorrect values when fixing tests. these were the initial values, i just reverted back to them

PaulRitter:
> does sloth's 'pvs turned off' optimization still work?

yes. i didn't touch the "pvs turned off" part at all

> have you tested for any regressions for previous PVS bugs? (this kinda makes me wish we had testmerges)

you got a list of those? i addressed a bunch simply be rewriting & unfucking a bunch of shit.

> was all the PVS entity anchoring stuff removed? i dont really see any mentions of anchoring but I could just be blind? if so was it removed because it was forgotten/not viable with the current architecture/unneeded optimization/overcomplicated and will be done in a future PR?

it was a scuffed optimization ontop of the old system & was an attempt at making it a bit more performant, given old pvs makes an entitylookup every tick. simply not needed anymore, it was basically what i am doing but only for anchored entities.

PaulRitter:
> another misc question: benchmarks?

havent done any so far

Robust.Server/GameStates/PVSSystem.cs
PaulRitter: this check also exists in old pvs. however, i think that crash might be related to the chunking optimization, so i'd think it is solved (i havent tried ingame yet)

Robust.Server/GameStates/PVSSystem.cs
PaulRitter: yes. i plan to document pvs & also work on it further. eventually pvscollections will be automatically synced between client & server pvssystem & other systems can just grab the collections. i plan to expose methods to register pvscollections with the serverside pvssystem. those are private for now tho

metalgearsloth:
Getting crashes with 2 clients:
image

Robust.Shared/Map/MapManager.cs
```diff
@@ -502,6 +502,16 @@ public IMapGrid GetGrid(GridId gridID)
...
 
+        public bool IsGrid(EntityUid uid)
+        {
+            return _grids.Any(x => x.Value.GridEntityId == uid);
+        }
+
+        public bool IsMap(EntityUid uid)
+        {
+            return _mapEntities.Any(x => x.Value == uid);
+        }
```
metalgearsloth, line 513: Should probably check for the components instead?

metalgearsloth (approved):

PR #2249, opened 2021-11-20
Super Pausing World

### LINKED ISSUE #1444: A grid added after a map is paused has its grid entity unpaused
- Create a map (sMapManager.CreateMap).
- Pause it (sPauseManager.SetMapPaused(mapId, true)).
- Create a grid on that map (sMapManager.CreateGrid(mapId)).
- IPauseManager.IsMapPaused will return true for that map id.
- IPauseManager.IsGridPaused will return true for that grid id.
- IEntity.Paused will be false for that grid's entity.

Pausing the map after the grid is created does pause the grid's entity though.

### LINKED ISSUE #1445: Entities added to a paused grid afterwards are not themselves paused
See https://github.com/space-wizards/RobustToolbox/issues/1444

Follow the steps on that PR and then spawn an entity, it will not be paused.

Acruid:
Bringing map pausing into the EntityManager.

Resolves https://github.com/space-wizards/RobustToolbox/issues/1444.
Resolves https://github.com/space-wizards/RobustToolbox/issues/1445.

### DIFF
--- a/Robust.Client/ClientIoC.cs
+++ b/Robust.Client/ClientIoC.cs
@@ -51,6 +51,7 @@ namespace Robust.Client
             IoCManager.Register<IReflectionManager, ClientReflectionManager>();
             IoCManager.Register<IConsoleHost, ClientConsoleHost>();
             IoCManager.Register<IClientConsoleHost, ClientConsoleHost>();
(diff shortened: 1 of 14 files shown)


Acruid:
Turns out that you cannot hold a ref to a valtype in a field, so this PR is dead.

PR #2254, opened 2021-11-23
Slight tweak to version.py

metalgearsloth:
Allows me to run it from RobustToolbox or Tools (I tried the latter originally)


PJB3005 (commented):
Resolve the path by looking at the script directory instead if you're gonna do this, IMO.

PR #2260, opened 2021-11-25
Switch back to an NFluidSynth fork just to fix Linux library resolution

20kdc:
This was my first experience with NuGet.
And by the time all the PRs involved in this mess merge, it'll be Easter, so happy Christmas.


20kdc:
Closed for internal reasons.

PR #2263, opened 2021-11-26
Unwrap ParsedMain

wrexbe:
What the previous code does:
1. Creates a new thread, and makes the main thread do nothing forever, waiting on the new thread.
2. Messes up exception stack traces

What it does now:
1. Uses the main thread to run the code, instead of making a new thread to do it. So the main thread isn't wasted.
2. Lets the debugger find the correct place when an exception happens.

### DIFF
--- a/Robust.Server/Program.cs
+++ b/Robust.Server/Program.cs
@@ -40,15 +40,7 @@ namespace Robust.Server
 
             ThreadPool.SetMinThreads(Environment.ProcessorCount * 2, Environment.ProcessorCount);
 
-            // this sets up TaskScheduler.Current for new tasks to be scheduled off the main thread
-            // the LongRunning option causes it to not be scheduled on the thread pool, so it's just a plain thread with a task context
-            new TaskFactory(
-                    CancellationToken.None,
-                    TaskCreationOptions.LongRunning,
-                    TaskContinuationOptions.None,
-                    RobustTaskScheduler.Instance
-                ).StartNew(() => ParsedMain(parsed, contentStart, options))
-                .GetAwaiter().GetResult();
+            ParsedMain(parsed, contentStart, options);
         }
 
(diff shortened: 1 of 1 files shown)


gradientvera:
DNM until @PJB3005 look at this

PJB3005:
How did  you solve the thread scheduling problem with PVS that the comment mentions? Is it not a problem anymore?

wrexbe:
It just looks like a noop to me. It's starting a new thread, and makes the main method wait on it. It's not going to schedule something that could be blocked  on the main thread, because there isn't a synchronization context. There is a slight difference, because it's running off of the default task scheduler now, instead of the RobostTaskScheduler, which seems to be an older copy of the built in one, and is avoiding inlining, so it would be slower, because it'll make a new task, and then go find another task, instead of continuing with what it has. https://github.com/Microsoft/referencesource/blob/master/mscorlib/system/threading/Tasks/ThreadPoolTaskScheduler.cs

PJB3005:
The reason this was done was so that any uses of `Parallel` would not schedule jobs on the main thread because the main thread is perpetually seen as "busy" by the .NET task scheduler. Without this, PVS jobs will schedule to the main thread and then bulldozer the IoC context.

wrexbe:
It shouldn't ever schedule to the main thread, because it's not a thread pool thread.

PJB3005:
> It shouldn't ever schedule to the main thread, because it's not a thread pool thread.

It still does because the thread would otherwise just be blocked waiting on all the other threads to finish, so they opportunistically use the thread anyways.

PJB3005:
I just realized at some point, somebody made PVS explicitly avoid bulldozing the main-thread IoC context.

In that case, this PR is fine.

PJB3005:
image

Just for reference, `Parallel` does schedule workers on the main thread again with this PR. That's fine though because it guards against this now.

PR #2282, opened 2021-11-28
Fix multi-subscriber crash for PVS

### LINKED ISSUE #2198: Overlapping PVS view subscribers cause crash on client if anchored entities change

metalgearsloth:
Not the most elegant solution but PaulVS sometime soon-ish so the perf impact is immaterial and it's mainly to avoid stalling the opendream branch.

Fixes #2198

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -373,6 +373,10 @@ namespace Robust.Server.GameStates
 
                     chunk.FastGetAllAnchoredEnts(uid =>
                     {
+                        // Given the anchored ent is parented to the chunk we already have its parents
+                        // Could maybe also add vis check here? Though PaulVS soon TM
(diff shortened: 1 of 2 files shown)


Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -373,6 +373,10 @@ private HashSet<EntityUid> GetSessionViewers(ICommonSession session)
 
                     chunk.FastGetAllAnchoredEnts(uid =>
                     {
+                        // Given the anchored ent is parented to the chunk we already have its parents
+                        // Could maybe also add vis check here? Though PaulVS soon TM
+                        if (!visibleEnts.Add(uid)) return;
```
metalgearsloth, line 378: Easiest way to do a duplicate check, we just need to filter it out later

PR #2283, opened 2021-11-28
Fix `AudioParams` DataDefinition

ElectroJr:
Currently `AudioParams` is a data definition, so that it can be defined in yaml, but AFAIK it can't actually be used at the moment. When you try to use them you get a null reference exception when trying to get the default MaxDistance. The default `AudioParams` MaxDistance calls a static function that runs:
```cs
var args = new QueryAudioSystem();
IoCManager.Resolve<IEntityManager>().EventBus.RaiseEvent(EventSource.Local, args);
``` 
In order to access a `public readonly int` that is defined separately in the client & server, and in the case of the server just points to a `private const int`??

This PR just replaces all this with a single `public const int` defined in `SharedAudioSystem`. I have no idea why this was set up the way that it was, if there's a reason for it then something else needs to be done to fix `AudioParams` data-definitions.

The PR also updates `AudioParams.PopulateDefaultValues()` to use the same values as the static `AudioParams.Default`.

### DIFF
--- a/Robust.Shared/Audio/AudioParams.cs
+++ b/Robust.Shared/Audio/AudioParams.cs
@@ -3,6 +3,7 @@ using System;
 using System.Diagnostics.Contracts;
 using Robust.Shared.Serialization.Manager;
 using Robust.Shared.Serialization.Manager.Attributes;
+using Robust.Shared.GameObjects;
 
 namespace Robust.Shared.Audio
 {
@@ -84,7 +85,7 @@ namespace Robust.Shared.Audio
         /// <summary>
         ///     The "default" audio configuration.
         /// </summary>
-        public static readonly AudioParams Default = new(0, 1, "Master", SoundSystem.DefaultSoundRange, 1, 1, false, 0f);
+        public static readonly AudioParams Default = new(0, 1, "Master", SharedAudioSystem.DefaultSoundRange, 1, 1, false, 0f);
 
         public AudioParams(
             float volume,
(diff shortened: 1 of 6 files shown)


Robust.Shared/Audio/AudioParams.cs
```diff
@@ -216,9 +217,15 @@ public AudioParams WithPlayOffset(float offset)
...
+            Attenuation = Default.Attenuation;
+            Volume = Default.Volume;
+            PitchScale = Default.PitchScale;
+            BusName = Default.BusName;
+            MaxDistance = Default.MaxDistance;
+            RolloffFactor = Default.RolloffFactor;
+            ReferenceDistance = Default.ReferenceDistance;
+            Loop = Default.Loop;
+            PlayOffsetSeconds = Default.PlayOffsetSeconds;
         }
```
ElectroJr, line 229: Is there a better way to just copy all the values of some static default struct? `this = Default` seems to work, but also seems like black magic.  I imagine using `this = Default` means that if this struct ever had a reference types added without updating this function, then it would copy the reference and potentially cause issues? So maybe copying each element explicitly is better?

PJB3005: It would, but I'm not sure that's a major concern.

Acruid (approved):
`QueryAudioSystem` was a bad hack to get around not being able to get a ref to the server/client audio system in a shared context. There is no Grand Design about it, it was set up like that because I did not catch how dumb it was when I was creating SoundSystem.

PR #2295, opened 2021-11-30
The End of Entity (Komm, Süsser Todd: Part 1.0)

gradientvera:
~~Merge https://github.com/space-wizards/RobustToolbox/pull/2296 first.~~
Required by https://github.com/space-wizards/space-station-14/pull/5624

- Makes `IEntity` a sealed class instead of an interface.
- Removes `Entity`.
- Makes all `Entity` usages use `IEntity` instead.

See commits.
I won't change `IEntity` to be called `Entity` because it'd be a massive, breaking refactor for a type we're removing soon anyway.

"It all returns to nothing...
It all comes tumbling down,
tumbling down, tumbling down"

PR #2296, opened 2021-12-01
Remove Entity Transform/MetaData Component Cache

Acruid:
Removes transform/metadata caching on Entity, disposes components after calling event so systems can properly access comps to cleanup.

This fixes a problem that prevented https://github.com/space-wizards/RobustToolbox/pull/2295 from being complete.

### DIFF
--- a/Robust.Shared/GameObjects/Entity.cs
+++ b/Robust.Shared/GameObjects/Entity.cs
@@ -77,7 +77,7 @@ namespace Robust.Shared.GameObjects
         public bool Deleted => LifeStage >= EntityLifeStage.Deleted;
 
         [ViewVariables]
-        public bool Paused { get => MetaData.EntityPaused; set => MetaData.EntityPaused = value; }
+        public bool Paused { get => Deleted || MetaData.EntityPaused; set => MetaData.EntityPaused = value; }
 
         /// <inheritdoc />
         [ViewVariables]
--- a/Robust.Shared/GameObjects/Entity.cs
+++ b/Robust.Shared/GameObjects/Entity.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using Robust.Shared.IoC;
 using Robust.Shared.Network;
 using Robust.Shared.Physics;
 using Robust.Shared.Prototypes;
@@ -17,7 +18,7 @@ namespace Robust.Shared.GameObjects
         #region Members
 
         /// <inheritdoc />
-        public IEntityManager EntityManager { get; }
+        public IEntityManager EntityManager => IoCManager.Resolve<IEntityManager>();
 
         /// <inheritdoc />
         [ViewVariables]
@@ -26,7 +27,17 @@ namespace Robust.Shared.GameObjects
         /// <inheritdoc />
         EntityLifeStage IEntity.LifeStage { get => LifeStage; set => LifeStage = value; }
 
-        public EntityLifeStage LifeStage { get => MetaData.EntityLifeStage; internal set => MetaData.EntityLifeStage = value; }
+        public EntityLifeStage LifeStage
+        {
+            get
+            {
+                if (!EntityManager.EntityExists(Uid))
+                    return EntityLifeStage.Deleted;
+
+                return MetaData.EntityLifeStage;
+            }
+            internal set => MetaData.EntityLifeStage = value;
+        }
 
         [ViewVariables]
         GameTick IEntity.LastModifiedTick { get => MetaData.EntityLastModifiedTick; set => MetaData.EntityLastModifiedTick = value; }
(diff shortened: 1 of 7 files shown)


gradientvera:
Depends on https://github.com/space-wizards/space-station-14/pull/5657 being merged so tests pass on content

Robust.Shared/GameObjects/Entity.cs
```diff
@@ -68,20 +79,24 @@ public string Name
...
-        public TransformComponent Transform => _transform ??= GetComponent<TransformComponent>();
-
-        private MetaDataComponent? _metaData;
+        public TransformComponent Transform
+        {
+            get
+            {
+                return (TransformComponent)IoCManager.Resolve<IEntityManager>().GetComponent(Uid, typeof(TransformComponent));
+            }
+        }
```
PaulRitter, line 90:
```suggestion
        public TransformComponent Transform => EntityManager.GetComponent<TransformComponent>(Uid);
```

Robust.Shared/GameObjects/Entity.cs
```diff
@@ -68,20 +79,24 @@ public string Name
...
         [ViewVariables]
         public MetaDataComponent MetaData
         {
-            get => _metaData ??= GetComponent<MetaDataComponent>();
-            internal set => _metaData = value;
+            get
+            {
+                return (MetaDataComponent)IoCManager.Resolve<IEntityManager>().GetComponent(Uid, typeof(MetaDataComponent));
+            }
         }
```
PaulRitter, line 100: should be
```csharp
        public MetaDataComponent MetaData => EntityManager.GetComponent<MetaDataComponent>(Uid);
```

Robust.Shared/GameObjects/Entity.cs
```diff
@@ -90,7 +105,6 @@ public MetaDataComponent MetaData
 
         public Entity(IEntityManager entityManager, EntityUid uid)
```
PaulRitter, line 106:
```suggestion
        public Entity(EntityUid uid)
```

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -402,7 +409,6 @@ private protected Entity AllocEntity(EntityUid? uid = null)
 
             // Create the MetaDataComponent and set it directly on the Entity to avoid a stack overflow in DEBUG.
             var metadata = new MetaDataComponent() { Owner = entity };
-            entity.MetaData = metadata;
 
             // add the required MetaDataComponent directly.
             AddComponentInternal(uid.Value, metadata);
```
PaulRitter, line 414: i think this means we can now replace this with
```csharp
AddComponent<MetadataComponent>(entity);
```

PaulRitter (changes requested):

gradientvera:
@PaulRitter Except for the `AddComponent<MetaDataComponent>()` changes, all the changes you proposed are already part of #2295 so I'm not gonna do them here as well

Robust.Shared/GameObjects/EntityManager.cs
gradientvera: We can't do this yet because of how `AddComponent<T>(EntityUid)` works
```csharp
        public T AddComponent<T>(EntityUid uid) where T : Component, new()
        {
            if (!TryGetEntity(uid, out var entity)) throw new ArgumentException("Entity is not valid or deleted.", nameof(uid));

            return AddComponent<T>(entity);
        }
```
However, I think we will be able to in a future:tm: PR for this whole IEntity removal ordeal

PaulRitter: In L401, the entity gets added to the internal cache meaning the trygetent will go through. thats also why in L411, right below this, the `AddComponent<TransformComponent>(entity);` works without a hitch

gradientvera: I literally tested and it didn't work...

PaulRitter: I assume you used `AddComponent<MetadataComponent>(uid.Value);` instead of `AddComponent<MetadataComponent>(entity);`, what i said. Using the uid variant will try to get the entity and call the other variant.
I was wrong on the trygetent tho, weird that it falls through. Only reason i could see it being marked as Deleted but uhhhh, that shouldnt happen should it.

Robust.Shared/GameObjects/Entity.cs
```diff
@@ -77,7 +77,7 @@ public string Name
         public bool Deleted => LifeStage >= EntityLifeStage.Deleted;
 
         [ViewVariables]
-        public bool Paused { get => MetaData.EntityPaused; set => MetaData.EntityPaused = value; }
```
PaulRitter, line 69: absolutely based

Robust.Shared/GameObjects/EntityManager.cs
PaulRitter: nvm

PR #2299, opened 2021-12-01
Make AudioSystem no longer automatically add players in range. 

ElectroJr:
Currently, the audio system has this bit of code:
```cs
if (range > 0.0f)
    playerFilter = playerFilter.Clone().AddInRange(coordinates.ToMap(EntityManager), range);
```
This functionally prevents you from using the player filter to restrict who receives the audio. 
Given that this can lead to confusion (e.g. in the traitor uplink system), and that almost every use of `SoundSystem.Play()` in content specifies it's own filter, this should probably just be removed.

Would be useful to have for space-wizards/space-station-14/pull/5638


Acruid:
Setting the audioparams max range to <= 0 should already have the same effect?

ElectroJr:
> Setting the audioparams max range to <= 0 should already have the same effect?

If the distance is <= 0 (or if audio params is null) then the distance is just set to the default distance:
```cs
var range = audioParams is null || audioParams.Value.MaxDistance <= 0 ? DefaultSoundRange : audioParams.Value.MaxDistance;
```
so it still includes players in range. I guess a simpler fix would have been to have negative distances imply that it shouldn't add players in range.E.g., replacing the above with just:
```cs
var range = audioParams is null ? DefaultSoundRange : audioParams.Value.MaxDistance;
```
Should I just update the PR to do that instead? Could setting the `AudioParams` max distance to negative numbers lead to odd effects elsewhere?

Acruid:
Coming at the problem from a different direction, does it even make sense for the function to modify the filter like it does? There are no other functions that do that, and it seems like having the AudioSystem adding more clients like this is just going to confuse people.

Adding the `addInRange` flag to disable the snowflake functionality seems like a work around when the function should not be doing it's snowflake shit of modifying the filter to begin with. 

TBH I think you should just remove the lines that modify the filter to solve the problem, so your work around would not be required in the first place. Almost every callsite already filters the receiving clients to the PVS set, and if they want to actually filter to the arbitrary audio range they can with the AddInRange function.

ElectroJr:
Yeah so looks like every use of `SoundSystem.Play()` specifies it's own filter and don't rely on automatically adding players in range, so I've just removed it.

edit: actually some systems/components use `Filter.Local()` which is empty for the server. I'll update those in a content PR


>  and it seems like having the AudioSystem adding more clients like this is just going to confuse people.

So found a fun fact while checking `SoundSystem.Play()` usages: the traitor uplink, which uses `Filter.SinglePlayer()` to only play the buying-sounds to the user, was making a sound audible to everyone within the default audio range. So yeah definitely caused some confusion.

Acruid (approved):

Clyybber (approved):

PR #2301, opened 2021-12-02
Better ToPrettyString with EntityStringRepresentation

gradientvera:
Basically, instead of returning a plain string, we return a `readonly struct record` that contains a bunch of needed info for SS14 admin logging purposes. This type implicitly converts to `string` and `EntityUid` for convenience.

### DIFF
--- a/Robust.Shared/GameObjects/EntityStringRepresentation.cs
+++ b/Robust.Shared/GameObjects/EntityStringRepresentation.cs
@@ -2,6 +2,19 @@ using Robust.Shared.Players;
 
 namespace Robust.Shared.GameObjects;
 
+/// <summary>
+///     A type that represents an entity, and allows you to get a string containing human-readable information about it.
+///     This type converts implicitly to string, for convenience purposes.
+/// </summary>
+/// <remarks>
+///     This can be used to pretty-print information about entities and also to log various information regarding an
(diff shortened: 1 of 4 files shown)


metalgearsloth (approved):
probably fine

Robust.Shared/GameObjects/EntityStringRepresentation.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public override string ToString()
+    {
+        if (Deleted && Name == null)
+            return $"{Uid}D";
+
+        return $"{Name} ({Uid}{(Prototype != null ? $", {Prototype}" : "")}{(Session != null ? $", {Session.Name}" : "")}){(Deleted ? "D" : "")}";
+    }
+
+    public static implicit operator EntityUid(EntityStringRepresentation rep) => rep.Uid;
```
PaulRitter: why this implicit operator? not sure i like it

PaulRitter (changes requested):

Robust.Shared/GameObjects/EntityStringRepresentation.cs
gradientvera: honestly it's unused so I'll just remove it

PaulRitter (approved):

PR #2303, opened 2021-12-03
Optimise some physics stuff

metalgearsloth:
Also fixed a ray penetration thing.


Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -787,6 +787,15 @@ public TransformComponent GetMapTransform()
         /// Get the WorldPosition, WorldRotation, and InvWorldMatrix of this entity faster than each individually.
         /// </summary>
         public (Vector2 WorldPosition, Angle WorldRotation, Matrix3 InvWorldMatrix) GetWorldPositionRotationInvMatrix()
+        {
+            var (worldPos, worldRot, _, invWorldMatrix) = GetWorldPositionRotationMatrixWithInv();
+            return (worldPos, worldRot, invWorldMatrix);
```
PaulRitter, line 795: really just a nitpick but shouldn't GetWorldPositionRotationMatrixWithInv be calling GetWorldPositionRotationInvMatrix to get the 3 values & just return the localmatrix with them instead of doing it the other way around, throwing away one value

PaulRitter (approved):

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
metalgearsloth: How does it get the worldmatrix out of the method then?

gradientvera:
Conflicts

PR #2306, opened 2021-12-03
Graphics/Clyde: add option to submit GLSL `#defines` for shaders

Efruit:
Still does what it says on the tin. Does absolutely nothing to prevent malformed `#define`s, so probably some room for improvement there. There *may* be a more clean option provided by the OpenGL library, but I didn't bother looking.


Acruid:
The code is perfect, every PR should strive to be on this level.

Efruit:
> The code is perfect, every PR should strive to be on this level.

this makes me particulary nervous

gradientvera:
> The code is perfect, every PR should strive to be on this level.

reading this gave me body system PTSD

PR #2309, opened 2021-12-04
Allow tight grid fits

metalgearsloth:
I tried using up to 3/4 of the poly radius for the shrinking but that still made it nigh impossible to move. This allows the movement to be smooth and games can override it to whatever they want.

This does lead to visible bound seams *but* it shouldn't matter outside of grid collisions. When we get half-tiles TryFindGridAt will need updating but we can cross that bridge when we get to it.

Next map save will also update these.

image


Efruit (approved):
:thumbsup:

PR #2310, opened 2021-12-04
Add new EntitySystem proxy methods.

gradientvera:
Consider this the fabled `IEntity` replacement.
Adds a *lot* of new entity system methods for convenience and QOL.
Most methods have been called with very short names so you have to type less without bulldozing readability.
The idea is that we will move more code that deals with entities to entity systems, and that this will replace the convenience of ol' `IEntity`.

Note: The methods have been made public for now, since they're really just proxy methods that interact with `EntityManager` and the MetaData/Transform components. Resolves have also been made public as well for consistency.

Some Examples:
BEFORE
`EntityManager.GetComponent<MetaDataComponent>(uid).EntityName;`
AFTER
`Name(uid);`

BEFORE
`EntityManager.GetComponent<TransformComponent>(uid).Coordinates;`
AFTER
`Transform(uid).Coordinates;`

BEFORE
`EntityManager.GetComponent<FooComponent>(uid);`
AFTER
`Comp<FooComponent>(uid);`

BEFORE
`if(!EntityManager.TryGetComponent(uid, out FooComponent? foo) {}`
AFTER
`if(!TryComp(uid, out FooComponent? foo) {}`

And much more.

### DIFF
--- a/Robust.Shared/GameObjects/EntitySystem.Proxy.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.Proxy.cs
@@ -1,0 +1,523 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using Robust.Shared.Map;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Timing;
+
+namespace Robust.Shared.GameObjects;
+
+public partial class EntitySystem
+{
+    #region Entity LifeStage
+
+    /// <inheritdoc cref="IEntityManager.EntityExists" />
+    [MethodImpl(MethodImplOptions.AggressiveInlining | MethodImplOptions.AggressiveOptimization)]
+    public bool Exists(EntityUid uid)
+    {
+        return EntityManager.EntityExists(uid);
+    }
+
+    /// <summary>
+    ///     Retrieves whether the entity is initializing. Throws if the entity does not exist.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining | MethodImplOptions.AggressiveOptimization)]
+    public bool Initializing(EntityUid uid, MetaDataComponent? metaData = null)
+    {
+        return LifeStage(uid, metaData) == EntityLifeStage.Initializing;
+    }
+
+    /// <summary>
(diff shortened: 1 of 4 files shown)


mirrorcult (approved):
Blessed

Robust.Shared/GameObjects/EntitySystem.Proxy.cs
```diff
@@ -0,0 +1,523 @@
...
+
+        return metaData.EntityLifeStage;
+    }
+
+    /// <summary>
+    ///     Attempts to retrieve whether the entity is initializing.
+    /// </summary>
+    /// <returns>Whether it could be retrieved.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining | MethodImplOptions.AggressiveOptimization)]
+    public bool TryInitializing(EntityUid uid, [NotNullWhen(true)] out bool? initializing, MetaDataComponent? metaData = null)
```
metalgearsloth, line 73: I foresee in 6 months some newcoder is gonna go "woe is me why is it not tryget" but I prefer the brevity

Robust.UnitTesting/Server/Maps/MapLoaderTest.cs
```diff
@@ -75,6 +76,11 @@ protected override void OverrideIoC()
 
             IoCManager.RegisterInstance<IEntitySystemManager>(mock.Object, true);
             //IoCManager.RegisterInstance<ICustomFormatManager>(mockFormat.Object, true);
+
+            // Mocking moment...
+            IoCManager.BuildGraph();
+            IoCManager.RegisterInstance<SharedBroadphaseSystem>(broady);
```
metalgearsloth, line 82: :imfuckingscreaming:

metalgearsloth (approved):

metalgearsloth:
also not sure on the methodimpls so someone who knows more should comment

Robust.UnitTesting/Server/Maps/MapLoaderTest.cs
gradientvera: mocking fucking sucks, and also I have *no clue* why this started failing out of nowhere on this PR...

metalgearsloth (approved):

metalgearsloth (approved):

PR #2311, opened 2021-12-05
Console/SystemConsoleManager: Short-circuit console input handling when not interactive

Efruit:
Falls back to a plain `ReadLine()` if there's no one there to type commands.

### DIFF
--- a/Robust.Server/Console/SystemConsoleManager.cs
+++ b/Robust.Server/Console/SystemConsoleManager.cs
@@ -38,7 +38,7 @@ namespace Robust.Server.Console
         // This is ridiculously expensive to fetch for some reason.
         // I'm gonna just assume that this can't change during the lifetime of the process. I hope.
         // I want this ridiculous 0.1% CPU usage off my profiler.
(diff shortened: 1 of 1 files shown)


Robust.Server/Console/SystemConsoleManager.cs
```diff
@@ -104,13 +111,43 @@ private string UpdateBps()
...
+                    async () =>
+                    {
+                        while (!rlc.IsCancellationRequested)
+                        {
+                            var str = await Con.In
+                                .ReadLineAsync()
+                                .WaitAsync(TimeSpan.FromSeconds(2.0), rlc.Token);
+
+                            if (str != null)
+                                _conShell.ExecuteCommand(str);
```
Acruid: Isn't this running ExecuteCommand on a worker thread?

Efruit: Yep. I wasn't able to find even ONE standard way to read a single line of input without blocking. No way around it without writing our own `ReadLine`, as far as I've found.

cc @PJB3005 in case you want to take a whack at it.

PR #2315, opened 2021-12-08
BQL optimisations

metalgearsloth:
I just went for low-hanging fruit. Haven't checked in game because there doesn't seem to be an easy way to benchmark it.

IEnumerable still sucks in its own way but it's much better than LINQ chaining.

@moonheart08


moonheart08:
lgtm.

moonheart08 (approved):

PR #2316, opened 2021-12-08
Content build tools

komunre:
# Why do we need this
I think that tools for building game content should be in engine, sooo... I hope it will get merged

Please check if you want any change requests, specially in Mac OS code


metalgearsloth (changes requested):
2 things:
1. On discord I mentioned that it would be preferable that any updates are made to the existing script and not copy to a new one
2. Following up on 1 it's not clear what was changed.

metalgearsloth:
We can re-open if you get around to it.

Issue #1855, opened 2021-07-07
API to get the color of a specific pixel in a texture


gradientvera:
Mostly needed for a possible OpenDream port

gradientvera:
Fixed by #2320

PR #2320, opened 2021-12-09
Texture GetPixel implemented

localcc:
Related issue https://github.com/space-wizards/RobustToolbox/issues/1855


Acruid (approved):
I'm just going to note that pulling texture data from the vram like this is *really* slow, so this function needs to be called rarely. This was the first idea before PJB created the click map system for testing clicks against sprite alpha masks.

ike709:
I could be mistaken but IIRC I think OpenDream implemented click maps to make this unnecessary. I don't remember.

PR #2322, opened 2021-12-10
Fixes pvs culling & make budgets more granular

PaulRitter:
split the entitybudget up into a total streaming budget (entities entering pvs), as well as a new entity budget (entities never before seen by client -> will need to be allocated)

also refactors pvs culling to just send a transformcompstate sending the entitiy to nullspace. this should fix the bugs people saw of stuff disappearing.

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -23,6 +23,7 @@ internal partial class PVSSystem : EntitySystem
     [Shared.IoC.Dependency] private readonly IMapManager _mapManager = default!;
     [Shared.IoC.Dependency] private readonly IPlayerManager _playerManager = default!;
     [Shared.IoC.Dependency] private readonly IConfigurationManager _configManager = default!;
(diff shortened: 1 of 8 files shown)


Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -396,22 +410,32 @@ private bool TryAddToVisibleEnts(EntityUid uid, Dictionary<EntityUid, PVSEntityV
...
             return false; //we failed? suppose we dont get added either
 
         //did we already get added through the parent call?
         if (toSend.ContainsKey(uid)) return true;
 
-        //did the previous tick see us?
-        var @new = !previousVisibleEnts.Remove(uid);
+        //are we new?
+        var @new = !seenSet.Contains(uid);
+        var entered = @new | !previousVisibleEnts.Remove(uid);
```
metalgearsloth, line 421: I assume intended?

gradientvera: image

PR #2324, opened 2021-12-11
Some minor PVS optimizations

gradientvera:
Reduces some redundant dictionary lookups.

`EntityExists` and `GetComponent<MetaDataComponent>` can be turned into a single `TryGetComponent<MetaDataComponent>` check
`GetEntities()` and `GetComponent<MetaDataComponent>` can be turned into an `EntityQuery<MetaDataComponent>()`


metalgearsloth (approved):
im not hot but I can review

PR #2336, opened 2021-12-13
Optimise SharedGridTraversal slightly

metalgearsloth:
At least for the typical use case of something being parented to the grid or map.

The other half was the getcomp on the container check which can't be optimised out when it needs to happen (unless we cache "in container" on all children).


metalgearsloth:
How the hell is clickable failing from this.

PR #2340, opened 2021-12-13
UserInterface/RichTextEntry: Rich Text Fixes (Round 2)

Efruit:
Fixes:
- Whatever the hell was going wrong with the Tool status tooltip thing
- Spacing issues in `RoundEndSummaryWindow`

PR #2343, opened 2021-12-14
Rich text cleanup

Efruit:
~~Includes and requires #2340.~~ _(Since merged to master.)_

### DIFF
--- a/Robust.Client/Console/Completions.cs
+++ b/Robust.Client/Console/Completions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using Robust.Client.UserInterface.Controls;
 using Robust.Client.UserInterface.CustomControls;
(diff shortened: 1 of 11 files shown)


ElectroJr (commented):
Fiy there is also some sort of issue with `RichTextLabel` fonts that this PR doesn't yet seem to have fixed. The issue I've noticed is that for whatever reason the verbs in the verb menu don't have their fonts applied (bold/italic/etc).

Maybe unrelated, but `RichTextLabel._getFont()` ignores the "font" style property. It does a TryGet but does nothing with it.

Efruit:
@ElectroJr Very nice catch. Haven't tested it yet, but I've included a fix in #2340 @ https://github.com/space-wizards/RobustToolbox/pull/2340/commits/bc1454f9bba07c5696099eef9443cfd7993da6df

Efruit:
No luck, they're still style-less. But it did fix the tool status text!

PR #2348, opened 2021-12-15
Add nullable overrides for the the Deleted() & Exists() proxy methods.

ElectroJr:
The `EntityManager.Deleted()` & `EntityExists()` functions have overrides that take nullable uids with `NotNullWhen` attributes for the input. This just adds similar overrides for the new proxy methods.

### DIFF
--- a/Robust.Shared/GameObjects/EntitySystem.Proxy.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.Proxy.cs
@@ -19,6 +19,13 @@ public partial class EntitySystem
         return EntityManager.EntityExists(uid);
     }
 
+    /// <inheritdoc cref="IEntityManager.EntityExists" />
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public bool Exists([NotNullWhen(true)] EntityUid? uid)
+    {
+        return EntityManager.EntityExists(uid);
+    }
+
     /// <summary>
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/EntitySystem.Proxy.cs
```diff
@@ -19,6 +19,13 @@ public bool Exists(EntityUid uid)
         return EntityManager.EntityExists(uid);
     }
 
+    /// <inheritdoc cref="IEntityManager.EntityExists" />
+    [MethodImpl(MethodImplOptions.AggressiveInlining | MethodImplOptions.AggressiveOptimization)]
+    public bool Exists([NotNullWhen(true)] EntityUid? uid)
+    {
+        return EntityManager.EntityExists(uid);
+    }
```
ElectroJr, line 27: So obviously the nullable version could just functionally replace the non-nullable version. But I don't know if that impacts performance, which might matter given how often this function gets called.

Would it waste time making a non-nullable `EntityUid` nullable and then check `HasValue()`? or is the compiler black magic smart enough to not do that?

gradientvera: It's better to keep them separate, it could hurt performance a bit ye

gradientvera (approved):

gradientvera:
also for the record, the aggressive optimization flag isn't ideal-- but I have yet to remove it oops

PR #2352, opened 2021-12-17
UserInterface/TextLayout: Add word splitting for rich text

Efruit:
↑

Requires and includes #2351

### DIFF
--- a/Robust.Client/UserInterface/TextLayout.cs
+++ b/Robust.Client/UserInterface/TextLayout.cs
@@ -355,7 +355,6 @@ namespace Robust.Client.UserInterface
                     else if (wq.Work.wt != WordType.Normal)
                         wq.Flush();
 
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/TextLayout.cs
```diff
@@ -313,6 +355,7 @@ private static (int gapPri, int adv) TransitionWeights (TextAlign l, TextAlign r
                     else if (wq.Work.wt != WordType.Normal)
                         wq.Flush();
 
+                    #warning TODO: Check this w/ someone that knows C# string semantics
                     sbo += r.Utf16SequenceLength;
```
Efruit: This is how you're supposed to track the offset in a unicode `string`, right?
Won't this make sure `sbo` always points to the start of a `Rune`?

PR #2327, opened 2021-12-12
Broadphase fix take 2

metalgearsloth:
Original PR https://github.com/space-wizards/RobustToolbox/pull/2318

PR #2354, opened 2021-12-18
Fix destroyproxies crashing when getting transform

wrexbe:
When a ghost was on top of something, and it gets deleted, this was crashing,  because the transform was already gone. This seems to work better for getting the map id (it's not crashing), but I don't know enough about the system to know if this alternative way of getting it is ok.

### DIFF
--- a/Robust.Shared/Physics/SharedBroadphaseSystem.cs
+++ b/Robust.Shared/Physics/SharedBroadphaseSystem.cs
@@ -775,7 +775,7 @@ namespace Robust.Shared.Physics
             }
 
             var proxyCount = fixture.ProxyCount;
-            var moveBuffer = _moveBuffer[_entityManager.GetComponent<TransformComponent>(broadphase.Owner).MapID];
+            var moveBuffer = _moveBuffer[fixture.Body.PhysicsMap!.MapId];
 
             for (var i = 0; i < proxyCount; i++)
             {


Robust.Shared/Physics/SharedBroadphaseSystem.cs
```diff
@@ -775,7 +775,7 @@ internal void DestroyProxies(BroadphaseComponent broadphase, Fixture fixture)
             }
 
             var proxyCount = fixture.ProxyCount;
-            var moveBuffer = _moveBuffer[_entityManager.GetComponent<TransformComponent>(broadphase.Owner).MapID];
+            var moveBuffer = _moveBuffer[fixture.Body.PhysicsMap!.MapId];
```
metalgearsloth, line 778: As a bandaid it's fine but the fact these are different is a bug that should be fixed.

wrexbe: I think this is an improvement, because it's using it's own arguments, instead of looking up something it already has, even if the real bug is that the deletion is out of order somewhere.

metalgearsloth: Ideally I'd prefer PhysicsMap not to exist at all long-term but that's a whole project on its own.

metalgearsloth: For now just add an assert that they match IMO.

wrexbe: I can only do that if the component is not null

wrexbe:
Sloths PR resolves this issue in a different way
https://github.com/space-wizards/RobustToolbox/pull/2327

PR #2363, opened 2021-12-20
Revert Rich Text

PaulRitter:
@Efruit we (maintainers) talked about the rich text refactor and decided we'd rather you open one big pr, refactoring rich text, instead of you piecemealing features to master. hope you understand.


Efruit:
I'M FREE

Efruit:
For the record, it was one big PR (plus bugfixes as discovered), and that was probably half the problem.

PR #2365, opened 2021-12-21
Entity deletion stuff

metalgearsloth:
- ToArray to be consistent with the non-generic method
- Avoid unnecessary GetComponent<TransformComponent> on children
- Pass in IEntityManager ref to containers

### DIFF
--- a/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -54,7 +54,7 @@ namespace Robust.Shared.GameObjects
         [ViewVariables(VVAccess.ReadWrite)]
         internal bool ActivelyLerping { get; set; }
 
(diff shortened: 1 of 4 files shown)


Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -54,7 +54,7 @@ public sealed class TransformComponent : Component, IComponentDebug
         [ViewVariables(VVAccess.ReadWrite)]
         internal bool ActivelyLerping { get; set; }
 
-        [ViewVariables] private readonly SortedSet<EntityUid> _children = new();
+        [ViewVariables] internal readonly SortedSet<EntityUid> _children = new();
```
metalgearsloth, line 57: Ideally transformcomp would use friends but that's too powerful at this stage. Iterating the children directly is extremely useful as the Children property is much slower.

PaulRitter (approved):

PR #2367, opened 2021-12-23
Engine changes for more server-pushed lighting variables

Matz05:
More of the variables on a client-side PointLight are now replicated from the server.
There's some controversy over whether this is the right way to do things, and even if it is I'm not sure I did it right as I just imitated existing structures in the code with limited understanding of their inner workings, but it seems to work in my testing.
This is necessary for bulbs inserted into PoweredLights to change more than just the colour, and have the values replicate to clients correctly, using the same logic already used for colour changes. An implementation of this is found in the corresponding content PR.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
@@ -90,7 +90,7 @@ namespace Robust.Shared.GameObjects
 
         [ViewVariables(VVAccess.ReadWrite)]
         [Animatable]
-        public virtual float Energy
+        public float Energy
         {
             get => _energy;
             set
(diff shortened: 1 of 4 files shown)


Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
```diff
@@ -81,6 +87,36 @@ public Vector2 Offset
...
+        {
+            get => _softness;
+            set
+            {
+                if (_softness.Equals(value)) return;
+                _softness = value;
+                Dirty();
+            }
+        }
     }
```
metalgearsloth: These don't need to be virtual.

metalgearsloth (changes requested):
Some of the others are virtual because the rendertree needs to update their bounds (and for color it can probably be removed).

Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
Matz05: Fixed in the following commit.

PaulRitter (approved):

metalgearsloth (approved):

PR #2368, opened 2021-12-24
Adds help method to C# scripting.

gradientvera:
Because my memory sucks.

- Added `help` method to scripting, which returns all scripting global methods and properties with proper highlighting.
- Added `writesyntax` method to scripting. It allows you to write anything with C# syntax highlighting to the scripting output. Please note that this method compiles whatever you pass into it to a C# script to perform highlighting, making it fairly slow.
- Added various extension methods to pretty-print `MethodInfo`, `PropertyInfo`, `ParameterInfo` and `Type` with their proper C# signatures.
- Added `IsEmpty` property to `FormattedMessage`.  
- Changed `ScriptInstance`'s `OutputBuffer` to a `FormattedMessage`, for color output purposes.

Peek 24-12-2021 14-51
Peek 24-12-2021 14-27

### DIFF
--- a/Robust.Client/Console/ScriptConsoleClient.cs
+++ b/Robust.Client/Console/ScriptConsoleClient.cs
@@ -210,7 +210,7 @@ namespace Robust.Client.Console
                 vvm.OpenVV(a);
             }
 
-            protected override void writesyntax(object toString)
+            protected override void WriteSyntax(object toString)
             {
                 var code = toString.ToString();
 
--- a/Robust.Client/Console/ScriptConsoleClient.cs
+++ b/Robust.Client/Console/ScriptConsoleClient.cs
@@ -210,7 +210,7 @@ namespace Robust.Client.Console
                 vvm.OpenVV(a);
             }
 
-            public override void writesyntax(object toString)
+            protected override void writesyntax(object toString)
             {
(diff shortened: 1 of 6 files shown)


PaulRitter (approved):
lgtm but i'd like @PJB3005 to review this too

Robust.Client/Console/ScriptConsoleClient.cs
```diff
@@ -210,6 +210,23 @@ public override void vv(object a)
                 vvm.OpenVV(a);
             }
 
+            public override void writesyntax(object toString)
```
PJB3005: Does this really need to be a helper method on script globals? When are you ever outputting C# code to the console outside of the help command?

gradientvera: The only reason I did it this way is because the client and server output this in a different way, and it seemed the cleanest way to do it since `help` is in shared. I could make this a protected method instead and rename it tho

gradientvera: I made it protected, should be fine now

gradientvera:
PJB said I could merge it, but in return I'll have to play xonotic with him in the future, sometime:tm:

PR #2372, opened 2021-12-26
The Return of Rich Text (Core)

Efruit:
Re-introduces `TextLayout` and friends, but in a manner that doesn't require blowing up the whole codebase.


Efruit:
woe rich text upon ye


@PJB3005

PR #2373, opened 2021-12-27
Adds config file support for Content.Client

ike709:
This is just lifted directly from an old commit on Acruid's fork.

Useful for OpenDream development so we don't have to tell people to pass any cvars as program arguments.

### DIFF
--- a/Robust.Client/GameController/GameController.cs
+++ b/Robust.Client/GameController/GameController.cs
@@ -281,6 +281,22 @@ namespace Robust.Client
 
             if (Options.LoadConfigAndUserData)
             {
+                // Load content config
+                if (_resourceCache.TryGetDiskFilePath(new ResourcePath(Options.ContentConfigFileName),
+                        out var contentPath))
+                {
+                    _configurationManager.LoadFromFile(contentPath);
+                }
+                else
+                {
+                    var contentFile = Path.Combine(userDataDir, Options.ContentConfigFileName);
+                    if (File.Exists(contentFile))
+                    {
+                        _configurationManager.LoadFromFile(contentFile);
+                    }
+                }
+
+                // Load client config
                 var configFile = Path.Combine(userDataDir, Options.ConfigFileName);
                 if (File.Exists(configFile))
                 {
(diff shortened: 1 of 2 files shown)


Robust.Client/GameController/GameController.cs
```diff
@@ -322,6 +322,10 @@ internal bool StartupSystemSplash(GameControllerOptions options, Func<ILogHandle
                 Options.AssemblyDirectory,
                 Options.LoadContentResources, _loaderArgs != null && !Options.ResourceMountDisabled, ContentStart);
 
+            // Load content config
+            if(_resourceCache.TryGetDiskFilePath(new ResourcePath(Options.ContentConfigFileName), out var contentPath))
+                _configurationManager.LoadFromFile(contentPath);
```
PJB3005: 1. This should be loaded *before* appdata config, or at least not override the settings in it.
2. You should make this work even if there's no disk path available. It would be really confusing if this file only worked in development builds.

PJB3005 (changes requested):

metalgearsloth:
@ike709 you coming back to this?

ike709:
oops I forgot it existed, yeah I'll get back to it soon

Robust.Client/GameController/GameController.cs
```diff
@@ -281,6 +281,22 @@ internal bool StartupSystemSplash(GameControllerOptions options, Func<ILogHandle
...
             {
+                // Load content config
+                if (_resourceCache.TryGetDiskFilePath(new ResourcePath(Options.ContentConfigFileName),
+                        out var contentPath))
+                {
+                    _configurationManager.LoadFromFile(contentPath);
+                }
+                else
+                {
+                    var contentFile = Path.Combine(userDataDir, Options.ContentConfigFileName);
```
ike709, line 292: If this isn't what PJB meant then I have no idea what he did mean.

PJB3005: This still isn't loading the config file from VFS, now you just made it fall back and try to load it from AppData if VFS is a zip file.

PJB3005: You need to somehow allow ConfigurationManager to load from a stream (or something) and then read it in with `IResourceManager.ContentFileRead()`. None of this.

PR #2375, opened 2021-12-27
ByRefEvent attribute for events.

### LINKED ISSUE #2266: By-ref events should use an attribute.
As discussed on the September 19th meeting, by-ref events should use an attribute so working with them is less painful and you can tell at a glance if they're by-ref or by-value.

gradientvera:
Fixes #2266
Requires you to mark by-ref events with an attribute to use them as such.
Really just for convenience, so you can tell if an event is by-ref or not just by looking at its declaration.
Opens the possibility for a roslyn analyzer that checks by-ref vs by-value subscriptions in the future.

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -628,6 +628,7 @@ internal partial class PVSSystem : EntitySystem
     }
 }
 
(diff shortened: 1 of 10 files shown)


PaulRitter (approved):
now to add a funny analyzer

PR #2378, opened 2021-12-28
No events on stopping components

wrexbe:
This change prevents a system from receiving an event for a component if its life cycle is stopping, or later. The lifecycle events seem to go through the bus on a different path, so they are unaffected by this change.

### DIFF
--- a/Robust.Shared/GameObjects/EntityEventBus.Directed.cs
+++ b/Robust.Shared/GameObjects/EntityEventBus.Directed.cs
@@ -389,6 +389,8 @@ namespace Robust.Shared.GameObjects
                     var (component, reg) = tuple.Value;
                     if (reg.ReferenceEvent != dispatchByReference)
                         ThrowByRefMisMatch();
+                    
+                    if (component.LifeStage >= ComponentLifeStage.Stopping) continue;
 
                     reg.Handler(euid, component, ref args);
                 }


ShadowCommander:
Does player detached and mind detached still work?

gradientvera:
What's the rationale behind this?

wrexbe:
@ShadowCommander Seems to be fine, I went ghost, and back, suicided, and respawned.
@Zumorica It prevents about 7500 events from going off when the round restarts, and it lets a unit test pass in a different PR I made that is failing because a power event is happening after the entity is partially deleted.

metalgearsloth:
I've already run into issues where the eventbus is silently failing on other events so I'm not a huge fan of the change https://github.com/space-wizards/space-station-14/issues/5609

Acruid (changes requested):
https://github.com/space-wizards/space-station-14/blob/master/Content.Server/Atmos/EntitySystems/AirtightSystem.cs#L58
Existing code depends on the ability to raise an event from inside the `ComponentShutdown` and `ComponentRemove` lifestage events. This atmos code would break if this PR is merged.

> It prevents about 7500 events from going off when the round restarts

If this can be optimized, it needs to be done on a per-system basis, and address *why* that is happening.

> failing because a power event is happening after the entity is partially deleted.

You can check the component's lifestage inside the event for the same effect.

This PR as it is right now should not be merged.

PaulRitter:
> This PR as it is right now should not be merged.


damn, if even acruid is saying it.

jokes aside though, i also think this is a bad idea, same reasoning as sloth.

PR #2380, opened 2021-12-29
Run appearancevisualizers on dummy entities again

metalgearsloth:
The issue is that appearancesystem defers this until frameupdate but we're spawning the entity and immediately deleting it.

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -2074,6 +2074,11 @@ namespace Robust.Client.GameObjects
             var dummy = entityManager.SpawnEntity(prototype.ID, MapCoordinates.Nullspace);
             var spriteComponent = entityManager.EnsureComponent<SpriteComponent>(dummy);
 
+            if (entityManager.TryGetComponent(dummy, out ClientAppearanceComponent? appearanceComponent))
+            {
+                EntitySystem.Get<AppearanceSystem>().OnChangeData(appearanceComponent);
+            }
+
             var anyTexture = false;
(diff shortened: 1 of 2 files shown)


gradientvera (approved):
Sure, sounds good to me

Robust.Client/GameObjects/EntitySystems/AppearanceSystem.cs
```diff
@@ -21,13 +21,17 @@ public override void FrameUpdate(float frameTime)
...
-                {
-                    visualizer.OnChangeData(appearance);
-                }
-
+                OnChangeData(appearance);
                 appearance.UnmarkDirty();
             }
         }
+
+        public void OnChangeData(ClientAppearanceComponent appearanceComponent)
```
PaulRitter: smth smth entityuid smth smth resolve pattern

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -2074,6 +2074,11 @@ public static IEnumerable<IDirectionalTextureProvider> GetPrototypeTextures(Enti
             var dummy = entityManager.SpawnEntity(prototype.ID, MapCoordinates.Nullspace);
             var spriteComponent = entityManager.EnsureComponent<SpriteComponent>(dummy);
 
+            if (entityManager.TryGetComponent(dummy, out ClientAppearanceComponent? appearanceComponent))
+            {
+                EntitySystem.Get<AppearanceSystem>().OnChangeData(appearanceComponent);
+            }
```
PaulRitter: if you'd use the resolve pattern this wouldn't be needed

PaulRitter (changes requested):

PR #2382, opened 2021-12-29
add CanInsertIfEmpty()

ElectroJr:
This adds a function to `ContainerSlot` that allows you to check whether an item could be inserted into a container slot if it were empty.

Currently, you need to actually empty the slot before `CanInsert()` provides useful information, which is annoying if you want know whether you could swap out an item.


PaulRitter (changes requested):
caninsert should call this instead of base.caninsert to avoid having to duplicate code later on

PR #2384, opened 2021-12-30
TeleportTo upgrades

ShadowCommander:
Cleaned up duplicate code in the TeleportTo command.
Add teleporting to EntityUids in addition to usernames.
Renamed TeleportToPlayerCommand to TeleportToCommand.


PaulRitter (dismissed):
move the uname/uid parsing to a seperate function to avoid duplicated code. otherwise lgtm

PaulRitter (approved):

PR #2386, opened 2021-12-31
Automatic Component Names

wrexbe:
Figure out, and cache name per type.

### DIFF
--- a/Robust.Shared/GameObjects/Component.cs
+++ b/Robust.Shared/GameObjects/Component.cs
@@ -22,13 +22,32 @@ namespace Robust.Shared.GameObjects
         {
             get
             {
-                if (Attribute.GetCustomAttribute(GetType(), typeof(ComponentProtoNameAttribute)) is ComponentProtoNameAttribute attribute)
-                    return attribute.PrototypeName;
-
-                // Legacy code requires all components have a name, even though this should not be required.
-                throw new InvalidOperationException($"{GetType().FullName} does not have a defined Data Name.");
+                var type = GetType();
+                if (!DefaultNames.TryGetValue(type, out string? name))
+                {
+                    name = CalculateDefaultName(type);
+                    DefaultNames.Add(type, name);
+                }
+                return name;
             }
         }
+
+        private static string CalculateDefaultName(Type type)
+        {
+
+            if (Attribute.GetCustomAttribute(type, typeof(ComponentProtoNameAttribute)) is ComponentProtoNameAttribute attribute)
+                return attribute.PrototypeName;
+            const string component = "Component";
+            var typeName = type.Name;
+            if (typeName == component) return component;
+            if (typeName.EndsWith(component, StringComparison.Ordinal))
+            {
+                return typeName[..^component.Length];
+            }
+            return typeName;
+        }
+
+        private static Dictionary<Type, string> DefaultNames = new();
 
         /// <inheritdoc />
         [ViewVariables]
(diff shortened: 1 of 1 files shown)


mirrorcult (approved):

Robust.Shared/GameObjects/Component.cs
```diff
@@ -22,14 +23,33 @@ public virtual string Name
...
+            var typeName = type.Name;
+            if (typeName == component) return component;
+            if (typeName.EndsWith(component, StringComparison.Ordinal))
+            {
+                return typeName[..^component.Length];
             }
+            return typeName;
         }
 
+        private static Dictionary<Type, string> DefaultNames = new();
```
PJB3005: Use a `ConcurrentDictionary` for this. This approach isn't thread safe at all.

PJB3005: Also make this `readonly`.

PJB3005 (changes requested):

Acruid:
* Being able to break prototypes by changing the class name of a component is pretty terrible.
* How is this supposed to work with shared prototypes that need a name to map to both `ServerFooComponent` and `ClientFooComponent`?
* If I name one component `Bar` and have another completely different unnamed component with the class name `BarComponent`, won't that collide?
* What is the reason for caching the names?

wrexbe:
@Acruid 
> Being able to break prototypes by changing the class name of a component is pretty terrible.
- Almost every single string is just the same text as the classes name currently, and having all of them repeat their own name in a string is terrible, it's a lot of boilerplate for nothing.
- If a PR changes a class name for a component, it'll be easy to see. Changing a class name is going to show up in several files, and will be deliberate. Changing one character in a string, on the other hand is much easier to miss.
- Having the class, and file name match what is in the yml can be considered a feature, because it makes finding it easier.
- The system has to create, and throw away each component currently to figure out the name.

> If I name one component Bar and have another completely different unnamed component with the class name BarComponent, won't that collide?
- Realistically, if I was to do that, someone else would reject the PR, and tell me to rename the class. In this hypothetical situation, you really should just come up with a new name for the class, it's going to be confusing, even without this change.
- If your following what most of the components are currently doing, you would stick "Bar" in the string, and have the same collision the system has today.
- If you really needed the class named like that, you would add the attribute. This change is to remove the 400 components with names that match, not ones that are trying to be named weird.

> How is this supposed to work with shared prototypes that need a name to map to both ServerFooComponent and ClientFooComponent?
- This PR isn't removing the existing way of naming things with attributes, it's just making it unnecessary 99% of the time.
- They should just not have Client, and Server in the name, for the same reason every class in the Client project doesn't start with Client. It is really easy to tell if the file is on the server, or client, it doesn't need to have it in the class name.
- If supporting this kind of name to also be automatic is wanted, changing the auto-generated name to look for Client, Server, and Shared would be easy enough.

> What is the reason for caching the names?

- The reason for caching names is for performance, and to avoid allocating new strings whenever it replaces the word Component. I'm not sure how frequently these get called, but caching it makes the cost nearly free, so nobody needs to worry about it.

Robust.Shared/GameObjects/Component.cs
wrexbe: That made it look nicer.

metalgearsloth:
> it's just making it unnecessary 99% of the time.

I feel like it's better for new people to either require it 100% of the time or not at all, and it's not possible to not require it at all due to component inheritance existing (and yes, we do need it even if we try to avoid it).

> Having the class, and file name match what is in the yml can be considered a feature, because it makes finding it easier.

Has this ever been an issue for ctrl-F?

> The reason for caching names is for performance,

I've never seen it on a profiler but it's probably a non-issue.

wrexbe:
Closing, in favor of alternate implementation.
https://github.com/space-wizards/RobustToolbox/pull/2389

PR #2389, opened 2022-01-02
Automatic Component Names (Alternate ending)

wrexbe:
This is a different implementation of automatic component names, based off of @Acruid's suggestions.
I also have it handle Client, Server, and Shared in the component name, and enforce components ending with Component, if they are not manually overriding the name with an attribute.

For backward compatibility, if it detects the name is overridden, it falls back to creating the instance to get the name.


Robust.Shared/GameObjects/Component.cs
```diff
@@ -17,18 +17,8 @@ namespace Robust.Shared.GameObjects
...
-            {
-                if (Attribute.GetCustomAttribute(GetType(), typeof(ComponentProtoNameAttribute)) is ComponentProtoNameAttribute attribute)
-                    return attribute.PrototypeName;
-
-                // Legacy code requires all components have a name, even though this should not be required.
-                throw new InvalidOperationException($"{GetType().FullName} does not have a defined Data Name.");
-            }
-        }
+        [ViewVariables(VVAccess.ReadOnly)]
+        public string Name => IoCManager.Resolve<IComponentFactory>().GetComponentName(GetType());
```
Acruid: Needs to be virtual if you want to keep compatibility with content (and to fix failing tests).

Acruid (approved):

Robust.Shared/GameObjects/Component.cs
wrexbe: Yeah, I have an idea. I'm doing it

Robust.Client/GameObjects/ClientComponentFactory.cs
```diff
@@ -29,13 +29,9 @@ public ClientComponentFactory(IDynamicTypeFactoryInternal typeFactory, IReflecti
             RegisterClass<EyeComponent>();
             RegisterClass<AnimationPlayerComponent>();
             RegisterClass<TimerComponent>();
-
-#if DEBUG
```
PaulRitter, line 33: this

Robust.Server/GameObjects/ServerComponentFactory.cs
```diff
@@ -29,12 +29,9 @@ public ServerComponentFactory(IDynamicTypeFactoryInternal typeFactory, IReflecti
             RegisterClass<ServerUserInterfaceComponent>();
             RegisterClass<TimerComponent>();
             RegisterClass<MapSaveIdComponent>();
-
-#if DEBUG
```
PaulRitter, line 33: this

Robust.Shared/GameObjects/Components/DebugExceptionComponents.cs
```diff
@@ -1,9 +1,7 @@
 using System;
-using Robust.Shared.Serialization;
 
 namespace Robust.Shared.GameObjects
 {
-#if DEBUG
```
PaulRitter, line 6: this

PaulRitter (changes requested):
why remove the compiler-if?

PR #2391, opened 2022-01-03
EntityManager methods for working with components based on name

mirrorcult:
See https://discord.com/channels/310555209753690112/560845886263918612/927682504956211250

This is used pretty rarely (I think some wizard spell does it in SS14, and status effects do it as well) but it's still pretty useful and reduces boilerplate a good amount in those cases. Used rarely enough that I didn't feel the need to add entitysys proxy methods for these

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.Components.cs
+++ b/Robust.Shared/GameObjects/EntityManager.Components.cs
@@ -192,6 +192,15 @@ namespace Robust.Shared.GameObjects
             return newComponent;
         }
 
+        public void AddComponent(EntityUid uid, string name)
+        {
+            if (!_componentFactory.TryGetRegistration(name, out _))
+                throw new ArgumentException("Component name is invalid.", name);
+            var newComponent = (Component) _componentFactory.GetComponent(name);
+            newComponent.Owner = uid;
+            AddComponent(uid, newComponent);
+        }
+
         public void AddComponent<T>(EntityUid uid, T component, bool overwrite = false) where T : Component
         {
             if (!uid.IsValid() || !EntityExists(uid))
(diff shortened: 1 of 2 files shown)


Acruid:
> Random artifact should pick one of the components which will define how artifact will be activated

This is the terrible legacy way to do it, with the equivalent of `IActivate` interfaces.

> So I have CompA, CompB, CompC that should be added at runtime.

Needing a separate component for each interaction permutation is a bad idea.

You solution would work, but it's trying to solve a problem that is a bad idea in the first place. I bet the spells and status effects are doing the same things (component per spell/status effect in the strategy pattern), which is why content code would actually need a way to bind components to yaml using the component name.

> "Yeah, but my case is so unique, I don't think it will be widely used anywhere else"

https://en.wikipedia.org/wiki/Strategy_pattern, you will find this is many places, it's one of the basic 23 patterns.

Basically merging this is just going to reinforce bad code. This is going the opposite direction of https://github.com/space-wizards/RobustToolbox/pull/2389.

mirrorcult:
binding to a component using a datafield and just taking the name in YAML would be nice but I don't really think we have a way to do that easily?

mirrorcult:
that doesn't really solve everything though, statuseffects in content has to have some way to send which component is being dynamically added/removed over the wire and the name works for that. i guess that could use the netid instead of the name and then this isn't an issue?

PaulRitter:
> binding to a component using a datafield and just taking the name in YAML would be nice but I don't really think we have a way to do that easily?

afaict componentmanager is already initialized at the time of serialization, so this would very much be possible, but what do you mean by binding? resolving it into a type? or resolving it into a componentdependency of sorts?

PaulRitter (changes requested):
i agree with acruid. if anything, we should add a customtypeserializer for <Type, ValueDataNode> that'll resolve componentnames into a type.

if you want, i can quickly add that, shouldn't be too hard

mirrorcult:
this is prolly a bad idea

PR #2394, opened 2022-01-04
Expose ConsoleHost.RegisteredCommand and its callback.

gradientvera:
Needed so content can retrieve attributes specified in the callback method.
Required by https://github.com/space-wizards/space-station-14/pull/6025

### DIFF
--- a/Robust.Shared/Console/ConsoleHost.cs
+++ b/Robust.Shared/Console/ConsoleHost.cs
@@ -108,9 +108,9 @@ namespace Robust.Shared.Console
         /// A console command that was registered inline through <see cref="IConsoleHost"/>.
         /// </summary>
         [Reflect(false)]
-        private class RegisteredCommand : IConsoleCommand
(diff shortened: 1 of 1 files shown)


Acruid:
We already have `AdminFlags` -> `IConsoleCommand` mappings defined in the congroup perm files, why is content trying to redefine them in code? It would be better to use the existing system than to make a completely separate one.

gradientvera:
Convenience. I'd rather slap an attribute in-code than write the command name to some YAML file.
This is something we use *heavily* in content, I'm just making it available for RegisteredCommand.

Acruid (approved):
There are already multiple ways for content to bypass console security and directly execute console commands. This adds another way, but at this point it does not really matter. IF we want to address this issue in the future, it's better to redesign the entire system with this PR's requirement.

As far as I can tell the security bypass is the only potential problem with this, if content wants to hard code concommand permissions they can. Because this security bypass isn't any worse than what we have now, there is no real reason to deny this PR.

PR #2402, opened 2022-01-07
Make entity manager proxy methods protected

ElectroJr:
Non-systems can just use entity manager directly. Also allows `HasComp` to take in nullable entity UIDs.

See https://discord.com/channels/310555209753690112/310555209753690112/925981399230586920
discord


mirrorcult (approved):

PaulRitter (approved):
rising from the grave solely to merge this

PR #2407, opened 2022-01-09
Entity lookup fixes

metalgearsloth:
Needed for loadmap to work again with docked stuff. The main fix is doing it on EntityInit and not EntityStartup.

EntityLookup is still pretty fucken shit but it's 1% less shit.


Robust.Shared/GameObjects/Systems/EntityLookup.cs
```diff
@@ -791,17 +797,10 @@ public bool UpdateEntityTree(EntityUid entity, Box2? worldAABB = null)
...
 
-            if (float.IsNaN(center.X) || float.IsNaN(center.Y))
-            {
-                RemoveFromEntityTrees(entity);
-                return true;
-            }
-
-            var transform = _entityManager.GetComponent<TransformComponent>(entity);
-            DebugTools.Assert(transform.Initialized);
+            DebugTools.Assert(!float.IsNaN(center.X) && !float.IsNaN(center.Y));
```
mirrorcult, line 803: is there a reason these can be replaced with asserts rather than explicit checks? is it just cause entity lookup gets ensured or were these fixed earlier and just not removed until now

mirrorcult (approved):

Robust.Shared/GameObjects/Systems/EntityLookup.cs
metalgearsloth: They existed because PVS existed and actively used NaNs. Ideally they should never happenTM.

PR #2408, opened 2022-01-09
Fix lifestage for loading initialized map

metalgearsloth:
These weren't getting marked as MapInitialized.

I also deduplicated most of loadbp and loadmap as ideally they function identically with loadbp throwing if more than 1 grid is on the requested map file.

### DIFF
--- a/Robust.Server/Maps/MapLoader.cs
+++ b/Robust.Server/Maps/MapLoader.cs
@@ -68,6 +68,10 @@ namespace Robust.Server.Maps
             stream.Save(new YamlMappingFix(new Emitter(writer)), false);
         }
 
+        /*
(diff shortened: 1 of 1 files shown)


Robust.Server/Maps/MapLoader.cs
```diff
@@ -68,6 +68,10 @@ public void SaveBlueprint(GridId gridId, string yamlPath)
             stream.Save(new YamlMappingFix(new Emitter(writer)), false);
         }
 
+        /*
+         * TODO: LoadBlueprint and Loadmap should be almost identical methods but right now it's copypaste spaghet everywhere
+         */
+
```
metalgearsloth:
```suggestion

```

metalgearsloth: FFS I committed this fork.

PR #2410, opened 2022-01-09
Joint serialization

metalgearsloth:
Only did welds for now because it's used for docking.


metalgearsloth:
@PaulRitter you're probably the best person to review this?

PaulRitter (approved):

PR #2414, opened 2022-01-10
Raise an attempt-event when receiving client BUI messages.

ElectroJr:
A general problem in SS14 is that most BUI don't check action blockers when receiving UI messages from the client. For example currently a dead person next to the nuke can enter the code and arm it (assuming they had the UI open before dying). Instead of expecting every system to individually check action blockers, it makes sense to check them before passing messages along to systems/components.

This PR adds an attempt-event so that the interaction system can do that, along with an opt-out data-field in case someone does actually need a ghost/dead interactable UI. It uses an attempt event instead of directly checking action blockers because those are all SS14/content features.

As to why it should check individual messages rather than just closing the UI when players die / can no longer interact: because IMO ghosts/spectators should be able to observe consoles, PDAs, etc, but obviously shouldn't be able to modify them.

Currently this does lead to some jank, where ghosts can edit text/spin boxes which might give the illusion of allowing input, but it doesn't have any actual effect. Ideally those GUI elements would all be disabled by the client by checking the same attempt-event when opening the UI.

### DIFF
--- a/Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
+++ b/Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
@@ -7,9 +7,9 @@ using Robust.Server.Player;
 using Robust.Shared.Enums;
 using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
-using Robust.Shared.Log;
 using Robust.Shared.Serialization;
 using Robust.Shared.Serialization.Manager.Attributes;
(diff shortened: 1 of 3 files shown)


mirrorcult:
the eventual solution is to probably just move BUIs to content but this is good until then

metalgearsloth (approved):

PaulRitter:
> the eventual solution is to probably just move BUIs to content but this is good until then

why would we move buis to content? imo this is definitely something other games could gain alot from.

PR #2418, opened 2022-01-12
Prevent disembowelling on tile removal.

ElectroJr:
Slight oversight.

Now also checks if the entity center is on the tile.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedTransformSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedTransformSystem.cs
@@ -67,7 +67,7 @@ namespace Robust.Shared.GameObjects
             var grid = _mapManager.GetGrid(gridId);
             var gridUid = grid.GridEntityId;
             var mapTransform = Transform(_mapManager.GetMapEntityId(grid.ParentMapId));
-            var aabb = new Box2(tileIndices * grid.TileSize, (tileIndices + 1) * grid.TileSize);
+            var aabb = _entityLookup.GetLocalBounds(tileIndices, grid.TileSize);
 
             foreach (var entity in _entityLookup.GetEntitiesIntersecting(gridId, tileIndices).ToList())
             {
--- a/Robust.Shared/GameObjects/Systems/SharedTransformSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedTransformSystem.cs
@@ -67,35 +67,13 @@ namespace Robust.Shared.GameObjects
             var grid = _mapManager.GetGrid(gridId);
             var gridUid = grid.GridEntityId;
             var mapTransform = Transform(_mapManager.GetMapEntityId(grid.ParentMapId));
-
-            // Given that we need to check if the entity's CENTER is on the tile, and not just whether it intersects,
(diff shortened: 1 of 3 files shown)


metalgearsloth:
Something else I thought of is that this doesn't consider whether the center of the entity is over the tile or not, just if any part of it is touching. I think the best solution for that is probably add a flag to entitylookup to support it and we can deal with it later.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.cs
```diff
@@ -58,15 +58,44 @@ private void MapManagerOnTileChanged(object? sender, TileChangedEventArgs e)
...
+
+            _entityLookup.FastEntitiesIntersecting(lookup, ref aabb, (uid) =>
+            {
+                if (Deleted(uid))
+                    return;
+
+                var transform = Transform(uid);
+                if (transform.ParentUid == gridUid && aabb.Contains(transform.LocalPosition))
+                    results.Add(transform); // cannot de-parent directly, else modified-while-enumerating error.
+            });
```
metalgearsloth: Can this not use the entitylookup method?

ElectroJr: My loose thought process was that if eventually `GetEntitiesIntersecting` were updated to get the transform & do it's own aabb.Contains() then this would just be inefficient if it weren't updated. But yeah for now it could just do that, and I've switched it back to using `GetEntitiesIntersecting`.

metalgearsloth: I haven't foreseen a need for it for anything else but if it ever happens we can just update.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.cs
```diff
@@ -58,15 +58,22 @@ private void MapManagerOnTileChanged(object? sender, TileChangedEventArgs e)
...
         private void DeparentAllEntsOnTile(GridId gridId, Vector2i tileIndices)
         {
-            var mapId = _mapManager.GetGrid(gridId).ParentMapId;
-            var mapTransform = Transform(_mapManager.GetMapEntityId(mapId));
-         
-            foreach (var uid in _entityLookup.GetEntitiesIntersecting(gridId, tileIndices).ToList())
+            var grid = _mapManager.GetGrid(gridId);
+            var gridUid = grid.GridEntityId;
+            var mapTransform = Transform(_mapManager.GetMapEntityId(grid.ParentMapId));
+            var aabb = new Box2(tileIndices * grid.TileSize, (tileIndices + 1) * grid.TileSize);
```
metalgearsloth: The EntityLookup method should be public so you can just use that, otherwise LGTM.

metalgearsloth (approved):

Robust.Shared/GameObjects/Systems/SharedTransformSystem.cs
ElectroJr: If you mean using `GetLocalBounds` to get the aabb, they were private & not part of the `IEntityLookup` interface. I've changed that now.

Also slightly tweaked `GetWorldBounds` so that it supports non standard grid tile sizes

PR #2420, opened 2022-01-13
Give container attempt events an optional reason output.

ElectroJr:
Useful for giving feedback to players when they are attempting to insert an entity into some container.

### DIFF
--- a/Robust.Shared/Containers/BaseContainer.cs
+++ b/Robust.Shared/Containers/BaseContainer.cs
@@ -86,9 +86,10 @@ namespace Robust.Shared.Containers
         }
 
         /// <inheritdoc />
-        public virtual bool CanInsert(EntityUid toinsert, IEntityManager? entMan = null)
+        public virtual bool CanInsert(EntityUid toinsert, out string? reason, IEntityManager? entMan = null)
         {
             DebugTools.Assert(!Deleted);
+            reason = null;
 
             // cannot insert into itself.
             if (Owner == toinsert)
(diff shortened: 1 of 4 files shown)


PaulRitter (changes requested):
could do with a [notnullwhen(false)], otherwise lgtm.

ElectroJr:
> could do with a [notnullwhen(false)] 

I don't think you should be required to provide a reason string every time you want to cancel an insertion (or removal). While its useful as feedback for directly user initiated interactions, for things like lockers closing and failing to grab/insert other nearby lockers, I don't think you should have to provide localized reason strings for that. Similarly you'd have to add localised reason strings for failing to insert a map or grid entity into a container, which ideally a user should never even see.

And AFAIK there is no `[AlwaysNullWhen(true)]` or `[MaybeNotNullWhen(false)]` attribute.

metalgearsloth:
I'm a bit suss having this in engine tbh

ElectroJr:
> I'm a bit suss having this in engine tbh

It does feel a bit weird having a reason field (which transparently exists for the content's `PopupSystem`) to exist in engine. But the alternative is for the storage, item-slots, and maybe eventually other systems to each create duplicates of these 4 events, which also feels icky to me. Especially as the more general engine event would then probably just go unused. But if that is preferred I'm ok with just doing that instead.

metalgearsloth:
@PJB3005 any thoughts?

PaulRitter:
>I don't think you should be required to provide a reason string every time you want to cancel an insertion (or removal). 

ur right

>I'm a bit suss having this in engine tbh

i think its reasonable supporting having a reason for containerinsert/remove failing. for one, i dont see a better way than this to implement it, secondly this would contribute to making inventorysystem even less of a wrapper of containersystem, moving the can(un)equip fully to containerevents

PJB3005:
Seconding that I don't like this being in-engine. Honestly I feel like the actual "Can insert" code here should be a content-side thing instead since it's so game specific.

If you absolutely have to, don't make the reason a string. Make it an `object`.

ElectroJr:
Ok, I'll just go with content-side events specific to a given component/interaction. Currently very few things use the insert-attempt events anyways.

PR #2421, opened 2022-01-14
Add IsQueuedForDeletion

ElectroJr:
Allows systems to check if an entity is about to be deleted before doing potentially unnecessary & expensive computations (e.g., ExtensionCableSystem)


Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -34,8 +34,7 @@ public partial class EntityManager : IEntityManager
         /// <inheritdoc />
         public virtual IEntityNetworkManager? EntityNetManager => null;
 
-        protected readonly Queue<EntityUid> QueuedDeletions = new();
```
gradientvera: The reason we kept a queue as well was to maintain deletion order...

gradientvera (changes requested):

Robust.Shared/GameObjects/EntityManager.cs
ElectroJr: ok, reverted that change

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -297,10 +297,12 @@ private void RecursiveDeleteEntity(EntityUid uid)
 
         public void QueueDeleteEntity(EntityUid uid)
         {
-            if(QueuedDeletionsSet.Add(uid))
+            if (QueuedDeletionsSet.Add(uid))
                 QueuedDeletions.Enqueue(uid);
         }
 
+        public bool IsQueuedForDeletion(EntityUid uid) => QueuedDeletions.Contains(uid);
```
ShadowCommander: Is it faster to check the `QueuedDeletionsSet`?

ElectroJr: AFAIK checking the hashset is faster, this was an accidental leftover from before where I had removed the queue and just had the hashset. I'll update this now.

ShadowCommander (approved):

metalgearsloth (approved):
conflicts

PR #2422, opened 2022-01-14
Stop updating entity tree on pre-delete un-anchor event

ElectroJr:
Currently when an anchored entity is deleted, it will first un-anchor itself and be added to the entity lookup tree. then, when `EntityManager.EntityDeleted` gets invoked, it gets removed from that tree. This just stops the entity from being added to the tree in the first place.

Though on a more general note: how vital is the unanchoring / detach-to-null before an entity is deleted?
Time is still wasted on other components, for example the `CollideOnAnchorComponent` needlessly updates the physics collision. AFAIK any entity that would care about un-anchor events should also have the required logic present in their shutdown-logic. Should the unanchor event even be raised on shutdown?


metalgearsloth:
The issue with DetachParentToNull (aside from the fact it's expensive) is a bunch of stuff listens to parent changes that needs to be altered (e.g. dynamictrees, anchored ents).

PR #2429, opened 2022-01-17
Store grids in a DynamicTree

metalgearsloth:
Removes the nasty O(n) check from FindGridsIntersecting which is much better for large grid counts.

I wasted a lot of time trying to debug issues until I realised ClientMapManager never applied the grid's rotation.

TODO:

- [x] Anchoring test for rotated grid (I think this already gets caughhhttt?)
- [x] Move grids off of the map's broadphase

### DIFF
--- a/Robust.Client/GameController/GameController.cs
+++ b/Robust.Client/GameController/GameController.cs
@@ -134,8 +134,8 @@ namespace Robust.Client
             _prototypeManager.Initialize();
             _prototypeManager.LoadDirectory(Options.PrototypeDirectory);
             _prototypeManager.Resync();
-            _mapManager.Initialize();
             _entityManager.Initialize();
+            _mapManager.Initialize();
             _gameStateManager.Initialize();
             _placementManager.Initialize();
             _viewVariablesManager.Initialize();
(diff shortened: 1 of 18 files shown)


moonheart08:
merg when

Robust.Client/GameController/GameController.cs
```diff
@@ -134,8 +134,8 @@ internal bool StartupContinue(DisplayMode displayMode)
             _prototypeManager.Initialize();
             _prototypeManager.LoadDirectory(Options.PrototypeDirectory);
             _prototypeManager.Resync();
-            _mapManager.Initialize();
             _entityManager.Initialize();
+            _mapManager.Initialize();
```
mirrorcult, line 138: this seems sus, whats the rationale

mirrorcult (approved):

Robust.Client/GameController/GameController.cs
metalgearsloth: Eventbus isn't instantiated on entitymanager until then so mapmanager can't subscribe to events (this is also the order the server initialises them in).

mirrorcult: ok sgtm

PR #2430, opened 2022-01-17
Move MapGrid data from MapManager to MapGridComponent

Acruid:
What if, like, grids were components man?

IMapGrids were moved from the MapManager to MapGridComponent. Now instead of the components being synchronized by the map manager, the existence of a grid is determined by the existence of the component.

MapManager still stores the GridId -> EntityUid mapping, but now GridIds are optional to use, and all grids can be properly referenced by EntityUid.


Robust.Client/GameObjects/EntitySystems/ClientOccluderSystem.cs
```diff
@@ -85,7 +85,7 @@ private void HandleDirtyEvent(OccluderDirtyEvent ev)
             }
```
metalgearsloth, line 86:
```suggestion
            }
```

Robust.Shared/Map/MapManager.MapCollection.cs
```diff
@@ -0,0 +1,279 @@
...
+
+    public IEnumerable<MapId> GetAllMapIds()
+    {
+        return _maps;
+    }
+
+    public bool IsMap(EntityUid uid)
+    {
+        return _mapEntities.Any(x => x.Value == uid);
+    }
```
metalgearsloth: Should this just check for MapComponent? See https://github.com/space-wizards/RobustToolbox/pull/2486 that this is pretty high up.

Robust.Shared/Map/MapManager.MapCollection.cs
```diff
@@ -0,0 +1,279 @@
...
+            CreateMap(MapId.Nullspace);
+        else
+        {
+            if (!_mapEntities.TryGetValue(MapId.Nullspace, out var mapEntId))
+                return;
+
+            foreach (var childTransform in EntityManager.GetComponent<TransformComponent>(mapEntId).Children.ToArray())
+            {
+                EntityManager.DeleteEntity(childTransform.Owner);
+            }
```
metalgearsloth: Use ChildEntities as it avoids the additional GetComponent<TransformComponent> when only the entityuid is needed, though I think I updated this in one of my recent PRs...

metalgearsloth (approved):

metalgearsloth:
merge it merge it merge it

Issue #2431, opened 2022-01-17
Sprite rendering rotation/offset issues


ElectroJr:
There are currently two bugs related two rendering rotated sprites:

Firstly, shaders cut off part of the sprite when the sprite is rotated. This is most easily visible when the sprite has some offset, but is also noticeable for any square-shaped sprite with an outline, where the corners get cut off at ~45 degrees. In the video below this is just shown using the interaction outline shader. The issue is probably somewhere in `DrawEntities()` within `Clyde.HLR.cs` and has to do with the `screenSpriteSize` calculation. Brute forcing a solution by increasing `postShadeScale` works but the calculation should probably just be fixed. I'm happy to poke around but don't necessarily know enough about clyde/render- targets/shaders to fix this

Secondly, directional-sprites with offsets apply their offsets incorrectly, though this is only noticeable when the sprite is rotated relative to the observer. See the air alarm in the video. #2432 should fix this.

https://user-images.githubusercontent.com/60421075/149833474-9222314c-5dc1-4e0c-8537-eff5ac7c7bc5.mp4

edit:
Apparently already known (space-wizards/space-station-14/issues/5455), but as that issue is in content and as this is an engine problem I'll leave this up.

PR #2432, opened 2022-01-17
Sprite offset rendering fix

ElectroJr:
Fixes the second issue mentioned in #2431

This does mess up content's air alarms, because these were using bad offsets to render in the correct locations.
So space-wizards/space-station-14/pull/6216 should be merged around the same time.


metalgearsloth (approved):
Assuming it looks okay in-game

PR #2433, opened 2022-01-17
Fixes warnings

wrexbe:
I fixed the warnings I could see

### DIFF
--- a/Robust.Client/Debugging/DebugPhysicsSystem.cs
+++ b/Robust.Client/Debugging/DebugPhysicsSystem.cs
@@ -207,7 +207,7 @@ namespace Robust.Client.Debugging
 
                     const float AlphaModifier = 0.2f;
 
-                    foreach (var fixture in physBody.Fixtures)
+                    foreach (var fixture in _entityManager.GetComponent<FixturesComponent>(physBody.Owner).Fixtures.Values)
                     {
                         // Invalid shape - Box2D doesn't check for IsSensor
                         if (physBody.BodyType == BodyType.Dynamic && fixture.Mass == 0f)
@@ -274,7 +274,7 @@ namespace Robust.Client.Debugging
                     const float AlphaModifier = 0.2f;
                     Box2? aabb = null;
 
-                    foreach (var fixture in physBody.Fixtures)
+                    foreach (var fixture in _entityManager.GetComponent<FixturesComponent>(physBody.Owner).Fixtures.Values)
                     {
                         for (var i = 0; i < fixture.Shape.ChildCount; i++)
                         {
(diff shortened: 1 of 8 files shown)


Robust.Client/Debugging/DebugPhysicsSystem.cs
```diff
@@ -207,7 +207,7 @@ private void DrawWorld(DrawingHandleWorld worldHandle, OverlayDrawArgs args)
 
                     const float AlphaModifier = 0.2f;
 
-                    foreach (var fixture in physBody.Fixtures)
+                    foreach (var fixture in _entityManager.GetComponent<FixturesComponent>(physBody.Owner).Fixtures.Values)
```
ShadowCommander, line 210: Is there a way to make `_physicsSystem.GetCollidingEntities()` also return the FixturesComponent? If not then this should be fine since its' a debug overlay.

Robust.Server/ServerStatus/StatusHost.cs
```diff
@@ -79,10 +79,12 @@ public async Task ProcessRequestAsync(HttpListenerContext context)
 
         public event Action<JsonNode>? OnInfoRequest;
 
+#pragma warning disable CS0618 // For backwards compatibility
```
ShadowCommander: Maybe add a TODO to remove once MoMMI is refactored? If MoMMI can be refactored to async.

Robust.UnitTesting/TestingModLoader.cs
```diff
@@ -51,7 +51,8 @@ public void AddEngineModuleDirectory(string dir)
         {
             // Only used for ILVerify, not necessary.
         }
-
+#pragma warning disable CS0067 // Needed by interface
```
ShadowCommander, line 54: Another TODO maybe?

ShadowCommander (changes requested):

Robust.Client/Debugging/DebugPhysicsSystem.cs
wrexbe: I don't see a simple way to do it, would push this PR too far out of scope.

Robust.UnitTesting/TestingModLoader.cs
wrexbe: It's kind of a false positive, this is for testing. The real class uses it, so I don't think removing this is a goal.

Robust.Server/ServerStatus/StatusHost.cs
wrexbe: Added a todo

ShadowCommander (approved):

PR #2434, opened 2022-01-17
Make OffsetRsiDir() and GetDir() public

ElectroJr:
Makes `SpriteComponent`'s `OffsetRsiDir()` public and moves `GetDir()` function into direction extensions.

Nice to have for space-wizards/space-station-14#6221

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -1547,7 +1547,7 @@ namespace Robust.Client.GameObjects
             return rsi["error"];
         }
 
-        private static RSI.State.Direction OffsetRsiDir(RSI.State.Direction dir, DirectionOffset offset)
+        public static RSI.State.Direction OffsetRsiDir(RSI.State.Direction dir, DirectionOffset offset)
         {
             // There is probably a better way to do this.
             // Eh.
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -1526,51 +1526,6 @@ namespace Robust.Client.GameObjects
             return rsi["error"];
         }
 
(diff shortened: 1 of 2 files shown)


Robust.Client/Utility/DirExt.cs
```diff
@@ -109,5 +109,16 @@ public static Direction TurnCcw(this Direction dir)
...
             return (Direction)(((int)dir + 1) % 8);
         }
+
+        public static RSIDirection ToRsiDirection(this Angle angle, RSI.State.DirectionType type)
+        {
+            return type switch
+            {
+                RSI.State.DirectionType.Dir1 => RSIDirection.South,
+                RSI.State.DirectionType.Dir4 => angle.GetCardinalDir().Convert(type),
+                RSI.State.DirectionType.Dir8 => angle.GetDir().Convert(type),
```
PaulRitter: just do angle.getdir.convert for both dir4 and dir8 since it accounts for that internally already

PaulRitter (changes requested):

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1550,7 +1526,7 @@ internal static RSI.State GetFallbackState(IResourceCache cache)
             return rsi["error"];
         }
 
-        private static RSI.State.Direction OffsetRsiDir(RSI.State.Direction dir, DirectionOffset offset)
+        public static RSI.State.Direction OffsetRsiDir(RSI.State.Direction dir, DirectionOffset offset)
```
PJB3005: This shouldn't be in `SpriteComponent` if it's made public.

PJB3005 (changes requested):

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
ElectroJr: Fair point. I've made it an extension method and moved it to `DirExt.cs`

PJB3005 (approved):

PR #2438, opened 2022-01-18
Maybe fix effect spam

metalgearsloth:
https://discord.com/channels/310555209753690112/770682801607278632/932843798952837121

IsValid check on the coordinates.


metalgearsloth:
Now that I think about it entitycoordinates exist so why the hell do we even care about attachedentityuid anymore

PR #2440, opened 2022-01-20
Add acruid's funny entity query

### LINKED ISSUE #2227: Suss out exposing _entTraitDict for transforms
Take WorldRotation for example. To calculate it we continuously need to go through our parent uid + get the transformcomponent for the parent recursively. Every time we get transformcomponent we need to do _entTraitDict[typeof(T)] when we already know we just need the transformcomponents.

Granted exposing the internal workings of this is somewhat dirty it's still worthwhile investigating the benchmarks on using this just for transform stuff.

@PJB3005 any thoughts?

metalgearsloth:
tl;dr You don't have to typecast back into the dictionary every time.

Resolves https://github.com/space-wizards/RobustToolbox/issues/2227

In my shitty PVS profiling with 20 clients it's roughly a third faster (TryGetValue is still like 40% of the time though) so it's not gonna let us hit 300 clients but it'll be a nice saving in multiple places in the engine once more stuff is converted. Should also help physics a little bit as I converted the helper methods to use it.

(On release, left is pr im just a dumby)
image

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -278,12 +278,26 @@ internal partial class PVSSystem : EntitySystem
 
         visibleEnts.Clear();
 
+        var eyeQuery = EntityManager.GetEntityQuery<EyeComponent>();
+        var transformQuery = EntityManager.GetEntityQuery<TransformComponent>();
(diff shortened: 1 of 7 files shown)


metalgearsloth:
@PaulRitter relevant4ubb

PaulRitter (approved):
ballin' pr bb

PR #2442, opened 2022-01-20
Allow sprite layer data to be set using `PrototypeLayerData`.

ElectroJr:
This PR does two main things. Firstly, it allow layer information to be set using `PrototypeLayerData`. This is useful for systems/components that need to modify layers using data defined in yaml somewhere, without having to individually call all of `LayerSetColor`, `LayerSetRsi`, `LayerSetState`, etc. This was done by just moving most of the logic in the `LayerDatums` setter into a separate function called `LayerSetData()`. It also just makes `HandleComponentState` use this same function.

Secondly, it makes it so that layers now store information about the shader prototype. There was a `todo` about doing this in `ToPrototypeData()`, so I guess this was desirable? The real reason I want this is to be able to generally identify if a layer is `unshaded`, though maybe I should use another means of marking/selecting these layers.

Another thing I'm somewhat unsure about is the significance of `_layerMapShared` and `_layerMapEnsurePrivate()`, and whether `LayerSetData()` should set/call these. My guess is that these exists in case two sprites share the same layer map, but I don't see how that could happen, seeing as the `CopyFrom()` function creates a seperate dictionary anyways?


metalgearsloth (approved):
@PJB3005

PR #2445, opened 2022-01-22
Fix grid deletion

metalgearsloth:
Edit: Alright test added and checked it fails on master so good 4 merg


metalgearsloth:
you seem like a trustworthy contributor free merge

PR #2446, opened 2022-01-22
Only include shadow casting lights in the "max lights per scene" count

weaversam8:
_(Please correct me if I've misunderstood anything in Clyde, this is my best attempt at parsing this code.)_

When rendering lights, we currently enforce two lighting limits. One is a **"hard" cap** for the number of lights (`LightsToRenderListSize = 2048`) based on memory usage (since lights are stored in an array that we keep in memory and reuse), while the other is a **"soft" cap** (`_maxLightsPerScene = 128`) based on performance.

The soft cap is used to determine which lights to render on the client. When the soft cap is exceeded, the client can continue rendering frames and chooses to sacrifice lights based on the distance from the player (rendering is skipped for lights from furthest to closest.)

As of right now, the soft cap applies for all lights in the scene, but realistically, the most performance intensive lights are those which cast shadows, which are far more CPU intensive to render. Lights that don't cast shadows are much cheaper.

This PR updates the light rendering code such that the soft cap _only applies_ to _shadow casting_ lights. Non-shadow casting lights no longer count against the soft cap (but continue to count against the hard cap.)

### Screenshots (before)

Limit on non-shadow casting lights:
image

### Screenshots (after)

No limit on non-shadow casting lights:
image

Limit on shadow-casting lights (notice non-shadow casting lights in the distance (bottom right) still rendering):
image


mirrorcult:
@PJB3005

PJB3005 (approved):

PR #2453, opened 2022-01-24
Add dirty method to entmanager for components

metalgearsloth:
Only existed for entities previously. Something something something ECS.


Robust.Shared/GameObjects/Component.cs
```diff
@@ -165,7 +165,7 @@ internal void LifeRemoveFromEntity()
 
         /// <inheritdoc />
         [ViewVariables]
-        public GameTick LastModifiedTick { get; private set; }
+        public GameTick LastModifiedTick { get; internal set; }
```
metalgearsloth, line 168: Ideally this would only be settable via EntityManager but I couldn't figure out a good way to do it.

PR #2456, opened 2022-01-25
Fix spinbot CBT

metalgearsloth:
- Cursor no longer resets to position 0 on input
- Scrolling will update cursor position as required

### DIFF
--- a/Robust.Client/UserInterface/Controls/LineEdit.cs
+++ b/Robust.Client/UserInterface/Controls/LineEdit.cs
@@ -215,7 +215,7 @@ namespace Robust.Client.UserInterface.Controls
         /// <remarks>
         /// Does not fix cursor positions, those will have to be adjusted manually.
         /// </remarks>>
-        protected bool SetText(string newText)
+        public bool SetText(string newText)
         {
             if (IsValid != null && !IsValid(newText))
             {
--- a/Robust.Client/UserInterface/Controls/SpinBox.cs
+++ b/Robust.Client/UserInterface/Controls/SpinBox.cs
@@ -33,7 +33,18 @@ namespace Robust.Client.UserInterface.Controls
                     return;
                 }
                 _value = value;
-                _lineEdit.Text = value.ToString();
+
+                var edgeCursor = _lineEdit.CursorPosition == _lineEdit.Text.Length;
+
+                _lineEdit.SetText(value.ToString());
+
+                var textLength = _lineEdit.Text.Length;
+
+                if (edgeCursor && textLength != _lineEdit.CursorPosition)
+                    _lineEdit.CursorPosition++;
+                else if (_lineEdit.CursorPosition > textLength)
+                    _lineEdit.CursorPosition = textLength;
+
                 ValueChanged?.Invoke(this, new ValueChangedEventArgs(value));
             }
         }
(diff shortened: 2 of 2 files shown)


Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -215,7 +215,7 @@ public void InsertAtCursor(string text)
         /// <remarks>
         /// Does not fix cursor positions, those will have to be adjusted manually.
         /// </remarks>>
-        protected bool SetText(string newText)
+        public bool SetText(string newText)
```
metalgearsloth, line 218: The Text property defaults the cursor position to 0 which I guess is okay for copy-paste situations?

mirrorcult:
spinbot

Robust.Client/UserInterface/Controls/SpinBox.cs
```diff
@@ -33,7 +33,22 @@ public int Value
...
+                var edgeCursor = _lineEdit.CursorPosition == _lineEdit.Text.Length;
+
+                _lineEdit.SetText(value.ToString());
+
+                var textLength = _lineEdit.Text.Length;
+
+                if (edgeCursor && textLength != _lineEdit.CursorPosition)
+                    _lineEdit.CursorPosition++;
+                else if (_lineEdit.CursorPosition > textLength)
+                    _lineEdit.CursorPosition = textLength;
```
ShadowCommander, line 50: Could you rename `LineEdit.SetText` to `LineEdit.InternalSetText` and move this code to a new `LineEdit.SetText`? Since this code looks like it can be used in other places that replace text.

ShadowCommander (changes requested):

Robust.Client/UserInterface/Controls/SpinBox.cs
```diff
@@ -33,7 +33,22 @@ public int Value
...
+                // to the end again
+                // If we scroll down and have fewer characters then it will shift it left to the end.
+                var edgeCursor = _lineEdit.CursorPosition == _lineEdit.Text.Length;
+
+                _lineEdit.SetText(value.ToString());
+
+                var textLength = _lineEdit.Text.Length;
+
+                if (edgeCursor && textLength != _lineEdit.CursorPosition)
+                    _lineEdit.CursorPosition++;
```
ShadowCommander, line 48: Does this work if the `textLength` jumps more than one character? (e.g. 1 to 3, or 3 to 12)

ShadowCommander (changes requested):

Robust.Client/UserInterface/Controls/SpinBox.cs
```diff
@@ -72,7 +89,7 @@ public SpinBox()
             _lineEdit.OnTextChanged += (args) =>
             {
                 if (int.TryParse(args.Text, out int i))
-                    Value = i;
+                    OverrideValue(i);
```
ShadowCommander, line 92: This needs to call the `ValueChanged` event, so either use Value again or call the event here.

ShadowCommander (changes requested):

Robust.Client/UserInterface/Controls/LineEdit.cs
PJB3005: The dangerous thing about this function is that if cursor position isn't adjusted manually, you can get out of bounds errors I think.

Robust.Client/UserInterface/Controls/SpinBox.cs
PJB3005: Yeah that should be a while loop instead.

metalgearsloth:
spinbox will remain cbt

PR #2460, opened 2022-01-26
Hopefully fix appearance component mispredict reconciliation

### LINKED ISSUE #2127: AppearanceComponent doesn't reconcile properly after mispredicts.
Right now, `Appearance` data isn't reconciled at all when the client mispredicts. This can cause *hilarious* effects in SS14 such as mobs staying horizontal forever when slipping mispredicts, etc.

ElectroJr:
fixes #2127.

The appearance component state handling code sets the component's appearance data dictionary equal to the state's dictionary. As a result, if the client then updates and dirties the component during prediction, it actually also modifies the most recently received server state. So when a client mispredicts and does not receive a new server state, it will incorrectly reset itself to the dirty value.

A partial fix would be to just make the state-handling take a shallow clone of the appearance dictionary, but that could in principle still cause difficult to diagnose reconciliation issues. So this PR just restricts the appearance data values to be value-type or `ICloneable` and makes the state handling code duplicate the appearance dictionary

I'm really not sure if this is the correct way to go about this. It would be nice to just automatically support things like `List<ICloneable>` & similarly for dictionaries. But I don't know how to fully generalize it to allow going from something like `List<string>` -> `object` -> `List<Cloneable>`, and then easily reversing that in visualizers without iterating & casting. So just requiring that complex appearance data must be a by-value struct or cloneable class seems fine?

~~This also requires a content PR to avoid breaking some visualizers (space-wizards/space-station-14/pull/6319).~~
Has been merged.


Robust.Shared/Utility/CollectionExtensions.cs
```diff
@@ -20,7 +20,7 @@ public static class Extensions
         /// <param name="self">The list to shallow clone.</param>
         /// <typeparam name="T">The type of the list's elements.</typeparam>
         /// <returns>A new list with the same elements as <paramref name="list" />.</returns>
-        public static List<T> ShallowClone<T>(this List<T> self)
+        public static List<T> ShallowClone<T>(this IReadOnlyList<T> self)
```
PaulRitter: whats this about

PaulRitter (changes requested):

Robust.Shared/Utility/CollectionExtensions.cs
ElectroJr: IIRC I wanted this for space-wizards/space-station-14/pull/6319/files, for the `IReadOnlyList` in `SharedMapLayerData.cs`. But I changed that around and its not needed anymore, so I will get rid of this change.

metalgearsloth (approved):
@PJB3005 speak now or forever hold your peace son

PJB3005 (approved):

PaulRitter (approved):

PR #2461, opened 2022-01-26
Fix entity map change crash

metalgearsloth:
All of the entities on the map have DetachParentToNull run which is !FUN!

Ideally they wouldn't (as this will also help make grid deletion MUCH faster) but for now we're worried about fixing the crash.

Edit: Should also note only way I could replicate is by closing the client window.


ElectroJr (approved):
Haven't tested or tried to reproduce the crash, but changes seem fine.

PR #2465, opened 2022-01-27
Automatically pause server if no players connected.

### LINKED ISSUE #2383: CVar to automatically pause game loop if no players are connected
Would be nice to reduce idle CPU usage of the server.

PJB3005:
Configurable via CVar. Intent here is that leaving a server running in the background of your PC shouldn't draw that much CPU.

Fixes #2383

### DIFF
--- a/Robust.Server/BaseServer.cs
+++ b/Robust.Server/BaseServer.cs
@@ -1,6 +1,6 @@
 using System;
 using System.IO;
-using System.Runtime;
+using System.Linq;
 using System.Threading;
 using Prometheus;
 using Robust.Server.Console;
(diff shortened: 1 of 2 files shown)


Robust.Server/BaseServer.cs
```diff
@@ -543,6 +546,49 @@ private void LoadSettings()
             Logger.InfoS("srv", $"Name: {ServerName}");
             Logger.InfoS("srv", $"TickRate: {_time.TickRate}({_time.TickPeriod.TotalMilliseconds:0.00}ms)");
             Logger.InfoS("srv", $"Max players: {MaxPlayers}");
+
+            cfgMgr.OnValueChanged(CVars.GameAutoPauseEmpty, UpdateAutoPause, true);
```
gradientvera, line 550: Unsub from this on cleanup?

PJB3005: The other value up there wasn't getting unsubbed either and it's not like any of this cleanup logic works so...

gradientvera: Fair, doesn't matter much for now then

PR #2473, opened 2022-01-30
Don't use worldbounds for PVS

metalgearsloth:
@PaulRitter 

Don't look until exams done or I cut you.

Main goal is if a grid is spinning then no new chunks should be going in range as a result.


Robust.Shared.Maths/Vector2.cs
```diff
@@ -130,6 +130,12 @@ public Vector2(float x, float y)
             return new((int) MathF.Floor(X), (int) MathF.Floor(Y));
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public readonly Vector2i Ceiling()
```
PaulRitter:
```suggestion
        public readonly Vector2i Ceiled()
```

Robust.Server/GameStates/ChunkIndicesEnumerator.cs
```diff
@@ -5,30 +5,31 @@ namespace Robust.Server.GameStates;
...
-        _topLeft = (viewBox.TopLeft / chunkSize).Floored();
-        _bottomRight = (viewBox.BottomRight / chunkSize).Floored();
+        _bottomLeft = ((viewPos - range) / chunkSize).Floored();
+        // Also floor this as we get the whole chunk anyway.
+        _topRight = ((viewPos + range) / chunkSize).Floored();
 
-        _x = _topLeft.X;
-        _y = _bottomRight.Y;
+        _x = _bottomLeft.X;
+        _y = _topRight.Y;
```
PaulRitter: danger danger. this one should be `bottomLeft`

PaulRitter (changes requested):
get owned nerd

Robust.Server/GameStates/ChunkIndicesEnumerator.cs
metalgearsloth: A

Robust.Shared.Maths/Vector2.cs
metalgearsloth: Ceiled's not even a word m8

PaulRitter (approved):

metalgearsloth:
This will probably be superseded by whatever gets used for https://github.com/space-wizards/RobustToolbox/issues/2527 but okay for now.

PR #2476, opened 2022-01-30
Fix InteractionOutline for rotated eyes

metalgearsloth:
Bounds transform moment.

On master the offset is just applied to worldpos so falls over when the grid rotates, so we just do a transform on it using the entity's world matrix and get the correct rotated offset.

master:
image

fixed:
image

Fixes the other half of https://github.com/space-wizards/RobustToolbox/issues/2431 also https://github.com/space-wizards/space-station-14/issues/5455


metalgearsloth:
Deprecated in favor of https://github.com/space-wizards/RobustToolbox/pull/2511

PR #2477, opened 2022-01-30
Don't serialize grid fixtures

metalgearsloth:
Thought about it a bit. Ideally long-term custom ones would be serialized but I'd need to make a bunch of changes to gridfixturesystem to support determining whether they are custom.

For now:

Pros:
- Faster yml reading for map loads
- Faster map writing
- 5% reduced map line count (for saltern)

Cons:
- Any custom data around restitution etc set via vv isn't getting serialized.
- Serialization may or may not be faster overall? (Resolve IEntityManager for every body vs having to serialize >100 fixtures).

### DIFF
--- a/Robust.Shared/Physics/FixturesComponent.cs
+++ b/Robust.Shared/Physics/FixturesComponent.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using Robust.Shared.Analyzers;
 using Robust.Shared.GameObjects;
 using Robust.Shared.GameStates;
+using Robust.Shared.IoC;
 using Robust.Shared.Physics.Dynamics;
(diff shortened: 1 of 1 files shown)


Acruid:
Why would the engine need physics engine fixtures be serialized for girds? We already have the actual "model" of the grid that the fixtures can be generated for at any time, and the client already *somehow* gets any restitution values from the server, which means that value is coming from the same place the fixtures are getting it from (PhysicsComponent?). Basically you should not need the fixtures in order to generate the fixtures right?

Needing an IoCManager resolve on a component is pretty cringe, which says a lot about the whole `BeforeSerialization` function on components.

Acruid (approved):

PR #2480, opened 2022-01-31
Cleanup contact code a lot

metalgearsloth:
- A lot of this stuff was farseer porting box2d verbatim.
- Most of the cleanup is just doing it using existing dotnet stuff (i.e. LinkedList<T>) and moving more stuff onto ContactManager.
- I also moved contact storage directly onto fixtures which box2d has an open issue for. This removes the particularly slow O(n) check for AddPair but that really only matters with lots of contacts on a body. Previously this would iterate every single contact on a body checking if it was for the relevant fixture, then would check if the other fixture was the other one we're trying to add a pair for. Now this is a dictionary lookup so even 100 contacts is still going to be constant.
- References count is probably similar.
- I also found a multi-threading bug with contact code doing the IoC resolve so I moved that out of the multithreaded part and just used an array to check what needs doing.

Requires https://github.com/space-wizards/RobustToolbox/pull/2482


Robust.Shared/Robust.Shared.csproj
```diff
@@ -9,6 +9,7 @@
   <Import Project="..\MSBuild\Robust.DefineConstants.targets" />
   <ItemGroup>
     <PackageReference Include="JetBrains.Annotations" Version="2020.3.0" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="5.0.4" />
```
metalgearsloth, line 12: This is already used in Robust.Server for PVS. I haven't actually profiled contact pooling vs no-pooling because box2d had its own implementation but contacts are destroyed and created extremely frequently.

PR #2482, opened 2022-01-31
Fix edge collisions

metalgearsloth:
Sloth is very angery.

I just updated edge code to latest on box2d and stacking is stable again. This is also getting a test to assert the boxes eventually sleep this time.


metalgearsloth:
Something weird is going on where the edge normal might be inverted to box2d now? But stacking is stable so not sure how much I care right now after spending hours on this?

metalgearsloth:
Okay I'm an idiot but it's fixed in the followup PR anyway. I'm invoking cronyism and merging as it has a test again.

PR #2490, opened 2022-02-01
Reduce gamestate allocs

metalgearsloth:
Parallel.ForEach has no right allocating that much with no partitioner specified

Before:
image

After:
image


metalgearsloth:
Ran with PVS profile and it didn't seem to explode; can just revert if something goes fucky wucky.

mirrorcult (approved):

PR #2491, opened 2022-02-01
remove GetBus()

ElectroJr:
Currently any component add/init/startup/shutdown/remove calls will result in one or more `IoCManager.Resolve<IEntityManager>()` calls, mostly due to `GetBus()`. The end result is that when performing Queued entity deletions after nuking a station, 5.5% is spent on IoCManager.Resolve() calls. Please, no more.

### DIFF
--- a/Robust.Shared/GameObjects/Component.cs
+++ b/Robust.Shared/GameObjects/Component.cs
@@ -37,11 +37,13 @@ namespace Robust.Shared.GameObjects
         /// Increases the life stage from <see cref="ComponentLifeStage.PreAdd" /> to <see cref="ComponentLifeStage.Added" />,
         /// calling <see cref="OnAdd" />.
         /// </summary>
-        internal void LifeAddToEntity()
+        internal void LifeAddToEntity(IEntityManager entManager)
         {
             DebugTools.Assert(LifeStage == ComponentLifeStage.PreAdd);
 
             LifeStage = ComponentLifeStage.Adding;
+            CreationTick = entManager.CurrentTick;
+            entManager.EventBus.RaiseComponentEvent(this, CompAddInstance);
             OnAdd();
 
 #if DEBUG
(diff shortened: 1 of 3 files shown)


Robust.Shared/GameObjects/Component.cs
```diff
@@ -159,23 +166,11 @@ internal void LifeRemoveFromEntity()
         private static readonly ComponentShutdown CompShutdownInstance = new();
         private static readonly ComponentRemove CompRemoveInstance = new();
 
-        private IEventBus GetBus()
-        {
-            // Apparently components are being created outside of the EntityManager,
-            // and the Owner is not being set correctly.
-            // ReSharper disable once RedundantAssertionStatement
-            DebugTools.AssertNotNull(Owner);
```
ElectroJr, line 167: I have no idea what this is talking about.
When looking at all calls to add/init/startup/shutdown/remove, they all originate from entity manger. Is this just an old comment?

metalgearsloth (approved):

Robust.Shared/GameObjects/Component.cs
metalgearsloth: Yes. I believe everything that used to do that was culled a couple months ago (e.g. dummy sprites).

gradientvera: indeed, this is fine now

gradientvera (approved):
Eventually, these methods could be moved to `IEntityManager` itself. For now though, it's fine.

Acruid (approved):
Be aware that this changes the ordering of the calls, now the Lifestage event is raised before the function on the component is called. I think the only way this could cause a problem is if a component both overrode the legacy component functions and subscribed to the event, and that code implicitly depended on the ordering.

If this PR causes a problem like this, the solution is to move everything to an ECS system anyways, so it seems like a non-issue.

PR #2487, opened 2022-02-01
Funny struct enumerator for xform kids

metalgearsloth:
Before
image

After
image

Would need to look at SortedSet to remove the top 2 allocs.

PR #2492, opened 2022-02-01
Funny struct enumerator for xform kids Part 2

metalgearsloth:
Also includes https://github.com/space-wizards/RobustToolbox/pull/2487 so merge that first.

This is on client this time but it essentially removes the Enumerator allocs on client again.
Before:
image

After:
image


Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
```diff
@@ -169,36 +169,46 @@ public void RemoveExpectedEntity(EntityUid uid)
...
-        private void UpdateEntityRecursively(EntityUid entity)
+        private void UpdateEntityRecursively(
+            EntityUid entity,
+            EntityQuery<TransformComponent> xformQuery,
+            EntityQuery<PointLightComponent> pointQuery,
+            EntityQuery<SpriteComponent> spriteQuery)
         {
             // TODO: Since we are recursing down,
             // we could cache ShowContents data here to speed it up for children.
             // Am lazy though.
```
ElectroJr: I guess if this is being optimized with the component queries, this might as well be done now?
Should help with nested entities like stuff in backpacks equipped by players and whatnot.

I.e., instead of recursing down through children and having each one individually recurse upwards looking for occluding containers, just recurse downwards while looking for any occluding containers. Once one is found at some depth, keep recursing down and but stop doing container checks, & just set the sprite & light data.

Also that double recursion, once for sprites and once for lights is yucky

Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -170,15 +170,17 @@ private void OnEntityDeleted(object? sender, EntityUid e)
...
-        UpdateEntityRecursive(uid, component);
+        var xformQuery = EntityManager.GetEntityQuery<TransformComponent>();
+        UpdateEntityRecursive(uid, xformQuery, component);
     }
 
-    private void UpdateEntityRecursive(EntityUid uid, TransformComponent? transformComponent = null)
+    private void UpdateEntityRecursive(EntityUid uid, EntityQuery<TransformComponent> xformQuery, TransformComponent? transformComponent = null)
     {
         if(!Resolve(uid, ref transformComponent))
             return;
```
ElectroJr: This doesn't actually use the `xformQuery` here.

PaulRitter (approved):
what electro said, otherwise lgtm

Robust.Server/GameStates/PVSSystem.cs
PaulRitter: can resolves even use entityquery? since i assume not, do we have a resolve-method on entityquery?

ElectroJr: I guess you could "manually" do the resolve logic here, but yeah If the query stuff is gone be used a lot in other systems it might be useful to have that.

Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
metalgearsloth: Wouldn't you need to inherently recurse upwards first to determine if there's another container at some other level?

ElectroJr: Yeah I think you're right. You'd have to recourse upwards at least once.

metalgearsloth: This is gonna be a massive pain so I might come back to this PR in the future.

metalgearsloth: Okay passing by value worked at least so not awful.

Robust.Server/GameStates/PVSSystem.cs
metalgearsloth: future pr for resolve

PR #2495, opened 2022-02-02
Replace `EntityDirtyEvent` with C# event.

ElectroJr:
Currently whenever any entity is dirtied, `EntityManager` will raise a "normal" `EntityDirtyEvent` via the event bus. This event is currently used by both PVS and the `TransformSystem`. However the transform system just uses it to set `LastAnchoredModifiedTick`, which is currently completely unused (an old PVS thing?). So currently it just results in wasting time on meta-data, transform and map-grid component fetching whenever an entity is dirtied.

So this PR just removes `LastAnchoredModifiedTick`. It also replaces `EntityDirtyEvent` with a C# event, which AFAIK should be faster, which matters for something as common as just dirtying an entity.

I also moved the PVS system's ` EntityLifeStage` check into `EntityManager`'s `Dirty()`, so that dirty events are only ever be raised if the entity has been initialized, but currently that makes no difference as only PVS actually uses this event, and doing so saves a meta-data resolve.

Also has an edit to `RecursiveDeleteEntity()` to avoid retrieving the metadata component twice.

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.Dirty.cs
+++ b/Robust.Server/GameStates/PVSSystem.Dirty.cs
@@ -39,6 +39,7 @@ namespace Robust.Server.GameStates
         private void ShutdownDirty()
         {
             EntityManager.EntityAdded -= OnEntityAdd;
+            EntityManager.EntityDirtied -= OnEntityDirty;
         }
 
         private void OnEntityAdd(object? sender, EntityUid e)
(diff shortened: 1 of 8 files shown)


Robust.Server/GameStates/PVSSystem.Dirty.cs
```diff
@@ -33,7 +33,7 @@ private void InitializeDirty()
                 _dirtyEntities[i] = new HashSet<EntityUid>(32);
             }
             EntityManager.EntityAdded += OnEntityAdd;
-            SubscribeLocalEvent<EntityDirtyEvent>(OnDirty);
+            EntityManager.EntityDirtied += OnEntityDirty;
```
metalgearsloth, line 36: Should be unsubscribed on shutdown.

metalgearsloth (approved):
Curious on C# event vs struct performance, would be nice to have profile.

PR #2496, opened 2022-02-02
Misc physics/shutdown related optimizations

ElectroJr:
- Move `CollisionWakeComponent` shutdown to system and only run it if entity is not terminating.
- Move `PhysicsComponent`'s `EntParentChangedMessage` subscription into SharedPhysicsSystem 
  - Previously in `SharedBroadPhaseSystem`, but this subscription is also required for another function (duplicate subscription issues).
  - New subscription calls `SharedBroadPhaseSystem.UpdateBroadphase()` & the old `HandleParentChange` (now called `HandleParentChangeVelocity`
  - Less transform resolves, and avoids updating entity velocity when its being detached to null during deletion.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -101,9 +101,12 @@ namespace Robust.Shared.GameObjects
 
         private void HandleParentChange(EntityUid uid, PhysicsComponent body, ref EntParentChangedMessage args)
         {
-            if (LifeStage(uid) != EntityLifeStage.MapInitialized || !TryComp(uid, out TransformComponent? xform))
+            if (LifeStage(uid) is < EntityLifeStage.Initialized or > EntityLifeStage.MapInitialized
+                || !TryComp(uid, out TransformComponent? xform))
+            {
                 return;
-
+            }
+                
             if (body.CanCollide)
                 _broadphase.UpdateBroadphase(body, xform: xform);
             
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -1,6 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
+using System;
 using Prometheus;
 using Robust.Shared.Configuration;
 using Robust.Shared.Containers;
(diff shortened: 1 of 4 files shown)


Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -104,18 +99,26 @@ private void HandlePhysicsMapRemove(EntityUid uid, SharedPhysicsMapComponent com
             component.ContactManager.Shutdown();
         }
 
-        private void HandleParentChange(ref EntParentChangedMessage args)
+        private void HandleParentChange(EntityUid uid, PhysicsComponent body, ref EntParentChangedMessage args)
         {
-            var entity = args.Entity;
+            if (LifeStage(uid) != EntityLifeStage.MapInitialized)
+                return;
```
metalgearsloth: This isn't going to work for mapping I don't think?

Robust.Shared/Physics/SharedBroadphaseSystem.cs
```diff
@@ -294,22 +291,11 @@ internal void Cleanup()
...
-
-            UpdateBroadphase(component);
-        }
-
         /// <summary>
         /// If our broadphase has changed then remove us from our old one and add to our new one.
         /// </summary>
         /// <param name="body"></param>
-        private void UpdateBroadphase(PhysicsComponent body, FixturesComponent? manager = null, TransformComponent? xform = null)
+        public void UpdateBroadphase(PhysicsComponent body, FixturesComponent? manager = null, TransformComponent? xform = null)
```
metalgearsloth: internal, content should never call this.

metalgearsloth (commented):
I see you're getting that authentic physics experience so I wish you luck.

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
ElectroJr: Previously broadphase had a 
```cs
if (lifestage is < EntityLifeStage.Initialized or > EntityLifeStage.MapInitialized) return;
```
and I don't think the whole parent-change velocity transfer stuff matters pre map init.
Or was the broadphase check just bad in the first place?

ElectroJr: ah wait never mind, mistook < for <=

metalgearsloth (approved):

PR #2501, opened 2022-02-02
EntityQueryEnumerator "fix"

metalgearsloth:
Unfortunately it's still slower than the IEnumerable version no matter what I do.


Acruid:
If it is slower than just passing an IEnumerable, you should remove it.

metalgearsloth:
Probably yeah.

PR #2503, opened 2022-02-03
Fix multiline errors, add some tests

Ygg01:
There were some errors with reporting multiline errors.

Adds counting newlines to ensure that the `ErrorSpan.StartMark` matches the line it was created on.

Old behavior:
```
a = b {
  err = x 
                 ^ Unbalanced closing brace
```
New behavior:
```
a = b {
  err = x 
  ^ Unbalanced closing brace
```


ElectroJr (approved):
Don't really know anything about fluent or how `FormatCompileErrors()` is used, but this seems fine and does appear to fix formatting for the warning that used to be generated with #2532 before space-wizards/space-station-14/pull/6672 was merged.

Ygg01:
> Don't really know anything about fluent or how `FormatCompileErrors()` is used

A brief tour then. Fluent on error returns a `ParseError`.
We convert that to `ErrorSpan` for fancy errors.

`ErrorSpan` has the following fields:

`Row` to show row error occurred. Currently broken.
`StartSpan`/`EndSpan` to know what part to slice out of resource for displaying errors.
`StartMark`/`EndMark` to know what part of a previous string to mark with `^`.

Previous errors worked fine for single line errors, but as shown in the example, they don't take newlines into account. E.g. span says 0..14 and mark says 6..7. Before the algo just prints 0..14 newlines and whatnot.
Then prints 6 spaces followed by `^`.

PR #2504, opened 2022-02-03
Appearance data event

metalgearsloth:
We agreed that AppearanceVisualizers should probably be deprecated in favour of just using a system instead.


Robust.Client/GameObjects/EntitySystems/AppearanceSystem.cs
```diff
@@ -22,18 +76,33 @@ public override void FrameUpdate(float frameTime)
...
         }
     }
+
+    /// <summary>
+    /// Raised whenever the appearance data for an entity changes.
+    /// </summary>
+    public sealed class AppearanceChangeEvent : EntityEventArgs
+    {
+        public IReadOnlyDictionary<object, object> AppearanceData = default!;
+    }
```
metalgearsloth: The relevant bit.

PR #2506, opened 2022-02-04
Replace xform children sortedset with hashset

### LINKED ISSUE #2488: SortedSet<T> probably needs re-implementing
image

If I spawn 40 dummy pathfinding mobs it's no.1 and no.2 on allocs.

metalgearsloth:
We don't need the ordering and SortedSet allocates out the ass when iterating so this leads to massive savings.

Resolves https://github.com/space-wizards/RobustToolbox/issues/2488


metalgearsloth:
Somehow nothing blew up when I tried it in-game and acruid's okay with it.

Worst case we revert if there's some hidden bugs or we need the sorting.

mirrorcult (approved):

PR #2511, opened 2022-02-05
Fix InteractionOutline 2: Electric boogaloo

ElectroJr:
Branched off from #2476 as that had some issues.

Most notable changes:
- Originally, did not properly account for no-rot sprites. This fixes it by replacing `SpriteComponent.CalculateBoundingBox()` with `CalculateRotatedBoundingBox()` and makes it eye-dependent.
- `screenSpriteSize` calculation in `Clyde.HLR` was flawed. It took the width/height of a box as a vector and rotated that vector to get the width/height of the rotated box. This could easily result zero width/height sprites. Now it directly takes the sprite bounding box in screen coordinates and converts that into a size.

some comments/discussion happened in metalgearsloth/RobustToolbox/pull/1.

## Videos
Demo showing no-rot & direction sprites, with duplicates that have scaling, off-set, & rotation, showing both hover-outlines & sprite bounding box overlays. 

Current master:

https://user-images.githubusercontent.com/60421075/152624495-0ae3e8d0-af35-408e-9215-6e96b8762751.mp4

This version:

https://user-images.githubusercontent.com/60421075/152624522-7b93ce2d-3e7e-49a9-9a22-ab45382cc0a2.mp4

Note that there is currently an issue where stretched directional sprites look funky as they rotate. The sprite bounding box also does not match the actual drawn sprite, but IMO the bounding box is "correct" and the sprite rendering needs to be updated to not stretch/squash the sprite as it rotates. There are also some click map issues for offset/rotated sprites, but that is a sperate problem.


Robust.Client/Graphics/Clyde/Clyde.HLR.cs
```diff
@@ -369,30 +354,37 @@ private void DrawEntities(Viewport viewport, Box2Rotated worldBounds, Box2 world
...
             var xforms = _entityManager.GetEntityQuery<TransformComponent>();
 
+            // Construct a matrix equivalent for Viewport.WorldToLocal()
+            eye.GetViewMatrix(out var viewMatrix, view.RenderScale);
+            var uiProjmatrix = Matrix3.Identity;
+            uiProjmatrix.R0C0 = EyeManager.PixelsPerMeter;
+            uiProjmatrix.R1C1 = -EyeManager.PixelsPerMeter;
+            uiProjmatrix.R0C2 = view.Size.X / 2f;
+            uiProjmatrix.R1C2 = view.Size.Y / 2f;
+            var worldToLocal = viewMatrix * uiProjmatrix;
```
ElectroJr, line 369: So there might be much easier ways of doing this.

For one, if DrawEntities is only ever called from RenderViewport like it is ATM, then _currentMatrixView should always be equal to the `viewMatrix` that is calculated here. But I don't know if I can assume that or if it really matters.

Also there's probably an easier way to get uiProjmatrix? but `EyeManager.GetScreenProjectionMatrix` uses the display-size rather then the viewport size, which I assume matters?

Robust.Client/GameObjects/Components/Renderable/SpriteBoundsOverlay.cs
```diff
@@ -84,10 +84,17 @@ protected internal override void Draw(in OverlayDrawArgs args)
 
                 foreach (var sprite in comp.SpriteTree.QueryAabb(localAABB))
                 {
-                    var worldPos = _entityManager.GetComponent<TransformComponent>(sprite.Owner).WorldPosition;
-                    var bounds = sprite.CalculateBoundingBox(worldPos);
+                    var (worldPos, worldRot) = _entityManager.GetComponent<TransformComponent>(sprite.Owner).GetWorldPositionRotation();
+                    var bounds = sprite.CalculateRotatedBoundingBox(worldPos, worldRot);
+
+                    // Get scaled down bounds used to indicate the "south" of a sprite.
+                    // Or at least thats what I think they are meant to do.
```
metalgearsloth: Yeah it is

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1567,21 +1570,33 @@ public Box2 CalculateBoundingBox(Vector2 worldPos)
...
+            }
+            else
+            {
+                finalRotation = Rotation + worldRotation;
+            }
+
+            // I think there might be a problem here when you have directional-sprites with unique bounding boxes, where
+            // maaaaybe the final rotation requires a CalcRectWorldAngle(). But currently, I don't even know if a sprite
+            // like this exists anywhere for testing (sprites-sizes are generally 32x32 pixels, transparency is not
+            // accounted for in the layer bounding box).
```
metalgearsloth: I believe one of the old xeno sprites was 64 x 32

Robust.Client/GameObjects/EntitySystems/RenderingTreeSystem.cs
```diff
@@ -369,13 +369,10 @@ public override void FrameUpdate(float frameTime)
...
             var xforms = EntityManager.GetEntityQuery<TransformComponent>();
-
             var xform = xforms.GetComponent(value.Owner);
             var (worldPos, worldRot) = xform.GetWorldPositionRotation();
-            var bounds = new Box2Rotated(value.CalculateBoundingBox(worldPos), worldRot, worldPos);
-            var tree = GetRenderTree(value.Owner, xforms);
 
-            return tree == null ? bounds.CalcBoundingBox() : xforms.GetComponent(tree.Owner).InvWorldMatrix.TransformBox(bounds);
+            return SpriteAabbFunc(value, worldPos, worldRot, xforms);
         }
```
metalgearsloth, line 376: Does this cause problems not being in tree space? 

Because the AABB being passed into the tree for this is in tree space.

The light function below is also used in a similar fashion for non-approx queries.

Robust.Client/Graphics/Clyde/Clyde.HLR.cs
metalgearsloth: @PJB3005

Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
```diff
@@ -1019,7 +1019,8 @@ public int Compare(int x, int y)
                     return cmp;
                 }
 
-                cmp = _drawList[y].Item4.CompareTo(_drawList[x].Item4);
+                // compare the top of the sprite's BB for y-sorting. Because screen coordinates are flipped, the "top" of the BB is actually the "bottom".
+                cmp = _drawList[x].Item4.Top.CompareTo(_drawList[y].Item4.Top);
```
metalgearsloth, line 1023: Does rotation cause problems for this? Ideally it would use the lowest point (which afaik would require calculating the AABB and then taking the bottom of that).

Robust.Client/GameObjects/EntitySystems/RenderingTreeSystem.cs
ElectroJr: I'm not sure what you mean. If you're talking about the removal of the `xforms.GetComponent(tree.Owner).InvWorldMatrix.TransformBox(bounds)`, that's still done inside of `SpriteAabbFunc(value, worldPos, worldRot, xforms)`.

The coordinate system shouldn't have changed, or was this already a problem before?

Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
ElectroJr: WDYM would rotations cause issues? Assuming the bounding box calculation & screen transformation is correct (and rotations didn't cause issues there), then the `Item4`s should be the aabbs of each sprite in screen-coordinates.

The "bottom" of each aabb should the lowest-point of any sprite on the screen. But its just using `Box2.Top` because the Y-coordinates is flipped in screen coordinates.

Just testing with two players, one of which had a double-scaled sprite, the layering seemed to properly swap when the feet of a player moved below the other.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
ElectroJr: dammit
oh no

So clickable still has some rounding bug. breaks at -135 degrees but -135.01  or -134.99 work fine.

Robust.Client/GameObjects/EntitySystems/RenderingTreeSystem.cs
metalgearsloth: Ehh it's probably fine.

Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
metalgearsloth: Are there any sprites that are flipped to check with? Because if the sprite is inverted then we should check bottom instead I would imagine.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
ElectroJr: But yeah looks like non-rectangular normal-rotation 4-dir sprites have some problems I need to fix. So adding a DNM label for now

some problem

Robust.Client/Graphics/Clyde/Clyde.Rendering.cs
ElectroJr: By flipped do you just mean the sprite component has a 180 degree rotation?

As far as I can tell it works fine with both sprite and world rotation. Though for some angles the bottom of the sprite bb on screen isn't actually the same as the bottom of the visible part of the sprite on screen (see center clown) 

But other than switching from a sprite-bb to something more click-map-esque where it finds the bottom-most visible part of a sprite, this seems good enough.

https://user-images.githubusercontent.com/60421075/153240145-2da0ba8e-8238-4487-9195-6b40b8b7dd50.mp4

ElectroJr: A better demo with rotations and both no-rot and normal sprites. The draw-order should always prefer sprites with the BB that reaches the lowest point on screen.

https://user-images.githubusercontent.com/60421075/153240625-d0272933-8a3a-4072-9c17-1d40d8347858.mp4

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
ElectroJr: Okay so I think its fixed by 7604915. 

For reference, this is what an 8-directional non-square sprite looked like before that commit:

https://user-images.githubusercontent.com/60421075/153411981-42993481-504d-419a-8cfe-7140761ebbdc.mp4

And this is after, with no-rot sprites in the second column, offset & rotated in the middle row, and with a direction override in the last row.

https://user-images.githubusercontent.com/60421075/153412385-95d404b1-0028-479d-af20-a3cba4d3fc48.mp4

PR #2519, opened 2022-02-08
More transform optimisations

metalgearsloth:
Take 2. In my shitty benchmarking across every entity on saltern it's like 80-90% of the speed so it's probably similar where there's 1 or 2 parents but faster for anymore.


Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -211,15 +217,21 @@ public Matrix3 WorldMatrix
...
+                var xformQuery = _entMan.GetEntityQuery<TransformComponent>();
+                var parent = _parent;
+                var myMatrix = GetLocalMatrix();
+
+                while (parent.IsValid())
                 {
-                    var parentMatrix = Parent!.WorldMatrix;
-                    var myMatrix = GetLocalMatrix();
+                    var parentXform = xformQuery.GetComponent(parent);
+                    var parentMatrix = parentXform.GetLocalMatrix();
```
ElectroJr: Seeing as `parentXform.GetLocalMatrix()` literally just returns `parentXform._localMatrix`, would it be faster to just directly pass that to `Matrix3.Multiply`? And I guess same for the inverse matrix.

metalgearsloth: Seems like it would make more sense to aggressively inline it then?

Acruid: It should already be inlined, though IDK if Rider can look at IL and ASM to check.

PR #2520, opened 2022-02-08
Map Pausing Fixes

### LINKED ISSUE #1444: A grid added after a map is paused has its grid entity unpaused

### LINKED ISSUE #1445: Entities added to a paused grid afterwards are not themselves paused

Acruid:
This actually works on the server now.

Resolves #1444
Resolves #1445

Also merged PauseManager into MapManager, and made GridIds convertable to EntityUids.

### DIFF
--- a/Robust.Shared/GameObjects/Components/IgnorePauseComponent.cs
+++ b/Robust.Shared/GameObjects/Components/IgnorePauseComponent.cs
@@ -1,25 +1,24 @@
 using Robust.Shared.IoC;
-using Robust.Shared.Timing;
+using Robust.Shared.Map;
 
 namespace Robust.Shared.GameObjects
 {
     [RegisterComponent]
     public sealed class IgnorePauseComponent : Component
     {
-        [Dependency] private readonly IEntityManager _entMan = default!;
-
         protected override void OnAdd()
         {
             base.OnAdd();
-            _entMan.GetComponent<MetaDataComponent>(Owner).EntityPaused = false;
+            IoCManager.Resolve<IEntityManager>().GetComponent<MetaDataComponent>(Owner).EntityPaused = false;
         }
 
         protected override void OnRemove()
         {
             base.OnRemove();
-            if (IoCManager.Resolve<IPauseManager>().IsMapPaused(_entMan.GetComponent<TransformComponent>(Owner).MapID))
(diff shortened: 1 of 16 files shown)


Robust.Shared/GameObjects/Components/IgnorePauseComponent.cs
```diff
@@ -1,25 +1,24 @@
 using Robust.Shared.IoC;
-using Robust.Shared.Timing;
+using Robust.Shared.Map;
 
 namespace Robust.Shared.GameObjects
 {
     [RegisterComponent]
     public sealed class IgnorePauseComponent : Component
```
metalgearsloth, line 7: I should really nuke this component.

Robust.Shared/Map/MapManager.Pause.cs
```diff
@@ -2,101 +2,121 @@
...
+        {
+            // RunMapInit can modify the TransformTree
+            // ToArray caches deleted euids, we check here if they still exist.
+            if(!metaQuery.TryGetComponent(entity, out var meta))
+                return;
 
-                // MapInit could have deleted this entity.
-                if(_entityManager.TryGetComponent(entity, out MetaDataComponent? meta))
-                    meta.EntityPaused = false;
+            entity.RunMapInit();
```
metalgearsloth, line 77: This can probably be ECS'd now I think if you want to in this PR? IMapInit is dead.

metalgearsloth (approved):

Robust.Shared/Map/MapManager.Pause.cs
Acruid: ECSing MapManager.Pause is out of scope, and I tried to not mess with the implementation here. Removing the RunMapInit helper function can be another PR.

PR #2521, opened 2022-02-08
Make AppearanceSystem public

ElectroJr:
Is there any good reason for it to be internal? It would be useful in some instances to expose `MarkDirty()` for client-side systems

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AppearanceSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AppearanceSystem.cs
@@ -6,7 +6,7 @@ using Robust.Shared.GameStates;
 namespace Robust.Client.GameObjects
 {
     [UsedImplicitly]
-    internal sealed class AppearanceSystem : SharedAppearanceSystem
+    public sealed class AppearanceSystem : SharedAppearanceSystem
     {
         private readonly Queue<ClientAppearanceComponent> _queuedUpdates = new();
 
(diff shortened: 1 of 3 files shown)


metalgearsloth:
This would be mainly for stuff using reference types where it's unable to be set properly right?

Though I'm not sure if this interferes with photography.

ElectroJr:
> This would be mainly for stuff using reference types where it's unable to be set properly right?

In this case it was for space-wizards/space-station-14/pull/6565, for its use of the toggle-subfloor command. But after looking again, it might not be needed? I'll try fix it without this PR and see how I go, and maybe close this one.

ElectroJr:
> But after looking again, it might not be needed? I'll try fix it without this PR and see how I go, and maybe close this one.

Nope. Turns out the way the subfloor command currently works exploits the issue that #2460 fixes, and that fix will stop the current command from working properly.

So the annoying thing is that the visibility of individual layers is determined by visualizer data. If we want a client-side toggle-subfloor command we can't just override the appearance data, because then its state is dirty and it gets reset. Currently, because of the issue in @2460 this reset doesn't work properly and the subfloor command works fine currently.

So I don't know of a nice way to do the subfloor command without either having the command re-implement whatever pipe / subfloor visualizers may determine layer visibility and keeping that updated with new visualizers, OR: just making the command mark the client's appearance component as dirty (in the appearance, not network sense). Then all the required visualizers get run automatically. But the only way to do that is to make `SharedAppearanceSystem.MarkDirty()` public. The actual function already is, but the system is internal. Hence this PR.

If there's a good reason it shouldn't be, I guess I need input on what to do with space-wizards/space-station-14/pull/6565,

As to whether it would interfere with photography... I guess thats a general question that applies to the toggle sub-floor command. I think with this change, how it would actually work is that the command would toggle the visiblity of the subfloor in the photo?

PR #2522, opened 2022-02-08
Abstract VisualizerSystem

mirrorcult:
This introduces an abstract EntitySystem inheritor in the same vein as `VirtualController` or `GameRuleSystem` in SS14.

It just abstracts away having to create the event subscriptions for ComponentInit and OnChangeData, which is nice.

This also makes `AppearanceChangeEvent` ref struct, so it requires space-wizards/space-station-14#6571.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AppearanceSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AppearanceSystem.cs
@@ -100,9 +100,9 @@ namespace Robust.Client.GameObjects
     }
 
     /// <summary>
-    /// Raised whenever the appearance data for an entity changes.
+    ///     Raised whenever the appearance data for an entity changes.
     /// </summary>
-    public sealed class AppearanceChangeEvent : EntityEventArgs
+    public struct AppearanceChangeEvent
     {
         public AppearanceComponent Component = default!;
         public IReadOnlyDictionary<object, object> AppearanceData = default!;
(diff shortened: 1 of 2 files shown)


mirrorcult:
open question: should the methods be abstract or virtual, abstract may be better perf (????? not profiled), though we seal stuff now, and virtual means you can omit those methods if unnecessary

mirrorcult:
removed the init stuff since it can lead to double subscriptions

ElectroJr (approved):

Robust.Client/GameObjects/EntitySystems/AppearanceSystem.cs
```diff
@@ -100,9 +100,10 @@ public void OnChangeData(EntityUid uid, ClientAppearanceComponent? appearanceCom
     }
 
     /// <summary>
-    /// Raised whenever the appearance data for an entity changes.
+    ///     Raised whenever the appearance data for an entity changes.
     /// </summary>
-    public sealed class AppearanceChangeEvent : EntityEventArgs
```
ElectroJr, line 105: I think you also need to update AppearanceSystem to raise as a ref?

mirrorcult: oh true

Robust.Client/GameObjects/EntitySystems/VisualizerSystem.cs
```diff
@@ -0,0 +1,20 @@
...
+    where T: Component
+{
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<T, AppearanceChangeEvent>(OnAppearanceChange);
+    }
+
+    protected virtual void OnAppearanceChange(EntityUid uid, T component, ref AppearanceChangeEvent args) {}
```
metalgearsloth, line 19: I benchmarked and abstract / virtual looked identical to me; sealing the classes themselves had a significant impact though.

PR #2524, opened 2022-02-09
Optimise PVS recursion

metalgearsloth:
Before:
old_pvs

After:
new_pvs

40 pathfinding dummies with gear on over a minute.

Gonna go for the easy wins before I start trying to get the hard ones.

@PaulRitter

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -24,6 +24,7 @@ internal sealed partial class PVSSystem : EntitySystem
     [Shared.IoC.Dependency] private readonly IPlayerManager _playerManager = default!;
     [Shared.IoC.Dependency] private readonly IConfigurationManager _configManager = default!;
     [Shared.IoC.Dependency] private readonly IServerGameStateManager _stateManager = default!;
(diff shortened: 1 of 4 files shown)


Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -170,32 +170,62 @@ private void OnEntityDeleted(object? sender, EntityUid e)
...
+
+        // Parented directly to the grid
+        if (grid.GridEntityId == xform.ParentUid)
+            return xform.Coordinates;
+
+        // Parented to grid so convert their pos back to the grid.
+        var gridPos = Transform(grid.GridEntityId).InvWorldMatrix.Transform(xform.WorldPosition);
+        return new EntityCoordinates(grid.GridEntityId, gridPos);
+    }
+
```
metalgearsloth: I might end up re-using this for entitylookup so it might get moved to SharedTransformSystem, but also if I do that then I need to fix a bunch of unit tests (most of them are mocks that can be deleted but some need updating).

PaulRitter (approved):

metalgearsloth:
Gonna add a test first

PR #2528, opened 2022-02-10
Allow content to override occlusion directions

metalgearsloth:
Useful for OD.
Also avoided the 3(!) transform resolves per occluder update.

Triggering an update requires raising an OccluderDirtyEvent still.


metalgearsloth:
Might end up just deleting the test as there's no way for the client to do it on its own and there's no easy way to connect it to server for robust integration.

PR #2530, opened 2022-02-11
Allow specifying that sprite layer offsets shouldn't rotate

mirrorcult:
This is useful for 'absolute' translations of a layer where you dont actually want it to change with rotation, like translating clothing sprites for SS14 (space-wizards/space-station-14#6644)

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -211,6 +211,7 @@ namespace Robust.Client.GameObjects
                     layer.Color = layerDatum.Color;
                     layer.Rotation = layerDatum.Rotation;
                     layer._offset = layerDatum.Offset;
+                    layer.RotateOffset = layerDatum.RotateOffset;
                     // If neither state: nor texture: were provided we assume that they want a blank invisible layer.
                     layer.Visible = anyTextureAttempted && layerDatum.Visible;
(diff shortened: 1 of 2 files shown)


mirrorcult:
wait okay this stopped working for some reason

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1271,7 +1272,9 @@ private void RenderInternal(DrawingHandleWorld drawingHandle, Angle eyeRotation,
 
                 var numDirs = GetLayerDirectionCount(layer);
                 var layerRotation = worldRotation + layer.Rotation;
-                var layerPosition = worldPosition + layerRotation.RotateVec(layer._offset + offset);
+                var layerPosition = worldPosition + (layer.RotateOffset
+                    ? layerRotation.RotateVec(layer._offset + offset)
+                    : layer._offset + offset);
```
ElectroJr, line 1277: Will probably require updating clickable/clickmap logic, but thats in content and I'm happy to do that.

This should probably be merged first and then #2511 can be updated to properly do layer rotations.

ElectroJr: Actually I don't think this offset feature is need at all. Layer transforms just need to be fixed.

mirrorcult: yeah

PR #2532, opened 2022-02-11
Enable Fluent logs for only .ftl files

Ygg01:
Due to overly restrictive log levels, `.ftl` parsing errors were silenced.

This commit will break `Content` because it was already broken.


ElectroJr:
> This commit will break Content because it was already broken.

Even though its already broken, I'd rather merge this once there's a complementary content PR to get rid of the test failures, because until that's done test results will just not be as useful

Any idea roughly how many content fluent errors there are?

Ygg01:
> Any idea roughly how many content fluent errors there are?

One in suspicion mode.

PR #2550, opened 2022-02-16
Upgrade NUnit nuget packages

ZeWaka:
See commits for upgrade considerations (there are none)


ZeWaka:
lol wut that deps failure makes 0 sense

PJB3005:
You'll need to update the packages in content too.

PR #2557, opened 2022-02-18
Force get PVS parents for viewers

metalgearsloth:
So replicating this on master is CBT (vv'ing your parent to something across the station is unreliable) but AFAICT when you go over-budget there's no guarantee the viewer parents are sent which is bad.

Don't merge without @PaulRitter approval

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -477,8 +477,7 @@ internal sealed partial class PVSSystem : EntitySystem
 
         foreach (var viewerEntity in viewerEntities)
         {
-            var xform = xformQuery.GetComponent(viewerEntity);
-            var parent = xform.ParentUid;
+            var parent = viewerEntity;
 
             while (parent.IsValid())
             {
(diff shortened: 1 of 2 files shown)


Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -471,8 +477,16 @@ public void ReturnToPool(Dictionary<EntityUid, MetaDataComponent>?[] chunkCache,
...
+            var parent = viewerEntity;
+
+            while (parent.IsValid())
+            {
+                TryAddToVisibleEnts(in parent, seenSet, playerVisibleSet, visibleEnts, fromTick, ref newEntitiesSent,
+                    ref entitiesSent, mQuery, dontSkip: true);
+
+                var parentXform = xformQuery.GetComponent(parent);
+                parent = parentXform.ParentUid;
+            }
```
metalgearsloth, line 489: Originally I had the viewer done at the end but this causes some ordering bug where you just stop getting rendered. If your parent becomes some far off entity you flicker for like a frame but that also happens on master.

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -130,6 +130,7 @@ public void SendGameStateUpdate()
 
             //todo paul oh my god make this less shit
             EntityQuery<MetaDataComponent> metadataQuery = default!;
+            EntityQuery<TransformComponent> transformQuery = default!;
```
metalgearsloth, line 133: :imfuckingscreaming:

metalgearsloth:
Superseded by https://github.com/space-wizards/RobustToolbox/pull/2558

PR #2560, opened 2022-02-20
Optimise PVS cangetcompstate

metalgearsloth:
Most of it is just going to come from the ordering swap where we check if there's even a compstate to add or get before raising the event.

Content PR https://github.com/space-wizards/space-station-14/pull/6815

@PaulRitter


Robust.Shared/GameObjects/Components/MetaDataComponent.cs
```diff
@@ -192,4 +173,14 @@ internal override void ClearTicks()
...
+
+    [Flags]
+    public enum MetaDataFlags : byte
+    {
+        None = 0,
+        /// <summary>
+        /// Whether the entity has states specific to a particular player.
+        /// </summary>
+        EntitySpecific = 1 << 0,
+    }
```
mirrorcult, line 185: why not just like a component

metalgearsloth: Fastest would just be caching the flag when we get metadata earlier to avoid the getcomponent entirely. I also foresee container culling using it unless we get better ideas but see what paul thinks.

gradientvera (commented):
I'm not particularly happy about `MetaDataSystem`'s `AddFlag` and `RemoveFlag` methods not using the resolves convention, but I guess if you want it to be as performant as possible it's fine.

Robust.Shared/GameObjects/Components/MetaDataComponent.cs
PaulRitter: we can definitely grab the metadatacomp from earlier in 99% of cases

PaulRitter (changes requested):
lgtm but i dont see you setting the flag on any entity yet, or am i missing smth?

metalgearsloth:
> I'm not particularly happy about `MetaDataSystem`'s `AddFlag` and `RemoveFlag` methods not using the resolves convention, but I guess if you want it to be as performant as possible it's fine.

That should be fine to change, as long as it's not in PVS it's whatever.

metalgearsloth:
Gonna re-profile before merge sometime but should be good now.

PaulRitter (approved):
a

PR #2570, opened 2022-02-27
Makes Robusttree Children nullable to save allocations

PaulRitter:


Robust.Server/GameStates/RobustTree.cs
```diff
@@ -95,11 +109,16 @@ public TreeNode Set(T child, T parent)
         if (!_nodeIndex.TryGetValue(parent, out var parentNode))
             parentNode = Set(parent);
 
+        if (parentNode.Children == null)
+        {
+            _nodeIndex[parent] = parentNode = parentNode.WithChildren(_setProvider());
+        }
```
metalgearsloth: For this particular bit the parent of the parent needs to update its reference to point to the new node. For testing I used this particular piece of jank (I also removed the pooling as well):

```
TreeNode parentsParentNode = default;

if (_parents.TryGetValue(parent, out var parentsParent))
{
    parentsParentNode = _nodeIndex[parentsParent];
    parentsParentNode.Children?.Remove(parentNode);
}

_nodeIndex[parent] = parentNode = parentNode.WithChildren(new HashSet<TreeNode>());

// Possible that the parent was made in Set() above but no children present yet for the child.
if (parentsParentNode.Children != null)
{
    parentsParentNode.Children.Add(parentNode);
    _nodeIndex[parentsParent!] = parentsParentNode;
}
```

I re-profiled and with 20 clients it's like 20-50% slower than old PVS at but I'm pretty sure it would still scale better even without other optimisations

metalgearsloth (approved):

PR #2572, opened 2022-02-28
loadmap command tweak

ElectroJr:
Make the loadmap command only report success if the map loading actually worked.

### DIFF
--- a/Robust.Server/Console/Commands/MapCommands.cs
+++ b/Robust.Server/Console/Commands/MapCommands.cs
@@ -218,7 +218,11 @@ namespace Robust.Server.Console.Commands
             }
 
             IoCManager.Resolve<IMapLoader>().LoadMap(mapId, args[1]);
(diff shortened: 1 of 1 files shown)


Robust.Server/Console/Commands/MapCommands.cs
```diff
@@ -218,7 +218,11 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
             }
 
             IoCManager.Resolve<IMapLoader>().LoadMap(mapId, args[1]);
-            shell.WriteLine($"Map {mapId} has been loaded from {args[1]}.");
+
+            if (mapManager.MapExists(mapId))
+                shell.WriteLine($"Map {mapId} has been loaded from {args[1]}.");
+            else
+                shell.WriteLine($"Error while loading map from {args[1]}.");
```
metalgearsloth:
```suggestion
                shell.WriteError($"Error while loading map from {args[1]}.");
```

PR #2573, opened 2022-02-28
EntityLookup as a system

metalgearsloth:
Builds on https://github.com/space-wizards/RobustToolbox/pull/2568 and https://github.com/space-wizards/RobustToolbox/pull/2571

I did end up deleting one test because I'm pretty sure the thing it was testing for would actually fail and the only reason it was passing was because of specific unit testing setup to prevent it.


Robust.Shared/GameObjects/Components/Collidable/PhysicsComponent.Physics.cs
```diff
@@ -286,13 +282,15 @@ public Box2 GetWorldAABB(Vector2? worldPos = null, Angle? worldRot = null)
             return bounds;
         }
 
-        public Box2 GetWorldAABB(Vector2 worldPos, Angle worldRot, EntityQuery<FixturesComponent> fixtures)
+        public Box2 GetWorldAABB(Vector2? worldPos = null, Angle? worldRot = null)
         {
-            var transform = new Transform(worldPos, (float) worldRot.Theta);
+            worldPos ??= _entMan.GetComponent<TransformComponent>(Owner).WorldPosition;
+            worldRot ??= _entMan.GetComponent<TransformComponent>(Owner).WorldRotation;
```
ElectroJr: If both are null, use `GetWorldPositionRotation`?

PR #2576, opened 2022-03-01
BaseContainer.Insert() will now reset the entity's rotation after res…

### LINKED ISSUE #6886: Point lights don't reset rotation when picked up
## Description

Point lights on rotated items keep their rotation when picked up which causes the light to not point forward.

**Screenshots**

image

**Additional context**

Tested with a flashlight and a PDA light.

RemberBM:
BaseContainer.Insert() will now reset the entity's rotation after resetting its position. This fixes https://github.com/space-wizards/space-station-14/issues/6886


ElectroJr (approved):
If there is ever a container that needs to preserve rotation, or just have non-zero rotation in general (something like a display cabinet?), they can just manually set the angle after insertion, so this fix seems fine to me

PR #2582, opened 2022-03-04
Layer draw depths / sorting

ElectroJr:
This PR adds a draw depth to sprite layers, which determines the order in which they are drawn. It also adds RSI-direction based overrides, so that you could choose to have a layer behind another if an entity is facing east or west.

The change is motivated bv space-wizards/space-station-14/issues/4387 and space-wizards/space-station-14/issues/4563, but is also required to fix some bugs that were introduced by space-wizards/space-station-14/pull/6252

I'm not sure about how I did the sorting. Initially I just kept an up to date list of pre-sorted layer indices and iterated over those in `RenderInternal()`. However, after adding support for the directional draw depths, you can't sort the layers without knowing the angle entity's angle, so afaik the sorting has to be done inside of `RenderInternal()`. I just did the sorting by creating a temporary list and using `,Sort()`, but I feel like there might be a better way of doing that?

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -1375,7 +1375,7 @@ namespace Robust.Client.GameObjects
             var angle = worldRotation + eyeRotation; // angle on-screen. Used to decide the direction of 4/8 directional RSIs
 
             // Sort layers by draw depth. This can be dependent on the RSI direction, and so it isn't pre-sorted.
-            List<QueuedLayer> sortedLayers = new(Layers.Count);
+            List<LayerEntry> sortedLayers = new(Layers.Count);
             foreach (var layer in Layers)
             {
                 if (!layer.Visible)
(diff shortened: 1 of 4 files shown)


Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1322,10 +1373,49 @@ private void RenderInternal(DrawingHandleWorld drawingHandle, Angle eyeRotation,
...
+        }
+
+        internal readonly struct QueuedLayer
+        {
+            public QueuedLayer(int depth, Layer layer, RSI.State? rsiState, RSIDirection direction)
+            {
+                Depth = depth;
+                Layer = layer;
+                RsiState = rsiState;
+                Direction = direction;
```
ElectroJr: In order to get the layer's direction, and thus the draw depth, it needs to get the RSI state, which is also required for the `Layer.Render()` function.

I'm not sure whether its better to cache them for each layer in this weird struct, or to just call `layer.GetActualState()` a second time inside `Layer.Render()`

ElectroJr:
Will come back to this eventually, probably with a cached pre-sorted list for each direction or something like that

PR #2585, opened 2022-03-05
Add a is-In-container metadata flag.

ElectroJr:
This PR makes a couple of changes with the goal to speed up some container functions.

The main change is to just give transform component a bool for quickly checking whether an entity is in a container. This helps speed up `IsEntityInContainer` and `TryGetContainingContainer` calls, which are used for most most interactions and several entity-lookups (e.g., explosions). Adding or removing an entity to a container, and the existing is-in-container checks all already involve fetching the transform component anyways. Previously the IsInContainer functions would then try to resolve a `ContainerManagerComponent` on the entity's parent, and if one exists, iterate over all containers and their contained entities to check that the given entity is actually inside of one (instead of just parented to something that happens to have containers),

Other changes include:
- Making Insert, remove, and some other functions take in optional transform component arguments.
- Mark some `ContainerHelpers` with direct `ContainerSystem` replacements as obsolete
- Slightly extending `ContainerSystem.EnsureContainer` so that it has the same functionality as the container helper function
	- Namely: adding an `out bool` that specifies whether the container already existed, or was newly created.


metalgearsloth:
what about storing it on metadata flags?

ElectroJr:
> what about storing it on metadata flags?

Yeah could also do that, though that would then require one additional resolve on insert/remove. But that's probably fine. I'll update the PR to do that.

Having more than one flags also means the flag being removed has to be sent in the removal attempt event, which probably should have been done in the first place.

ElectroJr:
Now using metadata flags. Requires a content PR now, because some container functions had their optional transform component replaced with a metadata component. Also generalised metadata flag removal to support more than one flag

see space-wizards/space-station-14/pull/7003

Robust.Shared/Containers/BaseContainer.cs
```diff
@@ -55,24 +55,27 @@ public abstract class BaseContainer : IContainer
         protected BaseContainer() { }
 
         /// <inheritdoc />
-        public bool Insert(EntityUid toinsert, IEntityManager? entMan = null)
+        public bool Insert(EntityUid toinsert, IEntityManager? entMan = null, TransformComponent? transform = null, TransformComponent? ownerTransform = null, MetaDataComponent? meta = null)
         {
             DebugTools.Assert(!Deleted);
+            DebugTools.Assert(transform == null || transform.Owner == toinsert);
+            DebugTools.Assert(ownerTransform == null || ownerTransform.Owner == Owner);
```
mirrorcult, line 62: resolves already assert this on debug

Robust.Shared/Containers/BaseContainer.cs
```diff
@@ -189,10 +194,13 @@ public virtual void Shutdown()
         /// </summary>
         /// <param name="toinsert"></param>
         /// <param name="entMan"></param>
-        protected virtual void InternalInsert(EntityUid toinsert, IEntityManager entMan)
+        protected virtual void InternalInsert(EntityUid toinsert, IEntityManager entMan, MetaDataComponent? meta)
```
mirrorcult:
```suggestion
        protected virtual void InternalInsert(EntityUid toinsert, IEntityManager entMan, MetaDataComponent? meta=null)
```

Robust.Shared/Containers/BaseContainer.cs
```diff
@@ -203,13 +211,16 @@ protected virtual void InternalInsert(EntityUid toinsert, IEntityManager entMan)
         /// </summary>
         /// <param name="toremove"></param>
         /// <param name="entMan"></param>
-        protected virtual void InternalRemove(EntityUid toremove, IEntityManager entMan)
+        protected virtual void InternalRemove(EntityUid toremove, IEntityManager entMan, MetaDataComponent? meta)
```
mirrorcult:
```suggestion
        protected virtual void InternalRemove(EntityUid toremove, IEntityManager entMan, MetaDataComponent? meta=null)
```

mirrorcult (changes requested):

Robust.Shared/Containers/BaseContainer.cs
ElectroJr: Yeah, but this isn't in a system, and is just using `??= entMan.GetComponent<>()` so those checks aren't being done without this assert.

Robust.Shared/Containers/ContainerHelpers.cs
```diff
@@ -16,24 +16,17 @@ namespace Robust.Shared.Containers
     public static class ContainerHelpers
     {
         /// <summary>
-        /// Am I inside a container?
+        /// Am I inside a container? Only checks the direct parent.
```
metalgearsloth: Should put a TODO on this because functionally containers are nested everywhere else regardless if the children are also in containers.

ElectroJr: I've just added a variant that does a recursive check.

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -106,21 +119,69 @@ public ContainerManagerComponent.AllContainersEnumerable GetAllContainers(Entity
...
+            if (xform == null)
+            {
+                xforms ??= EntityManager.GetEntityQuery<TransformComponent>();
+                xform = xforms.Value.GetComponent(uid);
+            }
+
+            if (!xform.ParentUid.Valid)
+                return false;
+
+            return IsEntityOrParentInContainer(xform.ParentUid, metas: metas, xforms: xforms);
```
metalgearsloth, line 184: This is similar to TryGetOuterContainer so can just update that with the flag and have this method call that one?

Robust.Shared/GameObjects/Systems/SharedGridTraversalSystem.cs
```diff
@@ -53,7 +53,8 @@ private void HandleMove(ref MoveEvent moveEvent, EntityQuery<TransformComponent>
         {
             var entity = moveEvent.Sender;
 
-            if (Deleted(entity) ||
+            if (!TryComp(entity, out MetaDataComponent? meta) ||
+                meta.EntityDeleted ||
```
metalgearsloth: Might as well pass EntityQuery in (grid traversal pretty expensive already sooooo yeahhhh)

metalgearsloth (approved):

Robust.Shared/Containers/SharedContainerSystem.cs
ElectroJr: TryGetOuterContainer will always iterate all the way up to the map looking for the outermost container, whereas this should terminate whenever it finds the first entity inside of a container. This also doesn't need to fetch the container manager component, so keeping them separate should keep this slightly faster.

But I will add flag-checks to TryGetOuterContainer.

edit: actually the metadata check might make things slower in this specific instance? It basically just adds an extra component-fetch

PR #2593, opened 2022-03-09
makes budgets clientside & replicated

PaulRitter:
DISCLAIMER: afaict the client replication for cvars doesnt get used anywhere else so im not sure if it even works. i did try this out ingame and pvs still works but i never tried changing the cvar & debugging if it actually arrived on the server.

@moonheart08


metalgearsloth (approved):

metalgearsloth:
> so im not sure if it even works

It works

PR #2594, opened 2022-03-10
Fix grid traversal velocities

ElectroJr:
This PR fixes three issues:
- The calculations for the total map velocity incorrect. `MapLinearVelocity` and `MapVelocities` don't even agree with each other
- The grid-traversal velocity/momentum conservation function is wrong, and will lead to you being flung off of a moving grid.
- World rotation is not conserved when moving across grids

These issues can be seen in this video, which is just master engine & content with a slowed-down RPG projectile:

https://user-images.githubusercontent.com/60421075/157578124-bbeb4d74-279d-4fa4-9ce5-3a32653cb1d1.mp4

This PR fixes the map velocity calculations, and moves them to a function in the physics system (rather than a component getter). The grid traversal calculations now just require that the map velocities remain unchanged as your parent changes. Gird traversal now looks like this:

https://user-images.githubusercontent.com/60421075/157578966-dbbd2827-7a64-469e-9883-cb63fb48bce3.mp4

## Caveats

This does not fix an existing issue with mispredicts that happen when players move across grids. This is really noticeable with high ping or rapidly rotating grids. I'm not sure what causes the mispredicts, but it only seems to happen for player movement.
I suspect it might have something to do with a player resetting transform comp state, which leads to a parent-change, which will then modify the velocities during resetting, leading to a bad physics comp reset? Maybe?

This also doesn't fix another existing issue: parent changes due to teleportation will lead to potentially unexpected changes to velocities. IMO this is probably fine. If you want to teleport a moving object elsewhere, it should be the teleporters responsibility to deal with momentum/velocity conservation.


Robust.UnitTesting/Shared/Physics/MapVelocity_Test.cs
```diff
@@ -21,42 +21,138 @@ public sealed class MapVelocity_Test : RobustIntegrationTest
     bodyType: Dynamic
   - type: Fixtures
 ";
-
         [Test]
-        public async Task Test()
+        public async Task TestMapVelicities()
```
metalgearsloth:
```suggestion
        public async Task TestMapVelocities()
```

your kiwi is showing

metalgearsloth (approved):

PR #2600, opened 2022-03-13
Dubious entity deletion prediction

ElectroJr:
So I came up with this while trying to get item-stack prediction to work. Basically, instead of deleting server side entities, this just sending them to null space works well enough in most situations. If they actually get deleted by the server, they get deleted as normal. If there was a mispredict the transform-reset **should** fix it,

This PR makes it so that the client overrides the entity manager deletion. There is a separate function function that **actually** deletes the entity. The way I've done this is questionable and maybe very stupid. E.g.:
- This "deletion" doesn't trigger any of the normal component shutdown stuff, it doesn't even mark the entity as deleted (until normal / "real" deletion happens)
- The dubious way I've implemented it might mean people accidentally call the wrong function from client-side code, resulting in an entity not actually being deleted when it should be
- The client currently only **properly** deletes an entity when either when receiving a list of deletions from the server, or when calling `FlushEntities()`. I might've missed something.

So yeah not sure about this. It is also not strictly required. E.g., to get stack-prediction working I could just make the client-side stack system detach-to-null rather than calling `Delete()`. At least then it's a deliberate action that shouldn't lead to unintended side effects when `Delete()` is used elsewhere.

### DIFF
--- a/Robust.Client/GameObjects/ClientEntityManager.cs
+++ b/Robust.Client/GameObjects/ClientEntityManager.cs
@@ -48,6 +48,41 @@ namespace Robust.Client.GameObjects
             base.StartEntity(entity);
         }
 
+        void IClientEntityManagerInternal.ActuallyDeleteEntity(EntityUid uid)
+        {
+            DeleteEntityInternal(uid);
+        }
+
+        public override void DeleteEntity(EntityUid uid)
+        {
+            if (uid.IsClientSide())
+            {
+                DeleteEntityInternal(uid);
+                return;
+            }
+
+            // Client-side code is predicting that a server-side entity will get deleted.
+            // Until confirmation is received from the server, simply mark it as deleted and move it to null-space.
+            // If there was a mispredict, the metadata & transform component resets should fix it.
+
+            // This is somewhat of a hack for predicting entity deletions, and will not properly fire things like
+            // component shutdown & remove events. It doesn't even mark the entity as actually deleted, because the
+            // meta-data component doesn't netowork / reset that (though AFAIK there is no reason it couldn't?).
+
+            if (!TryGetComponent(uid, out MetaDataComponent? meta) || meta.EntityDeleted)
+                return;
+
+            if (meta.EntityLifeStage == EntityLifeStage.Terminating)
+#if !EXCEPTION_TOLERANCE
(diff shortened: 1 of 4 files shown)


Acruid (changes requested):
I'm not sure what problem this is trying to solve, but the way this is implemented doesn't work at all. It's not even predicting entity deletion, it's just detaching the entity to nullspace.

metalgearsloth:
> It's not even predicting entity deletion, it's just detaching the entity to nullspace.

> Basically, instead of deleting server side entities, this just sending them to null space works well enough in most situations. If they actually get deleted by the server, they get deleted as normal. If there was a mispredict the transform-reset should fix it,

It's trying to solve having to undelete entities that were predicted to be deleted.

ElectroJr:
Previously the client would only delete entities once it receives a new server state, and calling Delete() in shared code would just cause errors. This is is just making the client hide the entity in null space until it receives the deletion instruction from the server. If that doesn't arrive because it was a mispredict or something, the entity should be properly reset back to where it was. So yeah this isn't really predicting deletion, just side stepping it by making the client hide the entity.

As to the problem its trying to solve: its to allow predicting interactions that involves entity deletion. One example being stack merging. If you try to pick up a floor tile by adding it to a held stack, one of the entities gets deleted. So currently the interaction has to be all server side, and you get the normal latency induced interaction delay after clicking on things which feels shit. As mentioned, an alternative to this (or just having proper deletion & un-deletion) would be to just make the client-side stack system hide the entity (w/o any engine changes). Giving `IEntityManager.Delete()` this behaviour is really just for convenience, so that deletions can easily be done in shared code without accidentally breaking things.

Acruid:
>During prediction, I want a way to just mark an entity as will-soon-be-deleted-by-server & hide it from the client, and then either free it when the server state comes in, or revert the deletion if a mispredict happens.

This is the problem that the PR is trying to solve. Components are marked for deletion by setting their lifestage to `Deleted`, but EntityManager code has been changed to immediately free components from memory when they are deleted in prediction. Previously deferred deletion of components allegedly was causing bugs, which prompted the change. The component Lifestage is also not serialized, so it is unknown how reverting the lifestage would work when a mispredict happens.

PR #2605, opened 2022-03-15
New Serv3 api just dropped

PaulRitter:
requires https://github.com/space-wizards/space-station-14/pull/7145

- removed all read-overrides and variants in favor of two new `Read`-methods (generic and non-generic) which return `T`/`object?`
- `ITypeReader` now returns `T` too
- removes populate, you can now pass in a value into `Read` if you wish for that to be populated (optional ofc)
- moves inheritances to `Datanode`-level. custom inheritancelogic is definable using `ITypeInheritanceHandler`
- starting with this pr, id, parent & abstract-datafields for `IPrototype` & `IInheritingPrototype` need to use the attributes `IdDatafield`, `ParentDatafield` & `AbstractDatafield`. there exists a check & error incase someone forgets this ofc
- the game will now throw for required fields. this is temporarily the only way to detect this error until i make yamllinter push inheritance before validating
- abstract has been split up into abstract & noSpawn:
	- abstract now means you can never actually adress a prototype in code.
	- to get entityprototypes to not show up in the spawnpanel but still be able to indexed, use noSpawn


metalgearsloth (approved):
image

PR #2606, opened 2022-03-17
More transform ECS methods

metalgearsloth:


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -29,6 +29,45 @@ public Matrix3 GetWorldMatrix(EntityUid uid, EntityQuery<TransformComponent> xfo
...
+    {
+        return GetWorldPosition(xformQuery.GetComponent(uid));
+    }
+
+    [Pure]
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public Vector2 GetWorldPosition(TransformComponent component, EntityQuery<TransformComponent> xformQuery)
+    {
+        return component.WorldPosition;
+    }
```
metalgearsloth, line 69: Added these versions because at some stage they'll actually use the xformquery.

PR #2610, opened 2022-03-18
fix sequence node Except()

ElectroJr:
Currently `SequenceDataNode.Except()` is broken and will always return null. If you fix that, there is a separate issue where the lack of an equality override results in it not properly removing entries (the `HashSet.Contains()` check passes when it shouldn't).

This PR tries to fix both of these,  I'm not 100% sure what the all the ramifications of the data-node changes are so someone who knows about this stuff should definitely have a look, 

Fixing this bug also requires fixing content bugs (space-wizards/space-station-14/pull/7177):
- `EntityStorageComponent` uses Container.ID before the containers have been initialised (ID = null).
- systems that add fixtures don't check if the fixture already exists. To help with this, this PR also adds a TryCreateFixture() function

Finally, there is an issue where fixture IDs get automatically generated, which means that when fixtures get properly serialized the fixture definitions bloat map files. Instead of updating content to ensure that all prototypes specify fixture IDs, this just sorta dodges the problem by not serializing auto-generated Ids.


Robust.Shared/Serialization/Markdown/Sequence/SequenceDataNode.cs
```diff
@@ -114,7 +114,7 @@ public override int GetHashCode()
         {
             var set = new HashSet<DataNode>(node._nodes);
             var newList = new List<DataNode>();
-            foreach (var nodeNode in node._nodes)
+            foreach (var nodeNode in _nodes)
```
PaulRitter: no need to copy node._nodes here. remove the `set` variable and directly use it imo

PaulRitter (changes requested):
lgtm besides the nitpick

Robust.Shared/Physics/FixtureSystem.cs
```diff
@@ -93,6 +93,21 @@ private void OnInit(EntityUid uid, FixturesComponent component, ComponentInit ar
 
         #region Public
 
+        /// <summary>
+        ///     Attempts to add a new fixture. Will do nothing ia a fixture with the requested ID already exists.
```
metalgearsloth:
```suggestion
        ///     Attempts to add a new fixture. Will do nothing if a fixture with the requested ID already exists.
```

metalgearsloth (approved):

PR #2619, opened 2022-03-20
Fixes file dialog manager in macOS

vulppine:
Does this via adding the ability to call actions from Clyde to run in the main render/window thread. Thanks to @PJB3005  for helping me figure some of this out.


PJB3005:
So I fixed this myself before realizing you had a PR for this open, since I was doing some stuff like compiling Zstd on macOS anyways: c9ccd0b873933a4e84bf8d8e9ad031f65e7d8398

Whoops, my bad

vulppine:
that's fine! it's pretty much the same thing anyways since we discussed it beforehand, as long as it works

PR #2622, opened 2022-03-21
Cache Sprite AABBs

### LINKED ISSUE #2615: Sprites should cache their AABBs
Not their dynamictree one mind you, just the calculated one

Total render time (in debug, so a bunch of this probably gets inlined for release, but still some people develop on potatos)

image

snowsignal:
Fixes #2615.

Essentially, this adds a `Box2 Bounds` field to `SpriteComponent` that is automatically updated when sprite layers are added, modified or removed. This will take some pressure off the CPU by not having to re-calculate bounding boxes for all sprites every frame.

~~I also removed some related code that calculated sprite RSI directions, since that didn't seem to be affecting the actual render whatsoever.~~ It was used to calculate the bounds for 8-directional sprites. For those cases, the possible bounding boxes for these sprites are unioned into one larger bounding box.

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -601,7 +601,7 @@ namespace Robust.Client.GameObjects
         private void RebuildBounds()
         {
             _bounds = new Box2();
-            foreach (var layer in Layers)
+            foreach (var layer in Layers.Where(layer => layer.Visible))
             {
                 _bounds = _bounds.Union(layer.CalculateBoundingBox());
             }
@@ -1128,7 +1128,6 @@ namespace Robust.Client.GameObjects
             }
 
             Layers[layer].SetVisible(visible);
-            RebuildBounds();
         }
 
         public void LayerSetVisible(object layerKey, bool visible)
@@ -1872,6 +1871,7 @@ namespace Robust.Client.GameObjects
                 Visible = value;
 
                 _parent.QueueUpdateIsInert();
+                _parent.RebuildBounds();
             }
 
             public void SetRsi(RSI? rsi)
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -1974,7 +1974,7 @@ namespace Robust.Client.GameObjects
                 // we can take the quick path of just making a box the size of the texture.
                 if (_parent.NoRotation && _rotation != 0)
                 {
-                    return Box2.CenteredAround(Offset, textureSize);
+                    return Box2.CenteredAround(Offset, textureSize).Scale(_scale);
                 }
 
                 var rsiState = GetActualState();
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -815,6 +815,8 @@ namespace Robust.Client.GameObjects
                 default:
                     throw new NotImplementedException();
             }
+
+            RebuildBounds();
         }
 
         public void LayerSetSprite(object layerKey, SpriteSpecifier specifier)
@@ -840,6 +842,7 @@ namespace Robust.Client.GameObjects
 
             var theLayer = Layers[layer];
             theLayer.SetTexture(texture);
+            RebuildBounds();
         }
 
         public void LayerSetTexture(object layerKey, Texture texture)
@@ -905,6 +908,7 @@ namespace Robust.Client.GameObjects
 
             var theLayer = Layers[layer];
             theLayer.SetState(stateId);
+            RebuildBounds();
         }
 
         public void LayerSetState(object layerKey, RSI.StateId stateId)
@@ -952,6 +956,8 @@ namespace Robust.Client.GameObjects
                     theLayer.Texture = null;
                 }
             }
+
+            RebuildBounds();
         }
 
         public void LayerSetState(object layerKey, RSI.StateId stateId, RSI rsi)
@@ -1009,6 +1015,7 @@ namespace Robust.Client.GameObjects
 
             var theLayer = Layers[layer];
             theLayer.SetRsi(rsi);
+            RebuildBounds();
         }
 
         public void LayerSetRSI(object layerKey, RSI rsi)
@@ -1066,6 +1073,7 @@ namespace Robust.Client.GameObjects
 
             var theLayer = Layers[layer];
             theLayer.Scale = scale;
+            RebuildBounds();
         }
 
         public void LayerSetScale(object layerKey, Vector2 scale)
@@ -1092,6 +1100,7 @@ namespace Robust.Client.GameObjects
 
             var theLayer = Layers[layer];
             theLayer.Rotation = rotation;
+            RebuildBounds();
         }
 
         public void LayerSetRotation(object layerKey, Angle rotation)
@@ -1116,6 +1125,7 @@ namespace Robust.Client.GameObjects
             }
 
             Layers[layer].SetVisible(visible);
+            RebuildBounds();
         }
 
         public void LayerSetVisible(object layerKey, bool visible)
@@ -1141,6 +1151,8 @@ namespace Robust.Client.GameObjects
 
             var theLayer = Layers[layer];
             theLayer.Color = color;
+
+            RebuildBounds();
         }
 
         public void LayerSetColor(object layerKey, Color color)
@@ -1166,6 +1178,8 @@ namespace Robust.Client.GameObjects
 
             var theLayer = Layers[layer];
             theLayer.DirOffset = offset;
+
+            RebuildBounds();
         }
 
         public void LayerSetDirOffset(object layerKey, DirectionOffset offset)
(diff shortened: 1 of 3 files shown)


ElectroJr (commented):
> I also removed some related code that calculated sprite RSI directions, since that didn't seem to be affecting the actual render whatsoever.

It makes no difference for 1- or 4-directional square sprites, but for either 8-directional or rectangular sprites the directions/rotations do still matter (or for sprites that have layers with offsets). This is most noticeable for things that use a post shader (like the interaction outline) which use the spritebbs for drawing. W/O accounting for directions, the entity gets cut off  when you hover over it:

https://user-images.githubusercontent.com/60421075/159211411-f05a229c-0420-4100-a1d9-7a410a7eff67.mp4

I think in general you might just need to have a different cached bound for all 8 directions. Or just a single cached bound that is made up of the union of all 8 of those. Though then rendering with shaders would be slightly less efficient because its using a larger area to draw than it needs to? No idea how significant that would be. Though AFAIK most sprites don't usually have a post shader anyways.

Caching the union of the 8 directional BBs rather than each one individually might also cause issues for y-sorting, because that currently just uses the bottom of the BB to sort entities.

ElectroJr (commented):
The bounds should also be updated whenever a layer has its visibility, scale, rotation, texture etc updated.

snowsignal:
Thanks for mentioning that @ElectroJr, I didn't include 8-directional sprites in my testing. I like your idea of making a union of rotated bounding boxes for 8-direction sprites, and as you said I don't think it'll be much of a performance hit since we'd only do it on sprites that have 8 directions. In terms of y-sorting, shouldn't there be a way to manually put the BB origin at the bottom of the sprite after we make a union from the directional BBs?

ElectroJr (commented):
> In terms of y-sorting, shouldn't there be a way to manually put the BB origin at the bottom of the sprite after we make a union from the directional BBs?

I'm not sure what you mean, but I probably also wasn't very clear. My concern was that currently, the y sorting is based on the y-coordinate of the bottom of the BB, so the black line in this image:
y-sorting-1

but if you only cache the union of the different BBs, the BB will be bigger and the lowest y coordinate will be lower (the new black line in this mockup):
y-sorting-3

The result is that an entity might get drawn above another, even though it appears to be further-up on the screen and would normally be sorted drawn below the other.

Realistically, because the vast majority of sprites are square and 1- or 4-directional, and those that won't will probably be something like no-rot mob sprites, I don't think this is a severe problem. Really, Y-sorting probably just needs be be done differently somehow. Its already odd having the sorting suddenly change because you rotated a couple of degrees and the BB changed as the sprite direction updated.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1943,32 +1963,26 @@ public Vector2i PixelSize
...
                 {
                     var textureSize = PixelSize / EyeManager.PixelsPerMeter;
 
-                    // this switch block is basically an explicit version of the `rsiDirectionMatrix` in `GetLayerDrawMatrix()`.
-                    var box = dir switch
+                    var box = Box2.CenteredAround(Offset, textureSize);
+
+                    var rsiState = GetActualState();
+
+                    if (rsiState is {Directions: RSI.State.DirectionType.Dir8})
```
ElectroJr: if rsi.State.Size is not square, then you also need to do a union for 4-dir sprites. (or just directly use a square box with sides of length max(width,height)

ElectroJr: And for 8-dir I think you can just directly use a box with sides of length `max(width, height, (width+ height) / sqrt(2))`

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1943,32 +1963,26 @@ public Vector2i PixelSize
...
-                        // rotate 90 degrees:
-                        RSIDirection.East or RSIDirection.West => Box2.CenteredAround(Offset, (textureSize.Y, textureSize.X)),
-                        // rotated 45 degrees (any 45 degree rotated rectangle has a square bounding box with sides of length (x+y)/sqrt(2) )
-                        _ => Box2.CenteredAround(Offset, Vector2.One * (textureSize.X + textureSize.Y) / MathF.Sqrt(2))
-                    };
+                        box = box.Union(Box2.CenteredAround(Offset, (textureSize.Y, textureSize.X)).Union(Box2.CenteredAround(Offset, Vector2.One * (textureSize.X + textureSize.Y) / MathF.Sqrt(2))));
+                    }
 
                     return _scale == Vector2.One ? box : box.Scale(_scale);
                 }
```
ElectroJr: also for the code below this, you can't just pass dir=RSIDirection.South. Just like the code above (which was just a shortcut/simplification if rotation = 0), it needs to account for both non-square and 8 dir sprites.

I think the brute force approach would be to just call it with something like dir =  south, southeast, east and then union those results?

Or I guess just use dir=south, and then union with copies of that BB rotated 45 & 90 degrees (ie, basically what is should be done above). AFAICT that should always contain the union of the separate direction BBs? or maybe its even equivalent?

ElectroJr: Also looking back at `GetLayerDrawMatrix()`, the `rsiDirectionMatrix` should probably just be static/const matrix with a dir switch, which should optimize that a bit.

ElectroJr (commented):
Couple more things:
- `RebuildBounds()` also has to be called in `SpriteComponent.LayerSetData()`
- Invisible layers shouldn't contribute to the BB 
  - Also means RebuildBounds() has to be called when toggling layer visibility
- Dir4 & Dir8 layers can be treated the same as Dir1 layers if the parent sprite has `NoRotation == true`
  - Not strictly required I guess, but currently this means some BBs will just be bigger than they need to be. In the video below, it would mean that two of the no-rot sprites would be square boxes, instead the smaller rectangles that they have in the video.
  - Would also mean RebuildBounds() has to be called when `NoRotation` is set

Other than that, AFAICT it should be good to merge. For the record with the no-rot changes, the sprite BBs for a rectangular Dir8 sprite using a combination of transforms & no-rot options now look like this:

https://user-images.githubusercontent.com/60421075/159970787-dd684870-73ea-4560-bd22-ec138e051ebb.mp4

So most of the time unchanged, but for some scaled or rotated layers the BBs will be bigger than they need to be, which might mess with how y-sorting is done currently. But that should hopefully be a rare problem and worth it for not having to re calculate the BBs constantly.

ElectroJr (commented):
Also, the MS-paint test entities I've been using all this time aren't in ss14 content. But If they keep being useful for testing sprite prs, maybe I should just open a content PR to add them (or some variation of them) for testing/debugging use until someone actually adds 8-dir rectangular sprites to the game?

snowsignal:
> Also, the MS-paint test entities I've been using all this time aren't in ss14 content. But If they keep being useful for testing sprite prs, maybe I should just open a content PR to add them (or some variation of them) for testing/debugging use until someone actually adds 8-dir rectangular sprites to the game?

Please do :pray:

snowsignal:
> Couple more things:
> 
>     * `RebuildBounds()` also has to be called in `SpriteComponent.LayerSetData()`
> 
>     * Invisible layers shouldn't contribute to the BB
>       
>       * Also means RebuildBounds() has to be called when toggling layer visibility
> 
>     * Dir4 & Dir8 layers can be treated the same as Dir1 layers if the parent sprite has `NoRotation == true`
>       
>       * Not strictly required I guess, but currently this means some BBs will just be bigger than they need to be. In the video below, it would mean that two of the no-rot sprites would be square boxes, instead the smaller rectangles that they have in the video.
>       * Would also mean RebuildBounds() has to be called when `NoRotation` is set
> 
> 

Done! `RebuildBounds()` was already being called in `LayerSetVisibility()` but the other two issues you raised have been addressed.

ElectroJr:
The no-rotation case also needs to re-scale the BB if the layer has a scale

> RebuildBounds() was already being called in LayerSetVisibility() 

Ah I didn't check that function, but there is still a problem because I was looking at the layer's `SetVisible` function, which can be, and sometimes is, called separately from the sprite's `LayerSetVisibility()`. Given that `LayerSetVisibility()` actually calls `SetVisible` internally, I guess `RebuildBounds()` doesn't actually have to be called in `LayerSetVisibility()`. 

Also `RebuildBounds()` function doesn't actually have a `if (!layer.Visible) continue` yet

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1943,39 +1966,40 @@ public Vector2i PixelSize
...
-                // Other than some special cases for simple layers, this will basically just apply the matrix obtained
-                // via GetLayerDrawMatrix() to this layer's bounding box.
+                var textureSize = PixelSize / EyeManager.PixelsPerMeter;
+
+                // If the parent has locked rotation and we don't have any rotation,
+                // we can take the quick path of just making a box the size of the texture.
+                if (_parent.NoRotation && _rotation != 0)
+                {
+                    return Box2.CenteredAround(Offset, textureSize);
+                }
```
ElectroJr: .Scale(_scale)

snowsignal: Sorry, forgot to add that!

PR #2624, opened 2022-03-22
Struct enumerator for GetAllGrids

metalgearsloth:
The IEnumerable version allocates a loootttt via NetworkedMapManager

### DIFF
--- a/Robust.Shared/Map/MapManager.GridCollection.cs
+++ b/Robust.Shared/Map/MapManager.GridCollection.cs
@@ -116,9 +116,20 @@ internal partial class MapManager
         OnGridCreated?.Invoke(mapGrid.ParentMapId, mapGrid.Index);
     }
 
+    public GridEnumerator GetAllGridsEnumerator()
+    {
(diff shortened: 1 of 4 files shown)


Robust.Shared/Map/MapManager.GridCollection.cs
```diff
@@ -116,9 +116,20 @@ public void OnGridAllocated(MapGridComponent gridComponent, MapGrid mapGrid)
...
+    public GridEnumerator GetAllGridsEnumerator()
+    {
+        var query = EntityManager.GetEntityQuery<MapGridComponent>();
+        return new GridEnumerator(_grids.GetEnumerator(), query);
+    }
+
     public IEnumerable<IMapGrid> GetAllGrids()
     {
-        return EntityManager.EntityQuery<IMapGridComponent>(true).Select(c => c.Grid);
+        var compQuery = EntityManager.GetEntityQuery<MapGridComponent>();
```
metalgearsloth, line 127: The Select was pretty bad so I also nuked that from this version too.

PR #2632, opened 2022-03-24
Increase OpenAL source limit to 512, adjustable by CVar

### LINKED ISSUE #2388: CreateAudioSource throws exceptions when there are more than 256 sources
Once audio source count passes 256, it throws an exception `System.ArgumentException: An item with the same key has already been added. Key: 0`

This can be reproduced by spawning into a large map like Packedstation with PVS off.
The sounds are caused by lights turning on and/or entities being inserted into storage entities.

snowsignal:
Fixes https://github.com/space-wizards/RobustToolbox/issues/2388

When creating the OpenAL context, we now hint that we may create up to 512 audio sources at once (512 was chose because it was double the previous limit, which already took a lot of effort to hit). This also has the nice benefit of making audio sources faster to allocate and use.


moonheart08 (approved):

metalgearsloth:
Does the same thing happen at 512? If so probably better to make it a cvar and allow other games to override it.

PJB3005:
We don't ship our own OpenAL on macOS, instead using the system one. Does this work on that? I know macOS' has a much lower limit than OpenAL Soft.

moonheart08:
Just looked, MacOS is hardcapped at 256.

moonheart08:
so if this is made a CVar we should introduce some way to report back if it was actually set to that value or if it has hit some kind of cap. This would introduce a method of fingerprinting the OS, though.

snowsignal:
> so if this is made a CVar we should introduce some way to report back if it was actually set to that value or if it has hit some kind of cap. This would introduce a method of fingerprinting the OS, though.

Wouldn't OpenAL just fail if we try to set a number of sources larger than what it can support?

snowsignal:
@metalgearsloth @PJB3005 @moonheart08 I made the requested changes - source limits can be set through the `audio.sourceMaximum` CVar and we will log a warning if the limit was unable to be set.

PJB3005:
Does OpenAL soft and macOS just set a lower limit if you try to go too high, or does it fail to create the device in the first place? I can check tonight I guess if necessary.

snowsignal:
> Does OpenAL soft and macOS just set a lower limit if you try to go too high, or does it fail to create the device in the first place? I can check tonight I guess if necessary.

I'm not sure. Unfortunately, I don't have a suitable way to test this.

metalgearsloth:
@PJB3005

> I can check tonight I guess if necessary.

Sad...

PJB3005:
From testing: Attempts to query the context attributes all give fuckall for some reason, meaning we can't find out the real sources count. It doesn't actually crash or anything though so 🤷.

I still think this PR is a bad idea though because 256 is already a *ridiculous* amount of audio sources and the real fix is to fix our shitty audio engine to do some basic prioritization.

PR #2638, opened 2022-03-25
Map loader saving change

wrexbe:
I was exploring, and felt like changing this


metalgearsloth (commented):
Could also use the EntityQuery<T> structs too should help as well.

PR #2644, opened 2022-03-25
Add Fluent hot reloading

DrSmugleaf:
https://user-images.githubusercontent.com/10968691/160124038-ede19cb8-3520-44f0-add3-781a4c716bad.mp4

The sandboxing receipt:
image

### DIFF
--- a/Robust.Shared/ContentPack/IDirRoot.cs
+++ b/Robust.Shared/ContentPack/IDirRoot.cs
@@ -1,0 +1,6 @@
+﻿namespace Robust.Shared.ContentPack;
+
+public interface IDirRoot : IContentRoot
+{
+    string FullPath { get; }
+}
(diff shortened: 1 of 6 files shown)


Robust.Shared/ContentPack/IDirRoot.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Robust.Shared.ContentPack;
+
+public interface IDirRoot : IContentRoot
```
PJB3005, line 3: When I said "public as long as the class stays internal" that means that the property ITSELF is also internal. This won't fly for sandboxing.

PJB3005 (changes requested):
Can some of this code be shared with the prototype watcher & window focus system?

PR #2652, opened 2022-03-28
caching trees in pvs

### LINKED ISSUE #2636: Cache RobustTree across ticks
If the chunk hasn't updated that tick and clients still need it then it's safe to re-use the robusttree the next tick and save a decent chunk of work for PVS.

PaulRitter:
ur welcome sloth
fixes #2636

### DIFF
--- a/Robust.Server/GameStates/PVSCollection.cs
+++ b/Robust.Server/GameStates/PVSCollection.cs
@@ -98,6 +98,11 @@ public sealed class PVSCollection<TIndex> : IPVSCollection where TIndex : ICompa
     /// </summary>
     private HashSet<TIndex> _changedIndices = new();
 
+    /// <summary>
+    /// A set of all chunks changed last tick
+    /// </summary>
+    private HashSet<IChunkIndexLocation> _dirtyChunks = new();
+
     public PVSCollection()
     {
         IoCManager.InjectDependencies(this);
@@ -105,14 +110,14 @@ public sealed class PVSCollection<TIndex> : IPVSCollection where TIndex : ICompa
 
     public void Process()
     {
+        _dirtyChunks.Clear();
         _changedIndices.EnsureCapacity(_locationChangeBuffer.Count);
 
-        foreach (var (key, _) in _locationChangeBuffer)
+        foreach (var (key, loc) in _locationChangeBuffer)
         {
             _changedIndices.Add(key);
         }
 
-        var changedChunkLocations = new HashSet<IIndexLocation>();
         foreach (var (index, tick) in _removalBuffer)
         {
             //changes dont need to be computed if we are removing the index anyways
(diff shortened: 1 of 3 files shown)


Robust.Server/GameStates/PVSCollection.cs
```diff
@@ -177,6 +186,7 @@ private void AddIndexInternal(TIndex index, IIndexLocation location)
                 if(!_gridChunkContents[gridChunkLocation.GridId].ContainsKey(gridChunkLocation.ChunkIndices))
                     _gridChunkContents[gridChunkLocation.GridId][gridChunkLocation.ChunkIndices] = new();
                 _gridChunkContents[gridChunkLocation.GridId][gridChunkLocation.ChunkIndices].Add(index);
```
metalgearsloth, line 188: Should probably be a trygetvalue here. Does C# re-use these or does it re-index it every time?

Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -411,8 +411,48 @@ private void OnMapCreated(MapChangedEvent e)
         return (_chunkList, playerChunks, viewerEntities);
     }
 
+    private Dictionary<(uint visMask, IChunkIndexLocation location), (Dictionary<EntityUid, MetaDataComponent> metadata,
+        RobustTree<EntityUid> tree)?> _previousTrees = new();
+
+    private HashSet<(uint visMask, IChunkIndexLocation location)> _reusedTrees = new();
```
metalgearsloth, line 417: This might show up on a profile as I'm not sure how C# handles valuetuple hashcodes but I guess we'll see in future.

Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -411,8 +411,53 @@ private void OnMapCreated(MapChangedEvent e)
...
+
     public (Dictionary<EntityUid, MetaDataComponent> mData, RobustTree<EntityUid> tree)? CalculateChunk(IChunkIndexLocation chunkLocation, uint visMask, EntityQuery<TransformComponent> transform, EntityQuery<MetaDataComponent> metadata)
     {
+        if (!_entityPvsCollection.IsDirty(chunkLocation) && _previousTrees.TryGetValue((visMask, chunkLocation), out var previousTree))
+        {
+            lock (_reusedTrees)
+            {
+                _reusedTrees.Add((visMask, chunkLocation));
+            }
+            return previousTree;
```
metalgearsloth: yucky lock, i'll profile using the array instead and get back to you.

Robust.Server/GameStates/PVSCollection.cs
PaulRitter: it might get optimized away, but ye, ur right, should be a tryget

metalgearsloth (approved):

PR #2654, opened 2022-03-29
Fix physics parenting bug

metalgearsloth:
Thanks @ElectroJr


Robust.Shared/Physics/Dynamics/SharedPhysicsMapComponent.cs
```diff
@@ -121,6 +121,23 @@ public Vector2 Gravity
...
+            // Look at my note under ProcessWakeQueue
+            if (body.Awake && body.BodyType != BodyType.Static)
+            {
+                _queuedWake.Remove(body);
+                AwakeBodies.Add(body);
+            }
+
+            Bodies.Add(body);
+            body.PhysicsMap = this;
+        }
```
metalgearsloth, line 139: I just moved this because why the fuck did I ever keep it separate.

PR #2662, opened 2022-03-31
Don't store contained entities on entitylookup

metalgearsloth:
The movement updates on SS14 are killer.

If any other games using Robust really needs it we can add a preprocessor or cvar or something.

API cleaned up a lot too.

Will need content PR.


Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -20,16 +16,35 @@ namespace Robust.Shared.GameObjects
...
+        /// </summary>
         IncludeAnchored = 1 << 1,
+
+        /// <summary>
+        /// Include entities that are currently in containers.
+        /// </summary>
+        Contained = 1 << 2,
         // IncludeGrids = 1 << 2,
+        // IncludePhysics (whenever it gets split off)
+        // Include maps
```
ElectroJr, line 35: When would a lookup return maps? or is this for something like the table-top / chess entities that have maps?

Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -189,12 +206,14 @@ private void OnTerminate(ref EntityTerminatingEvent args)
...
+                _mapManager.IsGrid(uid) ||
+                _container.IsEntityInContainer(uid)) return;
+
             var xformQuery = EntityManager.GetEntityQuery<TransformComponent>();
 
             if (!xformQuery.TryGetComponent(uid, out var xform) ||
-                xform.Anchored ||
-                _mapManager.IsMap(uid) ||
-                _mapManager.IsGrid(uid)) return;
+                xform.Anchored) return;
```
ElectroJr: Given that relatively veeery few entities will be grids or maps, wouldn't it be better to check this after checking if an entity is anchored? that should make initialising anchored entities ever so slightly faster.

Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -248,12 +267,17 @@ private bool CanMoveUpdate(EntityUid uid)
...
-                EntityManager.GetComponent<MetaDataComponent>(args.Entity).EntityLifeStage < EntityLifeStage.Initialized) return;
+                _mapManager.IsGrid(args.Entity)) return;
 
-            EntityLookupComponent? oldLookup = null;
-            var xformQuery = EntityManager.GetEntityQuery<TransformComponent>();
+            var meta = MetaData(args.Entity);
+
+            // Parent change gets raised after container insert so we'll just drop it and let OnContainerInsert handle.
+            if (meta.EntityLifeStage < EntityLifeStage.Initialized ||
+                _container.IsEntityInContainer(args.Entity, meta)) return;
```
ElectroJr: I guess same goes for this. Parent change due to initialization or insertion should be much much more common than a map or grid changing parents, so that should probably be checked before wasting time looking for grid/map components.

Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -248,12 +267,17 @@ private bool CanMoveUpdate(EntityUid uid)
         private void OnParentChange(ref EntParentChangedMessage args)
         {
             if (_mapManager.IsMap(args.Entity) ||
-                _mapManager.IsGrid(args.Entity) ||
-                EntityManager.GetComponent<MetaDataComponent>(args.Entity).EntityLifeStage < EntityLifeStage.Initialized) return;
+                _mapManager.IsGrid(args.Entity)) return;
 
-            EntityLookupComponent? oldLookup = null;
-            var xformQuery = EntityManager.GetEntityQuery<TransformComponent>();
+            var meta = MetaData(args.Entity);
```
ElectroJr, line 270: For the current order of operations, unless this is a map or grid, this system will always get the metadata comp & check the in-container flag. Physics will also do that for almost all objects moving across grids. 

IMO the event should probably just have a `is-now-in-container` bool, so that the meta data & flags only need to be checked once. And passing basic data about the new parent relationship in the parent-changed event makes sense.

Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -305,34 +348,20 @@ private void OnParentChange(ref EntParentChangedMessage args)
...
-                        // TODO: If we have 0 position and not contained can optimise these further, but future problem.
-                        var childAABB = GetAABBNoContainer(child.Value, coordinates.Position, childXform.WorldRotation - lookupXform.WorldRotation);
-                        AddToEntityTree(lookup, childXform, childAABB, xformQuery);
-                    }
+                    if (conManager.ContainsEntity(child.Value)) continue;
+
+                    var coordinates = _transform.GetMoverCoordinates(xform.Coordinates, xformQuery);
+                    var childXform = xformQuery.GetComponent(child.Value);
+                    // TODO: If we have 0 position and not contained can optimise these further, but future problem.
+                    var childAABB = GetAABBNoContainer(child.Value, coordinates.Position, childXform.WorldRotation - lookupXform.WorldRotation);
```
ElectroJr: lookupXform.WorldRotation should probably be cached outside of the while loop

ElectroJr: Same for GetMoverCoordinates(). AFAIK it should be independent of the child and can just be obtained outside of the while loop. Also applied to the while loop below this for the case where the entity does not have a `ContainerManagerComponent`.

Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
metalgearsloth: They're still entities so I think it would be prudent to include.

Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
metalgearsloth: It does save the entityquery if the entity is contained (i.e. mobs have potentially dozens involved) so would need to investigate further.

Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
metalgearsloth: Realistically I think the parent change should get suppressed entirely during container operations (same with moveevent) or alternatively remove ent change messages and have the flag which probably needs more bikeshedding.

PR #2663, opened 2022-03-31
Physicsmap tweaks

metalgearsloth:
Should be slightly more stable and at least log things a bit more. Awake / cancollide event ordering needs a bit more cleanup still but this is a step in the right direction.


Robust.Shared/Physics/Dynamics/SharedPhysicsMapComponent.cs
```diff
@@ -130,7 +126,6 @@ public void AddBody(PhysicsComponent body)
             // Look at my note under ProcessWakeQueue
```
ElectroJr, line 126: Should probably remove this comment now that ProcessWakeQueue is being disappeared

ElectroJr: Should probably remove this comment now that ProcessWakeQueue is being disappeared

ElectroJr (approved):

PR #2664, opened 2022-03-31
Fix PVS crash

metalgearsloth:
@PaulRitter

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -114,6 +114,7 @@ internal sealed partial class PVSSystem : EntitySystem
 
         _playerManager.PlayerStatusChanged += OnPlayerStatusChanged;
         SubscribeLocalEvent<MoveEvent>(OnEntityMove);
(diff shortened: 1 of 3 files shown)


Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -123,6 +124,21 @@ public override void Initialize()
...
+        // If parent changes then the RobustTree for that chunk will probably no longer be valid and need to force it as dirty.
+        // Should still be at its old location as moveevent is called after.
+        var xform = Transform(ev.Entity);
+
+        // As it's still at the old location this means it may be somewhere invalid such as nullspace.
+        if (xform.MapID == MapId.Nullspace) return;
+
+        // TODO: Could check the chunk of old vs new parent and see if we need to dirty at all.
+        var index = _entityPvsCollection.GetIndex(xform.Coordinates);
+        _entityPvsCollection.RegisterUpdate(ev.Entity, index);
```
metalgearsloth, line 140: I don't think we need to dirty the parent's chunk as sending a parent without its children should be fine, it's just vice versa is the problem?

PR #2666, opened 2022-03-31
Cleans up and improves MIDI code significantly.

gradientvera:
Required by https://github.com/space-wizards/space-station-14/pull/7365

- `MidiEvent` has been renamed to `RobustMidiEvent` to avoid ambiguity with NFluidsynth's `MidiEvent`.
- `RobustMidiEvent` has been cleaned up significantly. We pack the MIDI data as compact as we can. It now weighs 7 bytes, down from 16 bytes.
- `MidiRenderer` doesn't get injected IoC dependencies anymore, they're passed through the constructor.
- `IMidiManager` now allows you to create non-mono MIDI renderers.
- Added `RobustMidiCommand` helper enum for readability in switch statements.
- Fixed a bug where sometimes MIDI playback would never stop even after reaching the end of the MIDI track.
- Fixed a bug where pitchbend events had incorrect values, resulting in some audio distortion.
- Fixed a bug where old sequencer events would be ignored.
- ...And just a lot of cleanup, in general.


Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -538,70 +541,65 @@ public void SendMidiEvent(Shared.Audio.Midi.MidiEvent midiEvent)
...
-                        case 5:
-                        // MetaEvent -- SetTempo - 0x51
-                        case 81:
+                        case 0x00:
+                        case (RobustMidiCommand) 0x01:
+                        case (RobustMidiCommand) 0x05:
+                        // MetaEvent -- SetTempo
+                        case (RobustMidiCommand) 0x50:
                             // Already handled by the player.
                             return;
```
DubiousDoggo: No idea what the hell this is trying to do here but it doesn't make any sense. If any of these cases ever trigger something has gone very wrong.

Robust.Shared/Audio/Midi/RobustMidiEvent.cs
```diff
@@ -0,0 +1,89 @@
...
+        #region Properties
+
+        public int Channel => Status & LowNibbleMask;
+        public int Command => Status & HighNibbleMask;
+        public RobustMidiCommand MidiCommand => (RobustMidiCommand) Command;
+        public byte Key => Data1;
+        public byte Velocity => Data2;
+        public byte Control => Data1;
+        public byte Value => Data2;
+        public int Pitch => ((Data2 << 8) + Data1);
```
DubiousDoggo:
```suggestion
        public int Pitch => ((Data2 << 8) & Data1);
```

Robust.Shared/Audio/Midi/RobustMidiEvent.cs
```diff
@@ -0,0 +1,89 @@
...
+namespace Robust.Shared.Audio.Midi
+{
+    /// <summary>
+    ///     This class is a lightweight data representation of a Midi Event.
+    /// </summary>
+    [Serializable, NetSerializable]
+    public readonly struct RobustMidiEvent
+    {
+        private const byte LowNibbleMask = 0x0F;
+        private const byte HighNibbleMask = 0xF0;
```
DubiousDoggo: I'm all for consts but this is a bit silly

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Not at all. In my experience testing many, many MIDI tracks, these are triggered *all the time* and spam the console for no good reason. Resolving, I'm not deleting these.

Robust.Shared/Audio/Midi/RobustMidiEvent.cs
gradientvera: It's not, it makes the code more readable for people coming in with no clue what's going on.

Robust.Shared/Audio/Midi/RobustMidiEvent.cs
gradientvera: ...That's incorrect. In any case, it'd be a bitwise OR, not a bitwise AND.
It doesn't matter much because the math ends up being the same and personally I think this looks more readable.

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: And for the record, these are not part of this PR. These have been there for a *long* time.

metalgearsloth (approved):
I have no idea what I'm doing.

Robust.Shared/Audio/Midi/RobustMidiEvent.cs
DubiousDoggo: ope, my bad

gradientvera (commented):
:heavy_check_mark: Looks good to me yes, yes!

PR #2667, opened 2022-03-31
ProcGen Terrain Grids

Acruid:
This allows content to mark a grid to generate chunks when viewed by a player.
The grid raises an ECS event that can be handled by content, where any sort of planet terrain generation can happen.


moonheart08:
Needs unload, but i will be extremely happy with this once in.

PR #2671, opened 2022-04-01
Fix PVS take 3

metalgearsloth:
Doesn't fix the crashing but was another bug I found.


ElectroJr (approved):
Don't know enough about PVS to really know what effect this change has, but apparently its safe and I trust sloth

PR #2672, opened 2022-04-01
Add error tolerance to physics map

ElectroJr:
This PR prevents PhysicsMapComponent from entering an infinite loop of errors when a deleted entity gets added to the list of awake entities.

It also adds some code to make sure that if ever a parent of an entity is detached to null before the child is, that the child will still properly remove itself from the physics map. Basically removes code that was relying on the old-parent's MapId to determine an entity's old-map ID. Even after these changes, a deleted entity will sometimes still appear in a client's awake bodies list. Still not sure what causes that, but at least it won't enter an error loop.


ElectroJr (commented):
Hasn't been properly reviewed, but I want proper error tolerance for the playtest, because explosions (mass entity deletion) tends to trigger this, so I'm just gonna merge.

PR #2678, opened 2022-04-03
prototype composition

PaulRitter:
requires https://github.com/space-wizards/space-station-14/pull/7403

still requires a check to look for circular composition


Robust.Server/Bql/BqlQuerySelector.Builtin.cs
```diff
@@ -268,7 +268,7 @@ public override IEnumerable<EntityUid> DoSelection(IEnumerable<EntityUid> input,
                 if ((metaData.EntityPrototype?.ID == name) ^ isInverted)
                     return true;
 
-                return (metaData.EntityPrototype?.Parent == name) ^ isInverted; // Damn, can't actually do recursive check here.
+                return metaData.EntityPrototype?.Parents?.Any(x => x == name) == true ^ isInverted; // Damn, can't actually do recursive check here.
```
PaulRitter, line 271: @moonheart08 unsure if this is correct

Robust.Shared/Localization/LocalizationManager.Entity.cs
```diff
@@ -125,10 +126,10 @@ private EntityLocData CalcEntityLoc(string prototypeId)
                     }
                 }
 
-                if (prototype?.Parent == null)
+                if (prototype?.Parents == null)
                     break;
 
-                prototypeId = prototype.Parent;
+                prototypeId = prototype.Parents.First();
```
PaulRitter, line 129: @RemieRichards unsure if this is correct or how to handle this

Robust.Server/Bql/BqlQuerySelector.Builtin.cs
moonheart08: This is correct.

metalgearsloth:
Well that master merge went well

PR #2687, opened 2022-04-05
Local event for BUI opening

mirrorcult:
there was one for closing but not opening

### DIFF
--- a/Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
+++ b/Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
@@ -183,8 +183,8 @@ namespace Robust.Server.GameObjects
                 return false;
             }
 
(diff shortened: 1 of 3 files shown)


Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
```diff
@@ -184,6 +184,7 @@ public bool Open(IPlayerSession session)
             }
 
             _subscribedSessions.Add(session);
+            IoCManager.Resolve<IEntityManager>().EventBus.RaiseLocalEvent(Owner.Owner, new BoundUIOpenedEvent(UiKey, Owner.Owner, session));
             SendMessage(new OpenBoundInterfaceMessage(), session);
```
ElectroJr, line 188: Seeing as the close-event is also called in `ClientUserInterfaceComponent`, and as both events are defined in shared, the ClientUserInterfaceComponent.OpenInterface() function should probably also raise a open-event

ElectroJr: I am illiterate

ElectroJr (approved):

PR #2691, opened 2022-04-05
Don't allow velocity to be set where an obj can't collide

metalgearsloth:
Mover code was doing this client-side. Really CanCollide should be renamed to Enabled like it is on box2d I think.


PaulRitter:
> Really CanCollide should be renamed to Enabled like it is on box2d I think.

do it then no balls

PaulRitter (approved):

PR #2693, opened 2022-04-05
testmerge

PaulRitter:
paging @PJB3005 for a vibecheck on publishing the client under the custom version, as well as how to remove the version afterwards.
the idea is that the workflow can get triggered manually & will provide a git commit hash to change the submodule to temporarily.
also im p sure i need to push the removal of the branch in the remove-testmerge workflow somehow, gotta look that up

### DIFF
--- a/.github/workflows/remove_testmerge.yml
+++ b/.github/workflows/remove_testmerge.yml
@@ -1,0 +1,22 @@
+﻿name: Remove Testmerge
+on:
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: 'PR Number'
+        required: true
+        type: number
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Delete branch
+        id: parse_version
+        shell: pwsh
+        run: git branch -d testmerge-${{ github.event.inputs.pr }}
+      - name: Remove the client build off of centcomm (todo)
--- a/.github/workflows/testmerge.yml
+++ b/.github/workflows/testmerge.yml
@@ -1,0 +1,56 @@
+﻿name: Testmerge PR
+on:
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: 'PR Number'
+        required: true
+        type: number
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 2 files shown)


metalgearsloth:
@PJB3005 is this a merge or a close.

.github/workflows/remove_testmerge.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Delete branch
+        id: parse_version
+        shell: pwsh
+        run: git branch -d testmerge-${{ github.event.inputs.pr }}
```
PJB3005, line 21: This is only gonna remove the testmerge branch off the build-runner's local repo, not the actual repo on GitHub.

.github/workflows/remove_testmerge.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Remove the client build off of centcomm (todo)
```
PJB3005, line 22: Not sure whether this is worth the trouble, we can just keep them up and have content not use them.

PJB3005 (changes requested):
You aren't publishing a Git tag in any of this, so there's no way for the content build scripts to correctly determine which engine version they need to use. 

Also, it isn't going to be possible to publish Robust Modules with this since the version resolution system won't be able to avoid using them. I feel like we should figure that out.

I'm also not a fan of relying on PR numbers explicitly. It may be nicer to allow any branch to be test merged and have an easy way to create test merge branches from PRs. This would allow you to merge multiple PRs into a single test merge.

.github/workflows/testmerge.yml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          submodules: true
+      - name:
+        id: parse_version
+        shell: pwsh
+        run: |
+          $rawver = git tag --sort=committerdate --list "v*" | Select-Object -Last 1
+          $ver = [regex]::Match($rawver, "v?(.+)").Groups[1].Value
+          echo ("::set-output name=version::{0}-testmerge${{ github.event.inputs.pr }}" -f $ver)
+      - name: Checkout PR and create new branch
```
PaulRitter, line 25: im never checking out the pr branch here + this should prob just override the branch if it already exists(?)

PR #2694, opened 2022-04-06
Diagonal tile support

metalgearsloth:
Theoretically it supports any form of arbitrary tile collision as long as the vertices can overlap and it can at least prune it (i.e. half / quarter tiles).

TODO:
- [x] Optimise
- [ ] Re-implement fixture enlargement
- [x] Test
- [x] Double-check TryFindGridAt works

Edit:
So some notes:
- I culled pendingtiles from placementmanager as it made placement annoying
- Tests should catch most of the stuff I found bugs with
- Content is going to need updating for anchoring but that's going to be a large piece of work like grid rotation was so doubtful everything is going to work overnight


Robust.Shared/Map/Tile.cs
```diff
@@ -120,7 +117,12 @@ public override int GetHashCode()
...
-public enum TileRenderFlag : byte
+public enum TileFlag : byte
 {
-
+    None = 0,
+    Full = 1 << 0,
+    BottomLeft = 1 << 1,
+    BottomRight = 1 << 2,
+    TopRight = 1 << 3,
+    TopLeft = 1 << 4,
```
moonheart08: Should these not be two bitfields? Would allow normal tiles to have directional states as bonus futureproofing. Should probably add getter/setters for these fields while at it.
```suggestion
    BottomLeft = 0 << 0, // 2-bit direction
    BottomRight = 1 << 0,
    TopRight = 2 << 0,
    TopLeft = 3 << 0,
    Full = 0 << 2, // 1 bit type.
    Diagonal = 1 << 2,
```

metalgearsloth:
Couldn't figure out a good way to do fixture enlargement so might chuck up an issue for it and come back to it, PR should be good for a review otherwise.

Robust.Shared/Map/Tile.cs
moonheart08: Re-opening this, because honestly I think full support for directional tiles is a better thing (if you really don't want to code it, i will, but i don't want to have to be the first person to figure out map format conversion)

moonheart08:
https://github.com/space-wizards/space-station-14/issues/5188
Please generalize rotation to be a thing for all tiles instead of making it so only diagonals can be rotated.
Just doing rotation with variants gets very tricky if the tile both has diagonals and rotation states.

PR #2696, opened 2022-04-07
Better deletion check

metalgearsloth:


Robust.Shared/Physics/Dynamics/SharedPhysicsMapComponent.cs
```diff
@@ -250,7 +250,7 @@ private void Solve(float frameTime, float dtRatio, float invDt, bool prediction)
             {
                 // TODO: When this gets ECSd add a helper and remove
 
-                if (seed.Deleted)
+                if (seed.Deleted || _entityManager.Deleted(seed.Owner))
```
ElectroJr, line 253: if the owner has been deleted, the component should also be. So something else would have gone wrong here, and this could probably use extra logging. E.g: what is the life stage of the component? Is it pre-add or something?

PR #2705, opened 2022-04-10
Sync current time

ElectroJr:
Currently the server and a client's `IGameTiming.CurTime` can differ. The current time is inferred from the current tick, tick-rate, and some reference point (a cached time & tick). Whenever the server cvar gets updated, the cached time & tick get updated, but otherwise it just starts at (1st tick = 0 seconds). This causes problems if a player joins a game after the tick-timing has been adjusted, because they just use that initial reference point, effectively assuming that the server has been running at that tick rate the whole time, and thus evaluating the wrong current-time. 

To reproduce the use-delay, do-after, lobby-timer and some other general client-side prediction bugs encountered during the stress test, just start a server with a client and let it run for a bit. Then raise the tick rate and connect a new client, which will then encounter those bugs.

This PR is a pretty straightforward fix that just includes the current time in the game state that the server sends out. I don't fully understand all the client-side code, with all the extrapolation / interpolation & other stuff so someone should definitely check that I haven't missed something obvious.

### DIFF
--- a/Robust.Client/GameStates/GameStateProcessor.cs
+++ b/Robust.Client/GameStates/GameStateProcessor.cs
@@ -123,7 +123,13 @@ namespace Robust.Client.GameStates
             }
 
             if (applyNextState && !curState!.Extrapolated)
+            {
                 LastProcessedRealState = curState.ToSequence;
+
+                // Keep the client and the server current-time calculation in sync 
+                if (curState.ToTime != null)
+                    _timing.CachedCurTimeInfo = (curState.ToTime.Value, curState.ToSequence);
+            }
 
             if (!_waitingForFull)
             {
@@ -204,6 +210,9 @@ namespace Robust.Client.GameStates
                     Logger.DebugS("net", $"Resync CurTick to: {_lastFullState.ToSequence}");
 
                 var curTick = _timing.CurTick = _lastFullState.ToSequence;
+
+                if (_lastFullState.ToTime != null)
+                    _timing.CachedCurTimeInfo = (_lastFullState.ToTime.Value, _lastFullState.ToSequence);
 
                 if (Interpolation)
                 {
--- a/Robust.Server/GameStates/ServerGameStateManager.cs
+++ b/Robust.Server/GameStates/ServerGameStateManager.cs
@@ -171,6 +171,7 @@ namespace Robust.Server.GameStates
             const int BatchSize = 2;
             var batches = (int) MathF.Ceiling((float) players.Length / BatchSize);
 
+            var curTime = _gameTiming.CurTime;
             Parallel.For(0, batches, i =>
             {
                 var start = i * BatchSize;
@@ -180,7 +181,7 @@ namespace Robust.Server.GameStates
                 {
                     try
                     {
-                        SendStateUpdate(j);
+                        SendStateUpdate(j, curTime);
                     }
                     catch (Exception e) // Catch EVERY exception
                     {
@@ -189,7 +190,7 @@ namespace Robust.Server.GameStates
                 }
             });
 
-            void SendStateUpdate(int sessionIndex)
+            void SendStateUpdate(int sessionIndex, TimeSpan curTime)
             {
                 var session = players[sessionIndex];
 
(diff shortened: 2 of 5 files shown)


PJB3005:
This definitely isn't the right way to fix this. The client should get the told the baseline offset at connection startup or something, but this shouldn't be sent along with every `GameState`.

ElectroJr:
Sending it with ever game state is probably unnecessary, but only sending it on first connection also doesn't 100% fix the problem. If the tickrate cvar is changed, it can sometimes result in the offset/reference time being incorrect. 

E.g., in this image the client gets the cvar message late and updates the tickrate at the wrong tick, leading to a different curtime. Though to get this you have to either have a very high packet loss rate or just get really unlucky when an admin changes the cvar.
tick
 
Sending on first connect & changing how cvars are applied should probably fix it. Other than connection or cvar changes, I don't see any other way that the times could drift apart. 

But it also makes me uncomfortable that there's no real way of checking that they haven't drifted without just periodically sending the time+tick information to clients. If this bug does somehow happen again, and the time difference isn't bad enough to completely break the do-afters & other cooldown UIs, then it would probably lead to bugs that are really hard to track down. Especially because most local testing isn't done with latency & lossy networking.

PJB3005:
ConVars should still come in in-order, so if the `OnValueChanged` hook for tick rate had a way to know the tick it was *supposed* to change at this should be fixable like that. I'll ask Acruid what his take on this is though.

Acruid:
> This causes problems if a player joins a game after the tick-timing has been adjusted, because they just use that initial reference point, effectively assuming that the server has been running at that tick rate the whole time, and thus evaluating the wrong current-time.

Every client needs the reference point info when they join so they can properly derive CurTime. If the reference point is not being sent to new clients after it has happened, that needs to be changed. Using the current tick, tick-rate, and the reference point, a client at any time should deterministically be able to calculate CurTime.

I have no idea how the reference point works with saving and loading, because if it is lost every serialized instance of CurTime would be wrong.

The reference point is a (tick, time) tuple. I doubt we can sanely reverse it once it overwrites the previous one, so it should not be applied before it's defined server tick is processed, like immediately when received or inside prediction. It should be applied on the tick defined inside of it, and if the client has received a previous reference the local sim CurTime should equal the time in the new reference. If the reference point is received late, the best you can do is immediately apply it and deal with any mispredicts.

Because the Tickrate cvar and the reference point are related, they must be processed on the same tick. At the moment I think the processing order of those two messages is undefined, so you may need to change the code so we can define an order. It would make sense for the server to be storing the reference point for new clients and broadcasting the reference point change from inside the OnValueChanged, guaranteeing this all happens in one tick.

There should be no problem passing the tick it was supposed to be applied to in OnValueChanged, so any timing-sensitive cvar can detect it is late and try and cope, but remember this is trying to cope with poor network conditions so there is only so much it can do.

Note that none of this is related to real/wall/network time. If the client and server cannot keep their gameloops in sync, and in turn keeping CurTick and LastServerTick synchronized causing late/really early packets, that is a completely different issue.

> Sending on first connect & changing how cvars are applied should probably fix it.

Yep. Without late messages, this whole process of keeping the CurTime synchronized should be completely feasible and deterministic.

moonheart08:
as of now significant tick drift can occur over short periods if the server lags (which is still a concern even if a minute one), so we should probably make sure the game resyncs every once and a while.

PJB3005:
> as of now significant tick drift can occur over short periods if the server lags (which is still a concern even if a minute one), so we should probably make sure the game resyncs every once and a while.

Well if that's bug 3, once again I'd rather fix it at the source

ElectroJr:
> There should be no problem passing the tick it was supposed to be applied to in OnValueChanged, so any timing-sensitive cvar can detect it is late and try and cope,

I have a branch that adds support for `OnValueChanged` & `UnsubValueChanged` & invoke to work with either `Action<T>` or `Action<T, GameTick>`. But its a bit of a mess, especially seeing as (afaik) tick-rate is currently the only tick-rate sensitive cvar. So instead: 

> It would make sense for the server to be storing the reference point for new clients and broadcasting the reference point change from inside the OnValueChanged, guaranteeing this all happens in one tick.

This is probably a better way of doing this, unless there is reason to expect other tick-sensitive cvars. And I guess the same message/event can be sent on client connection.

On that note, I have no idea where/how to handle on-connect data syncing, other than maybe adding something similar to the `SyncConnectingClient()` call for cvars in `PlayerManager.NewSession()`. It might be better if someone who is more familiar with cvars & networking handles fixing this.

PR #2711, opened 2022-04-10
Final grid movement optimisation

metalgearsloth:
Completely avoids the expensive broadphase updates for large maps and uses its own system.


metalgearsloth:
Drafted because I was concerned about *the bug* but turns out it's just lerping.

PR #2716, opened 2022-04-11
Fix map change crash

metalgearsloth:
Follower really out here exposing all the ordering bugs.

How to replicate:
- Follow entity
- Restart round

Doesn't fix the broadphase one I don't think.


Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -146,7 +146,7 @@ private void HandleMapChange(PhysicsComponent body, TransformComponent xform, Ma
             {
                 var oldMapEnt = MapManager.GetMapEntityId(oldMapId);
 
-                if (MetaData(oldMapEnt).EntityLifeStage < EntityLifeStage.Terminating)
+                if (TryComp<MetaDataComponent>(oldMapEnt, out var meta) && meta.EntityLifeStage < EntityLifeStage.Terminating)
```
gradientvera, line 149: Maybe use the `TryTerminating` proxy method here instead?

metalgearsloth: Tried using it but it has a resolve that logs if it fails which it probably should still do but in this case we know the old map may be deleted already.

PR #2720, opened 2022-04-11
Adds custom MIDI soundfont support

vulppine:
Adds custom MIDI soundfont support via a new CVar, `midi.soundfont`.

### DIFF
--- a/Robust.Client/Audio/Midi/IMidiManager.cs
+++ b/Robust.Client/Audio/Midi/IMidiManager.cs
@@ -48,7 +48,7 @@ public interface IMidiManager
     /// <summary>
     ///     Redirect for the MIDI library's soundfont checker
     /// </summary>
-    public bool IsSoundFont(string filename);
+    bool IsSoundFont(string filename);
 
     /// <summary>
     ///     Method called every frame.
--- a/Robust.Client/Audio/Midi/IMidiManager.cs
+++ b/Robust.Client/Audio/Midi/IMidiManager.cs
@@ -46,6 +46,11 @@ public interface IMidiManager
     RobustMidiEvent FromSequencerEvent(SequencerEvent midiEvent, uint tick);
 
     /// <summary>
+    ///     Redirect for the MIDI library's soundfont checker
+    /// </summary>
+    public bool IsSoundFont(string filename);
+
+    /// <summary>
     ///     Method called every frame.
     ///     Should be used to update positional audio.
     /// </summary>
--- a/Robust.Client/Audio/Midi/MidiManager.cs
+++ b/Robust.Client/Audio/Midi/MidiManager.cs
@@ -226,7 +226,7 @@ internal sealed partial class MidiManager : IMidiManager
             }
 
             var customFont = _cfgMan.GetCVar(CVars.MidiSoundfont);
-            if (customFont.Length > 0)
+            if (!string.IsNullOrEmpty(customFont))
             {
                 if (File.Exists(customFont) && SoundFont.IsSoundFont(customFont))
                     renderer.LoadSoundfont(customFont, true);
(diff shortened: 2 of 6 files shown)


Robust.Shared/CVars.cs
```diff
@@ -1034,6 +1034,9 @@ protected CVars()
         public static readonly CVarDef<float> MidiVolume =
             CVarDef.Create("midi.volume", 0f, CVar.CLIENTONLY | CVar.ARCHIVE);
 
+        public static readonly CVarDef<string> MidiSoundfont =
+            CVarDef.Create("midi.soundfont", "", CVar.CLIENTONLY | CVar.ARCHIVE);
```
gradientvera:
```suggestion
            CVarDef.Create("midi.soundfont", string.Empty, CVar.CLIENTONLY | CVar.ARCHIVE);
```

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -225,6 +225,13 @@ private void LoggerDelegate(NFluidsynth.Logger.LogLevel level, string message, I
                     renderer.LoadSoundfont(WindowsSoundfont, true);
             }
 
+            var customFont = _cfgMan.GetCVar(CVars.MidiSoundfont);
+            if (customFont.Length > 0)
```
gradientvera:
```suggestion
            if (!string.IsNullOrEmpty(customFont))
```

Robust.Client/Audio/Midi/IMidiManager.cs
```diff
@@ -45,6 +45,11 @@ public interface IMidiManager
     /// </summary>
     RobustMidiEvent FromSequencerEvent(SequencerEvent midiEvent, uint tick);
 
+    /// <summary>
+    ///     Redirect for the MIDI library's soundfont checker
+    /// </summary>
+    public bool IsSoundFont(string filename);
```
gradientvera: `public` is not really needed here
```suggestion
    bool IsSoundFont(string filename);
```

gradientvera (commented):
This won't load the custom soundfont on existing renderers, meaning this will only be applied after a game restart/the renderer gets recreated.

vulppine:
> This won't load the custom soundfont on existing renderers, meaning this will only be applied after a game restart/the renderer gets recreated.

yeah, this is why i asked earlier - is this OK behavior, or should this be accounted for upon a soundfont change?

gradientvera:
> > This won't load the custom soundfont on existing renderers, meaning this will only be applied after a game restart/the renderer gets recreated.
> 
> yeah, this is why i asked earlier - is this OK behavior, or should this be accounted for upon a soundfont change?

I'd prefer that to be accounted for, but for a first PR this is good enough. Up to you whether you want to code that now or not! 

Keep in mind that if you do that, you should also handle unloading the old custom soundfont for those cases. Fluidsynth's API has functions for that, but I'm not entirely sure if nfluidsynth exposes them-

vulppine:
> > > This won't load the custom soundfont on existing renderers, meaning this will only be applied after a game restart/the renderer gets recreated.
> > 
> > 
> > yeah, this is why i asked earlier - is this OK behavior, or should this be accounted for upon a soundfont change?
> 
> I'd prefer that to be accounted for, but for a first PR this is good enough. Up to you whether you want to code that now or not!
> 
> Keep in mind that if you do that, you should also handle unloading the old custom soundfont for those cases. Fluidsynth's API has functions for that, but I'm not entirely sure if nfluidsynth exposes them-

so this does exist, but the synth API didn't return the ID of the loaded soundfont: i've fixed that in atsushieno/nfluidsynth#15, so that way we can get the ID of the soundfonts loaded in

vulppine:
pending nfluidsynth making it available in its package (unless it already is), i think this is OK if you also consider it OK to merge - the SS14 PR merge isn't necessarily required since this is a cvar, and it could be configured through the config file (at least, iirc)

gradientvera (approved):

PR #2721, opened 2022-04-11
overflowqueue

PaulRitter:
i wrote this bc i thought i needed it for pvs, then realized i actually dont need it. oops


PJB3005:
What advantage does this have over `System.Collections.Generic.Queue<T>`? `Queue<T>` is also implemented as a circular queue underneath I think.

moonheart08:
> What advantage does this have over `System.Collections.Generic.Queue<T>`? `Queue<T>` is also implemented as a circular queue underneath I think.

just checked, it is in fast a circular queue internally. I don't see any benefit here?

PaulRitter:
i might be missing smth but afaict, the generic queue automatically grows its internal array and doesnt have a circular functionality like i provide here. the integer param being exposed by its ctor is to allocate the initial capacity, not to set the limit.
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.queue-1.enqueue?view=net-6.0#system-collections-generic-queue-1-enqueue(-0)

example: https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYAMBYAUBgRj1UwAJUCAWAbjzwDcBDMUgR1IF5SoBTAd1IBFCDxEAeAJZRgAPgAUBAJS1cAMwD2YOVOCkJnUpmr6xpAkb2JEigN55SDtgDoAolFYiR25XgC+xAgBOOVYnABEeD1EeOUVlIA
this should output 9 if the queue were to actually be circular

PJB3005:
Automatic growing behavior has nothing to do with being circular or not. .NET's grows if the buffer is completely full, yours silently discards unhandled entries. Both act the same if the buffer is not saturated.

PaulRitter:
>  .NET's grows if the buffer is completely full, yours silently discards unhandled entries.

yes, thats the behaviour i intended with this datastructure. its a fixed-size queue that discards the oldest entry if a new entry is enqueued when the queue is full. Queue<T> has no such behaviour afaict

PJB3005:
In that case I would recommend renaming this to something like `OverflowQueue` and be very clear about the behavior in a doc comment.

PaulRitter:
sgtm, way better name than circularqueue. also ye, needs xmldoc

Robust.Shared/Collections/OverflowQueue.cs
```diff
@@ -0,0 +1,149 @@
...
+/// <typeparam name="T"></typeparam>
+public sealed class OverflowQueue<T>
+{
+    private readonly T[] _queue;
+    private int _currentIdx = 0;
+    private int _length = 0;
+    /// <summary>
+    /// The size of the queue-buffer.
+    /// </summary>
+    public readonly int Size;
```
PJB3005: Generally most of these types of collections do `Size => _queue.Length`, since it always has to match the backing array. It would save 8 bytes per object instance.

PJB3005 (approved):

Robust.Shared/Collections/OverflowQueue.cs
PaulRitter:
```suggestion
    public int Size => _queue.Length;
```

Robust.Shared/Collections/OverflowQueue.cs
```diff
@@ -0,0 +1,149 @@
...
+    /// <summary>
+    /// The size of the queue-buffer.
+    /// </summary>
+    public int Size => _queue.Length;
+
+    /// <param name="size">size of the queue-buffer</param>
+    public OverflowQueue(int size)
+    {
+        _queue = new T[size];
+        Size = size;
```
PaulRitter:
```suggestion
```

PR #2727, opened 2022-04-13
Make MappingDataNode.Except() not recursive

ElectroJr:
Currently there is an issue with map saving. `MapLoader` uses the the `MappingDataNode.Except()` function to remove data that is already present in the entity prototype in order to reduce map sizes. However because this function recursively removes data, and because of how map loading works, this can lead to errors when saving & loading entities.

E.g. if you have some entity prototype with a component that has as some dictionary datafields:
```yaml
- type: MyComponent
  data:
    A: 1
  unchangedData: 
    B: 2
``` 
If some system adds an entry to the `data` dictionary, the dictionary SHOULD get saved to the map as something like:
```yaml
- type: MyComponent
  data:
    A: 1
    C: 3
``` 
But because of the recursive `Except()` it instead just saves it as:
```yaml
- type: MyComponent
  data:
    C: 2 
``` 
As the map loading stuff uses the `data` dictionary defined in the map to overwrite prototype-defined dictionary, if you then load that map you get a missing `A: 1` entry.

Because map saving is currently really only done pre-map init, this would only cause problems for systems that add or remove data from a non-empty dictionary/list during component add/init/startup. The only place I know of where this currently causes issues is for docking airlocks, where the docking system adds a fixture on init, resulting in the airlock getting saved with only that fixture, and not the default door fixture. This just wasn't noticed before because prior to #2610 those sequence mappings just weren't getting saved in the first place.

So this PR updates `Except()` to not be recursive while adding a separate `RecursiveExcept()` that keeps the old behaviour. AFAIK `Except()` was only really used in `MapLoader`, so AFAIK this shouldn't cause issues? But again I don't really know all the implications of messing with all this data node stuff.

This PR also fixes the equality override added in #2610, which just wasn't sufficient to ensure equality for sequences and mappings.

### DIFF
--- a/Robust.Server/Maps/MapLoader.cs
+++ b/Robust.Server/Maps/MapLoader.cs
@@ -866,15 +866,16 @@ namespace Robust.Server.Maps
 
                     var md = metaQuery.GetComponent(entityUid);
 
+                    Dictionary<string, MappingDataNode>? cache = null;
                     if (md.EntityPrototype is {} prototype)
                     {
                         mapping.Add("type", prototype.ID);
-                        if (!prototypeCompCache.ContainsKey(prototype.ID))
+                        if (!prototypeCompCache.TryGetValue(prototype.ID, out cache))
                         {
-                            prototypeCompCache[prototype.ID] = new Dictionary<string, MappingDataNode>();
+                            prototypeCompCache[prototype.ID] = cache =  new Dictionary<string, MappingDataNode>();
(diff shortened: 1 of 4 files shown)


PaulRitter (approved):
i will change this back as soon as i code yaml-annotations to override the inheritancebehaviour for nodes so i can properly annotate nodes/dicts in the map to merge with the default value

mirrorcult (approved):

PR #2733, opened 2022-04-14
make pvs account for lastack properly

PaulRitter:


Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -627,43 +623,41 @@ public void ReturnToPool(HashSet<int>[] playerChunks)
...
+                if (!EntityManager.EntityExists(entityUid))
+                    continue;
 
-            entityStates.Add(new EntityState(entityUid, new NetListAsArray<ComponentChange>(new []
-            {
-                ComponentChange.Changed(_stateManager.TransformNetId, new TransformComponent.TransformComponentState(Vector2.Zero, Angle.Zero, EntityUid.Invalid, false, false)),
-            }), true));
+                entityStates.Add(new EntityState(entityUid, new NetListAsArray<ComponentChange>(new []
+                {
+                    ComponentChange.Changed(_stateManager.TransformNetId, new TransformComponent.TransformComponentState(Vector2.Zero, Angle.Zero, EntityUid.Invalid, false, false)),
```
metalgearsloth: Just thinking should we make this state a static or something.

Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -29,6 +30,7 @@ internal sealed partial class PVSSystem : EntitySystem
     [Shared.IoC.Dependency] private readonly INetConfigurationManager _netConfigManager = default!;
 
     public const float ChunkSize = 8;
+    public const int TickBuffer = 10;
```
metalgearsloth, line 33: So technically at my highest ping was 350ms to vera's rpi which is technically higher than this so not sure if it matters.

Robust.Shared/Collections/OverflowDictionary.cs
```diff
@@ -0,0 +1,99 @@
...
+
+namespace Robust.Shared.Collections;
+
+public sealed class OverflowDictionary<TKey, TValue> : IDictionary<TKey, TValue>, IDisposable where TKey : notnull
+{
+    private TKey[] _insertionQueue;
+    private int _currentIndex = 0;
+    private IDictionary<TKey, TValue> _dict;
+    public int Capacity => _insertionQueue.Length;
+    private Action<TValue>? _valueDisposer;
```
metalgearsloth: comments my man

Robust.Shared/Collections/OverflowDictionary.cs
```diff
@@ -0,0 +1,99 @@
...
+        _valueDisposer = valueDisposer;
+        _dict = new Dictionary<TKey, TValue>(capacity);
+        _insertionQueue = new TKey[capacity];
+    }
+
+    private int GetArrayStartIndex()
+    {
+        var idx = _currentIndex - Count;
+        if (idx < 0)
+            idx += Capacity;
```
metalgearsloth: Could you do _currentIndex % Capacity?

metalgearsloth (approved):
my brain can't process overflowdictionary rn so my review was kinda shit

Robust.Shared/Collections/OverflowDictionary.cs
PaulRitter: yes :P

Robust.Shared/Collections/OverflowDictionary.cs
PaulRitter: never

Robust.Server/GameStates/PVSSystem.cs
PaulRitter: lemme try to calculate.
assuming a roundtrip for an ack takes Xms (your ping), and assuming you have Y ticks per second, you'd have one state produced each 1000/Y ms, meaning while we wait for the ack, X/(1000/Y) states get generated and sent, which would be our requirement for the tickbuffer afaict. for a ping of 350ms with 30tps, that'd sum up to 10.5.

this means, the question would be: what amount of latency should this account for on our default tps, 30. (if we lower tps, the tolerated latency goes even higher so thats not an issue. for example ,at 20tps and 350ms, you'd need a tickbuffer of 7)
do note that pvs still works even if the tickbuffer is exceeded, it'll just send the state as if you just entered your view.
tldr: we need to pin down what amount of *consistent* latency we want to account for. spikes/max dont matter as much.

Robust.Server/GameStates/PVSSystem.cs
PaulRitter: ye, prob true

PaulRitter:
> my brain can't process overflowdictionary rn so my review was kinda shit

dw about it, it has unittests so it must be fine

PR #2737, opened 2022-04-15
ZSTD game states + other improvements

PJB3005:


Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -60,6 +65,43 @@ public void Initialize()
...
+            for (var i = 0; i < _threadResourcesPool.Length; i++)
+            {
+                ref var res = ref _threadResourcesPool[i];
+                res.CompressionContext = new ZStdCompressionContext();
+            }
+
+            UpdateZStdParams();
+        }
+
+        private void UpdateZStdParams()
```
PaulRitter, line 91:
```suggestion
        private void UpdateZStdParams(int compressionLevel)
```

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -60,6 +65,43 @@ public void Initialize()
...
+        {
+            var compressionLevel = _cfg.GetCVar(CVars.NetPVSCompressLevel);
```
PaulRitter, line 93:
```suggestion
```

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -60,6 +65,43 @@ public void Initialize()
...
+            }
+
+            _threadResourcesPool = new PvsThreadResources[_parallelMgr.ParallelProcessCount];
+            for (var i = 0; i < _threadResourcesPool.Length; i++)
+            {
+                ref var res = ref _threadResourcesPool[i];
+                res.CompressionContext = new ZStdCompressionContext();
+            }
+
+            UpdateZStdParams();
```
PaulRitter, line 88:
```suggestion
            UpdateZStdParams(_cfg.GetCVar(CVars.NetPVSCompressLevel));
```

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -60,6 +65,43 @@ public void Initialize()
             _networkManager.Disconnect += HandleClientDisconnect;
 
             _pvs = EntitySystem.Get<PVSSystem>();
+
+            _parallelMgr.AddAndInvokeParallelCountChanged(ParallelChanged);
+
+            _cfg.OnValueChanged(CVars.NetPVSCompressLevel, _ => UpdateZStdParams(), true);
```
PaulRitter, line 71:
```suggestion
            _cfg.OnValueChanged(CVars.NetPVSCompressLevel, UpdateZStdParams, true);
```

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -168,33 +211,27 @@ public void SendGameStateUpdate()
...
-            Parallel.For(0, batches, i =>
-            {
-                var start = i * BatchSize;
-                var end = Math.Min(start + BatchSize, players.Length);
-
-                for (var j = start; j < end; ++j)
+            _parallelMgr.ParallelForWithResources(
+                0, players.Length,
+                _threadResourcesPool,
+                (int i, ref PvsThreadResources resource) =>
```
PaulRitter, line 217: this will no longer batch players. intended?

PaulRitter (changes requested):

Robust.Server/GameStates/ServerGameStateManager.cs
PJB3005: `Parallel.For` basically already does a very similar internal for loop to do the range anyways, so I figured it wasn't worth the extra noise.

Robust.Server/GameStates/ServerGameStateManager.cs
PJB3005: I did all the `UpdateZStdParams` like this so that if we need to add more zstd parameters in the future (idk for what) they assignment logic is self contained. This seems more much organized and self-contained IMO.

PaulRitter: fair enough, that makes sense

Robust.Server/GameStates/ServerGameStateManager.cs
PaulRitter: looking at the source, appears it does. interesting. lgtm

PaulRitter (approved):

PR #2742, opened 2022-04-17
ECS even more transform

metalgearsloth:
Had more in this PR but was getting a weird map serialization failure. Needs content PR


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -10,6 +10,142 @@ namespace Robust.Shared.GameObjects;
...
+
+        component.GridID = component.GetGridIndex(xformQuery);
+        component.RebuildMatrices();
+    }
+
+    private void OnCompStartup(EntityUid uid, TransformComponent component, ComponentStartup args)
+    {
+        // Re-Anchor the entity if needed.
+        if (component.Anchored)
+            component.Anchored = true;
```
ElectroJr: Was this code originally just there to run the `if LifeStage == ComponentLifeStage.Starting` clause in the `Anchored` setter function? Is there some reason that this logic can't just be moved into the startup function?

ElectroJr (approved):

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
metalgearsloth: Probably, the anchoring property is icky right now and should be cleaned up.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -434,4 +570,5 @@ public MapId GetMapId(EntityUid? uid, TransformComponent? xform = null)
 
         return xform.MapID;
     }
+
```
ShadowCommander:
```suggestion
```

ShadowCommander (approved):

PR #2744, opened 2022-04-18
MapLoader and mapping command changes

ElectroJr:
- Adds a `storeUids` option to the loadmap command
- Makes the order of arguments for the `loadmap` and `loadbp` commands the same.
- Fixes some args.Length checks for the `loadmap` command.
- Slightly improves map deserialization in `MapLoader.cs`
  - Somebody who understands map saving/loading better than me should check I haven't done something stupid.

### DIFF
--- a/Robust.Server/Maps/MapLoader.cs
+++ b/Robust.Server/Maps/MapLoader.cs
@@ -683,6 +683,8 @@ namespace Robust.Server.Maps
                         type = typeNode.Value;
                     }
 
+                    // TODO Fix this. If the entities are ever defined out of order, and if one of them does not have a
+                    // "uid" node, then defaulting to Entities.Count will error.
                     var uid = Entities.Count;
                     if (entityDef.TryGet<ValueDataNode>("uid", out var uidNode))
                     {
--- a/Robust.Server/Maps/MapLoader.cs
+++ b/Robust.Server/Maps/MapLoader.cs
@@ -513,10 +513,12 @@ namespace Robust.Server.Maps
                 // get ents that the grids will bind to
                 var gridComps = new Dictionary<GridId, MapGridComponent>(_readGridIndices.Count);
 
+                var gridQuery = _serverEntityManager.GetEntityQuery<MapGridComponent>();
+
                 // linear search for new grid comps
                 foreach (var tuple in _entitiesToDeserialize)
                 {
-                    if (!_serverEntityManager.TryGetComponent(tuple.Item1, out MapGridComponent gridComp))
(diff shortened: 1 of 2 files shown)


Robust.Server/Maps/MapLoader.cs
```diff
@@ -597,17 +599,14 @@ private void AttachMapEntities()
             private void FixMapEntities()
             {
                 var pvs = EntitySystem.Get<PVSSystem>();
-                foreach (var entity in Entities)
+                foreach (var grid in Grids)
                 {
-                    if (_serverEntityManager.TryGetComponent(entity, out IMapGridComponent? grid))
-                    {
```
ElectroJr, line 603: This same iteration over all entities was already done previously in `ReadGridSection()`, so this just re-uses those results. AFAIK this should cause no issues.

Robust.Server/Maps/MapLoader.cs
```diff
@@ -1084,14 +1083,14 @@ public MapLoadException(string? message)
 
                 var val = int.Parse(node.Value);
 
-                if (val >= Entities.Count || !UidEntityMap.ContainsKey(val) || !Entities.TryFirstOrNull(e => e == UidEntityMap[val], out var entity))
+                if (!UidEntityMap.TryGetValue(val, out var entity))
```
ElectroJr, line 1088: The `val >= Entities.Count` can't be used if the `MapSaveIdComponent` was used when writing entities (i.e., entity uids are not necessarily consecutive).

Also AFAICT if an entry exists in `UidEntityMap` there will always be a matching entry in `Entities`. I don't really understand why this LINQ stuff was being done here. If this is needed for some weird edge case this should probably be re-done to avoid doing that LINQ for every entity that needs to be read.

Robust.Server/Maps/MapLoader.cs
```diff
@@ -684,6 +683,8 @@ private void AllocEntities()
                         type = typeNode.Value;
                     }
 
+                    // TODO Fix this. If the entities are ever defined out of order, and if one of them does not have a
+                    // "uid" node, then defaulting to Entities.Count will error.
                     var uid = Entities.Count;
                     if (entityDef.TryGet<ValueDataNode>("uid", out var uidNode))
```
ElectroJr, line 689: Currently the "uid" field gets written when saving maps, but based on the loading code here it doesn't **have** to b and can just be inferred from the order. Though that can currently throw errors in some situations.

Should the uid field just be required? Or would that break things elsewhere?

metalgearsloth (approved):

ElectroJr:
Whenever this gets merged, space-wizards/space-station-14/pull/7625 should get un-reverted

PR #2745, opened 2022-04-18
benchmark script

PaulRitter:
move benchmarks to a script that gets executed on centcom.
the idea is that content/forks can also use this script to run their benchmarks.

needs @PJB3005 to:
- create key & secret
- create user (dont forget nice perms)
- adjust secrets so it will properly connect to the db


PJB3005:
What could go wrong.

PR #2747, opened 2022-04-19
Add PVSOverrideSystem

ElectroJr:
Add a barebones system that allows content to add pvs overrides.


PaulRitter (approved):
you can add another for a localoverride while ur at it if you want

ElectroJr:
> you can add another for a localoverride while ur at it if you want

Added that and a function to clear the override

PaulRitter (approved):
probably also interesting for @ike709 since you can use this to always send entities to a client

PR #2753, opened 2022-04-20
Remove ManagedHttpListener from csproj

ElectroJr:


PJB3005:
This appears to literally be the same fix as dec1495e1ec95251eda31111dbcb6057f7f9bf6b and the only thing that puzzles me is why GitHub doesn't mark this as conflicting. Closing.

PR #2769, opened 2022-04-24
Provide a generic way to expose tags on the status endpoint

20kdc:
This is so that there's a way to get things such as language information and content severity information to Launcher at some later date.

There is no use-case as of this moment, I don't have a Launcher PR using this.

Basically this gets the preparatory work over with so that if anyone wants to add such features to Launcher, they don't have to get an engine PR in place to actually write them.

### DIFF
--- a/Robust.Server/ServerStatus/StatusHost.Handlers.cs
+++ b/Robust.Server/ServerStatus/StatusHost.Handlers.cs
@@ -48,10 +48,10 @@ namespace Robust.Server.ServerStatus
             var tagsCache = _serverTagsCache;
             if (tagsCache != null)
             {
-                var tags = new JsonObject();
+                var tags = new JsonArray();
                 foreach (var tag in tagsCache)
                 {
-                    tags[tag] = true;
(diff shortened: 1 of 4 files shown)


ike709:
Shouldn't this be a dict so you could do things like `Region = US-W` or `Roleplay = Medium` and the launcher can easily filter them later?

20kdc:
> Shouldn't this be a dict so you could do things like `Region = US-W` or `Roleplay = Medium` and the launcher can easily filter them later?

See https://discord.com/channels/310555209753690112/310555209753690112/967858770300522536 for my attempted "is this a sane format" check of the format.

Something to keep in mind is that this tags system is, above all else, meant for simple flagging/filtering usecases where "my RobustToolbox version is too old" shouldn't be a valid way to dodge hub hosting rules.
It's not really meant as an *absolute* general metadata solution - at some point or another that's just re-inventing JSON.

If maintainers in general are sure it should be a key/value dictionary, I'll make it key/value, but I think it'll complicate things.

PR #2771, opened 2022-04-25
Make the server/client libraries

wrexbe:
Was playing with robust pong, and noticed two exe files in my build output. Changing the robust client, and server to be a dll seems to have a very small file size improvement, and has a slight usability improvement if you want to run it from the folder. I'm not sure if they are exes because of something from the past, or there is some reason it needs to be an exe.


gradientvera:
I mean, launching the Robust server/client and having it load content *is* something we support...

PaulRitter:
what vera said, this is intended

PR #2780, opened 2022-05-01
Unpause the map when initialized

ShadowCommander:
When mapping commands were run in this order, the spawned player would be unable to move:
```
mapping 2 maps/whatever.yml
respawn
mapinit 2
tp 0 0 2
```
This was due to `mapping` setting MapComponent.MapPaused to true, but `mapinit` did not set MapPaused back to false. This PR rectifies the situation.


ShadowCommander:
Closing in favor of #2781

ElectroJr:
Ah my bad, I didn't check recent Prs when opening mine and didn't see you'd already made one to fix the map-init

PR #2781, opened 2022-05-01
Speed up map init, and unpause map on init

ElectroJr:
Currently the `mapinit` command un-pauses all entities, but not the map itself. This updates `MapManager.DoMapInitialize()` so that it also un-pauses the map.

This PR also removes the `RunMapInit` EntityUid extension, in favour of adding a method to EntityManager. This also fixes a problem where `MapManager.DoMapInitialize()` would do an IEntityManager resolve for every entity on the map.

It also changes some init/startup function so that they either use a metadata query or accept a metadata component as an input.

### DIFF
--- a/Robust.Server/Maps/MapLoader.cs
+++ b/Robust.Server/Maps/MapLoader.cs
@@ -128,13 +128,20 @@ namespace Robust.Server.Maps
             }
             else if (_mapManager.IsMapInitialized(mapId))
             {
-
                 foreach (var entity in context.Entities)
                 {
                     var meta = query.GetComponent(entity);
                     _serverEntityManager.RunMapInit(entity, meta);
                     if (isPaused)
                         meta.EntityPaused = true;
+                }
+            }
+            else if (isPaused)
+            {
+                foreach (var entity in context.Entities)
+                {
+                    var meta = query.GetComponent(entity);
+                    meta.EntityPaused = true;
                 }
             }
         }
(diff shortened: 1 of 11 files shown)


Robust.Shared/Map/MapManager.Pause.cs
```diff
@@ -54,16 +54,18 @@ public void DoMapInitialize(MapId mapId)
             if (IsMapInitialized(mapId))
                 throw new ArgumentException("That map is already initialized.");
 
-            ClearMapPreInit(mapId);
-
```
ShadowCommander, line 58: Remove the ClearMapPreInit method if there's no other references.

ShadowCommander (changes requested):

ElectroJr:
I'm gonna be busy for the next few days and I probably wont have time to update this and fix whichever test failed, so in the meantime feel free either update or replace this PR

Robust.Server/Maps/MapLoader.cs
```diff
@@ -115,26 +115,25 @@ private ResourcePath Rooted(string path)
...
-            {
                 foreach (var entity in context.Entities)
                 {
-                    _serverEntityManager.GetComponent<MetaDataComponent>(entity).EntityPaused = true;
+                    var meta = query.GetComponent(entity);
+                    _serverEntityManager.RunMapInit(entity, meta);
+                    if (isPaused)
+                        meta.EntityPaused = true;
                 }
             }
```
metalgearsloth: Entities aren't being marked as paused for a non-mapinitialized map so saltern runs for a few ticks in the failed test.

PR #2790, opened 2022-05-04
caches reflection in serv3 datadefinition delegates

PaulRitter:
cuts down on a few percent of entityspawning


Robust.Shared/Serialization/Manager/Definition/DataDefinition.Emitters.cs
```diff
@@ -6,15 +6,19 @@
...
 using Robust.Shared.Utility;
 
 namespace Robust.Shared.Serialization.Manager.Definition
 {
     public partial class DataDefinition
     {
-        private PopulateDelegateSignature EmitPopulateDelegate()
+        private PopulateDelegateSignature EmitPopulateDelegate(IDependencyCollection collection)
         {
+            var isServer = !collection.Resolve<INetManager>().IsServer;
```
wrexbe: This is backwards?

PR #2804, opened 2022-05-09
RobustToolbox side of Redial API.

20kdc:
Please be aware that as far as I know testing this code is nigh-impossible until it ships.
RedialApi itself was already tested, though I suppose the full stack hasn't been tested yet.
We'll have to see, won't we?


20kdc:
Just altered this to add a bit more paranoia because I don't trust people not to do weird stuff to bypass the one redial rule

PR #2805, opened 2022-05-10
Add fluent function for indefinite article

mirrorcult:
Mostly stolen from here: https://stackoverflow.com/a/8044744

Works on either string or entity loc values. For entity just takes it from metadata name

Proof it works:

image

image

(these two are just tested, not actually implementing that)
image

image

### DIFF
--- a/Robust.Shared/Localization/LocalizationManager.Functions.cs
+++ b/Robust.Shared/Localization/LocalizationManager.Functions.cs
@@ -2,6 +2,8 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
+using System.Text.RegularExpressions;
 using Linguini.Bundle;
 using Linguini.Bundle.Types;
 using Linguini.Shared.Types.Bundle;
@@ -35,6 +37,7 @@ namespace Robust.Shared.Localization
             // Misc
             AddCtxFunction(bundle, "ATTRIB", args => FuncAttrib(bundle, args));
             AddCtxFunction(bundle, "CAPITALIZE", FuncCapitalize);
+            AddCtxFunction(bundle, "INDEFINITE", FuncIndefinite);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Robust.Shared/Localization/LocalizationManager.Functions.cs
```diff
@@ -56,6 +59,92 @@ private ILocValue FuncCapitalize(LocArgs args)
             else return new LocValueString("");
         }
 
+        private static string[] IndefExceptions = { "euler", "heir", "honest" };
+        private static char[] IndefCharList = { 'a', 'e', 'd', 'h', 'i', 'l', 'm', 'n', 'o', 'r', 's', 'x' };
+        private static string[] IndefRegexes = { "^e[uw]", "^onc?e\b", "^uni([^nmd]|mo)", "^u[bcfhjkqrst][aeiou]" };
```
PJB3005: These regexes should be pre-parsed into `Regex` instances.

Robust.Shared/Localization/LocalizationManager.Functions.cs
```diff
@@ -56,6 +59,92 @@ private ILocValue FuncCapitalize(LocArgs args)
...
+        private static char[] IndefVowels = { 'a', 'e', 'i', 'o', 'u' };
```
PJB3005: Make these all `static readonly`.

Robust.Shared/Localization/LocalizationManager.Functions.cs
```diff
@@ -56,6 +59,92 @@ private ILocValue FuncCapitalize(LocArgs args)
...
+            if (wordi.StartsWith("hour") && !wordi.StartsWith("houri"))
+                return an;
+
+            if (wordi.Length == 1)
+            {
+                return wordi.IndexOfAny(IndefCharList) == 0 ? an : a;
+            }
+
+            if (Regex.Match(word,
+                    "(?!FJO|[HLMNS]Y.|RY[EO]|SQU|(F[LR]?|[HL]|MN?|N|RH?|S[CHKLMNPTVW]?|X(YL)?)[AEIOU])[FHLMNRSX][A-Z]")
```
PJB3005: Pre-cache regex here.

Robust.Shared/Localization/LocalizationManager.Functions.cs
```diff
@@ -56,6 +59,92 @@ private ILocValue FuncCapitalize(LocArgs args)
...
+                return an;
+            }
+
+            foreach (var regex in IndefRegexes)
+            {
+                if (Regex.IsMatch(wordi, regex))
+                    return a;
+            }
+
+            if (Regex.IsMatch(word, "^U[NK][AIEO]"))
```
PJB3005: Pre-cache regex here.

Robust.Shared/Localization/LocalizationManager.Functions.cs
```diff
@@ -56,6 +59,92 @@ private ILocValue FuncCapitalize(LocArgs args)
...
+            {
+                return wordi.IndexOfAny(IndefCharList) == 0 ? an : a;
+            }
+
+            if (wordi.IndexOfAny(IndefVowels) == 0)
+            {
+                return an;
+            }
+
+            return Regex.IsMatch(wordi, "^y(b[lor]|cl[ea]|fere|gg|p[ios]|rou|tt)") ? an : a;
```
PJB3005: Pre-cache regex here.

PJB3005 (commented):
So one concern I have about this is that this is a very English-centric approach. Now obviously you'd have to change this code for other languages, but there are *many* languages where this type of guesswork isn't possible. I'm just wondering if it would be better to implement this in a more fundamental part of the localization system, somehow.

mirrorcult:
> So one concern I have about this is that this is a very English-centric approach. Now obviously you'd have to change this code for other languages, but there are _many_ languages where this type of guesswork isn't possible. I'm just wondering if it would be better to implement this in a more fundamental part of the localization system, somehow.

we already have many fluent functions that are english-centric that are widely used, the solution remie had was to just have locale-specific functions, which is fine, but not particularly needed (in this PR) right now

PJB3005:
That's not what I mean. The problem is that I'm concerned we should maybe have a *thorough* way of passing through the indefinite article (or any other such data) in a general purpose manner, instead of having an English-specific workaround to guess it. Then we could declare whether something is indefinite or not in the localization files and it would pass through.

This would probably be necessary for many languages. For example in Dutch the equivalent of "the" is either "het" or "de" depending on the word, and there's no rhyme or reason to it. All interpolated nouns would basically need to be annotated *somehow*.

PJB3005:
Another instance where this would be necessary is noun genders in French.

mirrorcult:
im gonna be real i have no idea how that would be accomplished in any way. for entities you could just use an attribute on the grammar component (like this has an override for) but how are you at all supposed to pass that data in with just, strings? especially if those are just strings defined in YAML?

the easiest way to do what you want with dutch IMO is to literally just have a big list somewhere of what nouns correspond to what article (if it is that random, idk dutch) and then have the function just lookup that

wrexbe:
Not a cool programming solution, but it's possible to reword sentences so it doesn't matter
(p)assistant will be selected if your preference is unavailable.
There is no way to deconstruct the (not) APC

PJB3005:
That, too, is a hack that probably isn't gonna work long-term. Try using English without "the".

wrexbe:
What about putting all these functions behind some kind of ILanguageAnnotationHandler, and make a DefaultEnglishHandler. Some of these annotations won't even make sense depending on the language.

PJB3005:
Current status has been "let downstreams help us out figuring it out". If need be I could do some work to add more exemplary Dutch support as a test case though.

PR #2818, opened 2022-05-14
Localized string serializer 

### LINKED ISSUE #2787: Custom type serializer for localized strings
So right now we do this for localized strings in prototypes and components:
```csharp
[DataField("name", required: true)]
private string Name { get; } = default!;

[ViewVariables(VVAccess.ReadOnly)]
public string LocalizedName => Loc.GetString(Name);
```

That could be replaced by:
```csharp
[DataField("name", customTypeSerializer: typeof(LocalizedStringSerializer))]
public string Name = default!;
```
or for list:
```csharp
[DataField("name", customTypeSerializer: typeof(LocalizedStringListSerializer))]
public List<string> Names = new();
```

We can even run ingegration test, to see if content has all localized strings in english .ftl files.

Macoron:
Fix #2787

I'm not familiar with magical world of robust serialization, so draft for now.

### DIFF
--- a/Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringListSerializer.cs
+++ b/Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringListSerializer.cs
@@ -1,0 +1,71 @@
+﻿using System.Collections.Generic;
+using Robust.Shared.IoC;
+using Robust.Shared.Serialization.Manager;
+using Robust.Shared.Serialization.Markdown;
+using Robust.Shared.Serialization.Markdown.Sequence;
+using Robust.Shared.Serialization.Markdown.Validation;
+using Robust.Shared.Serialization.Markdown.Value;
+using Robust.Shared.Serialization.TypeSerializers.Interfaces;
+
+namespace Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Localization;
+
+public sealed class LocalizedStringListSerializer : ITypeSerializer<List<string>, SequenceDataNode>
+{
+    private LocalizedStringSerializer LocalizedSerializer => new();
+
+    public ValidationNode Validate(ISerializationManager serializationManager, SequenceDataNode node,
+        IDependencyCollection dependencies, ISerializationContext? context = null)
+    {
+        var list = new List<ValidationNode>();
+
+        foreach (var dataNode in node.Sequence)
+        {
+            if (dataNode is not ValueDataNode value)
+            {
+                list.Add(new ErrorNode(dataNode, $"Cannot cast node {dataNode} to ValueDataNode."));
+                continue;
+            }
+
+            list.Add(LocalizedSerializer.Validate(serializationManager, value, dependencies, context));
+        }
+
+        return new ValidatedSequenceNode(list);
+    }
+
+    public DataNode Write(ISerializationManager serializationManager, List<string> value, bool alwaysWrite = false,
+        ISerializationContext? context = null)
+    {
+        var list = new List<DataNode>();
+
+        foreach (var str in value)
+        {
+            list.Add(LocalizedSerializer.Write(serializationManager, str, alwaysWrite, context));
+        }
+
+        return new SequenceDataNode(list);
+    }
+
+    public List<string> Read(ISerializationManager serializationManager, SequenceDataNode node, IDependencyCollection dependencies,
+        bool skipHook, ISerializationContext? context = null, List<string>? value = default)
+    {
+        value ??= new List<string>();
+
+        foreach (var dataNode in node.Sequence)
+        {
+            value.Add(LocalizedSerializer.Read(
+                serializationManager,
+                (ValueDataNode) dataNode,
+                dependencies,
+                skipHook,
+                context));
+        }
+
+        return value;
+    }
+
+    public List<string> Copy(ISerializationManager serializationManager, List<string> source, List<string> target, bool skipHook,
+        ISerializationContext? context = null)
+    {
+        return new(source);
+    }
+}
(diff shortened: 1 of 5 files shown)


Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringSerializer.cs
```diff
@@ -0,0 +1,40 @@
...
+using Robust.Shared.Serialization.TypeSerializers.Interfaces;
+
+namespace Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Localization;
+
+public sealed class LocalizedStringSerializer : ITypeSerializer<string, ValueDataNode>
+{
+    public ValidationNode Validate(ISerializationManager serializationManager, ValueDataNode node,
+        IDependencyCollection dependencies, ISerializationContext? context = null)
+    {
+         return dependencies.Resolve<ILocalizationManager>().HasString(node.Value)
```
Macoron, line 16: When does `Validate` function called? In some tests?

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringSerializer.cs
```diff
@@ -0,0 +1,40 @@
...
+             ? new ValidatedValueNode(node)
+             : new ErrorNode(node, $"Can't find localization string for {node.Value}");
+
+    }
+
+    public string Read(ISerializationManager serializationManager, ValueDataNode node, IDependencyCollection dependencies,
+        bool skipHook, ISerializationContext? context = null, string? value = default)
+    {
+        var isValid = dependencies.Resolve<ILocalizationManager>().TryGetString(node.Value, out var res);
+        return isValid ? res! : throw new InvalidMappingException($"Localization string {node.Value} not found");
```
Macoron, line 26: Is it ok to throw exception in `Read` function if loc string not found?

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringSerializer.cs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public DataNode Write(ISerializationManager serializationManager, string value, bool alwaysWrite = false,
+        ISerializationContext? context = null)
+    {
+        return new ValueDataNode(value);
```
Macoron, line 32: This one will write localized string in node. We can't revert localization, so reading this node again will result an error. I'm not sure that it's desired behavior.

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringSerializer.cs
```diff
@@ -0,0 +1,40 @@
...
+         return dependencies.Resolve<ILocalizationManager>().HasString(node.Value)
+             ? new ValidatedValueNode(node)
+             : new ErrorNode(node, $"Can't find localization string for {node.Value}");
+
+    }
+
+    public string Read(ISerializationManager serializationManager, ValueDataNode node, IDependencyCollection dependencies,
+        bool skipHook, ISerializationContext? context = null, string? value = default)
+    {
+        var isValid = dependencies.Resolve<ILocalizationManager>().TryGetString(node.Value, out var res);
```
Macoron, line 25: It won't support localization hot reloading. Is that ok?

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringListSerializer.cs
```diff
@@ -0,0 +1,71 @@
...
+using Robust.Shared.Serialization.Manager;
+using Robust.Shared.Serialization.Markdown;
+using Robust.Shared.Serialization.Markdown.Sequence;
+using Robust.Shared.Serialization.Markdown.Validation;
+using Robust.Shared.Serialization.Markdown.Value;
+using Robust.Shared.Serialization.TypeSerializers.Interfaces;
+
+namespace Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Localization;
+
+public sealed class LocalizedStringListSerializer : ITypeSerializer<List<string>, SequenceDataNode>
```
Macoron, line 12: Is that intended way to implement collections for custom serializes?

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringSerializer.cs
ElectroJr: Pretty sure it gets called by the yaml linter. Not sure if it also gets called elsewhere.

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringSerializer.cs
ElectroJr: So I don't know much about serialization, I've just poked around and edited existing code, but throwing errors during read when failing to parse the input seems to be the standard, so it should be fine.

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringSerializer.cs
ElectroJr: If the point is just to use the serializer for yaml linting / validation (e.g., like the prototype id string serializers), you could just return the non-localized key

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Localization/LocalizedStringSerializer.cs
ElectroJr: it should definitely read and write equivalent values. So it either needs to somehow be able to revert back to the key, or this just needs to read & write the key itself (i.e., have the serializer just be for validation)

ElectroJr: After actually reading the linked issue, I think to achieve this you'd have to replace the List<string> with some kind of List<LocalizedString> where LocalizedString has both the key, and can implicitly convert to a normal string

Acruid:
Is this still alive? Undraft it if you are ready for a review.

Macoron:
I will probably closed it for now, because I really don't sure what is the best way to implement it. Maybe will return to that latter.

PR #2826, opened 2022-05-15
Significantly optimise prediction resetting

metalgearsloth:
Turns out just tracking dirty entities and not iterating 30k is much faster.

This is gonna be very sussy and I'm not sure the best way to add a test to make sure functionality doesn't change.


Robust.Client/GameStates/PVSSystem.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            yield return ent;
+        }
+
+        ents.Clear();
+        _dirtyPool.Return(ents);
+    }
+
+    private void OnEntityDirty(object? sender, EntityUid e)
+    {
```
ElectroJr: probably should check if `IsClientSide` here and just not add those entities in the first place

metalgearsloth:
Ready for review I guess; I'm still a bit sussy on it as I'm not familiar with clientgamestatemanager's flow but none of my asserts ever tripped even with fakelag...

Robust.Client/GameStates/PVSSystem.cs
```diff
@@ -0,0 +1,108 @@
...
+        foreach (var (tick, dirty) in _dirtyEntities)
+        {
+            if (tick < currentTick) continue;
+            foreach (var ent in dirty)
+            {
+                // DebugTools.Assert(!ent.IsClientSide());
+
+                if (!metaQuery.TryGetComponent(ent, out var meta) ||
+                    meta.EntityLastModifiedTick < currentTick) continue;
+
```
ElectroJr: Assuming everything it working properly, the `if (...) continue` should always fail, right?
So why not just make it a debug assert and speed up release slightly.

also the comment can probably just be removed.

ElectroJr: And if its about checkingwhether the entity has been deleted, the comment in the ClientGameStateManager around the `_processor.TryGetLastServerStates()` implies that it would handle that situation

Robust.Client/GameStates/PVSSystem.cs
```diff
@@ -0,0 +1,108 @@
...
+    internal void Reset(GameTick curTick)
+    {
+        var toRemove = new RemQueue<GameTick>();
+
+        foreach (var (tick, _) in _dirtyEntities)
+        {
+            if (tick >= curTick - 1) continue;
+            toRemove.Add(tick);
+        }
```
ElectroJr: I think the logic here should be reversed? Currently it's only ever removing a single tick and leaving behind entities that were predicted to get dirtied in future ticks. What it should be doing when resetting back to curTick is to remove any entries with a tick value equal to or greater than curTick.

Actually, when would partial resets actually be used? AFAIK, currently `ResetPredictedEntities()` only ever resets to the "real" server state, so this should just always clear all dirtied entities.

If partial-resetting is actually needed eventually, there probably needs to be some sort of check to make sure that very old (small tick value) entries don't somehow get left behind and fill up the dictionary,

Robust.Client/GameStates/PVSSystem.cs
metalgearsloth: It does get hit for reasons; I don't really know what happens with resetting entity modified ticks.

Robust.Client/GameStates/PVSSystem.cs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    internal void Reset()
+    {
+        foreach (var (tick, sets) in _dirtyEntities.ToArray())
+        {
+            sets.Clear();
+            _dirtyPool.Return(sets);
+            _dirtyEntities.Remove(tick);
+        }
```
ElectroJr: maybe just 
```suggestion
        foreach (var (tick, sets) in _dirtyEntities)
        {
            sets.Clear();
            _dirtyPool.Return(sets);
        }
        _dirtyEntities.Clear();
```

ElectroJr (approved):
It seeeems to work fine? But maybe best to get someone else to also look at it.

Robust.Client/GameStates/PVSSystem.cs
```diff
@@ -0,0 +1,100 @@
...
+using Robust.Shared.IoC;
+using Robust.Shared.Timing;
+using Robust.Shared.Utility;
+
+namespace Robust.Client.GameStates;
+
+internal sealed class PVSSystem : SharedPVSSystem
+{
+    // Why is tracking dirty entities for prediction on PVS, you ask?
+    // Because the server does something identical in PVS sooooo
```
PJB3005: That doesn't sound like a very good reason to me. We're not doing PVS here so don't call it that.

PJB3005 (changes requested):

Robust.Shared/GameStates/SharedPVSSystem.cs
```diff
@@ -0,0 +1,8 @@
+using Robust.Shared.GameObjects;
+
+namespace Robust.Shared.GameStates;
+
+internal abstract class SharedPVSSystem : EntitySystem
```
PaulRitter: why does this exist? i see no reason for it

Robust.Client/GameStates/PVSSystem.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            if (comp.Owner.IsClientSide() || comp.EntityLastModifiedTick < currentTick)
+            {
+                continue;
+            }
+            DebugTools.Assert(ents.Contains(comp.Owner));
+        }
+#endif
+        */
+
```
PaulRitter: mucho comments

PaulRitter (changes requested):
nitpicks, otherwise lgtm

PR #2830, opened 2022-05-16
Stop double AnchorStateChangedEvent

ElectroJr:
Currently transform component state handling raises two `AnchorStateChangedEvent` when setting something to anchored.


metalgearsloth (approved):

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -393,7 +393,7 @@ internal void OnHandleState(EntityUid uid, TransformComponent component, ref Com
             {
                 var iGrid = Comp<MapGridComponent>(_mapManager.GetGridEuid(component.GridID));
                 AnchorEntity(component, iGrid);
-                component.SetAnchored(true);
+                DebugTools.Assert(component.Anchored);
```
ElectroJr, line 396: The `AnchorEntity()` call should itself call `component.SetAnchored(true)`, which raises the anchor changed event. If it ever fails to anchor the entity, something else probably needs to be fixed.

PR #2831, opened 2022-05-16
Structured disconnect messages

20kdc:
This allows using JSON for disconnect messages on an opt-in basis.
This is used for server shutdown messages and for server key mismatches.
(A "redial flag" is set on these two.)

### DIFF
--- a/Robust.Shared/Network/NetStructuredDisco.cs
+++ b/Robust.Shared/Network/NetStructuredDisco.cs
@@ -1,0 +1,106 @@
+﻿using System;
+using System.Net;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+
+namespace Robust.Shared.Network;
+
+/// <summary>
+/// Structured disconnection utilities.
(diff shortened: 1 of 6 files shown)


Robust.Shared/Network/NetStructuredDisco.cs
```diff
@@ -0,0 +1,109 @@
...
+namespace Robust.Shared.Network;
+
+/// <summary>
+/// Structured disconnection utilities.
+/// These use JsonNode so that Content may add it's own data.
+/// Note that to prevent encoding a NetStructuredDisco value within a NetStructuredDisco value,
+/// these should be encoded at the "highest level".
+/// Whatever generates the final "reason" value is responsible for performing NetStructuredDisco.Encode.
+/// </summary>
+public static class NetStructuredDisco
```
PJB3005: `Disco` is weird, just write the name out I'd say.

20kdc: Fixed in bdb37ce

PR #2832, opened 2022-05-16
Filtered Itemlist + Map & Gridlist controls

PaulRitter:
items can now be hidden on the itemlist to avoid recreating the list on each filter input
makes the filter + list combo a standalone control things can inherit
creates map & gridlists to use in admin/mapping tooling
needs content pr (https://github.com/space-wizards/space-station-14/pull/8210) bc of the playerlist


Robust.Client/UserInterface/Controls/FilteredItemList.xaml.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Robust.Client.AutoGenerated;
+using Robust.Client.UserInterface.XAML;
+
+namespace Robust.Client.UserInterface.Controls;
+
+[GenerateTypedNameReferences, Virtual]
+public partial class FilteredItemList : BoxContainer
```
PJB3005, line 7: This control shouldn't publically inherit from `BoxContainer` directly. It should be a child control.

PJB3005 (commented):
Is this intended for future mapping mode:tm: work? If so I'd prefer if these were internal or something.

We're keeping some facade of API stability now, leaving random classes with unclear use hanging around like this is sus.

PaulRitter:
> Is this intended for future mapping mode™️ work? If so I'd prefer if these were internal or something.
> 
> We're keeping some facade of API stability now, leaving random classes with unclear use hanging around like this is sus.

I'd disagree on the unclear use. These can be used for mapping, but also for admin, hence they should be public imo.

PJB3005:
There are concerns about how this would even work when PVS is enabled, since presumably we'd want to remove the hack of always synchronizing grids and maps eventually. How would this work then?

PaulRitter:
> There are concerns about how this would even work when PVS is enabled, since presumably we'd want to remove the hack of always synchronizing grids and maps eventually. How would this work then?

we'll cross that bridge when we get there imo. there are multiple approaches to this, either networking grids seperately for these lists, or just sending all grids like we do now but to adminclients or smth

moonheart08:
Sending all grids to admins would be bad for admin performance on servers that use very large numbers of grids simultaneously.

PJB3005:
> either networking grids seperately for these lists

As soon as you go there these lists stop being any sort of "general purpose" solution and start being tied *heavily* to whatever UI/system (mapping mode etc) is using them. I do not think a blanket "list of all grids" control ever makes sense in that instance.

metalgearsloth:
Paul's busy so.

PR #2833, opened 2022-05-16
Reflection Commands POC

PaulRitter:
PoC of an idea of mine i recently mentioned about how we could making defining commands easier.
Uses Serv3 to try to read the type from a valuedatanode created using the arg value.

Usage:
```csharp
[RegisterCommand("test")]
public void Test(IConsoleShell shell, EntityUid uid, Color color, ResourcePath path)
{
    shell.WriteLine($"{uid} | {color} | {path}");
}

[RegisterCommand("test_static")]
public static void TestStatic(IConsoleShell shell, EntityUid uid, Color color, ResourcePath path)
{
    shell.WriteLine($"{uid} | {color} | {path}");
}
```

This can also be used to easily autogenerate help text. for example:
- for arg explanations, function params can be annotated OR xmldocs could be used(? dunno how much we can access xmldocs at runtime)
- for helptext in general, xmldoc (see previous point) or another param on the registercommandattribute could be used

TODO
- log errors if read fails and dont just throw
- integrate into current console system

### DIFF
--- a/Robust.Shared/Console/ReflectionCommands/CommandManager.cs
+++ b/Robust.Shared/Console/ReflectionCommands/CommandManager.cs
@@ -1,0 +1,100 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using Robust.Shared.GameObjects;
+using Robust.Shared.IoC;
+using Robust.Shared.Maths;
+using Robust.Shared.Reflection;
+using Robust.Shared.Serialization.Manager;
+using Robust.Shared.Serialization.Markdown.Value;
+using Robust.Shared.Utility;
+
+namespace Robust.Shared.Console.ReflectionCommands;
+
+public sealed class CommandManager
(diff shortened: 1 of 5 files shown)


Robust.Shared/Console/ReflectionCommands/CommandManager.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        RegisterCommand(typeof(CommandManager).GetMethod("Test")!, instance:this);
+        RegisterCommand(typeof(CommandManager).GetMethod("TestStatic")!);
+
+        var shell = new ConsoleShell(IoCManager.Resolve<IConsoleHost>(), null);
+        ExecuteCommand(shell, "test 1 red path");
+        ExecuteCommand(shell, "test_static 1 red path");
+    }
+
+    public void RegisterCommand(MethodInfo info, RegisterCommandAttribute? attribute = null, object? instance = null)
```
PJB3005, line 47: I'm scared.

PR #2836, opened 2022-05-17
Getting debug tests to work again

wrexbe:
Working on making the debug tests work again

These changes make debug tests go from never working to working most of the time


Robust.Client/Player/PlayerManager.cs
```diff
@@ -149,6 +149,12 @@ private void UpdateAttachedEntity(EntityUid? entity)
                 LocalPlayer.DetachEntity();
                 return;
             }
+            var metadata = IoCManager.Resolve<IEntityManager>().GetComponentOrNull<MetaDataComponent>(entity.Value);
+            if (metadata == null || metadata.EntityDeleted)
+            {
+                LocalPlayer.DetachEntity();
+                return;
+            }
```
metalgearsloth: This should be logging an error or something as they should never be getting a deleted entity.

PR #2844, opened 2022-05-19
Remove ambiguity in sprite specifier

ElectroJr:
Currently it is ambiguous whether a sprite specifier defined in yaml using just a single string is meant to be an entity prototype or a texture path.

Currently this is resolved by doing a `IoCManager.Resolve<Prototypes.IPrototypeManager>().HasIndex()`, and if it fails, it assumes that the string is a resource path for a texture. This significantly slows down entity loading / spawning for some entities.

So this PR makes entity prototype sprite specifiers use a `MappingDataNode` instead. So now:
- `ValueDataNode` implies a texture path
- `MappingDataNode` with an "entity" key implies an entity prototype
- Any other `MappingDataNode` implies an RSI

This will require a content PR to update mapping actions. Otherwise, AFAIK entity prototype sprite specs aren't really used in any yaml prototypes or maps?

### DIFF
--- a/Robust.Shared/Serialization/TypeSerializers/Implementations/SpriteSpecifierSerializer.cs
+++ b/Robust.Shared/Serialization/TypeSerializers/Implementations/SpriteSpecifierSerializer.cs
@@ -81,10 +81,7 @@ namespace Robust.Shared.Serialization.TypeSerializers.Implementations
             IDependencyCollection dependencies,
             ISerializationContext? context)
         {
-            var texNode = ((ITypeReader<Texture, ValueDataNode>) this).Validate(serializationManager, node, dependencies, context);
-            if (texNode is ErrorNode) return texNode;
-
-            return new ValidatedValueNode(node);
+            return ((ITypeReader<Texture, ValueDataNode>) this).Validate(serializationManager, node, dependencies, context);
         }
 
         ValidationNode ITypeValidator<EntityPrototype, ValueDataNode>.Validate(ISerializationManager serializationManager,
@@ -92,7 +89,7 @@ namespace Robust.Shared.Serialization.TypeSerializers.Implementations
             IDependencyCollection dependencies,
             ISerializationContext? context)
         {
-            return !IoCManager.Resolve<Prototypes.IPrototypeManager>().HasIndex<Prototypes.EntityPrototype>(node.Value)
+            return !dependencies.Resolve<Prototypes.IPrototypeManager>().HasIndex<Prototypes.EntityPrototype>(node.Value)
                 ? new ErrorNode(node, $"Invalid {nameof(EntityPrototype)} id")
                 : new ValidatedValueNode(node);
         }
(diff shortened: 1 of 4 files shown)


wrexbe (approved):

metalgearsloth:
@PaulRitter when you get time bb

Robust.Shared/Serialization/TypeSerializers/Implementations/SpriteSpecifierSerializer.cs
```diff
@@ -98,9 +84,6 @@ public sealed class SpriteSpecifierSerializer :
             var texNode = ((ITypeReader<Texture, ValueDataNode>) this).Validate(serializationManager, node, dependencies, context);
             if (texNode is ErrorNode) return texNode;
 
-            var protNode = ((ITypeReader<EntityPrototype, ValueDataNode>) this).Validate(serializationManager, node, dependencies, context);
-            if (protNode is ErrorNode) return texNode;
-
             return new ValidatedValueNode(node);
```
PaulRitter: just return whatever the validate call of the itypereader<tex, val> returns

Robust.Shared/Serialization/TypeSerializers/Implementations/SpriteSpecifierSerializer.cs
```diff
@@ -128,6 +110,9 @@ public sealed class SpriteSpecifierSerializer :
             IDependencyCollection dependencies,
             ISerializationContext? context)
         {
+            if (node.TryGet("entity", out var entityNode) && entityNode is ValueDataNode entityValueNode)
+                return ((ITypeReader<EntityPrototype, ValueDataNode>)this).Validate(serializationManager, entityValueNode, dependencies, context);
+
```
PaulRitter: you should have an explicit error if the entitynode is not a valuedatanode instead of just throwing it at the rsi-validator, resulting in a somewhat useless error message.

Robust.Shared/Serialization/TypeSerializers/Implementations/SpriteSpecifierSerializer.cs
```diff
@@ -114,7 +97,6 @@ public sealed class SpriteSpecifierSerializer :
                 : new ValidatedValueNode(node);
```
PaulRitter, line 94: tbqh this should not ioc.resolve but should either use the dependencycollection, or even create a dependency-field. iirc serv3 injects the typeserializers with dependencies

PaulRitter (changes requested):

Robust.Shared/Serialization/TypeSerializers/Implementations/SpriteSpecifierSerializer.cs
ElectroJr: It doesn't seem to inject, so for now I've switched it and two other IoCManager uses in other files to use the dependency collection.

PaulRitter: ahhh wait i remember, we explicitely *dont* do that bc the dependencycollection might differ iirc. either or, using the dependencycollection is good

PaulRitter (approved):

PR #2846, opened 2022-05-20
Ignore missing components on the client

wrexbe:
Made it so client doesn't care if the component isn't registered.

Content half: https://github.com/space-wizards/space-station-14/pull/8308


PJB3005 (changes requested):
Put this behind a CVar or something. Single player games would be screwed over by this in its current state.

wrexbe:
It is now opt-in on the content client.

mirrorcult (approved):

PJB3005 (approved):

PR #2852, opened 2022-05-23
Add editor copy keybinding

ElectroJr:
Adds a keybinding that starts placing the type of entity or tile that a user's mouse is hovering over.


ElectroJr:
Moved to content. Its weird to have all but one editor keybinding logic in engine, but this keybinding should really check sandbox/admin permissions.

PR #2858, opened 2022-05-24
Sync CurTime base.

PJB3005:
Synchronized on connect, and change of net.tickrate mid game correctly reports intended tick to client so it can make sure it's always synced.

Supersedes #2705

### DIFF
--- a/Robust.Shared/Configuration/NetConfigurationManager.cs
+++ b/Robust.Shared/Configuration/NetConfigurationManager.cs
@@ -1,11 +1,13 @@
 using System;
 using System.Collections.Generic;
+using System.Threading.Tasks;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
 using Robust.Shared.Network;
 using Robust.Shared.Network.Messages;
 using Robust.Shared.Timing;
 using Robust.Shared.Utility;
+using Robust.Shared.Utility.Collections;
 
 namespace Robust.Shared.Configuration
 {
@@ -123,7 +125,7 @@ namespace Robust.Shared.Configuration
 
                 // apply the initial set immediately, so that they are available to
                 // for the rest of connection building
-                ApplyNetVarChange(message.MsgChannel, message.NetworkedVars);
+                ApplyNetVarChange(message.MsgChannel, message.NetworkedVars, message.Tick);
                 ReceivedInitialNwVars?.Invoke(this, EventArgs.Empty);
             }
             else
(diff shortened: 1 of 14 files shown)


Robust.Shared/Timing/GameTiming.cs
```diff
@@ -205,32 +194,50 @@ private TimeSpan CalcFrameTime()
...
 
             TimeSpan newTime;
 
-            if (CurTick.Value >= lastTimeTick.Value)
-              newTime = cachedTime + (TickPeriod * (CurTick.Value - lastTimeTick.Value));
+            if (atTick.Value >= lastTimeTick.Value)
+              newTime = cachedTime + (TickPeriod * (atTick.Value - lastTimeTick.Value));
             else
-              newTime = cachedTime - (TickPeriod * (lastTimeTick.Value - CurTick.Value));
+              newTime = cachedTime - (TickPeriod * (lastTimeTick.Value - atTick.Value));
```
ElectroJr: The if else isn't even needed, is it? The two possible `newTime` expressions are equivalent.

Robust.Shared/Network/Messages/MsgSyncTimeBase.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Robust.Shared.Network.Messages;
+
+/// <summary>
+/// Sent server -> client to synchronize initial TimeBase on connection.
+/// Any further time synchronizations are done through the configuration system.
+/// </summary>
+internal sealed class MsgSyncTimeBase : NetMessage
+{
+    public override MsgGroups MsgGroup => MsgGroups.String;
```
ElectroJr, line 13: I have no idea what `MsgGroups` are for, but based on the doc strings this doesn't seem right?

Robust.Shared/Configuration/NetConfigurationManager.cs
```diff
@@ -136,21 +138,35 @@ public void TickProcessMessages()
...
+                toApply.Add(msg);
 
                 _netVarsMessages.RemoveSwap(i);
                 i--;
             }
+
+            if (toApply.Count == 0)
+                return;
+
+            toApply.Sort();
```
ElectroJr: `MsgConVars` Doesn't appear to have `IComparable` or something else to that effect. Is this actually sorting by tick?

ElectroJr (approved):
Other than the sorting and minor nitpick looks all good, AFAICT this fixes the curtime issues. I haven't managed to break it so far while testing with multiple clients or high packet loss & late cvar messages.

Robust.Shared/Configuration/NetConfigurationManager.cs
PJB3005: You're completely correct, no idea how I missed that.

Robust.Shared/Network/Messages/MsgSyncTimeBase.cs
PJB3005: Basically it controls whether the messages are reliable ordered or not etc. It's a spaghetti nightmare but I can't be bothered to fix it right now so I just made it String which is ReliableOrdered.

Robust.Shared/Timing/GameTiming.cs
PJB3005: You appear to be correct in that.

PR #2860, opened 2022-05-24
Minor performance improvements

PJB3005:
Big blocker for stutter issues is still serv3


mirrorcult:
basado

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -421,7 +422,7 @@ private protected EntityUid AllocEntity(EntityUid uid = default)
         private protected virtual EntityUid CreateEntity(string? prototypeName, EntityUid uid = default)
         {
             if (prototypeName == null)
-                return AllocEntity(uid);
+                return AllocEntity(out _, uid);
 
             var entity = AllocEntity(prototypeName, uid);
             try
```
ElectroJr: Maybe this should also have an `out MetaDataComponent`, given it has to resolve it right after this.

Robust.Shared/GameObjects/ComponentEventArgs.cs
```diff
@@ -24,50 +20,11 @@ public abstract class ComponentEventArgs : EventArgs
...
-        /// <param name="component">The relevant component</param>
-        /// <param name="uid">EntityUid of the entity this component belongs to.</param>
-        public AddedComponentEventArgs(IComponent component, EntityUid uid) : base(component, uid) { }
-    }
-
-    /// <summary>
-    /// Arguments for an event related to a component being removed.
-    /// </summary>
-    public sealed class RemovedComponentEventArgs : ComponentEventArgs
-    {
```
ElectroJr: I was going to add an entity-is-terminating bool to comp remove/delete, so it might still be worth keeping those separate from `ComponentEventArgs`

PJB3005: I re-introduced these types. They're structs now but that should be fine.

PR #2868, opened 2022-05-25
Add helper function in entity prototype

Macoron:
Add new `TryGetComponent` overload to get component from `EntityPrototype` without component name parameter. It will be resolved automatically from `IComponentFactory.GetComponentName()`.


PJB3005 (changes requested):
Definitely not keen on adding new helpers that use IoC resolves like this.

Macoron:
> Definitely not keen on adding new helpers that use IoC resolves like this.

I could probably pass it as an optional parameter.
```csharp
public bool TryGetComponent<T>(IComponentFactory factory = null, [NotNullWhen(true)] out T? component) where T : IComponent
```

PJB3005 (approved):

PR #2871, opened 2022-05-26
Reset last modified tick

ElectroJr:
Currently a component and the entity-metadata's last modified tick don't get reset when resetting predicted entities. 
Before #2826 this didn't really cause issues, other than occasionally unnecessarily resetting an entity that hadn't been modified.


metalgearsloth (approved):
fucken legend

PR #2873, opened 2022-05-26
ACZ general speedups

20kdc:
+ Removes the legacy `client.zip` download method, speeding up ACZ initialization and hopefully halving memory usage.
+ Internal refactoring to make this possible.


Robust.Server/ServerStatus/OnDemandFile.cs
```diff
@@ -0,0 +1,95 @@
...
+using Robust.Shared.Utility.Collections;
+using SharpZstd.Interop;
+
+namespace Robust.Server.ServerStatus;
+
+/// <summary>
+/// An attempt to mitigate the amount of RAM usage caused by ACZ-related operations.
+/// The idea is that Dictionary<string, OnDemandFile> should become the standard interchange.
+/// </summary>
+internal abstract class OnDemandFile
```
PJB3005: I would put this in nested with status host or something, instead of on the global namespace.

Also the naming could be a bit more clear? Something like "AczFileSource"

PJB3005 (commented):
Not sure how I feel about renaming *some* of the usages of "ACZ" to "ACM". I say just leave it all at ACZ to avoid confusion.

20kdc:
ACZ/ACM separation was from back when I expected the codepaths to co-exist. Obviously a vestigial change now, so I'll adjust that, but hopefully it explains things.

Robust.Server/ServerStatus/OnDemandFile.cs
20kdc: Naming not adjusted, but it's been moved into `StatusHost.Acz.Sources.cs` and into the StatusHost class.

20kdc:
Adjustments made

PR #2876, opened 2022-05-27
Predict component adding and removal

### LINKED ISSUE #2144: AddComponent prediction and reconciliation
If you have code in shared that adds a component to an entity but the client mispredicts, the added component will remain.
This should probably need a new `AddPredictedComponent` method or something specifically for situations like this.

ElectroJr:
Fixes #2144. Required for, and tested in-game using space-wizards/space-station-14/pull/8475, but can be merged on its own.

The client-side ClientDirtySystem now tracks removed components. The way in which it was tracking dirty entities was also greatly simplified, given that per-tick dirty hashsets are not required, so all the hashset pooling was unnecessary.


ElectroJr:
I might just not properly understand the gamestate manager/processor, but I think that when `_waitingForFull` is true it uses the wrong tick when calling `ResetPredictedEntities()`? Or more generally, because `CurTick` is sometimes set inside `ProcessTickStates()` before the predicted entities are reset, using `ResetPredictedEntities(_timing.CurTick)` is generally wrong?

I am no longer sure about the differences between `GameStateProcessor.LastProcessedRealState`, `ClientGameStateManager.CurServerTick` (aka `_lastProcessedTick`),  `IGameTiming.LastRealTick`, and how they relate to `IGameTiming.CurTick`.

I'm not sure why `LastProcessedRealState` and `CurTick` are being set by `ProcessTickStates()` before the states are applied, because without also updating `LastRealTick`. it results in `IGameTiming.InPrediction` being true while applying states, which it really shouldn't be.

Somebody please help.

PR #2881, opened 2022-05-28
Revert ShortestDistance changes

### LINKED ISSUE #8497: The camera misbehaves heavily on moving grids.
## Description
Spins violently!

## Reproduction
Try flying a shuttle.

ElectroJr:
reverts #2869. IMO shortest distance doesn't really make sense without reducing the angle, and I think just about every time it or the angle lerp get used you'll want that behaviour.

Instead this PR adds an option to Angle.Lerp to not call shortest distance in the first place and makes animation tracks use that.

Fixes space-wizards/space-station-14/issues/8497

### DIFF
--- a/Robust.Client/Animations/AnimationTrackProperty.cs
+++ b/Robust.Client/Animations/AnimationTrackProperty.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


mirrorcult:
...so ghost orbit lerp is broken again now that animation tracks don't use shortest distance?

PR #2887, opened 2022-05-31
Expose RemoveComponentDeferred

wrexbe:
There are at least 5 places in content that are rolling their own method of deferring component deletion, because you can't call RemoveComponent inside of an EntityQuery. The functionality to defer was already there, so I added a way to call it.


PaulRitter (approved):
might as well add a generic variant while ur at it

PaulRitter (approved):

PR #2889, opened 2022-05-31
Default lag

metalgearsloth:
As agreed in the 2022/05/14 maintainer meeting

image

Might need to be 0.075 instead?


PJB3005:
Maybe set this in SS14 instead?

PJB3005:
Also this isn't disabling it when on RELEASE anymore. Release clients have fakelag compiled in now so this needs to be disabled there.

metalgearsloth:
> Maybe set this in SS14 instead?

Probably a good idea though not sure how to override CVars in content easily?

> Also this isn't disabling it when on RELEASE anymore.

A

PR #2896, opened 2022-06-01
Remove allocations from physics ComputeDistance()

PJB3005:


metalgearsloth (approved):
Will know if it works if steptriggers stop working

PR #2899, opened 2022-06-02
Fix grid deletion crashing

### LINKED ISSUE #2750: Source of the FindNewContacts crashes
Need to delete a grid (on a paused map?) and occasionally you can get it to happen, same with spawning a new one on top of the previously deleted spot.

### LINKED ISSUE #2839: Server crash from grid entity deletion
image

https://central.spacestation14.io/grafana/explore?orgId=1&left=%7B%22datasource%22:%22Loki%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7BApp%3D%5C%22Robust.Server%5C%22,Server%3D%5C%22wizards_den_lizard%5C%22%7D%22%7D%5D,%22range%22:%7B%22from%22:%221652896022769%22,%22to%22:%221652896028474%22%7D%7D

Seemed to cause a sequence of exceptions all the way into PVS where it got stuck.

metalgearsloth:
Fixes https://github.com/space-wizards/RobustToolbox/issues/2750
Fixes https://github.com/space-wizards/RobustToolbox/issues/2839

### DIFF
--- a/Robust.Shared/Physics/SharedBroadphaseSystem.cs
+++ b/Robust.Shared/Physics/SharedBroadphaseSystem.cs
@@ -413,27 +413,6 @@ namespace Robust.Shared.Physics
 
             if (broadphase == null) return;
 
-            // Juussttt in case anything slips through
-            if (!broadphase.Owner.IsValid() ||
-                !TryComp(broadphase.Owner, out MetaDataComponent? meta) ||
-                meta.EntityLifeStage >= EntityLifeStage.Terminating)
-            {
-                // Don't log because this may happen due to recursive deletions.
-                body.Broadphase = null;
-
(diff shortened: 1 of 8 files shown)


Robust.Shared/Physics/SharedBroadphaseSystem.cs
```diff
@@ -413,27 +413,6 @@ private void DestroyProxies(PhysicsComponent body, FixturesComponent? manager =
...
-                    {
-                        proxy.ProxyId = DynamicTree.Proxy.Free;
-                    }
-
-                    fixture.ProxyCount = 0;
-                }
-
-                return;
-            }
-
```
metalgearsloth, line 436: Don't remember why I wrote this garbage but I don't think the bandaid is needed anymore.

PR #2905, opened 2022-06-04
Make TryGetComponent output nullable

ElectroJr:
Requires https://github.com/space-wizards/space-station-14/pull/8634

Can cause bugs, e.g. space-wizards/space-station-14/pull/8632

### DIFF
--- a/Robust.Server/Console/Commands/PlayerCommands.cs
+++ b/Robust.Server/Console/Commands/PlayerCommands.cs
@@ -100,7 +100,7 @@ namespace Robust.Server.Console.Commands
                     return;
                 }
 
-                if (!entMan.TryGetComponent(player.AttachedEntity, out TransformComponent playerTransform))
+                if (!entMan.TryGetComponent(player.AttachedEntity, out TransformComponent? playerTransform))
                 {
                     shell.WriteError("You don't have an entity.");
                     return;
--- a/Robust.Server/Console/Commands/SpinCommand.cs
+++ b/Robust.Server/Console/Commands/SpinCommand.cs
@@ -50,7 +50,7 @@ public sealed class SpinCommand : IConsoleCommand
         }
 
         // Try get physics
-        if (!entMan.TryGetComponent(target, out PhysicsComponent physics))
+        if (!entMan.TryGetComponent(target, out PhysicsComponent? physics))
         {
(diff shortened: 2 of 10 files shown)


Visne:
I think you should change the a parameter to use `[MaybeNullWhen(false)]` instead of `[NotNullWhen(true)]`, then you don't need to use `!` everywhere (which is very bad, in my opinion).

Visne:
Then I think you can also undo space-wizards/space-station-14#8632, after chaning `||` to `&&`.

PJB3005:
> then you don't need to use ! everywhere (which is very bad, in my opinion).

In which cases do you need to use `!`? The C# compiler already tracks the not null correctly.

I'm fairly confident `NotNullWhen` is the correct attribute to use here.

Visne:
> In which cases do you need to use `!`? The C# compiler already tracks the not null correctly.

I count at least nine in the changed files.

Visne:
I guess that's because there's no if-statements surrounding them. I'm pretty sure that `[NotNullWhen(true)] out type? value` is essentially equivalent to `[MaybeNullWhen(false)] out type value`, so I guess it doesn't matter then.

ElectroJr:
> I guess that's because there's no if-statements surrounding them. I'm pretty sure that [NotNullWhen(true)] out type? value is essentially equivalent to [MaybeNullWhen(false)] out type value, so I guess it doesn't matter then.

Yeah I think it just needs an if around it. The `Assert.That()` in tests don't seem to make the nullability / error checker happy. Though it seems to work fine for `DebugTools.Assert()`, but I think that's because it has a `[AssertionMethod]` attribute?
Then again, this whole time my VS error checking was throwing a fit, so maybe it just complained without cause and refused to build for no good reason.

> I'm fairly confident NotNullWhen is the correct attribute to use here.

AFAIK it is.

> Then I think you can also undo https://github.com/space-wizards/space-station-14/pull/8632, after chaning || to &&.

I think it needs to be ||. Otherwise the ignore weightless component would only affect static or kinematic physics objects, which doesn't make sense. AFAIK it is only meaningful for Dynamic or KinematicController bodies to be weightless.

Visne:
> I think it needs to be ||.

Yeah, that was if you would change it to `MaybeNullWhen`, I think, but now it's fine.

PR #2910, opened 2022-06-05
No more spoopy invisible ghosts

metalgearsloth:
IsEntityInContainer is not recursively done (i.e. container -> ent -> ent2, ent2 has no knowledge of being in a container) hence we need to do this.


Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
```diff
@@ -25,14 +25,14 @@ public override void Initialize()
         {
             base.Initialize();
 
-            SubscribeLocalEvent<UpdateContainerOcclusionMessage>(UpdateContainerOcclusion);
+            SubscribeLocalEvent<EntParentChangedMessage>(OnParentChange);
```
ElectroJr: The shared system already has this subscription, and the handler already gets overridden by the client side system, so this new handlers should just be added to that

ElectroJr: also for performance reasons, should avoid queuing the update if the entity gets moved to null space

ElectroJr (approved):
Other than merging the parent-change handlers, seems fine.

ElectroJr:
While reading the code I realised there's an existing bug, where a `while (parent.IsValid() && !spriteOccluded && !lightOccluded)` which should be a `while (parent.IsValid() && ( !spriteOccluded || !lightOccluded))`

Might as well include a fix for that in this PR

PR #2912, opened 2022-06-05
Switch YamlDotNet to submodule

wrexbe:
Need to try this with content

Requires https://github.com/space-wizards/YamlDotNet/pull/1
Going to use my own submodule in the mean time


PJB3005:
Still not sure whether this should be a submodule or a NuGet package. I'm worried about IDE and compile overhead from throwing too many projects into the solution.

wrexbe:
Closing in the hope the official one is going to get updated again.

PR #2914, opened 2022-06-05
Add generic visualizer system

ElectroJr:
Currently SS14 has a generic enum visualizer, which allows someone to use yaml to map some appearance component data onto a sprite layer state. However there is no way to specify other sprite layer data (e.g., shader, scale, etc). Also, given that this is generally useful and not at all ss14 specific, this functionality should probably be in engine.

So this PR adds a generic visualizer system to the engine, which extends the generic enum visualizer so that it can be used to fully specify the sprite layer data. It also uses string keys, instead of enums. This adds support for things like the `PowerDeviceVisuals.Powered` appearance data, which is a bool. 

For example. the yaml for the PAI visualizer used to look like this:
```yaml
- type: Appearance
  visuals:
  - type: GenericEnumVisualizer
    key: enum.PAIVisuals.Status
    layer: 1
    states:
      enum.PAIStatus.Off: pai-off-overlay
      enum.PAIStatus.Searching: pai-searching-overlay
      enum.PAIStatus.On: pai-on-overlay
```
Now it looks like:
```yaml
  - type: GenericVisualizer
    visuals:
      enum.PAIVisuals.Status:
        unshaded-layer:
          Off: { state: pai-off-overlay }
          Searching: { state: pai-searching-overlay }
          On: { state: pai-on-overlay }
```

This didn't and still doesn't have any yaml linter checking. Given that the keys & values for the various dictionaries & maps can be either strings, enums, or generic objects, I don't really think there is any reasonable way of linting that without adding restrictions on the types.

This PR also:
- Adds `SetData()` and `TryGetData()` functions to the appearance system, which should be used instead of the component-functions. 
- Makes some changes to sprite component, so that more `PrototypeLayerData` fields are nullable.

### DIFF
--- a/Robust.Client/GameObjects/Components/Appearance/GenericVisualizerComponent.cs
+++ b/Robust.Client/GameObjects/Components/Appearance/GenericVisualizerComponent.cs
@@ -20,6 +20,6 @@ public sealed class GenericVisualizerComponent : Component
     ///
     ///     In most instances, each of these dictionaries will probably only have a single entry.
     /// </summary>
-    [DataField("data", required:true)]
-    public Dictionary<string, Dictionary<string, Dictionary<string, PrototypeLayerData>>> Data = default!;
+    [DataField("visuals", required:true)]
+    public Dictionary<string, Dictionary<string, Dictionary<string, PrototypeLayerData>>> Visuals = default!;
 }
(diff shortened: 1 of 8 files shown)


mirrorcult (approved):
IMO the pattern of deserializing `PrototypeLayerData` and just doing `LayerSetData` needs to be used more in content

PR #2916, opened 2022-06-06
Add State application bool to IGameTiming

ElectroJr:
Currently there is no way for systems to know whether or not the game is currently in the process of applying a game state. E.g., the grid traversal system listens for `MoveEvent`, but these could be raised either while actually predicting entity movement or during game state application (new server state or prediction resetting). If the movement was due to state handling, the grid traversal induced parent change should be handled by the transform component's own state handling code. In some instances, this can mess up the state application. So this PR adds a `ApplyingState` bool to `IGameTiming` and adds some checks in the physics and grid traversal systems.

Also makes debugging some movement related mispredicts much easier.


metalgearsloth:
Zased.

When we stop physics from updating during FrameUpdate on the client we can probably move the deferral loop too for grid traversal.

Robust.Client/Timing/ClientGameTiming.cs
```diff
@@ -62,5 +62,15 @@ public void EndPastPrediction()
...
+
+        public void StartStateApplication()
+        {
+            ApplyingState = true;
+        }
+
+        public void EndStateApplication()
+        {
+            ApplyingState = false;
+        }
```
metalgearsloth, line 74: I don't have the pr checked out but is this ever called?

ElectroJr: It should get called when the `StateApplicationGuard` gets disposed.
I just coped the `PredictionGuard` set up with the `using`. I assume its preferred over just encolsing it all in a try {} except{}? I'm not even sure if the end result is different. Or if the game can even recover if state application/resetting results in an exception.

ElectroJr: i.e., if there is an error while applying the game state, I think the client will just enter an error loop. So AFAIK there's currently no point in the `using` s

PR #2919, opened 2022-06-06
Matrix3 and Angle: add readonly and ref->in

ElectroJr:
Adds some `readonly` keywords, and replaces some `ref`s with `in`s


PaulRitter (approved):
just out of curiousity, what does making a function readonly do? is this some sort of variant of the pure-attribute? and since when is this possible, lol

PJB3005:
> just out of curiousity, what does making a function readonly do? is this some sort of variant of the pure-attribute? and since when is this possible, lol

It enforces that the function doesn't mutate its `this` reference directly. Without it, functions on an `in` parameter need to do expensive shadow copies every time.

PR #2921, opened 2022-06-06
Make Eye use Matrix3.CreateTransform

ElectroJr:
Avoids unnecessary matrix multiplication and inversion.

Also removes `MathF.SinCos` from those functions, because it is apparently slower than getting sin and cos separately.

### DIFF
--- a/Robust.Client/Graphics/ClientEye/Eye.cs
+++ b/Robust.Client/Graphics/ClientEye/Eye.cs
@@ -78,8 +78,8 @@ namespace Robust.Client.Graphics
         public void GetViewMatrixInv(out Matrix3 viewMatrixInv, Vector2 renderScale)
         {
             viewMatrixInv = Matrix3.CreateTransform(
-                _coords.Position.X,
-                _coords.Position.Y,
+                _coords.Position.X + Offset.X,
+                _coords.Position.Y + Offset.Y,
                 (float)-Rotation.Theta,
                 1 / (_scale.X * renderScale.X),
                 1 / (_scale.Y * renderScale.Y));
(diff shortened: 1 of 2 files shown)


Robust.Client/Graphics/ClientEye/Eye.cs
```diff
@@ -56,27 +56,33 @@ public Vector2 Scale
         /// <inheritdoc />
         public void GetViewMatrix(out Matrix3 viewMatrix, Vector2 renderScale)
         {
-            var scaleMat = Matrix3.CreateScale(_scale.X * renderScale.X, _scale.Y * renderScale.Y);
-            var rotMat = Matrix3.CreateRotation(_rotation);
-            var transMat = Matrix3.CreateTranslation(-_coords.Position - Offset);
-
-            viewMatrix = transMat * rotMat * scaleMat;
+            viewMatrix = Matrix3.CreateInverseTransform(
```
ElectroJr, line 59: CreateTransform is equivalent to scale * rotate * translate, which is not the same order of operations as the view matrix uses. So this instead uses CreateInverseTransform with modified inputs.

Robust.Shared.Maths/Matrix3.cs
```diff
@@ -367,8 +367,8 @@ public static Matrix3 CreateTranslation(float x, float y)
             var result = Identity;
 
             /* column major
-             0 0 x
-             0 0 y
+             1 0 x
+             0 1 y
              0 0 1
```
PaulRitter, line 372: bruh

PaulRitter (approved):
questioning the purpose of MathF.SinCos' existence rn

PR #2922, opened 2022-06-07
Adds a filter to all except origin

keronshb:
@mirrorcult coded it for my shield branch so I figured it would be nice to have as a general option.

### DIFF
--- a/Robust.Shared/Player/Filter.cs
+++ b/Robust.Shared/Player/Filter.cs
@@ -355,6 +355,14 @@ namespace Robust.Shared.Player
         }
 
         /// <summary>
+        ///     A filter with every player who's PVS overlaps this point except the original player.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


Robust.Shared/Player/Filter.cs
```diff
@@ -354,6 +354,14 @@ public static Filter Pvs(MapCoordinates origin, float rangeMultiplier = 2f)
             return Empty().AddPlayersByPvs(origin, rangeMultiplier);
         }
 
+        /// <summary>
+        ///     A filter with every player who's PVS overlaps this point except the original player.
```
ike709:
```suggestion
        ///     A filter with every player whose PVS overlaps this point except the original player.
```

keronshb: Not that I don't want to fix this but all the other ones say who's.

Should I fix those too?

PaulRitter (approved):

PR #2923, opened 2022-06-07
Remove almost all allocations from F3 menu.

PJB3005:
You'll have to agree it's a bit wack that the allocations counter skews itself significantly.

### DIFF
--- a/Robust.Client/UserInterface/Controls/Label.cs
+++ b/Robust.Client/UserInterface/Controls/Label.cs
@@ -38,10 +38,13 @@ namespace Robust.Client.UserInterface.Controls
         /// <remarks>
         /// Replaces <see cref="TextMemory"/> when set.
         /// </remarks>
+        /// <exception cref="InvalidOperationException">
+        /// Thrown if <see cref="TextMemory"/> was set directly and there is no backing string instance to fetch.
+        /// </exception>
         [ViewVariables]
         public string? Text
         {
-            get => _text;
+            get => _text ?? (_textMemory.Length > 0 ? throw new InvalidOperationException("Label uses TextMemory, cannot fetch string text.") : null);
             set
             {
                 _text = value;
@@ -60,7 +63,7 @@ namespace Robust.Client.UserInterface.Controls
         /// If you modify the backing storage, re-assign the property to invalidate these.
         /// </para>
         /// <para>
-        /// Sets <see cref="Text"/> to null when assigned, as there is no backing string instance anymore.
+        /// Sets <see cref="Text"/> to throw an exception if read, as there is no backing string to retrieve.
         /// </para>
         /// </remarks>
         public ReadOnlyMemory<char> TextMemory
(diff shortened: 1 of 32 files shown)


Robust.Client/UserInterface/Controls/Label.cs
```diff
@@ -34,13 +35,41 @@ public Label()
...
+        /// <summary>
+        /// The text to display, set as a read-only memory.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Note that updating the backing memory while the control is using it can result in incorrect display due to caching of measure information and similar.
+        /// If you modify the backing storage, re-assign the property to invalidate these.
+        /// </para>
+        /// <para>
+        /// Sets <see cref="Text"/> to null when assigned, as there is no backing string instance anymore.
```
PaulRitter: whats stopping us from making the text property return the textmemory span so make this more consistent to deal with. imo, if i would be using the label control, i'd expect Text to always return the content. this will definitely confuse some ppl for little to no gain in my eyes. just do

```
    get => _text ?? _textMemory.Span;
```

or however else you get the string from the textmemory obj

Robust.Client/UserInterface/Controls/Label.cs
```diff
@@ -34,13 +35,41 @@ public Label()
...
+        /// Replaces <see cref="TextMemory"/> when set.
+        /// </remarks>
         [ViewVariables]
         public string? Text
         {
             get => _text;
             set
             {
                 _text = value;
+                _textMemory = value.AsMemory();
```
PaulRitter, line 51: this can cause a nullpointerexception i think

PaulRitter (changes requested):
solid pr but the label api change is kinda icky

Robust.Client/UserInterface/Controls/Label.cs
PJB3005: `AsMemory()` is an extension method and returns an empty memory for null, so no it cannot.

Robust.Client/UserInterface/Controls/Label.cs
PJB3005: 1. Returning a span would make no sense since we're storing a `Memory<char>`, not a span.
2. I didn't want to take a thermonuclear breaking change for a lower level performance API like this.

PJB3005: There are two "clean" solutions here:
1. Take a thermonuclear breaking change and remove `string Text` entirely in favor of `ReadOnlyMemory<char>`. Probably a massive PITA for everybody in the future, all for this somewhat low-level performance API.
2. Rewrite the entire UI styling system (and I have no idea what that would look like) so it's easier for me to make an alternative control with different semantics. The latter part here was my original idea until I realized the styling system would eat my face.

I think you can see the problem here.

PaulRitter: > Returning a span would make no sense since we're storing a `Memory<char>`, not a span.

see my post

> or however else you get the string from the textmemory obj

if i had to choose between having slightly less allocations during label usage or having the api be sane to work with, i'd pick the former.

PJB3005: We came to an agreement on Discord, exception it is 
image

PR #2930, opened 2022-06-09
Various test changes

wrexbe:
Tests fall apart when you try to reuse instances, these changes make them fall apart less often, and also let me bypass the current testing pool, so I can setup/clean/pool them differently.
Contains:
* Duct tape
* * Ping handling ServerChannel being null so it doesn't blow up in the tests sometimes
* * UpdateAttachedEntity being better at dealing with garbage (although I think garbage still gets through sometimes)
* * Wrap _fromInstanceReader.ReadAsync(cancellationToken); in a try, catch. It just makes it more graceful if it cancels.
* Pooling fixes
* * Make DataDefinitions not static, they are not the same between different instances (ex: client vs server, or different settings)
* * Add DoConnect to the _taskManager, which makes stuff wait for DoConnect to finish, which makes it more reliable, instead of based on luck.
* Bypass Pooling
* * Made ServerIntegrationInstance, and ClientIntegrationInstance public, so I can avoid the built in pooling. (which is very unreliable right now)
* Ignore missing types, and ids when removing prototypes with a string
* * If I add prototypes to both the client/server, they may not all be added because of differences between types
* * So I need remove to ignore missing things to undo what add is doing.

Content side (should not be needed for this PR): https://github.com/space-wizards/space-station-14/pull/8737


Robust.Shared/Serialization/Manager/SerializationManager.cs
```diff
@@ -50,6 +51,7 @@ public void Initialize()
             _initializing = true;
 
             DependencyCollection = IoCManager.Instance ?? throw new NullReferenceException();
+            var isServer = DependencyCollection.Resolve<INetManager>().IsServer;
```
PaulRitter: ?

Robust.Shared/Serialization/Manager/SerializationManager.cs
```diff
@@ -34,7 +35,7 @@ public sealed partial class SerializationManager : ISerializationManager
         private bool _initialized;
 
         // Using CWT<,> here in case we ever want assembly unloading.
-        private static readonly ConditionalWeakTable<Type, DataDefinition> DataDefinitions = new();
+        private readonly ConditionalWeakTable<Type, DataDefinition> DataDefinitions = new();
```
PaulRitter, line 38: uhhhhhhhhhh why

Robust.Client/Player/PlayerManager.cs
```diff
@@ -144,7 +144,9 @@ private void UpdateAttachedEntity(EntityUid? entity)
                 return;
             }
 
-            if (entity == null)
+            if (entity == null ||
+                !IoCManager.Resolve<EntityManager>().TryGetComponent<MetaDataComponent>(entity.Value, out var metaData) ||
```
PaulRitter: couldnt you use MetaData() here instead of the tryget?

Robust.Shared/Serialization/Manager/SerializationManager.cs
wrexbe: DataDefinitions can be different between client, and server instances (or even the same kind with different options), when I was running them in tests, it was sometimes deserializing wrong, and crashing.

Robust.Shared/Serialization/Manager/SerializationManager.cs
wrexbe: This was code I forgot to remove when I was testing the issue with the static.

Robust.Client/Player/PlayerManager.cs
wrexbe: PlayerManager doesn't have access to that, but I found a better method I can use on EntityManager

PaulRitter (approved):

Robust.Client/Player/PlayerManager.cs
```diff
@@ -143,8 +143,8 @@ private void UpdateAttachedEntity(EntityUid? entity)
             {
                 return;
             }
-
-            if (entity == null)
+            if (entity == null ||
+                !IoCManager.Resolve<EntityManager>().HasComponent<MetaDataComponent>(entity.Value))
```
metalgearsloth: I mentioned this before but I'm pretty sure this is a flaw of playerstate merging where it desyncs from entity states and should be marked as such IMO (i.e. via github issue and / or a TODO).

wrexbe: Don't need it now. Maybe something fixed it.

ElectroJr (approved):

PR #2932, opened 2022-06-10
Fix some grid traversal misprediction

ElectroJr:
Grid traversal would queue events and process them during a system update, but this could lead to movement events being raised during frame updates, prediction getting reset, and then that future-event being processed during the next tick.

This just moves the grid traversal processing to happen at the end of `PhysicsSystem.SimulateWorld()`. Apparently having traversal processed during a frame update used to cause issues at some point, but I haven't seen anything weird so far.

There is still a separate issue causing some grid traversal issues, this is only a partial fix.

### DIFF
--- a/Robust.Client/GameStates/ClientGameStateManager.cs
+++ b/Robust.Client/GameStates/ClientGameStateManager.cs
@@ -225,6 +225,9 @@ namespace Robust.Client.GameStates
                     using var _ = _timing.StartStateApplicationArea();
 
                     ResetPredictedEntities(_timing.CurTick);
+
+                    // This check shouldn't be necessary anymore, but I am paranoid.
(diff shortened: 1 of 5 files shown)


Robust.Client/GameStates/ClientGameStateManager.cs
```diff
@@ -225,6 +225,9 @@ public void ApplyGameState()
                     using var _ = _timing.StartStateApplicationArea();
 
                     ResetPredictedEntities(_timing.CurTick);
+
+                    // I hate this..
+                    _entitySystemManager.GetEntitySystem<SharedGridTraversalSystem>().QueuedEvents.Clear();
```
metalgearsloth, line 230: :stare:

Really need stop calling SimulateWorld in FrameUpdate, it causes too many issues.

PR #2937, opened 2022-06-11
Add simple raycast method

metalgearsloth:


Robust.Shared.Maths/MathHelper.cs
```diff
@@ -615,6 +616,72 @@ public static double InterpolateCubic(double preA, double a, double b, double po
...
+        // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+        // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+        // SOFTWARE.
+
+        /// <summary>
+        /// Gets the intersection between a line and a circle.
+        /// Essentially a reduced raycast.
+        /// </summary>
+        /// <returns></returns>
+        public static bool TryGetIntersecting(Vector2 start, Vector2 end, float radius, [NotNullWhen(true)] out Vector2? point)
```
moonheart08, line 648: Shouldn't this take a Circle instead of just the radius, or do you expect the user to adjust.

metalgearsloth: I don't really expect anyone to use it besides docking TBH. The raycast method from box2d takes in a circle and a transform so seemed unnecessary.

PR #2942, opened 2022-06-12
Make Grid Uid Nullable

ElectroJr:
- Replaces `EntityUid GridEntityId` with `EntityUid? GridUid`. 
- Makes the EntityUid not use a get-comp. Accessing the old GridId is now slow.
- Adds some more function variants that accept EntityUIid instead of GridId
- Add several `Obsolete` attributes

Requires space-wizards/space-station-14/pull/8798


metalgearsloth (approved):

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -72,33 +75,22 @@ public sealed class TransformComponent : Component, IComponentDebug
...
-                var xformQuery = _entMan.GetEntityQuery<TransformComponent>();
+        [Access(typeof(SharedTransformSystem))]
+        internal EntityUid? _gridUid = null;
 
-                while (childEnumerator.MoveNext(out var child))
-                {
-                    xformQuery.GetComponent(child.Value).GridID = value;
-                }
-            }
+        [Obsolete("Use GridEntityId")]
```
gradientvera:
```suggestion
        [Obsolete("Use GridUid")]
```

PR #2945, opened 2022-06-13
Fix entity manager shutdown

wrexbe:
Changed some stuff that looked sus.
Events got removed on shutdown, but never got re-added. I moved removing the events to cleanup, which looks like a destructive action, where as shutdown looks like it's just trying to reset things. I moved the queue clearing stuff to the method that is deleting all the entities, which looks like a more stable thing to do. I made Shutdown not re-allow initialization, because that would just break it. I also added a check to Startup, that Initialize was already called, because why not.


Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -103,16 +105,15 @@ public virtual void Shutdown()
...
 
         public void Cleanup()
         {
-            QueuedDeletions.Clear();
-            QueuedDeletionsSet.Clear();
+            _componentFactory.ComponentAdded -= OnComponentAdded;
+            _componentFactory.ComponentReferenceAdded -= OnComponentReferenceAdded;
+            FlushEntities();
             _entitySystemManager.Clear();
-            Entities.Clear();
```
ElectroJr, line 115: Its probably worth keeping this, just in case some entity fails to get deleted?

ElectroJr, line 115: Its probably worth keeping this, just in case some entity fails to get deleted?

ElectroJr (approved):

Robust.Shared/GameObjects/EntityManager.cs
PaulRitter: i'd much rather this be an assert than to just silently clear an error away since if this were to ever happen i could imagine it having some pretty nasty consequences elsewhere

PR #2946, opened 2022-06-13
Make fixture init event order explicit

### LINKED ISSUE #2907: Physics relies on unguaranteed local EventBus ordering
`CollisionWakeSystem` and `FixtureSystem` both subscribe to `PhysicsInitializedEvent` (the former with a directed sub). These two systems implicitly rely on `CollisionWakeSystem` getting fired *first* due to implementation details in EventBus, despite not having any explicit ordering specified. If `CollisionWakeSystem` fires second then non-colliding fixtures get into broadphase which explodes *very violently*.

wrexbe:
Closes https://github.com/space-wizards/RobustToolbox/issues/2907


Robust.Shared/Physics/FixtureSystem.cs
```diff
@@ -29,7 +29,7 @@ public override void Initialize()
             SubscribeLocalEvent<FixturesComponent, ComponentGetState>(OnGetState);
             SubscribeLocalEvent<FixturesComponent, ComponentHandleState>(OnHandleState);
 
-            SubscribeLocalEvent<PhysicsInitializedEvent>(OnPhysicsInit);
+            SubscribeLocalEvent<PhysicsInitializedEvent>(OnPhysicsInit, after: new[] { typeof(CollisionWakeSystem) });
```
metalgearsloth: Probably better off just explicitly calling it rather than using ordered events tbh.

wrexbe: I think I can do that... give me a bit

wrexbe:
Is this what you had in mind?

Robust.Server/GameObjects/EntitySystems/PhysicsSystem.cs
```diff
@@ -22,6 +24,18 @@ public override void Initialize()
...
+        protected override void OnPhysicsInitialized(EntityUid uid)
+        {
+            if (EntityManager.TryGetComponent(uid, out CollisionWakeComponent? wakeComp))
+            {
+                _collisionWakeSystem.OnPhysicsInit(uid, wakeComp);
+            }
+            if (EntityManager.TryGetComponent(uid, out FixturesComponent? fixtureComp))
+            {
+                _fixtureSystem.OnPhysicsInit(uid, fixtureComp);
+            }
```
metalgearsloth: Client should also be doing this, rest is probably okay.

PR #2948, opened 2022-06-13
Misc rendering changes

ElectroJr:
- Make overlays use OverlayDrawArgs bounds
- Stop Clyde calculating World bounds twice (once for bounds, once for aabbs).
- Add WorldToLocalMatrix to ClydeViewport
  - Make `ProcessSpriteEntities` use this instead.
- Ever so slightly speed up ClydeHandle equality.
  - Very minor, but this gets called very frequently,
  - Then again maybe it was just a debug / profiling artefact that isn't even present during release?


metalgearsloth:
Debug speedups are still nice, as long as release isn't made worse.

PR #2950, opened 2022-06-14
"Fix" another client connection crash

ElectroJr:
Again, this doesn't actually fix the underlying networking issue, this just makes the client handle it. Instead of crashing, it now just reports that there was an error and you can just hit the connect button again

### DIFF
--- a/Robust.Shared/Network/NetManager.ClientConnect.cs
+++ b/Robust.Shared/Network/NetManager.ClientConnect.cs
@@ -115,6 +115,7 @@ namespace Robust.Shared.Network
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


Robust.Shared/Network/NetManager.ClientConnect.cs
```diff
@@ -176,7 +177,7 @@ public async void ClientConnect(string host, int port, string userNameRequest)
                 if (keyBytes.Length != CryptoBox.PublicKeyBytes)
                 {
                     connection.Disconnect("Invalid public key length");
-                    return;
+                    throw new Exception("Invalid public key length");
```
moonheart08: Include logging for the length and the expected length.

PR #2953, opened 2022-06-15
Don't run client physics in frameupdate

metalgearsloth:
It's the source of too many headaches and means the client sim differs a fair amount from the server sim. Help wanted.

So AFAIK the only issue is that lerping only works for server states coming in, not client physics. Essentially what we need to do is whenever client sets LocalPosition of an entity it attempts to lerp it. It might be worthwhile making a new event for this instead of MoveEvent, maybe FrameMoveEvent?

This should also be separate I think from server lerping where we may not want to linearly lerp.

Also as a result of this PR we should probably add substepping after (i.e. dividing the frametime by n and running n steps).

Ready for review in the sense that "please look at it and consider" as I imagine some reviewers will probably ignore because it's draft.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/TransformSystem.Component.cs
+++ b/Robust.Client/GameObjects/EntitySystems/TransformSystem.Component.cs
@@ -1,0 +1,16 @@
+using Robust.Shared.GameObjects;
+using Robust.Shared.Maths;
+
+namespace Robust.Client.GameObjects;
+
+public sealed partial class TransformSystem
+{
+    public override void SetLocalPosition(TransformComponent xform, Vector2 value)
+    {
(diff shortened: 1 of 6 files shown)


moonheart08:
How do we keep movement feeling responsive?

metalgearsloth:
> How do we keep movement feeling responsive?

You're never going to get as responsive as same frame, best you could do is have the client doing its substeps in frameupdate maybe, but you're also not going to have to deal with the headaches that lead to prediction issues in gridtraversal, movementspeedmodifier, movercontroller, and pulling as a result.

Also if we want to do a slight movement acceleration it should be much more feasible now as you won't get the integration issues you would with the client doing it every frame.

PR #2955, opened 2022-06-16
Tile rotation

ElectroJr:
Allows `TileRenderFlags` to change tile rotation & mirroring, and makes placement manager set render flags based on the current placement rotation.

https://user-images.githubusercontent.com/60421075/173976962-83c8313a-63f7-4ed4-af4f-fd5604a091e4.mp4

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
@@ -53,10 +53,7 @@ namespace Robust.Client.Graphics.Clyde
 
                 while (enumerator.MoveNext(out var chunk))
                 {
-                    if (_isChunkDirty(grid, chunk))
-                    {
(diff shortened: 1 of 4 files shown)


moonheart08 (changes requested):
While I get you wanted it done, this just physically rotates the sprite which isn't great, and I was intending to implement rotation states ala RSI, so tiles could have dedicated rotational variants. Hence me wanting to look into RSI support here. Flip's fine though.

ElectroJr:
>  implement rotation states ala RSI

Do you just mean so that each rotation for a tile can be visually distinct, or actually having directional textures that depend on the viewers relative eye rotation?

moonheart08:
> > implement rotation states ala RSI
> 
> Do you just mean so that each rotation for a tile can be visually distinct, or actually having directional textures that depend on the viewers relative eye rotation?

Actual directional textures would be optimal but visual distinction is fine.

PR #2970, opened 2022-06-21
Sprite layer draw depths

ElectroJr:
Revival of #2582.

This PR adds a draw depths to sprite layers, which determines the order in which they are drawn. It also adds RSI-direction based overrides, so that you could choose to have a layer behind another if an entity is facing a specific direction.
 The change is motivated by https://github.com/space-wizards/space-station-14/issues/4387 and https://github.com/space-wizards/space-station-14/issues/4563,

Previously layers would just get drawn in the order they were added to the sprite's layer list. Now it sorts them based on a draw depth & render order, similar to how entities/sprites get drawn. In order to avoid having to sort a list every time a sprite gets drawn, this caches sorted layer arrays, with one array per relevant direction. Overall this will probably be slightly slower than sprite rendering was before, but not by much.

Also makes some misc sprite/layer changes, like renaming `Layer.RSI` to `Layer.Rsi`, so that it is consistent with `ISpriteLayer.Rsi`

Requires space-wizards/space-station-14/pull/8989

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/ISpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/ISpriteComponent.cs
@@ -211,9 +211,9 @@ namespace Robust.Client.GameObjects
         /// </summary>
         RSI? LayerGetActualRSI(object layerKey);
 
-        ISpriteLayer this[int layer] { get; }
-        ISpriteLayer this[Index layer] { get; }
-        ISpriteLayer this[object layerKey] { get; }
+        SpriteComponent.Layer this[int layer] { get; }
+        SpriteComponent.Layer this[Index layer] { get; }
+        SpriteComponent.Layer this[object layerKey] { get; }
 
         IEnumerable<ISpriteLayer> AllLayers { get; }
 
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -179,7 +179,7 @@ namespace Robust.Client.GameObjects
                 for (var i = 0; i < Layers.Count; i++)
                 {
                     var layer = Layers[i];
-                    if (!layer.State.IsValid || layer.RSI != null)
+                    if (!layer.State.IsValid || layer.Rsi != null)
                     {
                         continue;
                     }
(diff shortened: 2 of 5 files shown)


Robust.Client/GameObjects/Components/Renderable/ISpriteComponent.cs
```diff
@@ -211,9 +211,9 @@ public interface ISpriteComponent : IComponent, IAnimationProperties
         /// </summary>
         RSI? LayerGetActualRSI(object layerKey);
 
-        ISpriteLayer this[int layer] { get; }
-        ISpriteLayer this[Index layer] { get; }
-        ISpriteLayer this[object layerKey] { get; }
+        SpriteComponent.Layer this[int layer] { get; }
```
PJB3005, line 214: You should probably just obsolete this interface if you're going to do this.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1322,6 +1354,125 @@ internal void Render(DrawingHandleWorld drawingHandle, Angle eyeRotation, Angle
...
+        ///     If true, the layer draw-depths have changed and need to be updated before this sprite gets rendered.
+        /// </summary>
+        internal bool LayerOrderDirty = true;
+
+        /// <summary>
+        ///     If true, this sprite has at least one layer that has a direction dependent draw depth (e.g., a player's in-hand layers).
+        /// </summary>
+        private bool HasDirectionalSorting => _comparer.HasDirectionalSorting;
+
+        private Dictionary<RSIDirection, Layer[]> _sortedLayers = new(1);
```
PJB3005: Store this in a `Layer[]?[8]` instead and index the array with the `(int)` values of the enum. Less complex and still takes less memory than the overhead of dictionary.

Robust.Client/GameObjects/Components/Renderable/ISpriteComponent.cs
ElectroJr: I added an [Obsolete]. After this & the other sprite/rendering PR gets merged I'm planning to work on just removing it & make sprite comp client-only.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
ElectroJr: Done

ElectroJr:
I'll get back to this sometime after rejigging sprite component logic, which in turn first requires a server sprite component purge.

PR #2971, opened 2022-06-21
Allow sprites to override container occlusion

mirrorcult:
Needed for content to allow ghosts to not be occluded

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -211,6 +211,7 @@ namespace Robust.Client.GameObjects
         ///     Should this entity show up in containers regardless of whether the container can show contents?
         /// </summary>
         [DataField("overrideContainerOcclusion")]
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -207,13 +207,20 @@ private List<PrototypeLayerData> LayerDatums
...
+        public bool OverrideContainerOcclusion;
+
         [ViewVariables(VVAccess.ReadWrite)]
         public bool ContainerOccluded
         {
             get => _containerOccluded;
             set
             {
                 if (_containerOccluded == value) return;
+                if (OverrideContainerOcclusion) return;
```
ElectroJr: IMO it makes more sense to override the getter than the setter. Otherwise if someone wanted to temporarily enable the override, they'd have to redo container occlusion checks afterwards, in case it should have changed.

mirrorcult: sure

PR #2973, opened 2022-06-21
Add enum and custom hash set serializers

ElectroJr:
This adds two new custom type serializers.

Firstly, `EnumStringSerializer` which attempts to map strings onto enums via `IReflectionManager.TryParseEnumReference`. This is useful for things like sprite layers & appearance component keys, which can be either strings or enums. Currently anything that reads these from yaml has to manually call the reflection manager & has no yaml linting.

Secondly, this adds a variant of the hash set serializer that uses another custom type serializer to serialize the actual values. There should really also be list & dictionary variants of these, but initially this is just hash sets. This avoids the need to generate set/list/dict variants of each custom serializer, like is currently done for prototype Ids. So for example the `PrototypeIdHashSetSerializer` could now just be replaced with a `CustomHashSetSerializer<string, PrototypeIdSerializer<DamageGroupPrototype>>`

### DIFF
--- a/Robust.Shared/Serialization/Manager/SerializationManager.cs
+++ b/Robust.Shared/Serialization/Manager/SerializationManager.cs
@@ -325,7 +325,7 @@ namespace Robust.Shared.Serialization.Manager
         {
             var underlyingType = Nullable.GetUnderlyingType(type) ?? type;
 
-            if (value == null) return new MappingDataNode();
+            if (value == null) return new ValueDataNode("null");
 
             if (underlyingType.IsEnum)
             {
--- a/Robust.Shared/Serialization/Manager/SerializationManager.cs
+++ b/Robust.Shared/Serialization/Manager/SerializationManager.cs
@@ -29,6 +29,8 @@ namespace Robust.Shared.Serialization.Manager
     {
         [IoC.Dependency] private readonly IReflectionManager _reflectionManager = default!;
 
+        public IReflectionManager ReflectionManager => _reflectionManager;
+
         public const string LogCategory = "serialization";
 
         private bool _initializing;
--- a/Robust.Shared/Serialization/Manager/SerializationManager.cs
+++ b/Robust.Shared/Serialization/Manager/SerializationManager.cs
@@ -29,8 +29,6 @@ namespace Robust.Shared.Serialization.Manager
     {
         [IoC.Dependency] private readonly IReflectionManager _reflectionManager = default!;
 
-        public IReflectionManager ReflectionManager => _reflectionManager;
-
         public const string LogCategory = "serialization";
 
         private bool _initializing;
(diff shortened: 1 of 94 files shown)


Robust.Shared/Serialization/Manager/SerializationManager.cs
```diff
@@ -29,6 +29,8 @@ public sealed partial class SerializationManager : ISerializationManager
     {
         [IoC.Dependency] private readonly IReflectionManager _reflectionManager = default!;
 
+        public IReflectionManager ReflectionManager => _reflectionManager;
```
PaulRitter, line 32: kinda sus on this, i'd rather you use the dependencycollection provided to the serializer, although this saves some resolves ofc

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/EnumStringSerializer.cs
```diff
@@ -0,0 +1,55 @@
...
+using Robust.Shared.Serialization.Markdown.Value;
+using Robust.Shared.Serialization.TypeSerializers.Interfaces;
+
+namespace Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
+
+/// <summary>
+///     Attempts to resolve a string into an enum. If it fails, it simply reads the string. Useful for both both sprite
+///     layer and appearance data key, which both simultaneously support enums and strings. 
+/// </summary>
+public sealed class EnumStringSerializer : ITypeSerializer<object, ValueDataNode>
```
PaulRitter: either make this return a enum or throw, dont return the string. for special cases like the layer and appearance data key, create a seperate serializer

PaulRitter (changes requested):
either make the enumstringserializer truly generic, making it only parse enums, or create a serializer thats explicitely for layer and appearance data keys. this will just confuse ppl

ElectroJr:
> either make the enumstringserializer truly generic, making it only parse enums, or create a serializer thats explicitely for layer and appearance data keys. this will just confuse ppl

Making it only parse enums would render it useless for appearance & sprite key datafields, so I guess I'll just rename it to make its purpose more explicit. I'll rename it to `AppearanceKeySerializer`, given that sprite keys can technically be any object (though I'll probably change this eventually).

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/EnumStringSerializer.cs
ElectroJr: I'll just explicitly make this a more specialised serializer

Robust.Shared/Serialization/Manager/SerializationManager.cs
ElectroJr: Would there be an easy way to somehow have the serialization manager automatically use the collection to perform some common resolves (prototype, reflection), and somehow pass that along so that they don't need to get re-resolved for every read?

PaulRitter: not as of now, sadly. but feel free to open an issue, this could improve perf a bit

ElectroJr: Actually, I just realised the write-function doesn't even get a dependency collection passed into it, so either way it would need to access ISerializationManager separately

PR #2974, opened 2022-06-22
Misc Shader changes

ElectroJr:
This makes various changes to shaders:
- Clear functions can now specify whether to clear the stencil buffer & the stencil value.
- Sprite post shaders can receive the screen texture & raise events prior to drawing 
- Shader parameters only get updated if they've been dirtied.


PJB3005 (changes requested):
> * Setting an entity's post shader now needs to go through the sprite system and raises a cancellable event.
>   * Can be used to prevents interaction outline component from just overriding a shader.

This is not the correct way to fix this. Either put in a workaround in content or allow the engine to render multiple post shaders, but NOT this.

ElectroJr:
The shader setting changes have been reverted

PJB3005 (approved):

PR #2976, opened 2022-06-23
Fix hotreload

wrexbe:
Puts AccessPermissions, and AccessAttribute under a different namespace when compiled on Robust.Analysers. This prevents hot reload from getting confused because the same classes appear in two different dlls. Hot reload is still buggy with rider, but it works some of the time after this change. I have better luck using hot reload with run, rather then debug.


metalgearsloth (approved):

PaulRitter (approved):

PJB3005:
This should not be behind a `NETSTANDARD2_0` switch, there should be a separate define like `ROBUST_ANALYZER_IMPL`.

PR #2980, opened 2022-06-24
More `CompletionOption` helpers

mirrorcult:
Some ones I thought would be generic/useful while working on space-wizards/space-station-14#9114

### DIFF
--- a/Robust.Shared/Console/CompletionHelper.cs
+++ b/Robust.Shared/Console/CompletionHelper.cs
@@ -1,6 +1,10 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using JetBrains.Annotations;
 using Robust.Shared.ContentPack;
+using Robust.Shared.IoC;
+using Robust.Shared.Players;
+using Robust.Shared.Prototypes;
 using Robust.Shared.Utility;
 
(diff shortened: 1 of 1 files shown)


Robust.Shared/Console/CompletionHelper.cs
```diff
@@ -63,4 +74,30 @@ public static IEnumerable<CompletionOption> UserFilePath(string arg, IWritableDi
...
+        var protoOptions = proto.EnumeratePrototypes<T>().Select(p => new CompletionOption(p.ID));
+        return sorted ? protoOptions.OrderBy(o => o.Value) : protoOptions;
+    }
+
+    /// <summary>
+    ///     Returns a list of connected session names.
+    /// </summary>
+    public static IEnumerable<CompletionOption> SessionNames(bool sorted = true, ISharedPlayerManager? players = null)
+    {
+        IoCManager.Resolve(ref players);
```
mirrorcult, line 98: idk the structure of playermanager session storage but does this break(/not show all sessions) if called on the client or something? since this isnt using `ServerSessions`

mirrorcult: looks like it works how i expect

PR #2986, opened 2022-06-27
Fix client transform anchoring

### LINKED ISSUE #8955: IconSmooth on grid load issue
If you load in a grid next to you the icon smoothing doesn't work properly.

For me I just use `callemergencyshuttle on the evac shuttle PR and you can see the back half doesn't smooth until you go out of range and go back in range again.

metalgearsloth:
Installment 435 on transform causing sloth physical pain:

So the client in some instances (where a grid is re-parented) unanchors entities. The actual problem is:
1. Client gets transform states for everything. Move events are deferred
2. Client runs startup on particular entity. This unanchors the entity because it doesn't find a grid at that spot (as the moveevent is deferred).
3. Moveevent is now run but too late client has unanchored stuff.

I'd prefer not to defer moveevents but it is what it is for now. I removed the broadcast as only pvssystem subscribes to it anyway (and via directed). Bandaids / fixes https://github.com/space-wizards/space-station-14/issues/8955

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
@@ -3,6 +3,7 @@ using System.Runtime.CompilerServices;
 using JetBrains.Annotations;
 using Robust.Shared.GameStates;
 using Robust.Shared.IoC;
+using Robust.Shared.Log;
 using Robust.Shared.Map;
 using Robust.Shared.Maths;
 using Robust.Shared.Physics;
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -220,7 +221,7 @@ static MapId FindMapIdAndSet(TransformComponent xform, IEntityManager entMan, En
     private void OnCompStartup(EntityUid uid, TransformComponent component, ComponentStartup args)
     {
         // Re-Anchor the entity if needed.
-        if (component._anchored && _mapManager.TryFindGridAt(component.MapPosition, out var grid))
+        if (component._anchored && _mapManager.IsGrid(component.ParentUid) && _mapManager.TryGetGrid(component.GridUid, out var grid))
```
ElectroJr, line 224: `IsGrid()` should probably just be replaced with `component.Parentuid == component.GridUid`

metalgearsloth:
Fixing the tests is probably going to be as painful as not deferring moveevents.

PR #2991, opened 2022-06-28
Avoid resetting mass

wrexbe:
Bug: When something gets anchored, it becomes static, and the cached mass gets set to 0, and if that thing has a temperature component, it starts getting very hot.

Hack: If something goes from Non-Static to Static, we don't set the mass to 0, so it keeps the mass from before it went static.
I don't see a way the non-zero value could be seen, except on FixturesMass where it's supposed to be exposed.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Collidable/PhysicsComponent.Physics.cs
+++ b/Robust.Shared/GameObjects/Components/Collidable/PhysicsComponent.Physics.cs
@@ -735,7 +735,6 @@ namespace Robust.Shared.GameObjects
 
         public void ResetMassData(FixturesComponent? fixtures = null)
         {
-            _mass = 0.0f;
             _invMass = 0.0f;
             _inertia = 0.0f;
             InvI = 0.0f;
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Components/Collidable/PhysicsComponent.Physics.cs
```diff
@@ -746,10 +745,13 @@ public void ResetMassData(FixturesComponent? fixtures = null)
...
             }
 
+            // We need the mass to not get reset, unless we plan on recalculating it
+            // So it can be used in temperature calculations
+            _mass = 0.0f;
+
             // Temporary until ECS don't @ me.
             fixtures ??= IoCManager.Resolve<IEntityManager>().GetComponent<FixturesComponent>(Owner);
             var localCenter = Vector2.Zero;
-            var shapeManager = _sysMan.GetEntitySystem<FixtureSystem>();
```
metalgearsloth, line 752: Aren't you better off moving the bodytype check below it if you really need the mass? Coz if it's a static bodytype and it updates its mass then it won't get updated.

PR #3004, opened 2022-06-29
Moves fixture check in physics

keronshb:
client side PR that needs this: https://github.com/space-wizards/space-station-14/pull/8584

Reorders the fixture check so that it fixes an infinite temperature bug if a player was set to anchored.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Collidable/PhysicsComponent.Physics.cs
+++ b/Robust.Shared/GameObjects/Components/Collidable/PhysicsComponent.Physics.cs
@@ -762,7 +762,7 @@ namespace Robust.Shared.GameObjects
                 localCenter += data.Center * data.Mass;
                 _inertia += data.I;
             }
-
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Components/Collidable/PhysicsComponent.Physics.cs
```diff
@@ -768,6 +763,11 @@ public void ResetMassData(FixturesComponent? fixtures = null)
                 _inertia += data.I;
             }
 
+            if (((int) _bodyType & (int) (BodyType.Kinematic | BodyType.Static)) != 0)
```
metalgearsloth:
```suggestion
            }
			// Update this after re-calculating mass as content may want to use the sum of fixture masses instead.
            if (((int) _bodyType & (int) (BodyType.Kinematic | BodyType.Static)) != 0)
```

PR #3008, opened 2022-07-01
Make recursive delete detach parents before children

### LINKED ISSUE #2591: Attaching an entity to another that's currently being terminated doesn't delete it.
Example of this happening in content:
1. Map starts being terminated, recursive delete kicks in.
2. An entity with "followers" starts being terminated.
3. The entity's followers are attached to the map correctly.
4. Because RecursiveDelete doesn't account for new entities being attached to the entity being deleted (it uses a `ToArray` on its children) it doesn't delete the follower entities. The map is deleted successfully. 
5. The follower entities are now in an invalid state, parented to an entity that has been deleted. Because of many assumptions across the codebase, attempting to delete them will throw one of a few different possible exceptions.

See https://github.com/space-wizards/space-station-14/pull/7048 for a reproduction. Use without https://github.com/space-wizards/RobustToolbox/pull/2592 to trigger.

ElectroJr:
Previously recursive delete would:
- Raise a terminate event
- Recursively delete children
- Shutdown components
- Detach to null
- Dispose components
- Delete the entity

This PR changes the ordering so that entities get detached to null after raising the termination event, but before iterating through children.

The main motivator for this was that termination would cause the entity lookup system to recurse up the transform hierarchy to find a parent with a look-up component, and it would do this for every deleted entity, rather than just directly recursively removing the parent.

But this is also a pretty significant change that might cause unexpected issues when deleting something like a grid or map. AFAICT  thing still seem to work fine, round restart works, etc. 

Requires https://github.com/space-wizards/space-station-14/pull/9316

Also ran into several issues related to #2591.
I've added extra logging, and when something tries to parent to a terminating entity it should now try to delete that entity directly if exception tolerance is enabled. So I guess:
Fixes #2591 (really, whatever system was responsible for a re-parent-on-delete should just be more careful in the first place).

### DIFF
--- a/Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
@@ -73,8 +73,6 @@ namespace Robust.Shared.GameObjects
             SubscribeLocalEvent<EntityLookupComponent, ComponentShutdown>(OnLookupShutdown);
             SubscribeLocalEvent<GridInitializeEvent>(OnGridInit);
 
-            SubscribeLocalEvent<EntityTerminatingEvent>(OnTerminate);
-
             EntityManager.EntityInitialized += OnEntityInit;
             SubscribeLocalEvent<MapChangedEvent>(OnMapCreated);
         }
@@ -198,12 +196,6 @@ namespace Robust.Shared.GameObjects
         #endregion
 
         #region Entity events
-
-        private void OnTerminate(ref EntityTerminatingEvent args)
-        {
-            RemoveFromEntityTree(args.Owner, false);
-        }
-
         private void OnEntityInit(EntityUid uid)
         {
             if (_container.IsEntityInContainer(uid)) return;
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -221,6 +221,8 @@ namespace Robust.Shared.GameObjects
                 newMap?.AddAwakeBody(body);
                 DebugTools.Assert(body.Awake);
             }
+            else
+                DebugTools.Assert(oldMap?.AwakeBodies.Contains(body) != true);
 
             if (fixturesQuery.TryGetComponent(uid, out var fixtures) && body._canCollide)
             {
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -183,7 +183,9 @@ namespace Robust.Shared.GameObjects
 
             _broadphase.TryGetMoveBuffer(oldMapId, out var oldMoveBuffer);
 
-            RecursiveMapUpdate(xform, body, newMapId, null, newMap, oldMap, oldMoveBuffer, bodyQuery, xformQuery, fixturesQuery, jointQuery, broadQuery);
+            var newBroadphase = _broadphase.GetBroadphase(xform, broadQuery, xformQuery);
+
+            RecursiveMapUpdate(xform, body, newMapId, newBroadphase, newMap, oldMap, oldMoveBuffer, bodyQuery, xformQuery, fixturesQuery, jointQuery, broadQuery);
         }
 
         /// <summary>
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
@@ -71,7 +71,7 @@ namespace Robust.Shared.GameObjects
             SubscribeLocalEvent<PhysicsSleepEvent>(OnSleep);
             SubscribeLocalEvent<CollisionChangeEvent>(OnCollisionChange);
             SubscribeLocalEvent<PhysicsComponent, EntGotRemovedFromContainerMessage>(HandleContainerRemoved);
-            SubscribeLocalEvent<PhysicsComponent, EntParentChangedMessage>(OnParentChange);
+            SubscribeLocalEvent<EntParentChangedMessage>(OnParentChange);
             SubscribeLocalEvent<SharedPhysicsMapComponent, ComponentInit>(HandlePhysicsMapInit);
             SubscribeLocalEvent<SharedPhysicsMapComponent, ComponentRemove>(HandlePhysicsMapRemove);
             SubscribeLocalEvent<PhysicsComponent, ComponentInit>(OnPhysicsInit);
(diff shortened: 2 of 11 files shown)


Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -73,8 +73,6 @@ public override void Initialize()
             SubscribeLocalEvent<EntityLookupComponent, ComponentShutdown>(OnLookupShutdown);
             SubscribeLocalEvent<GridInitializeEvent>(OnGridInit);
 
-            SubscribeLocalEvent<EntityTerminatingEvent>(OnTerminate);
```
ElectroJr, line 76: Now just handled by parent change / detach to null

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -139,6 +139,10 @@ private void OnParentChange(EntityUid uid, PhysicsComponent body, ref EntParentC
 
             var xform = args.Transform;
 
+            // Is this entity getting recursively detached after it's parent was already detached to null?
+            if (args.OldMapId == MapId.Nullspace && xform.MapID == MapId.Nullspace)
+                return;
```
ElectroJr, line 147: If an entity gets moved to null-space, it should now recursively remove any of its children from the physics guff.
So if an entity is already in null-space, it can skip getting components, queries, etc.

Robust.Shared/Map/MapManager.GridTrees.cs
```diff
@@ -143,25 +143,24 @@ private void OnGridParentChange(EntityUid uid, MapGridComponent component, ref E
         var lifestage = EntityManager.GetComponent<MetaDataComponent>(uid).EntityLifeStage;
 
         // oh boy
-        // Want gridinit / gridremoval to handle this hence specialcase those situations.
-        if (lifestage is < EntityLifeStage.Initialized or >= EntityLifeStage.Terminating) return;
```
ElectroJr, line 147: Grid removal now happens in null-space, where it no longer has info about what map it came from. So now letting the parent-change also handle the grid tree update for deleting grids

Robust.Shared/Physics/SharedBroadphaseSystem.cs
```diff
@@ -394,10 +394,11 @@ internal void FindNewContacts(MapId mapId)
         /// <summary>
         /// If our broadphase has changed then remove us from our old one and add to our new one.
         /// </summary>
-        internal void UpdateBroadphase(PhysicsComponent body, FixturesComponent? manager = null, TransformComponent? xform = null)
+        internal void UpdateBroadphase(PhysicsComponent body, MapId oldMapId, FixturesComponent? manager = null, TransformComponent? xform = null)
         {
             if (!Resolve(body.Owner, ref manager, ref xform) ||
-                _mapManager.IsGrid(body.Owner)) return;
+                _mapManager.IsGrid(body.Owner) && xform.MapID != MapId.Nullspace)
+                return;
```
ElectroJr: If a grid gets moved to null-space, we still want to iterate through its children.
Why is this actually usually disabled for grids?

Robust.Shared/Physics/SharedBroadphaseSystem.cs
```diff
@@ -439,7 +440,9 @@ private void UpdateBodyBroadphase(PhysicsComponent body, FixturesComponent manag
 
                 if (oldBroadphase == newBroadphase) return;
 
-                DestroyProxies(body, manager);
+                DebugTools.Assert(oldMapId != MapId.Nullspace);
```
metalgearsloth: This assert triggers for me when stuff comes out of pvs.

metalgearsloth:
Also I seem to get the re-parenting issue more frequently as when I deleted Saltern when following on it 3 deleted bodies made it into physics.

ShadowCommander:
Since tests are faster now, it'd be nice to have tests written for more critical things like this. I can try to work on some, but I'll need to read up on how all the deletion code works.

metalgearsloth:
I've been writing some for entitylookup, broadphase, and physicsmap.

Robust.Shared/Physics/SharedBroadphaseSystem.cs
ElectroJr: Oh, this assert was originally for when I was trying to debug tests failing, this shouldn't even be here anymore

ElectroJr:
> Also I seem to get the re-parenting issue more frequently as when I deleted Saltern when following on it 3 deleted bodies made it into physics.

Ah, followers: the one instance where deleting an entity is not meant to delete all children. That should be fixed now, by making sure termination events get raised before detaching to null.

Also made some other misc physics changes including combining the recursive broadphase & physics map changes, and making the recursive broadphase update not call GetBroadphase for each child (which iterates back up through parents), and instead track the broadphase while going down through children. Also adds some more component queries (broadphase & joint)

Things seem to be working fine, and the new tests seem happy.

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -151,97 +155,92 @@ private void OnParentChange(EntityUid uid, PhysicsComponent body, ref EntParentC
...
-                {
-                    DebugTools.Assert(oldMap?.AwakeBodies.Contains(childBody) != true);
-                }
-
-                _broadphase.RemoveFromMoveBuffer(childBody, oldMapId);
-                DestroyContacts(childBody, oldMapId);
-                _joints.ClearJoints(childBody);
-                RecursiveMapUpdate(childXform, oldMapId, oldMap, map, xformQuery, bodyQuery, metaQuery);
+                if (xformQuery.TryGetComponent(child, out var childXform) && bodyQuery.TryGetComponent(child, out var childBody))
+                    RecursiveMapUpdate(childXform, childBody, newMapId, newBroadphase, newMap, oldMap, oldMoveBuffer, bodyQuery, xformQuery, fixturesQuery, jointQuery, broadQuery);
```
metalgearsloth: Thinking on it this might still need to iterate if there's a physics entity parented to something that doesn't have physicscomponent?

Robust.Shared/Physics/SharedJointSystem.cs
```diff
@@ -381,6 +381,11 @@ public void ClearJoints(PhysicsComponent body)
             _dirtyJoints.Add(joint);
         }
 
+        public void ClearJoints(JointComponent joint)
+        {
+            _dirtyJoints.Add(joint);
+        }
```
metalgearsloth: I realised this doesn't actually clear said joints as I think it got destroyed in some merge.

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
ElectroJr: Both map & broadphase updating should now iterate through children, even if they don't have physics

metalgearsloth (approved):

PR #3035, opened 2022-07-10
Allow ignore missing to specify postfix

wrexbe:
Ex: IgnoreMissingComponents("Visuals")
So Content.Server doesn't have to add them to a manual ignore list.

### DIFF
--- a/Robust.Shared/GameObjects/ComponentFactory.cs
+++ b/Robust.Shared/GameObjects/ComponentFactory.cs
@@ -14,7 +14,6 @@ namespace Robust.Shared.GameObjects
     [Virtual]
     internal class ComponentFactory : IComponentFactory
     {
-        private bool _ignoreMissingComponents;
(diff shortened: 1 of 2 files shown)


ElectroJr (approved):

Robust.Shared/GameObjects/ComponentFactory.cs
```diff
@@ -198,9 +203,13 @@ private void RegisterReference(Type target, Type @interface)
             ComponentReferenceAdded?.Invoke(registration, idx);
         }
 
-        public void IgnoreMissingComponents()
+        public void IgnoreMissingComponents(string? postfix = "")
         {
```
mirrorcult: why have the param be nullable at all if it just throws?

wrexbe: I decided to make it not null while I was implementing it, but forgot to remove the question mark.

PR #3038, opened 2022-07-12
AudioSystem Rejig

ElectroJr:
This PR moves sound specifiers from content into robust (along with the serializer and sound collection prototype).
This is partly to make it easier to specify audio params for sounds via yaml,  which most content systems don't bother to implement, and partly to get rid of all the boiler place of having to resolve `IRobustRandom`, `IPrototypeManager`, and then calling `specifier.GetSound(_random, _prototypeManager)`. While this could also be achieved by having a content specific audio system or some entity system helper functions, nothing about this is unique to SS14 and AFAIK would be useful for anything using the engine.

General changes:
- Removes `IAudioSystem`
- Makes the static `SoundSystem` directly resolve & call functions in `AudioSystem`
- Marks functions in `SoundSystem` as obsolete.
- `AudioParams` now have a `variation` datafield, which should be used in place of `AudioHelpers.WithVariation()`

Audio system specific changes:
- One of the `Play()` functions is now called `PlayGlobal()` to avoid accidental global audio.
- Added a `PlayPvs()` variant, which removes the need to constantly add `Filter.Pvs(uid, entityManager: EntityManager)`.
- Added `PlayPredicted()`. This just plays audio normally for local clients, while for servers it will sends the audio to all players in pvs except for some "user" entity. 
  - This is very commonly required for interaction or foot-step sound prediction. 
- Added overrides that takes in a sound specifier.
  - Removes the need to constantly call `specifier.GetSound()` & automatically uses the specifier's audio params as a default, which is frequently forgotten about.
- Reduces transform component fetching & world-position calculations in some play functions. 

Requires space-wizards/space-station-14/pull/9635


metalgearsloth:
I think it'd be more consistent having soundspecifier in engine considering content uses it so extensively and there's no reason other games couldn't take advantage of it (even if they make their own specifiers).

metalgearsloth (approved):
@PJB3005 soundspecifier in engine yeah/nah

PR #3041, opened 2022-07-12
Make typereader stabler

### LINKED ISSUE #3039: Esoteric serv3 issue
https://github.com/space-wizards/space-station-14/runs/7304111972?check_suite_focus=true

Randomly get this on starting up the game.

@PaulRitter

wrexbe:
The problem seems to be that the dictionary has multiple things in it that matched, and since dictionaries don't promise the order, sometimes something is picked that doesn't work. Adding the generic argument check seems to prevent duplicate matches, so I think this resolves it.

Fixes https://github.com/space-wizards/RobustToolbox/issues/3039


PaulRitter (changes requested):
this is the wrong way to go about this. on how to solve this, see #3039

PR #3045, opened 2022-07-14
Dirty PVS chunk when visibility mask changes

ElectroJr:
- Adds an internal `MarkDirty(EntityUid)`  to pvs system
- Makes VisibilityMask a friend of the metadata system, as it should be set via a new function in that system.
- Makes that function dirty the pvs chunk.
  - Requires a server & client version of the meta-data system.
- Minor unrelated changes to PVS's `OnParentChange` (less component fetching/checking).

Required so that an entity that has its visibility mask changed gets added to / or removed from a client's view without needing to wait for entities to move around in the pvs chunk. 

Edit:
So adding PVS system as a dependency to MetadataSystem made unit tests dependency stuff unhappy. I fixed it, but maybe not how I should've done.

### DIFF
--- a/Robust.Server/GameObjects/EntitySystems/VisibilitySystem.cs
+++ b/Robust.Server/GameObjects/EntitySystems/VisibilitySystem.cs
@@ -59,16 +59,8 @@ namespace Robust.Server.GameObjects
 
         public void RefreshVisibility(EntityUid uid, MetaDataComponent? metaDataComponent = null, VisibilityComponent? visibilityComponent = null)
         {
-            if (!Resolve(uid, ref metaDataComponent, false))
-            {
-                // This means it's deleting or some shit; I'd love to make it a GetComponent<T> in future.
-                return;
-            }
-
-            var visMask = 1;
-            GetVisibilityMask(uid, ref visMask, visibilityComponent);
-
-            _metaSys.SetVisibilityMask(uid, visMask, metaDataComponent);
+            if (Resolve(uid, ref metaDataComponent, false))
+                _metaSys.SetVisibilityMask(uid, GetVisibilityMask(uid, visibilityComponent), metaDataComponent);
         }
 
         public void RefreshVisibility(VisibilityComponent visibilityComponent)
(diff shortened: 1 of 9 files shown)


Robust.Server/GameStates/PVSSystem.Dirty.cs
```diff
@@ -44,7 +44,8 @@ private void ShutdownDirty()
 
         private void OnEntityAdd(EntityUid e)
         {
-            DebugTools.Assert(_currentIndex == _gameTiming.CurTick.Value % DirtyBufferSize);
+            DebugTools.Assert(_currentIndex == _gameTiming.CurTick.Value % DirtyBufferSize ||
+                _gameTiming.GetType().Name == "IGameTimingProxy");// Look I have NFI how best to excuse this assert if the game timing isn't real (a Mock<IGameTiming>).
```
ElectroJr, line 48: I have NFI how to properly disable this assert during unit tests, without just removing it or doing this.

metalgearsloth (approved):
@PaulRitter

Robust.Server/GameStates/PVSSystem.Dirty.cs
metalgearsloth: Can probably remove the assert; ideally it'd just be a test instead. I think this only matters for pvs off but that's been stable for a while now?

Robust.Server/GameObjects/EntitySystems/VisibilitySystem.cs
```diff
@@ -63,12 +66,14 @@ public void RefreshVisibility(EntityUid uid, MetaDataComponent? metaDataComponen
...
+            _metaSys.SetVisibilityMask(uid, visMask, metaDataComponent);
         }
 
         public void RefreshVisibility(VisibilityComponent visibilityComponent)
         {
-            RefreshVisibility(visibilityComponent.Owner);
+            RefreshVisibility(visibilityComponent.Owner, null, visibilityComponent);
         }
 
         private int GetVisibilityMask(EntityUid uid, ref int visMask, VisibilityComponent? visibilityComponent = null, TransformComponent? xform = null)
```
PaulRitter: why is this using ref to return the vismask as well as returning it? just return it tbqh

Robust.Server/GameStates/PVSSystem.Dirty.cs
PaulRitter: i'd rather it stay, at least until we have a test for it. this solution sgtm for the shortterm

PaulRitter (approved):

Robust.Server/GameObjects/EntitySystems/VisibilitySystem.cs
ElectroJr: I've changed it to just return it. But I also noticed that the way the logic works is that everything will always have a vismask with 1 set (i.e., `visMask & 1 = 1`), there is no way to avoid having 1.

I updated the logic so that it doesn't do that, but does give an entity a mask of 1 if it has no vis mask component.
I think this won't change anything?

metalgearsloth: It's set to 1 for opendream from memory.

ElectroJr: ok, ill undo that change so that everything always has the first bit/flag set.
also I realize I never pushed the changes described in that comment, but they're there now.

PR #3048, opened 2022-07-14
Don't defer moveevents anymore

metalgearsloth:


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -436,6 +434,12 @@ internal void OnHandleState(EntityUid uid, TransformComponent component, ref Com
                 component.RebuildMatrices();
             }
 
+            if (oldPos != default)
+            {
+                var ev = new MoveEvent(uid, oldPos, component.Coordinates, component, _gameTiming.ApplyingState);
```
ElectroJr: If they're not being deferred anymore, you can just remove the `ApplyingState` argument and have the one subscription that uses it check it directly.

metalgearsloth:
yeet

PR #3051, opened 2022-07-17
Fix linear velocity cap rounding

metalgearsloth:
Here we use it as a hard cap.


ElectroJr:
Why is `PhysicsIsland._maxLinearVelocity` and `IslandCfg.MaxLinearVelocity` even called linear velocity, when its really a max displacement computed from the max velocity? I guess the same also applies to max angular velocity, where the variables should be something like "max angle change squared"?

metalgearsloth:
It's what box2d calls it though in our case it updates whenever tickrate updates.

PR #3054, opened 2022-07-18
Fix cross-grid/map container inserts

ElectroJr:
Container insert events were removing an entity from the lookup tree corresponding to the current-transform/parent, rather than the one it was in prior to insertion.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
@@ -272,9 +272,22 @@ namespace Robust.Shared.GameObjects
         {
             var meta = MetaData(args.Entity);
 
-            // Parent change gets raised after container insert so we'll just drop it and let OnContainerInsert handle.
+            // If our parent is changing due to a container-insert, we let the container insert event handle that. Note
+            // that the in-container flag gets set BEFORE insert parent change, and gets unset before the container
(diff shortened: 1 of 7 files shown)


Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -272,9 +272,22 @@ private void OnParentChange(ref EntParentChangedMessage args)
...
+            // that the in-container flag gets set BEFORE insert parent change, and gets unset before the container
+            // removal parent-change. So if it is set here, this must mean we are getting inserted.
+            //
+            // However, this means that this method will still get run in full on container removal. Additionally,
+            // because not all container removals are guaranteed to result in a parent change, container removal events
+            // also need to add the entity to a tree. So this generally results in:
+            // add-to-tree -> remove-from-tree -> add-to-tree.
+            // Though usually, `oldLookup == newLookup` for the last step. Its still shit though.
+            //
+            // TODO IMPROVE CONTAINER REMOVAL HANDLING
```
ElectroJr, line 285: I think if you changed removal so that it unsets the metadata flag and raises the got-removed events after the (possible) parent change, then this wouldn't be an issue.

But I don't know if that causes issues elsewhere?

metalgearsloth (approved):

PR #3055, opened 2022-07-18
Allow maps to be saved

### LINKED ISSUE #3053: MapLoader can't load maps
I tried fixing it and the amount of bs you gotta fix is pretty high. I tried using SetMapEntityId but that just explodes because none of the component init etc are run on the deserialized map entity.

metalgearsloth:
Fixes https://github.com/space-wizards/RobustToolbox/issues/3053

Need to clean it up coz I was pulling my hair out throwing crap at the wall.


Robust.Server/Maps/MapLoader.cs
```diff
@@ -602,12 +610,22 @@ private static MapGrid AllocateMapGrid(MapGridComponent gridComp, MappingDataNod
...
+
+                if (TargetMapUid != null)
+                {
+                    mapEntity = TargetMapUid.Value;
+                    _mapManager.SetMapEntity(TargetMap, TargetMapUid.Value);
+                }
+                else
+                {
+                    mapEntity = _mapManager.GetMapEntityIdOrThrow(TargetMap);
+                }
```
metalgearsloth, line 623: So we create a map above but we don't really have the data to already attach it to an entity, so we just set it here but unfortunately it needs the kinda hack in SetMapEntity.

moonheart08 (approved):
evil but it works + seems sensible.

PR #3064, opened 2022-07-21
Fix nullspace entity creation error

ElectroJr:


moonheart08:
you deserve a cookie

ElectroJr:
Actually I think I just completely misunderstood how null-space was handled Somehow I thought it wasn't a real map/entity? So this isn't how this should be fixed

metalgearsloth:
> Actually I think I just completely misunderstood how null-space was handled Somehow I thought it wasn't a real map/entity? So this isn't how this should be fixed

Nullspace is schrodinger's entity, it may or may not have an entity attached and you encounter this funness when you go digging into map tests.

PR #3065, opened 2022-07-21
Add entitylookup for components

### LINKED ISSUE #2952: Add component queries to entitylookup
i.e. "GetComponentsInRange".

Generally internally it should essentially check the component count for it and determine whether it's probably better to iterate every component or iterate all entities in range checking for the component.

metalgearsloth:
Technically entitylookup is now just "lookup" but changing it would be a pain.

I used the same API as entitylookup and it's kinda boilerplatey but don't know of an easier way to do it. The main advantage is it internally checks if it should just iterate every component directly or check the dynamic trees. This also means content has a consistent way to do component checks even if we change the internal implementation as right now people call the entitylookup methods and use their own entityquery.

Entitylookup is also pretty slow due to the hashsets everywhere but I'm mainly worried about API stability atm and we can optimise later (as PVS doesn't use this anymore it's pretty much just content calling it).

Fixes https://github.com/space-wizards/RobustToolbox/issues/2952


Robust.Shared/GameObjects/Systems/EntityLookupSystem.ComponentQueries.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        var xformQuery = GetEntityQuery<TransformComponent>();
+        var intersecting = new HashSet<T>();
+
+        if (!UseBoundsQuery<T>(worldAABB.Height * worldAABB.Width))
+        {
+            foreach (var (comp, xform) in EntityQuery<T, TransformComponent>(true))
+            {
+                if (xform.MapID != mapId || !worldAABB.Contains(_transform.GetWorldPosition(comp.Owner, xformQuery))) continue;
+                intersecting.Add(comp);
```
ElectroJr: This doesn't respect `LookupFlags.Contained`. If the contained flag is not set, this should get the metadata & check the in-container flag. It should also check xform.Anchored & the anchored flag.

ElectroJr (changes requested):

Robust.Shared/GameObjects/Systems/EntityLookupSystem.ComponentQueries.cs
metalgearsloth: For containers it's similar to the entitylookup side but I added the anchored check so not sure if both should change for that.

ElectroJr (approved):

PR #3072, opened 2022-07-24
Adds a check to HandleContainerRemoved to handle different canCollide states.

Mirino97:
This was needed for [#9357](https://github.com/space-wizards/space-station-14/pull/9357)

The component CollideOnAnchor was acting properly. A collideable item would be impassable when anchored and passable when unanchored. But when you picked it and placed it down again, it would once again become impassable, even if not anchored.

That kinda defeated the purpose of the fix in the mentioned PR.
With this, everything works normally.

(10 mb only for videos? Srsly github?)
https://youtu.be/qVtJ-PX96Dk


Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -302,7 +301,12 @@ private void HandleContainerRemoved(EntityUid uid, PhysicsComponent physics, Ent
             // If entity being deleted then the parent change will already be handled elsewhere and we don't want to re-add it to the map.
             if (MetaData(uid).EntityLifeStage >= EntityLifeStage.Terminating) return;
 
-            SetCanCollide(physics, true, false);
+            bool canCollide = true;
+
+            if (TryComp(uid, out PhysicsComponent? physicsComponent))
+                canCollide = physicsComponent._canCollide;
```
ElectroJr: This is just fetching the phyics component again? I.e. `physics == physicsComponent`.

AFAIK the point of this was to stop container removal from auto-enabling collisions for entities that should only collide if they are anchored, so just check for the `CollideOnAnchorComponent`. If it exists and is enabled, then return early & don't enable collosions.

Mirino97: I think I have correctly addressed your request!
Please take another look at your earliest convenience!

PR #3073, opened 2022-07-24
More window open overrides

ElectroJr:
- Adds a generic function that attempts to place the center of a window at some relative position.
- Makes `OpenCentered()` & `OpenToLeft()` use this function.
- Adds `OpenCenteredLeft()`, `OpenCenteredRight()`, and `OpenToRight()`.

### DIFF
--- a/Robust.Client/UserInterface/CustomControls/BaseWindow.cs
+++ b/Robust.Client/UserInterface/CustomControls/BaseWindow.cs
@@ -227,39 +227,50 @@ namespace Robust.Client.UserInterface.CustomControls
             Opened();
         }
 
-        public void OpenCentered()
+        public void OpenCentered() => OpenCentered((0.5f, 0.5f));
+        public void OpenToLeft() => OpenCentered((0, 0.5f));
+        public void OpenCenteredLeft() => OpenCentered((0.25f, 0.5f));
+        public void OpenToRight() => OpenCentered((1, 0.5f));
+        public void OpenCenteredRight() => OpenCentered((0.75f, 0.5f));
+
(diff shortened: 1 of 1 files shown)


Mirino97:
Very, very based.

wrexbe (approved):

Robust.Client/UserInterface/CustomControls/BaseWindow.cs
```diff
@@ -227,39 +227,50 @@ public void Open()
...
+        public void OpenCenteredLeft() => OpenCentered((0.25f, 0.5f));
+        public void OpenToRight() => OpenCentered((1, 0.5f));
+        public void OpenCenteredRight() => OpenCentered((0.75f, 0.5f));
+
+        /// <summary>
+        ///     Opens a window, attempting to place the center of the window at some relative point on the screen.
+        /// </summary>
+        /// <param name="relativePosition">Fractional screen position. So (0,0) is the upper left, and (1,1) is the
+        /// lower right.</param>
+        public void OpenCentered(Vector2 relativePosition)
```
mirrorcult: to me 'opencentered' taking a vector2 implies that this would be offset from the center of the screen, not like just fractional screen position. so this should probably just be called `OpenAt` or `Open`

ElectroJr: which is better: `OpenAt` or `OpenCenteredAt`

wrexbe: OpenCenteredAt

PR #3075, opened 2022-07-25
Validate sprite specifier states

ElectroJr:
Currently sprite specifiers only check if an RSI's meta.json exists, but not whether the given state exists. This PR makes the server 
validation look for the required .png file, while the client actually tries to load the RSI state.

I'm not 100% if this is the right way to go about creating server & client variations of a type serializer, but it seems to work.

Requires space-wizards/space-station-14/pull/10015 (yaml linter will start failing)

### DIFF
--- a/Robust.Client/Serialization/ClientSpriteSpecifierSerializer.cs
+++ b/Robust.Client/Serialization/ClientSpriteSpecifierSerializer.cs
@@ -1,0 +1,45 @@
+using Robust.Client.ResourceManagement;
+using Robust.Shared.GameObjects;
+using Robust.Shared.IoC;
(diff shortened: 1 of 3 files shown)


metalgearsloth:
@PaulRitter

PaulRitter (approved):
should be fine, maybe also check if a meta.json exists on server

ElectroJr:
> should be fine, maybe also check if a meta.json exists on server

The ResourcePathSerializer looks for things ending in `.rsi` and then automatically looks for the meta.json instead, so it already did & still does that.

PaulRitter (approved):

PR #3080, opened 2022-07-26
Timed Event System

wrexbe:
Proof of concept

Todo:

- [x] Pausing
- [ ] Serializing

### DIFF
--- a/Robust.Shared/Timing/TimedEventSystem.cs
+++ b/Robust.Shared/Timing/TimedEventSystem.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using JetBrains.Annotations;
 using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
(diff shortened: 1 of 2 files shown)


metalgearsloth:
what about pausing

wrexbe:
todo I guess

Robust.Shared/Timing/TimedEventSystem.cs
```diff
@@ -0,0 +1,47 @@
...
+using JetBrains.Annotations;
+using Robust.Shared.GameObjects;
+using Robust.Shared.IoC;
+
+namespace Robust.Shared.Timing;
+public sealed class TimedEventSystem : EntitySystem
+{
+    [Dependency] private readonly IGameTiming _gameTiming = default!;
+    private readonly ComponentTimedEvent _componentTimedEventInstance = new();
+    private readonly PriorityQueue<(Component Component, string Key), TimeSpan> _timeQueue = new ();
```
gradientvera, line 13: You shouldn't store references to components.
And also, this wouldn't be serializable. Storing data in systems is bad, data should be in components instead.

gradientvera (commented):
This should really just be a component on entities instead...

wrexbe:
> This should really just be a component on entities instead...

It can't really work per component. The advantage of this system is how it is orders of magnitude faster then what systems are currently doing. Most systems that are keeping track of time are looping through each component, incrementing a value, and checking if a time elapsed. That method has several disadvantages. It is O(N) on speed, and memory, and it adds a lot of boiler plate code to systems. This method in contrast will use almost no resources for bookkeeping each update, regardless of the number of components that exist, and is very easy to add to a system. Many systems will be able to switch from checking every update to only running on events.

>This wouldn't be serializable

I think it could be changed to be serializable. It would require a little bit of transforming, and it would need to tap into the map serialization.
Would maybe be stored as a component on the map that serialized as a list of (Id, ComponentType, Key, Duration).

>You shouldn't store references to components.

There are advantages to having the component reference:
* They don't need to be looked up.
* If the component gets replaced on the entity, I don't want the timer to fire on a different component.
* I can add a check for the lifecycle of the component later.

>Storing data in systems is bad

I could make it into a manager class, but I'm not sure that would improve the code.

wrexbe:
>>This wouldn't be serializable

>I think it could be changed to be serializable. It would require a little bit of transforming, and it would need to tap into the map serialization.
Would maybe be stored as a component on the map that serialized as a list of (Id, ComponentType, Key, Duration).

That idea didn't work. Putting components all over everything is going to suck.

metalgearsloth:
(I am still keen)

PR #3083, opened 2022-07-27
Automatic simple component state networking

mirrorcult:
## Why

In content a common pattern emerges, where you have some very simple component state, and the get/handling process entirely just involves getting/setting the fields on the state (which have the same name as the component's fields) as-is, like so:

```csharp
    private void OnComponentGetState(EntityUid uid, IdCardComponent component, ref ComponentGetState args)
    {
        args.State = new IdCardComponentState(component.FullName, component.JobTitle);
    }

    private void OnComponentHandleState(EntityUid uid, IdCardComponent component, ref ComponentHandleState args)
    {
        if (args.Current is IdCardComponentState state)
        {
            component.FullName = state.FullName;
            component.JobTitle = state.JobTitle;
        }
    }
```

This feels terrible and makes it a pain to network basic components, and can also be a little confusing to explain why this is needed at all. All in all should be able to see why reducing the boilerplate here is good

## How

`NetworkedComponentAttribute` optionally takes in a type which inherits from `ComponentState`:
```csharp
[NetworkedComponent(typeof(IdCardComponentState))]
```

The data you want to auto-network is defined on the component state type, and must have the same name as the fields on the component:

```csharp
        // Component fields
		...
        [DataField("fullName")]
        [Access(typeof(SharedIdCardSystem), typeof(SharedPDASystem), typeof(SharedAgentIdCardSystem),
            Other = AccessPermissions.ReadWrite)] // FIXME Friends
        public string? FullName;

        [DataField("jobTitle")]
        public string? JobTitle;
		...

		// Component state fields
		...
		public string? FullName;
        public string? JobTitle;
		...
```

For get-state:
- First, attempt to get the state through a component event & override
- If fail, check if a state type is specified on the attribute and autogenerate
- Create an instance of the new state, then go over all the fields on the state type, match them to fields on the component, & fill them in on the state

For handle-state:
- Attempt to auto-handle state first for easy setting of values
- Check for the specified state type, go over all the fields on it, match to component fields, then fill in fields on the component
- Then, handle state through the component event & override

## Caveats / Misc

- This only works with fields, not properties. Shouldn't be an issue as components should be data-only anyway.
- Component states that want this now require a parameterless ctor. It's possible (and would be nice) to circumvent this and just call the normal ctor instead of `Activator.CreateInstance` (like EFcore can https://docs.microsoft.com/en-us/ef/core/modeling/constructors) but I'll figure it out later I'm lazy.
- There is probably a very minor perf hit, in that handlestate now needs to get the networked attribute regardless of if the auto handling is done at all (getstate can early out 99% of the time). But this is so negligible that its almost certainly fine.
- This will require documentation and a #codebase-changes post since it's perhaps unintuitive. I'm willing to write both of these obviously
- `GetComponentState` was made to return a nullable state instead of an empty component state, and a check in gamestate manager was changed which used to just check for it being the base type which was ugly.

### DIFF
--- a/Robust.Client/GameStates/ClientGameStateManager.cs
+++ b/Robust.Client/GameStates/ClientGameStateManager.cs
@@ -401,6 +401,7 @@ namespace Robust.Client.GameStates
 
                     // TODO: Handle interpolation.
                     var handleState = new ComponentHandleState(compState, null);
+                    TryAutoHandleState(comp, compState);
                     _entities.EventBus.RaiseComponentEvent(comp, ref handleState);
                     comp.HandleComponentState(compState, null);
                     comp.LastModifiedTick = curTick;
@@ -666,6 +667,7 @@ namespace Robust.Client.GameStates
                     try
                     {
                         var handleState = new ComponentHandleState(cur, next);
+                        TryAutoHandleState(component, cur);
                         bus.RaiseComponentEvent(component, ref handleState);
                         component.HandleComponentState(cur, next);
                     }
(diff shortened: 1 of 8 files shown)


mirrorcult:
will probably make a counterpart content PR pruning all basic usages of get/handle state

metalgearsloth:
As on discord gib profile first because pvs is expensive as and checking an attribute + iterating fields is probably expensive.

metalgearsloth:
@PaulRitter 

For reference I used to use 20 pathfinding dummies + 10/20 clients as a rough pvs benchmark

wrexbe:
I like the idea, but I think implementing a C# generator that makes a state class, and state method would be a better way to do it. I'll volunteer to make it, if you want.

mirrorcult:
ye im gonna try source gen cause i dont think it would even be that hard

Issue #3052, opened 2022-07-18
MapLoader needs love


metalgearsloth:
There's pretty much 2 code paths to entity creation in RT: one via maploader deserializing it and another via spawning it ingame which are completely detached from each other where one can break when the other is fine.

I think ideally maploader would build a hierarchy of entities and then spawn them in order rather than initializing everything at once then starting everything at once etc so hopefully we can avoid all of the specific maploader hacks that exist right now.

Issue #3032, opened 2022-07-07
Custom type serializers should have some way of directly accessing IPrototypeManager & IReflectionManager instances.


ElectroJr:
Currently custom type serializers should resolve these using the dependency collection, but this can noticeably slow down entity spawning. Ideally there'd be some way to just have the serialization manager resolve these from the collection and pass them to the serializers. Some examples where this would be useful include the `SpriteSpecifierSerializer` and `AppearanceKeySerializer`.

Also, the writing functions don't even get the dependency collection passed in, so these currently can't even use that to resolve these managers. This was a problem for #2973, where the `AppearanceKeySerializer` currently just directly uses the serialization manager's `IReflectionManager`.

PR #3084, opened 2022-07-27
replaces read with copy in entityspawning and redos alot of the serv3 api

PaulRitter:
requires https://github.com/space-wizards/space-station-14/pull/12308

i expanded the scope of this pr to kinda unfuck alot of the serv3 api, sorry not sorry

tldr:
- entityspawning uses copy, meaning it wont read literal mapping on each entityspawn anymore
- copy is now seperated into copyto and createcopy
- typeserializers no longer needs to implement any copying behaviour (i will definitely not regret this)
- typeserializers, serializationcontext and the manager have been cleaned up a bunch in the background
- validation, copying and a big part of serialization have been rewritten as expression trees
- copying does not box anymore

should partially adress both #3052 and #3032

basically nukes all read calls in entityspawning and uses copy instead. should allow us to speed up entityspawning quite a bit since we only need to do dumb copies instead of expensive deserialization and inheritance pushing logic.


Robust.Server/Maps/MapLoader.cs
```diff
@@ -754,26 +754,38 @@ private void FinishEntitiesLoad()
...
+                            if (prototype?.Components?.TryGetValue(value, out var protData) == true)
+                            {
+                                datanode =
+                                    _serializationManager.PushCompositionWithGenericNode(
+                                        compType,
+                                        new[] { protData.Mapping }, datanode, this);
+                            }
+                            CurrentComponentData[value] = (IComponent) _serializationManager.Read(compType, datanode, this)!;
                         }
                     }
```
ElectroJr: AFAIK, this behaviour is such that if an entity saved to a map has any data that differs from the prototype, this will create a new component data node and then will re-read/deserialize the whole component, including re-deserializing all the fields that haven't changed.

AFAICT `ISerializationManager.Read` has the option to pass in an existing value, though I don't know if thats used anywhere. So would/could it be better to instead just copy the prototype component and then read into that, instead of the push node & full read?

ElectroJr: Also, question about the "read into" functionality. Is it intended to read into a copy of the passed in value, or is it intended to directly read into it (i.e., modify the passed in object)?

To me, the documentation implies it should modify the value, but then It should probably be using a `ref` keyword? If its meant to read into a copy, then the parameter description should probably be more explicit about that. 

I guess the same question applies to the copy-into functionality. It's documentation at least states that the returned value `is not necessarily the same instance`, which TBH is even more confusing, as it implies that it may or may not be? So there is no standard? Yet the new copying relies on the fact that it modifies the target, and completely ignores the returned value.

PaulRitter: copying is a massive clusterfuck rn, i agree that i should change it to just use a ref.

as for:
> AFAICT ISerializationManager.Read has the option to pass in an existing value, though I don't know if thats used anywhere. So would/could it be better to instead just copy the prototype component and then read into that, instead of the push node & full read?

no, we want proper inheritance behavior which is not implemented for copying & passing the object into read to be populated. we *need* pushcomposition for this.

PaulRitter:
superseded by #3527

PR #3087, opened 2022-07-28
Fix CustomHashSetSerializer bug

ElectroJr:
Fixes a bug in the CustomHashSetSerializer's copy function that is currently causing test failures for #3084.

Also fixes what I think are two other bugs in copying, where a failed attempt to copy will instead return a reference to the source, rather than actually copying.


metalgearsloth:
@PaulRitter

Robust.Shared/Serialization/Manager/SerializationManager.cs
```diff
@@ -402,7 +402,7 @@ internal bool TryGetDefinition(Type type, [NotNullWhen(true)] out DataDefinition
         {
             if (source == null || target == null)
             {
-                return source;
```
PaulRitter: no! source might have a value but target may not. but what i just realized: this should not return source, this should create a copy of source if target is null. oops!

Robust.Shared/Serialization/Manager/SerializationManager.cs
```diff
@@ -502,7 +502,8 @@ internal bool TryGetDefinition(Type type, [NotNullWhen(true)] out DataDefinition
         public object? CopyWithTypeSerializer(Type typeSerializer, object? source, object? target,
             ISerializationContext? context = null, bool skipHook = false)
         {
-            if (source == null || target == null) return source;
+            if (source == null || target == null)
+                return null;
```
PaulRitter: same here

PaulRitter (changes requested):
god i really just need to convert the copy api to use ref

Robust.Shared/Serialization/Manager/SerializationManager.cs
ElectroJr: For this function I guess I should just re-use `CreateCopy()` to create a copy of the source.

But I'm not sure what to do with `CopyWithTypeSerializer()`, because type serializers don't currently support copying into a null target. I guess I can just hope it is value type or has a parameterless constructor, but the correct fix is probably to allow custom type serializers to be able copy into null (or alternatively: make them provide a constructor and just copy into a newly created object).

But that's probably out of scope for this PR. So unless there's something else that should be done, I'll just use something similar to `CreateCopyInternal()` and slap a //TODO onto it.

Also, I now have it ignoring `_copyByRefRegistrations`, in case someone wants to use a custom type serializer to override that functionality. Or should that registration override the serializer?

Robust.Shared/Serialization/Manager/SerializationManager.cs
```diff
@@ -503,7 +503,29 @@ public void Copy<T>(T source, ref T target, ISerializationContext? context = nul
         public object? CopyWithTypeSerializer(Type typeSerializer, object? source, object? target,
             ISerializationContext? context = null, bool skipHook = false)
         {
-            if (source == null || target == null) return source;
+            if (source == null)
+                return null;
+
+            // TODO should this respect _copyByRefRegistrations? Or should the type serializer be allowed to override this?
```
PaulRitter, line 509: it should be allowed to override

PaulRitter (approved):

PR #3088, opened 2022-07-29
Fix awake-physics map change bug

ElectroJr:
If an entity gets inserted into a container on another map, this can result in it not being removed from the original map's list of awake bodies. The container check was setting awake to false before the map change logic was getting run.

This PR fixes that and also makes some other misc changes:
- The Awake setter now just calls `SetAwake()` directly. Some of the checks from the setter were moved into `SetAwake()`
- ForceAwake is now just an optional bool for SetAwake

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.Components.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.Components.cs
@@ -27,7 +27,6 @@ public partial class SharedPhysicsSystem
 
         if (component._canCollide && xform.MapID != MapId.Nullspace)
         {
-            bool awake;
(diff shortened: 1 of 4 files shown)


Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.Components.cs
```diff
@@ -28,27 +28,15 @@ private void OnPhysicsInit(EntityUid uid, PhysicsComponent component, ComponentI
         if (component._canCollide && xform.MapID != MapId.Nullspace)
         {
             bool awake;
```
metalgearsloth: Is this still used?

ElectroJr: No its not. Removed now.

metalgearsloth (approved):

PR #3091, opened 2022-07-29
More joint fixes

ElectroJr:
Fixes another joint-init related bug that wasn't fixed in #3068.
Also adds more verbose error logs when duplicate joints are added.


Elijahrane:
Tested locally, fixed the issue I was having completely. Very good PR.

PR #3092, opened 2022-07-30
Update ImageSharp and fix incompatibilities

juliangiebel:
I updated ImageSharp to the latest version. This is required for changes to the map renderer. Esp. for exporting to lossless webp.

The only breaking change I could find is that `TryGetSinglePixelSpan` is now `DangerousTryGetSinglePixelMemory` which returns a memory object that contains the span and that `PreferContiguousImageBuffers` needed to be set to true.

I started up the game, ran tests and ran the map renderer and everything works fine.


juliangiebel:
The check failure is because the Content.MapRenderer PR and this need to be merged at the same time.

juliangiebel:
The corresponding content PR is up now

PR #3099, opened 2022-07-31
BUI range check changes

ElectroJr:
- Uses transform queries
- Allows the max range to be set by each BUI
- No longer uses `ICommonSession.AttachedEntityTransform` (hidden `IoCManager.Resolve<IEntityManager>()`).
  - Also removes a use in placement manager.

### DIFF
--- a/Robust.Server/GameObjects/EntitySystems/UserInterfaceSystem.cs
+++ b/Robust.Server/GameObjects/EntitySystems/UserInterfaceSystem.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


Robust.Server/GameObjects/EntitySystems/UserInterfaceSystem.cs
```diff
@@ -101,45 +99,46 @@ private void OnMessageReceived(BoundUIWrapMessage msg, EntitySessionEventArgs ar
         /// <inheritdoc />
         public override void Update(float frameTime)
         {
+            var query = GetEntityQuery<TransformComponent>();
             foreach (var userInterface in _activeInterfaces.ToList())
             {
-                CheckRange(userInterface);
+                CheckRange(userInterface, query);
                 userInterface.DispatchPendingState();
```
metalgearsloth, line 106: Would be nice as a component as you can then also do EntityQuery<BUIComp, XFormComp> but future problem.

metalgearsloth (approved):

Robust.Server/GameObjects/EntitySystems/UserInterfaceSystem.cs
ElectroJr: I'll probably chuck that into #3102

PR #3106, opened 2022-08-05
Fix nullable yaml validation

ElectroJr:
Currently the yaml linter will error on some "null" values.


metalgearsloth:
thank god

ElectroJr:
> thank god

This is just a linter thing, there still other nullable jank

PR #3107, opened 2022-08-06
Allow entities to be spawned directly in null-space

ElectroJr:
AFAIK this causes no issues, and avoids people needing to spawn into some arbitrary map and then immediately detaching the entity, as is done in space-wizards/space-station-14/pull/10348.

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.cs
+++ b/Robust.Shared/GameObjects/EntityManager.cs
@@ -191,6 +191,14 @@ namespace Robust.Shared.GameObjects
         {
             var newEntity = CreateEntity(prototypeName);
             var transform = GetComponent<TransformComponent>(newEntity);
+
+            if (coordinates.MapId == MapId.Nullspace)
+            {
+                transform._parent = EntityUid.Invalid;
+                transform.Anchored = false;
+                return newEntity;
+            }
+
             transform.AttachParent(_mapManager.GetMapEntityId(coordinates.MapId));
 
             // TODO: Look at this bullshit. Please code a way to force-move an entity regardless of anchoring.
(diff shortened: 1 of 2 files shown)


Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -191,7 +191,25 @@ public virtual EntityUid CreateEntityUninitialized(string? prototypeName, MapCoo
...
-            transform.AttachParent(_mapManager.GetMapEntityId(coordinates.MapId));
+
+            var mapEnt = _mapManager.GetMapEntityId(coordinates.MapId);
+            TryGetComponent(mapEnt, out TransformComponent? mapXform);
+
+            // If the entity is being spawned in null-space, we will parent the entity to the null-map, IF it exists.
+            // For whatever reason, tests create and expect null-space to have a map entity, and it does on the client, but it
+            // intentionally doesn't on the server??
+            if (coordinates.MapId == MapId.Nullspace &&
+                mapXform == null)
```
ElectroJr, line 202: I really don't understand this. The `MapManagerTests` create and require a null-space map entity, the client creates a null-space map entity, but the server doesn't and there are comments that explicitly state that null-space shouldn't be tied to an entity???

I have no idea if there is some reason for this split, or if there isn't, which method would be preferred. I assume not having an entity is better?

If there is meant to be an entity that things get parented to, the DetachToNull needs to be updated to actually parent to the null-map entity, and the transform init changes need to removed.

metalgearsloth: Yeah a lot of what mapmanager does doesn't make sense to me, especially the parts where tests do hyper-specific things that crash when you actually try it ingame.

PR #3113, opened 2022-08-08
Make tiledefinitions use spritespecifier

metalgearsloth:


Robust.Shared/Utility/SpriteSpecifier.cs
```diff
@@ -51,6 +54,11 @@ public override int GetHashCode()
             {
                 return RsiPath.GetHashCode() ^ RsiState.GetHashCode();
             }
+
+            public override ResourcePath GetPath()
+            {
+                return new ResourcePath(RsiPath + RsiState);
```
PJB3005: Reading the raw .png files inside a .rsi is not acceptable. It does not play nice with tons of systems current and future.

Using sprite specifiers for this does not make much sense to me.

PJB3005 (changes requested):

Robust.Shared/Utility/SpriteSpecifier.cs
metalgearsloth: I'm fine changing it back to a ResourcePath on the tile definition but I'm at least removing the hardcoded tiles path.

PJB3005: That's fine

PR #3114, opened 2022-08-08
Fix issue with Fluents ??? for empty suffix.

Rinkashikachi:
## About the PR
This fix addresses [issue 10438](https://github.com/space-wizards/space-station-14/issues/10438) from SS14 repo.

Previously there was no Fluent support for entities in the issue. So the engine used only what .yml files gave it. With .ftl files added it seems like Fluent defaults empty values for optional fields to `{???}` instead of `null` or just whitespace like with pure .yml.

Due to the value being neither `null` nor whitespace the check fails and weird `[{???}]` appears next to the names of entities in the Entity spawn panel for all Fluent-capable entities.

As mentioned in the issue by Morb0, one way to fix this is to add optional empty .suffix fields to .ftl descriptions for all entities. Obviously if this field is optional it should remain optional, so parsing changes are needed. That's what this fix is for.

### DIFF
--- a/Robust.Client/UserInterface/CustomControls/EntitySpawnWindow.cs
+++ b/Robust.Client/UserInterface/CustomControls/EntitySpawnWindow.cs
@@ -318,8 +318,7 @@ namespace Robust.Client.UserInterface.CustomControls
             button.ActualButton.OnToggled += OnItemButtonToggled;
             var entityLabelText = string.IsNullOrEmpty(prototype.Name) ? prototype.ID : prototype.Name;
 
-            // "{???}" value happens when Fluent can't find a value in .ftl files for EditorSuffix variable.
-            if (!string.IsNullOrWhiteSpace(prototype.EditorSuffix) && !"{???}".Equals(prototype.EditorSuffix))
+            if (!string.IsNullOrWhiteSpace(prototype.EditorSuffix))
             {
                 entityLabelText += $" [{prototype.EditorSuffix}]";
             }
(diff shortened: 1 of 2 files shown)


Ygg01:
The underlying issue here is that you are passing [FluentNone](https://github.com/Ygg01/Linguini/blob/898b04294883b4dc8942fd075565d6cddf8e5de3/Linguini.Shared/Types/Bundle/FluentNone.cs) and getting its string value.

Where is that string being generated?

Ygg01:
Did some digging. [According to Fluent test fixtures](https://github.com/Ygg01/Linguini/blob/89ec0ae3ec71ca28d8a3cb70815c5c1d0cb4502e/Linguini.Bundle.Test/fixtures/patterns.yaml#L66-L97), asking for an attribute with value null is supposed to return an error and `{???}` string, so Linguini is in the right on this one.

When the attribute isn't found in `TryGetMsg` that method returns `false`. So rather than relying on hackery in `EntitySpawnWindow`, it's easier to set the suffix to `""` or `null` when `TryGetMsg` returns false.

Here are the lines in question. 

https://github.com/space-wizards/RobustToolbox/blob/9e32eb46e50de3ee23b8ab10d2a51672b0a01cd9/Robust.Shared/Localization/LocalizationManager.Entity.cs#L102-L106)

You will need to make nested ifs. E.g.

```cs
if (suffix == null) {
   if(!bundle.TryGetMsg(locId, "suffix", null, out var err, out suffix) {
     ....
   }
}

PR #3118, opened 2022-08-10
Add private ctor for ResourcePath

metalgearsloth:
Disease outbreaks exception on ss14 atm whenever a copy of ResourcePath is tried due to no ctor existing for this.


PJB3005 (changes requested):
This should probably be equivalent to `new ResourcePath("")` instead?

PR #3123, opened 2022-08-12
Fixed crash in ApplySizeConstraints

### LINKED ISSUE #3029: Setting MaxWidth on a DefaultWindow causes an exception if the window is resized past the set max width
As said in the title. Resultant exception (without rest of stack trace):

image

DoctorBeard:
Fixes #3029

This was caused by ApplySizeConstraints trying to use the current size of a control as its minimum size and then trying to clamp with the minimum value being larger than the maximum value. 

MinSize is by default 0 so that value is just taken for the minimum constraints as is, regardless of what is set. After that it either takes MaxSize, if set, or SetSize, if set and actually larger or equal to MinSize as the maximum constraint. If neither of these is true the maximum constraint will end up being float.PositiveInfinity (the default of MaxSize).

ApplySizeConstraints isn't at all touched by the UI refactor last I checked so this PR should not conflict with that.

### DIFF
--- a/Robust.Client/UserInterface/Control.Layout.cs
+++ b/Robust.Client/UserInterface/Control.Layout.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using Robust.Client.Graphics;
 using Robust.Client.UserInterface.Controls;
 using Robust.Shared.Map;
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/Control.Layout.cs
```diff
@@ -689,28 +689,22 @@ protected virtual Vector2 ArrangeOverride(Vector2 finalSize)
...
+                    Math.Clamp(avail.Y, minH, maxH));
 
-            minConstraint = float.IsNaN(setH) ? 0 : setH;
-            minH = MathHelper.Clamp(maxH, minConstraint, minH);
+        }
+
+        private static float GetActualMaxSize(float minSize, float maxSize, float setSize)
+        {
+            if (float.IsPositiveInfinity(maxSize) && !float.IsNaN(setSize) && setSize >= minSize)
+                return setSize;
```
ElectroJr, line 705: This no longer just uses set size even when they are all specified and min < set < max ?

PR #3130, opened 2022-08-14
Add event for entity placement

reece-hanham:
Raises an event on entity placement, containing the freshly placed entity and the session responsible for placing the entity. This is used in https://github.com/space-wizards/space-station-14/pull/10206 to automatically attach entities to the ghost role groups.

### DIFF
--- a/Robust.Server/Placement/PlacementManager.cs
+++ b/Robust.Server/Placement/PlacementManager.cs
@@ -122,6 +122,7 @@ namespace Robust.Server.Placement
                 var created = _entityManager.SpawnEntity(entityTemplateName, coordinates);
 
                 _entityManager.GetComponent<TransformComponent>(created).LocalRotation = dirRcv.ToAngle();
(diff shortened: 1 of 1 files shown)


metalgearsloth:
How come this is used over component startup?

reece-hanham:
I needed to associate an entity that was just placed down with the player that had placed it down. I couldn't see anything currently existing that did this, so I implemented it as this event. Is there a better way of doing this?

ElectroJr:
Just briefly looking at the ghost role PR, its not clear to me what you are trying to achieve. Spawn-menu entity spawning shouldn't depend on the player that is doing the spawning and should just act as if the entity had been spawned via EntityManager functions.

If this for a some mob ability to spawn allies/minions using the placement manager, you should use something like the construction system's highjack stuff (see `ConstructionPlacementHijack`). Its not great but should be able to achieve this?

reece-hanham:
Cheers, I'll have a look at that.

PR #3156, opened 2022-08-21
Physics Substepping

### LINKED ISSUE #3116: Physics substepping
Another means to address tunneling. Right now physics runs at 30tps but most engines recommend 50+.

Ideally you'd have some cvar for the target minimum physics tickrate (60) and then somewhere in physics you would get the deltatime from that. So for example if the game is running at 40tps then it would do 2 substeps for 80 tps physics (because physics steps should always be equal length for stability).

Probably need to also include the movement controllers in this as well I guess maybe? Though mob movement inputs may not like it so need to check that.

keronshb:
closes #3116

Adds Substepping to physics, which should help with some collision issues but not all.

https://user-images.githubusercontent.com/54602815/198683010-6efd3273-279a-4f90-b93a-d72317908b3f.mp4


metalgearsloth:
Mentioned on disc but for any other reviewer:
- Client lerping needs fixing (blocker)
- We should really add a test for this

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -323,26 +324,30 @@ private void HandleContainerRemoved(EntityUid uid, PhysicsComponent physics, Ent
...
-            // Go through and run all of the deferred events now
-            foreach (var comp in EntityManager.EntityQuery<SharedPhysicsMapComponent>(true))
-            {
-                comp.ProcessQueue();
-            }
+                foreach (var comp in EntityManager.EntityQuery<SharedPhysicsMapComponent>(true))
+                {
+                    comp.Step(frameTime, prediction);
+                    var updateAfterSolve = new PhysicsUpdateAfterSolveEvent(prediction, frameTime);
+                    RaiseLocalEvent(ref updateAfterSolve);
```
ElectroJr: AFAIK this should just be raised once (I assume per sub-step?), not once per map? Especially seeing as there is a separate `PhysicsUpdateAfterMapSolveEvent` that gets raised for each map

Robust.Shared/GameObjects/Systems/SharedPhysicsSystem.cs
```diff
@@ -323,26 +324,30 @@ private void HandleContainerRemoved(EntityUid uid, PhysicsComponent physics, Ent
...
+
+                    comp.ProcessQueue();
```
ElectroJr: Can this just be run once per update, or is there some reason it has to be run in each subtick? And if it does have to be run per subtick, would it be better to have a separate sub-tick move event? I imagine the vast majority of move event subscribers don't care about sub-tick movement.

ElectroJr: Apparently this might be required to add things to the move buffer so that contacts are generated? But again, that should then just be a separate event or directly invoked. Calling the fixture synchronize function directly would also mean it only updates once per entity, whereas AFAICT at the moment it can get called twice, for movement + rotation

ElectroJr: I think the deferred move events & move-buffer / contact generation may just need rejigging in general, so its probably fine to leave as is.

ElectroJr (commented):
I haven't really been keeping up with the discord discussions, is this still a WIP / draft, or is it ready for reviews?

keronshb:
I've hit a wall on this. Lerping still has some issues IIRC and to fix it, it needs "displacement in the broadphase" according to @metalgearsloth  but I've tried some things and haven't been able to make any progress.

keronshb:
Going back to draft for a bit while it's figured out.

PR #3174, opened 2022-08-23
Hopefully fix container issues?

ElectroJr:
I've managed to somewhat reliably reproduce the current container bug, and AFAICT this seems to fix it? At least it hasn't reoccurred despite repeatedly attempting to trigger it.

Aside from the fix, this PR also adds a bunch of `DebugTools.Asserts()` and changes some of the code a bit to ensure that an entity can never be both inside of a container and detached to null-space simultaneously.

Without space-wizards/space-station-14/pull/10783, these asserts fail for SS14.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using Robust.Client.Player;
 using Robust.Shared.Collections;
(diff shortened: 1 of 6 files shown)


Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
```diff
@@ -140,8 +140,11 @@ private void HandleComponentState(EntityUid uid, ContainerManagerComponent compo
                         continue;
                     }
 
-                    if (!container.ContainedEntities.Contains(entity))
-                        container.Insert(entity);
+                    if (container.Contains(entity))
+                        continue;
+
+                    RemoveExpectedEntity(entity, out _);
+                    container.Insert(entity);
```
ElectroJr, line 147: This is the actual fix. Inserting without first removing expected can lead to the same entity getting inserted into a single container more than once (why is it even a list instead of a set?)

metalgearsloth (approved):

PR #3181, opened 2022-08-24
Fix load grid errors

### LINKED ISSUE #3177: If the maploader loads a map as a blueprint, it will delete reality.
Apparently one of SS14's salvage maps does this and it bricks the game on the spot by deleting the station map and station.

### LINKED ISSUE #3179: BP commands need renaming to Grid
They're for saving / loading specific grids but for as long as I've been here they've just been called grids / mapgrids so the commands + methods should just be "savegrid", "loadgrid", etc.

ElectroJr:
- Renames load/savebp commands to load/savegrid (fixes #3179)
- Using loadgrid to load map file now outputs a descriptive error and should no longer unintentionally delete maps (fixes #3177)
  - ~~There will still be errors, but in release mode, the game still seems to function & rounds can re-start after a bad grid load.~~
  - Correction: after #3160 & the LoadMap = false option, it doesn't even error anymore.
- This PR also updates the client-side session's attached entity to null when the player is detached from an entity
  - Makes it clearer in logs that the player has been detached & the old entity has been deleted, rather than the somewhat confusing "\<OldPlayerUid\> does not have XYZ component" errors. AFAIK this doesn't cause any issues.

Requires space-wizards/space-station-14/pull/10817, for the command permissions and because the `LoadBlueprint` -> `LoadGrid` renaming.

### DIFF
--- a/Robust.Server/Maps/MapLoader.cs
+++ b/Robust.Server/Maps/MapLoader.cs
@@ -84,7 +84,9 @@ namespace Robust.Server.Maps
         public (IReadOnlyList<EntityUid> entities, EntityUid? gridId) LoadGridBlueprint(MapId mapId, string path, MapLoadOptions options)
         {
             DebugTools.Assert(_mapManager.MapExists(mapId));
-            options.LoadMap = false;
+
+            var oldLoadMapOpt = options.LoadMap; // lets not mutate the default options
+            options.LoadMap = false; 
 
             var resPath = Rooted(path);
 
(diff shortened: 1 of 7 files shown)


Robust.Server/Maps/MapLoader.cs
```diff
@@ -770,8 +778,18 @@ private void FinishEntitiesLoad()
 
                     if (mapQuery.HasComponent(entity))
                     {
-                        DebugTools.Assert(TargetMapUid == null);
-                        TargetMapUid = entity;
+                        if (TargetMapUid == null)
+                        {
+                            TargetMapUid = entity;
+                        }
```
ElectroJr: This is the main bit that caused the map to get deleted. Later on it would assign this newly loaded map entity as the entity associated with the target map ID, deleting the old map entity in the process.

metalgearsloth (approved):

PR #3186, opened 2022-08-25
Add support for snapping sprites to south

### LINKED ISSUE #2262: 1 direction sprites need a way to snap at cardinals
So something like an airlock can snap at 45 degrees to try to face south instead of being able to go upside down. Not really sure on the best way to store this on a component TBH, maybe an enum to override the behavior to act like it has 4 / 8 directions?

metalgearsloth:
Tested it on my camera branch.

Closes https://github.com/space-wizards/RobustToolbox/issues/2262


ElectroJr (approved):
Also requires changes to bounding boxes & clickmaps, but seeing as nothing uses this functionality, I'll just do those myself in a separate PR

PR #3192, opened 2022-08-26
Fix client state handling bug

ElectroJr:


Robust.Client/GameStates/GameStateProcessor.cs
```diff
@@ -114,13 +114,20 @@ public bool AddNewState(GameState state)
                 return true;
             }
 
-            if (LastFullState == null || state.ToSequence <= LastFullState.ToSequence)
+            if (LastFullState != null && state.ToSequence <= LastFullState.ToSequence)
```
ElectroJr, line 117: This is the actually important change.

metalgearsloth (approved):

PR #3195, opened 2022-08-28
Update sprite bounds for SnapCardinals

ElectroJr:
Currently rectangular sprites have incorrect bounds when SnapCardinals is true.


moonheart08 (approved):
merg?

PR #3199, opened 2022-08-30
Move MoveBuffer to the physicsmapcomponent

metalgearsloth:
Baby steps in ECSifying physics fully.

Should also avoid the movebuffer crashes whenever something goes awry with a map.


Robust.Shared/Physics/BroadphaseComponent.cs
```diff
@@ -2,6 +2,9 @@
 
 namespace Robust.Shared.Physics
 {
+    /// <summary>
+    /// Stores the broadphase structore for the relevant grid / map.
```
mirrorcult:
```suggestion
    /// Stores the broadphase structure for the relevant grid / map.
```

PR #3211, opened 2022-09-01
Box2 from two points

Macoron:
Add a new utility function that creates Box2 from two points. Find min-max of x and y and create a new box from them.

It can be written shorter using `Mathf.Min`, `Mathf.Max` or tuples `(minX, maxX) = ...`, but my Rider IL Viewer says that it's the fastest version.

### DIFF
--- a/Robust.Shared.Maths/Box2.cs
+++ b/Robust.Shared.Maths/Box2.cs
@@ -126,6 +126,13 @@ namespace Robust.Shared.Maths
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Box2 FromTwoPoints(Vector2 a, Vector2 b)
+        {
+            return new Box2(MathF.Min(a.X, b.X), MathF.Min(a.Y, b.Y),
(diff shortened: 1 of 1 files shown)


Robust.Shared.Maths/Box2.cs
```diff
@@ -125,6 +125,34 @@ public static Box2 CentredAroundZero(Vector2 size)
...
+                maxY = b.Y;
+            }
+            else
+            {
+                minY = b.Y;
+                maxY = a.Y;
+            }
+
+            return new Box2(minX, minY, maxX, maxY);
+        }
```
metalgearsloth, line 154: Vector2.ComponentMin / Vector2.ComponentMax?

Macoron:
```csharp
            return new(
                MathF.Min(a.X, b.X),
                MathF.Min(a.Y, b.Y)
            );
```

Maybe I'm digging too deeply, but that's 4 compare operations (2 for min and 2 for max), while in my implementation there are only 2. It doesn't look that compiler optimize it somehow, but I might be wrong.

PR #3222, opened 2022-09-05
Fix input packet loss

### LINKED ISSUE #3031: Client -> Server keybinding/input messages handle packet loss poorly.
Having high packet loss leads to very noticeable stuttering when starting or stopping movement in some direction. This Isn't fixed by #3000, that only deals with Server -> client packet loss.

The fix is probably to just separate inputs out from entity events, and having the client send inputs for a range of ticks, rather than single ticks.

ElectroJr:
This PR tries to fix input packet loss issues. Previously inputs were just sent reliably as a normal ECS event. Whenever a message gets dropped it leads to the player seemingly lurching forwards (dropped key-up, server thinks player is still moving), or being stuck in place (dropped key-down server doesn't think player should be moving).

For example, this is me trying my best to run clockwise around a racetrack with silly high packet loss:
```
cvar net.fakeloss 0.1
cvar net.predict_tick_bias 2
cvar net.fakelagmin 0.125
sudo cvar net.fakelagmin 0.125
```

https://user-images.githubusercontent.com/60421075/188358356-6d874225-eedd-4268-8b15-34a1516bbbf8.mp4

This PR tries to fix this by just sending inputs in a separate unreliable event, where there is one message every tick containing the unacknowledged inputs.

Given that inputs are no longer normal networked system events, a `BeforeTickUpdateEvent` was added in order to make sure that inputs are handled before systems update. Alternatively, I guess `InputSystem` could've just been merged with InputManager, with its own non-system tick update function?

Fixes #3031


Robust.Shared/Serialization/RobustSerializer.cs
```diff
@@ -32,7 +33,8 @@ public sealed partial class RobustSerializer : IRobustSerializer
 
         private static Type[] AlwaysNetSerializable => new[]
         {
-            typeof(Vector2i)
+            typeof(Vector2i),
+            typeof(NetListAsArray<FullInputCmdMessage>)
         };
```
ElectroJr, line 38: AFAIK this is currently the only way to mark something like `NetListAsArray<T>` as serializable via IRobustSerializer without just creating some dummy `[NetSerializable]` class that contains this type.

If this isn't what this list is meant to be used for, I guess I can just use a dummy class or mark the new NetMessage as serializable?

ElectroJr:
Reworking this and will just reopen as another PR

PR #3224, opened 2022-09-05
Use TryComp instead of GetComp for animation playback

metalgearsloth:
It at least means we won't get exception spam in some instances.


Robust.Client/Animations/AnimationTrackComponentProperty.cs
```diff
@@ -20,7 +20,12 @@ protected override void ApplyProperty(object context, object value)
             }
 
             var entity = (EntityUid) context;
-            var component = IoCManager.Resolve<IEntityManager>().GetComponent(entity, ComponentType);
+            var entManager = IoCManager.Resolve<IEntityManager>();
+
+            if (!entManager.TryGetComponent(entity, ComponentType, out var component))
+            {
+                return;
+            }
```
ElectroJr, line 30: It should probably still at least print an error to the console

PR #3235, opened 2022-09-06
Fix fixture clearing on SetTiles

metalgearsloth:
Oop.

Might fix grid collision.

### DIFF
--- a/Robust.Shared/Map/MapGrid.cs
+++ b/Robust.Shared/Map/MapGrid.cs
@@ -969,11 +969,13 @@ namespace Robust.Shared.Map
                 }
             }
 
+            // May have been deleted from the bulk update above!
+            if (_entityManager.Deleted(GridEntityId))
+                return;
+
             // TODO: Move this to the component when we combine.
             _mapManager.OnGridBoundsChange(GridEntityId, this);
-            // May have been deleted from the bulk update above!
-            if (!_entityManager.Deleted(GridEntityId))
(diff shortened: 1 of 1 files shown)


Robust.Shared/Map/MapGrid.cs
```diff
@@ -976,49 +980,7 @@ public void RegenerateCollision(IReadOnlySet<MapChunk> chunks)
...
-                {
-                    var gridBounds = chunkBounds.Translated(chunk.Indices * chunk.ChunkSize);
-                    LocalAABB = LocalAABB.Union(gridBounds);
-                }
-            }
-
-            if (!_entityManager.EntitySysManager.TryGetEntitySystem(out SharedGridFixtureSystem? system) ||
-                _entityManager.Deleted(GridEntityId)) return;
-
-            // TODO: Move this to the component when we combine.
```
ShadowCommander, line 1018: Is this comment still relevant?

ShadowCommander (approved):

Robust.Shared/Map/MapGrid.cs
metalgearsloth: Yeah. There's a lot of hacks around grid code because of mapmanager being its own thing.

ShadowCommander: Then it should be moved to the combined code at 967

PR #3240, opened 2022-09-08
Drop SQLitePCL dependency

Partmedia:
Nothing in Robust.Server actually uses Sqlite, so drop this unused dependency. Any RobustToolbox users will need to add this back if they actually use Sqlite.

### DIFF
--- a/Robust.Server/Robust.Server.csproj
+++ b/Robust.Server/Robust.Server.csproj
@@ -13,7 +13,6 @@
     <PackageReference Include="JetBrains.Annotations" Version="2021.3.0" PrivateAssets="All" />
     <PackageReference Include="SpaceWizards.HttpListener" Version="0.1.0" />
     <!--  -->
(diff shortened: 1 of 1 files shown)


PJB3005:
This dependency exists so the packaging system picks up on the native library when packaging the server. We would need to revise it to be smart enough to copy `e_sqlite` over manually otherwise, which isn't all that trivial when you get into RIDs.

Partmedia:
> This dependency exists so the packaging system picks up on the native library when packaging the server. We would need to revise it to be smart enough to copy `e_sqlite` over manually otherwise, which isn't all that trivial when you get into RIDs.

I see. Should I leave this open as "planned eventually when someone gets around to it" or close it because it'll break too many things for now?

PJB3005:
Probably better to just open an issue instead unless you're planning on tackling this right now, yeah.

Partmedia:
Closed for now due to more complicated packaging issues.

PR #3253, opened 2022-09-13
Add alt use key

metalgearsloth:
Not sure if I should repurpose "OpenContextMenu" on content for this but the default keybind in ss14 is going to be right-click.


mirrorcult:
this kinda heavily conflicts (at least name wise) with the alt click that already exists in content

metalgearsloth:
> this kinda heavily conflicts (at least name wise) with the alt click that already exists in content

Could call it UseSecondary?

PR #3256, opened 2022-09-15
ECS 90% of physics component

metalgearsloth:
I also networked damping which I forgor to get around to.

Also found applylinearimpulse from point was missing a property + re-implemented linear velocity setting when massdata is reset (this really only matters for stuff that changes COM like a grid split).

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
+++ b/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
@@ -126,12 +126,16 @@ public partial class SharedPhysicsSystem
 
     #endregion
 
+    private bool IsMoveable(PhysicsComponent body)
+    {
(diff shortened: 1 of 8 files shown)


Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
```diff
@@ -89,29 +118,181 @@ private void OnPhysicsHandleState(EntityUid uid, PhysicsComponent component, ref
...
     }
 
+    #endregion
+
+    #region Impulses
+
+    public void ApplyAngularImpulse(PhysicsComponent body, float impulse)
+    {
+        if ((body._bodyType & (BodyType.Dynamic | BodyType.KinematicController)) == 0x0 ||
+            !WakeBody(body))
```
ShadowCommander: Should this eventually be another method since it's a lot of code reuse?

ShadowCommander (approved):

PR #3259, opened 2022-09-15
Don't sleep initialised items

metalgearsloth:
It means some stuff spawns stuff into walls on content. As nice as it is to avoid waking these bodies it's kinda necessary.

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
+++ b/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
@@ -14,7 +14,6 @@ namespace Robust.Shared.Physics.Systems;
 
 public partial class SharedPhysicsSystem
 {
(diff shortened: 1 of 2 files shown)


ShadowCommander (approved):

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
```diff
@@ -53,10 +47,6 @@ private void OnPhysicsInit(EntityUid uid, PhysicsComponent component, ComponentI
 
     private void OnPhysicsInitialized(EntityUid uid)
     {
-        if (EntityManager.TryGetComponent(uid, out CollisionWakeComponent? wakeComp))
-        {
-            _collisionWakeSystem.OnPhysicsInit(uid, wakeComp);
-        }
```
ElectroJr, line 59: Shouldn't the `OnPhysicsInit()` function itself also be removed?

PR #3263, opened 2022-09-17
Prevent PVS error spam and add more debugging asserts

ElectroJr:
Currently an error while handling player disconnects will result in pvs spamming errors while generating mail. This PR changes the `AddPlayer` & `RemovePlayer` functions so that they return bools rather than throwing exceptions, so that this just logs an error.

The original exception that caused the error spam on grafana was seemingly due to a player disconnecting twice? or disconecting without having fully connected?
```
Caught exception in OnDisconnected handler:
System.Collections.Generic.KeyNotFoundException: The given key 'Ponch' was not present in the dictionary.
   at Robust.Server.GameStates.PVSCollection`1.RemovePlayer(ICommonSession session) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Server/GameStates/PVSCollection.cs:line 271
   at Robust.Server.GameStates.PVSSystem.OnPlayerStatusChanged(Object sender, SessionStatusEventArgs e) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Server/GameStates/PVSSystem.cs:line 400
   at Robust.Server.Player.PlayerManager.EndSession(Object sender, NetChannelArgs args) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Server/Player/PlayerManager.cs:line 448
   at Robust.Shared.Network.NetManager.HandleDisconnect(NetPeerData peer, NetConnection connection, String reason) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/Network/NetManager.cs:line 790
 Sawmill=net
```

This PR also adds some more deletion related checks/asserts, because there is still some bug related to PVS & entity deletion and I have NFI what causes it.

### DIFF
--- a/Robust.Server/GameStates/PVSCollection.cs
+++ b/Robust.Server/GameStates/PVSCollection.cs
@@ -18,10 +18,17 @@ public interface IPVSCollection
     /// Processes all previous additions, removals and updates of indices.
     /// </summary>
     public void Process();
+
+    /// <summary>
+    ///     Adds a player session to the collection. Returns false if the player was already present.
+    /// </summary>
     public bool AddPlayer(ICommonSession session);
     public void AddGrid(GridId gridId);
     public void AddMap(MapId mapId);
 
+    /// <summary>
+    ///     Removes a player session from the collection. Returns false if the player was not present in the collection.
+    /// </summary>
     public bool RemovePlayer(ICommonSession session);
 
     public void RemoveGrid(GridId gridId);
--- a/Robust.Server/GameStates/PVSCollection.cs
+++ b/Robust.Server/GameStates/PVSCollection.cs
@@ -18,11 +18,11 @@ public interface IPVSCollection
     /// Processes all previous additions, removals and updates of indices.
     /// </summary>
     public void Process();
-    public void AddPlayer(ICommonSession session);
+    public bool AddPlayer(ICommonSession session);
     public void AddGrid(GridId gridId);
     public void AddMap(MapId mapId);
 
-    public void RemovePlayer(ICommonSession session);
+    public bool RemovePlayer(ICommonSession session);
 
     public void RemoveGrid(GridId gridId);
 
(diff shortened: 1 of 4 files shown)


Robust.Server/GameStates/PVSCollection.cs
```diff
@@ -249,10 +249,9 @@ private void AddIndexInternal(TIndex index, IIndexLocation location, HashSet<ICh
     #region Init Functions
 
     /// <inheritdoc />
-    public void AddPlayer(ICommonSession session)
+    public bool AddPlayer(ICommonSession session)
     {
-        _localOverrides[session] = new();
-        _lastSeen[session] = new();
+        return _localOverrides.TryAdd(session, new()) & _lastSeen.TryAdd(session, new());
```
ShadowCommander, line 261: One `&` made me look twice, but thinking about it `&&` would short-circuit if the first TryAdd fails. Not sure which is preferred, but one `&` seems fine.

Robust.Server/GameStates/PVSCollection.cs
```diff
@@ -266,14 +265,17 @@ public void AddPlayer(ICommonSession session)
     #region ShutdownFunctions
 
     /// <inheritdoc />
```
ShadowCommander, line 274: Maybe update doc comments?

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -349,6 +349,7 @@ private void RecursiveFlagEntityTermination(MetaDataComponent metadata, EntityQu
 
         private void RecursiveDeleteEntity(MetaDataComponent metadata, EntityQuery<MetaDataComponent> metaQuery, EntityQuery<TransformComponent> xformQuery, SharedTransformSystem xformSys)
         {
+
```
ShadowCommander:
```suggestion
```

ShadowCommander (approved):
Other than formatting stuff. Looks good.

Robust.Server/GameStates/PVSCollection.cs
ElectroJr: Yeah the choice to use `&` instead of `&&` was intentional.

Robust.Server/GameStates/PVSCollection.cs
ElectroJr: Done. They didn't even actually have docs before.

PR #3273, opened 2022-09-23
Allow startup to continue if prototype watching fails

SpaceManiac:
Example:
```
[ERRO] root: Watching resources in path /d/archives/projects/ss14/space-station-14/Resources/Prototypes threw an exception:
System.IO.IOException: The configured user limit (128) on the number of inotify instances has been reached, or the per-process limit on the number of open file descriptors has been reached.
   at System.IO.FileSystemWatcher.StartRaisingEvents()
   at System.IO.FileSystemWatcher.StartRaisingEventsIfNotDisposed()
   at System.IO.FileSystemWatcher.set_EnableRaisingEvents(Boolean value)
   at Robust.Client.Prototypes.ClientPrototypeManager.WatchResources() in /d/archives/projects/ss14/space-station-14/RobustToolbox/Robust.Client/Prototypes/ClientPrototypeManager.cs:line 124
```

### DIFF
--- a/Robust.Client/Prototypes/ClientPrototypeManager.cs
+++ b/Robust.Client/Prototypes/ClientPrototypeManager.cs
@@ -119,7 +119,14 @@ namespace Robust.Client.Prototypes
                     });
                 };
 
-                watcher.EnableRaisingEvents = true;
+                try
+                {
+                    watcher.EnableRaisingEvents = true;
+                }
+                catch (IOException ex)
+                {
+                    Logger.Error($"Watching resources in path {path} threw an exception:\n{ex}");
+                }
                 _watchers.Add(watcher);
             }
 #endif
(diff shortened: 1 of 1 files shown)


Robust.Client/Prototypes/ClientPrototypeManager.cs
```diff
@@ -119,7 +119,14 @@ private void WatchResources()
...
-                watcher.EnableRaisingEvents = true;
+                try
+                {
+                    watcher.EnableRaisingEvents = true;
+                }
+                catch (IOException ex)
+                {
+                    Logger.Error($"Watching resources in path {path} threw an exception:\n{ex}");
+                }
                 _watchers.Add(watcher);
```
PaulRitter: ig this should also be in the try clause

PaulRitter (changes requested):

Robust.Client/Prototypes/ClientPrototypeManager.cs
SpaceManiac: Makes sense thinking about it.

PaulRitter (approved):

PR #3278, opened 2022-09-25
Grid line enumerator

Macoron:
Added grid line enumerator. Used for radiation system, but might be used for other purposes.

It use bresenhams line algorithm and wrapped into struct enumerator. Should be fast enough.

### DIFF
--- a/Robust.Shared/Utility/GridLineEnumerator.cs
+++ b/Robust.Shared/Utility/GridLineEnumerator.cs
@@ -1,0 +1,81 @@
+using System;
+using Robust.Shared.Maths;
+
+namespace Robust.Shared.Utility;
+
+// bresenhams line algorithm
+// this is slightly rewritten version of code bellow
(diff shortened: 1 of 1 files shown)


PJB3005:
Do we really want this in engine, or should this go in content or something?

Macoron:
> Do we really want this in engine, or should this go in content or something?

Well, it's possible that some other project that using robust could use draw line algorithm. Though they also probably will want to modify it for they needs.

So it's up to maintainers to decide. I'm ok with both options.

metalgearsloth:
Personally I think it should be alongside the other grid tile methods like GetTilesIntersecting as some other game may want linecasts on tiles.

moonheart08:
Frankly, keep it in this form and then add something atop it specialized for grids.
This is itself really useful. (and saves me from having to memorize the line algorithm every time i need it aaaa)

ElectroJr (approved):

PR #3279, opened 2022-09-26
Remove obsolete GridId and all references to it

SpaceManiac:
From 385 to 204 warnings (47% reduction). The subtlest part of this is in the map loader: MapGridComponent still has a serializable `index` field, but it's treated simply as a linear index into the `grids` section of the YAML, and not used after map loading has completed.

Also fixed 3 minor other warnings and suppresses one more.

Content PR: https://github.com/space-wizards/space-station-14/pull/11531


SpaceManiac:
> Test content master against engine / build (pull_request) 

Am I expected to do a two-phase thing here or would merging the two PRs (assuming they otherwise pass muster) at the same time good enough?

metalgearsloth:
We merge them at the same time. On the content PR you need to put "Requires <engine pr>" and link it for the CI to use the pr instead. See https://docs.spacestation14.io/en/getting-started/engine-changes

ElectroJr (approved):

metalgearsloth:
I thought the content side was passing so that needs fixing before remerging this.

> The subtlest part of this is in the map loader: MapGridComponent still has a serializable index field, but it's treated simply as a linear index into the grids section of the YAML, and not used after map loading has completed.

By ommitting the index mapgrids no longer get serialized as they don't have custom data (and atm there's no way to force a component's serialization).

PR #3281, opened 2022-09-26
Open window size fix

wrexbe:
Fix a bunch of UI sizing issues with this one simple trick.


PJB3005:
What are the bugs fixed by this?

wrexbe:
Because it locks the setsize to desiredsize it breaks ui scaling.

PJB3005:
How is this related to UI scaling? These sizes are all in virtual pixel units.

wrexbe:
Set size clamps the max width and height for the window. So whenever it measures the window it will never let it get bigger again, even if it needs it. The windows figure out their size just fine without this line, so bugs went away, and I didn't see any new ones happen. Mainly noticed this on the sandbox and option window, which were getting cutoff sometimes.

PR #3286, opened 2022-09-28
Add system to turn entites into error-signs

ElectroJr:
This adds a client-side system that provides a function to update an entity's sprite, name, and raise an event to indicate that some error has occurred.

The main use of this is to try combat niche/rare client-side errors that simply don't get reported because they might not have a clearly noticeable impact on the game (and most people won't casually check the console for errors). I dunno if this is the best way to go about it, there might be better ways of trying track them.

Also, maybe this should just be an event defined in engine, with content event handlers?

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/ClientErrorSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/ClientErrorSystem.cs
@@ -1,0 +1,37 @@
+using System;
+using Robust.Shared.GameObjects;
+using Robust.Shared.Localization;
+
+namespace Robust.Client.GameObjects;
+
+/// <summary>
+///     Basic system for visually communicating client-side errors to players, so that they hopefully get reported
+///     instead of just being unread error logs somewhere in the console.
+/// </summary>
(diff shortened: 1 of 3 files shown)


Robust.Client/GameObjects/EntitySystems/ClientErrorSystem.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        // SubnauticaSystem when?
+        RaiseLocalEvent(uid, new EntityErrorEvent(uid, e), true);
+
+        if (!updateEntity || !TryComp(uid, out MetaDataComponent? meta))
+            return;
+
+        meta.NetSyncEnabled = false;
+        meta.EntityName = $"ERROR - {ToPrettyString(uid)}";
+        meta.EntityDescription = Loc.GetString("error-system-entity-description");
```
ElectroJr, line 28: Unfortunately the client-side description of networked entities isn't accessible in SS14 (for normal players), but that might change eventually.

PJB3005:
I feel like a telemetry system for error reporting would go a much longer way than something like this.

PR #3289, opened 2022-09-28
Fix a bunch of warnings

metalgearsloth:


PJB3005 (approved):

metalgearsloth:
Gonna revert the transform change because way too many damn test issues.

PR #3296, opened 2022-10-06
Update OpenTK.OpenAL to 4.7.5

Partmedia:
Update OpenTK.OpenAL dependency to 4.7.5. Among other things, 4.7.5 has upstreamed support for FreeBSD platform detection. See: https://github.com/opentk/opentk/pull/1500

Between the pre-release versions, OpenTK changed its package namespace from OpenToolkit to OpenTK. Most of this patch is just chasing this namespace update.


PJB3005:
Well, guess we'll see if this breaks anything.

PR #3298, opened 2022-10-07
Add build property to use system SQLite3

Partmedia:
SQLitePCLRaw.bundle_e_sqlite3 does not bundle native SQLite3 shared libraries for all .NET platforms. Add a build property that uses the system SQLite3 library PCL and enable it by default for FreeBSD.

### DIFF
--- a/MSBuild/Robust.Properties.targets
+++ b/MSBuild/Robust.Properties.targets
@@ -16,6 +16,11 @@
         <ActualOS>MacOS</ActualOS>
       </PropertyGroup>
     </When>
+    <When Condition="$([MSBuild]::IsOSPlatform('FreeBSD'))">
+      <PropertyGroup>
+        <ActualOS>FreeBSD</ActualOS>
+      </PropertyGroup>
+    </When>
     <Otherwise>
       <PropertyGroup>
         <ActualOS>Linux</ActualOS>
@@ -30,5 +35,6 @@
     <EnableClientScripting>True</EnableClientScripting>
     <!-- Client scripting is disabled on full release builds for security and size reasons. -->
     <EnableClientScripting Condition="'$(FullRelease)' == 'True'">False</EnableClientScripting>
+    <UseSystemSqlite Condition="'$(TargetOS)' == 'FreeBSD'">True</UseSystemSqlite>
   </PropertyGroup>
 </Project>
--- a/Robust.Client/Robust.Client.csproj
+++ b/Robust.Client/Robust.Client.csproj
@@ -14,7 +14,9 @@
   <ItemGroup>
     <PackageReference Include="DiscordRichPresence" Version="1.0.175" />
     <PackageReference Include="JetBrains.Annotations" Version="2021.3.0" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.Data.Sqlite" Version="6.0.2" />
+    <PackageReference Include="Microsoft.Data.Sqlite.Core" Version="6.0.9" />
+    <PackageReference Include="SQLitePCLRaw.provider.sqlite3" Version="2.1.2" Condition="'$(UseSystemSqlite)' == 'True'" />
+    <PackageReference Include="SQLitePCLRaw.bundle_e_sqlite3" Version="2.1.2" Condition="'$(UseSystemSqlite)' != 'True'" />
     <PackageReference Include="nfluidsynth" Version="0.3.1" />
     <PackageReference Include="NVorbis" Version="0.10.1" />
     <PackageReference Include="SixLabors.ImageSharp" Version="2.1.3" />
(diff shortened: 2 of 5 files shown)


Robust.Client/Robust.Client.csproj
```diff
@@ -24,6 +23,19 @@
...
+        <PackageReference Include="Microsoft.Data.Sqlite.Core" Version="6.0.8" />
+        <PackageReference Include="SQLitePCLRaw.provider.sqlite3" Version="2.1.1" />
+      </ItemGroup>
+    </When>
+    <Otherwise>
+      <ItemGroup>
+        <PackageReference Include="Microsoft.Data.Sqlite" Version="6.0.2" />
+      </ItemGroup>
+    </Otherwise>
+  </Choose>
```
PJB3005: Simpler way to do it.

```suggestion
    <PackageReference Include="Microsoft.Data.Sqlite.Core" Version="6.0.9" />
    <PackageReference Include="SQLitePCLRaw.provider.sqlite3" Version="2.1.2" Condition="'$(UseSystemSqlite)' == 'True'" />
    <PackageReference Include="SQLitePCLRaw.bundle_e_sqlite3" Version="2.1.2" Condition="'$(UseSystemSqlite)' != 'True'" />
  </ItemGroup>
```

Robust.Server/Robust.Server.csproj
```diff
@@ -13,12 +13,24 @@
...
+        <PackageReference Include="Microsoft.Data.Sqlite.Core" Version="6.0.8" />
+        <PackageReference Include="SQLitePCLRaw.provider.sqlite3" Version="2.1.1" />
+      </ItemGroup>
+    </When>
+    <Otherwise>
+      <ItemGroup>
+        <PackageReference Include="SQLitePCLRaw.bundle_e_sqlite3" Version="2.0.7" />
+      </ItemGroup>
+    </Otherwise>
+  </Choose>
```
PJB3005: Simpler way to do it.

```suggestion
    <PackageReference Include="Microsoft.Data.Sqlite.Core" Version="6.0.9" />
    <PackageReference Include="SQLitePCLRaw.provider.sqlite3" Version="2.1.2" Condition="'$(UseSystemSqlite)' == 'True'" />
    <PackageReference Include="SQLitePCLRaw.bundle_e_sqlite3" Version="2.1.2" Condition="'$(UseSystemSqlite)' != 'True'" />
  </ItemGroup>
```

MSBuild/Robust.Properties.targets
```diff
@@ -30,5 +30,6 @@
     <EnableClientScripting>True</EnableClientScripting>
     <!-- Client scripting is disabled on full release builds for security and size reasons. -->
     <EnableClientScripting Condition="'$(FullRelease)' == 'True'">False</EnableClientScripting>
+    <UseSystemSqlite Condition="$([MSBuild]::IsOSPlatform('FreeBSD'))">True</UseSystemSqlite>
```
PJB3005: This is going to check for the host system being built on, I think. It would probably be better to check the targeted [RID](https://learn.microsoft.com/en-us/dotnet/core/rid-catalog) instead and see if it's based on FreeBSD, though I don't know the best way to do that (I guess a shoddy way would be a string contains).

Robust.Client/Robust.Client.csproj
Partmedia: Thanks for this suggestion. Let me double check that the version bump doesn't break anything, and I'll update the PR.

Robust.Server/Robust.Server.csproj
Partmedia: I've force-pushed with the suggested change; this works fine on FreeBSD.

Since this adds Microsoft.Data.Sqlite.Core and bumps SQLitePCLRaw.provider.sqlite3 to 2.1.2 on the non-system SQLite build, could you check that this doesn't mess up something else (e.g. publish, etc.) on the other platforms?

Robust.Client/Robust.Client.csproj
Partmedia: I've force-pushed with the suggested change; this works fine on FreeBSD.

For non-system SQLite platforms, this changes Microsoft.Data.Sqlite 6.0.2 -> Microsoft.Data.Sqlite.Core 6.0.8 and adds QLitePCLRaw.bundle_e_sqlite3 2.1.2. Would you mind checking that this doesn't do anything surprising on the other platforms?

MSBuild/Robust.Properties.targets
Partmedia: I agree. I'll update this with a 'freebsd' string contains check.

Partmedia: I opted to use `TargetOS`, and added an `ActualOS` check for FreeBSD based on `[MSBuild]::IsOSPlatform('FreeBSD')` so that `dotnet build` (without any a RID) continues to work. Doing this is also compatible with the build script. Force pushed and updated.

PR #3311, opened 2022-10-09
Got Inserted Event for containers

keronshb:
This event is raised on the entity that got inserted. It's good for entities that need to run code when they're inserted, but not on an attempt or on the owner of the container.

Needed for https://github.com/space-wizards/space-station-14/pull/11770

### DIFF
--- a/Robust.Shared/Containers/BaseContainer.cs
+++ b/Robust.Shared/Containers/BaseContainer.cs
@@ -127,6 +127,9 @@ namespace Robust.Shared.Containers
             transform.LocalPosition = Vector2.Zero;
             transform.LocalRotation = Angle.Zero;
 
+            var gotInsertedEvent = new GotInsertedEvent(toinsert, this);
+            entMan.EventBus.RaiseLocalEvent(toinsert, gotInsertedEvent, true);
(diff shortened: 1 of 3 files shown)


Robust.Shared/Containers/BaseContainer.cs
```diff
@@ -127,6 +127,9 @@ public abstract class BaseContainer : IContainer
             transform.LocalPosition = Vector2.Zero;
             transform.LocalRotation = Angle.Zero;
 
+            var gotInsertedEvent = new GotInsertedEvent(toinsert, this);
+            entMan.EventBus.RaiseLocalEvent(toinsert, gotInsertedEvent, true);
```
ElectroJr: While the class names are somewhat longwinded, the new event name should be consistent with all the other container events, i.e., EntGotInsertedIntoContainerMessage, so that it matches EntInsertedIntoContainerMessage, EntRemovedFromContainerMessage, and EntGotRemovedFromContainerMessage.

The event should also be raised at the same point as EntInsertedIntoContainerMessage.

ElectroJr (approved):

PR #3312, opened 2022-10-09
Fix saving entities without prototypes

ElectroJr:
Fixes a bug related to #3279


mirrorcult:
this allows it to be un-reverted, yes?

mirrorcult (approved):

ElectroJr:
> this allows it to be un-reverted, yes?

As far as I know. I was testing the GridId PR with this but got side-tracked by heisenbugs so haven't actually tested yet.

PR #3313, opened 2022-10-09
Unrevert "Remove obsolete GridId and all references to it" (#3279)

ElectroJr:
Unreverts #3279.

AFAIK the only issue was the maps serialization, which should have been fixed by #3312 and commit fc67de0640c192450fdc8faacbc235fdbdd80ed3.

I also made some other slight changes in commit d529973cf16da6b77a6d966ae3c88504fd8d82dc so that the grid index can eventually just be removed altogether, but that requires existing maps to be updated


SpaceManiac:
Thanks

PR #3315, opened 2022-10-09
engine changes for localization cvar

Elijahrane:
It refused to overwrite the default locale if it was already written once. Not quite sure why, I didn't notice any ill effects testing. Requires https://github.com/space-wizards/space-station-14/pull/11807.


Elijahrane:
In case localization not being a cvar was more than just 'no one coded it', I have a few notes here:
1. Clients load the default language first before they connect to a server, then they'll load a second language if the server language is different. AFAICT this is only because of the connection screen, which is... unused outside of development?
2. I'm not aware of what small perf effects loading 2 locales might have, but the benefits to ease of translations probably outweigh it

Some questions:
1. Is the connect menu something that end users are ever going to see? Is there another reason clients need to load up en_US before they even connected?
2. Why is _contexts a dictionary? By that I mean, is there a use case for loading multiple locs at once? If it's fine to unload the other locales that could save a bit of memory, which I guess is the primary concern other than clients loading a loc and then immediately loading a different one when they join a server.

PR #3323, opened 2022-10-10
Use invariant culture in serialization

igorsaux:
Fixes serialization when using locales with comma separated floats


PJB3005 (commented):
Damn I thought I went through and fixed these at one point.

I made a bunch of helpers in the `Parse` class back then, so it's just `Parse.TryFloat()` instead of the whole 60-column mess for doing it otherwise. Maybe switch over to those?

igorsaux:
@PJB3005 Done!

PaulRitter (approved):

PR #3327, opened 2022-10-11
required changes for content benchmarks

PaulRitter:
the pathoffset is needed due to benchmarkdotnet nesting its working directory in like 5 different layers. there probably is a more elegant solution, but this was hastily shitcoded.


PaulRitter:
pj said its gucci

PR #3332, opened 2022-10-13
Don't allocate a list per physics contact

metalgearsloth:
Forgot to get around to this. no.2 on server allocs on tumbler testbed.

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs
+++ b/Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs
@@ -197,8 +197,8 @@ namespace Robust.Shared.Physics.Systems
                     if (proxyA.Fixture.Hard && other.Fixture.Hard &&
                         (gridMoveBuffer.ContainsKey(proxyA) || gridMoveBuffer.ContainsKey(other)))
                     {
-                        proxyABody.WakeBody();
-                        otherBody.WakeBody();
+                        _physicsSystem.WakeBody(proxyABody);
+                        _physicsSystem.WakeBody(otherBody);
                     }
 
                     contactManager.AddPair(proxyA, other);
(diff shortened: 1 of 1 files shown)


Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs
```diff
@@ -324,35 +324,35 @@ internal void FindNewContacts(SharedPhysicsMapComponent component, MapId mapId)
...
                 DebugTools.Assert(other.Fixture.Body.CanCollide);
                 // Logger.DebugS("physics", $"Checking {proxy.Fixture.Body.Owner} against {other.Fixture.Body.Owner} at {aabb}");
 
-                // Do fast checks first and slower checks after (in ContactManager).
-                if (proxy == other ||
-                    proxy.Fixture.Body == other.Fixture.Body ||
-                    !ContactManager.ShouldCollide(proxy.Fixture, other.Fixture)) continue;
+                if (tuple.proxy == other ||
+                    !ContactManager.ShouldCollide(tuple.proxy.Fixture, other.Fixture) ||
+                    tuple.proxy.Fixture.Body == other.Fixture.Body)
```
ElectroJr, line 339: why was the ordering flipped? I guess ShouldCollide is just bitmask check now, did it use to be slower?

ElectroJr (approved):

Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs
metalgearsloth: oop didn't see this in time.

Most of the time I think the mask itself is going to be the fail, at least in ss14, as most bodies are 1 fixture. The only exceptions to this off the top of my head are grids (which are special cased separately) and bullet flybys.

PR #3334, opened 2022-10-13
Entity query struct enumerators

metalgearsloth:
They're based.

Changed over the ones that were top of memory profile on debug. Content PR incoming.

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.Components.cs
+++ b/Robust.Shared/GameObjects/EntityManager.Components.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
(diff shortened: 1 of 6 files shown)


moonheart08:
Should we deprecate EntityQuery<>? I don't see a reason to keep it beyond ergonomics, and the footgun of alloc being avoided permanently would be nice.

metalgearsloth:
> Should we deprecate EntityQuery<>? I don't see a reason to keep it beyond ergonomics, and the footgun of alloc being avoided permanently would be nice.

Depends if we go ahead with flecs or not; if we are I wouldn't bother with the effort to refactor given both would probably get replaced. I just got these because they're several of the top 10 allocs on client and it only took me an hour.

PR #3335, opened 2022-10-13
Use static lambdas for sprites / lights

metalgearsloth:
Client didn't seem to explode when I ran it.


Ygg01:
This isn't a breaking change, right? One dependency added but no outside changes

metalgearsloth:
> This isn't a breaking change, right? One dependency added but no outside changes

nope

PR #3338, opened 2022-10-14
makes some loops for in pvs for free perf

PaulRitter:


Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -809,7 +809,7 @@ public void ReturnToPool(HashSet<int>[] playerChunks)
             return null;
 
         var leftView = new List<EntityUid>();
-        foreach (var uid in lastSent.Keys)
+        foreach (var (uid, _) in lastSent)
```
ElectroJr: Is this actually faster? I just always assumed that iterating over only keys instead of keys & values would be better. Or that at the very least the compiler would be smart enough to convert it into (keys, _) internally if it werent.

ElectroJr (approved):

Robust.Server/GameStates/PVSSystem.cs
PaulRitter: old change that snuck in there, but since you asked, i got curious:

grafik
as it turns out, key & valuecollections are actually **faster** than throwing away one value. will fix!

mirrorcult (approved):

PR #3346, opened 2022-10-17
Move CastShadows to SharedPointLightComponent

Partmedia:
This change makes it possible for the server to set CastShadows to false. Shadow casting is a relatively expensive operation that can be skipped for some lights.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
@@ -37,6 +37,12 @@ namespace Robust.Shared.GameObjects
         [DataField("softness")]
         private float _softness = 1f;
 
+        /// <summary>
+        ///     Whether this pointlight should cast shadows
+        /// </summary>
+        [DataField("castShadows")]
+        public bool CastShadows = true;
+
         [ViewVariables(VVAccess.ReadWrite)]
         public virtual bool Enabled
         {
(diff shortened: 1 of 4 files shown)


Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
```diff
@@ -37,6 +37,12 @@ public abstract class SharedPointLightComponent : Component
         [DataField("softness")]
         private float _softness = 1f;
 
+        /// <summary>
+        ///     Whether this pointlight should cast shadows
+        /// </summary>
+        [DataField("castShadows")]
+        public bool CastShadows = true;
+
```
ElectroJr, line 45: This needs to be made part of the `PointLightComponentState`, otherwise it doesn't actually get sent to clients.
Also the component has to be `Dirty()`ed every time this gets set anywhere.

ElectroJr (changes requested):

Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
Partmedia: Interesting, then I wonder how this ever worked when I tested it locally? In any case, I'll make the fix.

Partmedia: My testing (via a wheat bushel prototype) leaked CastShadows. Fixed via force push and really tested this time.

PR #3351, opened 2022-10-18
makes entity nullable for commandmessages

PaulRitter:
requires https://github.com/space-wizards/space-station-14/pull/12030


PJB3005:
I know we don't care much about network bandwidth, but this would somewhat worsen bandwidth usage of these messages.

PR #3354, opened 2022-10-20
makes all prototypes readonly record structs

PaulRitter:
also forces future prototypes to be structs

requires https://github.com/space-wizards/space-station-14/pull/12060


mirrorcult:
based

PJB3005:
What? Why?

PR #3360, opened 2022-10-22
PVS budget tweaks

ElectroJr:
Currently PVS limits the number of entities that get sent to players if they have newly entered their view. This is mostly done to limit stuttering due to entity spawning. This PR separates that limit into a new-entity and entered-entity budget, which should help reduce pop-in a bit when moving around previously seen areas. PVS-entry still has a limit because even though it is much faster than entity spawning, it will still cause the client to stutter if it has to re-parent 1000+ entities in a single tick. 

This also fixes a bug where PVS would sometimes incorrectly detach entities when the budget gets overwhelmed. To cause the bug on current master, just set the entity budget to 1, tickrate to 30, and walk around. You will notice that the player's own clothes occasionally disappear for a bit, despite the fact that they are always in PVS range.

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -685,8 +685,10 @@ internal sealed partial class PVSSystem : EntitySystem
         EntityUid[] viewerEntities)
     {
         DebugTools.Assert(session.Status == SessionStatus.InGame);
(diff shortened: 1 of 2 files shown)


metalgearsloth:
@PaulRitter

ElectroJr:
> @ PaulRitter

As paul pointed out on discord, these were actually already added earlier (e.g., they were present in #2593) and then removed in #2733 because of last-ack issues so I'll interpret that as approval re-adding them in this PR

PaulRitter:
ye sorry for not responding in time. good pr, sgtm

PR #3368, opened 2022-10-23
More Transform ECS

### LINKED ISSUE #3260: EntParentChanged messages shouldn't go out on initialising entities
It means we need a bunch of lifetime checks everywhere which is yucky.

ElectroJr:
Moves the `EntityCoordinates` setter and `AttachParent()` to system functions. Also reorganizes state handling logic a bit.

Misc changes that could have side effects:
- The coordinate setter now unanchors anchored entities when used, not just when changing parents.
- Movement, anchor, and parent change events are no longer raised before the transform component has initialized (fixes #3260)
  - Because of shitcode elsewhere that relied on this, the parent & anchor events are now raised explicitly during component startup
  - Eventually this should be removed


Robust.Shared/Map/MapManager.MapCollection.cs
```diff
@@ -273,6 +273,7 @@ protected MapId CreateMap(MapId? mapId, EntityUid entityUid)
 
                 var mapComp = EntityManager.AddComponent<MapComponent>(newEnt);
                 mapComp.WorldMap = actualId;
+                EntityManager.Dirty(mapComp);
```
ElectroJr, line 276: I have NFI why this is required, but everything started breaking and this fixed it......

PR #3369, opened 2022-10-23
Lookup movement handling changes.

ElectroJr:
- Removes the parent-change subscription in favour of just checking for parent changes in move events.
- Replaces GetMoverCoordinate() with the override that simultaneously returns the world rotation.


metalgearsloth (approved):

Robust.Shared/Physics/B2DynamicTree.cs
```diff
@@ -583,6 +583,9 @@ private void Balance(Proxy index)
                 indexNode.Height = Math.Max(child1Node.Height, child2Node.Height) + 1;
                 indexNode.Aabb = child1Node.Aabb.Union(child2Node.Aabb);
 
+                if (index == indexNode.Parent)
+                    throw new Exception($"Infinite loop in {nameof(B2DynamicTree)}.{nameof(Balance)}.");
+
```
ElectroJr: Is there anything wrong with this check, is it true that this should never be the case?

ElectroJr: Ill just move this to a separate PR

ElectroJr:
Finally passes tests and has one slightly outdated approval, so imma just merge

PR #3373, opened 2022-10-25
Obsolete .Owner

metalgearsloth:
If this is something we intend to do to make our engine not run like shit better to avoid the problem getting worse.


moonheart08:
How does this work with EntityQuery.

metalgearsloth:
Likely EntityQuery would return entityuid as its first arg and the current EntityQuery becomes ComponentQuery without returning the uid.

DrSmugleaf (approved):
Assuming it doesn't crash the Rider warning view

metalgearsloth:
I'll re-open it soon-ish with the updated API and after bringing down the outstanding prs a bit.

Issue #2273, opened 2021-11-26
Fluent Text type


gradientvera:
As discussed on the October 30th meeting, we should have a fluent text type to allow different locales on the server at the same time.

Efruit:
There's a bit of foundation for this in #2213.

Formatted text `Section`s have the option of including a `Localized` flag. The idea is that there would be an `ISectionable` proxy that sits in front of the real `ISectionable` (probably a `FormattedMessage`). If the proxy sees that flag set in the original message, it would be able to transparently replace the `Section`'s content with the localized version before it hits the layout or display system.

No provisions for arguments yet since that's serialization stuff and that's _Not My Department™_... not that anyone would trust code I write for it anyway.

PR #3374, opened 2022-10-25
Localize commands

Ygg01:
A continuation of #2331 and stepping stone for #2273.

Just without too much complication. Rather than a custom data structure, adds an abstract `LocalizeCommand` that preserves `IConsoleCommand` interface.
`LocalizeCommand` standardizes the `IConsoleCommand` localization to `cmd-{Command}-{help/desc}` for `Help` and `Description` respectively.

Related to https://github.com/space-wizards/space-station-14/issues/5543 (Expect a commit there too as well).

Non-breakable change at first glance.


Ygg01:
Hrm, why is entityQuery failing on transform in Content.IntegrationTests?

NVM. It was some intermittent failure.

Resources/Locale/en-US/commands.ftl
```diff
@@ -159,3 +159,339 @@ cmd-hint-savebp-id = <Grid EntityID>
...
+cmd-ldrsc-help = Usage: ldrsc <path> <type>
+
+cmd-rldrsc-desc = Reloads a resource.
+cmd-rldrsc-help = Usage: rldrsc <path> <type>
+
+cmd-gridtc-desc = Gets the tile count of a grid.
+cmd-gridtc-help = Usage: gridtc <gridId>
+
+
+# Client site commands
```
PJB3005:
```suggestion
# Client-side commands
```

Robust.Shared/Console/LocalizedCommands.cs
```diff
@@ -0,0 +1,20 @@
...
+    [Dependency] private readonly ILocalizationManager _loc = default!;
+
+    /// <inheritdoc />
+    public abstract string Command { get; }
+
+    /// <inheritdoc />
+    public string Description => _loc.TryGetString($"cmd-{Command}-desc", out var val) ? val : "";
+
+    /// <inheritdoc />
+    public string Help => _loc.TryGetString($"cmd-{Command}-help", out var val) ? val : "";
```
PJB3005: These should stay virtual IMO. Some commands dynamically build the help body up from enum members etc.

Ygg01: Thanks, that solved some problems with VV Commands.

Here is when I point SS14/RobustToolbox to this branch.
Content Client_pa77AcGAa6

ElectroJr (changes requested):
This currently breaks completion hints (e.g., see `/help`'s lack of completions)

PR #3388, opened 2022-10-27
Remove Proxy.Free getter

ElectroJr:
Unless there is some reason this was a getter rather than just a static readonly struct.

### DIFF
--- a/Robust.Shared/Physics/DynamicTree.Proxy.cs
+++ b/Robust.Shared/Physics/DynamicTree.Proxy.cs
@@ -12,11 +12,7 @@ namespace Robust.Shared.Physics
 
             private readonly int _value;
 
-            public static Proxy Free
-            {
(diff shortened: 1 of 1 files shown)


PJB3005:
I want to say that making it a dead-simple getter is easier for the JIT to inline etc. But I don't have tests for that.

I don't see any reason to change it either way, unless there's evidence that this helps the JIT in some way.

PJB3005:
I can't be arsed to ask on the C# discord, but my hunch is that the static constructor created by this change is *significantly* more complex to deal with for the JIT and requires tiered compilation to be able to inline it.

ElectroJr:
Huh, I would've though it'd be the other way around.

wrexbe:
inlining
Tried it for fun, the get does inline easier then then the static readonly.

Issue #3387, opened 2022-10-27
GetEntitiesIntersecting is returning deleted entities.


ElectroJr:
Just go around blowing up parts of the station:
```
[FATL] unhandled: System.Collections.Generic.KeyNotFoundException: Entity 13685 does not have a component of type Robust.Shared.GameObjects.TransformComponent
   at Robust.Shared.GameObjects.SharedTransformSystem.DeparentAllEntsOnTile(EntityUid gridId, Vector2i tileIndices) in space-station-14\RobustToolbox\Robust.Shared\GameObjects\Systems\SharedTransformSystem.cs:line 88
   at Robust.Shared.GameObjects.SharedTransformSystem.MapManagerOnTileChanged(TileChangedEvent e) in space-station-14\RobustToolbox\Robust.Shared\GameObjects\Systems\SharedTransformSystem.cs:line 64
   at Robust.Shared.GameObjects.EntityEventBus.<>c__DisplayClass4_0`1.<SubscribeEvent>b__0(Unit& ev) in space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntityEventBus.Broadcast.cs:line 169
   at Robust.Shared.GameObjects.EntityEventBus.ProcessSingleEventCore(EventSource source, Unit& unitRef, EventData subs, Boolean byRef) in space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntityEventBus.Broadcast.cs:line 324
   at Robust.Shared.GameObjects.EntityEventBus.RaiseLocalEventCore(EntityUid uid, Unit& unitRef, Type type, Boolean broadcast, Boolean byRef) in space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntityEventBus.Directed.cs:line 223
   at Robust.Shared.GameObjects.EntityEventBus.RaiseLocalEvent[TEvent](EntityUid uid, TEvent args, Boolean broadcast) in space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntityEventBus.Directed.cs:line 179
   at Robust.Shared.Map.MapManager.RaiseOnTileChanged(TileRef tileRef, Tile oldTile) in space-station-14\RobustToolbox\Robust.Shared\Map\MapManager.GridCollection.cs:line 247
[...]

PR #3390, opened 2022-10-27
Lookup fixes

ElectroJr:
- Body type and collision changed handling wasn't checking if entities were in containers, causing all contained entities to be added to the sundries tree
- Some is-in-container checks should have been is-ent-or-parent-in-container.
- One `StaticTree` query used the wrong bitflag check.
- collision-less anchored entities were added to the sundries tree, instead of the static tree
  - This is incompatible with old behaviour where excluding anchored entities would  uuhhh well exclude all anchored entities, leading to issues with storage lockers eating pipes
  - Adding collisionless entities to the static tree made physics cry, so This PR adds a StaticSundries tree instead.
  - There might be a better way of fixing this eventually, but I dunno enough and at least this fixes the issue for now.

This does NOTfix #3387, so there are more issues somewhere.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
@@ -1,9 +1,6 @@
-using System;
-using System.Collections.Generic;
 using JetBrains.Annotations;
 using Robust.Shared.Configuration;
 using Robust.Shared.Containers;
-using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
 using Robust.Shared.Map;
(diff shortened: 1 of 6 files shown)


Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -239,41 +242,72 @@ private void OnPhysicsUpdate(ref CollisionChangeEvent ev)
...
+
+            if (!ev.CanCollide && _container.IsEntityOrParentInContainer(ev.Body.Owner, null, xform, null, xformQuery))
+            {
+                // getting inserted, skip sundries insertion and just let container insertion handle tree removal.
+
+                // TODO: for whatever fucking cursed reason, this is currently required.
+                // FIX THIS, this is a hotfix
+                var b = GetBroadphase(ev.Body.Owner, xform, GetEntityQuery<BroadphaseComponent>(), xformQuery);
+                if (b != null)
+                    RemoveBroadTree(ev.Body, b, ev.Body.BodyType);
```
ElectroJr, line 256: This is cursed, but id rather just hotfix it than leave master broken for longer

PR #3392, opened 2022-10-27
Replaces ContainsEntity() with metadata flag check

ElectroJr:
Replaces a `ContainerManagerComponent.ContainsEntity()` check with a metadata-flag check. If an entity has a container, it probably stores more than one entity and/or has more than one container, so checking the metadata comp should be faster than doing a List.Contains() foreach entity.

Also adds metadata & ContainerManagerComponent queries.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
@@ -126,11 +126,13 @@ namespace Robust.Shared.GameObjects
 
             var coordinates = _transform.GetMoverCoordinates(xform.Coordinates, xformQuery);
             var lookupRotation = _transform.GetWorldRotation(lookup.Owner, xformQuery);
-            // If we're contained then LocalRotation should be 0 anyway.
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -531,7 +538,8 @@ private void OnContainerInsert(EntInsertedIntoContainerMessage ev)
             bool recursive = true)
         {
             var coordinates = _transform.GetMoverCoordinates(xform.Coordinates, xformQuery);
-            // If we're contained then LocalRotation should be 0 anyway.
+            var metaQuery = GetEntityQuery<MetaDataComponent>();
+            var contQuery = GetEntityQuery<ContainerManagerComponent>();
             var aabb = GetAABBNoContainer(xform.Owner, coordinates.Position, _transform.GetWorldRotation(xform, xformQuery) - lookupRotation);
             AddToEntityTree(lookup, xform, aabb, xformQuery, lookupRotation, recursive);
```
metalgearsloth:
```suggestion
            AddToEntityTree(lookup, xform, aabb, xformQuery, metaQuery, contQuery, lookupRotation, recursive);
```

PR #3396, opened 2022-10-28
speeds up gamestate a bit by optimizing collection usage

PaulRitter:
requires https://github.com/space-wizards/space-station-14/pull/12263

way too many hashsets, had to cut the usage down a bit. will reduce allocations as well as general runtime due to less overhead

### DIFF
--- a/Robust.Client/GameStates/ClientGameStateManager.cs
+++ b/Robust.Client/GameStates/ClientGameStateManager.cs
@@ -266,7 +266,7 @@ namespace Robust.Client.GameStates
                     _processor.UpdateFullRep(curState, _entities);
                 }
 
-                IEnumerable<EntityUid> createdEntities;
+                List<EntityUid> createdEntities;
                 using (_prof.Group("ApplyGameState"))
                 {
                     if (_timing.LastProcessedTick < targetProcessedTick && nextState != null)
@@ -285,7 +285,7 @@ namespace Robust.Client.GameStates
                     try
                     {
 #endif
-                    createdEntities = ApplyGameState(curState, nextState);
+                        createdEntities = ApplyGameState(curState, nextState);
 #if EXCEPTION_TOLERANCE
                     }
                     catch (MissingMetadataException e)
@@ -548,7 +548,7 @@ namespace Robust.Client.GameStates
         ///     initial server state for any newly created entity. It does this by simply using the standard <see
         ///     cref="IEntityManager.GetComponentState(IEventBus, IComponent)"/>.
         /// </remarks>
-        private void MergeImplicitData(IEnumerable<EntityUid> createdEntities)
+        private void MergeImplicitData(List<EntityUid> createdEntities)
         {
             var outputData = new Dictionary<EntityUid, Dictionary<ushort, ComponentState>>();
             var bus = _entityManager.EventBus;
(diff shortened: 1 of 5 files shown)


Robust.Client/GameStates/ClientGameStateManager.cs
```diff
@@ -606,21 +606,56 @@ private IEnumerable<EntityUid> ApplyGameState(GameState curState, GameState? nex
...
+
+            // Check next state
+            if (nextState != null)
+            {
+                foreach (var es in nextState.EntityStates.Span)
+                {
+                    var uid = es.Uid;
+
+                    if (!metas.TryGetComponent(uid, out var meta))
+                        continue;
```
ElectroJr, line 629: This check has to be done after first creating new entities, to ensure that things like newly spawned bullets properly lerp during the first tick. Otherwise they weirdly stay in place before starting to move.

Robust.Client/GameStates/ClientGameStateManager.cs
```diff
@@ -636,63 +671,35 @@ private IEnumerable<EntityUid> ApplyGameState(GameState curState, GameState? nex
...
+            foreach (var (uid, state) in nextStates)
+            {
+                toApply.Add(new ToApplyEntityData(uid, false, GameTick.Zero, null, state));
+            }
+
             // Create new entities
             if (toCreate.Count > 0)
             {
                 using var _ = _prof.Group("Create uninitialized entities");
                 _prof.WriteValue("Count", ProfData.Int32(toCreate.Count));
```
ElectroJr, line 689: The actual entity creation profile block is empty now? Given that if any entities are being created, then the majority of the `foreach (var es in curSpan)` loop will be spent on entity creation, I guess this profiler group can just be moved around that for loop?

Having entity creation be separated out in the client profiler isn't really necessary, its just nice to be able to see the breakdown for what's causing stuttering.

PaulRitter: i was under the impression that the profiler group doesnt track time spent in the group, but only the values written to it? or am i mistaken here?

PaulRitter: nvm, i was wrong. will fix

ElectroJr: For example, entity creation now shows up as "0" when its really the majority of the time spent in applying state:
state data

Also oh boy the lookup changes degraded detach to null and pvs-entry performance a smidgen

Robust.Client/GameStates/ClientGameStateManager.cs
```diff
@@ -606,21 +606,56 @@ private IEnumerable<EntityUid> ApplyGameState(GameState curState, GameState? nex
...
+                {
+                    var uid = es.Uid;
+
+                    if (!metas.TryGetComponent(uid, out var meta))
+                        continue;
+
+                    // Does the next state actually have any future information about this entity that could be used for interpolation?
+                    if (es.EntityLastModified != nextState.ToSequence)
+                        continue;
+
```
ElectroJr, line 634: AFAIK, this loop will `continue` most of the time, so AFAIK `new Dictionary(nextState.EntityStates.Span.Length)` significantly overestimates the size of the dict most of the time, but I have no idea if that's actually all that important.

ElectroJr (changes requested):
Seems fine other than reordering the next-state getting to happen after ent creation, and either removing the ent creation profiler group or fixing it.

Issue #3395, opened 2022-10-28
B2DynamicTree.Balance() can get stuck in infinite loops


ElectroJr:
E.g., this is happening in tests in #3369, where index, parent, child1, and child2 are all "{0}". The cause might be some other bug, but either way the fact that this can happen makes it very difficult to debug what is going wrong, and I'm pretty sure this was responsible for some client crashes that were happening earlier as well.

PR #3398, opened 2022-10-28
Add exception for infinite loop in Balance()

ElectroJr:
Dunno enough about how the dynamic tree works to know if there's any reason you'd ever have `index == indexNode.Parent` but it seems to work fine with this assert?

Related to #3395, but this obviously doesn't fix the issue.

### DIFF
--- a/Robust.Shared/Physics/B2DynamicTree.cs
+++ b/Robust.Shared/Physics/B2DynamicTree.cs
@@ -583,10 +583,8 @@ namespace Robust.Shared.Physics
                 indexNode.Height = Math.Max(child1Node.Height, child2Node.Height) + 1;
                 indexNode.Aabb = child1Node.Aabb.Union(child2Node.Aabb);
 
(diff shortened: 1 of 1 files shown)


Robust.Shared/Physics/B2DynamicTree.cs
```diff
@@ -583,6 +583,11 @@ private void Balance(Proxy index)
                 indexNode.Height = Math.Max(child1Node.Height, child2Node.Height) + 1;
                 indexNode.Aabb = child1Node.Aabb.Union(child2Node.Aabb);
 
+#if DEBUG
+                if (index == indexNode.Parent)
+                    throw new Exception($"Infinite loop in B2DynamicTree.Balance(). Trace: {Environment.StackTrace}");
+#endif
```
moonheart08: Shouldn't be if DEBUG, this branch is very cheap.

Issue #3376, opened 2022-10-25
MIDI broken again


S1ss3l:
So yea, I can't play nor hear midi. Yes I have libfluidsynth-dev, qsynth, fluidsynth all the synth.
I'm on Pop-OS 22.04 (Ubuntu based distro).
What I tried : 
- Installing libfluidsynth-dev (v2.2.5-1), libfluidsynth3 (v2.2.5-1), qsynth (v0.9.6-1)
- setting GLIBC_TUNABLES=glibc.rtld.dynamic_sort=1 as environment variable
- Downgrading glibc (package is named libc6 for my distro and I couldnt find any version prior to 2.35 still online)
- Doing all of the above on a fresh install (Yes I wanna hear music that badly)
- Praying ancient deities

So all this is happening on Wisard's Den Lizard [US West] server and for some unholy reason, on Wizard's Den Test [EU West 3???] I have absolutely no problem playing or hearing MIDI files.

Here's an extract of what I thought would be relevant of the logs : 

!!!WARNING!!!: glibc 2.35 or higher detected (you have 2.35).
If anything misbehaves (weird native crashes, library load failures), try setting GLIBC_TUNABLES=glibc.rtld.dynamic_sort=1 as environment variable.
This is a severe glibc bug introduced in glibc 2.35. See https://github.com/space-wizards/RobustToolbox/issues/2563 for details
[DEBG] root: Runtime: .NET 6.0.4 pop.22.04-x64
[DEBG] root: OS: Linux 6.0.2-76060002-generic #202210150739~1666289067~22.04~fe0ce53 SMP PREEMPT_DYNAMIC Thu O X64

[...]

[ERRO] midi: Failed to initialize fluidsynth due to exception, disabling MIDI support:
System.DllNotFoundException: Unable to load shared library '/home/sissel/SS14/bin/loader/libfluidsynth.so.2' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: /home/sissel/SS14/bin/loader/libfluidsynth.so.2: cannot open shared object file: No such file or directory
   at System.Runtime.InteropServices.NativeLibrary.LoadFromPath(String libraryName, Boolean throwOnError)
   at System.Runtime.InteropServices.NativeLibrary.Load(String libraryPath)
[...]

PR #3400, opened 2022-10-29
Add osx-arm64 support

drcatdoctor:
Adds osx-arm64 support. Still in testing stages and this is still very early. Draft for feedback.

In local tests this seems to all work, and I played a full game on Lizard and it worked almost perfectly the entire time and was pegged at 120 FPS... 

but some part of the map had a client-side entity that caused my screen to stop updating entirely. Though the game was still responsive, so I could walk away from the bugged area and everything would resume, it is pretty brutal. Traceback is below under Known Issues. I don't know more than that yet.

Still, things seem promising.

The number one open question of course is: is this a good idea at all, and now that I have it somewhat working, I can see arguments for both sides.

### Changes
- bump native dependencies for Robust.Client
- Robust.Server osx-arm64 now requires Zstd.
- explicitly set the library name of fluidsynth to have a `2` in it to be compatible with the dylib that is packed in the build deps, and to stop problems like https://github.com/space-wizards/RobustToolbox/issues/3376
- update package_client_build.py to add `mac-apple-silicon` 

#### Upstream requirements

- [ ] https://github.com/space-wizards/nativefiledialog/pull/2
- [ ] https://github.com/space-wizards/native-build/pull/4
- [ ] https://github.com/space-wizards/build-dependencies/pull/2
- [ ] https://github.com/space-wizards/SpaceWizards.Sodium/pull/1

### Known issues
- https://github.com/glfw/glfw/issues/1997
This has been happening a long time and also happens under Rosetta - e.g. https://github.com/space-wizards/space-station-14/issues/5186 - and the GLFW maintainers traced it back to bugs introduced by Apple. Doesn't seem like there is much we can do here. Some minor revisions of OSX remove the bug, but other ones seem to bring it back.
- Bugged client-side entity (light?), needs investigation.
```
[ERRO] runtime: Caught exception in "GameLoop Render"
System.Collections.Generic.KeyNotFoundException: Entity c2754 does not have a component of type Robust.Shared.GameObjects.TransformComponent
   at Robust.Client.Graphics.Clyde.Clyde.DrawLightsAndFov(Viewport viewport, Box2Rotated worldBounds, Box2 worldAABB, IEye eye) in .../Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs:line 436
   at Robust.Client.Graphics.Clyde.Clyde.<>c__DisplayClass204_0.<RenderViewport>b__0() in .../Robust.Client/Graphics/Clyde/Clyde.HLR.cs:line 574
   at Robust.Client.Graphics.Clyde.Clyde.RenderInRenderTarget(RenderTargetBase rt, Action a, Nullable`1 clearColor) in .../Robust.Client/Graphics/Clyde/Clyde.HLR.cs:line 446
   at Robust.Client.Graphics.Clyde.Clyde.RenderViewport(Viewport viewport) in .../Robust.Client/Graphics/Clyde/Clyde.HLR.cs:line 575
```
- changing the libsodium version might be a bad idea (https://github.com/jedisct1/libsodium/issues/1075) -- make a native runtime package instead.

### Questions
- Not actually sure if `$(ProcessArchitecture)` is doing what I want? It seems to work locally. Could just remove the conditional, it's not hurting anybody
- Seems like a really ugly place to set the fluidsynth library name but I can't figure out where would be better, suggestions welcome
- Does package_client_build.py do anything? Am I doing it right? I can't really tell.

### DIFF
--- a/.github/workflows/publish-client.yml
+++ b/.github/workflows/publish-client.yml
@@ -26,7 +26,7 @@ jobs:
         dotnet-version: 6.0.100
 
     - name: Package client
-      run: Tools/package_client_build.py -p windows mac linux
+      run: Tools/package_client_build.py -p windows mac-intel mac-apple-silicon linux
 
     - name: Shuffle files around
       run: |
--- a/Robust.Client/Audio/Midi/MidiManager.cs
+++ b/Robust.Client/Audio/Midi/MidiManager.cs
@@ -103,6 +103,12 @@ internal sealed partial class MidiManager : IMidiManager
 
     private const string ContentCustomSoundfontDirectory = "/Audio/MidiCustom/";
 
+    // explicitly look for version 2 where possible, or the fluidsynth module will just pick up anything called 'libfluidsynth'
+    // and we're incompatible with v3.
+    private const string DefaultLibrary = "fluidsynth";
+    private const string LinuxLibrary =  "libfluidsynth.2.so";
+    private const string OsxLibrary = "libfluidsynth.2.dylib";
+
     private const float MaxDistanceForOcclusion = 1000;
 
     private static ResourcePath CustomSoundfontDirectory = new ResourcePath("/soundfonts/");
@@ -148,6 +154,8 @@ internal sealed partial class MidiManager : IMidiManager
             _resourceManager.UserData.Rename(CustomSoundfontDirectory, CustomSoundfontDirectory.WithName(CustomSoundfontDirectory.Filename + ".old"));
             _resourceManager.UserData.CreateDir(CustomSoundfontDirectory);
         }
+
+        DllMapHelper.RegisterExplicitMap(typeof(NFluidsynth.Logger).Assembly, DefaultLibrary, LinuxLibrary, OsxLibrary);
 
         try
         {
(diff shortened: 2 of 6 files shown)


PJB3005:
> Robust.Server osx-arm64 now requires Zlib.

~~Where is this extra dependency coming from?~~

Wait did you mean zstd?

> Not actually sure if $(ProcessArchitecture) is doing what I want? It seems to work locally. Could just remove the conditional, it's not hurting anybody

It's only used for the Robust.Natives.Zstd package so frankly you can just include that dependency always. Would be nicer to be able to get it into the upstream `SharpZstd.Interop` package though.

(The "correct" way would be to use RIDs or something if you really care, but I don't think MSBuild makes that easy)

For the GLFW stuff, I've been planning to move to SDL2 for various reasons. I hope it'll be more reliable for stuff like this. We already have a good chunk of the backend code for it, but it's far from done.

drcatdoctor:
> > Robust.Server osx-arm64 now requires Zlib.
> 
> ~Where is this extra dependency coming from?~
> 
> Wait did you mean zstd?

I meant Zstd. Duh. My mistake. Fixed description.

> > Not actually sure if $(ProcessArchitecture) is doing what I want? It seems to work locally. Could just remove the conditional, it's not hurting anybody
> 
> It's only used for the Robust.Natives.Zstd package so frankly you can just include that dependency always. Would be nicer to be able to get it into the upstream `SharpZstd.Interop` package though.
> 
> (The "correct" way would be to use RIDs or something if you really care, but I don't think MSBuild makes that easy)

Hm it does seem better. Thanks for the pointer, I will look into that more when I have some more time to work on this.

> For the GLFW stuff, I've been planning to move to SDL2 for various reasons. I hope it'll be more reliable for stuff like this. We already have a good chunk of the backend code for it, but it's far from done.

Ok, if that's in progress then maybe it is tolerable that GLFW has some problems for now..

Thanks for the look.

PR #3407, opened 2022-10-29
Cache broadphases and container rejig

### LINKED ISSUE #3402: Broadphase related exceptions
- server-side Index out of bounds while processing `DeparentAllEntsOnTile()`
  - AKA `child1 == Proxy.Free` 
  - Can reliably be caused by just doing a  `/explosion 100000` somewhere on station
- Various bugs that can be triggered by just placing "space" tiles randomly around the station. Which one gets triggered first seems somewhat random.
  - (client-side only?) null broadphase while updating physics contacts
  - (client-side only?) Infinite loops (#3395)
  - (client-side only?) Entity with `CanCollide = false` in the physics trees.
  - (client-side only?) Same index out of bounds as the server's `DeparentAllEntsOnTile()`, seemingly caused by an entity leaving PVS?

ElectroJr:
This PR makes it so that the broadphase that an entity can be found on is cached in the transform component. It could be moved to a separate component, but given that this data is required every time any entity moves and fetching components is somewhat slow, it just makes sense to me to keep it on xform for now.

This should hopefully avoid various broadphase add/remove bugs, as well as making move event handling faster (replaces parent iteration with just a single get-broadphase). This PR also significantly changes container insert-remove logic to avoid redundant broadphase updates & move events.

In the process, there were also several container, pvs, & client state handling logic bugs. So this PR also just touches all of those a bit.. AFAIK these changes are stable? Round restart, explosions, and randomly spacing tiles all seem to work, So I'll claim that this:
Fixes #3402

This also changes how null-space is treated. Previously the client would create a nullspace map, while the server would not, and tests sometimes did and sometimes did not. Now nullspace maps are just illegal. Because of this, this requires a content PR:
Requires  space-wizards/space-station-14/pull/12290


Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -787,6 +785,10 @@ public void ReturnToPool(HashSet<int>[] playerChunks)
 
         foreach (var (uid, visiblity) in visibleEnts)
         {
+            // if an entity is visible, its parents should always be visible. This currently sometimes fails.
+            DebugTools.Assert((tQuery.GetComponent(uid).ParentUid is not { Valid: true } parent) || visibleEnts.ContainsKey(parent),
+                $"Attempted to send an entity without sending it's parents. Entity: {ToPrettyString(uid)}.");
+
```
ElectroJr, line 791: This can currently fail in some rare situations where entities in containers get deleted. In release mode, this just triggers an exception for the client but it manages to recover fine, so this PR should still make everything more stable.

I'm not even sure if this issue was introduced by this PR, or something like #3362 or #3368.

Issue #3403, opened 2022-10-29
Allow for out-of-source-tree builds


rene-descartes2021:
~I tried to build today but had an error as I'd redirected $OutDir to `/tmp/bin/$(MSBuildProjectName)` in Directory.Build.props. Also IntDir similarly changed. In general I often put build outputs into tmpfs for speedup & cleanup.~ ~I'll share this file tomorrow.~ ~**EDIT:** My Directory.Build.props in root of SS14 working copy:~
```msbuild
<Project>
	<PropertyGroup>
		<IntermediateOutputPath>/tmp/cache/MSBuild/obj/</IntermediateOutputPath>
	</PropertyGroup>
	<ImportGroup Label="PropertySheets"/>
	<PropertyGroup>
		<IntDir>/tmp/cache/MSBuild/obj/$(MSBuildProjectName)/</IntDir>
		<OutDir>/tmp/bin/$(MSBuildProjectName)/</OutDir>
	</PropertyGroup>
</Project>
```
~If you want to try it, it shouldn't be checked into repo, I have it ignored in .git/info/exclude, and you can organize with stuff like $(Configuration)~

~I get an error as [this Robust.Client.Injectors.dll path](https://github.com/space-wizards/RobustToolbox/blob/master/MSBuild/XamlIL.targets#L25) doesn't exist with OutDir changed.~
~That dll does exist in `/tmp/bin/Robust.Client.Injectors/` where I'd expect.~

~**EDIT:** Full error:~
```
RobustToolbox/MSBuild/XamlIL.targets|47 col 5| error MSB4062: The "CompileRobustXamlTask" task could not be loaded from the assembly /home/browser/dev/space-station-14/RobustToolbox/MSBuild/../bin/Client//Robust.Client.Injectors.dll. Could not load file or assembly '/home/browser/dev/space-station-14/RobustToolbox/bin/Client/Robust.Client.Injectors.dll'. The system cannot find the path specified. [/home/browser/dev/space-station-14/RobustToolbox/Robust.Client/Robust.Client.csproj]
```

~$OutDir within XamIL.targets is `/tmp/bin/Robust.Client` as it was included from that csproj.~

[...]

PR #3408, opened 2022-10-29
Allow for out-of-source-tree builds

rene-descartes2021:
~Allows build without errors with respect to #3403 but I need to think about it more before I consider this to be a fix.~

~Contingent on [XamlX#1](https://github.com/space-wizards/XamlX/pull/1). **Edit:** This PR build fails without it, that PR copies XamlX and Mono.Cecil assemblies.~

~Changed from AfterCompile to AfterBuild as I suppose copy task for ProjectReferences hadn't run yet. Seems fine. Would prefer links instead of copies to save space.~

~`$(OutDir)` setting in Directory.Build.props appears to override all `$(OutputPath)` settings which I see are used to curate stuff in `bin/stuff/` in both content and robusttoolbox. So that's a consideration, if any other tooling I may use requires that curated structure?~

**EDIT:** Initial proposal had problems. See new comment for new proposal.

### DIFF
--- a/MSBuild/XamlIL.targets
+++ b/MSBuild/XamlIL.targets
@@ -27,9 +27,13 @@
   </PropertyGroup>
 
   <UsingTask
-    Condition="'$(_RobustUseExternalMSBuild)' != 'true' And $(DesignTimeBuild) != true"
+    Condition="'$(_RobustUseExternalMSBuild)' != 'true' And $(DesignTimeBuild) != true And $(BuildPrefix) == ''"
     TaskName="CompileRobustXamlTask"
     AssemblyFile="$(MSBuildThisFileDirectory)\..\Robust.Client.Injectors\bin\$(RobustInjectorsConfiguration)\netstandard2.0\Robust.Client.Injectors.dll"/>
+  <UsingTask
+    Condition="'$(_RobustUseExternalMSBuild)' != 'true' And $(DesignTimeBuild) != true And $(BuildPrefix) != ''"
+    TaskName="CompileRobustXamlTask"
+    AssemblyFile="$(BuildPrefix)\Robust.Client.Injectors\bin\$(Configuration)\netstandard2.0\Robust.Client.Injectors.dll"/>
   <Target
     Name="CompileRobustXaml"
     Condition="Exists('@(IntermediateAssembly)')"
--- a/Avalonia.Base/Avalonia.Base.csproj
+++ b/Avalonia.Base/Avalonia.Base.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\MSBuild\Robust.Engine.props" />
   <Import Project="..\MSBuild\Robust.Common.props" />
 </Project>
--- a/Avalonia.Base/Avalonia.Base.csproj
+++ b/Avalonia.Base/Avalonia.Base.csproj
@@ -1,3 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\MSBuild\Robust.Engine.props" />
+  <Import Project="..\MSBuild\Robust.Common.props" />
 </Project>
--- a/MSBuild/Robust.Common.props
+++ b/MSBuild/Robust.Common.props
@@ -1,4 +1,4 @@
-<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="12.0">
+﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="12.0">
   <PropertyGroup Condition="'$(BuildPrefix)'==''">
     <RobustBuildPrefix>..\bin</RobustBuildPrefix>
     <ContentBuildPrefix>..\bin</ContentBuildPrefix>
--- a/MSBuild/Robust.Common.props
+++ b/MSBuild/Robust.Common.props
@@ -1,0 +1,10 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="12.0">
+  <PropertyGroup Condition="'$(BuildPrefix)'==''">
+    <RobustBuildPrefix>..\bin</RobustBuildPrefix>
+    <ContentBuildPrefix>..\bin</ContentBuildPrefix>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(BuildPrefix)'!=''">
+    <RobustBuildPrefix>$(BuildPrefix)\Content\RobustToolbox\bin</RobustBuildPrefix>
+    <ContentBuildPrefix>$(BuildPrefix)\Content\bin</ContentBuildPrefix>
+  </PropertyGroup>
+</Project>
--- a/MSBuild/Robust.OutOfSourceTree.targets
+++ b/MSBuild/Robust.OutOfSourceTree.targets
@@ -1,0 +1,29 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="12.0">
+  <Target
+    Name="OutOfSourceTreeResources"
+    Condition="'$(BuildPrefix)'!=''"
+    AfterTargets="AfterCompile"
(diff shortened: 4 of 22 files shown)


PJB3005:
> So that's a consideration, if any other tooling I may use requires that curated structure?

The engine pretty strongly relies on hardcoded relative paths to be able to load resources. Putting the final build results somewhere else wouldn't work without code changes.

MSBuild/XamlIL.targets
```diff
@@ -16,17 +16,18 @@
...
-    <ProjectReference Include="$(MSBuildThisFileDirectory)\..\Robust.Client.Injectors\Robust.Client.Injectors.csproj" ReferenceOutputAssembly="false"/>
+    <ProjectReference Include="$(MSBuildThisFileDirectory)\..\Robust.Client.Injectors\Robust.Client.Injectors.csproj" OutputItemType="Content" CopyToOutputDirectory="Always" ReferenceOutputAssembly="false"/>
+    <PackageReference Include="Mono.Cecil" Version="0.11.3" />
   </ItemGroup>
 
   <UsingTask
     Condition="'$(_RobustUseExternalMSBuild)' != 'true' And $(DesignTimeBuild) != true"
     TaskName="CompileRobustXamlTask"
-    AssemblyFile="$(MSBuildThisFileDirectory)\..\Robust.Client.Injectors\bin\$(Configuration)\netstandard2.0\Robust.Client.Injectors.dll"/>
+    AssemblyFile="$(OutDir)Robust.Client.Injectors.dll"/>
```
PJB3005: Copying these files to the output directory isn't a suitable workaround: this would get them mixed in with actual build output.

rene-descartes2021:
**EDIT:** Revised 2022-Nov-05: Refactored to use `Robust.Common.props` and not require changes in most submodules. Also `Resources/` folders copied among a couple others.
**EDIT:** Revised 2022-Nov-07: Edited comment below for clarity/details.
**EDIT:** Revised 2022-Dec-07: Added note on cefglue. Also will make sure will work for OpenDream.
**EDIT:** Revised 2022-Dec-09: Revised tree and example `Directory.Bulid.props`. Cleanup of comment.
**EDIT:** Revised 2023-Jan-04: Revised `Directory.Build.props`. Rebased and squashed as it has been months and there were conflicts. OpenDream and SS14 both appear to work... investigating OpenDream errors.
**EDIT:** Revised 2023-Jan-06: No OpenDream problems when this PR rebased onto the RobustToolbox version used there. So now this PR is ready for review.
**EDIT:** 2023 Apr 22. Added section on rationale. A thought: CMake generator of .net sdk project files might work around some MSBuild shortfalls. Would be alternative to this MSBuild OutputPath redirection, better platform detection and other goodies. Also upcoming newer .NET sdk may include this output redirection through their artifacts thingie.

> > So that's a consideration, if any other tooling I may use requires that curated structure?
> 
> The engine pretty strongly relies on hardcoded relative paths to be able to load resources. Putting the final build results somewhere else wouldn't work without code changes.

Revised to allow for those hardcoded relative paths, without code changes.

Introduced optional `$(BuildPrefix)` specifiable within content `Directory.Build.props`, as follows:
- `$(BuildPrefix)` or a derived variable are used as appropriate within csproj `$(OutputPath)` to ensure all outputs, including the curated/actual build outputs are out of the source tree and under `$(BuildPrefix)`, with expected relative structure.
- Logic to handle `$(BuildPrefix)` is within `MSBuild/Robust.Common.props`.
- ~Note: `$(OutDir)` no longer specified as in former proposal. Due to it overriding later `$(OutputPath)` specifications with csproj files it doesn't make sense to specify `$(OutDir)`. `$(OutputPath)` when specified within the csproj will take precedence over `$(OutputPath)` in `Directory.Build.props`, which results in a satisfactory outcome.~
- Note: There are some feature requests in MSBuild to handle such a feature out of the box, such as this proposed [$(RootOutputPath)](https://github.com/dotnet/sdk/issues/867). When such a feature is implemented that variable name can be used instead of `$(BuildPrefix)`.

Example out-of-source-tree build filesystem-tree, pruned for clarity:
```bash
/tmp/SS14$ tree -L 3
.
├── Avalonia.Base
│   └── bin/Debug/<Framework>  #Most folders below have this substructure
├── Content #Main folder of structured outputs, other build products are top-level
│   ├── bin
│   │   ├── Content.Benchmarks
│   │   ├── Content.Client
│   │   ├── Content.IntegrationTests
│   │   ├── Content.MapRenderer
│   │   ├── Content.Server
│   │   ├── Content.Server.Database
│   │   ├── Content.Tests
│   │   └── Content.YAMLLinter
│   ├── Resources
│   │   ├── Audio
│   │   ├── ...
│   │   ├── Maps
│   │   ├── Prototypes
│   │   ├── ...
│   │   └── Textures
│   └── RobustToolbox
│       ├── bin
│       │   ├── Benchmarks
│       │   ├── Client
│       │   ├── Server
│       │   ├── Shared.Maths
│       │   └── UnitTesting
│       ├── Resources
│       │   ├── EngineCredits
│       │   ├── EnginePrototypes
│       │   ├── ...
│       │   └── Textures
│       └── Schemas
├── Content.Packaging
├── Content.Shared
├── Content.Shared.Database
├── Content.Tools
├── Lidgren.Network
├── NetSerializer
├── OpenToolkit.GraphicsLibraryFramework
├── Pow3r
├── Robust.Analyzers
├── Robust.Client.Injectors
├── Robust.Client.NameGenerator
├── Robust.LoaderApi
├── Robust.Packaging
├── Robust.Physics
├── Robust.Shared
├── Robust.Shared.Maths
├── XamlX
├── XamlX.IL.Cecil
└── XamlX.Runtime
```

Example Content `Directory.Build.props`, which must be written by the developer who wants to do an out-of-source-tree build:
```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="12.0">
	<ImportGroup Label="PropertySheets"/>
	<PropertyGroup>
		<BaseIntermediateOutputPath>\tmp\cache\MSBuild\obj\$(MSBuildProjectName)</BaseIntermediateOutputPath>
		<BuildPrefix>\tmp\SS14</BuildPrefix>
		<BaseOutputPath>$(BuildPrefix)\$(MSBuildProjectName)\bin</BaseOutputPath>
		<AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
		<UseSymbolicLinksIfPossible>false</UseSymbolicLinksIfPossible>
	</PropertyGroup>
</Project>
```
Content changes:
* Here are the changes for the SS14 repository: space-wizards/space-station-14#13355
* Here are the changes for OpenDream: OpenDreamProject/OpenDream#1019

Shortcomings:
* With respect to `Resources/` msbuild doesn't seem to support symbolically linking directories and must instead recursively link (or copy if no symlink permissions) each file. MSBuild limitation, not worth improvement labour. The example property `$(UseSymbolicLinksIfPossible)` in the example `Directory.Build.props` allows for customization, if false resources will be copied not links.

Rationale:
* `Robust.Common.props` should be included in each csproj with curated/actual build output. E.g. Content.Client not Content.Tools or other top level outputs in example output tree above. I believe CMake generator would make this unnecessary; can't think of simpler alternative.
* `Robust.OutOfSourceTree.targets` is only included once and makes an assumption of content `Resources/`, maybe a different user of RobustToolbox would want a different set of content folders in the actual/curated build output than just that. Could rename this targets file or maybe put contents in different file... maybe that would be better organization.

Other changes:
* ~~Included lines for `$(TargetFramework)` bug workaround- when TargetFramework was included from an .props file included into the csproj. Comments on that bug are in other csproj files.~~ Decided to drop this as out of scope of this PR.

**Submodule change:** [cefglue#4](https://github.com/space-wizards/cefglue/pull/4). ~~Will attempt to do upstream cefglue patch, then get that pulled into space-wizards/cefglue, that seems like proper approach. I submitted a patch to their support forum.~~ Seems like no use to wait on upstream, better to put into space-wizards fork.

rene-descartes2021:
Well looks like this feature will be in .NET 8 via an 'artifacts path' setting, written about in a link from that $(RootOutputPath) link in last comment (.net design # 281). Is currently in .NET 8 Preview 3 and someone wrote a survey was sent soliciting design feedback. Hooray.

w.r.t this proposal: given relatively low user and maintainer interest in addition to csproj maintainance overhead (scrutiny of new csproj for compliance), I think I should close this and just use privately for now, until .NET 8 used.

A user in the linked OpenDream PR expressed interest in this proposal. I suppose I could squash these PRs for convenience for users to be able to use in a private workflow.

PR #3414, opened 2022-10-30
Broadphase update deferral

ElectroJr:
WIP

Tries to defer broadphase updates while applying server states. Should help significantly with stutter induced by entities coming into PVS range by avoid wasted broadphase insertions/removals. Currently the combination of moving, physics body type, physics collision, and container insertion can mean that an entity might be unnecessarily added & removed from broadphases 3+ times.


ElectroJr:
Mostly redundant now

PR #3421, opened 2022-10-31
Add lookup reparenting tests

metalgearsloth:
The 2nd one was failing on the old build as the 0-fixture entity was being added.


Robust.UnitTesting/Shared/Physics/Broadphase_Test.cs
```diff
@@ -13,6 +15,80 @@ namespace Robust.UnitTesting.Shared.Physics;
...
+        var mapId = mapManager.CreateMap();
+        var mapEnt = mapManager.GetMapEntityId(mapId);
+        var grid = mapManager.CreateGrid(mapId);
+
+        grid.SetTile(Vector2i.Zero, new Tile(1));
+        Assert.That(entManager.HasComponent<BroadphaseComponent>(grid.GridEntityId));
+        var broadphase = entManager.GetComponent<BroadphaseComponent>(grid.GridEntityId);
+
+        var ent = entManager.SpawnEntity(null, new EntityCoordinates(grid.GridEntityId, new Vector2(0.5f, 0.5f)));
+        Assert.That(broadphase.SundriesTree, Does.Contain(ent));
```
ElectroJr: Now that broadphases are cached, each of these should also check that the xform's cached broadphase is correct.

PR #3422, opened 2022-10-31
removes beforeserialization hook

PaulRitter:
requires https://github.com/space-wizards/space-station-14/pull/12319

### DIFF
--- a/Robust.Shared/Physics/FixturesComponent.cs
+++ b/Robust.Shared/Physics/FixturesComponent.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Robust.Shared.GameObjects;
 using Robust.Shared.GameStates;
+using Robust.Shared.IoC;
 using Robust.Shared.Physics.Dynamics;
 using Robust.Shared.Physics.Systems;
 using Robust.Shared.Serialization.Manager.Attributes;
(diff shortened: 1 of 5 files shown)


ElectroJr (approved):

Robust.Shared/Physics/FixturesComponent.cs
```diff
@@ -38,21 +34,12 @@ public sealed class FixturesComponent : Component, ISerializationHooks
...
-
-        void ISerializationHooks.BeforeSerialization()
+        internal List<Fixture> SerializedFixtures
         {
-            // Can't assert the count is 0 because it's possible it gets re-serialized before init!
-            if (SerializedFixtures.Count > 0) return;
-
-            // At some stage we'll serialize non-default grids (gridfixturesystem does support it)
-            // but for now just for smaller map file + reading speed we'll just globally cull them.
-            if (IoCManager.Resolve<IEntityManager>().HasComponent<MapGridComponent>(Owner)) return;
```
ElectroJr, line 50: Actually this bit was needed to avoid writing the grid's fixtures to a map file.

ElectroJr (changes requested):

Robust.Shared/Physics/FixturesComponent.cs
PaulRitter: should be fixed now

PR #3433, opened 2022-11-02
Aurora Borealis

metalgearsloth:
Makes ambientlightcolor per-map rather than on the LightManager. This means we can do separate planetmap lighting in the future and also means we can easily use stuff like RGBLightController.

Gonna move this to its own component so if content is using RGBLightController it doesn't reset the lighting whenever a map state comes in.


PJB3005:
Planet lighting is still gonna need somebody to figure out an acceptable solution for HDR.

PJB3005:
Uhh this looks PR looks kinda broken right now.

metalgearsloth:
I think it got merged in on a future pr so uhhh.

PR #3437, opened 2022-11-05
Infer default command names

ElectroJr:
Makes localized commands default to using a command string inferred from the class name . E.g. for something like the `DumpEntitiesCommand` class, you no longer need to specify `public override string Command => "dumpentities";`

This implicitly makes flatcase/mumblecase the standard convention, which it already appears to be (though there are some snake_case commands),


PJB3005:
👎 I like these being explicit and consistent. Automatic name folding like this sucks.

PR #3448, opened 2022-11-07
LoadFromDocument properly detects duplicate entities

### LINKED ISSUE #3079: serv3 doesn't error on duplicate prototype IDs
SS14 has several that have gone unnoticed, the most recent one is BoxSurvival.

Mervill:
I circled back around to this and discovered that the fix is (probably?) easy.

Closes #3079

I tested this locally but I'm pushing the change via the web editor so I don't have to go to the trouble of setting up a Robust fork of my own.

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.cs
@@ -786,7 +786,7 @@ namespace Robust.Shared.Prototypes
                 if (!datanode.TryGet<ValueDataNode>(IdDataFieldAttribute.Name, out var idNode))
                     throw new PrototypeLoadException($"Prototype type {type} is missing an 'id' datafield.");
 
-                if (!overwrite && _prototypes[prototypeType].ContainsKey(idNode.Value))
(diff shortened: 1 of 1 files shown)


Mervill:
I should point out that this only works because `_prototypeResults` is never cleared (i.e. after `ResolveResults()`) so as long as that's intended then this is good. (It _is_ cleared when the whole manager is reset)

Mervill:
The content test is failing because it's detecting the duplicates discussed in the content pr

mb1jxrc806141

PaulRitter:
> I should point out that this only works because `_prototypeResults` is never cleared (i.e. after `ResolveResults()`) so as long as that's intended then this is good. (It _is_ cleared when the whole manager is reset)

yep, its intended

PaulRitter (approved):
good catch!

PaulRitter:
to elaborate, because im not at my phone anymore: yep, prototyperesults dont get cleared because we are pushing composition (inheritance) using the datanodes. Because of hotreloading, we need to keep the mappings around so we can properly push composition again when a prototype gets new mappings. now that i think of it, we could probably make it clear them in release after ResolveResults runs to free up alot of memory. Another quick and easy pr i suppose :D

PR #3467, opened 2022-11-12
Warn about invalid placement type

Mervill:
Can't remember if this got logged anywhere but I did see this on discord. Currently if you enter an invalid placement type (i.e. you got the capitalization of `SnapgridCenter` wrong) you wont get any warning about it and it's frustrating to debug.

Eventually™️ this should have a proper linter-level validator, but until such time as the real work is done, we can just yell at the developer so they at least know what's wrong.


Robust.Client/Placement/PlacementManager.cs
```diff
@@ -470,14 +471,14 @@ public void BeginHijackedPlacing(PlacementInformation info, PlacementHijack? hij
...
+            if (!_modeDictionary.TryFirstOrNull(pair => pair.Key.Equals(CurrentPermission.PlacementOption), out KeyValuePair<string, Type>? placeMode))
             {
+                Logger.LogS(LogLevel.Warning, nameof(PlacementManager), $"Invalid placement mode `{CurrentPermission.PlacementOption}`");
                 Clear();
                 return;
             }
 
-            var modeType = _modeDictionary.First(pair => pair.Key.Equals(CurrentPermission.PlacementOption)).Value;
-            CurrentMode = (PlacementMode) Activator.CreateInstance(modeType, this)!;
+            CurrentMode = (PlacementMode) Activator.CreateInstance(placeMode.Value.Value, this)!;
```
Mervill, line 481: I thought nullables got unwrapped transparently, is that only for content?

PJB3005: I'm not sure what you mean with that.

Mervill: `placeMode.value` here is a nullable so I have to use `value.value` to get the actual value I want out of the nullable. Am I crazy or does content auto-magic this step away so you don't have to work with nullables like this?

Mervill: Anyway it doesn't have anything to do with this pr really. If it looks good then please merge when appropriate.

moonheart08: > I thought nullables got unwrapped transparently, is that only for content?

Only applies to classes (aka ref types that can be transparently null)
Structs implement nullability through a specific type (`Nullable<T>`) and cannot be transparently converted.

PR #3471, opened 2022-11-13
Audio and other misc replay related changes

ElectroJr:
This is part of getting the server to record replays. One of the main difficulties is in recording entity events, particularly audio events. This is because most audio events are sent via filters instead of globally or to a single session. So its often not clear whether any given event should be "recorded" so that it can be included in a replay. I.e., is the filter the result of a `Filter.Pvs()` or some custom filtering logic? Additionally, some system manually iterate over filter recipients when raising events, so simply blindly recording all events leads to lots of duplicates.

So this PR does the following:
- Audio system methods that take in a `Filter` now have to specify whether or not the sound should be "recorded"
- There are several new audio system overrides to play sounds, including variants that take in an `ICommonSession` or `EntityUid` recipient. These avoids the unnecessary `IoCManager.Resolve` calls  hidden in `Filter.Entities()` & automatically disables replay recording.
- Adds a new `RaiseNetworkEvent(args, ICommonSession)` entity system proxy method (which does not record).
-  The ` EntitySystem .RaiseNetworkEvent(args, filter)` proxy method now has a new recording optional argument that defaults to true.  Almost all uses of this function were either Filter.Pvs() or a single-entity filter (which now uses the new proxy method).
- There is a new `IReplayRecordingManager` which is currently just a dummy class for the sake of keeping the PRs separate without having to go back and re-update all the audio methods.

Requires space-wizards/space-station-14/pull/12578


Robust.Shared/Replays/IReplayRecordingManager.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Robust.Shared.Replays;
+
+public interface IReplayRecordingManager
+{
+    /// <summary>
+    ///     Queues some net-serializable data to be saved for replaying
+    /// </summary>
+    /// <remarks>
+    ///     The queued object is typically a something like an <see cref="EntityEventArgs"/>, so that replays can
```
metalgearsloth:
```suggestion
    ///     The queued object is typically something like an <see cref="EntityEventArgs"/>, so that replays can
```

metalgearsloth (approved):

PR #3472, opened 2022-11-13
Misc lerping changes

ElectroJr:
Renames some of the lerping related variables, Adds a method the clear all active lerps, removes `TransformStartLerpMessage()`.


Robust.Client/GameObjects/EntitySystems/TransformSystem.cs
```diff
@@ -56,39 +73,41 @@ public override void FrameUpdate(float frameTime)
             for (var i = 0; i < _lerpingTransforms.Count; i++)
             {
                 var transform = _lerpingTransforms[i];
+                if (transform.Deleted)
+                    continue;
```
metalgearsloth, line 77: Should this still clean it up from the list?

metalgearsloth: Also woops fatfingered the merge

PR #3476, opened 2022-11-13
Expose more net serializer stuff to content

ElectroJr:
~~Changes some parts of `IRobustSerializer` and `IRobustMappedStringSerializer` from `internal` to `public` and adds some public methods to retrieve hashes. Or is there some reason they should be internal only? Main rationale is to make it easier for content to use `IRobustSerializer` to deserialize data.~~

~~Maybe `IRobustMappedStringSerializer` isn't meant to be used outside of engine networking, but its sort of required for things like replays that just reuse existing networked entity events.~~

See comment below providing an updated overview of changes.


Robust.Shared/Serialization/IRobustMappedStringSerializer.cs
```diff
@@ -11,7 +11,7 @@
 namespace Robust.Shared.Serialization
 {
     [PublicAPI]
-    internal interface IRobustMappedStringSerializer
+    public interface IRobustMappedStringSerializer
```
PJB3005: If you need something like `MappedStringsHash` you should make those properties specifically and nothing else. A lot of the functions you're making public like `Handshake` shouldn't ever be public either way.

Robust.UnitTesting/IntegrationMappedStringSerializer.cs
```diff
@@ -70,6 +70,11 @@ private void HandleServerHandshake(MsgMapStrServerHandshake message)
             ClientHandshakeComplete?.Invoke();
         }
 
+        public (byte[] mapHash, byte[] package) GeneratePackage()
+        {
+            throw new NotImplementedException();
```
PJB3005: Should this be `NotSupportedException()` instead?

PJB3005 (changes requested):

Robust.UnitTesting/IntegrationMappedStringSerializer.cs
ElectroJr: yes

ElectroJr:
> If you need something like MappedStringsHash you should make those properties specifically and nothing else. A lot of the functions you're making public like Handshake shouldn't ever be public either way.

I've updated the PR so now:
- `IRobustMappedStringSerializer` and `MappedStringDict` stay internal
- `IRobustMappedStringSerializer` now has two internal functions to set and get the string mapping data.
- The public `IRobustSerializer` now has two public methods for getting hashes & string mapping data.
- Changes `RobustSerializer` from public to internal & abstract
- Added an internal `ClientRobustSerializer` and public `IClientRobustSerializer` which provides a public function that enables string-mapping data to be directly set in single player games.
- Also adds an empty internal `ServerRobustSerializer`

PR #3477, opened 2022-11-14
Client-, shared-, and server-side ConfigurationManager 

ElectroJr:
This splits ConfigurationManager into client and sever classes/interfaces. This is partly to reduce the number of _netManager.IsClient checks, and partly so that the client can check if `IBaseClient.RunLevel == ClientRunLevel.SinglePlayerGame` when trying to set "only server settable" cvars like tickrate.

After #3466 there is now also a weird interface combinatorics issue (`client`/`server` x `net` x `internal`)


Robust.Client/Configuration/ClientNetConfigurationManager.cs
```diff
@@ -0,0 +1,118 @@
...
+
+    protected override void ApplyNetVarChange(
+        INetChannel msgChannel,
+        List<(string name, object value)> networkedVars,
+        GameTick tick)
+    {
+        ApplyClientNetVarChange(networkedVars, tick);
+    }
+
+    public void ApplyClientNetVarChange(List<(string name, object value)> networkedVars, GameTick tick)
```
PJB3005: This shouldn't be public.

PJB3005 (changes requested):
Seems good, though `GetClientCVar` seems like a potential loss on the client: it could be useful for writing shared code and is sane to implement on both sides.

Robust.Client/Configuration/ClientNetConfigurationManager.cs
ElectroJr: The reason I initially made this public instead of private was for replays, so that `ReceivedInitialNwVars` gets invoked without having to receive a cvar net-message, as I though it might be required to properly set up the ss14 client. Though looking at it further, I guess it doesn't actually need it?

Currently `ReceivedInitialNwVars` is only used by `BaseClient` after having connected to a server (which I guess makes sense). So as long as it never gets used by any other kind of set up that might be required for replays to work properly, I guess it can stay private?

ElectroJr:
This now also adds an optional argument to ignore normal cvar flags when setting. (again, useful when forcing server authoritative cvars in replays)

PR #3483, opened 2022-11-14
Parallel physics broadphase

metalgearsloth:
It's ~twice as fast on the tumbler in my benchmarks with dottrace (~30% of physics to 15%).

The valuelist is a bit sussy but can deal with that later.


metalgearsloth:
Okay valuelist was too sussy so I just pooled it again.

PR #3484, opened 2022-11-14
Fix bad reference

amylizzle:
Deletes references to `GUITextEventArgs` which was causing a compile fail for OpenDream due to a missing reference.

This is probably not the right way to fix this, so feel free to fix it properly instead.


PJB3005:
I ought to set up CI for WebView...

PR #3485, opened 2022-11-14
WIP FText implementation

Ygg01:
Essentially will add new interface for sharing FText localization (name pending).

FText is just localization id + arguments needed. Rn I got the commands to serialize.

https://user-images.githubusercontent.com/1146204/201720286-bb1d5c69-aeb1-4f89-b2eb-828dceb28ecb.mp4

@PJB3005 you mentioned it needs some sort of recursion but didn't quite specify what for or how?

This is WIP. Expect SharedLocalizationManager and LocalizationManager to merge. Too much duplication.


PJB3005:
> you mentioned it needs some sort of recursion but didn't quite specify what for or how?

We need to be able to specify FTexts as parameters to other FTexts.

Furthermore, I really don't like that `FText` is a complex nested datastructure like this. It feels like the kind of thing that should really be stored as a single object in-memory to avoid GC churn on object count. This would entail the in-memory representation being some sort of serialized string.

Ygg01:
Closing until I have a better idea of how to do it.

PR #3492, opened 2022-11-16
Adds basic delta state support

ElectroJr:
also makes the get-state event always include the player session data. Null now implies no client (i.e., replays).


Robust.Client/GameStates/GameStateProcessor.cs
```diff
@@ -177,11 +177,29 @@ public void UpdateFullRep(GameState state, IEntityManager entMan)
...
+                    var compState = change.State;
+
+                    if (compState is IComponentDeltaState delta && !delta.FullState)
+                    {
+                        var ev = new ComponentMergeDeltaState(compData[change.NetID], compState);
+                        entMan.EventBus.RaiseComponentEvent(entMan.GetComponent(entityState.Uid, change.NetID), ref ev);
+                        DebugTools.AssertNotNull(ev.NewState);
+                        DebugTools.Assert(!ReferenceEquals(compData[change.NetID], ev.NewState));
+                        compState = ev.NewState!;
+                    }
```
PJB3005: This is not the place to do this. `GameStateProcessor` is very much a self-contained system solely concerned with shuffling component states around, it's not designed to call into the actual game state like this.

Doing this on actual entities is a no-go. It may be acceptable to figure out some way to wire up handling logic on systems, but it should be very much purely on the state, no looking at entity UIDs.

PJB3005 (changes requested):

Robust.Client/GameStates/GameStateProcessor.cs
ElectroJr: I've now changed this around so that the `IComponentDeltaState` interface provides an apply-state function, rather than raising this event. So the logic should be entirely confined to the actual states themselves, and not any system or entity.

PaulRitter (approved):

PR #3495, opened 2022-11-16
Switches order of post-insertion debug assertions and event raising in base container insertion logic.

TemporalOroboros:
This should prevent the debugger from crashing if a signal handler is created which relocates an entity in response to that entity being inserted into a container.
 
## TODO:
- [ ] Testing. Am currently phonecoding so it's needed more than usual.

### DIFF
--- a/Robust.Shared/Containers/BaseContainer.cs
+++ b/Robust.Shared/Containers/BaseContainer.cs
@@ -141,16 +141,16 @@ namespace Robust.Shared.Containers
             // Implementation specific insert logic
             InternalInsert(toinsert, entMan);
 
-            // Raise container events (after re-parenting and internal remove).
-            entMan.EventBus.RaiseLocalEvent(Owner, new EntInsertedIntoContainerMessage(toinsert, oldParent, this), true);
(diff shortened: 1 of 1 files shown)


ElectroJr:
You shouldn't be removing an entity from a container while handling the insertion events. If the entity is not meant to be in the container you should either block it from ever getting inserted, or just not use containers in the first place.

TemporalOroboros:
I'm attempting to figure out 'the thing I just locker smited instantly ate the locker because it is a black hole'. Using the insert attempt events requires canceling the event in the event that the container is consumed. This implies to the code attempting the insertion that the insertion has failed when it in fact did not, the black hold just ate its way back out.

ElectroJr:
It sounds like the problem there is that nothing is preventing the singularity from being put into a locker in the first place.  You should just stop that from happening in the first place rather than this.

EmoGarbage404:
> It sounds like the problem there is that nothing is preventing the singularity from being put into a locker in the first place. You should just stop that from happening in the first place rather than this.

just as a note, there is an event for attempted insertion inside entity storage. Just use that.

PR #3501, opened 2022-11-19
Change ComponentGetState args

ElectroJr:
Changes the session argument of the ComponentGetState event, so that the session is always included when sending the state to normal players.

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -1153,7 +1153,7 @@ internal sealed partial class PVSSystem : EntitySystem
             if (component.SessionSpecific && !EntityManager.CanGetComponentState(bus, component, player))
                 continue;
 
-            var state = EntityManager.GetComponentState(bus, component, component.SessionSpecific ? player : null);
+            var state = EntityManager.GetComponentState(bus, component, player);
             changed.Add(new ComponentChange(netId, state, component.LastModifiedTick));
 
             if (sendCompList)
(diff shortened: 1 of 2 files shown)


Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -1153,7 +1153,7 @@ private EntityState GetEntityState(ICommonSession player, EntityUid entityUid, G
             if (component.SessionSpecific && !EntityManager.CanGetComponentState(bus, component, player))
                 continue;
 
-            var state = EntityManager.GetComponentState(bus, component, component.SessionSpecific ? player : null);
```
metalgearsloth, line 1156: Does this get the state twice now?

Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -1187,7 +1187,7 @@ private EntityState GetFullEntityState(ICommonSession player, EntityUid entityUi
             if (component.SessionSpecific && !EntityManager.CanGetComponentState(bus, component, player))
                 continue;
 
-            changed.Add(new ComponentChange(netId, EntityManager.GetComponentState(bus, component, component.SessionSpecific ? player : null), component.LastModifiedTick));
```
metalgearsloth, line 1190: Same here

Robust.Server/GameStates/PVSSystem.cs
ElectroJr: I'm not sure what you mean, it should still only be fetching it twice? Or do you mean that can-get-state and get-state should just be merged? Because looking at this, they probably could just be merged at this point?

ElectroJr:
Closing this in favour of the delta state PR

PR #3502, opened 2022-11-19
Prototype load parallelization

PJB3005:
SPEED


Robust.Shared/Prototypes/IPrototypeManager.cs
```diff
@@ -0,0 +1,265 @@
...
+internal interface IPrototypeManagerInternal : IPrototypeManager
+{
+    event Action<DataNodeDocument>? LoadedData;
+}
+
+public sealed record PrototypesReloadedEventArgs(
+    IReadOnlyDictionary<Type, PrototypesReloadedEventArgs.PrototypeChangeSet> ByType)
+{
+    public sealed record PrototypeChangeSet(IReadOnlyDictionary<string, IPrototype> Modified);
+}
```
moonheart08, line 265: praise be it has a change set now.

PJB3005: That's been there for a while, I just moved it.

PR #3509, opened 2022-11-20
Add basic gamestate / replay recording.

ElectroJr:
This adds an actual `IReplayRecordingManager` implementation for recording server game states for replays. It doesn't have any complex saving or replay-distribution logic, nor does it automatically start recording replays during rounds. It requires a host to run a command and retrieve the recording from the server's user data directory.

## How to use
You can use the `/replaystart` command start a recording. `/replaystats` will print basic info about the current recording (duration, size). Recordings can be stopped via `/replaystop`, or automatically when they get too big (`CVars.ReplayMaxCompressedSize`).

## How it works
While a replay is being recorded the recording manager will basically create a game state that is effectively equivalent to what would get sent to clients if PVS was disabled. It then serializes & compresses that before eventually saving it to a file. The game state recording is done in parallel with all the other PVS/gamestate updating.

Networked events/messages get saved to replays via the `IReplayRecordingManager.QueueReplayMessage()` function, which can take in any `[NetSerializable]` object and will save them each tick alongside the gamestate. The replay client can then deserialize them and do whatever it wants with them. E.g., when deserializing an `EntityEventArgs` the client currently just raises that event locally as if it had been received via the network, triggering any networked event subscriptions.

The combination of just replaying game-states & broadcasted `EntityEventArgs` is just about all you need for basic SS14 replays.

## I need a database magician

The way the data is currently saved is pretty simple, it just throws everything into a subfolder (`CVars.ReplayDirectory`) in the server's UserData directory and overwrites the files every time a new recording starts. Eventually that should probably somehow be saved to some database that periodically culls old replays and is somehow accessible to admins or the general public. But IMO its good enough in its current form to get some recordings to test & work on a better replay client.

## Replay files / contents
Actual replays consist of several files:
- `replay.yml`, which has misc data bout the server/versions/hashes, as well as whatever data content wants to add to the file.
- `cvars.toml` has information about the initial replicated cvars that get sent to clients
- `strings.dat` which is a binary file with data required for `IRobustMappedStringSerializer` to read "networked" messages
- `init.dat` is a binary file with misc `[NetSerialized]` objects that get sent to clients when they first connect (e.g., admeme yaml prototype & resource uploads).
- Some arbitrary number of binary files containing tick data (`0.dat. 1.dat`, etc).
  - These contain the actual gamestate & networked messages required for replays
  -  Each file generally contains multiple ticks of data. The sever collects data until it is larger than 1 mb (`Cvars.ReplayTickBatchSize`) and then compresses & writes the file to the data directory

## Issues

I'm using `IRobustSerializer` as a crutch to serialize `NetMessage`s by just wrapping them in some other `[NetSerializable]` object. E.g., the cvar-update messages (`MsgConVars`) are just getting wrapped in a `CvarChangeMsg` object that then gets saved via `QueueReplayMessage()`. The replay client then deserializes that & converts it back to a `MsgConVars`. This is pretty jank, especially considering that all the `NetMessage` inheritors are already inherently serializable via their `WriteToBuffer` function. But I can't seem to figure out any easy way of actually using that to save the `NetMessage` data to a file, because that involves some internal/private lidgren stuff. I'm open to alternatives if anyone knows a better way of doing this.


Robust.Server/Replays/ReplayRecordingManager.cs
```diff
@@ -0,0 +1,306 @@
...
+        // Saving YAML data
+        {
+            _yamlMetadata = new MappingDataNode();
+
+            // TODO REPLAYS are these the right properties needed to uniquely identify content+engine builds?
+            // could this just point to a github hash?
+            _yamlMetadata["engineVersion"] = new ValueDataNode(_netConf.GetCVar(CVars.BuildEngineVersion));
+            _yamlMetadata["buildForkId"] = new ValueDataNode(_netConf.GetCVar(CVars.BuildForkId));
+            _yamlMetadata["buildVersion"] = new ValueDataNode(_netConf.GetCVar(CVars.BuildVersion));
+            _yamlMetadata["buildHash"] = new ValueDataNode(_netConf.GetCVar(CVars.BuildHash));
```
ElectroJr: The latter three fields are just empty when testing locally, but I assume they get populated in full release servers?
If not, maybe this requires some other version identifiers.

At the end of the day, it'd be nice to be able to just point to a git commit so if someone has a recording but no matching client they could in principle just clone & build it themselves.

ElectroJr: Actually apparently the build hash is just for the content pack, not some github thing.
Either way, the gist is that content can just add whatever yaml data it wants to uniquely identify fork via `StartingRecording?.Invoke((_yamlMetadata, extraData))`, so that in principle the data can be used to direct people to an appropriate place to get the right client version

PR #3518, opened 2022-11-22
Disable discord rich presence locally

metalgearsloth:


PJB3005 (changes requested):
Change the Cvar default or override it or something instead. There's no need to swap out the actual code with annoyi g `#if`s

Issue #3524, opened 2022-11-24
Contact linkedlist exceptions


metalgearsloth:
Noticed on FTL / cancollide stuff.

![image](https://user-images.githubusercontent.com/31366439/203668839-ecfa804f-33e9-4a91-977f-2a92f7c6ed74.png)

ElectroJr:
A similar error seems to have happened again recently:
```
Caught exception in entsys
System.InvalidOperationException: The LinkedList node does not belong to current LinkedList.
   at Robust.Shared.Physics.Systems.SharedPhysicsSystem.DestroyContact(Contact contact) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs:line 344
   at Robust.Shared.Physics.Systems.SharedBroadphaseSystem.RegenerateContacts(EntityUid uid, PhysicsComponent body, FixturesComponent fixtures, TransformComponent xform) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs:line 430
   at Robust.Shared.Physics.Systems.SharedPhysicsSystem.TrySetBodyType(EntityUid uid, BodyType value, FixturesComponent manager, PhysicsComponent body, TransformComponent xform) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs:line 459
   at Content.Shared.Construction.EntitySystems.AnchorableSystem.OnUnanchorComplete(EntityUid uid, AnchorableComponent component, TryUnanchorCompletedEvent args) in /home/runner/work/space-station-14/space-station-14/Content.Shared/Construction/EntitySystems/AnchorableSystem.cs:line 105
   at Robust.Shared.GameObjects.EntityEventBus.EntDispatch(EntityUid euid, Type eventType, Unit& args) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityEventBus.Directed.cs:line 584
   at Robust.Shared.GameObjects.EntityEventBus.RaiseLocalEvent(EntityUid uid, Object args, Boolean broadcast) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityEventBus.Directed.cs:line 186
[...]

PR #3525, opened 2022-11-24
Fix grid contact ordering

metalgearsloth:
The contact was destroyed before grabbing the next node.


metalgearsloth:
May address https://github.com/space-wizards/RobustToolbox/issues/3524 will need to see because it's very sporadic.

PR #3529, opened 2022-11-24
Clear out fixture proxies on broadphase deletion

### LINKED ISSUE #3528: Broadphase exception
AFAICT this only ever happens with stopfollowing:

image

metalgearsloth:
This can happen in some re-parent situations so need to avoid trying to move incorrect proxies.

Fixes https://github.com/space-wizards/RobustToolbox/issues/3528

Will write a test soon.


Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -72,9 +72,12 @@ public sealed partial class EntityLookupSystem : EntitySystem
         /// </summary>
         public const LookupFlags DefaultFlags = LookupFlags.Contained | LookupFlags.Dynamic | LookupFlags.Static | LookupFlags.Sundries;
 
+        private ISawmill _sawmill = default!;
+
         public override void Initialize()
         {
             base.Initialize();
+            _sawmill = Logger.GetSawmill("lookup");
```
metalgearsloth:
```suggestion
        public override void Initialize()
        {
            base.Initialize();
```

Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -445,6 +445,16 @@ private void UpdateParent(EntityUid uid, TransformComponent xform, EntityUid old
...
+            // Old broadphase was deleted so make sure fixture proxies are cleared
+            if (oldBroadphase == null && fixturesQuery.TryGetComponent(uid, out var fixtures))
+            {
+                foreach (var fixture in fixtures.Fixtures.Values)
+                {
+                    fixture.ProxyCount = 0;
+                    fixture.Proxies = Array.Empty<FixtureProxy>();
+                }
+            }
+
```
ElectroJr: I think it is safe if this code is only run a bit further up where there is the `// broadphase was probably deleted` comment (avoids extra fixture queries for client PVS pop-in).

ElectroJr: It should also be chucked into `TryGetCurrentBroadphase()`, then I think that covers any instance where an entity specified via some `xform.Broadphase` doesn't have a broadphase comp?

ElectroJr:
I'm gonna push some changes and then merge this

ElectroJr:
Actually I'll just combine it with #3538 for easier testing

PR #3531, opened 2022-11-24
Correct removal of PVS state for players without inGame status

Morb0:
The state for PVS is created only if the player's status went to `InGame`, but is deleted without checking whether the player was really in `InGame` state.
This fixes the problem when the user disconnect while in `Connected` state.
Checked in game and error from 412 line gone.

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -403,7 +403,7 @@ internal sealed partial class PVSSystem : EntitySystem
             return;
         }
 
-        if (e.NewStatus != SessionStatus.Disconnected)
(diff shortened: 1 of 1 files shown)


PaulRitter (changes requested):
correct me if im wrong, but i think a player can move back to the lobby after a round, making their playerstate inlobby, which would cause us to not properly clean up if they left at that point.

imo you should move the playervisiblesets tryremove in line 415 infront of the loop and not check the oldstatus like you are doing

PJB3005:
> correct me if im wrong, but i think a player can move back to the lobby after a round, making their playerstate inlobby

Lobby isn't a concept in-engine, they're still `InGame`.

PaulRitter (approved):

PR #3535, opened 2022-11-26
Use correct native GLFW library name on FreeBSD

Partmedia:
This change is actually already upstream, thanks to @PJB3005 :
https://github.com/opentk/opentk/blob/4.0.0-pre9.1/src/OpenToolkit.Windowing.GraphicsLibraryFramework/OpenToolkit.Windowing.GLFW.dll.config

However, because RobustToolbox carries around its own fork of GraphicsLibraryFramework, we need to make this change locally.

By the way, that does in fact work on FreeBSD.


PJB3005:
Just an FYI, you're pointing at a `.dll.config` file which were used by Mono for DLL mapping. CoreCLR doesn't use them.

PR #3540, opened 2022-11-28
Fix rootUid returning all loaded entities

ElectroJr:


Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
```diff
@@ -128,12 +128,12 @@ public void Load(MapId mapId, string path, MapLoadOptions? options = null)
         MapLoadOptions? options = null)
     {
         options ??= DefaultLoadOptions;
-        rootUids = new List<EntityUid>();
 
         var resPath = new ResourcePath(path).ToRootedPath();
 
         if (!TryGetReader(resPath, out var reader))
         {
+            rootUids = new List<EntityUid>(0);
```
metalgearsloth:
```suggestion
            rootUids = new List<EntityUid>();
```

If capacity is passed in then it checks if capacity == 0 then does the same code path as if no capacity was passed in, so it's actually slightly slower and doesn't save memory.

metalgearsloth (approved):

PR #3545, opened 2022-11-29
RSI meta atlas

ElectroJr:
This PR combines individual RSI atlases into one or more large combined atlases. It does this by just writing each RSI to some larger atlas left to right, top to bottom, with each row just being the height of the largest RSI atlas in that row. This can lead to gaps / bad packing, but seems to work well enough. Might hurt start-up times, but improves with rendering.

I don't know what typical max texture sizes are, but 8kx8k currently fits all of content. If it turns out that most users can only support 2k*2k (requires ~ 14 atlases), then maybe the way rsis are grouped into atlases needs to be more intelligent, to ensure that entities that are very common and have the same draw-depth are in the same atlas.

Needs testing on potato computers before merging. I did test it on my laptop, which has no dedicated GPU and it still got a significant improvement, but even that apparently still supports 8k textures.

### DIFF
--- a/Robust.Client/ResourceManagement/ResourceCache.Preload.cs
+++ b/Robust.Client/ResourceManagement/ResourceCache.Preload.cs
@@ -1,12 +1,18 @@
-﻿using System;
+using System;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
+using OpenToolkit.Graphics.OpenGL4;
 using Robust.Client.Graphics;
+using Robust.Client.Utility;
 using Robust.Shared;
 using Robust.Shared.Configuration;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
+using Robust.Shared.Maths;
+using Robust.Shared.Utility;
+using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.PixelFormats;
 
 namespace Robust.Client.ResourceManagement
 {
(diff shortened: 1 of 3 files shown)


PJB3005:
Reducing the max size to something like 2k would reduce the amount of wasted VRAM somewhat but yeah it'd reduce the effectiveness of the atlas drastically. Dunno what the best solution is.

ElectroJr:
As far as video memory goes, is a 2k x 8k atlas functionally equivalent to a 4x4k one, or does it just internally use a 8x8 texture anyways? Because if its the latter, the current approach wastes a bunch of video memory.

moonheart08:
> As far as video memory goes, is a 2k x 8k atlas functionally equivalent to a 4x4k one, or does it just internally use a 8x8 texture anyways? Because if its the latter, the current approach wastes a bunch of video memory.

this is a "what GPU do you have" question, but the answer should almost always be the former unless you're running ss14 on a weird mobile GPU that simplifies texture indexing logic significantly (doubt such a GPU could even handle the lighting much less the rest of the game). I'd only worry if one of the texture dimensions isn't a power of two.
However, it's likely better to pad to be an NxN texture either way, though you may prefer to avoid wasting *significant* texture space in favor of just having two textures. I'd say wasted space in a 4K texture is the least impactful as long as you don't create a dozen of them.

Common knowledge also seems to say that 4K is best, so I'd go with that.

ElectroJr:
> I'd say wasted space in a 4K texture is the least impactful as long as you don't create a dozen of them. ... Common knowledge also seems to say that 4K is best, so I'd go with that.

In the case of ss14, everything fits into one 8x8k atlas, but not a single 4x4 one (even if the singularity & chess boards are removed). So unless the atlas creation is improved to somehow try group common sprites / draw-depths together to avoid swapping textures, I guess I'll just make the atlas size upper limit configurable and set it to 8 for now?

PR #3546, opened 2022-11-30
Queue serializer

ElectroJr:
Adds serializers for queues. Effectively the same as the list/hash serializers.


Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/QueueSerializer.cs
```diff
@@ -0,0 +1,57 @@
...
+using Robust.Shared.Serialization.Markdown;
+using Robust.Shared.Serialization.Markdown.Sequence;
+using Robust.Shared.Serialization.Markdown.Validation;
+using Robust.Shared.Serialization.TypeSerializers.Interfaces;
+using System.Collections.Generic;
+
+namespace Robust.Shared.Serialization.TypeSerializers.Implementations.Generic;
+
+[TypeSerializer]
+public sealed class QueueSerializer<T> : ITypeSerializer<Queue<T>, SequenceDataNode>
```
PaulRitter: could also specify a itypecopier here while you're at it

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/QueueSerializer.cs
```diff
@@ -0,0 +1,57 @@
...
+        return queue;
+    }
+
+    ValidationNode ITypeValidator<Queue<T>, SequenceDataNode>.Validate(ISerializationManager serializationManager,
+        SequenceDataNode node, IDependencyCollection dependencies, ISerializationContext? context)
+    {
+        var list = new List<ValidationNode>();
+        foreach (var elem in node.Sequence)
+        {
+            list.Add(serializationManager.ValidateNode(typeof(T), elem, context));
```
PaulRitter: use the generic signature

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/QueueSerializer.cs
```diff
@@ -0,0 +1,57 @@
...
+    public DataNode Write(ISerializationManager serializationManager, Queue<T> value,
+        IDependencyCollection dependencies,
+        bool alwaysWrite = false,
+        ISerializationContext? context = null)
+    {
+        var sequence = new SequenceDataNode();
+
+        foreach (var elem in value)
+        {
+            sequence.Add(serializationManager.WriteValue(typeof(T), elem, alwaysWrite, context));
```
PaulRitter: use the generic signature

PaulRitter (changes requested):
fyi, if you use the nongeneric signature, the system will be forced to box the values creating more overhead. a small crime, but it can add up.
otherwise lgtm

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/QueueSerializer.cs
ElectroJr: I just pilfered the code from the hashset serializer, so I've also gone and changed that one to generics.

PR #3554, opened 2022-12-04
SIMD physics position integration.

ElectroJr:
Draft because this assumes FMA is supported.

This makes the position integration use SIMD. Performance wise, this should be compared with #3553 (which tries to avoid unnecessary clamping, whereas this will do it for all velocities). 

~~I don't actually know how to best benchmark this, so: Help me @metalgearsloth, you're my only hope.~~
Using 5 tumbler test beds, the overall improvement seems insignificant because most of the time is spent in solving constraints & finding contacts. Probably still worth adding though?

This does implicitly change the velocity limit cvar so that it goes from limiting the magnitude of the velocity, to limiting the magnitude of individual components. 

The same treatment could easily be given to the velocity damping if bodies are sorted by body type.

### DIFF
--- a/Robust.Shared/Physics/Dynamics/PhysicsIsland.cs
+++ b/Robust.Shared/Physics/Dynamics/PhysicsIsland.cs
@@ -490,16 +490,15 @@ stored in a single array since multiple arrays lead to multiple misses.
             {
                 var linearVelocity = velocities[i];
                 var velSqr = linearVelocity.LengthSquared;
+
                 if (velSqr > vLimSqr)
                 {
                     linearVelocity *= vLimit / MathF.Sqrt(velSqr);
+                    // TODO: is this actually required?
                     velocities[i] = linearVelocity;
                 }
 
-                velocities[i] += linearVelocity * frameTime;
-
-                // TODO: is this actually required?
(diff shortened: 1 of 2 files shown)


Robust.Shared/Physics/Dynamics/PhysicsIsland.cs
```diff
@@ -487,6 +457,89 @@ public void Solve(Vector2 gravity, float frameTime, float dtRatio, float invDt,
...
+            {
+                for (; i < count; i += 4)
+                {
+                    var v = Sse.LoadVector128(vPtr + i);
+                    var x = Sse.LoadVector128(xPtr + i);
+                    v = Sse.Max(Sse.Min(v, maxV), minV);
+                    Sse.Store(xPtr + i, Fma.MultiplyAdd(x, time, x));
+
+                    // TODO: is this actually required?
+                    Sse.Store(vPtr + i, v);
```
ElectroJr, line 528: What I mean by this comment is that (AFAIK) there is no actual reason to clamp the velocity, as long as its effects are clamped. In addition to just being faster, this also helps with velocity conservation when crossing grids.

ElectroJr:
Maybe one day

PR #3568, opened 2022-12-10
Fix scroll bars being visible when they shouldn't be.

### LINKED ISSUE #3567: Infinite UI measure loop when opening an OptionButton
## Description

It's an infinite loop of InvalidateMeasure on the Popup ScrollContainer due to the DesiredSize not being correct in Measure.

https://discord.com/channels/310555209753690112/1050983932532891688/1051075953390002266

**Reproduction**

Open a drop down menu

ElectroJr:
Currently scroll bars sometimes appear in scroll containers when they are exactly the desired size of their children due to floating point errors. This should fix that.

This also appears like it might resolve #3567 in the process? Adding logs it looks like the scroll container was fluctuating between showing and not showing the scroll bars? But I'm not 100% sure if this actually fully fixes that, or just stops it from happening for the specific case I was testing (which was showing a scroll bar when it shouldn't have been).

### DIFF
--- a/Robust.Client/UserInterface/Controls/ScrollContainer.cs
+++ b/Robust.Client/UserInterface/Controls/ScrollContainer.cs
@@ -150,7 +150,7 @@ namespace Robust.Client.UserInterface.Controls
                     sWidth -= vBarSize;
                 }
 
-                if (sWidth < cWidth && _hScrollEnabled)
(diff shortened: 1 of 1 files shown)


ShadowCommander:
Just tested it. Fixes #3567.

PR #3585, opened 2022-12-17
Increase granularity of ShouldCollide

TemporalOroboros:
This already runs for every pair of fixtures, now it actually takes which fixtures are colliding into account.


Robust.Shared/Physics/Events/PreventCollideEvent.cs
```diff
@@ -1,18 +1,45 @@
...
+
+    /// <summary>
+    /// The fixture on the first body to prevent the collision of if specified.
+    /// </summary>
+    public Fixture? FixtureA;
+
+    /// <summary>
+    /// The fixture on the other body to prevent the collision of if specified.
+    /// </summary>
+    public Fixture? FixtureB;
```
metalgearsloth: Shouldn't these be non-nullable?

metalgearsloth (changes requested):

Robust.Shared/Physics/Events/PreventCollideEvent.cs
TemporalOroboros: Probably. I made them nullable in case someone wanted to just check between the bodies themselves at some point, but I suppose that if they want to bother raising core physics events elsewhere they can work that out themselves.

PR #3586, opened 2022-12-17
Don't defer dirty matrices

ElectroJr:
I'm not sure if this will have weird side effects of if there was some reason it was like this, but this currently causes issues during rendering where the matrices are incorrect but not marked as dirty. Alternatively, anything that sets DeferUpdates to true needs to make sure that the matrices get properly updated.


metalgearsloth:
I ended up nuking physics use of DeferUpdates in https://github.com/space-wizards/RobustToolbox/pull/3479 so we can probably just remove it entirely soon.

mirrorcult (approved):

Issue #1914, opened 2021-07-30
Helper for batching


PaulRitter:
should do somewhat what #1897 does.

should be provided with:
- minimumThreads
- maximumThreads (optional, will be the processorscount if not provided) (must be < processorcount, duh)
- idealBatchSize

should provide a way to tell:
- should be split in batches
- how many batches á how many datapoints
- maybe even take care of (not) launching the threads by being passed in a delegate

PR #3587, opened 2022-12-17
Somewhat improve sprite processing

ElectroJr:
This PR improves the pre-rendering sprite processing (getting world position, rotation, and bounding boxes). It also moves most of the processing & sorting logic into a separate partial class. AFAICT this is generally not a bottleneck in most situations, but can becomes problematic if there are many sprites on screen (e.g., when zooming out). 

I just tested this using a basic test map with ~15k sprites on screen with random positions. On master release mode I get about 67fps, with a running average of 4.9ms spent inside `ProcessSpriteEntities()`. With these changes, I get ~90fps with 0.67ms spent processing sprites. About half of the speedup just comes from faster transform & bounding box calculations (4.9ms -> 2.5ms), the rest comes from parallelising (2.5ms -> 0.62ms, though it actually increases to ~3.3ms if only using a single thread).

When it comes to SS14 gameplay, when moving around box station (400-500 sprites on screen), my desktop & laptop framerate doesn't actually change all that much. But maybe this will help potato computers?

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.Sprite.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.Sprite.cs
@@ -55,7 +55,7 @@ internal partial class Clyde
         };
 
         // We need to batch the actual tree query, or alternatively we need just get the list of sprites and then
-        // prallelize the rotation & bounding box calculations.
+        // parallelize the rotation & bounding box calculations.
         var index = 0;
         var added = 0;
         var opts = new ParallelOptions { MaxDegreeOfParallelism = _parMan.ParallelProcessCount };
@@ -119,7 +119,7 @@ internal partial class Clyde
             DebugTools.Assert(data.Sprite.Visible);
 
             // To help explain the remainder of this function, it should be functionally equivalent to the following
-            // tree lines of code, but has been expanded & simplified to speed up the calculation:
+            // three lines of code, but has been expanded & simplified to speed up the calculation:
             // 
             // (data.WorldPos, data.WorldRot) = batch.Sys.GetWorldPositionRotation(data.Xform, batch.Query);
             // var spriteWorldBB = data.Sprite.CalculateRotatedBoundingBox(data.WorldPos, data.WorldRot, batch.ViewRotation);
--- a/Robust.Client/Graphics/Clyde/Clyde.Sprite.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.Sprite.cs
@@ -149,7 +149,7 @@ internal partial class Clyde
             pos = batch.ViewRotation.RotateVec(pos - batch.ViewPosition);
 
             // special casing angle = n*pi/2 to avoid box rotation & bounding calculations doesn't seem to give significant speedups.
-            data.SpriteScreenBB = TransformCenteredBoxSse(
+            data.SpriteScreenBB = TransformCenteredBox(
                 data.Sprite.Bounds,
                 finalRotation,
                 pos + batch.PreScaleViewOffset,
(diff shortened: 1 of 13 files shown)


Robust.Client/Graphics/Clyde/Clyde.Sprite.cs
```diff
@@ -0,0 +1,271 @@
...
+            Sys = _entityManager.EntitySysManager.GetEntitySystem<TransformSystem>(),
+            Query = query,
+            ViewRotation = eye.Rotation,
+            ViewScale = viewScale,
+            PreScaleViewOffset = view.Size / 2f / viewScale,
+            ViewPosition = eye.Position.Position + eye.Offset
+        };
+
+        // We need to batch the actual tree query, or alternatively we need just get the list of sprites and then
+        // prallelize the rotation & bounding box calculations.
```
metalgearsloth:
```suggestion
        // parallelize the rotation & bounding box calculations.
```

Robust.Client/Graphics/Clyde/Clyde.Sprite.cs
```diff
@@ -0,0 +1,271 @@
...
+            if (batches > 1)
+                Parallel.For(0, batches, opts, (i) => ProcessSprites(list, index + i * _spriteProcessingBatchSize, _spriteProcessingBatchSize, treeData));
+            else
+                batches = 0;
+
+            var remainder = added - _spriteProcessingBatchSize * batches;
+            if (remainder > 0)
+                ProcessSprites(list, index + batches * _spriteProcessingBatchSize, remainder, treeData);
+
+            index += batches * _spriteProcessingBatchSize + remainder;
```
metalgearsloth, line 101: Should batches be Math.Ceiling?

Robust.Client/Graphics/Clyde/Clyde.Sprite.cs
```diff
@@ -0,0 +1,271 @@
...
+        int count,
+        in BatchData batch)
+    {
+        for (int i = startIndex; i < startIndex + count; i++)
+        {
+            ref var data = ref list[i];
+            DebugTools.Assert(data.Sprite.Visible);
+
+            // To help explain the remainder of this function, it should be functionally equivalent to the following
+            // tree lines of code, but has been expanded & simplified to speed up the calculation:
```
metalgearsloth:
```suggestion
            // three lines of code, but has been expanded & simplified to speed up the calculation:
```

Robust.Client/Graphics/Clyde/Clyde.Sprite.cs
ElectroJr: AFAICT it is correct in its current form (i.e., it should be floor /  integer division). If it were changed so that the partial/remainder is also done in parallel, then yeah it should be ceiling, but that'd require passing in the current batch size so that the last batch could be smaller than all the others.

Alternatively, maybe I'll get around to adding #1914

ElectroJr:
> When it comes to SS14 gameplay, when moving around box station (400-500 sprites on screen), my desktop & laptop framerate doesn't actually change all that much. But maybe this will help potato computers?

image

Well never mind about 15k being an unreasonable test number. If botany is often this active the sprite count might actually approach 10k every now and then, though that's also a performance issue that needs fixing for many other reasons.

ike709:
> When it comes to SS14 gameplay, when moving around box station (400-500 sprites on screen), my desktop & laptop framerate doesn't actually change all that much. But maybe this will help potato computers?

If nothing else this will likely help OD where it's not unusual to have a few thousand sprites on screen at a time.

PJB3005:
> If nothing else this will likely help OD where it's not unusual to have a few thousand sprites on screen at a time.

I don't think any of the code here is hit by OpenDream

ElectroJr:
Currently there is a bug somewhere that results in transform matrices not being properly set during rendering.
#3586 fixed it by just ensuring matrices were always updated, but apparently this broke physics.
So I guess DNM until thats fixed somehow or #3479 just gets rid of the deferral

Issue #3575, opened 2022-12-12
client_config.toml is overwritten with defaults on Apply?


Errant-4:
## Description

Client UI settings are not saved to disk on Apply, and ALL get overwritten with the defaults instead. This seems to impact UI, Audio and Network settings, but not keybinds.

**Reproduction**

1. IMPORTANT! Create a backup of your client_config.toml located under %appdata% or ~/.local  (...)/Space Station 14/data/
2. Connect to any server
3. Change UI scaling, click Apply
4. Close and relaunch the game - all your UI settings are reset to the defaults. (If they did not reset, try again with changing more settings)
5. Restore your backup and never touch the Options again

**Additional context**

Issue was reproduced on both windows and linux, and by multiple people. Probably most people simply do not realise the issue exists because they have no need to touch the Options anymore, but this might annoy a lot of new players who then simply did not report it

Elijahrane:
I can confirm I've experienced this myself and received multiple complaints from it from users of all operating systems

Errant-4:
I have not been able to reliably reproduce how the config file gets corrupted (though it definitely happens at Apply) but as a workaround, any affected player can simply restore the contents of the config file and then things will work and possibly might even be changed ingame and saved without breaking. This for example restores some common settings:

```
[audio]
mastervolume = 1.00

[display]
windowmode = 1
uiScale = 1.25

[interface]
resolutionAutoScaleUpperCutoffY = 770
resolutionAutoScaleLowerCutoffY = 240

[ui]
layout = "Separated"
```

metalgearsloth:
Might be launcher related as this is unreproducible locally but I didn't investigate further.

PJB3005:
The launcher itself shouldn't touch these files, so if that's the case it must be related to FULL_RELEASE or something.

metalgearsloth:
I saw there were some toml changes a few weeks ago but I'm on master and couldn't repro it by messing with my options unless there's some specific setup for it?

Errant-4:
I have spent some literal hours testing this, and I found that
1, if you don't have config files yet (and this is going to be the case for any new players! but you can easily test this by deleting your ~~entire config Space Station 14 dir~~ enough if you delete client_config.toml), the game loads with default config options, and when you try to apply any changes, config.toml for some reason only gets populated with
```
[interface]
resolutionAutoScaleUpperCutoffY = 770
resolutionAutoScaleLowerCutoffY = 240
```
[...]

PR #3590, opened 2022-12-18
Fix bug where config files could be corrupted

eoineoineoin:
Code was previously using a File.OpenWrite, which as documented, in the API, does not truncate the file if it already exists; when a user restored some options to their defaults, this would cause the new config file to be shorter, and, depending on how lucky/unlucky they get, could result in some options being incorrectly set or (more likely) total failure to parse the config file the next time the application is launched (and the bug will repeat.)

### DIFF
--- a/Robust.Shared/Configuration/ConfigurationManager.cs
+++ b/Robust.Shared/Configuration/ConfigurationManager.cs
@@ -260,7 +260,7 @@ namespace Robust.Shared.Configuration
                 var memoryStream = new MemoryStream();
                 SaveToTomlStream(memoryStream, cvars);
                 memoryStream.Position = 0;
-                using var file = File.OpenWrite(_configFile);
+                using var file = File.Open(_configFile, FileMode.Create, FileAccess.Write);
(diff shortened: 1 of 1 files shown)


PJB3005:
Oh, so that's it.

Robust.Shared/Configuration/ConfigurationManager.cs
```diff
@@ -260,7 +260,7 @@ public void SaveToFile()
                 var memoryStream = new MemoryStream();
                 SaveToTomlStream(memoryStream, cvars);
                 memoryStream.Position = 0;
-                using var file = File.OpenWrite(_configFile);
+                using var file = File.Open(_configFile, FileMode.Create, FileAccess.Write);
```
PJB3005:
```suggestion
                using var file = File.Create(_configFile);
```

eoineoineoin:
Sounds like that was a known bug? I did a quick search through the issues, but didn't see anything to link.

Edit: Oh, I see #3575 - I'm just late to the party, sorry!

PR #3598, opened 2022-12-20
Generalize component trees

ElectroJr:
This PR generalizes component trees to avoid code duplication across the occluder, light, and sprite trees, and to make it easier for content to implement its own trees (e.g., for ambient sounds, see space-wizards/space-station-14/pull/13110). This PR should also make the tree updates themselves slightly faster by removing unnecessary coordinate transforms & component queries.

The code that was in the old `RenderTreeSystem` and `OccluderSystem` has effectively all been moved into the a generic abstract 
`ComponentTreeSystem<TTreeComp, TComp>`.  `RenderTreeSystem` is now split into `SpriteTreeSystem` & `LightTreeSystem` which are both very small systems that just inherit from the generic system.

I'm not sure that the current implementation with generics & interfaces is the best approach. The only other alternative I can currently think of that would still avoids code duplication would be to just have a single component tree system & comp that stores a list of trees, and other systems somehow register component that need to be tracked during system init? If that's preferred I can try use something like that instead.  

Note that some trees (light, sprites) recursively update whenever any entity moves, while the occluder tree currently assumes that active occluders are directly parented to a tree. So both of these have different movement handling, which actually matters because only the occluder tree is server-side, which means the server is slightly faster as it doesn't have to do the recursive updates for sprites & lights. Its maybe somewhat janky, but the current implementation should handle this via the `Recursive` option. 

I also haven't properly benchmarked this to see if this negatively effects performance, and there are some ominous comments with warnings about GC pressure and I have NFI how to best benchmark this.

Requires https://github.com/space-wizards/space-station-14/pull/13109

### Other misc changes
In the process of making these changes this also requires partially ECSing the point light and occluder components. This also fixes an occluder bug where the removal of nearby occluders didn't properly update the occluded directions leading to improper occlusion. E.g., see the image below, where the  outlines & other effects were unoccuded

Content Client_CPEQBSQGeT


Robust.Client/GameObjects/EntitySystems/RenderingTreeSystem.cs
```diff
@@ -1,391 +0,0 @@
...
-
-            // This recursive search is needed, as MoveEvent is defined to not care about indirect events like children.
-            // WHATEVER YOU DO, DON'T REPLACE THIS WITH SPAMMING EVENTS UNLESS YOU HAVE A GUARANTEE IT WON'T LAG THE GC.
-            // (Struct-based events ok though)
-            // Ironically this was lagging the GC lolz
-            if (spriteQuery.TryGetComponent(uid, out var sprite))
-                QueueSpriteUpdate(sprite);
-
-            if (pointQuery.TryGetComponent(uid, out var light))
-                QueueLightUpdate(light);
```
ElectroJr, line 148: Welp its a struct event now, but like I said in the PR description, now sure how to benchmark the GC effect,

Robust.Client/GameObjects/ClientComponentFactory.cs
```diff
@@ -22,8 +23,10 @@ public ClientComponentFactory(IDynamicTypeFactoryInternal typeFactory, IReflecti
             RegisterClass<ContainerManagerComponent>();
             RegisterClass<InputComponent>();
             RegisterClass<SpriteComponent>();
-            RegisterClass<ClientOccluderComponent>();
+            RegisterClass<OccluderComponent>();
             RegisterClass<OccluderTreeComponent>();
+            RegisterClass<SpriteTreeComponent>();
+            RegisterClass<LightTreeComponent>();
             RegisterClass<EyeComponent>();
```
metalgearsloth: Should just be using the attribute, then you just register it under robustunittest / robustserversim as needed.

PR #3602, opened 2022-12-21
Hotfix generic visualizers

metalgearsloth:
The issue is that the data is stored with an Enum as a key and not the string so this will always fail. The old behaviour converted it to a string after the fact but the generic in this case checks for the type which will always fail.


ElectroJr:
This'll still break stuff, because the values can be bools or whatever else as well, it just needs to get the generic object out and call ToString() like it did before.

PR #3609, opened 2022-12-24
Don't wrap prototype reload event under full_release

metalgearsloth:
A lot of content assumes this event will actually be issued. It's not ideal but it's better to have prototype reloading in the one spot rather than subbing to both a content event and an engine event. Not sure what a good impl. would look like as there's a lot of different code paths for loading.


ElectroJr:
how does the loadprototype command even work when most of this function is disabled in full release? Is it just not required in the first place?

metalgearsloth:
> how does the loadprototype command even work when most of this function is disabled in full release? Is it just not required in the first place?



> how does the loadprototype command even work when most of this function is disabled in full release? Is it just not required in the first place?

```
var changed = new Dictionary<Type, HashSet<string>>();
_prototypeManager.LoadString(message.PrototypeData, true, changed);
_prototypeManager.ResolveResults();
_localizationManager.ReloadLocalizations();
```

So before my content change this was how admin loading worked; it also had its own gameprototypeloaded C# event but only 1 system subscribed to it.

I'm not sure at what point this stopped working.

PR #3613, opened 2022-12-24
Add tile edge rendering support

metalgearsloth:
Not sure if I should write this into a buffer like chunk rendering does? If so will take me a bit longer to figure it out.

Content PR is https://github.com/space-wizards/space-station-14/pull/13168


moonheart08:
Can't really draw this to a buffer atm due to not having a guarantee that all edge pieces will be in the same texture (which is what allows tile buffering to work)

PR #3618, opened 2022-12-26
Remove client/server split on sharedphysicsmap

metalgearsloth:
Also removed unnecessary client / server comps for physics map.

Content PR is https://github.com/space-wizards/space-station-14/pull/13190


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -471,8 +468,6 @@ public void SetCoordinates(TransformComponent xform, EntityCoordinates value, An
             }
         }
 
-        DebugTools.Assert(!xform.DeferUpdates); // breaks anchoring lookup logic if deferred. If this changes, also need to relocate the `xform.MatricesDirty = true`
-
```
metalgearsloth: Not sure if this would break anything but seems fine ingame? @ElectroJr

PR #3620, opened 2022-12-26
Don't send every grid to clients

metalgearsloth:
Mainly an experiment to see if it works; I'm not sure if we'd have some kind of cvar on grid range or something.

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -446,8 +446,6 @@ internal sealed partial class PVSSystem : EntitySystem
         {
             pvsCollection.AddGrid(gridId);
         }
-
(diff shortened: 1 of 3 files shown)


moonheart08:
This breaks mass scanners.

metalgearsloth:
> This breaks mass scanners.

It sends grids on the same grid as viewers for now so should be fine?

ElectroJr:
I know there is some issue with broadphase/entity lookup if grids can be in nullspace. Theres a comment block around line 550 in EntityLookupSystem. But I think now that the map gets cached in the xform comp, that whole `if` block isn't required anymore and can just be removed?

The if block will currently cause issues if an entity that was previously on a grid that is now out of PVS range re-enters PVS range parented to something else while the grid is still in nullspace.

PR #3621, opened 2022-12-26
Make physics awake bodies a component

metalgearsloth:
Branched from
https://github.com/space-wizards/RobustToolbox/pull/3619
https://github.com/space-wizards/RobustToolbox/pull/3618
https://github.com/space-wizards/RobustToolbox/pull/3617

Awake bodies are no longer stored on physicsmaps and physics is entirely run as one world rather than separate maps. This also gives us the benefit of being able to run much more in parallel.


metalgearsloth:
will do this later

PR #3626, opened 2022-12-27
Remove manual comp registrations from client / server

metalgearsloth:
Still done a little bit on tests but at least we still cut down a lot.


Robust.Server/GameObjects/ServerComponentFactory.cs
```diff
@@ -1,34 +0,0 @@
...
-namespace Robust.Server.GameObjects
-{
-    internal sealed class ServerComponentFactory : ComponentFactory
-    {
-        public ServerComponentFactory(IDynamicTypeFactoryInternal typeFactory, IReflectionManager reflectionManager)
-            : base(typeFactory, reflectionManager)
-        {
-            RegisterIgnore("Input");
-            RegisterIgnore("AnimationPlayer");
-            RegisterIgnore("GenericVisualizer");
```
ElectroJr, line 16: I don't think there is a replacement for this yet? So this is still just required.

PR #3627, opened 2022-12-27
Add grid merging

metalgearsloth:
Just needs tests and maybe a nicer API.


metalgearsloth:
Bringing this bad boy back for xmas.

PR #3629, opened 2022-12-29
Add event dispatching benchmark

gradientvera:
TODO: also benchmark ordered event dispatching, perhaps have three handlers/subscribers per event to test a more "realistic" scenario.


metalgearsloth:
just uhhhhhhhhhh use arch

PR #3631, opened 2022-12-30
Add tickrate to debug monitor

ElectroJr:


moonheart08:
Please don't lock server time to only showing hours; instead let the other time periods show days too.
image
This is a 25hr round but it looks like a 1hr one due to precision.

PR #3645, opened 2023-01-02
physics ECS

metalgearsloth:
Lots of breaking API changes in the interest of cleanup.

- Remove IPhysBody
- Cleanup fixture APIs, access attributes etc
- Cleanup shape APIs, now they should update the relevant systems instead of relying on the caller to handle it
- Implement IEquatable for fixture shapes
- Remove all of the obsolete APIs on physicscomp so now it's actually ECS
- Add a few more resolves that should remove some trygetcomps
- Remove a lot of body.Owner
- Preparations for struct components and struct fixtures / shapes. Shapes can probably be made structs with minimal effort.


metalgearsloth:
Gonna cherry-pick coz I did something that affects the stack solver

PR #3653, opened 2023-01-04
Don't use current broadphase for physics contacts

metalgearsloth:
The 4 dictionary lookups per contact is 1ms of ticktime on the tumbler and I think we can just assume it's griduid / mapuid.


Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs
```diff
@@ -389,8 +389,8 @@ internal void CollideContacts()
 
             var proxyA = fixtureA.Proxies[indexA];
             var proxyB = fixtureB.Proxies[indexB];
-            var broadphaseA = _lookup.GetCurrentBroadphase(xformA);
-            var broadphaseB = _lookup.GetCurrentBroadphase(xformB);
+            var broadphaseA = xformA.GridUid ?? xformA.MapUid;
+            var broadphaseB = xformB.GridUid ?? xformB.MapUid;
```
ElectroJr: If you don't actually need the comp & you could also just do xform.Broadphase?.Uid. But I guess these are functionally equivalent in this situation. They should only differ if the entity is in the middle of changing parent or something like that.

ElectroJr (approved):

PR #3655, opened 2023-01-05
Add helper for SnapGridPosAt

metalgearsloth:
Added it to the system and renamed it for general use.


Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
```diff
@@ -86,6 +86,32 @@ public static Vector2i GetChunkRelative(Vector2i tile, byte chunkSize)
 
     #endregion
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
DrSmugleaf: You sure this is necessary/a benefit

DrSmugleaf (approved):

PR #3658, opened 2023-01-05
Ensure DestroyContacts is called on fixture deletion

metalgearsloth:
Don't think this fixes the exception in CollideContacts though I'm having a hard time reproing and it seems to be client-only maybe, assuming physics is the source?


Robust.Shared/Physics/Systems/FixtureSystem.cs
```diff
@@ -178,18 +178,17 @@ private void OnShutdown(EntityUid uid, FixturesComponent component, ComponentShu
                 return;
             }
 
+            foreach (var contact in fixture.Contacts.Values.ToArray())
+            {
+                _physics.DestroyContact(contact);
+            }
+
+            var xform = Transform(body.Owner);
```
metalgearsloth:
```suggestion
```

PR #3666, opened 2023-01-07
Made RSI validator account for RGA in RSIs

AJCM-git:
Temporary measure so its at least supported until we convert them all

### DIFF
--- a/Schemas/rsi.json
+++ b/Schemas/rsi.json
@@ -55,10 +55,12 @@
     ],
     "required": [
         "version",
-        "license",
-        "copyright",
         "size",
         "states"
+    ],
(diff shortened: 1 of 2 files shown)


EmoGarbage404:
Are RGAs meant to be used for textures? I thought they were pnly being used for audio files as of right now.

AJCM-git:
When i asked what should we do to handle multiple copyrights for different files in the same RSI sloth told me the plan was to move them all to RGA

ike709:
> Are RGAs meant to be used for textures?

No, I planned on writing a RSIv2 spec at some point after RGAs but never got around to it.

PR #3675, opened 2023-01-10
Fix incorrect size of second window in split container

eoineoineoin:
CL https://github.com/space-wizards/RobustToolbox/commit/45295d4b49ca16ba5ae8c06635051c257c3ec0e7 introduced a small bug where the second widget in a SplitContainer could have the wrong measure applied. This manifested in some strange-looking layouts of that widget's children:

Peek 2023-01-10 23-02

The additional size adjustment for the padding of the SplitContainer's second widget was being applied twice, and, in the second instance, was erroneously subtracting a scalar from a vector - really feels like that should be a domain error, but it looks like there's a bunch of code that relies on "Vector2.operator -(Vector2 a, float b)" (and operator+)

### DIFF
--- a/Robust.Client/UserInterface/Controls/SplitContainer.cs
+++ b/Robust.Client/UserInterface/Controls/SplitContainer.cs
@@ -291,7 +291,7 @@ namespace Robust.Client.UserInterface.Controls
                 else
                     size.X = availableSize.X - _splitStart - _splitWidth;
 
-                size = Vector2.ComponentMax(availableSize, Vector2.Zero);
+                size = Vector2.ComponentMax(size, Vector2.Zero);
                 second.Measure(size);
             }
             else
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/Controls/SplitContainer.cs
```diff
@@ -291,7 +291,7 @@ protected override Vector2 MeasureOverride(Vector2 availableSize)
                 else
                     size.X = availableSize.X - _splitStart - _splitWidth;
 
-                size = Vector2.ComponentMax(availableSize - _splitStart - _splitWidth, Vector2.Zero);
```
ElectroJr, line 294: Ah my bad, this bit was supposed to be replaced with the modified `size` calculated just above this code, but I guess I forgot to replace it. I'll modify this n then merge.

eoineoineoin: Ah, silly mistake on my part; I should have been paying more attention. Thanks for fixing!

PR #3676, opened 2023-01-10
Adds a WCvar to override the CEF user-agent

ike709:
OpenDream needs this to trick older implementations of TGUI into thinking it's Internet Explorer.

Here's a random OD window that I used to test setting it to `opendream useragent test`:

image


DrSmugleaf (approved):

moonheart08 (approved):
i read some of the cef api just to review this

PR #3681, opened 2023-01-13
Try fix cvar toml type parsing

### LINKED ISSUE #3680: Wrong rules.time config variable results in serious client's bugs
## Description
Client get a bugs with UI and chat channels when there are a wrong variable was set for rules.time cvar. This cvar requires `simple` type to be set, for example `100.0`, but not `100`. If you set `100`, you will get only error in client console and no rules will be shown. When game starts and player rejoins server, his chat channels disappears and UI no longer work. Why engine doesn't check cvars for correct variable types or not converting them? We didn't know that this bug appeared because of that shit for 1 month

**Reproduction**
Setup server and add this cvar in server config
```
[rules]
time = 45
```

**Screenshots**
The error in console
изображение
No UI windows and no chat channels
изображение

ElectroJr:
This tries to fix #3680. Cvar loading/registering & cvar def types in general are a bit of a mess & I'm not at all sure if this is the best way to do this, but it seem to fix the issue. 

It would be nice if there were just some kind of generic "can be cast from" but that doesn't exist AFAIK.

### DIFF
--- a/Robust.Shared/Configuration/ConfigurationManager.cs
+++ b/Robust.Shared/Configuration/ConfigurationManager.cs
@@ -299,6 +299,20 @@ namespace Robust.Shared.Configuration
                 if (cVar.Registered)
                     Logger.ErrorS("cfg", $"The variable '{name}' has already been registered.");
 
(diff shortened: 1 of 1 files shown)


ZoldorfTheWizard (approved):
better something than nothing.

PR #3684, opened 2023-01-13
Auto-componentstate source generator

mirrorcult:
basic concept:

Requires https://github.com/space-wizards/space-station-14/pull/14845

- `AutoGenerateComponentState` attr on component
- `AutoNetworkField` attr on component members (fields/properties)
- only works in shared
- will autogenerate something like this:

```csharp
// <auto-generated />
using Robust.Shared.GameStates;
using Robust.Shared.Analyzers;
using Robust.Shared.Serialization;

namespace Content.Shared.Access.Components;

public partial class IdCardComponent
{
    public class IdCardComponent_AutoState : ComponentState
    {
        public string FullName = default!;
        public string JobTitle = default!;
    }

    [RobustAutoGenerated]
    public class IdCardComponent_AutoNetworkSystem : EntitySystem
    {
        {
            SubscribeLocalEvent<IdCardComponent, ComponentGetState>(OnGetState);
            SubscribeLocalEvent<IdCardComponent, ComponentHandleState>(OnHandleState);
        }

        private void OnGetState(EntityUid uid, IdCardComponent component, ref ComponentGetState args)
        {
            args.State = new IdCardComponent_AutoState
            {
                FullName = component.FullName,
                JobTitle = component.JobTitle,
            };
        }

        private void OnHandleState(EntityUid uid, IdCardComponent component, ref ComponentHandleState args)
        {
            if (args.Current is not IdCardComponent_AutoState state)
                return;

            component.FullName = state.FullName;
            component.JobTitle = state.JobTitle;
        }
    }
}
```

ancillaray things that had to get done to support this:

- access analyzer was changed to ignore special types with the `RobustAutoGenerated` attribute, so that this doesn't step on the toes of access (which really should just work with nested classes, but there is zero support for finding that info out through roslyn was far as I can tell)
- component events were made public with a very stern warning to not use them. this is because of the `AfterAutoHandleState` behavior where a special component state can be raised after handling state is done, to facilitate things like updating appearance/raising events after state setting is done. an alternative solution is to partial the entity system instead and call a partial method that can be overriden, but partialing the component is much cleaner IMO so eventbus it is

todo
- [x] uhh uhmm access
- [x] itd be really nice if systems had a way to perform actions after the handlestate is done, this is a really common workflow in content (eg updating visuals). i had an implementation of this that used component events to raise an event after handle but then i promptly realized that component events are engine only so. idk. maybe i can get an exception for that
- [x] content PR 
- [x] & update docs & make issue to replace stuff (will do when me rged)


mirrorcult:
redy 4 review

vulppine (commented):
i was doing PR review in SS14 and was looking through resolved reviews for it, one of which was a mark for having a really big component state: is there a way to filter out which fields we wouldn't want to be auto-generated in a component state (i.e., a dictionary or something), or is that something that would call for a manual implementation of a component's networked state?

mirrorcult:
you have to manually mark fields you want with `[AutoNetworkedField]` so you can filter out stuff you dont want by just not giving it that attribute

PR #3694, opened 2023-01-15
Add support for singleplayer commands

ElectroJr:
## The issue
Currently there are several "server only" commands like `rmmap` and `tp` which should just be moved to shared code & restricted when not using robust as an engine for a single player game. This also applies to things like load & save map, though that currently relies on server-exclusive code.

## What this PR changes

This PR adds a new `RequireServerOrSingleplayer` bool to `IConsoleCommand` which will cause the client to functionally ignore the presence of a command while connected to a server (i.e. execution & command completions are automatically redirected to the server).

## Other changes
- Moves various commands from server code to shared code with `RequireServerOrSingleplayer` set to true. This is the bulk of the PR line change count.
- The client now clears server-side command when disconnecting from the server & stores them separately from other registered commands.
- IConsoleHost.RegisteredCommands has been renamed to available commands.
  -  Requires space-wizards/space-station-14/pull/14041
- Adds `RequireServerOrSingleplayer` as an optional argument of the `RegisterCommand()` functions
- Adds overrides of the `RegisterCommand()` functions that automatically fetch localized help & description strings based on the command name. Basically a sort of variant of `LocalizedCommand`.


Robust.Client/Console/ClientConsoleHost.cs
```diff
@@ -67,12 +70,49 @@ public void Initialize()
...
+            _availableServerCommands.Clear();
+            UpdateAvailableCommands();
+        }
+
+        private void UpdateAvailableCommands()
+        {
+            // so uhhh. AvailableCommands are all the currently registered commands.
+            // and uhhh RegisteredCommands are all the currently available commands.
+            // makes perfect sense.
+            // I cbf fixing this and bloating the PR.
```
ElectroJr: This is just bad variable naming that requires a content PR to fix

ElectroJr: fuck it, ill just make a content PR

PR #3696, opened 2023-01-15
Lets forces act on kinematic controllers.

TemporalOroboros:
Physics bodies with the kinematic controller bodytype allow forces, torques, and impulses to be applied similar to dynamic physics bodies. However, when the physics system runs it only applies forces to dynamic physics bodies, kinematic controllers are skipped. This results in some inconsistencies where applying an impulse to a kinematic controller works just fine since the change to the entities momentum is applied immediately but applying a force over time fails to have any effect at all.

This resolves the above inconsistency by allowing forces to apply to physics bodies.


metalgearsloth:
Kinematic bodies should be set directly.

TemporalOroboros:
> Kinematic bodies should be set directly.

In that case I would like some advice on how to make singularities smoothly pull on people.

PR #3704, opened 2023-01-16
Fix `SpriteComponent.CopyFrom()`

ElectroJr:
Converted to draft because its broken. I hate sprite code.

Copying from prototype sprites is currently broken because of recent changes that I endorsed. This PR fixes that.

This PR moves most of the initialization logic out of the `ISerializationHook` and into a sprite system function. The copying code now calls that function when copying from an uninitialized sprite. Eventually this should probably just become a serializer, but AFAIK the technology just isn't there yet.

This also gets rid of some old data-fields like `layerDatums`.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
@@ -1,14 +1,18 @@
-using System;
 using System.Collections.Generic;
+using System.Linq;
 using JetBrains.Annotations;
 using Robust.Client.ComponentTrees;
 using Robust.Client.Graphics;
+using Robust.Client.ResourceManagement;
 using Robust.Shared;
 using Robust.Shared.Configuration;
 using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
+using Robust.Shared.Log;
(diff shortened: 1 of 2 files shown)


Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
```diff
@@ -34,9 +38,65 @@ public override void Initialize()
...
+                else
+                    Logger.ErrorS(SpriteComponent.LogCategory, "Unable to load RSI '{0}'.", rsiPath);
+            }
+            else
+            {
+                sprite.BaseRSI = null;
+            }
+
+            // TODO this should be done via a custom type serializer for layer data
+            if (layers != null && layers.Count() > 0)
```
metalgearsloth, line 76: Would .Any() be better?

metalgearsloth (approved):

Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
ElectroJr: Actually it should just be `Count` not `Count()`. I assume `Any()` will be slower bc linq?

PR #3708, opened 2023-01-18
Use sprite offset for SpriteView

metalgearsloth:


ElectroJr:
Should probably be publicly settable options so that controls can choose whether or not to add the offset.

PR #3720, opened 2023-01-22
Set null shader if none specified

metalgearsloth:
Copies the LayerSetShader behaviour to the LayerSetState behavior

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -305,7 +305,7 @@ namespace Robust.Client.GameObjects
         void ISerializationHooks.AfterDeserialization()
         {
             // Please somebody burn this to the ground. There is so much spaghetti.
-            
+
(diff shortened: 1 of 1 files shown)


AJCM-git (approved):

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -765,6 +765,11 @@ public void LayerSetData(int index, PrototypeLayerData layerDatum)
                         layerDatum.Shader);
                 }
             }
+            else
+            {
```
ElectroJr, line 769: This will break visualizers because there will be no way to specify non shader related sprite data without overriding the existing shader. It should only clear the shader if its intentionally being cleared but not if its just the default / not-specified value (null).

Easiest way is to just do nothing if null and clear if only if its an empty string. Otherwise the layer datadef needs to somehow differentiate between "null because it it was set to null" and "null because it was not specified".

ElectroJr (changes requested):

PR #3723, opened 2023-01-22
Fix bug where ScrollContainer could cause layout engine to loop forever.

eoineoineoin:
Copy-pasting repro from https://github.com/space-wizards/space-station-14/issues/13641

> To repro (there's a slighly simpler way to repro it, I'm sure):
> 1. Write a paper with the contents "11111 11111i 11111ii 11111iii 11111iiii 11111iiiii 11111iiiiii 11111iiiiiii 11111iiiiiiii 11111iiiiiiiii 11111iiiiiiiiii 11111iiiiiiiiiii 11111iiiiiiiiiiii 11111iiiiiiiiiiiii 11111iiiiiiiiiiiiii 11111iiiiiiiiiiiiiii 11111iiiiiiiiiiiiiiii 11111iiiiiiiiiiiiiiiii"
> 2. Resize the paper so the entered text just fills up the the whole page
    Hang forever

What's happening is that the ScrollContainer and RichTextLabel don't get on. The ScrollContainer decides it needs to enable the scrollbar, which causes an InvalidateMeasure(), then, on next layout, the RichTextLabel decides it needs to add a line break to account for the slightly more narrow space. Then, on next layout, the scroll bar isn't necessary anymore, which causes an InvalidateMeasure() and....

ScrollContainer _attempts_ to account for the size of the scroll bars, but when they are `Visible==false`, their `DesiredSize` is 0. This adds the ability for ScrollBars to be Visible for the purposes of layout, but won't draw themselves if not necessary. Longer-term, this feels like it would be a nice to break Visibility out into `Invisible` `Invisible-but-takes-up-space` and `Visible`.

I tested this on a bunch of the other UIs which use ScrollContainer, and they seem to be behaving correctly.

### DIFF
--- a/Robust.Client/UserInterface/Control.cs
+++ b/Robust.Client/UserInterface/Control.cs
@@ -28,6 +28,9 @@ namespace Robust.Client.UserInterface
         private readonly List<Control> _orderedChildren = new();
 
         private bool _visible = true;
+        // Determines if this control requires space, even when
+        // it's visibility has been set to false
+        private bool _reservesSpace = true;
 
         // _marginSetSize is the size calculated by the margins,
         // but it's different from _size if min size is higher.
@@ -170,7 +173,7 @@ namespace Robust.Client.UserInterface
         /// <seealso cref="VisibleInTree"/>
         [ViewVariables(VVAccess.ReadWrite)]
         [Animatable]
-        public virtual bool Visible
+        public bool Visible
         {
             get => _visible;
             set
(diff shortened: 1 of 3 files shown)


PJB3005 (changes requested):
This should just be a new boolean property on `ScrollBar`, not making `Visible` virtual.

Robust.Client/UserInterface/Control.cs
```diff
@@ -207,6 +210,33 @@ private void _propagateVisibilityChanged(bool newVisible)
...
         }
 
+        /// <summary>
+        ///     Whether or not this control and its children require
+        ///     space to be reserved, even when not visible.
+        /// </summary>
+        /// <seealso cref="ReservesSpace"/>
+        [ViewVariables(VVAccess.ReadWrite)]
+        [Animatable]
+        public virtual bool ReservesSpace
```
PJB3005: Remove `virtual` here.

PJB3005 (changes requested):
Minor nit otherwise looks good.

Robust.Client/UserInterface/Control.cs
eoineoineoin: Whoops, my bad; sorry.

PJB3005 (approved):

PR #3724, opened 2023-01-22
Refactor NextWordPosition to EndWordPosition

Visne:
Instead of finding the start of the next word, it now finds the end of the current word. The new behaviour is consistent with any application I know of/have tried, and is less weird (before, moving a word left would not add the whitespace between the current and previous word, while moving a word right would add the whitespace in between the current and next word).


Visne:
Tests pass locally btw

Edit: while other commits are applied 😵‍💫

PR #3725, opened 2023-01-22
Implement deleting a full word at a time

Visne:
Requires #3724. Makes it possible to remove a word at a time, just like Ctrl+Backspace and Ctrl+Delete do in almost all programs.


metalgearsloth:
PR name?

Visne:
Oops, fixed

PR #3742, opened 2023-01-28
Rich Text refactor

juliangiebel:
This PR replaces the current FormatedText implementation with a more flexible one that supports adding new tags by implementing the `IMarkupTag` interface.

The new parser parses the markup into a list of `MarkupTag` nodes.

It's more flexible and supports the following features:
- Arbitrary tag names
- One parameter parsed into a nodes `Value` field
- An arbitrary number of attributes which are put into a nodes `Attributes` dictionary
- Escaping reserved characters using backslashes
- Self closing tags
- String, long and color parameters

New tags are created by implementing the `IMarkupTag` interface which contains methods for modifying the context for rendering the text (currently only color and font) and supplying additional text.

This PR also adds a `cmdlink` tag because I needed to test inline controls:
`[cmdlink="Click Here To Win 10000€" command="ghost"]`

And a `b` tag because of ike:

grafik
*The characters for starting and ending a tag are still `[` and `]`. I want to make that a CVar*

I wrote a doc about what this refactor does: https://hackmd.io/@dUUm5XdUR2Gd9I7x4_Weyg/r1lzQwPRss
Content PR: https://github.com/space-wizards/space-station-14/pull/13772


juliangiebel:
grafik

I added the `italic` and the `font` tag.
The font tag lets you specify a font and size to use.

*Allowing limited rich text in chat isn't hard to do but that's for another PR*

Robust.Client/UserInterface/RichText/ItalicTag.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Robust.Client.UserInterface.RichText;
+
+public sealed class ItalicTag : IMarkupTag
+{
+    [Dependency] private readonly IResourceCache _resourceCache = default!;
+    public string Name => "italic";
+
+    public void PushDrawContext(MarkupNode node, MarkupDrawingContext context)
+    {
+        var font = FontTag.CreateFont(context.Font, node, _resourceCache, "/Fonts/NotoSans/NotoSans-Italic.ttf");
```
moonheart08: Don't hardcode this. Need to figure out a better way to grab the required font from styles. @PJB3005 as i think you're the only one who'd know how.

Robust.Client/UserInterface/RichText/BoldTag.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Robust.Client.UserInterface.RichText;
+
+public sealed class BoldTag : IMarkupTag
+{
+    [Dependency] private readonly IResourceCache _resourceCache = default!;
+    public string Name => "bold";
+
+    public void PushDrawContext(MarkupNode node, MarkupDrawingContext context)
+    {
+        var font = FontTag.CreateFont(context.Font, node, _resourceCache, "/Fonts/NotoSans/NotoSans-Bold.ttf");
```
moonheart08: see italic

Robust.Client/UserInterface/RichText/FontTag.cs
```diff
@@ -0,0 +1,61 @@
...
+/// </summary>
+public sealed class FontTag : IMarkupTag
+{
+    [Dependency] private readonly IResourceCache _resourceCache = default!;
+
+    public string Name => "font";
+
+    public void PushDrawContext(MarkupNode node, MarkupDrawingContext context)
+    {
+        string fontPath = node.Value.StringValue ?? "/Fonts/NotoSans/NotoSans.ttf";
```
moonheart08: see italic

moonheart08 (changes requested):
Almost all looks good.

Robust.Client/UserInterface/RichText/ItalicTag.cs
juliangiebel: I implemented the `FontPrototype` which contains the fonts resource path. I hope that's decent enough as I still had to hardcode the prototype Ids

PR #3743, opened 2023-01-29
Make HistoryLineEdit invoke OnTextChanged

ElectroJr:


Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -57,32 +57,42 @@ public class LineEdit : Control
...
+        }
 
-                if (!InternalSetText(value))
-                {
-                    return;
-                }
+        /// <summary>
+        ///     Sets the current text, optionally invoking <see cref="OnTextChanged"/>.
+        /// </summary>
+        public void SetText(string value, bool invokeEvent = false)
```
DrSmugleaf, line 72: Should this be protected instead? Or is there a reason something else would want to control whether or not to invoke the event?

DrSmugleaf (approved):

Robust.Client/UserInterface/Controls/LineEdit.cs
ElectroJr: IMO it should just be public because there will probably be instances where controls want to update the data/text without unnecessarily invoking on-change events. Other controls like SpinBox also have a function to set without invoking.

PR #3744, opened 2023-01-30
VV invoke/paths rejig

ElectroJr:
There were a few things that bothered me with the new VV paths/commands and this PR tries to fix them, though I'm, not at all sure that I'm taking the best approach for each one.

Requires space-wizards/space-station-14/pull/13792

## Better filtering when listing paths

Currently vv-read/write/invoke console completion will return forbidden/invalid paths. I.e., vvinvoke will returns paths regardless of whether they are actually invokable. This PR fixes this by:
- Splitting `VVAccess` into read, write, and execute flags
- Making the invoke/read/write commands use the appropriate flags for `VVListPathOptions.MinimumAccess` (previously was always just `VVAccess.ReadOnly`)
- The reflection based paths just check the VV attributes's access to see if they satisfy `VVListPathOptions.MinimumAccess` 
- The `VVListPathOptions` are now also passed onto custom domain/type handlers so they can use that information however they like to restrict the returned paths.

In order to ensure that methods with a vv attribute are always executable (otherwise what is the point?) I decided to make the `ViewVariablesAttribute` no longer be applicable to methods, and there is now a separate `VVInvokableAttribute` that inherits from `ViewVariablesAttribute` but can only be applied to methods and has no access customization.

## inherited private members

Apparently `Type.GetMembers()` fails to return private members from an object's base class, which means that VV paths currently can't access any private methods defined in things like shared systems. This just makes the get-member functions use the same solution already used in old vv code (see `TypeHelper.GetAllFields()`).

## Error messages / feedback

Quite often if you execute an invalid vv command (e.g. failing to parse arguments for a vvinvoke) the command just returns "null" without any indication that something went wrong. This PR tries to fix that by:

- Making remote functions like `InvokeRemotePath()` return the `ViewVariablesResponseCode`.
- If the response code is not "Ok", the returned string is either null or some informative error message. 
- Replaced several `DoThing(path)` type functions with `bool TryDoThing(path, out result, out string? error)`
  - Maybe its better to just make this use exceptions & try-catches instead of having `out string?`s everywhere.
- The console commands print any errors returned by the vv functions. 

## EntityUid parsing

I CBF using `/entity/123` whenever I want to pass in an entity to a function, so I added a VV yaml serialization context. So now you can use either `123` or the old `/entity/123`.

## TODO
Other things that should probably be done:
- vvinvoke console completions should show a summary of argument & return types.
- On a related note: vvinvoke needs to be better at dealing with methods with the same name but different signatures.
- I guess the new errors should probably be localized?

### DIFF
--- a/Robust.Client/ViewVariables/ClientViewVariablesManager.Domains.cs
+++ b/Robust.Client/ViewVariables/ClientViewVariablesManager.Domains.cs
@@ -19,7 +19,7 @@ internal sealed partial class ClientViewVariablesManager
             : null, segments);
     }
 
-    private IEnumerable<string>? ListGuiHoverPaths(string[] segments)
+    private IEnumerable<string>? ListGuiHoverPaths(string[] segments, VVListPathOptions options)
     {
         return null;
     }
--- a/Robust.Client/ViewVariables/ViewVariablesCommand.cs
+++ b/Robust.Client/ViewVariables/ViewVariablesCommand.cs
@@ -16,6 +16,7 @@ namespace Robust.Client.ViewVariables
         [Dependency] private readonly IEntityManager _entities = default!;
 
         public override string Command => "vv";
+        protected override VVAccess RequiredAccess => VVAccess.ReadOnly;
 
         public override void Execute(IConsoleShell shell, string argStr, string[] args)
         {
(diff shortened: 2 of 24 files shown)


gradientvera:
Will review properly sometime this week but for now these two things stand out to me from reading the PR description and skimming through the code:
- I think it'd be best if the VV methods would just throw exceptions instead. The out string parameter for the error is not good practice, and not enough things are gonna consume this API so that using the TryX pattern here would make sense (In an ideal world we could just use something like `Result<T, E>` in Rust but sadly the technology isn't there yet)
- For the remote methods, I'd prefer using a struct over a tuple for the returned values. (although frankly this is yet again one of those cases where I *yearn* for a `Result<T, E>`)

ElectroJr:
I'll get around to redoing this sometime later.

PR #3747, opened 2023-02-01
Add capability to queue entity spawn for end of current update cycle

sephtasm:
### **What?**
Adds a new method to entity manager that allows a request to spawn an entity rather than doing it immediately.

### **Why?**
This makes it possible to queue up a request to spawn entities while some system is in the middle of an `EntityQuery`. Otherwise, you might create an entity that modifies the `EntityQuery` while it's being iterated through, which is an exception.

It doesn't happen normally but you can reproduce the issue as follows:
1. Spawn Urist
2. Replace all of Urist's "Blood" with "BuzzochloricBees"
3. Attack Urist with a slashing weapon
4. During the Bloodstream system's update call it will query all entities with a bloodstream
5. Urist's bloodstream component will spill some bee/blood out, which spawns a new bee
6. The bee is an entity with a Bloodstream, so the collection has been modified while being iterated, which throws

### **How?**
This commit adds a `QueueSpawnEntity` method to `IEntityManager `which takes the same arguments as `SpawnEntity` but does not immediately return anything. Instead the `EntityManager` will wait until the end of the current update cycle to spawn the entities, similar to how deleting works. You can pass the coords for the entity's spawn location as MapCoords or EntityCoords and it will convert them. This doesn't match what `SpawnEntity` does exactly but it looks better to me. I might be wrong though.

I don't really know the codebase that well so I didn't want to make any changes to existing behavior. I think improvements could be made to the entity spawning system that would completely avoid the need for this sort of change, but I'm not confident enough to try them.

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.cs
+++ b/Robust.Shared/GameObjects/EntityManager.cs
@@ -15,7 +15,7 @@ using System.Linq;
 namespace Robust.Shared.GameObjects
 {
     public delegate void EntityUidQueryCallback(EntityUid uid);
-
+    
     /// <inheritdoc />
     [Virtual]
     public partial class EntityManager : IEntityManager
@@ -48,6 +48,7 @@ namespace Robust.Shared.GameObjects
         /// <inheritdoc />
         public virtual IEntityNetworkManager? EntityNetManager => null;
 
+        protected readonly Queue<(string?, MapCoordinates)> QueuedEntitySpawns = new();
         protected readonly Queue<EntityUid> QueuedDeletions = new();
         protected readonly HashSet<EntityUid> QueuedDeletionsSet = new();
 
@@ -153,6 +154,15 @@ namespace Robust.Shared.GameObjects
                 QueuedDeletionsSet.Clear();
             }
 
+            using (histogram?.WithLabels("QueuedSpawn").NewTimer())
+            using (_prof.Group("QueueSpawn"))
+            {
+                while (QueuedEntitySpawns.TryDequeue(out (string? uid, MapCoordinates coordinates) t))
+                {
+                    SpawnEntity(t.uid, t.coordinates);
+                }
+            }
+
             using (histogram?.WithLabels("ComponentCull").NewTimer())
             using (_prof.Group("ComponentCull"))
             {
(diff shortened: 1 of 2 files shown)


metalgearsloth:
Might be better to have the offending systems handle it / provide a locking assertion that throws if systems try to do this?

As now you might get scenarios where for example the spawn map / entitycoordinates becomes invalid and the entity is never cleaned up.

sephtasm:
> Might be better to have the offending systems handle it / provide a locking assertion that throws if systems try to do this?

If I understand correctly, the effect that spawns the bee would call back to the chemistry reaction system asking it to queue an entity spawn, then during the chemistry reaction system's update loop it will spawn the entity safely?


> As now you might get scenarios where for example the spawn map / entitycoordinates becomes invalid and the entity is never cleaned up.

In either case I should still check that, right?

metalgearsloth:
> If I understand correctly, the effect that spawns the bee would call back to the chemistry reaction system asking it to queue an entity spawn, then during the chemistry reaction system's update loop it will spawn the entity safely?

Chemistry already needs adjusting for how reactions are done which needs doing content-side.

The problem with this approach is subject to even more issues:
- You convert EntityCoordinates to MapCoordinates but these are only valid when made. If someone decides to move the entire station for example then suddenly that mapcoordinates will spawn it in the original spot and not the new spot.
- If you do queue EntityCoordinates you get the aforementioned deletion issues.

sephtasm:
I'm willing to do this on the content side but I want to make sure I understand how to do it correctly. I think I'm starting to get what you meant about the locking assertion.

The problem is that delaying an entity spawn in any case is a bad idea because the spawn coordinates might be invalid by the time you get to it?

So things like the chemistry reaction / bloodstream interaction should be reworked to safely spawn an entity. Do you think a reasonable way to do that would be to simply have offending systems running an EntityQuery cache the results and then loop through them rather than looping through the EntityQuery itself?

```C#
        var bloodstreamToUpdate = EntityManager.EntityQuery<BloodstreamComponent>().ToList();
        foreach (var bloodstream in bloodstreamToUpdate)
        {
            ...
```
Or do you think something more comprehensive is needed?

metalgearsloth:
> So things like the chemistry reaction / bloodstream interaction should be reworked to safely spawn an entity. Do you think a reasonable way to do that would be to simply have offending systems running an EntityQuery cache the results and then loop through them rather than looping through the EntityQuery itself?

That or we get synchronisation points but that would require a lot of engine work I think.

PR #3748, opened 2023-02-01
Box2i serializer

metalgearsloth:
We have one for Box2 but not Box2i. This is just copied from Box2i with int parsers instead. I would not copy but I don't think there's an easy way around it.


Robust.Shared/Serialization/TypeSerializers/Implementations/Box2Serializer.cs
```diff
@@ -16,7 +16,7 @@ namespace Robust.Shared.Serialization.TypeSerializers.Implementations
     [TypeSerializer]
     public sealed class Box2Serializer : ITypeSerializer<Box2, ValueDataNode>, ITypeCopyCreator<Box2>
     {
-        private static void NextOrThrow(
+        internal static void NextOrThrow(
```
DrSmugleaf: You could make this class also implement the serializer methods for Box2i or add a partial that does it if you want to keep this private, it would make the file longer though

DrSmugleaf (approved):

PR #3751, opened 2023-02-03
Cleanup DynamicTypeFactory, remove redundant overloads.

gradientvera:


RELEASE-NOTES.md
```diff
@@ -51,7 +51,7 @@ END TEMPLATE-->
 
 ### Internal
 
-*None yet*
+- Cleaned up `DynamicTypeFactory` and its interface, removed redundant overloads.
```
PJB3005, line 54: This is not an internal change, `IDynamicTypeFactory` and its API are public.

PJB3005 (changes requested):
This is a breaking change and at the very least needs some `[Obsolete]` symbols if we're going to do this.

PR #3755, opened 2023-02-06
Fix bug where labels with `ClipText=true` don't render

eoineoineoin:
Fixes a bug where a `Label` wouldn't be visible in the UI if `ClipText` was enabled on the control.

### DIFF
--- a/Robust.Client/UserInterface/Controls/Label.cs
+++ b/Robust.Client/UserInterface/Controls/Label.cs
@@ -1,5 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using Robust.Client.Graphics;
 using Robust.Shared.Animations;
@@ -253,7 +254,7 @@ namespace Robust.Client.UserInterface.Controls
 
             if (ClipText)
             {
-                return (0, _cachedTextHeight / UIScale);
+                return (_cachedTextWidths.Max() / UIScale, _cachedTextHeight / UIScale);
             }
 
             var totalWidth = 0;


eoineoineoin:
This actually returns the same value as the code immediately below - it looks like it's actually sufficient to set RectClipContent=true, in order to get the effect that ClipText provides, so it's not clear to me why `_clipText` is a field at all? Objections to removing? Can keep the `ClipText` as a wrapper around `RectClipContent` just to keep the API easy to use?

PJB3005:
The problem here is that the label you're creating isn't given any sizing flags or minimum size, so the layout system assigns it a size of 0. The point of clip text is that overflow text (which gets clipped) does not affect layout of the rest of the UI. This is why it reports a size of 0 in measure: it can shrink as much as possible and take as much as assigned (standard layout behavior).

`RectClipContent` on the other hand is a rendering system flag that frankly shouldn't even be a property thing, but fixing that would require improving the UI rendering system drastically.

eoineoineoin:
That feels like a really strange API. If you have a `Label` with some long text, it can overflow past the edge of the `Control` - most of the time, I don't want to set a size (or minimum size) on a control, since I don't know the size of the children; but I also don't want the text to overflow into other elements. `ClipContentRect` seems to provide this behaviour, so it's really unclear to me how to use `ClipText` properly.

PJB3005:
To start with, I'd like to point out that the current change in the PR just makes `ClipText` do nothing, since both the code paths in `MeasureOverride` do exactly the same now.

> If you have a Label with some long text, it can overflow past the edge of the Control

This is only if the control is somehow restricted via a `MaxWidth` somewhere up the tree. Otherwise the control will take as much text as it needs to fit the full text.

A quick refresher on how the UI system works:
1. A control gives its *minimum desired* required size in `MeasureOverride()`
2. A control may then be given more space by the parent in `Arrange()` if available.

A button may only need as much space to fit the border + embedded text, but can be given more with `HAlignment.Stretch` which would give you a really big button with the label centered inside it.

`Label`'s default behavior (with `ClipText=false`) is to request a desired size that fits the whole text. This means that if you put the label in any sort of flexible container like a box container, it will push stuff out of the way horizontally to reach that. It also won't make any meaningful use of any additional space given to it.

The behavior with `ClipText=true` on the other hand is that label requests *no* horizontal space and instead takes the space given to it. This means that if you stick the label in a *horizontal* `BoxContainer`, it will be invisible with width zero (box containers only give out the desired space along their primary axis, unless the appropriate `*Expand` is set). However if you were to put the control in a position where it will get some sort of size (for example, a vertical box container, or a horizontal box container with a minimum width set, ...) then the control can still get some width and will have all text that's "too wide" clipped off.

Of course, some of those cases (horizontal box container with minwidth and clip on label) can be replicated with maxwidth instead, which will force the control to clip either way. However forcing a max width can't be made flexible/responsive in many uses.

metalgearsloth:
Re-open when you come back to this.

PR #3757, opened 2023-02-06
Revert "fix prototypemanager ignoring yaml files with the .yaml ending"

PJB3005:
This reverts commit 58c79b2c7bfe2cc54ceca3dcf81cf0754fcf7998.

Issues with this commit:
1. only updates *one* "yml" to also support "yaml". A cursory glance over the code for the string "yml" quickly shows 10+ cases that have been missed by the commit, in both prototype loading, prototype validation, tests, GitHub workflow configurations, texture loading, and probably something else I missed.
2. I mentioned on both Discord and GitHub this is not a good idea (as shown above). We should have a warning if .yaml is spotted anywhere, not allow loading it.

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
@@ -24,7 +24,7 @@ public partial class PrototypeManager
     {
         _hasEverBeenReloaded = true;
         var streams = Resources.ContentFindFiles(path)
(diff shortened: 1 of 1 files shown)


PaulRitter:
i updated the implementation to be more complete. imo its either maintaining a warning in multiple places or supporting .yaml in multiple places, i opted to support it.
if you and others believe its better to put a warning, feel free to open a pr implementing that, but dont just remove the .yaml support please since contribs have been stumbling over this issue.

PaulRitter (changes requested):
.

PJB3005:
Done by 1f1f58a1820d41294f260461fb046140c9d382a6

PR #3761, opened 2023-02-07
Add FastNoiseLite

metalgearsloth:
Also added a library credit for it.

https://github.com/Auburn/FastNoiseLite

I didn't drop FastNoise yet because Simplex4 support was dropped and the noise generator for parallax needs it so I left it as is.

The other benefit is FastNoiseLite has a GUI so it's much easier to play around with settings you want externally.


moonheart08:
Would appreciate it if you ping me before the full removal of old FastNoise @metalgearsloth as I may need to help some folks migrate.

PR #3763, opened 2023-02-07
Add mapsaveoptions for metadata

metalgearsloth:
Authors shouldn't get overwritten anymore but name would need content changes.

Will need mapper ping so they update their local names.


metalgearsloth:
Should be using RGA

PR #3798, opened 2023-02-22
Add option to hide scrollbars

ElectroJr:
Adds an option to scroll containers to hide invisible scrollbars.


eoineoineoin:
Looks like this change has caused https://github.com/space-wizards/space-station-14/issues/13641 to resurface. Repro steps are the same and setting ScrollContainer._reserveScrollbarSpace to `true` will fix it. It feels like any time this is `false` will potentially cause the bug where the layout engine loops forever. Not sure what to do here? Feels like this should have a more robust fix?

PR #3799, opened 2023-02-22
Floor Vector2 -> Vector2i conversions

metalgearsloth:
For consistency sake as 0.5 -> 0 yet -0.5 -> 0

### DIFF
--- a/Robust.Shared.Maths/Vector2i.cs
+++ b/Robust.Shared.Maths/Vector2i.cs
@@ -160,7 +160,7 @@ namespace Robust.Shared.Maths
 
         public static explicit operator Vector2i(Vector2 vector)
         {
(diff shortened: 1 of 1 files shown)


ElectroJr:
> For consistency sake

Isn't the current conversion the consistent one, seeing as the result is consistent with casting the individual components of the vector? ~~IMO it'd make more sense to just add a `Vector2.Floor()`.~~ nvm, `Vector2.Floored()` already exists.

metalgearsloth:
I'd be fine with that though I would probably want to remove the Vector2i cast as most of the engine should be flooring it due to tile conversions.

PJB3005:
Removing it definitely is not the best approach, whatever code using it for tile conversion should be fixed. Isn't that already a pretty small amount of code anyways?

PR #3800, opened 2023-02-22
Fix PanelContainer ArrangeOverride()

ElectroJr:
Currently the PanelContainer's `MeasureOverride()` treat style box margins/padding as physical pixels (i.e., not scaled by `UiScale`), while the `ArrangeOverride()` effectively treats them as virtual pixels. This PR changes the arrange method so that it also treats them like physical pixels, which fixes some UI layout bugs (e.g., scroll bar visibility in context menu pop-ups).

### DIFF
--- a/Robust.Client/UserInterface/Controls/PanelContainer.cs
+++ b/Robust.Client/UserInterface/Controls/PanelContainer.cs
@@ -34,13 +34,24 @@ namespace Robust.Client.UserInterface.Controls
 
         protected override Vector2 ArrangeOverride(Vector2 finalSize)
         {
-            var pixelSize = finalSize * UIScale;
-            var ourSize = UIBox2.FromDimensions(Vector2.Zero, pixelSize);
-            var contentBox = _getStyleBox()?.GetContentBox(ourSize) ?? ourSize;
+            var style = _getStyleBox();
+
+            UIBox2 contentBox;
+            if (style == null)
+            {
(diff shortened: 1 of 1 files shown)


PJB3005 (changes requested):
1. You would need to change `MeasureOverride()` as well.
2. I'm not sure it makes much sense to fix this without fixing the rest of the StyleBox rendering code to be properly scaled.

ElectroJr:
> 1. You would need to change MeasureOverride() as well.

AFAIK this change should make Arrange consistent with the current measure, they are currently inconsistent on master.

> 2. I'm not sure it makes much sense to fix this without fixing the rest of the StyleBox rendering code to be properly scaled.

If there's no real reason why they should even be using physical pixel's I'll just open a PR to make them use virtual pixels instead if I can get it to work easily with current style boxes.

PJB3005:
> AFAIK this change should make Arrange consistent with the current measure, they are currently inconsistent on master.

Ah wait, I see, you're right.

> If there's no real reason why they should even be using physical pixel's I'll just open a PR to make them use virtual pixels instead if I can get it to work easily with current style boxes.

I think when I wrote the UI scaling system I figured "style boxes don't scale" would be better than "style boxes scale blurrily" but please punch that past version of me if you meet them so.

AJCM-git:
@ElectroJr you coming back to this?

ElectroJr:
> @ElectroJr you coming back to this?

Yes, but it'll probably require starting from scratch anyways, so may as well close this for now

PR #3806, opened 2023-02-25
Set paused before ent startup when loading

metalgearsloth:
Means eventbus subs can do their own logic if the ent is paused.


Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
```diff
@@ -783,28 +791,6 @@ private void ResetNetTicks(EntityUid entity, MetaDataComponent metadata, Mapping
...
-    {
-        var isPaused = data.MapIsPaused;
-
-        if (data.MapIsPostInit)
-        {
-            metadata.EntityLifeStage = EntityLifeStage.MapInitialized;
-        }
-        else if (_mapManager.IsMapInitialized(data.TargetMap))
-        {
-            EntityManager.RunMapInit(uid, metadata);
```
DrSmugleaf, line 796: Is running RunMapInit not needed anymore?

metalgearsloth: ooooooooooooop

PR #3807, opened 2023-02-26
Refactor MSBuild spaghetti, new configurations.

PJB3005:
There is now a "DebugOpt" and "Tools" build configuration available, as discussed on the last maintainer meeting.

Project files and MSBuild stuff has been somewhat cleaned up.


moonheart08:
Could Release be made to fail on all non-obsolete warnings if I do the fixes?

PJB3005:
Is that a "is it technically possible" or a "should we do this" discussion?

moonheart08:
> Is that a "is it technically possible" or a "should we do this" discussion?

"should we do this", as we discussed prior this would be fine if GHA only (Which I opposed due to local testing of buildability being important), with this people could just build a Tools build if they still have warnings.

PJB3005:
Fine by me, there's little reason to use a release build for anything other than a reference anyways.

moonheart08 (approved):

DrSmugleaf (approved):

PR #3808, opened 2023-02-26
Adds an option to control rendering strategies on a layer basis for sprites

Menshin:
**Context**
While trying to make freezers/heaters rotatable again in the SS14 codebase, we've encountered a case where some layers needed to have differents rendering strategies (the machine body shouldn't rotate and always faces south while the pipes connectors can rotate freely or snap to cardinals).
See https://github.com/space-wizards/space-station-14/pull/14225#issuecomment-1440991771 and https://github.com/space-wizards/space-station-14/pull/14225#issuecomment-1442670391

The current "fix" is to put the whole sprite on No Rotation and duplicates all icons (for all states) so they exist for the four directions. That's obviously far from ideal.

**In this PR**
What i propose here is to add attributes to both the sprite component and the layer object that would allows layer-based rendering differentiation :

* the sprite's `GranularLayersRendering` is set to `false` by default, so every existing sprites will work as usual with the minimal overhead of checking a bool
* if `GranularLayersRendering` is set to `true` (only possible in the prototype), then, when the sprite renders, it computes the three matrices related to the default/norotation/snapcardinal strategies and apply them on a per-layer basis, using the layer's `LayerRenderingStrategy` (here again set in the prototype) to select which one to apply.

Unfortunately, the yaml data type for `LayerRenderingStrategy` is a string, because it proved a real pain to make it an enumeration.

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -44,6 +44,13 @@ namespace Robust.Client.GameObjects
         /// </summary>
         public static double DirectionBias = -0.05;
 
+        /// <summary>
+        ///     Whether the layers have independant drawing strategies, e.g some may snap to cardinals while others won't.
+        ///     The sprite should still set its global rendering method (e.g NoRot or SnapCardinals), this only gives finer control over how layers are rendered internally.
+        /// </summary>
+        [DataField("granularLayersRendering")]
+        public bool GranularLayersRendering = false;
+
         [DataField("visible")]
         private bool _visible = true;
 
@@ -338,6 +345,7 @@ namespace Robust.Client.GameObjects
                         Color = Color.White,
                         Scale = Vector2.One,
                         Visible = true,
+                        RenderingStrategy = LayerRenderingStrategy.UseSpriteStrategy,
                     });
                     state = null;
                     texture = null;
@@ -393,6 +401,7 @@ namespace Robust.Client.GameObjects
             }
 
             RenderOrder = other.RenderOrder;
+            GranularLayersRendering = other.GranularLayersRendering;
         }
 
         internal void UpdateLocalMatrix()
@@ -802,6 +811,8 @@ namespace Robust.Client.GameObjects
                 }
             }
 
+            layer.RenderingStrategy = layerDatum.RenderingStrategy ?? layer.RenderingStrategy;
+
             layer.Color = layerDatum.Color ?? layer.Color;
             layer._rotation = layerDatum.Rotation ?? layer._rotation;
             layer._offset = layerDatum.Offset ?? layer._offset;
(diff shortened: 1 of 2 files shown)


Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1275,9 +1312,32 @@ internal void RenderInternal(DrawingHandleWorld drawingHandle, Angle eyeRotation
...
+                    {
+                        case eLayerRenderingStrategy.NoRotation:
+                            layer.Render(drawingHandle, ref transformNoRot, angle, overrideDirection);
+                        break;
+                        case eLayerRenderingStrategy.SnapCardinals:
+                            layer.Render(drawingHandle, ref transformSnap, angle, overrideDirection);
+                        break;
+                        default:
+                            layer.Render(drawingHandle, ref transform, angle, overrideDirection);
+                        break;
```
ElectroJr: If the sprite has norot or snap enables, then this matrix won't be the "default" non-snap / non-norot matrix. So I think you also just need to calculate that separately

ElectroJr:
> Unfortunately, the yaml data type for LayerRenderingStrategy is a string, because it proved a real pain to make it an enumeration.

What was the issue? as long as the enum is defined in shared code it should work fine as a datafield.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
Menshin: You're right, i'm adding that.

Menshin:
> > Unfortunately, the yaml data type for LayerRenderingStrategy is a string, because it proved a real pain to make it an enumeration.
> 
> What was the issue? as long as the enum is defined in shared code it should work fine as a datafield.

Iirc, the enum wasn't automatically serialized, i'll check again now i understand the codebase better.

Menshin:
Ok, i've fixed the "default" behavior and defaulted to use the Sprite drawing strategy (meaning you only have to change specific layers and not give everyone of them a rendering strategy, less line to edit overall and easier to read for the yaml file).

The enum is still giving me trouble though : all is tested and working with integers in the prototype definition, but using things like `renderingStrategy: enum.LayersRenderingStrategy.NoRotation` instead produce a `enum value not found` error.

ElectroJr:
> The enum is still giving me trouble though : all is tested and working with integers in the prototype definition, but using things like renderingStrategy: enum.LayersRenderingStrategy.NoRotation instead produce a enum value not found error.

You only need to use the "full" enum string if you're reading a generic enum. I.e., if you had
```cs
[DataField("renderingStrategy")] public Enum? RenderingStrategy;
```
then you would have to use "enum.LayerRenderingStrategy.NoRotation" in yaml. But in this case the datafield is for an explicit enum type, i.e.,
```cs
[DataField("renderingStrategy")] public LayerRenderingStrategy? RenderingStrategy;
```
so you can just use "NoRotation" in yaml. Also in your GH comment you said "enum.LayersRenderingStrategy.NoRotation" instead of "enum.LayerRenderingStrategy.NoRotation" which might also have caused issues if you used that in yaml while testing.

Menshin:
Thanks for your detailed explanations, i realize i should have read the whole enumations handling code more thoroughly...

Tested and working as intended on my side :
image

Switching it back to ready for review.

PR #3810, opened 2023-02-27
Fix AnchorEntity replication when the coordinate doesn't change

ghost:
before:

https://user-images.githubusercontent.com/32521367/221577317-eea0ba30-c042-4257-aea6-a83168c196a9.mp4

after:

https://user-images.githubusercontent.com/32521367/221576311-6657dc4b-8a61-48ae-87e2-bab3598d36bf.mp4

EntityCoordinates() dirty the component if there is a coordinate change, it is why we have to use the Dirty function. I tried to refactor this function and directly return if the component is already anchored. But It breaks the station initialization. Do you know why ?

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
@@ -83,6 +83,7 @@ public abstract partial class SharedTransformSystem
             return false;
 
         var wasAnchored = xform._anchored;
+        Dirty(xform);
         xform._anchored = true;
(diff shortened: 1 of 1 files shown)


ElectroJr:
>  I tried to refactor this function and directly return if the component is already anchored. But It breaks the station initialization. Do you know why ?

The transform component/system has a lot of cursed old code when it comes to entity initialization and startup. I haven't checked what the actual issue is here but it's not surprising that it unexpectedly broke. It should hopefully get fixed whenever someone fixes the initialization spaghetti code.

PR #3817, opened 2023-03-03
Make map loader more robust

Partmedia:
## What
RobustToolbox's map loader is unrobust and cowardly bails out (usually with a massive backtrace in the console) when somebody tries to load old maps.

This PR makes RobustToolbox more robust by trying harder to load maps with missing prototypes and tiles (old maps). Drop entities with missing prototypes and replace missing tiles with Plating so that these maps at least load.

## Why
image

Obviously, mappers still have to go find and replace the missing things, but this is infinitely better than giving up and starting a new version of Moose based on screenshots of the original, which is what mappers currently do.

## Results
This PR allows the current version of SS14 to load the original Moose, which was made a Long Time Ago:

image

Notice that because somebody thought it was a good idea to rename all of the tiles (including space tiles), the backup Plating tile is loaded along chunkbb boundaries.

I wrote a vimscript to CamelCase the tilemap strings, which results in most of the correct tiles loading:

image

Due to personal safety reasons I am not submitting this vimscript at this time.

### DIFF
--- a/Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
+++ b/Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
@@ -404,7 +404,8 @@ public sealed class MapLoaderSystem : EntitySystem
             }
             catch (UnknownPrototypeException e)
             {
-                Logger.ErrorS("loader", "Ignoring unknown prototype " + type.ToString());
+                if (type != null)
+                    Logger.ErrorS("loader", "Ignoring unknown prototype " + type.ToString());
             }
         }
 
--- a/Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
+++ b/Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
@@ -253,10 +253,10 @@ public sealed class MapLoaderSystem : EntitySystem
         return true;
     }
 
-    private bool Deserialize(MapData data)
+    private bool Deserialize(MapData data, bool strict = false)
     {
         // Verify that prototypes for all the entities exist
-        if (!VerifyEntitiesExist(data))
+        if (strict && !VerifyEntitiesExist(data))
             return false;
 
         // First we load map meta data like version.
@@ -390,15 +390,21 @@ public sealed class MapLoaderSystem : EntitySystem
                 uid = uidNode.AsInt();
             }
 
-            var entity = _serverEntityManager.AllocEntity(type);
-            data.Entities.Add(entity);
-            data.UidEntityMap.Add(uid, entity);
-            data.EntitiesToDeserialize.Add(entity, entityDef);
+            try {
+                var entity = _serverEntityManager.AllocEntity(type);
+                data.Entities.Add(entity);
+                data.UidEntityMap.Add(uid, entity);
+                data.EntitiesToDeserialize.Add(entity, entityDef);
 
-            if (data.Options.StoreMapUids)
+                if (data.Options.StoreMapUids)
+                {
+                    var comp = _serverEntityManager.AddComponent<MapSaveIdComponent>(entity);
+                    comp.Uid = uid;
+                }
+            }
+            catch (UnknownPrototypeException e)
             {
-                var comp = _serverEntityManager.AddComponent<MapSaveIdComponent>(entity);
-                comp.Uid = uid;
+                Logger.ErrorS("loader", "Ignoring unknown prototype " + type.ToString());
             }
         }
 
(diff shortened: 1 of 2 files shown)


Partmedia:
Whoops, I had warnings off, will fixe if there is conceptual approval.

Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
```diff
@@ -253,10 +253,10 @@ private bool TryGetReader(ResourcePath resPath, [NotNullWhen(true)] out TextRead
         return true;
     }
 
-    private bool Deserialize(MapData data)
+    private bool Deserialize(MapData data, bool strict = false)
```
moonheart08, line 256: Strict by default please.

Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
```diff
@@ -390,15 +390,21 @@ private void AllocEntities(MapData data)
...
+                data.EntitiesToDeserialize.Add(entity, entityDef);
 
-            if (data.Options.StoreMapUids)
+                if (data.Options.StoreMapUids)
+                {
+                    var comp = _serverEntityManager.AddComponent<MapSaveIdComponent>(entity);
+                    comp.Uid = uid;
+                }
+            }
+            catch (UnknownPrototypeException e)
```
moonheart08, line 405: ...check if the prototype exists instead?
exceptions are multiple times slower than doing the check.

Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
Partmedia: Sure, I'll also have to revisit the patch to identify the paths that should not be strict, but I'll turn this on by default.

Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
Partmedia: `PrototypeManager`'s `TryIndex` method simply throws an `UnknownPrototypeException`. Is there a better public method for querying whether a prototype exists?

metalgearsloth: No it doesn't? It only does that if the type doesn't exist (and because it's an entityprototype it will).

metalgearsloth (changes requested):
Given how frequently mappers and newbie coders show up in discord asking for help without even having read the console for errors I don't really want this unless we make it extremely obvious because what's most likely going to happen is some error occurs and the mapper resaves the map blissfully unaware.

Partmedia:
> Given how frequently mappers and newbie coders show up in discord asking for help without even having read the console for errors I don't really want this unless we make it extremely obvious because what's most likely going to happen is some error occurs and the mapper resaves the map blissfully unaware.

To the typical SS14 user/mapper, you can either have (1) current behavior which is functionally throwing away all of their work by having their map not load, or (2) try to rescue as much of their map as you can, and permitting whatever is recovered to be saved.

While that in general I typically agree that a program should never do something that causes data to be silently discarded:

1. This isn't silent, it's on them for not reading the console
2. It should be the responsibility of the calling code to decide what to do; the loader should always be robust to finish loading the thing instead of bailing out on an exception
3. It is better than losing their whole map and causing them to start over. Do you know how the current Moose was mapped?

This situation is not ideal. Ideally, we would have never broken maps in the first place. Or made whoever broke maps responsible for writing a migration for every breaking change (like what happens with the SQL migrations). I think this solution, perhaps with some changes, is a whole lot better than what we have now.

metalgearsloth:
> It is better than losing their whole map and causing them to start over. Do you know how the current Moose was mapped?

The code only affects loading not saving, which means you already have your work saved and it's not being thrown away. I see far more people throw work away accidentally mapping on already initialized maps.

I agree they should read console but this is the reality based on discord help.

The only time I might want this is for centcomm and nukeops map loading but considering how infrequently grafana gets checked I would rather just have a specific test to handle those.

Partmedia:
> The only time I might want this is for centcomm and nukeops map loading but considering how infrequently grafana gets checked I would rather just have a specific test to handle those.

This is not about loading CentCom or the nukie planet. This is about making things easier for mappers loading in old maps. This code should never run for a production server, because they would already have been checked in strict mode.

I promised to change strict mode to be the default, I just haven't gotten around to this yet.

metalgearsloth:
I think it should be its own command as I don't think there's a good reason to use it upstream and don't want any of our mappers to get the wrong impression and use it for regular mapping, saying this as someone who has done this type of work previously (e.g. stationstation, gun yml update, saltern). Even better if it just parses component names without loading to validate it can load.

In most of those cases we renamed prototypes far more frequently than removed them which requires going into the yml and knowing what needs updating (e.g. generator renames, gun renames, wall rename, container culling off the top of my head for stuff I did), where deleting the prototype is not wanted and just getting a list of what needs updating is preferable.

metalgearsloth:
I talked with emisse about it and having more useful errors in console is preferred, particularly with all invalid tiles being replaced with plating, though I imagine that's out of scope as that would require a bigger PR.

PR #3819, opened 2023-03-04
EmptyContainer tweaks

ElectroJr:
This PR has some empty-container changes for a content PR (requires space-wizards/space-station-14/pull/14406).

- ContainerSystem's `EmptyContainer()` function now returns removed entities.
- The function's arguments have been renamed to be consistent with `IContainer.Remove()`
- Added a `TerminatingOrDeleted()` helper function
- Moved `AttachToGridOrMap()` from the transform component to the system & marked the component method as obsolete.


Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
```diff
@@ -63,14 +63,20 @@ private void HandleComponentState(EntityUid uid, ContainerManagerComponent compo
             if (args.Current is not ContainerManagerComponentState cast)
                 return;
 
+            var query = EntityManager.GetEntityQuery<TransformComponent>();
```
AJCM-git: proxy method

PR #3845, opened 2023-03-12
Make DataDefinitions check for private constructors

ElectroJr:


Robust.Shared/Serialization/Manager/SerializationManager.cs
```diff
@@ -119,9 +119,10 @@ IEnumerable<Type> GetImplicitTypes(Type type)
                 }
 
                 var isRecord = records.ContainsKey(type);
-                if (!type.IsValueType && !isRecord && !type.HasParameterlessConstructor())
+                if (!type.IsValueType && !isRecord && !type.HasParameterlessConstructor(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic))
                 {
-                    sawmill.Debug(
+                    // This should probably just be an error. But I CBF tying this to a content PR.
```
PaulRitter: imo this shouldnt be an error since we may have "unintended" signups through methods like ImplicitDataDefinitionForInheritors etc. i think simply logging a debug message for this is better than forcing people to exclude certain classes

PaulRitter (approved):

Robust.Shared/Serialization/Manager/SerializationManager.cs
ElectroJr: The issue is that this can introduce hidden errors because people seem to have just been ignoring this debug message, and it currently prevents some maps/entities from being serialized. At the very least it should be a warning so that it is more prominent.

~~Would you be fine with eventually making it an error if we exempt abstract classes?~~ Never mind, they are already excluded.

PaulRitter: abstract classes & interfaces should already be omitted for signup. i honestly dont know if making it an error eventually would be fine. i'd like to think so, im not opposed but i havent recently looked at signups to see how many usecases there are (and could be) that would need to be explicitely excluded

PR #3851, opened 2023-03-14
Fix audio IoC errors

ElectroJr:
Stops IoC threading exceptions when logging errors in frame update.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
@@ -8,6 +8,7 @@ using Robust.Client.Graphics;
 using Robust.Client.ResourceManagement;
 using Robust.Shared;
 using Robust.Shared.Audio;
+using Robust.Shared.Exceptions;
 using Robust.Shared.GameObjects;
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -398,13 +410,10 @@ private PlayingStream CreateAndStartPlayingStream(IClydeAudioSource source, Audi
         if (!source.SetPosition(fallbackCoordinates.Position))
         {
             source.Dispose();
-            Logger.Warning($"Can't play positional audio \"{stream.Name}\", can't set position.");
+            _sawmill.Warning($"Can't play positional audio \"{stream.Name}\", can't set position.");
             return null;
         }
 
-        if (!coordinates.IsValid(EntityManager))
-            coordinates = fallbackCoordinates;
```
ElectroJr, line 406: I'm not sure why this was here before. Its always possible that the entity that this sound is attached to will be sent to the client in the future, so it should just be using the fall-back as the actual fall-back, not as the main coordinate.

metalgearsloth (approved):

PR #3854, opened 2023-03-19
PSS - Persistence - Fixes and tweaks

ingles98:
~~**IMPORTANT**: The change to **ActorComponent** will break current SS14, the content-side PR that I am creating addresses this change accordingly.~~

Content PR: https://github.com/space-wizards/space-station-14/pull/14740

Please read the file comments.

Includes several engine-side changes for making persistence possible:
- Added missing List serializers - Required for persistence
- Fixed precision of Float Serializer - General impact
- ~~Changed ActorComponent's PlayerSession to nullable - The actor component holds data for a previously saved player's mob, after a persistence load, there is no PlayerSession, therefore makes sense having it as nullable.~~

### DIFF
--- a/Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
+++ b/Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
@@ -988,7 +988,7 @@ public sealed class MapLoaderSystem : EntitySystem
 
             foreach (var component in EntityManager.GetComponents(entityUid))
             {
-                if (component is MapSaveIdComponent)
+                if (component is MapSaveIdComponent || component is ActorComponent)
                     continue;
 
                 var compType = component.GetType();
--- a/Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
+++ b/Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
@@ -1018,6 +1018,44 @@ public sealed class MapLoaderSystem : EntitySystem
                         context: _context);
                 }
 
+                void Bandaid_MappingDataNode_RemoveNullsRecursive(MappingDataNode node)
+                {
+                    foreach (var m in node.ToArray())
+                    {
+                        if (m.Value.IsNull || m.Key.IsNull)
+                            node.Remove(m);
+                        if (m.Value is MappingDataNode mv)
+                        {
+                            Bandaid_MappingDataNode_RemoveNullsRecursive(mv);
+                            if (mv.IsEmpty) node.Remove(m);
+                        }
+                        else if (m.Value is SequenceDataNode sv)
+                        {
+                            Bandaid_SequenceDataNode_RemoveNullsRecursive(sv);
+                            if (sv.IsEmpty) node.Remove(m);
+                        }
+                    }
+                }
+                void Bandaid_SequenceDataNode_RemoveNullsRecursive(SequenceDataNode node)
+                {
+                    foreach (var m in node.ToArray())
+                    {
+                        if (m.IsNull)
+                            node.Remove(m);
+                        if (m is MappingDataNode mv)
+                        {
+                            Bandaid_MappingDataNode_RemoveNullsRecursive(mv);
+                            if (mv.IsEmpty) node.Remove(m);
+                        }
+                        else if (m is SequenceDataNode sv)
+                        {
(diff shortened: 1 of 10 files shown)


Robust.Server/GameObjects/EntitySystems/MapLoaderSystem.cs
```diff
@@ -1018,6 +1018,44 @@ private void WriteEntitySection(MappingDataNode rootNode, Dictionary<int, Entity
                         context: _context);
                 }
 
+                void Bandaid_MappingDataNode_RemoveNullsRecursive(MappingDataNode node)
```
ingles98, line 1021: This is purposefully named this way as a temporary solution, awaiting a better solution.

Currently there are assertions in Map Loading against null/empty values, therefore on Save I am clearing those. This also helps a little with save size I guess.

PJB3005:
> Changed ActorComponent's PlayerSession to nullable - The actor component holds data for a previously saved player's mob, after a persistence load, there is no PlayerSession, therefore makes sense having it as nullable.

Would it not make more sense to simply not store the `ActorComponent` and have it disappear upon load?

ingles98:
> > Changed ActorComponent's PlayerSession to nullable - The actor component holds data for a previously saved player's mob, after a persistence load, there is no PlayerSession, therefore makes sense having it as nullable.
> 
> Would it not make more sense to simply not store the `ActorComponent` and have it disappear upon load?

How would i persist user's mob attachment ?

ingles98:
> 

What i mean is, ActorComponent is NEEDED and holds the data needed for a persisted player to come back to its mob.

EmoGarbage404:
> > 
> 
> What i mean is, ActorComponent is NEEDED and holds the data needed for a persisted player to come back to its mob.

isn't that already supported without ActorComponent? Players can leave games and rejoin without the actor component being needed. And if it's a big issue, perhaps use another component?

ingles98:
> 

The issue is not about players leaving or joining games, it's about persisting through saves. Forks that would use this system could implement handling player mob's in a completely different way, this is simply barebones player persistence where a player mob will sit idle, sleeping after a save + reload where the player that owns that mob hasn't logged in yet.

On PSS13 for instance, players would usually be moved (or move themselves before going away) to cryo for deep storage where the mob is saved and disposed until the player comes back, but that's too specific and still should work even if the player's mob wasnt ever disposed/stored properly.

I think I understand the issue, perhaps actor components could simply be removed and then readded when the player joins back in as the mob ownership is stored in the MindComponent, I can look into that and block Actor from being saved, rolling back changes to Actor.PlayerSession

ingles98:
> > > 
> > 
> > 
> > What i mean is, ActorComponent is NEEDED and holds the data needed for a persisted player to come back to its mob.
> 
> isn't that already supported without ActorComponent? Players can leave games and rejoin without the actor component being needed. And if it's a big issue, perhaps use another component?

I rolled back the ActorComponent change, it was a dumb oversight. Tested it and is working as before.

Robust.Shared/Serialization/TypeSerializers/Implementations/Primitive/FloatSerializer.cs
```diff
@@ -24,14 +24,14 @@ public sealed class FloatSerializer : ITypeSerializer<float, ValueDataNode>
...
         {
-            return float.Parse(node.Value, CultureInfo.InvariantCulture);
+            return float.Parse(node.Value);
         }
 
         public DataNode Write(ISerializationManager serializationManager, float value,
             IDependencyCollection dependencies, bool alwaysWrite = false,
             ISerializationContext? context = null)
         {
-            return new ValueDataNode(value.ToString(CultureInfo.InvariantCulture));
```
metalgearsloth, line 34: Won't removing invariantculture cause issues with non-english serialization?

ingles98: Writing in G9 is Invariant. Parsing however I don't think it makes any difference, but I could be wrong. Will see about testing those.

ingles98: Also this Float fix isn't too specific for persistence though it does generate loss of precision. This can be removed and looked into in more detail on another topic

metalgearsloth:
sortedset would probably be good to get in its own PR but the other bits look a bit too bandaidy atm.

PR #3858, opened 2023-03-22
Multi-shader rendering

amylizzle:
Adds a path for rendering with multiple shaders at once, without requiring ping-pong use of `RenderInRenderTarget()`

Note: I fucked up
```
[StructLayout(LayoutKind.Explicit)]
private struct RenderCommand
```
because I don't entirely understand its purpose or how to fix it, so someone may want to check that out before merging.


amylizzle:
I don't understand how shaders work, so closing on the advice of someone who does (PJB)

PR #3865, opened 2023-03-24
Adjust center point on non 32px Sprites if SpriteOffset == false

jamessimo:
This is to fix an issue with setting SpriteOffset to false on sprites that are larger than 32px.

If SpriteOffset is false in the draw entity it would just pass in a Vec2.zero, however this was never tested as SpriteOffset=false has not been used by any content. It doesn't work as expected and this is designed to fix that.

For 32px this will set to 0,1. if 64 will be 0,0.5, if 96 will be 0,0.33 etc... This was designed to not impact the larger sprite system. This will not impact any sprites unless explicitly you set there SpriteOffet to false (true by default)

### DIFF
--- a/Robust.Client/UserInterface/Controls/SpriteView.cs
+++ b/Robust.Client/UserInterface/Controls/SpriteView.cs
@@ -61,8 +61,13 @@ namespace Robust.Client.UserInterface.Controls
             }
 
             var uid = Sprite.Owner;
-            //Add the Sprite.Offset and if SpriteOffset = false, adjust the SpriteOffset center point if the sprite is > 32px.
-            var  offsetAdj = (SpriteOffset ? Sprite.Offset : new Vector2(0,(1.0f/(32*Sprite.Bounds.Height))*32));
+            var offsetAdj = Vector2.Zero;
+
+            //Set SpriteOffset, If the sprite is > 32px, adjust the SpriteOffset center point
+            if(Sprite.Bounds.Height > 1)
+            {
+                offsetAdj = (SpriteOffset ? Sprite.Offset : new Vector2(0,(1.0f/(32*Sprite.Bounds.Height))*32));
+            }
 
             _spriteSystem ??= IoCManager.Resolve<IEntitySystemManager>().GetEntitySystem<SpriteSystem>();
             _spriteSystem?.ForceUpdate(uid);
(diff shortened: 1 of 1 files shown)


jamessimo:
Hey guys, anything blocking this? It’s needed in order to fix some in game issue. It’s a very small change and doesn’t impact any thing right now since nothing uses spriteoffset =false

Robust.Client/UserInterface/Controls/SpriteView.cs
```diff
@@ -61,11 +61,13 @@ internal override void DrawInternal(IRenderHandle renderHandle)
...
 
             var uid = Sprite.Owner;
+            //Add the Sprite.Offset and if SpriteOffset = false, adjust the SpriteOffset center point if the sprite is > 32px.
+            var  offsetAdj = (SpriteOffset ? Sprite.Offset : new Vector2(0,(1.0f/(32*Sprite.Bounds.Height))*32));
 
             _spriteSystem ??= IoCManager.Resolve<IEntitySystemManager>().GetEntitySystem<SpriteSystem>();
             _spriteSystem?.ForceUpdate(uid);
 
-            renderHandle.DrawEntity(uid, PixelSize / 2 + PixelSize * (SpriteOffset ? Sprite.Offset : Vector2.Zero), Scale * UIScale, OverrideDirection);
+            renderHandle.DrawEntity(uid, PixelSize / 2 + PixelSize * offsetAdj, Scale * UIScale, OverrideDirection);
```
metalgearsloth: This will break 0-offset entities that are 32px in size.

e.g. if sprite offset is disabled for context menu entities (the old behaviour) then they won't draw inside of the spriteview correctly.

metalgearsloth (dismissed):

Robust.Client/UserInterface/Controls/SpriteView.cs
jamessimo: why would you SpriteOffset false sprites with no offset? 

But to fix this should I just do check if 32 and if 32 just vector 0 it?

metalgearsloth: No offset should just draw the sprite from 0,0 of the control and should only potentially break if the sprite does have an offset.

jamessimo:
is there anything blocking this?

ElectroJr:
> is there anything blocking this?

No its just that there's not many active engine maintainers and its relatively low priority, largely because its just a band aid for something that needs to be fixed properly later on anyways

PR #3872, opened 2023-03-26
Remove MapId serializer and rename `WorldMap` to `MapId`

ElectroJr:
AFAIK, the MapId serializer isn't actually used by anything. If anything currently is using it, it should be replaced with just using the map's EntityUid.

Requires https://github.com/space-wizards/space-station-14/pull/14857


Robust.Shared/Map/Components/MapComponent.cs
```diff
@@ -11,19 +11,12 @@ namespace Robust.Shared.Map.Components
     [NetworkedComponent]
     public sealed class MapComponent : Component
     {
-        [ViewVariables(VVAccess.ReadOnly)]
-        [DataField("index")]
-        private MapId _mapIndex = MapId.Nullspace;
-
```
metalgearsloth, line 17: This was just a bandaid around non-edited components not being serialized which has since been fixed.

PR #3884, opened 2023-03-27
Wrap contact events in try-catch

metalgearsloth:
Content has broken it too many times so we need this. Unfortunately there's no comp not found exception for events.

See https://github.com/space-wizards/space-station-14/issues/14892

Hopefully without the contact exception spam that follows this I'll get some more reports with the root causes as people normally read the bottom exception.


moonheart08:
This almost definitely has performance impact, have you profiled?

metalgearsloth:
> This almost definitely has performance impact, have you profiled?

Currently admins are restarting the server daily but if they ping me instead I'd be fine with that.

PR #3895, opened 2023-03-31
Fix discord null errors

ElectroJr:
Prevents null exceptions when updating player lists in some niche situations.


moonheart08:
approved but i'm too lazy to do an engine release rn so you do it

PR #3900, opened 2023-04-01
Obsolete comp references

metalgearsloth:
Mum sloth is obsoleting code again.

The performance cost of basic component add / removal operations is not worth this.

### DIFF
--- a/Robust.Shared/GameObjects/ComponentReferenceAttribute.cs
+++ b/Robust.Shared/GameObjects/ComponentReferenceAttribute.cs
@@ -1,26 +1,30 @@
 using System;
 
-namespace Robust.Shared.GameObjects
+namespace Robust.Shared.GameObjects;
+
+/*
+ * Obsoleted because this has a massive performance cost and it is not worth it.
+ */
+
+/// <summary>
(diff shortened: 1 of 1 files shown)


mirrorcult:
also we will probably need this dead for arch anyway

metalgearsloth:
Yeah we do I was talking to smug a bit about it.

I realise getcomp / trygetcomp at least don't need it so I can prototype pvs reversion before arch.

DrSmugleaf:
image

ElectroJr:
This shouldn't be obsoleted until we have a replacement for server/client components that inherit from shared components. Because currently people still **need** to use it for that.

EmoGarbage404:
The main uses right now are just:
- gasmixture (should be in shared
- textures (honestly not a clue)

ElectroJr:
Is the plan to just forbit shared comp inheritance?

DrSmugleaf:
> Is the plan to just forbit shared comp inheritance?

With struct components you wouldn't be able to do it anyway. For those that really need it, you can put whatever data in a separate component, which is less expensive with archetypes.

PR #3919, opened 2023-04-09
Create entities before applying entity states.

metalgearsloth:
Doesn't seem to break anything?

The issue is due to some internal transform ordering that happened at least a month+ ago afaict. Because the entity state for the parent (the grid) isn't applied yet then the child doesn't have the correct griduid to use.

By applying the parent first (instead of in a separate step) it ensures all the data is available.


Robust.Client/GameStates/ClientGameStateManager.cs
```diff
@@ -629,16 +633,42 @@ public IEnumerable<EntityUid> ApplyGameState(GameState curState, GameState? next
...
-            var toCreate = new Dictionary<EntityUid, EntityState>();
             var enteringPvs = 0;
-
+            _toApply.Clear();
+            _toCreate.Clear();
             var curSpan = curState.EntityStates.Span;
+
+            // Create new entities
+            {
+                using var _ = _prof.Group("Create uninitialized entities");
```
metalgearsloth:
```suggestion

            // Create new entities
            // This is done BEFORE state application to ensure any new parents exist before existing children have their states applied, otherwise, we may have issues with entity transforms!
            {
                using var _ = _prof.Group("Create uninitialized entities");
```

Too lazy to checkout branch again.

PR #3923, opened 2023-04-10
Refactor fixture serialization

metalgearsloth:
ATM these get serialized as a sequence when practice it aligns more with a mapping node. Long-term we should pull the ID and Body fields off of fixtures but baby steps so there's some placeholder code in here as duct tape in the short-term.

Needs content PR because of the yaml changes to every fixture in the game.

https://github.com/space-wizards/space-station-14/pull/15254


Robust.Shared/Physics/Systems/FixtureSystem.cs
```diff
@@ -249,7 +245,13 @@ private void OnHandleState(EntityUid uid, FixturesComponent component, ref Compo
...
             }
 
-            component.SerializedFixtureData = null;
+            // State handling funnies, someday we'll remove fixture.ID and fixture.Body and it won't matter
+            foreach (var (id, fixture) in component.Fixtures)
+            {
+                fixture.ID = id;
+                fixture.Body = physics;
+            }
+
```
metalgearsloth: This bit is particularly skrunkly as state handling is being applied before init AFAICT so body is never set.

PR #3930, opened 2023-04-12
Add BroadphaseNetworkingTest

ElectroJr:
Adds a new test, which would have failed before #3919 was merged.

Also renames the broadphase `MapUid` field to `PhysicsMap` to make it somewhat clearer that this field is related to physics and won't be set to anything for sundries.

### DIFF
--- a/Robust.UnitTesting/Shared/Physics/BroadphaseNetworkingTest.cs
+++ b/Robust.UnitTesting/Shared/Physics/BroadphaseNetworkingTest.cs
@@ -1,0 +1,154 @@
+using System.Linq;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Robust.Server.Player;
+using Robust.Shared;
+using Robust.Shared.Configuration;
+using Robust.Shared.GameObjects;
+using Robust.Shared.Map;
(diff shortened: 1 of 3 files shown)


Robust.UnitTesting/Shared/Physics/BroadphaseNetworkingTest.cs
```diff
@@ -0,0 +1,154 @@
...
+            await client.WaitRunTicks(1);
+        }
+
+        // Player & server xforms should match.
+        broadphase = new BroadphaseData(grid2, map2, true, false);
+        Assert.That(cPlayerXform.GridUid, Is.EqualTo(grid2));
+        Assert.That(sPlayerXform.GridUid, Is.EqualTo(grid2));
+        Assert.That(cPlayerXform.MapUid, Is.EqualTo(map2));
+        Assert.That(sPlayerXform.MapUid, Is.EqualTo(map2));
+        Assert.That(cPlayerXform.Broadphase, Is.EqualTo(broadphase));
```
ElectroJr, line 150: `cPlayerXform.Broadphase` would have been null prior to #3919

PR #3940, opened 2023-04-17
Reorder initialization slightly.

moonheart08:
This allows postinit to set a game state that has a UIScreen attached.


moonheart08:
@Jezithyr as this may affect SS14, i only tested this on a different project.

metalgearsloth:
yeet

PR #3945, opened 2023-04-19
Crude IsDefault checks for entities

metalgearsloth:
I mainly just wanted SOMETHING for now for biomes and I'd have to dig in serv3 for a while to get proper component diffs, hence the obsolete warning.


ElectroJr:
FWIW `UninitializedSaveTest` has a more thorough difference check, although it'll obviously also be slower.

PR #3955, opened 2023-04-21
Add sf3 support

lever1209:
turns out fluidsynth supports sf3 sound fonts, so i added sf3 to the extensions of loaded sound font files


gradientvera (approved):

gradientvera:
Untested, but Fluidsynth has supported sf3 since version v1.1.7 and this seems to me like it should just work.

PR #3961, opened 2023-04-23
Fix pvs chunk update bug

ElectroJr:
Fixes a situations where PVS could add a single entity to both grid & map chunks. This then caused some debug asserts to fail with the message "Root node count is 2 instead of 1."


Robust.Server/GameStates/PVSCollection.cs
```diff
@@ -441,6 +441,8 @@ public void UpdateIndex(TIndex index, EntityUid gridId, Vector2i chunkIndices, b
             oldGrid.ChunkIndices == chunkIndices &&
             oldGrid.GridId == gridId)
         {
+            _locationChangeBuffer.Remove(index);
+
```
ElectroJr, line 445: the changes to this file is the actual bugfix. the rest is misc cleanup, more asserts, and a new test.

PR #3964, opened 2023-04-23
'fix' transform coordinates assert

metalgearsloth:
Planetmaps function by adding gridcomponent to them hence can be both maps and grids at the same time.

Easiest way to repro is to spawn the salvate expedition console and pick a destruction mission.

### DIFF
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -614,7 +614,7 @@ internal sealed partial class PVSSystem : EntitySystem
             if (chunkLocation is MapChunkLocation)
                 DebugTools.Assert(xform.GridUid == null || xform.GridUid == uid);
             else if (chunkLocation is GridChunkLocation)
-                DebugTools.Assert(xform.GridUid != null && xform.ParentUid != xform.MapUid);
+                DebugTools.Assert(xform.GridUid != null);
 #endif
         }
 
--- a/Robust.Server/GameStates/PVSSystem.cs
+++ b/Robust.Server/GameStates/PVSSystem.cs
@@ -614,7 +614,7 @@ internal sealed partial class PVSSystem : EntitySystem
             if (chunkLocation is MapChunkLocation)
                 DebugTools.Assert(xform.GridUid == null || xform.GridUid == uid);
             else if (chunkLocation is GridChunkLocation)
-                DebugTools.Assert(xform.GridUid != null);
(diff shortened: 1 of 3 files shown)


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -512,11 +512,6 @@ public void SetCoordinates(EntityUid uid, TransformComponent xform, EntityCoordi
         if (!xform.Initialized)
             return;
 
-#if DEBUG
-        // If an entity is parented to the map, its grid uid should be null (unless it is itself a grid)
-        if (xform.ParentUid == xform.MapUid)
-            DebugTools.Assert(xform.GridUid == null || xform.GridUid == uid);
-#endif
         var newPosition = xform._parent.IsValid() ? new EntityCoordinates(xform._parent, xform._localPosition) : default;
         var moveEvent = new MoveEvent(uid, oldPosition, newPosition, oldRotation, xform._localRotation, xform, _gameTiming.ApplyingState);
```
ElectroJr, line 521: assuming planet maps still set GridUid, maybe just replce this with
```suggestion
#if DEBUG
        // If an entity is parented to the map, its grid uid should be null (unless it is itself a grid or we have a map-grid)
        if (xform.ParentUid == xform.MapUid)
            DebugTools.Assert(xform.GridUid == null || xform.GridUid == uid || xform.GridUid == xform.MapUid);
#endif
        var moveEvent = new MoveEvent(uid, oldPosition, newPosition, oldRotation, xform._localRotation, xform, _gameTiming.ApplyingState);
```

Robust.Server/GameStates/PVSSystem.cs
```diff
@@ -614,7 +614,7 @@ private void OnMapCreated(MapChangedEvent e)
             if (chunkLocation is MapChunkLocation)
                 DebugTools.Assert(xform.GridUid == null || xform.GridUid == uid);
             else if (chunkLocation is GridChunkLocation)
-                DebugTools.Assert(xform.GridUid != null && xform.ParentUid != xform.MapUid);
+                DebugTools.Assert(xform.GridUid != null);
 #endif
```
ElectroJr: Do map-grids use GridChunkLocations instead of MapChunkLocation?
In which case, could the second condition just be replaced with `&& (xform.ParentUid != xform.MapUid || xform.GridUid == xform.MapUid)`

PR #3966, opened 2023-04-24
Add dupe prototype ID validation

metalgearsloth:
Twice in 2 days.

Needs https://github.com/space-wizards/space-station-14/pull/15722 to not fail content.

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
@@ -123,21 +123,20 @@ public partial class PrototypeManager
                     var mapping = node.ToDataNodeCast<MappingDataNode>();
                     var id = mapping["id"].ToString();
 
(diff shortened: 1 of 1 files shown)


Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
```diff
@@ -115,14 +118,27 @@ public partial class PrototypeManager
...
                     }
 
+                    var before = seen.GetOrNew(type);
+
                     var mapping = node.ToDataNodeCast<MappingDataNode>();
+                    var id = mapping["id"].ToString();
+
+                    if (!before.Add(id!))
+                    {
+                        throw new PrototypeLoadException($"Found duplicate prototype definition {id} for type {type}");
```
DrSmugleaf: This will stop any further prototypes from getting validated, it should add an error node instead.

DrSmugleaf (changes requested):

PR #3975, opened 2023-04-28
Add basic map loader support for entity renaming

ElectroJr:
This just makes map loader raise an event that fetches a dictionary to rename entity prototypes. Also has basic support for deleting entities, but might horribly break in some situations (e.g., if the entity is in a container or has entities contained inside of it).


metalgearsloth:
conflicts

PR #3978, opened 2023-04-29
Generalize sprite view

ElectroJr:
This PR is meant to try remove any hard coded assumptions about sprite sizes from the SpriteView control, among some other things:

- You can now specify a sprite's world-rotation
- You can set the "eye" rotation
- The control will try and to place the centre of the sprite's bounding box in the centre of the view
- Sprite offsets are now ignored by default.
- The view now properly measures the size required to draw the sprite.
- There is a new Stretch-mode field which determines how sprites are stretched
  - The default is `Fit`, which will scale sprits down if the available size is insufficient to draw the sprite
  - `Fill` will scale a sprite up in order to fill the available size
  - `None` does no sprite scaling, and may lead to the sprite getting cropped if there is insufficient size

I started going insane trying to test & fix all the misc transform issues, so I've also added a new sprite view tab to the UI test window to make it easier to look for bugs in the future. The UI code itself is pretty shitty, but it is a test UI. I CBF improving it so if its an issue I can just remove it instead.

https://user-images.githubusercontent.com/60421075/235285511-16802070-384a-49de-b1ee-5fe233edb482.mp4

Requires space-wizards/space-station-14/pull/15881

### DIFF
--- a/Robust.Client/UserInterface/Controls/SpriteView.cs
+++ b/Robust.Client/UserInterface/Controls/SpriteView.cs
@@ -5,6 +5,7 @@ using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
 using Robust.Shared.Maths;
 using Robust.Shared.Utility;
+using Robust.Shared.ViewVariables;
 
 namespace Robust.Client.UserInterface.Controls
 {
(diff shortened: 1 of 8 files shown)


Robust.Client/UserInterface/Controls/SpriteView.cs
```diff
@@ -11,9 +13,86 @@ namespace Robust.Client.UserInterface.Controls
...
+            get => _eyeRotation;
+            set
+            {
+                _eyeRotation = value;
+                InvalidateMeasure();
+            }
+        }
+
+        /// <summary>
+        /// used to override the entity's world rotation. Note that the desired size of the control will not
```
metalgearsloth:
```suggestion
        /// Used to override the entity's world rotation. Note that the desired size of the control will not
```

metalgearsloth (approved):

PR #3987, opened 2023-04-30
Make RGA field URL optional. 

ElectroJr:


gradientvera:
You should probably modify the [RGA specification](https://github.com/space-wizards/RobustToolboxSpecifications/tree/master/RobustGenericAttribution) as well.

PR #3990, opened 2023-05-01
Small shader code cleanup

ElectroJr:
Small shader cleanup. Should have no effect on functionality. Maybe some of this was intentionally like this for some reason I don't know about, but IMO these changes make shader code somewhat less convoluted to read. 

Shaders are still a mess of confusingly differentiated shader objects which make it a pain to parse the code. E.g., ShaderPrototype, ShaderSourceResource, ShaderInstance, LoadedShaderInstance, LoadedShader, ParsedShader, and GLShaderProgram & a bunch of ClydeHandles that could ambiguously be referring to several different kinds of shader objects.

- Combines the `StencilData` and `StencilParameters` structs into one. 
- Adds view-variables attributes to a bunch of things
- Moves the lighting & blend-mode data from `LoadedShader` to `LoadedShaderInstance`
  - The stencil data was already stored in `LoadedShaderInstance` 
  - I have no idea if there was a reason they were stored separately before, but it meant that you **needed** to compile a separate shader program despite only wanting to toggle lighting or blending properties.
- Makes the canvas shaders use the default sprite shader.
  - Before this they were using a separate but completely identical shaders, each with their own OpenGL program.


metalgearsloth:
@PJB3005

PR #3999, opened 2023-05-02
Remove texture IoC resolves

ElectroJr:
Remove `Texture.White` references in `RenderHandle`. No more IoC resolves every time you draw a square.


moonheart08:
i hate this engine

PR #4000, opened 2023-05-03
fix wrong attribute name in autogen state errors

deltanedas:
:warning: webedit engine pr real

this has trolled me when i go autopilot and copy paste the thing the error says but its from before final version of the pr was made

it said AutoNetworkField but its called AutoNetworkedField


deltanedas:
also #4000 unintentionally sniped like a boss

mirrorcult (approved):

PR #4011, opened 2023-05-07
Add padding to glyphs in atlas

eoineoineoin:
Fixes visual artifacts when anti-aliasing glyph textures; I wanted to multi-sample the texture used by glyph rendering, but at the borders of each glyph, was picking up pixels from adjacent glyphs. Note the artifacts that appear between some of the letters:

By adding a pixel of padding in between each one, the multisampled text looks like:

### DIFF
--- a/Robust.Client/Graphics/FontManager.cs
+++ b/Robust.Client/Graphics/FontManager.cs
@@ -188,7 +188,8 @@ namespace Robust.Client.Graphics
                 info.Texture = atlasTexture;
 
                 scaled.CurSheetMaxY = Math.Max(scaled.CurSheetMaxY, scaled.CurSheetY + bitmap.Rows);
-                scaled.CurSheetX += bitmap.Width;
+                // Add a pixel of horizontal padding, to avoid artifacts when aliasing
+                scaled.CurSheetX += bitmap.Width + 1;
             }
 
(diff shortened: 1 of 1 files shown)


Robust.Client/Graphics/FontManager.cs
```diff
@@ -188,7 +188,8 @@ private GlyphInfo EnsureGlyphCached(FontInstanceHandle instance, ScaledFontData
                 info.Texture = atlasTexture;
 
                 scaled.CurSheetMaxY = Math.Max(scaled.CurSheetMaxY, scaled.CurSheetY + bitmap.Rows);
-                scaled.CurSheetX += bitmap.Width;
+                // Add a pixel of horizontal padding, to avoid artifacts when aliasing
+                scaled.CurSheetX += bitmap.Width + 1;
```
ElectroJr: I'm not familiar with the glyph/text atlas, but wouldn't you also want to add padding to the y-axis?

eoineoineoin: Ah, yeah; I was expecting that most glyphs would get that "for free" due to differences between the font baseline and min descent, but, after taking a closer look, it does show up occasionally.

For what it's worth, this is what the atlas looked like before, with a single stamp:

And with a stamp that had an artificially large amount of characters:


And with the extra padding:

PR #4013, opened 2023-05-08
[Fixture] fixed bug with null pointer execution while updating dead bodies

778b:
### Description:
If u had dead bodies which is not visible in while died, and if they will update u will had null pointer on their bodies, and in this code this didn't handled.
image
some screenshots from stack trace
image

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs
+++ b/Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs
@@ -458,6 +458,9 @@ namespace Robust.Shared.Physics.Systems
                 contact.Flags |= ContactFlags.Filter;
             }
 
+            if (fixture.Body == null)
+                return;
+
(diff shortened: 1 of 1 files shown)


778b:
Or maybe we need to add assert to show this error, cuz some system work with fixtutes.body, like FixtureSystem 
image

ElectroJr:
What was the full stack trace for this error? `Refilter()` probably shouldn't be getting called before before the entity is initialized.
And the method should just be reworked to not rely on `fixture.Body` in the first place, seeing as it is being obsoleted.

778b:
closes in #4014

ElectroJr:
> closes in https://github.com/space-wizards/RobustToolbox/pull/4014

While it should stop the exception, its still true that Refilter() probably shouldn't be getting called before before the entity is initialized, and I haven't spent any time looking into when/why this was happening

PR #4014, opened 2023-05-08
Remove obsolete `Fixture.Body` references

ElectroJr:
Removes several references to `Fixture.Body` & `Fixture.Body.Owner`. This also adds new `PhysicsComponent` and EntityUid fields to the collision events and `FixtureProxy` & `Contact` classes.

Requires space-wizards/space-station-14/pull/16259

### DIFF
--- a/Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
+++ b/Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
@@ -71,6 +71,9 @@ namespace Robust.Shared.Physics.Dynamics.Contacts
         public Fixture? FixtureA;
         public Fixture? FixtureB;
 
+        public PhysicsComponent? BodyA;
+        public PhysicsComponent? BodyB;
+
         public Manifold Manifold;
 
         internal ContactType Type;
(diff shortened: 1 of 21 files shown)


Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
```diff
@@ -315,7 +318,7 @@ public override bool Equals(object? obj)
         public override int GetHashCode()
         {
             // TODO: Need to suss this out
-            return HashCode.Combine((FixtureA != null ? FixtureA.Body.Owner : EntityUid.Invalid), (FixtureB != null ? FixtureB.Body.Owner : EntityUid.Invalid));
+            return HashCode.Combine(EntityA, EntityB);
```
metalgearsloth, line 321: This is equally scrungly as bodies can have multiple contacts but probably fine for now

metalgearsloth (approved):

Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
ElectroJr: At least its not more scrungly than it was before

PR #4016, opened 2023-05-10
Try fix collision mispredicts

ElectroJr:
This PR tries to fix the collision mispredicts. The screen should no longer stutters for a tick when entering or leaving kudzu.

This fix works by regenerating contacts & updating `contact.IsTouching` after resetting predicted entities.
Note that this means that contacts have to be re-generated an extra time per newly predicted tick, as this step is (and has to be) done before applying the next server state.

In order avoid enumerating over every entity with physics to find predicted entities, this PR adds a new `PredictedPhysicsComponent` and changes how the `PhysicsComponent.Predict` bool is set. Previously physics state handling would set the bool to false repeatedly, while `MoverController` would conditionally enable it. Now Instead of just setting the bool every tick the value is actually cached and can be queued for updating by calling `PhysicsSystem.UpateIsPredicted()`, which will cause `UpdateIsPredictedEvent` to be raised.

As a consequence, this also requires a content PR which makes some changes `MoverController` and how input/movement relays are handled.  Requires space-wizards/space-station-14/pull/16298

I haven't thoroughly tested the mover changes, but pulling & vehicle prediction still seems to work.


Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
```diff
@@ -237,6 +237,19 @@ internal ContactStatus Update(Transform bodyATransform, Transform bodyBTransform
...
+        /// </summary>
+        internal void UpdateIsTouching(Transform bodyATransform, Transform bodyBTransform)
+        {
+            var shapeA = FixtureA!.Shape;
+            var shapeB = FixtureB!.Shape;
+
+            // TODO is this sufficient? or do hard/non-sensor contacts need a more in depth check?
+            // I.e., if a non-sensor contact is "touching", is it guaranteed that the sensor overlap test always returns true?
+            IsTouching = _manifoldManager.TestOverlap(shapeA,  ChildIndexA, shapeB, ChildIndexB, bodyATransform, bodyBTransform);
+        }
```
metalgearsloth: You might've already seen it but line 154 of contact has the updates. The sensor overlaps use GJK (TestOverlap) and the manifold one uses SAT (Evaluate) so may give different results (I think GJK is more expensive? Been a while since I've seen comparisons)

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
```diff
@@ -690,4 +689,9 @@ public Box2 GetHardAABB(EntityUid uid, FixturesComponent? manager = null, Physic
 
         return (layer, mask);
     }
+
+    public virtual void UpateIsPredicted(EntityUid? uid, PhysicsComponent? physics = null)
```
metalgearsloth: Update

Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
```diff
@@ -237,6 +237,19 @@ internal ContactStatus Update(Transform bodyATransform, Transform bodyBTransform
...
+        ///     Trimmed down version of <see cref="Update"/> that only checks if two shapes overlap.
+        /// </summary>
+        internal void UpdateIsTouching(Transform bodyATransform, Transform bodyBTransform)
+        {
+            var shapeA = FixtureA!.Shape;
+            var shapeB = FixtureB!.Shape;
+
+            // TODO is this sufficient? or do hard/non-sensor contacts need a more in depth check?
+            // I.e., if a non-sensor contact is "touching", is it guaranteed that the sensor overlap test always returns true?
+            IsTouching = _manifoldManager.TestOverlap(shapeA,  ChildIndexA, shapeB, ChildIndexB, bodyATransform, bodyBTransform);
```
ElectroJr: The update method for non-sensor contacts modifies the manifold via `Evaluate(ref Manifold, ... );` and then determines IsTouching based on `Manifold.PointCount > 0`. If the current method isn't sufficient for non-sensor contacts then I'd need to know if modifying the manifold has any weird side effects for stuff like warm starting or something.

metalgearsloth: You'd likely want to copy it like the update method does as the manifold does contain warm starting data (which also does mean something like serializing and deserializing the game would cause different solver results but god forbid I ever need to start serializing contacts...)

Realistically it's probably fine and should be giving similar results.

Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
ElectroJr: Yeah I saw it and originally used that, but the fact that it actually modifies the cached "manifold" struct made me unsure about whether it might have some unintended sideffects

Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
ElectroJr: Ill have a look at it tomorrow, for now I need sleep

metalgearsloth (approved):
I double-checked the new box2d to see if the changes would be painful (as erin is adding multi-threading to contacts + island solving) but shouldn't be too hard for me to update https://github.com/erincatto/box2c/blob/e105c0c3787d485ae3796f83cef301664dd88911/src/world.c#L266

PR #4021, opened 2023-05-12
Change prof tree style & add basic stylesheet support

ElectroJr:
Makes the alternating rows visible again after the background colour was changed to black.

Content Client_JszXtN6b54

### DIFF
--- a/Robust.Client/UserInterface/DevWindow/ProfAltBackground.cs
+++ b/Robust.Client/UserInterface/DevWindow/ProfAltBackground.cs
@@ -6,10 +6,19 @@ namespace Robust.Client.UserInterface;
 internal sealed class ProfAltBackground : Control
 {
     public bool IsAltBackground { get; set; }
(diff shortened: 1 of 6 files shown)


Robust.Client/UserInterface/DevWindow/ProfAltBackground.cs
```diff
@@ -6,10 +6,19 @@ namespace Robust.Client.UserInterface;
 internal sealed class ProfAltBackground : Control
 {
     public bool IsAltBackground { get; set; }
+    public Color Color = DefaultColor;
+    public static readonly Color DefaultColor = new (255, 255, 255, 16);
```
moonheart08: Not a great default. `#222222` maybe?

ElectroJr: Updated screenshot

PR #4026, opened 2023-05-13
vv perm tweaks

metalgearsloth:
I think this doesn't functionallity change anything except if you had access to vv (e.g. via the debug perm group) then you had access to the other vv commands.

Previously all the commands checked "CanViewVar" which only checked for "vv" perm.

### DIFF
--- a/Robust.Server/ViewVariables/ServerViewVariablesManager.cs
+++ b/Robust.Server/ViewVariables/ServerViewVariablesManager.cs
@@ -112,7 +112,7 @@ namespace Robust.Server.ViewVariables
             }
 
             var player = _playerManager.GetSessionByChannel(message.MsgChannel);
-            if (!_groupController.CanViewVar(player))
+            if (!_groupController.CanCommand(player, "vv"))
             {
                 Deny(ViewVariablesResponseCode.NoAccess);
                 return;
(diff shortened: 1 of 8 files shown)


Robust.Server/ViewVariables/ServerViewVariablesManager.cs
```diff
@@ -286,16 +286,16 @@ private bool TryReinterpretValue(object? input, [NotNullWhen(true)] out object?
...
-            return _playerManager.TryGetSessionByChannel(channel, out var session) && _groupController.CanViewVar(session);
+            return _playerManager.TryGetSessionByChannel(channel, out var session) && _groupController.CanCommand(session, command);
         }
 
         protected override bool TryGetSession(Guid guid, [NotNullWhen(true)] out ICommonSession? session)
         {
             if (guid != Guid.Empty
                 && _playerManager.TryGetSessionById(new NetUserId(guid), out var player)
-                && !_groupController.CanViewVar(player)) // Can't VV other admins.
+                && !_groupController.CanCommand(player, "vv")) // Can't VV other admins.
```
metalgearsloth, line 298: This is the main thing left as is.

Robust.Server/ViewVariables/ServerViewVariablesManager.cs
```diff
@@ -286,16 +286,16 @@ private bool TryReinterpretValue(object? input, [NotNullWhen(true)] out object?
             }
         }
 
-        protected override bool CheckPermissions(INetChannel channel)
+        protected override bool CheckPermissions(INetChannel channel, string command)
         {
-            return _playerManager.TryGetSessionByChannel(channel, out var session) && _groupController.CanViewVar(session);
+            return _playerManager.TryGetSessionByChannel(channel, out var session) && _groupController.CanCommand(session, command);
```
metalgearsloth, line 291: CanViewVar only checked for vv access so this just checks the specific command instead and decouples them slightly more.

PR #4035, opened 2023-05-14
Update old methods

Chief-Engineer:
New values shouldn't be relied on until they're evaluated for uniqueness. Though some values are expected to be highly unique, frequent collisions are expected with some values at moderate device counts. These values can be made useful by combining them with other values to reach a sufficient level of uniqueness. The uniqueness of certain values may change over time, so continuous reevaluation is recommended.

S1 is known to be sufficiently unique that it alone should **almost** always be useable to confirm a match for the foreseeable future, however care should be used when attempting to use it to exclude a match. The uniqueness of S1 does not require evaluation. S1 provides reverse compatibility for values from the old system.

Please ping me before merging, I want to do a final sanity check and maybe truncate some values. Also I'm interested in de-hashing some values, specifically ones that have been truncated since, depending on the value, at that point hashing isn't very useful, but I'm not sure if that'd be acceptable with the current privacy policy, or if changes are possible to the privacy policy.

### DIFF
--- a/Robust.Shared/Network/HWId.cs
+++ b/Robust.Shared/Network/HWId.cs
@@ -2,13 +2,12 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Management;
-using System.Runtime.Serialization;
 using System.Security.Cryptography;
 using System.Text;
-using System.Text.Json.Serialization;
 // ReSharper disable UseUtf8StringLiteral
 using Microsoft.Win32;
 using Robust.Shared.Console;
+using Robust.Shared.Log;
 
 namespace Robust.Shared.Network
 {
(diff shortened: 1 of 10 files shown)


Robust.Shared/Network/HWId.cs
```diff
@@ -1,34 +1,299 @@
...
+                    if (o != null)
+                    {
+                        try
+                        {
+                            var p = Convert.ToUInt32(o);
+                            m.Add(C(p, 16));
+                        }
+                        catch (FormatException e)
+                        {
+                            System.Console.WriteLine(e);
```
metalgearsloth: Logger for this + other consoles

PR #4040, opened 2023-05-15
Make map-grids set TransformComponent.GridUid

ElectroJr:
Previously the `GridUid` would be null if an entity is on a map-grid, which causes issues for some components which use the grid uid to fetch the grid component.

This change runs afoul of some convoluted entity init ordering issues, so it also reworks how GridUid is initially set, though the overall behaviour is the same. In particular, GridUid is still not guaranteed to be set correctly until after the transform component has initialized.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AppearanceSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AppearanceSystem.cs
@@ -17,6 +17,7 @@ namespace Robust.Client.GameObjects
             base.Initialize();
 
             SubscribeLocalEvent<ClientAppearanceComponent, ComponentInit>(OnAppearanceInit);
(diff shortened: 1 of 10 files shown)


Robust.Client/GameObjects/EntitySystems/AppearanceSystem.cs
```diff
@@ -103,10 +107,9 @@ public override void FrameUpdate(float frameTime)
             var metaQuery = GetEntityQuery<MetaDataComponent>();
             while (_queuedUpdates.TryDequeue(out var appearance))
             {
-                if (appearance.Deleted)
-                    continue;
-
                 appearance.AppearanceDirty = false;
+                if (!appearance.Running)
+                    continue;
```
ElectroJr, line 112: This is technically unrelated, but I can't be bothered opening a separate PR for this change.

PR #4042, opened 2023-05-15
Use theme resolve

Skyedra:
Use the same logic as OnThemeUpdated() does so that the path used considers the current theme, instead of only supporting absolute paths.

(This appeases the YAML linter causing issue in https://github.com/space-wizards/space-station-14/pull/15840 )

### DIFF
--- a/Robust.Client/UserInterface/Controls/TextureButton.cs
+++ b/Robust.Client/UserInterface/Controls/TextureButton.cs
@@ -54,8 +54,8 @@ namespace Robust.Client.UserInterface.Controls
         {
             set
             {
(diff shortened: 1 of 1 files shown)


eoineoineoin:
This broke UIs for paper, mech equipment and the targeting doll (also, possibly others I missed?):






It also changes the semantics of TextureButton.TexturePath; that field is now no longer a path, but a weird-subsection-of-a-path-which-limits-where-textures-can-be-located, which feels _real_ weird to me.

Skyedra:
Thanks for the heads up, please lemme know what you think of this other method in that thread when you can:  https://github.com/space-wizards/RobustToolbox/pull/4048

PR #4048, opened 2023-05-16
Split texture theme path and regular texture path

Skyedra:
Based on [the other thread](https://github.com/space-wizards/RobustToolbox/pull/4042#issuecomment-1549211753), it looks like not all locations for texturepath can depend on texture theme path resolver.  Also looks like there was some code kinda managing this before, but in a way that didn't save whether the path was a theme-relative or root-relative path.

This PR splits the two different paths into two different variables, instead of trying to overload one variable in a way that the theme update event can't reliably differentiate between.

Also fixes broken images on paper, etc

:warning:  Requires https://github.com/space-wizards/space-station-14/pull/16498 on the content side to prevent action buttons becoming errors

@eoineoineoin

image

### DIFF
--- a/Robust.Client/UserInterface/Controls/TextureButton.cs
+++ b/Robust.Client/UserInterface/Controls/TextureButton.cs
@@ -17,7 +17,16 @@ namespace Robust.Client.UserInterface.Controls
         public const string StylePseudoClassHover = "hover";
         public const string StylePseudoClassDisabled = "disabled";
         public const string StylePseudoClassPressed = "pressed";
+
+        /// <summary>
+        /// Path specified from root of resources.
+        /// </summary>
         private string? _texturePath;
+
+        /// <summary>
+        /// Path specified relative to current theme.
+        /// </summary>
+        private string? _textureThemePath;
 
 
         public TextureButton()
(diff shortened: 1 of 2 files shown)


ElectroJr:
I think a better fix would be to just make the theme resolve only try to interpret relative paths, while having having rooted paths and failed resolves try to use it as an absolute path.
I.e., replacing

```cs
public Texture ResolveTexture(IResourceCache cache, string texturePath)
{
    return cache.TryGetResource<TextureResource>( new ResPath($"{Path}/{texturePath}.png"), out var texture) ? texture :
        cache.GetResource<TextureResource>($"{DefaultPath}/{DefaultName}/{texturePath}.png");
}
```
with
```cs
public static ResPath DefaultThemePath = new ResPath($"{DefaultPath}/{DefaultName}");
public Texture ResolveTexture(IResourceCache cache, string texturePath)
{
    var texPath = new ResPath($"{texturePath}.png");
    if (texPath.IsRelative)
    {
        if (cache.TryGetResource( Path / texPath, out TextureResource? texture))
            return texture;

        if (cache.TryGetResource( DefaultThemePath / texPath, out texture))
            return texture;
    }

    return cache.GetResource<TextureResource>(texPath);
}
```

and then just removing the `TextureThemePath` setter altogether. Note that I haven't actually tested this, and this might also required some content changes.

ElectroJr:
Superseded by #4059

PR #4053, opened 2023-05-16
Remove theme sub

metalgearsloth:
I have a suspicion it's what's causing contextmenus to cause multi-second stutters over time though didn't get confirmation but testmerge for a day wouldn't hurt?

### DIFF
--- a/Robust.Client/UserInterface/UserInterfaceManager.Themes.cs
+++ b/Robust.Client/UserInterface/UserInterfaceManager.Themes.cs
@@ -8,7 +8,6 @@ namespace Robust.Client.UserInterface;
 
 internal partial class UserInterfaceManager
 {
-    public event Action? ThemeUpdated;
     private readonly Dictionary<string, UITheme> _themes = new();
 
     public UITheme CurrentTheme { get; private set; } = default!;
@@ -57,8 +56,14 @@ internal partial class UserInterfaceManager
     {
         if (newTheme == CurrentTheme) return; //do not update if the theme is unchanged
         CurrentTheme = newTheme;
-        if (ThemeUpdated != null)
-            ThemeUpdated();
+
+        foreach (var root in AllRoots)
+        {
+            foreach (var control in root.Children)
+            {
+                control.OnThemeUpdated();
+            }
+        }
     }
 
     //Try to set the current theme, if the theme is not found leave the previous theme
--- a/Robust.Client/UserInterface/Control.cs
+++ b/Robust.Client/UserInterface/Control.cs
@@ -75,7 +75,7 @@ namespace Robust.Client.UserInterface
 
         public UITheme Theme { get; set; }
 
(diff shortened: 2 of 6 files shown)


Robust.Client/UserInterface/UserInterfaceManager.Themes.cs
```diff
@@ -57,8 +56,14 @@ private void UpdateTheme(UITheme newTheme)
...
-        if (ThemeUpdated != null)
-            ThemeUpdated();
+
+        foreach (var root in AllRoots)
+        {
+            foreach (var control in root.Children)
+            {
+                control.OnThemeUpdated();
+            }
+        }
```
metalgearsloth, line 66: Technically this doesn't cover off stuff not under any root but realistically UIControllers should probably be handling theme stuff? Not sure. 90% of controls don't care about theme.

ElectroJr: At minimum, this would need the the recursive update that was removed in #3973 to be re-added.

As to the ui controller comment, because not all controls that care about themes are managed by UI controllers, you would still have to update all controls. Having ui controller separately also update all of their managed controls might fix the issue with orphaned controls not getting updated, but that sounds tedious to implement for each controller, and will probably introduce bugs because it seems like the kind of thing that is easy to overlook.

~~Are we even sure that the theme update subscription is the issue? AFAICT it should always be unsubscribing.~~

ElectroJr (changes requested):

Robust.Client/UserInterface/UserInterfaceManager.Themes.cs
ElectroJr: Maybe just add all orphaned controls to some orphan collection?

ElectroJr: > Are we even sure that the theme update subscription is the issue? AFAICT it should always be unsubscribing.

It does seem to have made the context menu stutter significantly worse, though there was still a slight stutter when examining many entities before. It also makes closing sub-menus like the verb menu stutter, which didn't use to happen.

Is the actual issue just that unsubbing from c# events with many subscriptions is very slow? Given that the order of invocations is fixed by the order in which subscriptions happened, I guess its akin to removing an element somewhere in the middle of a list? So when you have **many** controls and then dispose a bunch all at once it is very slow?

ElectroJr: > AFAICT it should always be unsubscribing.

Nevermind. It only unsubs when explicitly disposed and the subscription counts as a reference to the control so they never get collected by GC.

ElectroJr: I'm just going to revert the PR and close this one.

PR #4058, opened 2023-05-16
Update themes when adding root controls.

ElectroJr:
Replacement for #3973. Alternatively, maybe this should just become part of the base AddChild method (see #4059).


ElectroJr:
After thinking about it a bit more, #4059 is probably a better fix.

PR #4064, opened 2023-05-17
Add locale support for grammatical measure words

Vordenburg:
Information on what measure words are: https://en.wikipedia.org/wiki/Measure_word

**Before:**
image

**After:**
image
image
image


Robust.Shared/Localization/LocalizationManager.Functions.cs
```diff
@@ -227,6 +228,27 @@ private ILocValue FuncReflexive(LocArgs args)
...
+            {
+                EntityUid entity = (EntityUid)entity0.Value;
+
+                if (TryGetEntityLocAttrib(entity, "counter", out var counter))
+                {
+                    return new LocValueString(counter);
+                }
+            }
+
+            return new LocValueString(GetString("zzzz-counter-default"));
```
moonheart08, line 249: You never actually define this locale string in engine for en-US.

Vordenburg: It's not used in English. Will that be an issue?

PR #4065, opened 2023-05-17
Change default state reset behaviour

ElectroJr:
Full state resets now only delete client side entities if they are parented to a networked entity that is getting deleted. Should avoid issues where client-side nullspace entities that are being used for things like UIs unintentionally get deleted.

Breaking change because the `PartialStateReset()` arguments have changed:
- Changed the `deleteClientSideEnts` bool  argument to an optional argument that defaults to false.
- Added a new argument that controls whether or not to delete client-side entities parented to networked entities (defaults to true).


moonheart08:
@ElectroJr update engine whenever

PR #4071, opened 2023-05-19
Add `IContentRoot.FileExists()`

ElectroJr:
Generally faster than having to use `TryGetFile(path, out _)`

### DIFF
--- a/Robust.Shared/ContentPack/DirLoader.cs
+++ b/Robust.Shared/ContentPack/DirLoader.cs
@@ -58,6 +58,12 @@ namespace Robust.Shared.ContentPack
                 return true;
             }
 
+            public bool FileExists(ResPath relPath)
+            {
+                var path = GetPath(relPath);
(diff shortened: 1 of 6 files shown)


Robust.Shared/ContentPack/DirLoader.cs
```diff
@@ -58,6 +58,12 @@ public bool TryGetFile(ResPath relPath, [NotNullWhen(true)] out Stream? stream)
                 return true;
             }
 
+            public bool FileExists(ResPath relPath)
+            {
+                var path = GetPath(relPath);
+                return File.Exists(path);
+            }
```
metalgearsloth, line 65: Is this using dotnet File.Exists? If so might be an issue as content could read any file on their PC?

metalgearsloth: Ahh okay I just read it in my IDE and looks like GetPath sanitises.

PR #4079, opened 2023-05-21
Ecs's UserInterfaceComponent.

TemporalOroboros:
Makes UserInterfaceComponent ECS

Requires https://github.com/space-wizards/space-station-14/pull/17736


Robust.Server/GameObjects/Components/UserInterface/ServerUserInterfaceComponent.cs
```diff
@@ -15,23 +15,18 @@ namespace Robust.Server.GameObjects
...
+        [Access(typeof(UserInterfaceSystem))]
+        [ViewVariables]
+        public readonly Dictionary<Enum, BoundUserInterface> AllInterfaces = new();
 
-            foreach (var prototypeData in _interfaceData)
-            {
-                _interfaces[prototypeData.UiKey] = new BoundUserInterface(prototypeData, this);
-            }
-        }
+        public IReadOnlyDictionary<Enum, BoundUserInterface> Interfaces => AllInterfaces;
```
ElectroJr: Just remove `Interfaces`  and rename `AllInterfaces` to `Interfaces`

Robust.Client/GameObjects/Components/UserInterface/ClientUserInterfaceComponent.cs
```diff
@@ -1,122 +1,62 @@
...
+        public readonly Dictionary<Enum, BoundUserInterface> OpenInterfaces = new();
 
-        internal void SendMessage(BoundUserInterfaceMessage message, Enum uiKey)
-        {
-            _netMan.SendSystemNetworkMessage(new BoundUIWrapMessage(Owner, message, uiKey));
-        }
+        /// <summary>
+        /// An enumerator for <see cref="OpenInterfaces"/>
+        /// </summary>
+        public IEnumerable<BoundUserInterface> Interfaces => OpenInterfaces.Values;
```
ElectroJr: Id say just remove this. `OpenInterfaces` is already public and this will just cause confusion if the server-side `Dictionary<Enum, BoundUserInterface> Interfaces` is ever moved to shared code.

Robust.Shared/GameObjects/Components/UserInterface/SharedUserInterfaceComponent.cs
```diff
@@ -11,7 +11,7 @@ namespace Robust.Shared.GameObjects
     public abstract class SharedUserInterfaceComponent : Component
     {
         [DataField("interfaces")]
-        internal List<PrototypeData> _interfaceData = new();
```
ElectroJr: Why does this need to be public? Eventually this should probably just be removed, so it'd be easier if content didn't have references to it

TemporalOroboros:
Needs a few content modules to be ECSed before it's merged or it could break some things.

PR #4084, opened 2023-05-27
Add `IWritableDirProvider.OpenOsWindow()`

ElectroJr:
Adds a function that opens an external file browser for folder, so that you can do things like add a button that opens the screenshots or a save file directory. I'm not sure if `IWritableDirProvider` is the best place to put this, but I couldn't think of anything better.

### DIFF
--- a/Robust.Shared/ContentPack/VirtualWritableDirProvider.cs
+++ b/Robust.Shared/ContentPack/VirtualWritableDirProvider.cs
@@ -209,6 +209,11 @@ namespace Robust.Shared.ContentPack
             throw new NotImplementedException();
         }
 
+        public void OpenOsWindow(ResPath path)
+        {
+            throw new NotImplementedException();
+        }
+
         private bool TryGetNodeAt(ResPath path, [NotNullWhen(true)] out INode? node)
         {
             if (!path.IsRooted)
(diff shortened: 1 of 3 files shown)


Robust.Shared/ContentPack/WritableDirProvider.cs
```diff
@@ -130,6 +131,19 @@ public void Rename(ResPath oldPath, ResPath newPath)
...
+        {
+            if (!IsDir(path))
+                path = path.Directory;
+
+            var fullPath = GetFullPath(path);
+            Process.Start(new ProcessStartInfo
+            {
+                UseShellExecute = true,
+                FileName = fullPath,
+            });
```
ElectroJr, line 144: I just pilfered this code snippet from the launcher's "open log directory" button. works fine on windows and I assume it'll also work on mac & linux.

Robust.Shared/ContentPack/VirtualWritableDirProvider.cs
```diff
@@ -209,6 +209,11 @@ public void Rename(ResPath oldPath, ResPath newPath)
             throw new NotImplementedException();
         }
 
+        public void OpenOsWindow(ResPath path)
+        {
+            throw new NotImplementedException();
```
PJB3005: The act of opening the file explorer should be invisible to the rest of the game, so it'd make more sense to simply do nothing here.

PJB3005 (changes requested):

PJB3005 (approved):

PR #4085, opened 2023-05-27
Reduced damage option and multiplayer CVars

McFck:
Moved reduced damage and reduce multiplier as CVar. Made cables do damage on destroy. Fixed cables doing 0 damage.
Part of https://github.com/space-wizards/space-station-14/pull/16816


mirrorcult:
these should not be in engine just put them in content cvars

PR #4092, opened 2023-05-29
Replace Robust Vector2 with System.Numerics

metalgearsloth:
Perfectly good standard library right there.

Also includes sandboxing change for System.Numerics.Vector2 and also needs a NetSerializer PR for Vector2 maybe (or I guess I make it a dynamic type instead of using primitives?)

Also I inadvertantly broke the collision solver (flipped sign or some shit idk) so need to fix that, rest seems okay.


Robust.Shared.Maths/Vector2Helpers.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        var length = vec.Length();
+        return new Vector2(vec.X / length, vec.Y / length);
+    }
+
+    /// <summary>
+    /// Normalizes this vector if its length > 0, otherwise sets it to 0.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static float Normalize(this Vector2 vec)
```
ElectroJr:
```suggestion
    public static float Normalize(ref this Vector2 vec)
```

ElectroJr (changes requested):

Robust.Shared.Maths/Vector2Helpers.cs
ElectroJr: also the description should probably be changed as it doesn't "set it to 0" but just returns zero. Unless it is actually meant to set it to zero, in which case I guess that also a bug in the old method.

PR #4105, opened 2023-06-02
Fixed some coding mistakes

Redict:
- `XamlCompiler.cs`. In case this if is true `parameterlessCtor` will be always used
- `SharedPhysicsSystem.Island.cs`. Was checking `jointCompB` for `uidA`
- `SpriteComponent.cs`. Was checking only `X` in setter
- Others. `??` operator priority is lower than it's left part


PJB3005:
These changes all look good (oof) but the XamlIL change is causing invalid IL to be generated on content. I'll take a look.

PJB3005:
The code generated by the Xaml injector change here doesn't seem like it's ever actually getting used. Probably also explains why this obvious bug didn't cause any trouble. I'll just remove it entirely.

PJB3005 (approved):

PR #4106, opened 2023-06-03
Feat/exposure nightvision

tom-leys:
- Add Exposure to Eye which scales all lighting range and intensity. 
- Each frame, lighting brightness is measured in a small square near the player, this sampled brightness is written back into Eye for client code to use to adjust exposure
- Point lights easier to add in the lighting pass, since they rely on a simpler class than PointLightComponent
- Light measuring code might have a performance impact since we're reading a RenderTexture from the CPU. It is not noticable on my system in debug builds.
- Store texture format data in Clyde Textures
- GetPixel now works on float textures
- Engine changes alone don't add HDR exposure changes. You'll see no changes until SS14 client changes are also merged.
- Light limit pass - a single fullscreen pass compresses lights above 1.0 so extreme lighting conditions look much less bright.
- Overbrighten in standard shader - Bright lights will whiten sprites (looks far less garish and more correct)
- Extremely bright or stacked lights will always look better, as they will both be capped and will also whiten correctly.

---

Demos with Client code changes also
https://www.youtube.com/watch?v=8JrGEDx1S6Q

Powerful ghost night vision
image

Human dynamic exposure changes
https://github.com/space-wizards/RobustToolbox/assets/5285589/91cd7fde-30d5-4dac-90dd-795614cb04ef

I've also added good looking overexposure and light compression, so overbright regions look nicer
image

7 flare pointlights in the same place
image

Ghosts are much less sensitive to overbrightness, so you can see in the dark without blowing out regions of your screen
image


Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -544,9 +556,19 @@ private void DrawLightsAndFov(Viewport viewport, Box2Rotated worldBounds, Box2 w
             if (count >= state.clyde._maxLights)
                 return false;
 
-            var (light, transform) = value;
+            var (light_comp, transform) = value;
+
+            var powerMul = Math.Min(state.exposure, 1.2f + state.exposure * 0.05f);
+            var rangeMul = state.exposure * 0.2f + 0.8f;
```
ElectroJr: if you want exposure to modify the range you also need to update the lights bounding boxes to account for tha. I.e., put an upper bound on exposure and use that for the bounds, otherwise you'd need to constantly update the bounds of all lights as exposure changes (+ having a different tree per eye). Alternatively, just don't scale the range unless its actually necessary.

tom-leys: 1. I just expand the AABB that I pass into the tree as a query, which is much cheaper but equilivent to increasing the range of all lights in the tree. I over-expand the AABB and then rely on the second range filter here.
2. Note this is adjusting the range of a COPY of the light data, it's not modifying the PointLightComponent itself.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -567,18 +589,24 @@ private void DrawLightsAndFov(Viewport viewport, Box2Rotated worldBounds, Box2 w
...
-            var state = (this, count: 0, shadowCastingCount: 0, xformSystem, xforms, worldAABB);
+            var state = (this, count: 0, shadowCastingCount: 0, xformSystem, xforms, worldAABB, exposure);
 
             foreach (var comp in lightTreeSys.GetIntersectingTrees(map, worldAABB))
             {
                 var bounds = xformSystem.GetInvWorldMatrix(comp.Owner, xforms).TransformBox(worldBounds);
+                if (exposure > 1.5f)
+                {
+                    // Need to query a larger area because at higher exposure the radius of lights is increased.
+                    bounds = bounds.Enlarged(5.0f * exposure);
```
tom-leys: @ElectroJr Here's my conservative code that increases the AABB size quite a lot (probably too much now actually) to include more lights as exposure increases.

Expanding AABB size by X is like increasing the radius of all lights in the tree by X.

moonheart08:
- NightVision has zero business being in engine, allow content to supply non-entity lights instead
- AutoExposure also has no business being in engine, allow content to preprocess the to-be-rendered-lights list instead.
- Scaling non shadow casting lights is no bueno.

tom-leys:
> * NightVision has zero business being in engine, allow content to supply non-entity lights instead

I think you're right on this. "NightVision" here is very game specific, though it's kind of a scaling vignette?

Perhaps, though I'm unsure how to accomplish that. Some sort of callback on LightManager or Eye?

>     * AutoExposure also has no business being in engine, allow content to preprocess the to-be-rendered-lights list instead.

Being able to scale the power of the lighting system is a powerful feature. The mathematically best way to do that is to scale the lights (scaling the buffer directly loses resolution).

If you increase a light's power, you need to increase its range by sqrt(increase) because light diminishes to the sqrt of the power.

Having a callback that processes the lights seems complex and expensive, since any changes it makes can affect the filtering. It could be done, but I think it would be nicer to keep exposure specifically in-engine.

> 
>     * Scaling non shadow casting lights is no beuno.

They will look unnatural if we don't scale them but increase their power. I guess you're concerned that they will bleed through walls? Perhaps if the number of shadow-casting lights is low, some non-shadow casting lights gain shadows? 

Personally if I was a content author I'd want a powerful lighting system in the engine, but I'd appreciate the choice about how the exposure changed.

tom-leys:
I've moved AutoExposure and NightVision into client code.

Eye.Exposure is still in, I argue that it has good business being in the lighting logic. If you really really hate dynamic exposure I guess we could do a post process clientside, moving it seems ugly to me.

I'm not sure what to do with local clientside lighting, probably a list of lights in screenspace or worldspace in Eye? In the meantime I commented out the additional light.

Robust.Client/GameObjects/Components/Light/PointLightComponent.cs
```diff
@@ -7,9 +7,62 @@
 using Robust.Shared.Physics;
 using Robust.Shared.Serialization.Manager.Attributes;
 using Robust.Shared.ViewVariables;
+using TerraFX.Interop.Windows;
 
 namespace Robust.Client.GameObjects
 {
+    public struct PointLight
```
PJB3005: This struct is only used by Clyde. It should be in `Clyde.LightRendering` and be private.

Robust.Client/GameObjects/Components/Light/PointLightComponent.cs
```diff
@@ -7,9 +7,62 @@
...
+            Softness = light.Softness;
+        }
+
+        public PointLight(float energy, float radius, bool castShadows)
+        {
+            Energy = energy;
+            Radius = radius;
+            CastShadows = castShadows;
+            Color = Color.White;
+        }
```
PJB3005: These aren't used.

Robust.Client/Graphics/ClientEye/IEye.cs
```diff
@@ -14,7 +17,7 @@ public interface IEye
         /// <summary>
         /// Should the black FoV effect be drawn for this eye?
         /// </summary>
-        bool DrawFov { get; set; }
+        public bool DrawFov { get; set; }
```
PJB3005: Interface elements are by default `public`, you don't need to change this.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -544,41 +567,64 @@ private void DrawLightsAndFov(Viewport viewport, Box2Rotated worldBounds, Box2 w
...
+            // Increase light proportional to exposure.
+            var powerMul = state.exposure; // Math.Min(state.exposure, 1.2f + state.exposure * 0.05f);
+            var rangeMul = (float)Math.Sqrt(powerMul);
+
+            // Create a simplified light from the component then scale power & range using the exposure.
+            ref var light = ref state.clyde._lightsToRenderList[count++];
+            light.UpdateFrom(light_comp);
+
+            // Lights that have a radius beyond a certain limit create popping artifacts. Cap the max range to help prevent those here.
+            light.Radius = Math.Min(20.0f, light.Radius * rangeMul);
```
PJB3005: Make this a constant so it stands out more.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -544,41 +567,64 @@ private void DrawLightsAndFov(Viewport viewport, Box2Rotated worldBounds, Box2 w
...
-            var state = (this, count: 0, shadowCastingCount: 0, xformSystem, xforms, worldAABB);
+            var state = (this, count: 0, shadowCastingCount: 0, xformSystem, xforms, worldAABB, exposure);
 
             foreach (var comp in lightTreeSys.GetIntersectingTrees(map, worldAABB))
             {
                 var bounds = xformSystem.GetInvWorldMatrix(comp.Owner, xforms).TransformBox(worldBounds);
+                if (exposure > 1.5f)
+                {
+                    // Need to query a larger area because at higher exposure the radius of lights is increased.
+                    bounds = bounds.Enlarged(5.0f * exposure);
```
PJB3005: Pull these constants out too so they stand out better.

Robust.Client/Graphics/Clyde/Clyde.Textures.cs
```diff
@@ -653,17 +658,89 @@ public override Color GetPixel(int x, int y)
...
+                        {
+
+                            GL.GetTextureSubImage(loaded.OpenGLObject.Handle, 0, x, y, 0, 1, 1, 1, Format.pixFormat, Format.pixType, 4, (IntPtr) p);
+                        }
+                    }
+                    return new Color(rgba[0], rgba[1], rgba[2], rgba[3]);
+                }
+            }
+
+            public Color MeasureBrightness(int x, int y, int width, int height)
```
PJB3005, line 693: This implementation is not suitable and will ruin the game's performance. You're effectively having the CPU do nothing while the GPU finishes rendering, rather than allowing them to run properly asynchronously.

You're going to need to replicate what the game does for screenshots: copy the pixels into a Pixel Buffer Object, fence it, and then map the buffer on a later frame when the copy is done. This inherently makes the operation asynchronous and yes this means you'll have to accept that the results will come in on a bit of a delay.

(If you're wondering about `GetPixel()`... the code for that one sucks and I didn't review it at the time. It's not used by SS14 IIRC, only OpenDream)

PJB3005: Also move this code into a function on the `Clyde` class itself rather than on `ClydeTexture`.

PJB3005 (changes requested):
You need to update the sort calls on line ~640 of `Clyde.LightRendering.cs`, you missed them because `Array.Sort()` sucks.

Robust.Client/GameObjects/Components/Light/PointLightComponent.cs
tom-leys: Yeah, it was a constructor added for the nightvision lighting. We need a way to inject game-specific clientside point lights into the renderer (not in this PR yet)

Specifically I want to add a pointlight in the middle of the viewport to represent nightvision of the player.

Robust.Client/Graphics/Clyde/Clyde.Textures.cs
tom-leys: Thanks for the pointer on where to look for an example. Async is indeed ideal - I was also concerned about the performance impact and cpu wait nature of this operation but how to fix it was beyond me.

tom-leys:
This is a draft until I've resolved issues raised by the recent review.

tom-leys:
I've now moved the light exposure code into an async process in `Clyde.LightingSample.cs`

metalgearsloth:
I only skimmed it but wouldn't you async call it? Looks like the methods are still unsafe / void.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
metalgearsloth: Does this even work with PVS? The range is still just going to be 12.5

Robust.Client/Graphics/Clyde/Shaders/base-default.frag
```diff
@@ -16,5 +16,14 @@ void main()
...
+    // Whiten everything (even black source pixels) as lighting becomes too intense.
+    vec3 overBrightC = max(vec3(0.0), lightSample - vec3(1.3));
+    // Calculate a bright to white effect for strong lights
+    float overBright = sqrt((overBrightC.r + overBrightC.g + overBrightC.b) * 0.33);
+    vec3 bloomColor = min(vec3(1.0), vec3(0.002) * overBright + lightSample * overBright * 0.01);
+    // Add the white component to a slight colored component.
+    vec4 bloom = vec4(vec3(bloomColor), 0.0);
+    // bloom = vec4(1.0);
+
+    gl_FragColor = zAdjustResult(COLOR * VtxModulate * vec4(lightSample, 1.0) + bloom);
```
deathride58, line 28: It'd probably be far saner to do the tonemapping here in a postprocess pass (whether it's an overlay in content or something more sophisticated in engine), rather than having the tonemapping be a part of the base default fragment shader. While this is fine for most shaders currently in SS14, this will have a pretty significant impact on artistic flexibility in the future, particularly when it comes to object shaders that need to sample the underlying image, but have good reason to not be using raw.

Generally, it's ideal for HDR conversion to sRGB to be handled as late into a pipeline as feasible, as this provides the finest level of artistic control, among other positives.

tom-leys: While this is true, I wasn't comfortable changing the RGB render buffers to be HDR or adding a post-process pass to this. 

The key here is if the sprite shows a completely black pixel but we overbrighten it, does it remain black? I've pushed it to white regardless if there is enough lighting. Otherwise we get very ugly artifacts.

tom-leys: Please bear in mind that I don't feel comfortable changing the whole renderer. I also doubt that my PR would ever be merged if I DID substantially change the renderer more than I have already.

A small hack is more likely to get merged, could always refactor it and make it more flexible in a later PR.

metalgearsloth: Talked with bhijn and this is still outstanding.

Stokes52:
This is really amazing, and I hope that this makes it into the game. Excellent, excellent work!

ElectroJr:
I'm starting to get back into looking at light rendering stuff, mostly to try increase performance. While my motivation lasts I'll also try get around to looking at this PR again. But in the meantime, it might be best to split off the `struct PointLight` related changes into a separate PR, which would also help reduce conflicts.

dragonryan06:
I don't know much about the engine but from a more player-facing side, this PR would be amazing as it would allow for different creatures to have different eye settings, I did a lot of digging and as-is there isn't really a way for SS14 rats to be able to see in dark maintenance tunnels without allowing them to hold flashlights or disabling the lighting engine entirely, which is quite unfortunate, but this PR would look even cooler than the hacky stuff I tried!

I also just think the fake-hdr effect it gives off is pretty neat, though there would need to be a cap on how bright the bloom can get because those flares are a little out of control

ShatteredSwords:
Had a friend say this PR could make it so Diona could feed off light. That's a really cool change.

tom-leys:
> Had a friend say this PR could make it so Diona could feed off light. That's a really cool change.

Since the lighting calculations are currently clientside this actually wouldn't be compatible with non-visual gameplay changes. It would be possible (but hard) to do enough calculations on the server to add the functionality you want - just not with this PR.

Robust.Client/Graphics/Clyde/Shaders/base-default.frag
metalgearsloth: Having looked at this I'm also not really sure why you can't just use an overlay to adjust the rendertarget for tonemapping instead of baking it into the base shader?

K-Dynamic:
Are you still working on this

tom-leys:
No, and I have no plans to work on it in the future. I should probably turn off notifications here now.

metalgearsloth:
> Had a friend say this PR could make it so Diona could feed off light. That's a really cool change.

That would be https://github.com/space-wizards/RobustToolbox/pull/5574

PR #4110, opened 2023-06-04
Obsolete some metadata stuff

metalgearsloth:
I keep seeing it get used on content.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/MetaDataSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/MetaDataSystem.cs
@@ -39,6 +39,33 @@ public abstract class MetaDataSystem : EntitySystem
         component.PauseTime = state.PauseTime;
     }
 
+    public void SetEntityName(EntityUid uid, string value, MetaDataComponent? metadata = null)
+    {
+        if (!Resolve(uid, ref metadata) || value.Equals(metadata.EntityName))
+            return;
+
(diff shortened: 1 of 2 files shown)


Robust.Shared/GameObjects/Systems/MetaDataSystem.cs
```diff
@@ -39,6 +39,33 @@ private void OnMetaDataHandle(EntityUid uid, MetaDataComponent component, ref Co
...
+    }
+
+    public void SetEntityPrototype(EntityUid uid, EntityPrototype? value, MetaDataComponent? metadata = null)
+    {
+        if (!Resolve(uid, ref metadata) || value?.Equals(metadata._entityPrototype) == true)
+            return;
+
+        metadata._entityPrototype = value;
+        Dirty(metadata);
+    }
```
ElectroJr: This function should throw an error if the new prototype ID differs from the old one, because otherwise it will probably cause game state / networking errors if ever used in a multiplayer game. + maybe just make it internal?

ElectroJr (changes requested):

Robust.Shared/GameObjects/Systems/MetaDataSystem.cs
metalgearsloth: I'll just make it internal and not dirtyable probably.

PR #4122, opened 2023-06-06
Add support for client-side replays

ElectroJr:
This adds support for client-side replay recordings and playback. While a recording is active the client will write data just before it applies the server's game state. Requires space-wizards/space-station-14/pull/17168.

## Media

Here's a basic video showing client side replays working with sounds & pop-ups. There is a bug in the video, where the appearance of sub-floor wires doesn't seem to properly update, but that has since been fixed.

https://github.com/space-wizards/RobustToolbox/assets/60421075/8dfaae66-884f-4a09-9311-c7fc17007651

## Description of changes
- The majority of `ReplayRecordingManager` has been moved to shared code. The large diff is misleading, quite a few functions were just copied over unchanged.
- Replay playback can now handle with PVS-detached / out-of-range entities.
- Various private methods in `ClientGameStateManager` were exposed so that replay playback can use them.
- The client-side audio system has been changed to stop some sound effects from being recorded twice.
- `IWritableDirProvider` has new methods for async writing.
- Replays now write data via async tasks (see `SharedReplayRecordingManager.Write.cs`).
- A portion of `PrototypeLoadManager` code has been moved to shared.
- Some Cvars were added or changed.
- The `replay_recording_start` command has had the order of the arguments changed


Robust.Shared/ContentPack/WritableDirProvider.cs
```diff
@@ -170,5 +170,24 @@ private static string GetFullPath(string root, ResPath path)
...
+            var slice = new ReadOnlyMemory<byte>(bytes, offset, length);
+            return WriteBytesAsync(path, slice, cancellationToken);
+        }
+
+        public Task WriteBytesAsync(ResPath patch, ReadOnlyMemory<byte> bytes, CancellationToken cancellationToken = default)
+        {
+            var fullPath = GetFullPath(patch);
+            using var fs = new FileStream(fullPath, FileMode.Create, FileAccess.Write, FileShare.None, 4096, true);
+            return fs.WriteAsync(bytes, cancellationToken).AsTask();
+        }
```
ElectroJr: I'm not all that confident with async & file IO, but I think this asynchronously opens a file for writing and writes to it?

metalgearsloth: Should the method be async? If it has async in the name and returns a task I'd assume it's awaitable...

ElectroJr: I'm not sure about this, I'm not at all confident about how async stuff actually works, but I think the async postfix is used both for awaitable functions and functions that return tasks. E.g., `FileStream.WriteAsync()` isn't actually marked as async, it just returns a task.

I guess this question boils down to: are these two functionally equivalent, or is one wrong? if they are equivalent is one preferred for some reason? They both appear to work.

```cs
public void WriteStuff(ResPath path, ReadOnlyMemory<byte> bytes)
{
    Task.Run(() => OptionA(path, bytes));
    Task.Run(() => OptionB(path, bytes));
}

public async void OptionA(ResPath path, ReadOnlyMemory<byte> bytes)
{
    var fullPath = GetFullPath(path);
    await using var fs = new FileStream(fullPath, FileMode.Create, FileAccess.Write, FileShare.None, 4096, true);
    await fs.WriteAsync(bytes);
}

public ValueTask OptionB(ResPath patch, ReadOnlyMemory<byte> bytes)
{
    var fullPath = GetFullPath(patch);
    using var fs = new FileStream(fullPath, FileMode.Create, FileAccess.Write, FileShare.None, 4096, true);
    return fs.WriteAsync(bytes);
}
```

Option B is what I currently do, except for replacing ValueTask with task, though I'm also not sure what the difference between those is.

ElectroJr: @PJB3005

ElectroJr: async stuff scrambles my brain, but now I think "optionA" is how I'm supposed to do it, so I guess I'll change it,

ElectroJr: Welp I tested it by just having it write a ~100 MB file and it seems to be properly async?

PR #4125, opened 2023-06-08
Physics stuff

metalgearsloth:
Ended up adding includeMap just to make it easier to filter out maps from grid queries.


Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs
```diff
@@ -482,8 +488,8 @@ internal void CollideContacts()
 
                     var fixtureA = contact.FixtureA!;
                     var fixtureB = contact.FixtureB!;
-                    var bodyA = fixtureA.Body;
-                    var bodyB = fixtureB.Body;
```
ElectroJr, line 486: Why was this replaced with a `GetComponent(contact.EntityA)` here and elsewhere?

PR #4126, opened 2023-06-08
Don't raise contact events for qdel ents

metalgearsloth:
This may or may not be desireable behaviour for content but we'll just plug this in here for now unless we need to change it.

This doesn't affect entities qdeled inside of the contact events but will handle anything done prior to the physics step.


Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs
```diff
@@ -377,6 +377,13 @@ internal void CollideContacts()
                 continue;
             }
 
+            // Pending deletion from earlier in the tick so stop raising events any new contact events on it.
```
metalgearsloth:
```suggestion
            // Pending deletion from earlier in the tick so stop raising any new contact events on it.
```

metalgearsloth:
Might actually raise an event on qdel sub to that instead.

ElectroJr:
Does this cause issues with things like kudzu or glue that cause slowdown? They currently rely on collision-stop events being raised when an entity is deleted.

metalgearsloth:
> Does this cause issues with things like kudzu or glue that cause slowdown? They currently rely on collision-stop events being raised when an entity is deleted.

Entity deletion raises DestroyContacts on all of them when the deletion happens but I will double-check they work.

metalgearsloth:
Raising EndCollideEvent inside of potential StartCollideEvent seems okay, at least based on current content usages.

PR #4127, opened 2023-06-09
Deprecate IPlayerSession methods to attach/detach entities to a player.

gradientvera:
This has very few usages across content and engine already so might as well obsolete it before it gets used more.
Use ActorSystem methods instead.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -35,7 +35,7 @@ END TEMPLATE-->
 
 ### Breaking changes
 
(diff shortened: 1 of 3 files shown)


RELEASE-NOTES.md
```diff
@@ -35,7 +35,7 @@ END TEMPLATE-->
 
 ### Breaking changes
 
-*None yet*
+* Obsoleted the following methods from IPlayerSession: AttachToEntity, DetachFromEntity. Use the methods in ActorSystem instead.
```
PJB3005:
```suggestion
* Obsoleted the following methods from `IPlayerSession`: `AttachToEntity`, `DetachFromEntity`. Use the methods in `ActorSystem` instead.
```

PR #4137, opened 2023-06-10
Automatically create a sawmill for each entity system.

ElectroJr:
Entity systems now automatically create a Sawmill with a name based on the class name. E.g., `FixtureSystem` becomes `ILogMan.GetSawmill("system.fixture")`.  Originally I also added proxy methods (e.g. `LogError() -> Sawmill.Error())`), but decided to just name the sawmill `Log` instead, so that systems can just use `Log.Error()`.

The samill names can be overridden by systems, The automatic sawmill names don't currently properly convert to snake_case. E.g. `PVSOverrideSystem` just becomes `system.pvsoverride` instead of `system.pvs_override`.

### DIFF
--- a/Robust.Shared/GameObjects/EntitySystem.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.cs
@@ -39,8 +39,6 @@ namespace Robust.Shared.GameObjects
                 if (name.EndsWith("System"))
                     name = name.Substring(0, name.Length - "System".Length);
 
-
-
                 // Convert CamelCase to snake_case
                 name = string.Concat(name.Select(x => char.IsUpper(x) ? $"_{char.ToLower(x)}" : x.ToString()));
                 name = name.Trim('_');
--- a/Robust.Shared/GameObjects/EntitySystem.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.cs
@@ -43,7 +43,6 @@ namespace Robust.Shared.GameObjects
                 name = string.Concat(name.Select(x => char.IsUpper(x) ? $"_{char.ToLower(x)}" : x.ToString()));
                 name = name.Trim('_');
 
-
                 return $"system.{name}";
             }
         }
--- a/Robust.Shared/GameObjects/EntitySystem.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.cs
@@ -1,8 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using System.Threading;
-using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


Robust.Shared/GameObjects/EntitySystem.cs
```diff
@@ -20,16 +19,31 @@ namespace Robust.Shared.GameObjects
     ///     This class is instantiated by the <c>EntitySystemManager</c>, and any IoC Dependencies will be resolved.
     /// </remarks>
     [Reflect(false), PublicAPI]
-    public abstract partial class EntitySystem : IEntitySystem
+    public abstract partial class EntitySystem : IEntitySystem, IPostInjectInit
     {
         [Dependency] protected readonly EntityManager EntityManager;
         [Dependency] private readonly ISharedPlayerManager _playerMan = default!;
         [Dependency] private readonly IReplayRecordingManager _replayMan = default!;
+        [Dependency] private readonly ILogManager _logMan = default!;
```
gradientvera: Maybe make this protected? In case systems want to create 'sub-sawmills' or log to the root sawmill for some reason.

Robust.Shared/GameObjects/EntitySystem.cs
```diff
@@ -180,5 +194,10 @@ protected void RaiseLocalEvent(EntityUid uid, ref object args, bool broadcast =
         }
 
         #endregion
+
+        public void PostInject()
```
gradientvera: Make this a explicit implementation

gradientvera (commented):
I like this

0x6273:
> The samill names can be overridden by systems, The automatic sawmill names don't currently properly convert to snake_case. E.g. `PVSOverrideSystem` just becomes `system.pvsoverride` instead of `system.pvs_override`.

What if we enforced PascalCase for all class names? Converting PascalCase to snake_case is easy, just insert underscore before each uppercase letter except the first. Of course this fails if you use mix of PascalCase but UPPERCASE for acronyms, since you would end up with `system.p_v_s_override`, but they should probably just be renamed to use normal PascalCase.

PJB3005:
Yeah, according to Microsoft's naming guidelines, `PVSSystem` is incorrectly named, it should be `PvsSystem` instead. I agree renaming it is best in that case.

I'd still prefer a way to override the default name though, can't make it work automatically for everything.

ElectroJr:
> I'd still prefer a way to override the default name though, can't make it work automatically for everything.

It can already be overridden via `protected override string SawmillName => "foo.bar"`. I've not made it do a basic CamelCase -> snake_case conversion, fixing existing system names can be done in other PRs.

PR #4147, opened 2023-06-17
Add CollisionLayerChangeEvent

Vordenburg:
Intended for Content, to catch when closets change their collision layers, for dynamic pathfinding.

### DIFF
--- a/Robust.Shared/Physics/Systems/FixtureSystem.cs
+++ b/Robust.Shared/Physics/Systems/FixtureSystem.cs
@@ -9,6 +9,7 @@ using Robust.Shared.Log;
 using Robust.Shared.Physics.Collision.Shapes;
 using Robust.Shared.Physics.Components;
 using Robust.Shared.Physics.Dynamics;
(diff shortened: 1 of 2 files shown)


Robust.Shared/Physics/Systems/FixtureSystem.cs
```diff
@@ -349,6 +350,12 @@ public void FixtureUpdate(EntityUid uid, bool dirty = true, bool resetMass = tru
             if (resetMass)
                 _physics.ResetMassData(uid, manager, body);
 
+            if (body.CollisionLayer != layer)
+            {
+                var ev = new CollisionLayerChangeEvent(body);
+                RaiseLocalEvent(ref ev);
+            }
```
ElectroJr: This should get raised after the properties have actually been updated, so that the event handler can use the new information. I.e., the handler might need to know about the new collision mask.

Vordenburg: Done.

PR #4158, opened 2023-06-25
Add prototype counts to IPrototypeManager

metalgearsloth:
a

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.cs
@@ -57,11 +57,19 @@ namespace Robust.Shared.Prototypes
             ReloadPrototypeKinds();
         }
 
+        /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


Robust.Shared/Prototypes/PrototypeManager.cs
```diff
@@ -134,18 +146,23 @@ public IEnumerable<IPrototype> EnumerateParents(Type kind, string id, bool inclu
...
+                yield break;
 
             var queue = new Queue<string>(iPrototype.Parents);
             while (queue.TryDequeue(out var prototypeId))
             {
-                if (!TryIndex(kind, id, out var parent))
+                if (!TryIndex(kind, prototypeId, out var parent))
                     continue;
+
                 yield return parent;
```
metalgearsloth, line 162: Also this seemed to be a bug I spotted, it iterates the queue but never actually uses the queue prototypes.

PR #4165, opened 2023-07-01
Make StyleBoxTexture respect UI zoom level

eoineoineoin:
When changing the UI zoom level, any uses of StyleBoxTexture would remain the same size, rather than scaling with the UI. This leads to a few minor inconsistencies.

Here's the game before this change, on 100% UI:
NoRobustNoGame_100
And on 175% UI:
NoRobustNoGame_175

Some of the elements remain a constant size, which in turn causes slight layout differences; look closely at the highlighted pieces here:

ProblemsHighlighted

So, the stripey background is not scaled, scrollbars are always the same size on screen and angled buttons margins appear to change (but the "clipped" part of the button is always the same size).

Issue is that the StyleBoxTexture which is used by each of these elements is not a `Control` and does not have visibility to the current UIScale. This adds a minor refactor to allow providing it when drawing the StyleBoxTexture. The old method is marked obsolete, as it's called by one place (the `StripeBack`) in-game (dependent PR incoming. EDIT: game PR is https://github.com/space-wizards/space-station-14/pull/17763) With *this* change applied but *not* the game PR applied, the game looks like:

RobustNoGame_175

Note that the button sizes and scroll bars are much more consistent now. The stripes aren't scaled, as they need an additional game change. However, there is a minor artifact in that some of these images start to display jagged pixels at higher zoom levels - look closely at the corners of the button and paper. A minor change to configure texture filtering will make them look like this:

RobustAndGame_175

I tried this out with every UI I could find and it all seems to be good. When reviewing code which used the style box texture, there were some inconsistencies around whether some controls were working with real/virtual pixels, so I tried to clarify and clean up some of those along the way.

Edit: Accompanying game change is https://github.com/space-wizards/space-station-14/pull/17763 - game will run just fine without that; it just provides minor visual improvements.


Robust.Client/UserInterface/Controls/PanelContainer.cs
```diff
@@ -36,7 +36,7 @@ protected override Vector2 ArrangeOverride(Vector2 finalSize)
         {
             var pixelSize = finalSize * UIScale;
             var ourSize = UIBox2.FromDimensions(Vector2.Zero, pixelSize);
-            var contentBox = _getStyleBox()?.GetContentBox(ourSize) ?? ourSize;
+            var contentBox = _getStyleBox()?.GetContentBox(ourSize, UIScale) ?? ourSize;
 
             foreach (var child in Children)
             {
```
ElectroJr: This should probably just use virtual pixels. I.e., using UiScale = 1, and replacing `child.ArrangePixel((UIBox2i) contentBox);` with `child.Arrange(contentBox);`

ElectroJr:
> When reviewing code which used the style box texture, there were some inconsistencies around whether some controls were working with real/virtual pixels, so I tried to clarify and clean up some of those along the way.

Inconsistent use of virtual & real pixels was also causing issues with the arrange & measure functions for panel containers, so this PR should also fix those. I.e., this PR supersedes #3800, no more spurious scrollbars in context menus!

Before:
Content Client_1SgtyPVLfT

After:
Content Client_PDtRzbUnop

Robust.Client/UserInterface/Controls/PanelContainer.cs
ElectroJr: Ill just push these changes, AFAICT it doesn't cause any issues

PR #4166, opened 2023-07-01
Add EntityQuery<MetaDataComponent> to EntityManager

metalgearsloth:
Probably makes it faster, need to profile.

Okay as long as I take out the .Deleted calls from EntityQuery then it ends up marginally faster for now:

Old:
image

New:
image

I did this by just adding internal EntityQuery methods that bypass it.


Robust.Shared/GameObjects/EntityManager.Components.cs
```diff
@@ -1402,13 +1404,94 @@ public bool Resolve(EntityUid uid, [NotNullWhen(true)] ref TComp1? component, bo
...
+        {
+            if (_traitDict.TryGetValue(uid, out var comp))
+                return (TComp1) comp;
+
+            throw new KeyNotFoundException($"Entity {uid} does not have a component of type {typeof(TComp1)}");
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        [Pure]
+        internal bool TryGetComponentInternal([NotNullWhen(true)] EntityUid? uid, [NotNullWhen(true)] out TComp1? component)
```
ElectroJr, line 1464: Even if they're internal they should still have docs/comments to explain how they differ from the public methods.

PR #4176, opened 2023-07-09
Try prevent infinite measure/arrange loops

ElectroJr:
Its currently possible for a control's measure/arrange code to inadvertently trigger infinite measure/arrange loops, which causes the game to freeze even in release mode. E.g., #3827 added a bug like this for split containers, which can be triggered by just resizing the window while using the "old chat" layout.

This PR tries to stop that from being possible in the first place by preventing measure/arrange/style updates from being queued while a update is already taking place. It also adds an arbitrary upper limit on the number of updates. When the limit is reached it'll log a warning and continue processing next frame.


PJB3005:
> It also adds an arbitrary upper limit on the number of updates. When the limit is reached it'll log a warning and continue processing next frame.

This is basically what WPF and such do, if you're curious.

PR #4184, opened 2023-07-12
MidiRenderer master/puppet and various improvements/cleanup

gradientvera:
Required by https://github.com/space-wizards/space-station-14/pull/17995
- Add a Master property to MidiRenderer, which allows it to clone all events from another renderer.
- Add a property to filter out notes in selected MIDI channels.
- Some fixes with MidiBank and stuff--
- Fixes ItemList selection breaking with "Multiple" mode. Turns out there was no other code in the codebase using the "multiple" mode...

### DIFF
--- a/Robust.Client/Audio/Midi/MidiRenderer.cs
+++ b/Robust.Client/Audio/Midi/MidiRenderer.cs
@@ -70,8 +70,8 @@ internal sealed class MidiRenderer : IMidiRenderer
             {
                 for (byte i = 0; i < ChannelCount; i++)
                 {
-                    // Channel 9 is the percussion channel. Let's not change its instrument...
-                    if (i == 9)
+                    // Don't change percussion channel instrument.
+                    if (i == RobustMidiEvent.PercussionChannel)
                         continue;
 
                     SendMidiEvent(RobustMidiEvent.ProgramChange(i, value, SequencerTick));
@@ -96,8 +96,8 @@ internal sealed class MidiRenderer : IMidiRenderer
             {
                 for (byte i = 0; i < ChannelCount; i++)
                 {
-                    // Channel 9 is the percussion channel. Let's not change its bank...
-                    if (i == 9)
+                    // Don't change percussion channel bank.
+                    if (i == RobustMidiEvent.PercussionChannel)
                         continue;
 
                     SendMidiEvent(RobustMidiEvent.BankSelect(i, value, SequencerTick));
@@ -130,8 +130,8 @@ internal sealed class MidiRenderer : IMidiRenderer
     [ViewVariables(VVAccess.ReadWrite)]
     public bool DisablePercussionChannel
     {
-        get => FilteredChannels[9];
-        set => FilteredChannels[9] = value;
+        get => FilteredChannels[RobustMidiEvent.PercussionChannel];
+        set => FilteredChannels[RobustMidiEvent.PercussionChannel] = value;
     }
 
     [ViewVariables(VVAccess.ReadWrite)]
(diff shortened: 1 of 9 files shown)


Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -181,14 +188,60 @@ public bool LoopMidi
...
+                }
+                catch
+                {
+                    // ignored
+                }
+            }
+
+            _master = value;
+
+            if (_master == null)
```
PrPleGoo, line 234: Isn't the null check cheaper than the `is { Disposed: false }`?

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -181,14 +188,60 @@ public bool LoopMidi
...
+                return;
+
+            _master.OnMidiEvent += SendMidiEvent;
+            ApplyState(_master.RendererState, true);
+            MidiBank = _midiBank;
```
PrPleGoo, line 239: Should the logic be split off from the set so it doesn't look like you're doing nothing here?

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -534,7 +597,7 @@ public void SendMidiEvent(RobustMidiEvent midiEvent)
 
                     case RobustMidiCommand.ProgramChange:
                         if (DisableProgramChangeEvent)
-                            return;
+                            break;
```
PrPleGoo, line 604: there's also a return on :621

Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -487,7 +546,12 @@ public void ApplyState(MidiRendererState state)
         }
     }
 
-    public void SendMidiEvent(RobustMidiEvent midiEvent)
+    private void SendMidiEvent(RobustMidiEvent midiEvent)
+    {
+        SendMidiEvent(midiEvent, true);
+    }
+
+    public void SendMidiEvent(RobustMidiEvent midiEvent, bool raiseEvent)
```
PrPleGoo, line 558: bool param is a code smell, just do the call for the event from the caller or make a separate method that calls both

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: The returns on 621 and 649 are intended. In this switch expression, `break` raises an event and `return` doesn't.
Because of how the new `Master/Puppet` relationship works, we still want to raise an event for a program change, even if the master renderer has it disabled, as the puppet renderers might allow it.

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: I have plans to make a cleanup PR in the future, I'll do it then

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: Yes, but we need to check that the old master is not disposed. When a renderer gets disposed, it automatically clears all event subscribers so if we tried to unsubscribe again it'd throw.

Robust.Client/Audio/Midi/MidiRenderer.cs
gradientvera: This is fine, I think having to raise the event from the caller or having a separate method would be quite a pain, given that in 99 % of cases you're going to want `raiseEvent` to be true (which is already the case with the `IMidiRenderer` interface api)

PR #4187, opened 2023-07-19
more richtext

ghost:
Set a groundwork for player-written richtext with:

- more tags (`bolditalic`, `head`, `bullet`);
- combined tags (`bold`+`italic`);
- (dis)allowed tags; and
- a fix for client CTD on unrecognised escape sequence.

used by https://github.com/space-wizards/space-station-14/pull/18154

### DIFF
--- a/Robust.Client/UserInterface/Controls/RichTextLabel.cs
+++ b/Robust.Client/UserInterface/Controls/RichTextLabel.cs
@@ -22,24 +22,14 @@ namespace Robust.Client.UserInterface.Controls
             IoCManager.InjectDependencies(this);
         }
 
-        public void SetMessage(FormattedMessage message)
-        {
-            SetMessage(message, null);
-        }
-
-        public void SetMessage(FormattedMessage message, Type[]? tagsAllowed, Color? defaultColor = null)
+        public void SetMessage(FormattedMessage message, Type[]? tagsAllowed = null, Color? defaultColor = null)
         {
             _message = message;
             _entry = new RichTextEntry(_message, this, _tagManager, tagsAllowed, defaultColor);
             InvalidateMeasure();
         }
 
-        public void SetMessage(string message)
-        {
-            SetMessage(message, null, null);
-        }
-
-        public void SetMessage(string message, Type[]? tagsAllowed, Color? defaultColor = null)
+        public void SetMessage(string message, Type[]? tagsAllowed = null, Color? defaultColor = null)
         {
             var msg = new FormattedMessage();
             msg.AddText(message);
--- a/Robust.Client/UserInterface/Controls/RichTextLabel.cs
+++ b/Robust.Client/UserInterface/Controls/RichTextLabel.cs
@@ -1,9 +1,9 @@
-﻿using System.Numerics;
+﻿using System;
+using System.Numerics;
 using JetBrains.Annotations;
 using Robust.Client.Graphics;
 using Robust.Client.UserInterface.RichText;
 using Robust.Shared.IoC;
-using Robust.Shared.Maths;
 using Robust.Shared.Utility;
 
 namespace Robust.Client.UserInterface.Controls
@@ -23,13 +23,13 @@ namespace Robust.Client.UserInterface.Controls
 
         public void SetMessage(FormattedMessage message)
         {
-            SetMessage(message, false);
+            SetMessage(message, null);
         }
 
-        public void SetMessage(FormattedMessage message, bool safeMode)
+        public void SetMessage(FormattedMessage message, Type[]? tagsAllowed)
         {
             _message = message;
-            _entry = new RichTextEntry(_message, this, _tagManager, safeMode);
+            _entry = new RichTextEntry(_message, this, _tagManager, tagsAllowed);
             InvalidateMeasure();
         }
 
--- a/Robust.Client/UserInterface/Controls/RichTextLabel.cs
+++ b/Robust.Client/UserInterface/Controls/RichTextLabel.cs
@@ -23,8 +23,13 @@ namespace Robust.Client.UserInterface.Controls
 
         public void SetMessage(FormattedMessage message)
         {
+            SetMessage(message, false);
+        }
+
+        public void SetMessage(FormattedMessage message, bool safeMode)
+        {
             _message = message;
-            _entry = new RichTextEntry(_message, this, _tagManager);
+            _entry = new RichTextEntry(_message, this, _tagManager, safeMode);
             InvalidateMeasure();
         }
 
--- a/Robust.Client/UserInterface/Controls/RichTextLabel.cs
+++ b/Robust.Client/UserInterface/Controls/RichTextLabel.cs
@@ -4,6 +4,7 @@ using JetBrains.Annotations;
 using Robust.Client.Graphics;
 using Robust.Client.UserInterface.RichText;
 using Robust.Shared.IoC;
+using Robust.Shared.Maths;
 using Robust.Shared.Utility;
 
 namespace Robust.Client.UserInterface.Controls
@@ -26,18 +27,23 @@ namespace Robust.Client.UserInterface.Controls
             SetMessage(message, null);
         }
 
-        public void SetMessage(FormattedMessage message, Type[]? tagsAllowed)
+        public void SetMessage(FormattedMessage message, Type[]? tagsAllowed, Color? defaultColor = null)
         {
             _message = message;
-            _entry = new RichTextEntry(_message, this, _tagManager, tagsAllowed);
+            _entry = new RichTextEntry(_message, this, _tagManager, tagsAllowed, defaultColor);
             InvalidateMeasure();
         }
 
         public void SetMessage(string message)
         {
+            SetMessage(message, null, null);
+        }
+
+        public void SetMessage(string message, Type[]? tagsAllowed, Color? defaultColor = null)
+        {
             var msg = new FormattedMessage();
             msg.AddText(message);
-            SetMessage(msg);
+            SetMessage(msg, tagsAllowed, defaultColor);
         }
 
         public string? GetMessage() => _message?.ToMarkup();
--- a/Robust.Client/UserInterface/RichText/BoldTag.cs
+++ b/Robust.Client/UserInterface/RichText/BoldTag.cs
@@ -13,7 +13,6 @@ public sealed class BoldTag : IMarkupTag
     [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
 
     public string Name => "bold";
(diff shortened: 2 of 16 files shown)


ghost:
content test failing. I guess `IMarkupTag.IsUnsafe` could be made optional.

EDIT: passes on the other PR at least

juliangiebel:
> content test failing. I guess `IMarkupTag.IsUnsafe` could be made optional.
> 
> EDIT: passes on the other PR at least

That's because your content PR wasn't merged yet so any tags in content don't have the `IsUnsafe` Property.
Making the property be set to false on default would be better.

juliangiebel:
I'm not sure if setting wether a tag is not safe for user input or not is enough. Different systems using richtext might want to put their own restrictions on what tags can be used.

ghost:
> I'm not sure if setting wether a tag is not safe for user input or not is enough. Different systems using richtext might want to put their own restrictions on what tags can be used.

changed it to a whitelist that's passed down. for compatibility, it can be `null` to accept all tags, but otherwise it needs to be an exhaustive list of all the tags you want to allow to prevent new tags slipping in uninvited.

ghost:
I had two thoughts while working on this that i would like some input on:

1. is there a good reason why _any_ (new) tags would live in the engine? or should they be moved to content? or should they really all live in the engine (if they're generic and not domain-specific like guidebook links)?
2. when naming tags, i used the style that was already there. but typing `[bold][/bold]` instead of `[b][/b]` is cumbersome. could tags be made shorter? or is there a better reason they should remain spelled out (ignoring the effort phasing out the existing usages)?

juliangiebel:
> I had two thoughts while working on this that i would like some input on:
> 
> 1. is there a good reason why _any_ (new) tags would live in the engine? or should they be moved to content? or should they really all live in the engine (if they're generic and not domain-specific like guidebook links)?
> 
> 2. when naming tags, i used the style that was already there. but typing `[bold][/bold]` instead of `[b][/b]` is cumbersome. could tags be made shorter? or is there a better reason they should remain spelled out (ignoring the effort phasing out the existing usages)?

The tags that are currently in engine are there because they are very generic. New tags should should usually be put into content though. 
The tags like `bold` are named like that to prevent name conflicts with projects like OpenDream. The short version of those tags could be added to content though.

Robust.Client/UserInterface/RichTextEntry.cs
```diff
@@ -39,13 +39,15 @@ internal struct RichTextEntry
 
         private readonly Dictionary<int, Control> _tagControls = new();
 
-        public RichTextEntry(FormattedMessage message, Control parent, MarkupTagManager tagManager)
+        public RichTextEntry(FormattedMessage message, Control parent, MarkupTagManager tagManager, Type[]? tagsAllowed, Color? defaultColor = null)
         {
```
juliangiebel: You should give tagsAllowed a default value of null so you don't have to pass in null every time you want to create a rich text entry without a whitelist

Robust.Client/UserInterface/Controls/RichTextLabel.cs
```diff
@@ -22,17 +23,27 @@ public RichTextLabel()
         }
 
         public void SetMessage(FormattedMessage message)
+        {
+            SetMessage(message, null);
+        }
+
+        public void SetMessage(FormattedMessage message, Type[]? tagsAllowed, Color? defaultColor = null)
         {
```
juliangiebel: `tagsAllowed` should get a default value of null

Robust.Client/UserInterface/Controls/RichTextLabel.cs
```diff
@@ -22,17 +23,27 @@ public RichTextLabel()
...
             InvalidateMeasure();
         }
 
         public void SetMessage(string message)
+        {
+            SetMessage(message, null, null);
+        }
+
+        public void SetMessage(string message, Type[]? tagsAllowed, Color? defaultColor = null)
         {
```
juliangiebel: Same here

Robust.Client/UserInterface/RichText/BoldTag.cs
```diff
@@ -14,12 +15,19 @@ public sealed class BoldTag : IMarkupTag
...
 
+    /// <inheritdoc/>
     public void PushDrawContext(MarkupNode node, MarkupDrawingContext context)
     {
-        var font = FontTag.CreateFont(context.Font, node, _resourceCache, _prototypeManager, BoldFont);
+        string fontId;
+        if (context.Tags.Any(x => x is ItalicTag))
+            fontId = BoldItalicTag.BoldItalicFont;
+        else
+            fontId = BoldFont;
```
juliangiebel: The usage of linq is generally discouraged and you need to use curly braces for if else statements.
Here is a link to our code conventions: https://docs.spacestation14.io/en/getting-started/conventions

Robust.Client/UserInterface/RichText/HeadingTag.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    [Dependency] private readonly IResourceCache _resourceCache = default!;
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
+
+    public string Name => "head";
+
+    /// <inheritdoc/>
+    public void PushDrawContext(MarkupNode node, MarkupDrawingContext context)
+    {
+        if (!node.Value.TryGetLong(out var levelParam)) return;
```
juliangiebel: The return statement needs to be on the next line with a new line below it.

Robust.Client/UserInterface/RichText/ItalicTag.cs
```diff
@@ -11,14 +12,22 @@ public sealed class ItalicTag : IMarkupTag
...
 
+    /// <inheritdoc/>
     public void PushDrawContext(MarkupNode node, MarkupDrawingContext context)
     {
-        var font = FontTag.CreateFont(context.Font, node, _resourceCache, _prototypeManager, ItalicFont);
+        string fontId;
+        if (context.Tags.Any(x => x is BoldTag))
+            fontId = BoldItalicTag.BoldItalicFont;
+        else
+            fontId = ItalicFont;
```
juliangiebel: Same issue as the bold tag

Robust.Client/UserInterface/RichText/MarkupTagManager.cs
```diff
@@ -15,24 +16,29 @@ public sealed class MarkupTagManager
...
+    private readonly Dictionary<string, IMarkupTag> _markupTagTypes = new IMarkupTag[] {
+        new BoldItalicTag(),
+        new BoldTag(),
+        new BulletTag(),
+        new ColorTag(),
+        new CommandLinkTag(),
+        new FontTag(),
+        new HeadingTag(),
+        new ItalicTag()
+    }.ToDictionary(x => x.Name.ToLower(), x => x);
```
juliangiebel, line 28: You're using linq here as well. The way it was previously done doesn't use linq. That's why it was done that way.

Robust.Client/UserInterface/RichText/MarkupTagManager.cs
```diff
@@ -59,9 +65,14 @@ public void Initialize()
...
-    public bool TryGetMarkupTag(string name, [NotNullWhen(true)] out IMarkupTag? tag)
+    private Type[]? _tagsAllowed;
+    public bool TryGetMarkupTag(string name, Type[]? tagsAllowed, [NotNullWhen(true)] out IMarkupTag? tag)
     {
-        if (_markupTagTypes.TryGetValue(name, out var markupTag))
+        if (null == tagsAllowed)
+            tagsAllowed = _tagsAllowed ??= _markupTagTypes.Values.Select(x => x.GetType()).ToArray();
+        if (_markupTagTypes.TryGetValue(name, out var markupTag)
+            // Using a whitelist prevents new tags from sneaking in.
+            && tagsAllowed.Any(x => x == markupTag.GetType()))
```
juliangiebel: You're better of just doing something like 
```cs
if (_markupTagTypes.TryGetValue(name, out var markupTag)
&& (tagsAllowed == null ||  Array.IndexOf(tagsAllowed, markupTag.GetType()) > -1))
{
```

That way you're not using linq and you can get rid of the `_tagsAllowed` property.

Robust.Client/UserInterface/RichText/MarkupDrawingContext.cs
```diff
@@ -8,22 +8,26 @@ public sealed class MarkupDrawingContext
 {
     public readonly Stack<Color> Color;
     public readonly Stack<Font> Font;
+    public readonly List<IMarkupTag> Tags;
```
juliangiebel, line 11: What is this used for?

juliangiebel (changes requested):
I found a few things that need to be changed. 
You might want to look at our code conventions: https://docs.spacestation14.io/en/getting-started/conventions
This PR looks good otherwise 👍

Robust.Client/UserInterface/RichText/MarkupDrawingContext.cs
ghost: from ede44ba: instead of adding another property to `VectorFont`, this is a list of all the unclosed tags so they can be used for combined effects like **_bold+italic_**:

https://github.com/space-wizards/RobustToolbox/blob/a02fe3e617ecc60a4791b66a5cd5db9ea316e363/Robust.Client/UserInterface/RichText/BoldTag.cs#L22

it makes more semantic sense in my mind than checking the font.

(i'll later change that to not use Linq)

Robust.Client/UserInterface/RichText/BoldTag.cs
moonheart08: In this case, adding `static` to the lambda so it's `static x => x is ItalicTag` should be fine.
C#'s compiler will get to it.

Robust.Client/UserInterface/RichText/MarkupTagManager.cs
moonheart08: This is an initalizer, using linq here will be less error prone and it only runs once.
This is fine.

moonheart08:
@onoira politely overriding Julian here, your use of Linq is fine. We only care about it in performance sensitive scenarios.

ghost:
i hope you'll forgive me for the rebase. i did it out of habit and didn't want to pop open the reflog :c

i've addressed all the comments in 42aafeb. regarding the coding conventions: i read them before starting on this project and there was nothing in there that would've warned me about Linq or bracket style. i'll keep it in mind for the future, tho.

ElectroJr (approved):
Haven't tested in-game, but changes seem fine

juliangiebel (approved):

PR #4189, opened 2023-07-22
Add method to validate prototype ids in c# fields

ElectroJr:
This PR adds a method that will use reflection to check that fields marked with specific attributes correspond to valid prototype IDs. This includes any field marked with the new `[ValidatePrototypeId<TPrototype>]` attribute, as well as any data-field that uses the `PrototypeIdSerializer` custom type serializer.

The validation method will also error when encountering optional data-fields that default to an empty string, effectively imposing a new requirement that these fields should either be nullable or have a useable default value.

Hard coded prototype IDs should still be avoided wherever possible, but they are sometimes necessary and are frequently used, so validating them is important. This also encourages people to use a single constant string, rather than just repeatedly hardcoding the same string all over the place.

For an example of the validation being used, see space-wizards/space-station-14/pull/18224, This PR does not require the content PR.

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.YamlValidate.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.YamlValidate.cs
@@ -2,11 +2,14 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
+using Robust.Shared.Reflection;
 using Robust.Shared.Serialization.Manager.Attributes;
 using Robust.Shared.Serialization.Markdown;
 using Robust.Shared.Serialization.Markdown.Mapping;
 using Robust.Shared.Serialization.Markdown.Validation;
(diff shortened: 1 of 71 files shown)


Robust.Shared/Prototypes/PrototypeManager.YamlValidate.cs
```diff
@@ -78,7 +85,99 @@ public partial class PrototypeManager
...
+                if (!field.IsStatic && obj == null)
+                {
+                    // This is an instanced field. So we will attempt to get an instance and get the default values.
+
+                    if (t.IsAbstract)
+                        continue; // TODO somehow validate private instanced fields on abstract classes.
+
+                    try
+                    {
+                        obj = Activator.CreateInstance(t);
```
ElectroJr: Relying on being able to create an instance with a parameterless constructor is somewhat icky, but I don't know of a better way to get the "default value" of non-static fields.

PR #4198, opened 2023-07-26
Set sawmill default to info on release

metalgearsloth:
The new helper defaults to verbose but I only really care for that on debug mode; anyone that wants differently can just set it on system initialize.


wrexbe:
If what you are changing is the filtering from verbose to info for release, that makes sense.

If what you are changing is the default log level for created logs, then it doesn't make sense it was verbose debug or not. A log made without specifying a level should be impossible or default to info, a verbose log level indicates something noisy or something that you don't expect to care about normally.

metalgearsloth:
> If what you are changing is the filtering from verbose to info for release, that makes sense.

Yeah that.

Updating existing stuff to use the new helper is leading to logs that were otherwise unseen becoming seen.

PR #4199, opened 2023-07-26
Add warning for unhandled replay messages

ElectroJr:
Added a warning, as its a sign that replay has some message recorded for replaying but the content client doesn't handle it for whatever reason. Related to space-wizards/space-station-14/pull/18299


wrexbe:
One reason it could happen is the recording being from a different version (guessing). I would worry about this causing a huge flood of logs if this new or unsupported type is very common, it might make sense to only show a warning once, Something like "Type x not recognized, ignoring all future type x"

ElectroJr:
Playing back replays with a mismatched client version isn't supported, and will generally just not work in the first place.
Only showing the warning once might be fine, I can update the PR when I get home.

PR #4185, opened 2023-07-13
Start MapGrid ECS

metalgearsloth:
- Fixed a bunch of warnings in entitylookupsystem and added callbacks.
- Moved MapGrid stuff to the system but left the old ones in place (no warnings or I die).
- Added some other minor perf stuff around with EntityQuery<T> / EntityQueryEnumerator<T>.

PR #4200, opened 2023-07-26
Don't recreate _entTraitDict on disconnect

ElectroJr:
This was causing a bug where `MapManager._physicsQuery` and other entity queries introduced in #4185 (and possible other queries in other managers) pointed to an old component dictionary after disconecting & reconnecting to a server.

If for whatever reason this was required, then we need to stop caching entity queries outside of entity systems or somehow update them after calling `FillComponentDict()`.

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.Components.cs
+++ b/Robust.Shared/GameObjects/EntityManager.Components.cs
@@ -77,7 +77,10 @@ namespace Robust.Shared.GameObjects
             _netComponents.Clear();
             _entCompIndex.Clear();
             _deleteSet.Clear();
-            FillComponentDict();
+            foreach (var dict in _entTraitDict.Values)
+            {
+                dict.Clear();
+            }
         }
 
         private void AddComponentRefType(CompIdx type)


metalgearsloth (approved):

wrexbe:
Could the caches being cleared, and not removed from the places holding them, cause issues if a different entity gets the same key? (If that is possible)

ElectroJr:
Entities with the same UID should be fine, but components with a different `CompIdx` would cause issues, though I don't think that's possible? Not sure though, CompIdx is some black magic.

wrexbe:
Maybe it would be safer to also have the cached references removed on disconnect?
Also if there are a ton of different uuids, could the dictionary of dictionaries get too big if it's only clearing the inner dictionaries?

metalgearsloth:
> Maybe it would be safer to also have the cached references removed on disconnect?
Also if there are a ton of different uuids, could the dictionary of dictionaries get too big if it's only clearing the inner dictionaries?

I don't think CompIdx will have issuesTM, and the query will likely be updated by archetypes later on.

Can revisit this in a bit but seems a lot of people reporting my bug.

PR #4213, opened 2023-07-30
Truncate discord rich presence strings

ElectroJr:
Stops exceptions from being thrown while applying game states.

### DIFF
--- a/Robust.Client/Utility/DiscordRichPresence.cs
+++ b/Robust.Client/Utility/DiscordRichPresence.cs
@@ -2,7 +2,6 @@ using System;
 using System.Text;
 using DiscordRPC;
 using DiscordRPC.Logging;
-using Microsoft.VisualBasic;
 using Robust.Shared;
 using Robust.Shared.Configuration;
 using Robust.Shared.IoC;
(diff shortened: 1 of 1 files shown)


Robust.Client/Utility/DiscordRichPresence.cs
```diff
@@ -1,5 +1,8 @@
+using System;
+using System.Text;
 using DiscordRPC;
 using DiscordRPC.Logging;
+using Microsoft.VisualBasic;
```
moonheart08: UHHHHHHHHHHH

ElectroJr: Ooops. Was a bad namsepace import suggestion when trying to get Encoding.UTF8

Robust.Client/Utility/DiscordRichPresence.cs
```diff
@@ -95,18 +97,58 @@ private void _stop()
...
+            while (encoding.GetByteCount(output) > bytes)
+            {
+                if (value.Length == 0)
+                    return string.Empty;
+
+                value = value.Substring(0, value.Length - 1);
+                output = value + postfix;
+            }
+
+            return output;
```
metalgearsloth, line 151: Yeah I only know string slicing which I don't think does the same.

ElectroJr: Slicing would be an easy solution if it were limited by character count, Just iteratively removing the last character is pretty shit, but this also shouldn't be getting called much so it should be fine.

PR #4217, opened 2023-08-01
Fix bad disconnects in tests

ElectroJr:
Fixes a bug where using `session.ConnectedClient.Disconnect()` would only disconnect the client on the clients side, but not on the server side. Also adds a test to check that disconnect works properly.

### DIFF
--- a/Robust.UnitTesting/RobustIntegrationTest.NetManager.cs
+++ b/Robust.UnitTesting/RobustIntegrationTest.NetManager.cs
@@ -97,8 +97,6 @@ namespace Robust.UnitTesting
                 {
                     channel.Disconnect(reason);
                 }
-
-                _channels.Clear();
             }
 
             public void ProcessPackets()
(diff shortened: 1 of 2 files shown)


Robust.UnitTesting/RobustIntegrationTest.NetManager.cs
```diff
@@ -462,9 +436,8 @@ public void SendMessage(NetMessage message)
 
                 public void Disconnect(string reason)
                 {
-                    OtherChannel.TryWrite(new DisconnectMessage(RemoteUid));
-
-                    IsConnected = false;
+                    OtherChannel.TryWrite(new DisconnectMessage(RemoteUid, reason));
+                    _owner.MessageChannelWriter.TryWrite(new DisconnectMessage(ConnectionUid, reason));
                 }
```
ElectroJr, line 441: This is where the main bug was happening. If the server called this method, it would never actually do `Disconnect?.Invoke()`, so the client would sorta stay connected,

Robust.UnitTesting/Shared/Networking/DisconnectTest.cs
```diff
@@ -0,0 +1,98 @@
...
+        // Reconnect again
+        Assert.DoesNotThrow(() => client.SetConnectTarget(server));
+        await client.WaitPost(() => cNetMan.ClientConnect(null!, 0, null!));
+        await RunTicks();
+        AssertConnected();
+
+        // Disconnect again, but using the server-channel
+        session.ConnectedClient.Disconnect("test 2");
+        await RunTicks();
+        AssertDisconnected();
```
ElectroJr, line 61: This assert would fail on master, because even after disconnecting the player would still be in the list of sessions. I.e., this assert fails:
```cs
Assert.That(sPlayerMan.Sessions.Count(), Is.EqualTo(1));
```

PR #4218, opened 2023-08-01
Fix entitylookup recursion

metalgearsloth:
If someone were buckled for example their entitylookup is never updated.


ElectroJr:
Was this causing the server crashes?

ElectroJr:
Recursive defaulting to false was also causing errors when entities were removed from containers without reparenting them. 
I've made some more changes to the PR to make the default behaviour to always be recursive, and have added two tests which fail on master.

PR #4224, opened 2023-08-02
More serialization fixes

ElectroJr:
- Fixes serialization of arrays of virtual/abstract objects. Previously it would always skip writing the `!type:FooBar` part
- Adds tests to check that the above now works properly.
- Adds a special case to the `FlagSerializer` that handles serializing `-1`, which content sometimes uses to indicate "all flags".


Robust.Shared/Serialization/Manager/SerializationManager.Writing.cs
```diff
@@ -193,7 +192,7 @@ static object ValueFactory(Type baseType, Type actualType, bool notNullableOverr
         }
 
         var type = typeof(T);
-        if (type.IsAbstract || type.IsInterface)
+        if (type.IsAbstract || type.IsInterface || !type.IsSealed)
```
metalgearsloth: If a type isn't sealed doesn't mean it can't be abstract?

ElectroJr: Ah yeah good point

PR #4239, opened 2023-08-06
Add GetActorFromUserId to actor system

Chief-Engineer:
for space-wizards/space-station-14#18709

### DIFF
--- a/Robust.Server/GameObjects/EntitySystems/ActorSystem.cs
+++ b/Robust.Server/GameObjects/EntitySystems/ActorSystem.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics.CodeAnalysis;
 using JetBrains.Annotations;
 using Robust.Server.Player;
 using Robust.Shared.GameObjects;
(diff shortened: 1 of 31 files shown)


Robust.Server/GameObjects/EntitySystems/ActorSystem.cs
```diff
@@ -116,6 +120,19 @@ private void OnActorShutdown(EntityUid entity, ActorComponent component, Compone
             // The player is fully detached now that the component has shut down.
             RaiseLocalEvent(entity, new PlayerDetachedEvent(entity, component.PlayerSession), true);
         }
+
+        public (IPlayerSession? actor, EntityUid? actorEntity) GetActorFromUserId(NetUserId? userId)
```
EmoGarbage404: feels more sane to do a TryGet that returns a boolean and has out variables for the actor and actor entity.

Chief-Engineer: Would it return true or false if it was able to get a player session but no controlled entity

metalgearsloth: True for IPlayerSession not being null, then annotate both appropriately (i.e. notnullwhentrue for IPlayerSession and maybenullwhentrue for the entity).

Robust.Server/GameObjects/EntitySystems/ActorSystem.cs
```diff
@@ -116,6 +121,19 @@ private void OnActorShutdown(EntityUid entity, ActorComponent component, Compone
...
         }
+
+        public bool TryGetActorFromUserId(NetUserId? userId, [NotNullWhen(true)] out IPlayerSession? actor, [MaybeNullWhen(true)] out EntityUid? actorEntity)
+        {
+            actor = null;
+            actorEntity = null;
+            if (userId != null)
+            {
+                actor = _playerManager.GetSessionByUserId(userId.Value);
+                actorEntity = actor.AttachedEntity;
```
ElectroJr: This should use `TryGetSessionById`

PR #4242, opened 2023-08-06
Add ContentFileRead test and fix file reading on windows

ElectroJr:
- Fixes an issue where windows would throw exceptions while using `ContentFileExists()` or `TryContentFileRead()`, including adding some logic to return early when trying to read a folder as a file.
- Adds tests to ensure that these methods behave as expected
- Tweaks `ResPathSerializer` so that it doesn't unnecessarily try to read folders as files.

### DIFF
--- a/Robust.Shared/Utility/FileHelper.cs
+++ b/Robust.Shared/Utility/FileHelper.cs
@@ -82,8 +82,13 @@ internal static class FileHelper
             // Who the fuck though this was the right way of handling that? This should clearly just be a
             // ERROR_FILE_NOT_FOUND or other result like that,
             // https://github.com/dotnet/runtime/issues/70275
-            stream = null;
-            return false;
+            if (Directory.Exists(path))
+            {
+                stream = null;
+                return false;
+            }
+
+            throw;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Robust.Shared/Utility/FileHelper.cs
```diff
@@ -38,33 +38,52 @@ public static bool TryOpenFileRead(string path, [NotNullWhen(true)] out FileStre
...
+            }
+
+            var sf = new SafeFileHandle(file, ownsHandle: true);
+            stream = new FileStream(sf, FileAccess.Read);
+            return true;
+
+        }
+        catch (UnauthorizedAccessException)
+        {
+            // UnauthorizedAccessException aka this is a folder not a file because of course that is what that means.
```
moonheart08, line 81: This may not always be the case, you should probably check the type of the path if this happens.

moonheart08 (dismissed):

Robust.Shared/Utility/FileHelper.cs
ElectroJr: Okay, but I have no idea how to do that with the windows API functions so I'm just gonna use `Directory.Exists()`.

PR #4252, opened 2023-08-10
Network entity ids

### LINKED ISSUE #4251: Net EntityUids
In porting to Arch it has come up that client and server entityuids need to be detached.

To facilitate this we need a few things:
- A networked entityuid struct.
- Add helper methods for converting it back and forth with an EntityUid.
- Adding this struct to metadatacomponent so client and server can refer to the same entity.
- Replacing all EntityUid instances in component state methods with the networked version.

Might also want to remove NetSerializable from EntityUid potentially.

metalgearsloth:
Resolves https://github.com/space-wizards/RobustToolbox/issues/4251 but going to take me a while to get something going. Content PR is https://github.com/space-wizards/space-station-14/pull/18935

- [x] Test server-side replays
- [x] Test client-side replays
- [ ] Add test ensuring client and server NetEntities match even if Uids don't (Okay fuck this RobustIntegrationTest is annoying to use and I've already spent over 10 hours on this).
- [x] Clicking the server variables VV tab crashes.
- [x] Need some kind of strategy for commands.
- [x] Ensure networked entities aren't marked as clientside.
- [x] Ensure clientside entities don't have a corresponding NetEntity.
- [x] Update componentstate code from EntityUid to NetEntity.
~~- [ ] Serialization support for NetEntity (there's some areas that are really really annoying to avoid if using datadefs).~~ (I think bobby is the only thing close to needing this?)
- [x] Add some kind of safeguard (analyser, netserializable, idk?) to prevent EntityUids being used in comp states / elsewhere.
- [x] Remove Uid handling from AllocEntity and completely detach client + server.
- [x] Remove the "c123" clientside strings for clientside entities and add some method to EntityManager to determine if an entity is clientside (i.e. invalid NetEntity).

Areas I'd recommend reviewing:
- VV
- EntityManager
- Input commands (these were split into clientside versions so interacting with clientside ghosts during construction still works)
- do_after
- Some commands

I wanted to split EntityUid to have cNumber for the string representation but it broke a lot of stuff so I dumped it.


ElectroJr:
> Add some kind of safeguard (analyser, netserializable, idk?) to prevent EntityUids being used in comp states / elsewhere.

You could just remove the [NetSerializable] attribute. Though IIRC that doesn't lead to very user friendly errors if somebody does try to use them.

Robust.Client/Replays/ReplayRecordingManager.cs
```diff
@@ -110,7 +110,9 @@ public override void RecordServerMessage(object obj)
                 compData[j++] = new ComponentChange(id, compState, tick);
             }
 
-            entStates[i++] = new EntityState(uid, compData, tick, netComps);
+            // TODO: Do we even care about netentity for replays?
```
ElectroJr: Replays do still need to distinguish between "server-side" and client-side entities. So if the net id is going to replace the old is-client-side check, then yeah

Robust.Client/GameObjects/ClientEntityManager.Network.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// </summary>
+    protected override NetEntity GenerateNetEntity() => NetEntity.Invalid;
+
+    /// <summary>
+    /// Set the inverse lookup for a particular entityuid.
+    /// </summary>
+    public void SetNetEntity(EntityUid uid, NetEntity netEntity)
+    {
+        NetEntityLookup[netEntity] = uid;
+    }
```
ElectroJr: This should also be the setter method for the metadata component's field. This should also be a `Add()` or `TryAdd()` so it logs an error on duplicate ids.

Also, would it be worth keeping an inverse dictionary? or would uid -> component-> netId eventually be faster by the magic of non-dictionary comp fetching? Though even if it is eventually going to be faster, it wouldn't hurt to just make `GetNetEntity()` directly use a dictionary if no metadata comp is passed in.

Robust.Shared/GameObjects/NetEntity.cs
```diff
@@ -0,0 +1,222 @@
...
+
+    /// <summary>
+    ///     An Invalid entity UID you can compare against.
+    /// </summary>
+    public static readonly NetEntity Invalid = new(-1);
+
+    /// <summary>
+    ///     The first entity UID the entityManager should use when the manager is initialized.
+    /// </summary>
+    public static readonly NetEntity First = new(0);
```
ElectroJr: I'm pretty sure there is EntityUid code somewhere that relies on the default EntityUid value being invalid. Having default/unspecified ids be valid ids seems like its prone to introducing confusing bugs. Even if this is what arch uses I think we should stick with 0 being invalid.

Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
```diff
@@ -225,36 +231,42 @@ private IContainer ContainerFactory(ContainerManagerComponent component, string
...
 
-        public bool RemoveExpectedEntity(EntityUid uid, [NotNullWhen(true)] out IContainer? container)
+        public bool RemoveExpectedEntity(NetEntity entity, [NotNullWhen(true)] out IContainer? container)
         {
-            if (!ExpectedEntities.Remove(uid, out container))
+            if (!ExpectedEntities.Remove(entity, out container))
                 return false;
 
-            DebugTools.Assert(container.ExpectedEntities.Contains(uid),
-                $"While removing expected contained entity {ToPrettyString(uid)}, the entity was missing from the container expected set. Container: {container.ID} in {ToPrettyString(container.Owner)}");
```
ElectroJr, line 256: If there isn't already one, there should be a `ToPrettyString(NetEntity)`, and it should be used here.

Robust.Client/GameStates/ClientDirtySystem.cs
```diff
@@ -53,7 +53,7 @@ private void OnCompRemoved(RemovedComponentEventArgs args)
             return;
 
         var comp = args.BaseArgs.Component;
-        if (!_timing.InPrediction || comp.Owner.IsClientSide() || !comp.NetSyncEnabled)
+        if (!_timing.InPrediction || IsClientSide(comp.Owner) || !comp.NetSyncEnabled)
             return;
```
ElectroJr: Now that `IsClientSide()` is slow, the order here should be changed.

Robust.Client/Physics/JointSystem.cs
```diff
@@ -62,7 +62,7 @@ private void HandleComponentState(EntityUid uid, JointComponent component, ref C
                     continue;
                 }
 
-                var other = state.UidA == uid ? state.UidB : state.UidA;
+                var other = state.UidA == ToNetEntity(uid) ? ToEntity(state.UidB) : ToEntity(state.UidA);
```
ElectroJr:
```suggestion
                var uidA = ToEntity(state.UidA);
                var other = uidA == uid ? ToEntity(state.UidB) : uidA;
```
If this comes up frequently, it might be useful to add a helper method?

Robust.Client/Replays/Loading/ReplayLoadManager.Start.cs
```diff
@@ -68,10 +68,10 @@ public async Task StartReplayAsync(ReplayData data, LoadReplayCallback callback)
             var metaState = (MetaDataComponentState?)ent.ComponentChanges.Value?
                 .FirstOrDefault(c => c.NetID == _metaId).State;
             if (metaState == null)
-                throw new MissingMetadataException(ent.Uid);
+                throw new MissingMetadataException(ent.NetEntity);
 
-            _entMan.CreateEntityUninitialized(metaState.PrototypeId, ent.Uid);
-            entities.Add(ent.Uid);
+            var uid = _entMan.CreateEntityUninitialized(metaState.PrototypeId);
+            entities.Add(uid);
```
ElectroJr, line 74: Need to `SetNetEntity()`.

Robust.Shared/GameObjects/Components/MetaDataComponent.cs
```diff
@@ -63,6 +63,12 @@ public sealed partial class MetaDataComponent : Component
         [DataField("desc")] internal string? _entityDescription;
         internal EntityPrototype? _entityPrototype;
 
+        /// <summary>
+        /// Network identifier for this entity.
+        /// </summary>
+        [ViewVariables]
+        public NetEntity NetEntity { get; internal set; } = NetEntity.Invalid;
```
ElectroJr: Maybe add an access attribute for EntityManager. Also, this should get set inside of `SetNetEntity()` to ensure they never get set separately.

Robust.Client/Replays/Playback/ReplayPlaybackManager.Checkpoint.cs
```diff
@@ -82,19 +82,20 @@ private void EnsureDetachedExist(CheckpointState checkpoint)
...
-                throw new MissingMetadataException(es.Uid);
+                throw new MissingMetadataException(es.NetEntity);
 
-            _entMan.CreateEntityUninitialized(metaState.PrototypeId, es.Uid);
-            meta = metas.GetComponent(es.Uid);
-            _entMan.InitializeEntity(es.Uid, meta);
-            _entMan.StartEntity(es.Uid);
+            _entMan.CreateEntityUninitialized(metaState.PrototypeId, uid);
+            meta = metas.GetComponent(uid);
+            _entMan.InitializeEntity(uid, meta);
```
ElectroJr: set net id

Robust.Shared/Console/Commands/MapCommands.cs
```diff
@@ -81,9 +82,9 @@ public override void Execute(IConsoleShell shell, string argStr, string[] args)
             return;
         }
 
-        var gridId = EntityUid.Parse(args[0]);
```
ElectroJr, line 84: In all shared & client-side commands, this needs to be some other functions that can parse both net-ids and normal entity uids, and I guess we need some convention for how to distinguish them as strings.

Or alternatively, the client needs to assign valid (but client-only) net ids to client-side entities. I.e., basically how the old client side ids worked.

The latter is probably better for keeping parity between server, shared, & client side code?

Robust.Server/GameObjects/EntitySystems/UserInterfaceSystem.cs
```diff
@@ -368,9 +368,9 @@ public bool OpenUi(BoundUserInterface bui, IPlayerSession session)
                 return false;
 
             _openInterfaces.GetOrNew(session).Add(bui);
-            RaiseLocalEvent(bui.Owner, new BoundUIOpenedEvent(bui.UiKey, bui.Owner, session));
+            RaiseLocalEvent(bui.Owner, new BoundUIOpenedEvent(bui.UiKey, GetNetEntity(bui.Owner), session));
```
ElectroJr: The local events shouldn't ever need to use the net entity, and the client should resolve the net -> local uid when handling the BUI message

Robust.Server/GameStates/PvsSystem.cs
```diff
@@ -322,7 +327,7 @@ private void UpdateEntityRecursive(EntityUid uid, TransformComponent xform, Enti
         // directly.
         while (children.MoveNext(out var child))
         {
-            UpdateEntityRecursive(child.Value, _xformQuery.GetComponent(child.Value), coordinates, true, forceDirty);
+            UpdateEntityRecursive(child.Value, _metaQuery.GetComponent(child.Value), _xformQuery.GetComponent(child.Value), coordinates, true, forceDirty);
         }
```
ElectroJr, line 329: Even if arch does make comp fetching faster, would it be worth having the transform child list store a (uid, netUid) tuple? Because this will add another ~20+ component fetches anytime a player moves.

Robust.Server/GameStates/PvsSystem.cs
```diff
@@ -649,51 +648,53 @@ public void ReturnToPool(HashSet<int>[] playerChunks)
...
-        EntityQuery<MetaDataComponent> metadata)
+    private bool AddToChunkSetRecursively(in NetEntity netEntity, uint visMask, RobustTree<NetEntity> tree, Dictionary<NetEntity, MetaDataComponent> set)
     {
-        if (set.ContainsKey(uid))
+        if (set.ContainsKey(netEntity))
             return true;
 
-        var mComp = metadata.GetComponent(uid);
+        var uid = GetEntity(netEntity);
+        var mComp = _metaQuery.GetComponent(uid);
```
ElectroJr: This method should take in both the netId & uid. E.g., further down you have both parent & parentNetEntity, but you only pass the netId to the function which causes a redundant  GetEntity(netEntity);

Robust.Server/GameStates/PvsSystem.cs
```diff
@@ -649,51 +648,53 @@ public void ReturnToPool(HashSet<int>[] playerChunks)
...
 
         DebugTools.Assert(!_mapManager.IsGrid(uid) && !_mapManager.IsMap(uid));
 
         var parent = xform.ParentUid;
-        if (!set.ContainsKey(parent) && //was the parent not yet added to toSend?
-            !AddToChunkSetRecursively(in parent, visMask, tree, set, transform, metadata)) //did we just fail to add the parent?
+        var parentNetEntity = _metaQuery.GetComponent(parent).NetEntity;
+
+        if (!set.ContainsKey(parentNetEntity) && //was the parent not yet added to toSend?
+            !AddToChunkSetRecursively(in parentNetEntity, visMask, tree, set)) //did we just fail to add the parent?
```
ElectroJr: Not the fault of this PR, but this causes a redundant  set.ContainsKey() check, maybe just slap //TODO on it

Robust.Server/GameStates/PvsSystem.cs
```diff
@@ -712,7 +713,17 @@ public void ReturnToPool(HashSet<int>[] playerChunks)
...
+
+        if (deletions?.Count > 0)
+        {
+            foreach (var netEntity in deletions)
+            {
+                netDeletions.Add(netEntity);
+            }
+        }
 
+        _poolManager.Return(deletions);
```
ElectroJr: What is the point of this? This is just copying the list, and as far as I know GetDeletedIndices() isn't pooled and shouldn't be returned?

These changes should either just be removed, or  GetDeletedIndices() needs to be properly pooled & returned. In order to minimize the chance of accidentally adding misc pooling bugs, its probably better to just add a //TODO and leave the code unchanged for now.

Robust.Server/GameStates/PvsSystem.cs
```diff
@@ -980,32 +1004,35 @@ public void ReturnToPool(HashSet<int>[] playerChunks)
...
+            var childNetEntity = GetNetEntity(child);
+
+            if (!toSend.ContainsKey(childNetEntity))
             {
-                var (entered, _) = ProcessEntry(in child, lastAcked, lastSent, lastSeen, ref newEntityCount,
+                var (entered, _) = ProcessEntry(in childNetEntity, lastAcked, lastSent, lastSeen, ref newEntityCount,
                     ref enteredEntityCount, newEntityBudget, enteredEntityBudget);
 
-                AddToSendSet(in child, _metaQuery.GetComponent(child), toSend, fromTick, in entered, ref entStateCount);
+                AddToSendSet(in childNetEntity, _metaQuery.GetComponent(child), toSend, fromTick, in entered, ref entStateCount);
```
ElectroJr: Move outside of the if block and use in GetNetEntity()

Robust.Server/GameStates/PvsSystem.cs
```diff
@@ -1221,7 +1248,8 @@ private EntityState GetEntityState(ICommonSession? player, EntityUid entityUid,
         }
 
         DebugTools.Assert(meta.EntityLastModifiedTick >= meta.LastComponentRemoved);
-        var entState = new EntityState(entityUid, changed, meta.EntityLastModifiedTick, netComps);
+        DebugTools.Assert(meta.NetEntity.IsValid());
```
ElectroJr:
```
DebugTools.Assert(EntityManager.GetEntity(meta.NetEntity) == entityUid);
```

Robust.Shared.CompNetworkGenerator/ComponentNetworkGenerator.cs
```diff
@@ -115,27 +120,59 @@ private static string GenerateSource(in GeneratorExecutionContext context, IName
...
-                    getStateInit.Append($@"
+                    case GlobalEntityUidName:
+                    case GlobalNullableEntityUidName:
+                        stateFields.Append($@"
+        public NetEntity{nullableAnnotation} {name} = default!;");
+
+                        getStateInit.Append($@"
+                {name} = GetNetEntity(component.{name}),");
+                        handleStateSetters.Append($@"
+            component.{name} = GetEntity(state.{name});");
```
ElectroJr: I just realized there is an issue here that'll break comp state networking, thought it'd also break entity coordinate networking and thus also things like & sound sources.

Previously a component could reference some other entity that the client hasn't received yet. Initially any TryComp() and such would just fail, but when the client does eventually receive the entity, things should work fine. However, now the netid -> uid  will fail and the data in the component will just be stuck in a bad state until the server dirties the component and re sends it, which it might never do.

If I get around to it ill write an actual test for this, but for now a simple test would be to just have an overlay that draws a line between two entities, and then have them enter PVS one after the other. 

For now, the only way I can think of fixing it is to have the client-side `GetEntity()` generate/reserve an entity uid if it encounters an unknown net id. ApplyGameState() would then need to check if a uid has been reserved whenever it encounters an unknown netId, and then create an entity with the reserved uid (which also means re-adding that AllocEntity arg).

Robust.Shared/Console/CompletionHelper.cs
```diff
@@ -155,33 +155,41 @@ public static IEnumerable<CompletionOption> MapUids(IEntityManager? entManager =
         }
     }
 
-    public static IEnumerable<CompletionOption> EntityUids(string text, IEntityManager? entManager = null)
+    public static IEnumerable<CompletionOption> NetEntities(string text, IEntityManager? entManager = null)
     {
```
ElectroJr, line 159: If we don't create valid net-ids for client-side entities, then this should me moved to Robust.Server to try and ensure that people don't accidentally write invalid commands in shared code.

Robust.Client/GameStates/ClientGameStateManager.cs
```diff
@@ -844,7 +851,7 @@ public IEnumerable<EntityUid> ApplyGameState(GameState curState, GameState? next
             var toDelete = new List<EntityUid>(Math.Max(64, _entities.EntityCount - stateEnts.Count));
             foreach (var ent in currentEnts)
             {
-                if (ent.IsClientSide())
+                if (_entities.IsClientSide(ent))
                 {
```
ElectroJr: Move the `metas.TryGetComponent` up so you can pass it into IsClientSide.

Robust.Client/GameStates/ClientGameStateManager.cs
```diff
@@ -1015,44 +1028,48 @@ public void DetachImmediate(List<EntityUid> entities)
                     DebugTools.Assert((meta.Flags & MetaDataFlags.InContainer) == 0);
 
                     if (container != null)
-                        containerSys.AddExpectedEntity(ent, container);
+                        containerSys.AddExpectedEntity(_entities.GetNetEntity(ent), container);
```
ElectroJr: Just use the netEntity from the foreach loop

Robust.Client/GameObjects/ClientEntityManager.Network.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    public override bool IsClientSide(EntityUid uid, MetaDataComponent? metadata = null)
+    {
+        // Can't log false because some content code relies on invalid UIDs.
+        if (!MetaQuery.Resolve(uid, ref metadata, false))
+            return false;
+
+        return !metadata.NetEntity.IsValid();
+    }
```
ElectroJr: Is there any reason we can't just keep the old clientside check? It would make this significantly faster,

ElectroJr (changes requested):
I still intend to review the content PR

Robust.Client/GameObjects/ClientEntityManager.Network.cs
metalgearsloth: Do you mean the bitflag or?

If we're using genids then the id is going to point to its array index.

ElectroJr: gen ids work by having some bits for the generation, while others are the index, right? is there some spare bit that can still be used to store misc "metadata"?

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -637,24 +631,35 @@ public virtual void FlushEntities()
...
 
 #pragma warning disable CS0618
-            metadata = new MetaDataComponent { Owner = uid };
+            metadata = new MetaDataComponent
+            {
+                Owner = uid,
+                NetEntity = netEntity,
+            };
+
+            // TODO: Dump on server
```
ElectroJr:
```suggestion
```

Robust.Shared/GameObjects/EntityState.cs
```diff
@@ -9,7 +9,10 @@ namespace Robust.Shared.GameObjects
     [Serializable, NetSerializable]
     public sealed class EntityState
     {
-        public EntityUid Uid { get; }
```
ElectroJr, line 12: Seeing as the client will repeatedly use the state while resetting prediction, its probably worth keeping the uid field and resolving it only once. Thought that doesn't need to be done now and I might do this in a separate PR.

Robust.Shared/GameObjects/EntitySystem.Proxy.cs
```diff
@@ -875,4 +875,288 @@ protected IEnumerable<(TComp1, TComp2, TComp3, TComp4)> EntityQuery<TComp1, TCom
...
+    public bool TryGetEntity(NetEntity nEntity, out EntityUid entity)
+    {
+        return EntityManager.TryGetEntity(nEntity, out entity);
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public bool TryGetEntity(NetEntity? nEntity, [NotNullWhen(true)] out EntityUid? entity)
+    {
+        return EntityManager.TryGetEntity(nEntity, out entity);
+    }
```
ElectroJr: Having one use a nullable out, while the other is not nullable is a bit weird. It should just be one or the other.

Robust.Shared/GameObjects/Serializers/NetEntitySerializer.cs
```diff
@@ -0,0 +1,46 @@
...
+        return int.TryParse(nodeValue, CultureInfo.InvariantCulture, out _) ? new ValidatedValueNode(node) : new ErrorNode(node, "Failed parsing NetEntity.");
+    }
+
+    public NetEntity Read(ISerializationManager serializationManager, ValueDataNode node, IDependencyCollection dependencies,
+        SerializationHookContext hookCtx, ISerializationContext? context = null, ISerializationManager.InstantiationDelegate<NetEntity>? instanceProvider = null)
+    {
+        var entity = serializationManager.Read<EntityUid>(node);
+        var entManager = dependencies.Resolve<IEntityManager>();
+        return entManager.GetNetEntity(entity);
+    }
```
ElectroJr: This confused me for a bit, so its probably worth adding some comments noting that:
- This serializer relies on a serialization context that provides an entity uid serializer.
- All entities that are to be loaded must have been allocated prior to deserialization.

Also, are there actually net id data fields that get used by existing maps?  I.e., do the map load tests implicitly test this serializer or does this need a test? (I haven't gotten to the content PR yet)

Robust.Shared/GameObjects/Serializers/NetEntitySerializer.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        var entManager = dependencies.Resolve<IEntityManager>();
+        var entity = entManager.GetEntity(value);
+        return serializationManager.WriteValue(entity);
+    }
+
+    public NetEntity CreateCopy(ISerializationManager serializationManager, NetEntity source, IDependencyCollection dependencies,
+        SerializationHookContext hookCtx, ISerializationContext? context = null)
+    {
+        return new NetEntity(source.GetHashCode());
```
ElectroJr: Why not just make the int public? Also makes some debugging slightly easier

Robust.Shared/Map/EntityCoordinates.cs
```diff
@@ -10,10 +10,9 @@
 namespace Robust.Shared.Map
 {
     /// <summary>
-    ///     A set of coordinates relative to another entity.
+    ///     A set of coordinates relative to another entity. Used for net
```
ElectroJr: incomplete comment. I guess its meant to refer to the net ent coords?

Robust.Shared/GameObjects/EntityManager.Network.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// <inheritdoc />
+    public bool TryGetNetEntity(EntityUid uid, out NetEntity netEntity, MetaDataComponent? metadata = null)
+    {
+        if (uid == EntityUid.Invalid)
+        {
+            netEntity = NetEntity.Invalid;
+            return false;
+        }
+
+        // I wanted this to logMissing but it seems to break a loootttt of dodgy stuff on content.
```
ElectroJr, line 97: Ill get around to looking at the content PR next, but how much is a lot? Should it at least log a warning?
And what causes it?

Robust.Shared/GameObjects/EntityManager.Network.cs
```diff
@@ -0,0 +1,488 @@
...
+        return GetEntity(nEntity.Value);
+    }
+
+    /// <inheritdoc />
+    public NetEntity GetNetEntity(EntityUid uid, MetaDataComponent? metadata = null)
+    {
+        if (uid == EntityUid.Invalid)
+            return NetEntity.Invalid;
+
+        // I wanted this to logMissing but it seems to break a loootttt of dodgy stuff on content.
```
ElectroJr: Even moreso here, when its not a try-get function

metalgearsloth:
Talked about it on disc but tl;dr currently blocked on handling entities we don't know about yet.

Robust.Shared/GameObjects/NetEntity.cs
metalgearsloth: I think with arch they ended up changing it to start at 1 due to the propensity for default bugs so will change.

Robust.Client/GameObjects/ClientEntityManager.Network.cs
metalgearsloth: With genids it *should* be faster to index the archetype array then index metadatacomponent's type in it rather than do a dictionary lookup but will benchmark it later.

I did the other thing.

Robust.Client/Physics/JointSystem.cs
metalgearsloth: I don't think it's that common, mainly just joints code, and box2d is getting updated to v3 at some point which refactors the entire thing which will require me to re-write most of physics for it so I'm trying to avoid changes before then.

Robust.Server/GameObjects/EntitySystems/UserInterfaceSystem.cs
metalgearsloth: I'll make another abstract for it.

Robust.Server/GameStates/PvsSystem.cs
metalgearsloth: With archetypes GetComponent scales much better with getting multiple components at the same time as you have an array of component arrays, so you just index the same slot into 2 different arrays instead of 1.

With that being said I want to look at refactoring out MoveEvent from PVSSystem at some point so I'm fine dealing with the overhead in the interim.

Robust.Shared/GameObjects/EntityState.cs
metalgearsloth: Yeah I'm not sure where this done atm so could put the field in and put NonSerialized on it?

Robust.Shared/GameObjects/Serializers/NetEntitySerializer.cs
metalgearsloth: Yeah I'll change it.

Robust.Client/GameObjects/ClientEntityManager.Network.cs
metalgearsloth: I don't think there's an easy way to do it to my knowledge. If it's an issue we could keep a dummy clientside world that stores entityuids that are clientside if we need the lookups to be fast as it'd just be an array index (though obviously not going to be as fast as just setting an arbitrary bit like we do atm which is somewhat hacky anyway).

Robust.Shared/Map/EntityCoordinates.cs
metalgearsloth: Yeah it re

Robust.Shared/GameObjects/Serializers/NetEntitySerializer.cs
metalgearsloth: DoAfter I think was the main edge case of "I kinda sorta maybe need to serialize NetEntities or double stuff". There might be one or two more. It might be easier to just remove it and have content handle it.

Robust.Server/GameStates/PvsSystem.cs
metalgearsloth: Think it's a holdover from before I rewrote most of pvs for NetEntity so it's removeable.

ElectroJr:
I pushed some changes, should be easy to examine commit by commit:
- Moved `IEntityManager.IsClientSide(NetEntity)` into `NetEntity`
- Fixed partial state reset erroring when the state contains new entities
- Changed the `GetEntity()` in `ApplyEntityStates` and `ProcessDeletions` into a try-get
- Changed `GetEntity()` so that it returns the default value when trying to parse an unknown entity uid
  - This is what the docstrings said the method should do, and a bunch of code seemed to assume that that is how it would work.
  - If it should be changed back to throwing, then we should probably add a `GetEntityOrDefault()`
- Removed the colleciton pooling from `EntityManager.Network.cs`
  - There is no clear indication that these methods return pooled collections, and a bunch of code was just never returning the collections to the pool
  - IMo we should just add overrides that explicitly return pooled collections, or just add overrides that take in an existing collection, seeing as the caller already has to handle returning the collection to the pool anyways.

ElectroJr (approved):

PR #4267, opened 2023-08-15
Remove personally-identifiable file paths from client logs

Partmedia:
This removes PII from client logs, per [the discussion in #maintainers](https://discord.com/channels/310555209753690112/900426319433728030/1140754747566071828.) Removed are the full paths to the client config file, Robust directory for IL verification, and path to .NET root.

### DIFF
--- a/Robust.Shared/ContentPack/AssemblyTypeChecker.cs
+++ b/Robust.Shared/ContentPack/AssemblyTypeChecker.cs
@@ -69,6 +69,10 @@ namespace Robust.Shared.ContentPack
             {
                 _sawmill.Debug("Robust directory not available");
             }
+            else
+            {
+                loadDirs.Add(Path.GetDirectoryName(ourPath)!);
+            }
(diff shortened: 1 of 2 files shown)


Robust.Shared/ContentPack/AssemblyTypeChecker.cs
```diff
@@ -71,12 +71,9 @@ internal Resolver CreateResolver()
             }
             else
             {
-                _sawmill.Debug("Robust directory is {0}", ourPath);
```
DrSmugleaf, line 74: Are these debug logs enabled in a release build?

DrSmugleaf (approved):

Robust.Shared/ContentPack/AssemblyTypeChecker.cs
PJB3005: Yeah.

PR #4270, opened 2023-08-18
Remove FixtureId from fixtures

metalgearsloth:
The dictionary already has it. Just means content needs to be more careful. Also lets us remove the serializer hack around it (though we still need the serializer atm for grid fixtures).

ATM it's added to FixtureProxy so AddPairs might get marginally slower.


metalgearsloth:
Merge after serv4 obviously

PR #4273, opened 2023-08-20
Check if Zstd is installed

wrexbe:
Checks if Zstd is installed, and stops the startup if it isn't, with a fatal log.

### DIFF
--- a/Robust.Client/GameController/GameController.cs
+++ b/Robust.Client/GameController/GameController.cs
@@ -110,6 +110,13 @@ namespace Robust.Client
         {
             DebugTools.AssertNotNull(_resourceManifest);
 
+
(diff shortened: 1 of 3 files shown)


PJB3005:
What prompted this?

wrexbe:
Not having Zstd on my machine

PJB3005:
> Not having Zstd on my machine

What? The game ships zstd itself. Are you on ARM?

wrexbe:
Yes, my computer is Arm :)

PJB3005:
We don't currently officially support ARM. This isn't the only binary you're going to have issues with, so this PR makes no sense. We do not want to encourage people to rely on system packages.

wrexbe:
Alright

PR #4285, opened 2023-08-21
Advertise to multiple hubs simultaneously

PrPleGoo:
new CVar hub_urls
CVar master_url mentioned as deprecated
HubManager will now ping all hub URLs


Robust.Server/ServerHub/HubManager.cs
```diff
@@ -100,33 +114,34 @@ private async void SendPing()
...
-        var apiUrl = $"{_masterUrl}api/servers/advertise";
-
-        try
+        foreach (var hubUrl in _hubUrls)
         {
-            using var response = await _httpClient!.PostAsJsonAsync(apiUrl, new AdvertiseRequest(_advertiseUrl!));
+            var apiUrl = $"{hubUrl}api/servers/advertise";
 
-            if (!response.IsSuccessStatusCode)
+            try
```
Jezithyr, line 112: try to avoid using try-catch, an if check would be more efficient here. Throwing and handling exceptions is considerably more expensive than a simple if check

Robust.Server/ServerHub/HubManager.cs
```diff
@@ -100,33 +114,34 @@ private async void SendPing()
...
+                {
+                    _sawmill.Info($"Successfully advertised to hub {hubUrl} with address {_advertiseUrl}");
+                    _firstAdvertisementByHubUrl.Add(hubUrl, true);
+                }
             }
-
-            if (_firstAdvertisement)
+            catch (Exception e)
             {
-                _sawmill.Info("Successfully advertised to hub with address {ServerHubAddress}", _advertiseUrl);
```
Jezithyr, line 122: We should still log info that the server was successfully advertised. Also it would be useful to know which hub the server was advertised on.

Robust.Shared/CVars.cs
```diff
@@ -1304,10 +1304,17 @@ protected CVars()
 
         /// <summary>
         /// URL of the master hub server to advertise to.
+        /// Depricated: overwrites the value of hub.hub_urls
```
Jezithyr: why deprecate this cvar instead of removing it?

Jezithyr (changes requested):

Robust.Server/ServerHub/HubManager.cs
PrPleGoo: the try catch was already here

Robust.Shared/CVars.cs
PrPleGoo: Didn't want to cause breaking changes to a setting that might already be in use by some servers.
I can remove it if that's alright.

Robust.Server/ServerHub/HubManager.cs
PrPleGoo: see :137

Robust.Server/ServerHub/HubManager.cs
Jezithyr: Still it might be worth replacing it with an if check instead to improve perf, but that's up to you. I'd highly reccommend it but I won't block merging over a nitpick like that

Robust.Shared/CVars.cs
Jezithyr: Yep, removing it shouldn't break anything

Robust.Server/ServerHub/HubManager.cs
Jezithyr: I see that, yeah sorry I missed that the first time 😅 looks good to me

Jezithyr:
I'll need to do some tests to make sure this doesn't somehow brick the launcher when I get home from work. But this looks like it's good to merge so far

Robust.Server/ServerHub/HubManager.cs
```diff
@@ -38,7 +39,14 @@ public async void Start()
...
 
         _cfg.OnValueChanged(CVars.HubAdvertiseInterval, UpdateInterval, true);
-        _cfg.OnValueChanged(CVars.HubMasterUrl, s => _masterUrl = s, true);
+        _cfg.OnValueChanged(CVars.HubUrls, s => {
+            var hubUrls = s.Split(",", StringSplitOptions.RemoveEmptyEntries);
+            for (var i = 0; i < hubUrls.Length; i++)
+            {
+                hubUrls[i] = hubUrls[i].Trim();
+            }
+            _hubUrls = hubUrls;
```
Visne:
```suggestion
            _hubUrls = s.Split(",", StringSplitOptions.RemoveEmptyEntries).Select(u => u.Trim()).ToArray();
```

Robust.Server/ServerHub/HubManager.cs
```diff
@@ -100,33 +108,34 @@ private async void SendPing()
...
-                    errorText);
-                return;
+                using var response = await _httpClient!.PostAsJsonAsync(apiUrl, new AdvertiseRequest(_advertiseUrl!));
+
+                if (!response.IsSuccessStatusCode)
+                {
+                    var errorText = await response.Content.ReadAsStringAsync();
+                    _sawmill.Log(
+                        LogLevel.Error,
+                        $"Error status while advertising server: [{response.StatusCode}] {errorText}, to hub: {hubUrl}");
```
Visne: I know the code was already like that, but you can just use `_sawmill.Error()` here. Also, passing the args separately instead of string interpolation like it was before is probably better.

Robust.Server/ServerHub/HubManager.cs
```diff
@@ -100,33 +108,34 @@ private async void SendPing()
...
+                    return;
+                }
+
+                if (!_firstAdvertisementByHubUrl.ContainsKey(hubUrl))
```
Visne: This only checks if it has the key, not whether its value is `true`. It probably makes more sense anyways to just store the hub URLs that were advertised to, instead of a dictionary which will only ever have the values `true`.

Robust.Server/ServerHub/HubManager.cs
```diff
@@ -100,33 +108,34 @@ private async void SendPing()
...
+                    _firstAdvertisementByHubUrl.Add(hubUrl, true);
+                }
             }
-
-            if (_firstAdvertisement)
+            catch (Exception e)
             {
-                _sawmill.Info("Successfully advertised to hub with address {ServerHubAddress}", _advertiseUrl);
-                _firstAdvertisement = false;
+                _sawmill.Log(LogLevel.Error, e, $"Exception while trying to advertise server to hub {hubUrl}");
```
Visne: Probably better to use args instead of string interpolation for these two logs too.

Robust.Server/ServerHub/HubManager.cs
Visne: Try-catch is fine here imo.

Visne (changes requested):

Robust.Server/ServerHub/HubManager.cs
Visne: Or just use a `List<string>`, doesn't really have to be an array.

Visne (approved):

Visne:
Code looks good to me, but definitely needs some real testing before merge.

Robust.Server/ServerStatus/StatusHost.cs
```diff
@@ -46,11 +47,11 @@ internal sealed partial class StatusHost : IStatusHost, IDisposable
 
         private string? _serverNameCache;
         private string? _serverDescCache;
-        private string[]? _serverTagsCache;
+        private IReadOnlyList<string> _serverTagsCache = Array.Empty<string>();
```
PJB3005, line 50: Why change this file?

Visne: I think they originally copy-pasted the for-loop on line 113 to `HubManager.cs`, then I recommended to just use Linq there, and now they are also applying that here for consistency.

PrPleGoo: Yeah, exactly what Visne said.

Visne:
I tested this today and it seems to work fine, even with a pretty messed up config value with multiple commas, invalid addresses, and whitespace between commas.

PR #4289, opened 2023-08-22
Ability to change IPlayingStream audio parameters

ilinm1:
Added SetAudioParams() and GetAudioParams() methods


Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -55,6 +55,7 @@ public override void Initialize()
         SubscribeNetworkEvent<PlayAudioGlobalMessage>(PlayAudioGlobalHandler);
         SubscribeNetworkEvent<PlayAudioPositionalMessage>(PlayAudioPositionalHandler);
         SubscribeNetworkEvent<StopAudioMessageClient>(StopAudioMessageHandler);
+        SubscribeNetworkEvent<SetAudioParametersMessage>(SetAudioParamsHandler);
```
metalgearsloth: This should be handled as party of audio being dirtied not special-casing events.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -113,6 +114,15 @@ private void StopAudioMessageHandler(StopAudioMessageClient ev)
...
+    
+    private void SetAudioParamsHandler(SetAudioParametersMessage ev)
+    {
+        PlayingStream? stream = _playingClydeStreams.Find(p => p.NetIdentifier == ev.Identifier);
+        if (stream == null)
+            return;
+        
+        SetAudioParams(stream, ev.AudioParams);
+    }
     #endregion
```
metalgearsloth: As above and also using LINQ on hot paths is a nono.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -320,7 +330,9 @@ private PlayingStream CreateAndStartPlayingStream(IClydeAudioSource source, Audi
         source.StartPlaying();
         var playing = new PlayingStream
         {
+            AudioSystem = this,
             Source = source,
+            Stream = stream,
```
metalgearsloth: Streams shouldn't contain references to systems as we use ECS and data composition.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -518,7 +530,9 @@ private void ApplyAudioParams(AudioParams? audioParams, IClydeAudioSource source
 
     public sealed class PlayingStream : IPlayingAudioStream
     {
+        public AudioSystem AudioSystem = default!;
         public uint? NetIdentifier;
+        public AudioStream Stream = default!;
```
metalgearsloth: Isn't Stream gettable through source anyway.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -562,6 +576,16 @@ public void Stop()
...
+        public void SetAudioParams(AudioParams parameters)
+        {
+            AudioSystem.SetAudioParams(this, parameters);
+        }
+
+        public AudioParams GetAudioParams()
+        {
+            return AudioParams.Default;
+        }
     }
```
metalgearsloth: Looks placeholder? Why is it always returning default audioparams.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -17,21 +17,45 @@ public sealed class AudioSystem : SharedAudioSystem
...
+        {
+            return Parameters;
+        }
+
         public void Stop()
         {
-            _audioSystem.InternalStop(_id, _sessions);
+            _audioSystem.InternalStop(Id, Sessions);
         }
     }
```
metalgearsloth: These shouldn't be public as content shouldn't worry about audio source handling.

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -62,16 +86,19 @@ private uint CacheIdentifier()
...
         var id = CacheIdentifier();
+        
+        AudioParams parameters = audioParams ?? AudioParams.Default;
+        
         var msg = new PlayAudioGlobalMessage
         {
             FileName = filename,
-            AudioParams = audioParams ?? AudioParams.Default,
+            AudioParams = parameters,
             Identifier = id
```
metalgearsloth, line 82: Pulling this out is an unnecessary change.

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -562,6 +576,16 @@ public void Stop()
...
         }
+        
+        public void SetAudioParams(AudioParams parameters)
+        {
+            AudioSystem.SetAudioParams(this, parameters);
+        }
+
+        public AudioParams GetAudioParams()
+        {
+            return AudioParams.Default;
```
ilinm1: Idk how to retrieve current parameters since IClydeAudioSource only implements methods for setting them, if it's strictly required to make this in order for this to get merged tell me

metalgearsloth (changes requested):

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
ilinm1: no, atleast not without typecasting source interface to more specific types

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
ilinm1: yeah wrote about it above. I'll implement it later

Robust.Server/GameObjects/EntitySystems/AudioSystem.cs
ilinm1: this was made to not write the same ternary op twice

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
ilinm1: functional now

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
ilinm1: done

Robust.Client/GameObjects/EntitySystems/AudioSystem.cs
```diff
@@ -78,34 +78,79 @@ private void OnRaycastLengthChanged(float value)
...
+        stream = default!;
+        
+        PlayingStream? st = ResolveStream(identifier);
+        if (st == null)
+            return false;
+
+        stream = st;
+        return true;
+    }
+
```
metalgearsloth: Store them by netidentifier rather than iterating potentially hundreds of sounds every time.

Also just use the latter and call it TryGetStream.

metalgearsloth (changes requested):

metalgearsloth:
Also for any future changes we should just bite the bullet and make audio sources entities instead.

ilinm1:
this pr also requires https://github.com/space-wizards/space-station-14/pull/20397

Robust.Shared/Audio/AudioParams.cs
```diff
@@ -35,48 +35,48 @@ public partial struct AudioParams
         ///     The DistanceModel to use for this specific source.
         /// </summary>
         [DataField("attenuation")]
-        public Attenuation Attenuation { get; set; } = Default.Attenuation;
+        public Attenuation? Attenuation { get; set; } = Default.Attenuation;
```
Partmedia, line 38: Could you justify why these are now nullable, given that they are given a default? This is a breaking API change that requires a major engine version bump.

Robust.Shared/Audio/AudioParams.cs
```diff
@@ -91,23 +91,28 @@ public partial struct AudioParams
...
 
         public AudioParams(
-            float volume,
-            float pitchScale,
-            string busName,
-            float maxDistance,
-            float refDistance,
-            bool loop,
-            float playOffsetSeconds,
+            float? volume,
```
Partmedia, line 103: You can make the constructor accept nullable fields without making the public fields nullable as well.

Robust.Shared/Audio/AudioParams.cs
ilinm1: sloth has asked to change audio parameters by sending play sound message twice, instead of using separate message. so nullable fields in audio params are for comparing new parameters with the old ones. some of the parameters cannot be (and I think shouldn't be) retrieved after setting them, so it's the only way I saw to tweak only the necessary values. if there is a better way please tell me.

Robust.Shared/Audio/AudioParams.cs
ilinm1: uhm, but they should be nullable in current implementation.

ilinm1:
not working on the feature this pr was made for anymore

PR #4290, opened 2023-08-22
Raise MapInitEvent on components added after spawn

### LINKED ISSUE #4288: adding components in VV does not mapinit them
## Description
if the map has inited then it should run mapinitevent, currently it never does even if you want it to
ideally components added in code would do this too but its not *as* important

metalgearsloth:
Resolves https://github.com/space-wizards/RobustToolbox/issues/4288

Requires space-wizards/space-station-14/pull/19965


Robust.Client/GameObjects/ClientEntityManager.cs
```diff
@@ -90,6 +90,13 @@ public override void Dirty(EntityUid uid, Component component, MetaDataComponent
                 base.Dirty(uid, component, meta);
         }
 
+        protected override void ComponentMapInit(IComponent component, MetaDataComponent metadata)
+        {
+            /*
+             * NOOP on client as client doesn't run MapInit; runs the component event on server for components added after MapInit.
+             */
```
ElectroJr: it should still mapinit if its a client-side entity (e.g., singleplayer games).

Normal entity spawning mapinit logic is probably also broken, but thats a separate issue ill fix in a map loading PR

ElectroJr (changes requested):

ElectroJr (approved):

PR #4295, opened 2023-08-23
Fix serialization sharing instances when copying data definitions and not assigning null when the source is null

DrSmugleaf:
No more communitary lung bug


Jezithyr:
One of, if not the greatest SS14 bugs of all time.

PR #4300, opened 2023-08-23
Fix Timer drift

### LINKED ISSUE #4298: Robust.Shared.Timing.Timer has time drift
This issue has been exposed by extending the end-round timer for salamander to five minutes, but it takes way longer than five minutes, no actually accurate timing can be provided.

For a timer set to five minutes:
- On my local machine, no drift is created at all.
- On @Arimah 's machine, the clock drifts 12 seconds.

This is especially a problem for long-running timers, as the dilation is made worse there. The implementation should rely on a more reliable source of passed time to resolve this issue.

arimah:
Fixes #4298

As described in the issue, the cast to `int` causes fractions of a millisecond to be lost. This means the timer actually counts down slower than real time. If the game runs at a stable (exact) 60 fps, the timer is _guaranteed_ to drift by 4%. At an (exact) 59 fps, the drift increases to 5.6%.

Changing `_timeCounter` to a float and removing the cast fixes the problem and appears to have no performance impact whatsoever.

Note, however: due to floating point error accumulation, this fix will actually make timers run slightly _faster_. In my testing, a 300 s timer ran on average 30 ms (0.01%) faster. I would argue that a 0.01% speedup is acceptable compared to a _guaranteed_ slowdown.

I have been told the `Timer` class is not used much in newer code. Nevertheless, this drift issue noticeably affects the round restart time on Salamander, and I posit that it's better to fix a known problem.

### DIFF
--- a/Robust.Shared/Timing/Timer.cs
+++ b/Robust.Shared/Timing/Timer.cs
@@ -11,7 +11,7 @@ namespace Robust.Shared.Timing
         /// <summary>
         /// Counts the time (in milliseconds) before firing again.
         /// </summary>
-        private int _timeCounter;
+        private float _timeCounter;
 
         /// <summary>
         /// Time (in milliseconds) between firings.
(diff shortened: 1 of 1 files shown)


lzk228:
does that should fix https://github.com/space-wizards/space-station-14/issues/17131 ?

PJB3005:
No, that sounds unrelated.

PR #4307, opened 2023-08-24
Convert `Tile.TypeId` to an int

### LINKED ISSUE #4306: Change Tile Ids to use an int
OpenDream is running up against 65k tiles. No map updates required just need to update code in a few areas.

ike709:
Fixes #4306

OpenDream didn't explode:
image


Robust.Shared/Map/Tile.cs
```diff
@@ -41,37 +41,13 @@ namespace Robust.Shared.Map;
...
         Flags = flags;
         Variant = variant;
     }
 
-    /// <summary>
-    ///     Explicit conversion of <c>Tile</c> to <c>uint</c> . This should only
-    ///     be used in special cases like serialization. Do NOT use this in
-    ///     content.
-    /// </summary>
-    public static explicit operator uint(Tile tile)
```
ike709, line 56: image

ike709:
I'm not sure why these tests are failing still:
image

Robust.Server/Maps/GridSerializer.cs
```diff
@@ -69,7 +69,7 @@ internal sealed class MapChunkSerializer : ITypeSerializer<MapChunk, MappingData
         {
             for (ushort x = 0; x < chunk.ChunkSize; x++)
             {
-                var id = reader.ReadUInt16();
+                var id = reader.ReadInt32();
                 var flags = (TileRenderFlag)reader.ReadByte();
                 var variant = reader.ReadByte();
```
metalgearsloth: None of the old maps were saved with this in mind so you should bump the map format version and check if the version is earlier than latest then use the old method.

metalgearsloth (changes requested):

PR #4310, opened 2023-08-24
Add system to CompNetworkGenerator

metalgearsloth:
Robust doesn't global using this but content does so any automatic comp states on engine don't work.

@mirrorcult


DrSmugleaf (approved):

mirrorcult:
glug

PR #4311, opened 2023-08-24
Ignore deleted components while raising events.

PrPleGoo:
Fix an issue where removing a component and then triggering a subscription for that component on the same stack leads to an Exception.


metalgearsloth:
Are you deleting components in eventbus to trigger this?

PrPleGoo:
> Are you deleting components in eventbus to trigger this?

No, when handling the ToggleHealthBarsCommand in https://github.com/space-wizards/space-station-14/pull/17859 I either add or remove a component.
Adding the component is fine, but removing it triggers ComponentRemove.
This then raises a RefreshEquipmentHudEvent to get all the component that are relevant for the EquipmentHudSystem.
Despite the Component being deleted (I tried non-deferred delete as well) the subscription seems to still be alive in the MoveNext method.

Robust.Shared/GameObjects/EntityEventBus.Directed.cs
```diff
@@ -740,8 +740,7 @@ private struct SubscriptionsEnumerator
                     return false;
                 }
 
-                component = _entityManager.GetComponent(_uid, compType);
-                return true;
+                return _entityManager.TryGetComponent(_uid, compType, out component);
             }
```
ElectroJr: This will cause it to skip other subscriptions, which can introduce obscure & hard to diagnose bugs. It should either actually get the next event, or it should use GetComponentInternal, though I think it makes more sense to skip the event than to handle it with a component that has already been shut down and queued for deletion

PrPleGoo: Ended up using GetComponentInternal to get the potentially deleted components. If a component is deleted, it is just skipped.

Robust.Shared/GameObjects/EntityEventBus.Directed.cs
```diff
@@ -566,6 +569,9 @@ private void EntDispatch(EntityUid euid, Type eventType, ref Unit args, bool dis
 
             while (enumerator.MoveNext(out var component, out var reg))
             {
+                if (component.Deleted)
+                    continue;
+
                 if (reg.ReferenceEvent != byRef)
```
metalgearsloth: Isn't this going to cause inconsistent event handling between ordered and non-ordered events?

Ordered only checks comp deletion once up front which should never happen, but unordered is checking it every loop.

PrPleGoo: sorry, does the found result here have a lifespan longer than a tick? Edit: it doesn't

PrPleGoo: Oh you're worried about event handler A deleting the component for event handler B and in the unordered case it doesn't fire but in the ordered case it does?

PrPleGoo: Moved the check into the Func for ordered

ElectroJr (approved):

PR #4316, opened 2023-08-25
Remove lights component reference

metalgearsloth:
Requires https://github.com/space-wizards/space-station-14/pull/19531

I had to do some shit with textures but tried to minimise the damage.

Alternatively I just make it so PointLightSystem has a method to get the relevant component and we use that instead of moving textures around maybe? Keen for second thoughts.


Robust.Client/Graphics/Clyde/Clyde.RenderHandle.cs
```diff
@@ -149,7 +150,7 @@ public void SetScissor(UIBox2i? scissorBox)
                 Vector2 scale,
                 Angle? worldRot,
                 Angle eyeRot = default,
-                Direction? overrideDirection = null,
```
DrSmugleaf: The namespace I beg you

Robust.Client/Graphics/Clyde/Clyde.RenderHandle.cs
```diff
@@ -356,7 +357,7 @@ public override void DrawRect(UIBox2 rect, Color color, bool filled = true)
                     Vector2 scale,
                     Angle? worldRot,
                     Angle eyeRot = default,
-                    Direction? overrideDirection = null,
```
DrSmugleaf: Here too

DrSmugleaf (changes requested):

PR #4329, opened 2023-08-28
Delete EntityDeletedMessage

metalgearsloth:
There's already a C# event that engine uses something something 12 years no subscribers.


mirrorcult:
what wrong millennial zoomer? crying because you have 0 subscriber ?

PR #4332, opened 2023-08-28
Fix some multiline edit issues

juliangiebel:
This PR fixes an issue with multiline edits scrollbar margins and adds a multiline text submit keybind function.

### DIFF
--- a/Robust.Client/Input/EngineContexts.cs
+++ b/Robust.Client/Input/EngineContexts.cs
@@ -64,6 +64,7 @@ namespace Robust.Client.Input
             common.AddFunction(EngineKeyFunctions.TextWordDelete);
             common.AddFunction(EngineKeyFunctions.TextNewline);
             common.AddFunction(EngineKeyFunctions.TextSubmit);
+            common.AddFunction(EngineKeyFunctions.MultilineTextSubmit);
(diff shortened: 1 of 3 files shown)


Robust.Client/Input/EngineContexts.cs
```diff
@@ -64,6 +64,7 @@ public static void SetupContexts(IInputContextContainer contexts)
             common.AddFunction(EngineKeyFunctions.TextWordDelete);
             common.AddFunction(EngineKeyFunctions.TextNewline);
             common.AddFunction(EngineKeyFunctions.TextSubmit);
+            common.AddFunction(EngineKeyFunctions.MultilineTextSubmit);
```
mirrorcult, line 67: does this actually do anything? i dont see behavior added for it

juliangiebel: The behavior is in content. This PR changes the only thing that actually used submit on enter for multiline edits, `PaperBoundUserInterface`, to use the new keybind: space-wizards/space-station-14#19610
More things should use it. Hence the new keybind.

mirrorcult: aight

PR #4335, opened 2023-08-30
Localization support for any prototypes

ficcialfaint:
# About the PR
The PR adds ability to turn any prototype into `localized prototype` that supports localization in fluent file with custom prefix as entities do with `ent-` prefix. 
## How does it work?
There is now a new `ILocalizedPrototype` interface. You simply implement it in your prototype and add:
- `SetName` -- the default name you set in the prototype, in YAML files
- `SetDesc` -- same as `SetName` but for descriptions
- `Name` -- the localized name
- `Description` -- the localized description
- `CustomLocalizationPrefix` - the prefix that will be used in `.ftl` files. For example, entities use `ent-` prefix: `ent-BaseItem = base item`
Then you have to set `Name`'s and `Description`'s getters to `ILocalizationManager.GetPrototypeData<YourPrototype>(ID).Description` / `.Name`.

**Here is a small example for a localized prototype:**
```cs
[Prototype("construction")]
public sealed class ConstructionPrototype : IPrototype, ILocalizedPrototype, ISerializationHooks
{
    private ILocalizationManager _loc = default!;

    [DataField("name")]
    public string? SetName { get; private set; }

    [DataField("description")]
    public string? SetDesc { get; private set; }

    [ViewVariables]
    public string Name => _loc.GetPrototypeData<ConstructionPrototype>(ID).Name;

    [ViewVariables]
    public string Description => _loc.GetPrototypeData<ConstructionPrototype>(ID).Desc;

    public string CustomLocalizationPrefix { get; private set; } = "cnt"; // e.g. cnt-BaseCraft = cool name

    void ISerializationHooks.AfterDeserialization()
    {
        _loc = IoCManager.Resolve<ILocalizationManager>();
    }
}
```


Robust.Shared/Localization/LocalizationManager.Prototypes.cs
```diff
@@ -0,0 +1,123 @@
...
+    {
+        return CalcInheritingPrototypeLoc(typeof(T), prototypeId);
+    }
+
+    private PrototypeLocData CalcInheritingPrototypeLoc(Type kind, string prototypeId)
+    {
+        string? name = null;
+        string? desc = null;
+
+        foreach (var prototype in _prototype.EnumerateParents(kind, prototypeId, true))
```
ElectroJr, line 40: This doesn't recursively iterate up through a prototypes parents, this is just the direct parent list

Robust.Shared/Localization/LocalizationManager.Prototypes.cs
```diff
@@ -0,0 +1,123 @@
...
+    private PrototypeLocData CalcPrototypeLoc(Type kind, string prototypeId)
+    {
+        string? name = null;
+        string? desc = null;
+
+        // Return empty strings if no prototype found
+        if (!_prototype.TryIndex(kind, prototypeId, out var prototype))
+            return new PrototypeLocData("", "");
+
+        var lproto = (ILocalizedPrototype)prototype;
```
ElectroJr, line 95: try cast

Robust.Shared/Prototypes/IPrototype.cs
```diff
@@ -32,6 +33,20 @@ public interface IInheritingPrototype
...
+    public interface ILocalizedPrototype
+    {
+
+        [DataField("name")] public string? SetName { get; }
+
+        [DataField("description")] public string? SetDesc { get; }
+
+        [ViewVariables] public string Name { get; }
+
+        [ViewVariables] public string Description { get; }
```
ElectroJr, line 45: I might be wrong, but I don't think the data-field attributes do anything on interfaces?
I'm also not sure about whether VV works on interfaces.

ElectroJr: Also, these should have documentation comments

ElectroJr (changes requested):
The given example is bad because it uses ISerializationHooks which are obsolete.
The localized strings should just get resolved once after prototypes have been (re)loaded, which would also mean you could just get rid of the caching dictionary.

ficcialfaint:
I just noticed the review, I'll see what I can do, thanks

Robust.Shared/Localization/LocalizationManager.Prototypes.cs
```diff
@@ -0,0 +1,123 @@
...
+using Linguini.Bundle.Errors;
+using Robust.Shared.Prototypes;
+
+namespace Robust.Shared.Localization;
+
+internal sealed partial class LocalizationManager
+{
+    // Concurrent dict so that multiple threads "reading" .Name won't cause a concurrent write issue
+    // when the cache gets populated.
+    private readonly ConcurrentDictionary<string, PrototypeLocData> _prototypeCache = new();
```
ElectroJr, line 13: ~~Needs to be per-prototype~~
NVM see my earlier comment, just remove it altogether.

Robust.Shared/Localization/LocalizationManager.Prototypes.cs
```diff
@@ -0,0 +1,123 @@
...
+        _entityCache.Clear();
+    }
+
+    // Flush caches conservatively on prototype/localization changes.
+    private void OnPrototypesReloaded(PrototypesReloadedEventArgs args)
+    {
+        if (args.ByType.ContainsKey(typeof(EntityPrototype)))
+            FlushEntityCache();
+
+        FlushPrototypeCache();
```
ElectroJr, line 27: ~~This should enumerate over kinds and only flush the cache for prototypes that actually changed.
Also FlushPrototypeCache() is broken, its just a copy paste of FlushEntityCache()~~
NVM see my earlier comment, just remove it altogether.

LightVillet:
@ficcialfaint when

Ygg01:
This is probably too old to merge anymore. Close?

PR #4337, opened 2023-08-30
Added method to search nested containers for a specified component

chromiumboy:
Adds two new methods to SharedContainerSystem that allow you to recursively search the parent objects of a contained entity for a specified component

The first method returns the first matching component found, and the second returns a list of matching components found across all recursive parent entities

### DIFF
--- a/Robust.Shared/Containers/SharedContainerSystem.cs
+++ b/Robust.Shared/Containers/SharedContainerSystem.cs
@@ -226,28 +226,16 @@ namespace Robust.Shared.Containers
         public bool TryFindComponentOnEntityContainerOrParent<T>(
             EntityUid uid,
             out T? foundComponent,
-            MetaDataComponent? meta = null,
-            TransformComponent? xform = null,
-            EntityQuery<MetaDataComponent>? metas = null,
-            EntityQuery<TransformComponent>? xforms = null) where T : Component
+            EntityQuery<MetaDataComponent> metas,
+            EntityQuery<TransformComponent> xforms) where T : Component
         {
             foundComponent = null;
 
-            if (meta == null)
-            {
-                metas ??= EntityManager.GetEntityQuery<MetaDataComponent>();
-                meta = metas.Value.GetComponent(uid);
-            }
-
+            var meta = metas.GetComponent(uid);
             if ((meta.Flags & MetaDataFlags.InContainer) != MetaDataFlags.InContainer)
                 return false;
 
-            if (xform == null)
-            {
-                xforms ??= EntityManager.GetEntityQuery<TransformComponent>();
-                xform = xforms.Value.GetComponent(uid);
-            }
-
+            var xform = xforms.GetComponent(uid);
             if (!xform.ParentUid.Valid)
                 return false;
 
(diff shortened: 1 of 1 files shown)


Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -220,6 +220,83 @@ public bool IsEntityInContainer(EntityUid uid, MetaDataComponent? meta = null)
...
+        /// <summary>
+        ///     Finds the first instance of a component on the recursive parented containers that hold an entity  
+        /// </summary>
+        public bool TryFindComponentOnEntityContainerOrParent<T>(
+            EntityUid uid,
+            out T? foundComponent,
+            MetaDataComponent? meta = null,
+            TransformComponent? xform = null,
+            EntityQuery<MetaDataComponent>? metas = null,
+            EntityQuery<TransformComponent>? xforms = null) where T : Component
```
metalgearsloth: Just put the entityquery on the system and also require the user to pass in EntityQuery<T>

metalgearsloth (changes requested):

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -220,6 +226,70 @@ public bool IsEntityInContainer(EntityUid uid, MetaDataComponent? meta = null)
...
+                return foundComponents.Any();
+
+            if (entityQuery.Resolve(xform.ParentUid, ref foundComponent, false) && foundComponent != null)
+                foundComponents.Add(foundComponent);
+
+            if (TryFindComponentsOnEntityContainerOrParent(xform.ParentUid, entityQuery, out var extraComponents))
+                foundComponents = foundComponents.Concat(extraComponents).ToList();
+
+            return foundComponents.Any();
+        }
```
metalgearsloth: This is going to allocate a new list for every parent which is expensive, better to just pass it in and re-use it. This also avoids the additional Concat and ToList calls.

metalgearsloth (changes requested):

Robust.Shared/Containers/SharedContainerSystem.cs
chromiumboy: Fair point, the requested changes have been made

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -220,6 +226,64 @@ public bool IsEntityInContainer(EntityUid uid, MetaDataComponent? meta = null)
...
+                return foundComponents.Any();
+
+            if (!xform.ParentUid.Valid)
+                return foundComponents.Any();
+
+            if (entityQuery.Resolve(xform.ParentUid, ref foundComponent, false) && foundComponent != null)
+                foundComponents.Add(foundComponent);
+
+            return TryFindComponentsOnEntityContainerOrParent(xform.ParentUid, entityQuery, ref foundComponents);
+        }
```
metalgearsloth: You're setting the list to null every loop when it should just be passed in.

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -220,6 +226,64 @@ public bool IsEntityInContainer(EntityUid uid, MetaDataComponent? meta = null)
...
+        ///     Finds all instances of a component on the recursive parented containers that hold an entity  
+        /// </summary>
+        public bool TryFindComponentsOnEntityContainerOrParent<T>(
+            EntityUid uid,
+            EntityQuery<T> entityQuery,
+            ref List<T> foundComponents,
+            MetaDataComponent? meta = null,
+            TransformComponent? xform = null) where T : Component
+        {
+            T? foundComponent = null;
```
metalgearsloth: Remove this as it looks like you don't even use it.

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -220,6 +226,64 @@ public bool IsEntityInContainer(EntityUid uid, MetaDataComponent? meta = null)
...
+            return TryFindComponentOnEntityContainerOrParent(xform.ParentUid, entityQuery, ref foundComponent);
+        }
+
+        /// <summary>
+        ///     Finds all instances of a component on the recursive parented containers that hold an entity  
+        /// </summary>
+        public bool TryFindComponentsOnEntityContainerOrParent<T>(
+            EntityUid uid,
+            EntityQuery<T> entityQuery,
+            ref List<T> foundComponents,
```
metalgearsloth: Don't need to pass ref types by ref.

metalgearsloth (changes requested):

Robust.Shared/Containers/SharedContainerSystem.cs
chromiumboy: foundComponent is required for the entity query on line 281 and 282

Robust.Shared/Containers/SharedContainerSystem.cs
chromiumboy: Not sure what you mean here, don't we need to pass the ref for the recursion to work on line 284?

Robust.Shared/Containers/SharedContainerSystem.cs
metalgearsloth: Just use TryGetComponent instead of a dummy resolve.

PR #4350, opened 2023-09-04
Fix turn into Invalid direction

### LINKED ISSUE #4349: `Direction.TurnCw`/`TurnCcw` are bugged
image

TurnCw from south becomes invalid

Morb0:
Fixes #4349

### DIFF
--- a/Robust.Client/Utility/DirExt.cs
+++ b/Robust.Client/Utility/DirExt.cs
@@ -103,7 +103,7 @@ namespace Robust.Client.Utility
 
         public static Direction TurnCw(this Direction dir)
         {
-            return (Direction)(((int)dir - 1) % 8);
+            return (Direction)(((int)dir - 1 + 8) % 8);
         }
 
         public static Direction TurnCcw(this Direction dir)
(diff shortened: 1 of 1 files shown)


Robust.Client/Utility/DirExt.cs
```diff
@@ -103,7 +103,7 @@ public static RSIDirection Convert(this Direction dir, RSI.State.DirectionType t
 
         public static Direction TurnCw(this Direction dir)
         {
-            return (Direction)(((int)dir - 1) % 8);
+            return (Direction)(((int)dir - 1 + 8) % 8);
         }
```
metalgearsloth: Just do +7, also please make a test for it.

Morb0: For me correctly work only +8

metalgearsloth: do you mean - 1 + 8 or just doing + 7?

Morb0: I mean `(Direction)(((int)dir - 1 + 7) % 8)`

ElectroJr: I'm fairly sure sloth meant you should replace `- 1 + 8` with `+ 7`

Morb0: Okay, make sense

PR #4355, opened 2023-09-08
Add a command to hide replay UI

ShadowCommander:
Adds `replay_toggleui` which toggles the visibility on the ReplayWidget.


moonheart08 (dismissed):
Toolshed.

PR #4373, opened 2023-09-11
Allow handling by-value events by ref

DrSmugleaf:
Raising events by value is still supported.


deltanedas:
just curious whats the reason for preferring ref events? does raising a non ref event make it copy the data every time its handled or something

metalgearsloth:
> just curious whats the reason for preferring ref events? does raising a non ref event make it copy the data every time its handled or something

If it's a struct event it copies by value so you also can't get any data out of it.

deltanedas:
ok cool

PR #4409, opened 2023-09-16
More arch backports

metalgearsloth:
- ToPrettyString did a whacky cast that sometimes was annoying so I just made it more normal (i.e. pass the nullable value to a non-null method rather than doing a cast).
- Remove all of the _entTraitArray stuff and just use entitymanager's entityquery<T> directly.
- Sprinkly some speedholes AKA Unsafe.As<T> around the place.
- Fix some generic constraints on some entitymanager methods.

Content PR is https://github.com/space-wizards/space-station-14/pull/20241

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.Components.cs
+++ b/Robust.Shared/GameObjects/EntityManager.Components.cs
@@ -386,7 +386,7 @@ namespace Robust.Shared.GameObjects
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void RemoveComponent(EntityUid uid, IComponent component, MetaDataComponent? meta = null)
         {
-        	// Probably can't Unsafe.As<T> in case something on content overrides it.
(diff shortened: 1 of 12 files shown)


Robust.Shared/GameObjects/EntityManager.Components.cs
```diff
@@ -410,14 +411,14 @@ public bool RemoveComponentDeferred(EntityUid uid, ushort netId, MetaDataCompone
...
-            RemoveComponentDeferred((Component)comp, uid, false);
+            RemoveComponentDeferred(Unsafe.As<Component>(comp), uid, false);
             return true;
         }
 
         /// <inheritdoc />
         public void RemoveComponentDeferred(EntityUid owner, IComponent component)
         {
-            RemoveComponentDeferred((Component)component, owner, false);
+            RemoveComponentDeferred(Unsafe.As<Component>(component), owner, false);
```
moonheart08: Bad use of unsafe as, IComponent can be implemented by content.

PR #4410, opened 2023-09-16
Make AudioSystem stream Play methods public

Morb0:
Allow to play raw sound streams


Morb0:
Tested in game

metalgearsloth:
the ctor isn't public so how do you get these?

Morb0:
Implemented

PR #4421, opened 2023-09-17
Audio rework

metalgearsloth:
# Reviewing
It looks daunting but it's not that bad.
- There's an EAX preset file that's several thousand lines. This was regexed from OpenAL so should be fine.
- I moved all the audio out of clyde to its own namespace and cleaned up some duplicate classes. Most of this is the same (now called AudioManager) with some extra methods.
- IPlayingStream was culled and moved into AudioComponent.
- AudioSystem has some changes due to entity spawning and supporting other features like audio pausing / playback offsets.
- IResourceCache was moved to shared so server has access to audio to retrieve length. There's still a client local one now called IClientResourceCache which also has textures.

I'd recommend reviewing AudioComponent (and the EFX components) + AudioSystem predominantly.

Content PR is https://github.com/space-wizards/space-station-14/pull/20862

# Changes
- Removed SoundSystem static as this never seemed to be getting done.
- Removed some of the duplicate audio sources for headless.
- Removed IPlayingAudioStream and moved it all onto audio component.
- Cleaned up some of the duplication between AudioParams and IAudioSource so now audioparams data is networked and sets source directly.
- Split out IResourceCache from IResourceManager so it can be used in shared where the server wants it (i.e. getting audio length).
- Made audio into entities.
- Grid-local audio (no more WEOWEOWOEWE on FTL).
- Changed audio attenuation default to LinearDistanceClamped (this is what unreal engine uses as default afaik) as most people didn't seem to like inverse.
- Removed per-source audio attenuation and cleaned up ambient sounds around it (somewhat).
- Added support for pausing audio via entities.
- I reduced the z-axis offset for audio from -5 to -1 though may change this in future again.
- Z-axis is now considered for refdistance / maxdistance for audio. The attenuation won't be quite normal but this should work much better than master.
- Added support for audio to start at an offset if an audio entity comes into range so it won't always go from the start.
- Added rudimentary grid sounds so FTL won't be ear-blasting via global sound.
- Added preliminary EFX support via EAX reverb presets, these are also entities.

# Unsure
Filtered audio. If we play like a static sound for only a few entities I'm not sure how to handle this from a filtering perspective. For the moment this just won't work and we only handle predicted audio but like uhh... do we just have a set on audio when we spawn it and check that on compstate get or something?

TODO:
- [x] Figure out what's going on with ambient audio.
- [x] Implement 0% volume by stopping audio.
- [x] Fix replays by re-implementing events.
- [x] Add EFX support
- [x] Cleanup the API around playing streams.
~~- [ ] Remove the string versions of playing audio as it clutters the API.~~
~~- [ ] Remove AudioParams as it's already on SoundSpecifier and clutters methods.~~
- [x] Add filter-less versions of everything.
- [x] Add an audio overlay for easier debugging (will probably split this one out).
- [x] Turn audio into entities (this will probably hammer pvsoverride but will see how it goes).
- [x] Pause support, when they're turned into entities this should be extremely easy.
- [x] Fix whatever is going on with occlusion.
- [ ] Look at what ike sent me.


mirrorcult:
blessed

PR #4423, opened 2023-09-17
Add force ack threshold

ElectroJr:
Prevents very old client acks from slowing down the server. I don't know if this is the current cause of slowdowns, and even if it is, it doesn't actually fix whatever is causing the client to not send acks (or for disconnected clients to not disconnect)


metalgearsloth:
Problem seems to be fixed so gonna merge this too.

PR #4429, opened 2023-09-18
Use CollectionsMarshal in RobustTree

ElectroJr:
RobustTree isn't all that significant in terms of PVS performance, but might as well make it faster.

### DIFF
--- a/Robust.Server/GameStates/RobustTree.cs
+++ b/Robust.Server/GameStates/RobustTree.cs
@@ -1,6 +1,6 @@
 using System;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
 using Microsoft.Extensions.ObjectPool;
 using Robust.Shared.Utility;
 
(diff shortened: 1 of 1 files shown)


Robust.Server/GameStates/RobustTree.cs
```diff
@@ -89,56 +89,76 @@ public void Remove(T value, bool mend = false)
...
+            _parents.Add(child, parent);
+            return;
+        }
 
-            previousParentNode.Children?.Remove(existingNode.Value);
-            parentNode.Children!.Add(existingNode.Value);
-            _parents[child] = parent;
-            return existingNode;
+        ref var parentEntry = ref CollectionsMarshal.GetValueRefOrAddDefault(_parents, child, out var previousParentExists);
+        if (!previousParentExists || !_nodeIndex.TryGetValue(parentEntry!, out var previousParentNode))
```
ElectroJr, line 154: I'm fairly sure this was a bug. Previously it was `||` instead if `|| !`

PR #4433, opened 2023-09-19
Use CollectionsMarshal in DynamicTree

ElectroJr:
Also makes the nan check slightly faster Do we actually still need the nan check, or is it just a remnant from when nullspace just giving things nan coordinates?


metalgearsloth:
> Also makes the nan check slightly faster Do we actually still need the nan check, or is it just a remnant from when nullspace just giving things nan coordinates?

Yeah, I don't think NaNs are too common anymore, mainly if content stuffs something up.

PR #4435, opened 2023-09-19
Use CollectionsMarshal in AddComponentInternal()

ElectroJr:


metalgearsloth:
Arch conflicts my beloved.

PR #4441, opened 2023-09-21
Percentage range. Percentage and percentage range type serializers

ficcialfaint:
# Percentage range
Added a new **Percentage range** struct that keeps two values:
- Min value
- Max value
## What is it used for?
Can be used for getting a random discount of an item, for random chance of getting hit from something, as an example.
## What do Max and Min values have to be?
- Min value can’t be greater than Max value
- Min and Max values should be in 0-100% range:
  - 0.0f ≤ `Min/Max` ≤ 1.0f
  - 0% ≤ `Min/Max` ≤ 100%
## `NextFloat()`
This method is a part of an extension that is in `Robust.Shared.Random` that requires `IRobustRabdom` as the only parameter. What it does is basically `random.NextFloat(this.Min, this.Max`, where:
- `random` is the `IRobustRabdom` parameter
- `this` is a `PercentageRange` object
# Type serializers
## Percentage type serializer
Converts a YAML string into a float.
### Valid syntaxes
- `0.02`
  - would be `0.02` after reading it with serialization manager 
- `2%`
  - would also be `0.02` in C# code
## Percentage range serializer
Converts a YAML string into a `PercentageRange`.
### Valid syntaxes
- `0.05, 0.28`
  - `Min = 0.05` and `Max = 0.28`
- `20%, 75%`
  - `Min = 0.2` and `Max = 0.75`
- `12-25%`
  - `Min = 0.12` and `Max = 0.25`

### DIFF
--- a/Robust.Shared.Maths/PercentageRange.cs
+++ b/Robust.Shared.Maths/PercentageRange.cs
@@ -1,0 +1,134 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Robust.Shared.Maths;
+
+[Serializable]
+public struct PercentageRange : IEquatable<PercentageRange>
+{
+    /// <summary>
+    /// Minimum range value, e.g. 0.01f == 1%
+    /// </summary>
+    public float Min;
+
+    /// <summary>
+    /// Maximum range value, e.g. 1f == 100%
+    /// </summary>
+    public float Max;
+
+    public bool IsValid()
+    {
+        return IsValid(this, out _);
+    }
+
+    private bool IsValid([NotNullWhen(false)] out string? exception)
+    {
+        return IsValid(this, out exception);
+    }
+
+    private static bool IsValid(PercentageRange range, [NotNullWhen(false)] out string? exception)
+    {
+        return IsValid(range.Min, range.Max, out exception);
+    }
+
+    private static bool IsValid((float, float) range, [NotNullWhen(false)] out string? exception)
+    {
+        return IsValid(range.Item1, range.Item2, out exception);
+    }
+
+    private static bool IsValid(float min, float max, [NotNullWhen(false)] out string? exception)
+    {
+        exception = (min, max) switch
+        {
+            _ when min < 0 || max > 1 => $"Incorrect min and max values. Got: {min}, {max}; expected to be in 0.0 - 1.0 range",
+            _ when min > max => $"Incorrect min and max values. Min value is greater than max value: {min}>{max}",
+            _ => null
+        };
+
+        return exception == null;
+    }
+
+    public PercentageRange(float min = 0, float max = 1)
+    {
+        if (!IsValid(min, max, out var message))
+            throw new ArgumentException(message);
+
+        Min = min;
+        Max = max;
+    }
+
+    public PercentageRange((float, float) range) : this(range.Item1, range.Item2)
+    {
+    }
+
+    public PercentageRange(ReadOnlySpan<float> range)
+    {
+        if (range.Length != 2)
+            throw new ArgumentException($"An array with 2 items was expected, got: {range.Length}");
+
+        if (!IsValid(range[0], range[1], out var message))
+            throw new ArgumentException(message);
+
+        Min = range[0];
+        Max = range[1];
+    }
+
+    public PercentageRange(IReadOnlyList<float> range)
+    {
+        if (range.Count != 2)
+            throw new ArgumentException($"A list with 2 items was expected, got: {range.Count}");
+
+        if (!IsValid(range[0], range[1], out var message))
+            throw new ArgumentException(message);
+
+        Min = range[0];
+        Max = range[1];
+    }
+
+    public override bool Equals(object? obj)
+    {
+        return obj is PercentageRange other && Equals(other);
+    }
+
+    public override int GetHashCode()
+    {
+        return HashCode.Combine(Min, Max);
+    }
+
+    public bool Equals(PercentageRange range)
+    {
+        const double tolerance = 0.0001;
+        return Math.Abs(range.Max - Max) < tolerance && Math.Abs(range.Min - Min) < tolerance;
+    }
+
+    public static implicit operator (float, float)(PercentageRange r)
+    {
+        return (r.Min, r.Max);
+    }
+
+    public static explicit operator PercentageRange((float, float) r)
+    {
+        return new PercentageRange(r);
+    }
+
+    public static implicit operator (string, string)(PercentageRange r)
+    {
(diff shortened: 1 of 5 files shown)


Robust.Shared.Maths/PercentageRange.cs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public static bool operator !=(PercentageRange range1, PercentageRange range2)
+    {
+        return !(range1 == range2);
+    }
+
+    public override string ToString()
+    {
+        return $"{Min * 100}-{Max * 100}%";
```
PJB3005, line 132: Use percentage format specifiers: https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#standard-format-specifiers

Robust.Shared.Maths/PercentageRange.cs
```diff
@@ -0,0 +1,134 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Robust.Shared.Maths;
+
+[Serializable]
+public struct PercentageRange : IEquatable<PercentageRange>
```
PJB3005, line 8: You need to mark non-mutating functions in this struct as `readonly`.

Robust.Shared.Maths/PercentageRange.cs
```diff
@@ -0,0 +1,134 @@
...
+            _ => null
+        };
+
+        return exception == null;
+    }
+
+    public PercentageRange(float min = 0, float max = 1)
+    {
+        if (!IsValid(min, max, out var message))
+            throw new ArgumentException(message);
```
PJB3005, line 55: Add a `ThrowIfInvalid()` helper method for this.

Robust.Shared.Maths/PercentageRange.cs
```diff
@@ -0,0 +1,134 @@
...
+    public override int GetHashCode()
+    {
+        return HashCode.Combine(Min, Max);
+    }
+
+    public bool Equals(PercentageRange range)
+    {
+        const double tolerance = 0.0001;
+        return Math.Abs(range.Max - Max) < tolerance && Math.Abs(range.Min - Min) < tolerance;
+    }
```
PJB3005, line 103: This should use exact equality, not approximate.

Robust.Shared.Maths/PercentageRange.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    public static explicit operator PercentageRange((float, float) r)
+    {
+        return new PercentageRange(r);
+    }
+
+    public static implicit operator (string, string)(PercentageRange r)
+    {
+        return (((int)(r.Min * 100)).ToString() + '%', ((int)(r.Max * 100)).ToString() + '%');
+    }
```
PJB3005, line 118: This operator feels too specific to really be useful, I'd just axe it.

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Percents/PercentageRangeSerializer.cs
```diff
@@ -0,0 +1,40 @@
...
+            throw new InvalidMappingException("Could not parse percentage range");
+
+        return range.Value;
+    }
+
+    public DataNode Write(ISerializationManager serializationManager, PercentageRange value, IDependencyCollection dependencies,
+        bool alwaysWrite = false,
+        ISerializationContext? context = null)
+    {
+        return new ValueDataNode(value.ToString());
```
PJB3005, line 38: This needs to specify that it's formatting in culture-invariant mode.

Robust.Shared/Serialization/TypeSerializers/Implementations/Custom/Percents/PercentageSerializer.cs
```diff
@@ -0,0 +1,37 @@
...
+            throw new InvalidMappingException("Could not parse percentage");
+
+        return @float.Value;
+    }
+
+    public DataNode Write(ISerializationManager serializationManager, float value, IDependencyCollection dependencies,
+        bool alwaysWrite = false,
+        ISerializationContext? context = null)
+    {
+        return new ValueDataNode($"{value * 100}%");
```
PJB3005, line 35: Use percentage format specifier and format in culture-invariant mode.

Robust.Shared/Utility/PercentageSerializerUtility.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Robust.Shared.Utility;
+
+public sealed class PercentageSerializerUtility
+{
+    public static bool TryParse(string value, [NotNullWhen(true)] out float? result)
+    {
+        result = null;
+
+        if (float.TryParse(value, out var @float))
```
PJB3005, line 13: You are doing culture-sensitive parsing here.

Robust.Shared/Utility/PercentageSerializerUtility.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            if (@float is < 0 or > 1)
```
PJB3005, line 15: It seems plenty common for people to want to specify a value such as "200%". This feels too restrictive?

Robust.Shared/Utility/PercentageSerializerUtility.cs
```diff
@@ -0,0 +1,75 @@
...
+                return false;
+
+            result = @float;
+            return true;
+        }
+
+        if (value.Replace(" ", "")[^1] == '%' && int.TryParse(value[..^1], out var @int))
```
PJB3005, line 22: 1. If you want to cut out surrounding whitespace, you should use `Trim()`, not do it manually like this.
2. You aren't doing the parse on the trimmed result, so this won't work properly for untrimmed text.
3. Why can't I specify a value such as "10.5%"?

Robust.Shared/Utility/PercentageSerializerUtility.cs
```diff
@@ -0,0 +1,75 @@
...
+    public static bool TryParseRange(string value, [NotNullWhen(true)] out PercentageRange? result)
+    {
+        result = null;
+        var values = new List<float>();
+
+        var commas = value.Split(',');
+
+        if (commas.Length == 2)
+        {
+            foreach (var val in value.Split(','))
```
PJB3005, line 43: You are splitting this twice.

PJB3005 (changes requested):
I don't think there's any value in having multiple different syntaxes for the percentage ranges here. It adds a lot of complexity.

This entire thing needs way more comments.

PJB3005:
Derelict.

PR #4442, opened 2023-09-22
Infer `VVAccess.ReadWrite` for all datafields

mirrorcult:
reduces boilerplate since people usually readwrite these anyway

i cant think of a situation where it would be bad for a datafield to be readwriteable imo. maybe in the case of stuff needing to go through systems for some behavior but i'd still rather allow people to write to these always


moonheart08 (approved):

mirrorcult:
going to change this to infer vv readwrite for everything in debug mode

deltanedas:
that might lead to situations where someone thinks a field is rw when developing something then in a live round while making an admeme realises its some legacy component and nothing is writable

DrSmugleaf (approved):

PR #4444, opened 2023-09-24
Fix PVS bug and add new test

ElectroJr:
Fixes a PVS bug where dropped states due to lossy networking could sometimes result in the server never properly telling clients that an entity re-entered PVS range, which probably leads to something that looks like empty PVS chunks.

I don't think I've heard this bug reported, but I think that's just because its rare, not because it can't happen for some reason. 
AFAICT it requires a player to move around so that a stationary entity to leaves and then later re-enter PVS range while the client happens to drop all received states between when the entity left and re-entered PVS range. I've added a test that currently fails on master, though the way it handled packet loss is very artificial.

The current fix requires adding a new dictionary, which degrades PVS performance a bit. I think there might be some alternative way of fixing this bug that doesn't require that but I couldn't come up with anything.


Robust.UnitTesting/Server/GameStates/PvsReEntryTest.cs
```diff
@@ -0,0 +1,214 @@
...
+
+        for (int i = 0; i < 10; i++)
+        {
+            await server.WaitRunTicks(1);
+            await client.WaitRunTicks(1);
+        }
+
+        // Entity should be in PVS range, client should know about it:
+        Assert.That(timing.LastRealTick, Is.GreaterThan(lastRealTick));
+        Assert.That(meta!.Flags & MetaDataFlags.Detached, Is.EqualTo(MetaDataFlags.None));
```
ElectroJr, line 204: This is where the test would fail on master - the client still thinks the entity is outside of PVS range

PR #4450, opened 2023-09-26
Fix double contact deletion throwing

metalgearsloth:
Don't think this ever came up in content before.

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs
+++ b/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs
@@ -312,12 +312,17 @@ public abstract partial class SharedPhysicsSystem
 
     public void DestroyContact(Contact contact)
     {
+        // Don't recursive update or we're in for a bad time.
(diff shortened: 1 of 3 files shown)


Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs
```diff
@@ -312,12 +312,17 @@ internal static bool ShouldCollide(Fixture fixtureA, Fixture fixtureB)
...
+        if ((contact.Flags & ContactFlags.Deleting) != 0x0)
+            return;
+
         Fixture fixtureA = contact.FixtureA!;
         Fixture fixtureB = contact.FixtureB!;
         var bodyA = contact.BodyA!;
         var bodyB = contact.BodyB!;
         var aUid = contact.EntityA;
         var bUid = contact.EntityB;
+        contact.Flags |= ContactFlags.Deleting;
```
metalgearsloth, line 325: Flags get reset when it's returned to pool at the end of this method.

PR #4455, opened 2023-09-28
Allow force submitting line edits

mirrorcult:
required for content pr

### DIFF
--- a/Robust.Client/UserInterface/Controls/LineEdit.cs
+++ b/Robust.Client/UserInterface/Controls/LineEdit.cs
@@ -98,6 +98,11 @@ namespace Robust.Client.UserInterface.Controls
                 OnTextChanged?.Invoke(new LineEditEventArgs(this, _text));
         }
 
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -98,6 +98,11 @@ public void SetText(string value, bool invokeEvent = false)
                 OnTextChanged?.Invoke(new LineEditEventArgs(this, _text));
         }
 
+        public void ForceSubmitText()
```
DrSmugleaf, line 101: Should line 615 be calling this method too?

DrSmugleaf (dismissed):

Robust.Client/UserInterface/Controls/LineEdit.cs
mirrorcult: sure i can do that

PR #4471, opened 2023-10-05
Forbids EyeComponents from ever possessing a single visibility flag.

TemporalOroboros:
Forbids EyeComponents from ever setting the MSB on their visibility mask.
This makes setting the MSB on the visibility mask for _entities_ a safe way to prevent that entity from being broadcast to clients.
Required for https://github.com/space-wizards/space-station-14/pull/20744


ElectroJr:
I'm not sure why this is needed, why not just use a normal visibility layer? If something is carelessly setting the mask to ~0 somewhere that's a separate issue. I don't see any reason to create a "forbidden" layer that can never be seen, if an admin wants to view all hallucinations by giving themselves the layer they should be able to.

metalgearsloth:
Hallucinations should be using clientside entities for perf reasons so don't see a need for this and it adds a bit of cruft to pvs.

PR #4472, opened 2023-10-05
add type serializer for SortedSet

deltanedas:
shamelessly stolen from #3854

might be used in the future...


deltanedas:
ok i realise its 650 commits behind :trollface:

PR #4473, opened 2023-10-05
add SortedSetSerializer

deltanedas:
shamelessly stolen from https://github.com/space-wizards/RobustToolbox/pull/3854 and updated and put into its own serializer since its a rbtree not a list

might be used in the future...


deltanedas:
i think i did the test right but not sure how to make sure it actually ran

ok from non sorted version failing test it does run

PR #4474, opened 2023-10-06
Entity timers

mirrorcult:
basic idea. this is basically proof of concept
- remove old timercomponent API sucked and was old and thus no one used it,  replace with entity timers
- entity timers are entities which get parented to the entity they're relevant to
- entity timers store an event which they raise on their parent after a given delay
- goes through all the normal entity methods--delete it to end the timer, reparent it to change its target, so on. spawn has its own API.
- deleted when the entity its relevant to is deleted
- also handles repeating timers
- fully serializable handles saving/loading etc
- can be added to any entity, doesnt involve ensuring a comp, all good

draft bc:

todo
- [ ] phase out engine uses of timercomp. content uses later
- [ ] consider PVS like who should we be sending these to. also clientside entity timers
- [ ] add broadcast timers (entity timers which are in nullspace) opt-in you just have to accept that they wont be serializable, attach them to something if you want them to be you freak
- [ ] add `DeferEvent` helper which basically just sets the delay to minimum so it's raised next tick

### DIFF
--- a/Robust.Shared/GameObjects/Components/Timers/EntityTimerComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Timers/EntityTimerComponent.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using Robust.Shared.GameStates;
+using Robust.Shared.Serialization.Manager.Attributes;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
+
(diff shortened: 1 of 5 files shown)


moonheart08 (changes requested):
RepeatingTimer and Timer are identical aside from one having a repeat field. Couldn't a normal timer just be a repeating timer with a max repetition of 1?

PR #3963, opened 2023-04-23
Fix interp overlay

ElectroJr:
Makes the interpolation overlay properly account for rotation, and makes the boxes easier to see. Yellow boxes are the current / interpolated positions. Green boxes show the next-state positions.

Demo video shows various interactions on a spinning station with tickrate 5. This includes showing some issues related to https://github.com/space-wizards/space-station-14/issues/15676 & rotation lerping.

https://user-images.githubusercontent.com/60421075/233828616-149984dd-dfc6-4ed6-a7aa-d3ede689f38b.mp4

PR #4488, opened 2023-10-14
Transform interpolation fixes

### LINKED ISSUE #11446: RotateToFaceSystem / rotation lerping issues.
Something is wrong with the `RotateToFaceSystem` or rotation lerping. This affects player rotations, but is easiest to see for sprites that are not no-rot. A pretty reliable way of seeing this is to just control something like a crowbar, set tick rate to 30/20, and then start clicking:

https://user-images.githubusercontent.com/60421075/191630441-b41a5b18-57d4-45d6-8067-c49c5e613d46.mp4

### LINKED ISSUE #15676: Item dropping mispredicts
Probably 1/4 of the time when you drop an item you can see it flicker for a frame. It's probably some specific transform ordering code or the likes in engine but not entirely sure.

I'm also not entirely sure when it started but bisecting it shouldn't be hard.

ElectroJr:
Reworks how position & rotation interpolation/lerping works to fix a few bugs. This also adds some new transform system methods that take in an EntityUid and marks some old methods as obsolete. This PR should get rid of the issues with item dropping (fixes space-wizards/space-station-14/issues/11446), and the rotation lerping jumping all over the place (fixes space-wizards/space-station-14/issues/15676 and the bugs visible in the video in #3963).

Below is a before & after video video with the game running at 4 tps with time slowed down by 50%.
The inconsistent lerp speed is because some inputs are received towards the beginning of a tick, while others are received near the end. And the long delay when throwing an item is because it still isn't predicted.

Before:

https://github.com/space-wizards/RobustToolbox/assets/60421075/c7bdc793-9a57-40ce-a1be-9dae9d028422

After:

https://github.com/space-wizards/RobustToolbox/assets/60421075/053dfc5a-e6aa-4bc8-b164-b0defb9abf1a

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/TransformSystem.Component.cs
+++ b/Robust.Client/GameObjects/EntitySystems/TransformSystem.Component.cs
@@ -1,51 +1,41 @@
 using System.Numerics;
+using Robust.Client.Player;
 using Robust.Shared.GameObjects;
+using Robust.Shared.IoC;
 using Robust.Shared.Maths;
 
 namespace Robust.Client.GameObjects;
 
 public sealed partial class TransformSystem
 {
-    public override void SetLocalPosition(TransformComponent xform, Vector2 value)
+    public override void SetLocalPosition(EntityUid uid, Vector2 value, TransformComponent? xform = null)
     {
-        xform.PrevPosition = xform._localPosition;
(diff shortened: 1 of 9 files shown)


metalgearsloth:
thank god

PR #4491, opened 2023-10-15
UI prediction

metalgearsloth:
The only thing that really takes advantage of it atm is predicted UI opening / closing and storage interactions.

Still want to do a lot more cleanup.


metalgearsloth:
conflict nightmare.

PR #4493, opened 2023-10-15
Allow for ushort CVars

Cyberboss:


DrSmugleaf (changes requested):
This also needs to add a case for ushort values in the cvar command at https://github.com/space-wizards/RobustToolbox/blob/b6879869d6cbf4418b1286e4b31f1efdab5ca7e9/Robust.Shared/Configuration/ConfigurationCommands.cs#L130

DrSmugleaf (approved):

PR #4498, opened 2023-10-19
Add generic EntityUid, remove some usages of .Owner

DrSmugleaf:
Requires https://github.com/space-wizards/space-station-14/pull/21100

For the xeno antag pr btw


Robust.Shared/GameObjects/Entity.cs
```diff
@@ -0,0 +1,217 @@
+﻿using Robust.Shared.Utility;
+
+namespace Robust.Shared.GameObjects;
+
+public record struct Entity<T>
```
moonheart08, line 5: IEntity real?

moonheart08 (approved):
Approving the API changes, didn't read much else.

Robust.Shared/GameObjects/Entity.cs
DrSmugleaf: Return of the King

deltanedas:
a funny i found is that you cant pass `Entity<SomethingComponent>` to a function expecting `Entity<SomethingComponent?>` you need to (probably) use `(ent.Owner, ent.Comp)` which is annoying

deltanedas:
@DrSmugleaf is there a better way to do above

PR #4538, opened 2023-10-31
add contact to StartCollideEvent

notquitehadouken:
required for a pr in ss14 and having access to more collision data (like normals) is nice
keeping as draft because i might end up finding a solution that doesn't require an engine pr


notquitehadouken:
oh whoops i just realized i'm pr'ing from master

PR #4545, opened 2023-11-04
Allow Physics to get Bounding Boxes in Object Space

ChrisThirtle:
While defining bounding boxes in world-space is very valuable for determining if objects are in collision or should interact with each other, it is currently being misused in the game codebase. GetWorldAABB is being used to determine the true size of physics objects, but this produces inconsistent values as it means the size and shape of the bounding box varies with the rotation in world space.
By providing a way to define bounding boxes in object space, you can get a "canonical" bounding box for items that is based solely on the default assignment of their fixture shapes. For determining invariant values like total object size, this is invaluable.

I can cite the specific issues this causes if examples are wanted by anyone

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
+++ b/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
@@ -631,7 +631,7 @@ public partial class SharedPhysicsSystem
         if (!Resolve(uid, ref manager, ref body))
             return new Box2();
 
-        var transform = new Transform(Vector2.Zero, 0);
+        var transform = new Transform(0);
 
         var bounds = new Box2(Vector2.Zero, Vector2.Zero);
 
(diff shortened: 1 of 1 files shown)


metalgearsloth:
Isn't there already GetLocalAABB for this

ChrisThirtle:
> Isn't there already GetLocalAABB for this

No, there isn't. I just double checked both through the IDE and Github search. Grids have a LocalAABB variable, but I suspect that's not exactly right for individual physics objects. And I'm not sure how you'd navigate from PhysicsComponent to MapGridComponent either.

ChrisThirtle:
Okay I found _something_ that might be right after coming back around for another look. @metalgearsloth is EntityLookupSystem.GetAABB() what you were thinking of? That appears to be accessible and effective as far as I can tell naively testing it. Need to test it a bit further just to be sure but it _looks_ like I can force it to non-rotated values.

ChrisThirtle:
Double checked behavior, existing code does what's needed, was just fully unaware of its existence since it was in a completely different component than what was presently being used to calculate AABB for the known issue I was tracking down.

PR #4550, opened 2023-11-05
Add More `Entity<T>` query methods

ElectroJr:
Adds several `Resolve(ref Entity<T?>)` methods, and adds a new method to `EntityQuery<T>` that returns an `Entity<T>`.

### DIFF
--- a/Robust.Shared/GameObjects/EntitySystem.Resolve.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.Resolve.cs
@@ -34,44 +34,6 @@ namespace Robust.Shared.GameObjects
 
         /// <inheritdoc cref="Resolve{TComp}(Robust.Shared.GameObjects.EntityUid,ref TComp?,bool)"/>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        protected bool Resolve<TComp>(ref Entity<TComp?> entity, bool logMissing = true)
-            where TComp : IComponent
-        {
-            return Resolve(entity.Owner, ref entity.Comp, logMissing);
-        }
-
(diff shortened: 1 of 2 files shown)


Robust.Shared/GameObjects/EntitySystem.Resolve.cs
```diff
@@ -32,15 +32,53 @@ protected bool Resolve<TComp>(EntityUid uid, [NotNullWhen(true)] ref TComp? comp
             return found;
         }
 
-        /// <inheritdoc cref="Resolve{TComp}"/>
+        /// <inheritdoc cref="Resolve{TComp}(Robust.Shared.GameObjects.EntityUid,ref TComp?,bool)"/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        protected bool Resolve<TComp>(ref Entity<TComp?> entity, bool logMissing = true)
```
DrSmugleaf: This won't actually be enough for the IDE to think the component isn't nullable, so you would still need to suppress the warning.
There is a way to fix that which is putting the resolve method in Entity<T> and making it use the MemberNotNullWhen attribute. I'm not sure if that attribute works outside of the same type though.

DrSmugleaf (changes requested):

Robust.Shared/GameObjects/EntitySystem.Resolve.cs
ElectroJr: I was led astray by Rider's nullability analysis, which seems to have no issue with this, but build still fails.

Moving it into the entity class means you'd also need to pass in `IEntityManager`, and at that point it'd jus be easier to use `Resolve(ent.Owner, ref ent.comp)`. 

So I've just removed these functions. Maybe one day we'll have the technology.

DrSmugleaf (approved):

PR #4552, opened 2023-11-05
Add more implicit Entity<T> casts

ElectroJr:
Adds casts for converting `(EntityUid, T1)` tuple into a `Entity<T1?, T2?>`


Robust.Shared/GameObjects/Entity.cs
```diff
@@ -65,6 +65,11 @@ public Entity(EntityUid owner, T1 comp1, T2 comp2)
         return new Entity<T1, T2>(tuple.Owner, tuple.Comp1, tuple.Comp2);
     }
 
+    public static implicit operator Entity<T1?, T2?>((EntityUid Owner, T1? Comp1) tuple)
```
DrSmugleaf: These methods could be added to https://github.com/space-wizards/RobustToolbox/blob/master/Robust.UnitTesting/Shared/GameObjects/GenericEntityPrint.cs which makes it easier to adjust in the future for more or less generics

DrSmugleaf (approved):

Robust.Shared/GameObjects/Entity.cs
ElectroJr: I've updated it to also generate these casts.

DrSmugleaf: I just noticed these casts will result in the generic parameter of the new Entity struct being nullable regardless of whether or not it originally was in the source tuple.
The correct signature for this case would be:
`public static implicit operator Entity<T1, T2?>((EntityUid Owner, T1 Comp1) tuple)`
So that T1 is only returned as nullable when the source is.

Same for the rest, this one for example is correct:
`public static implicit operator Entity<T1?, T2?, T3?>(EntityUid owner)`

But `public static implicit operator Entity<T1?, T2?, T3?>((EntityUid Owner, T1 Comp1) tuple)`
Needs to be `public static implicit operator Entity<T1, T2?, T3?>((EntityUid Owner, T1 Comp1) tuple)` 

And `public static implicit operator Entity<T1?, T2?, T3?>((EntityUid Owner, T1 Comp1, T2 Comp2) tuple)`
Needs to be `public static implicit operator Entity<T1, T2, T3?>((EntityUid Owner, T1 Comp1, T2 Comp2) tuple)`

And so on. If the given T1 through T8 is nullable in the source, it will remain nullable in the produced cast.

DrSmugleaf (changes requested):
See https://github.com/space-wizards/RobustToolbox/pull/4552#discussion_r1386160274

Robust.Shared/GameObjects/Entity.cs
ElectroJr: This is kinda intentional. You can have either `implicit operator Entity<T1?, T2?>((EntityUid Owner, T1 Comp1) tuple)` OR ` implicit operator Entity<T1, T2?>((EntityUid Owner, T1 Comp1) tuple)` but not both because then you get `Member with the same signature is already declared` errors.

The rationale for preferring this one was that its more likely to be used for public methods. I.e., you are much more likely to have a public method that takes in an `Entity<T1?, T2?> ` instead of `Entity<T1, T2?>`.

I.e., I want to be able to do stuff like
```cs

public void DoStuff(Entity<TransformComponent?, MetaDataComponent?> ent)
{
}
public void A(EntityUid uid, TransformComponent xform)
{
    DoStuff((uid, xform));
}
```

ElectroJr: I also realise now that we probably also want the ability to cast `Entity<T1>` to `Entity<T1?,T2?>` and `Entity<T1,T2>` into `Entity<T1/2>`.

The latter would automatically support both nullable & non-nullable, but the former would also require choosing between all nullable or partially nullable.

I.e., so that you could use
```cs
public void A(Entity<TransformComponent?> ent)
{
    DoStuff(ent);
}
```

DrSmugleaf: See https://discord.com/channels/310555209753690112/900426319433728030/1172426838027550761

ElectroJr: I've updated the PR again, so now it casts tuples to `Entity<T1, T2?>` instead of `Entity<T1?, T2?>`. I've also added `Entity<T1>` -> `Entity<T1, T2?>` and `Entity<T1,T2>` -> `Entity<T1>` casts.

All of the casts just add or remove generics from the end of the list. I.e., while there is a `<T1,T2,T3>` -> `<T1,T2>` cast, there is no `<T1,T2,T3>` -> `<T1,T3>`, mainly because I couldn't be bothered adding code to create all of the combinations of possible casts.

Of the casts that I would like to work, there are still two that don't. E.g., this code snippet has two casts that error:

```cs
public void DoStuff(Entity<PhysicsComponent?, TransformComponent?> ent) { }
public void DoPartiallyNullableStuff(Entity<PhysicsComponent, TransformComponent?> ent) { }
public void DoOtherStuff(Entity<PhysicsComponent?> ent) { }
public void DoOtherStuffNotNull(Entity<PhysicsComponent> ent) { }

public void A()
{
    EntityUid uid = default;
    PhysicsComponent comp = default!;
    Entity<PhysicsComponent> ent = new(uid, default!);
    Entity<PhysicsComponent?> nEnt = new(uid, null);

    DoStuff((uid, comp));
    DoStuff((uid, null));
    DoPartiallyNullableStuff((uid, comp));

    DoStuff(ent); // <-- Errors
    DoStuff(nEnt);
    DoPartiallyNullableStuff(ent);

    Entity<PhysicsComponent, MetaDataComponent> ent2 = new(uid, default!, default!);
    Entity<PhysicsComponent?, MetaDataComponent?> nEnt2 = new(uid, null, null);
    DoOtherStuff(ent2); // <-- Errors
    DoOtherStuffNotNull(ent2);
    DoOtherStuff(nEnt2);
}
```
It doesn't seem to like casting `Entity<T1,T2>` ->  `Entity<T1?>` and  `Entity<T1>` -> `Entity<T1?, T2?>`.
I think that's probably fine? In most situations I assume people will be working with all nullable components anyways.

PR #4558, opened 2023-11-07
Fix transform test warnings

metalgearsloth:


Robust.UnitTesting/Server/GameObjects/Components/Transform_Test.cs
```diff
@@ -347,16 +347,16 @@ public void ParentRotationRoundingErrorTest()
...
             //NOTE: Yes, this does cause a non-zero error
 
             // Assert
 
             Assert.Multiple(() =>
             {
-                Assert.That(MathHelper.CloseToPercent(oldWpos.X, newWpos.Y));
-                Assert.That(MathHelper.CloseToPercent(oldWpos.Y, newWpos.Y));
+                Assert.That(MathHelper.CloseToPercent(oldWpos.X, newWpos.Y, 0.0001f));
+                Assert.That(MathHelper.CloseToPercent(oldWpos.Y, newWpos.Y, 0.0001f));
```
metalgearsloth, line 359: TransformComponent.WorldPosition uses a matrix transform which has hardcoded angles and avoids precision issues whereas GetWorldPosition just uses position + parent rotation instead. I'm making a separate pr with Vector2d backing GetWorldPosition.

PR #4563, opened 2023-11-09
Fix stack overflow error on planet station

JerryImMouse:
Planets threw an error when you are trying to make them stations(adding BecomesStationComponent) and forcing round start with forcemapped planet station

Just removed the recursion in func. to fix that

### DIFF
--- a/Robust.Shared/Map/Enumerators/GridTileEnumerator.cs
+++ b/Robust.Shared/Map/Enumerators/GridTileEnumerator.cs
@@ -26,33 +26,33 @@ public struct GridTileEnumerator
     }
 
     public bool MoveNext([NotNullWhen(true)] out TileRef? tileRef)
-    {
-        if (_index == _chunkSize * _chunkSize)
         {
-            if (!_chunkEnumerator.MoveNext())
+            while (true)
             {
-                tileRef = null;
-                return false;
+                if (_index == _chunkSize * _chunkSize)
+                {
+                    if (!_chunkEnumerator.MoveNext())
+                    {
+                        tileRef = null;
+                        return false;
+                    }
+
+                    _index = 0;
+                }
+
+                var (chunkOrigin, chunk) = _chunkEnumerator.Current;
+                var x = (ushort) (_index / _chunkSize);
+                var y = (ushort) (_index % _chunkSize);
+                var tile = chunk.GetTile(x, y);
+                _index++;
+
+                if (_ignoreEmpty && tile.IsEmpty)
+                    continue;
+
+                var gridX = x + chunkOrigin.X * _chunkSize;
+                var gridY = y + chunkOrigin.Y * _chunkSize;
+                tileRef = new TileRef(_gridUid, gridX, gridY, tile);
+                return true;
             }
-
-            _index = 0;
         }
-
-        var (chunkOrigin, chunk) = _chunkEnumerator.Current;
-
-        var x = (ushort) (_index / _chunkSize);
-        var y = (ushort) (_index % _chunkSize);
-        var tile = chunk.GetTile(x, y);
-        _index++;
-
-        if (_ignoreEmpty && tile.IsEmpty)
-        {
-            return MoveNext(out tileRef);
-        }
-
-        var gridX = x + chunkOrigin.X * _chunkSize;
-        var gridY = y + chunkOrigin.Y * _chunkSize;
-        tileRef = new TileRef(_gridUid, gridX, gridY, tile);
-        return true;
-    }
 }


ElectroJr:
Changing this method to avoid excessive recursion is probably a good idea, but I'd still like to know what caused this first. I can't seem to reproduce it. E.g., running these commands seems to load fine in release mode without any stack overflow:

- `/mapping 100 /Maps/aspid.yml`
- `/planet 100 Grasslands`
- `/savemap 100 aspid.yml`
- move the file and overwrite the original aspid.yml
- `/forcemap Aspid`
- `/restartroundnow`

Was this being caused by grids having many empty chunks? Because otherwise I'm not sure how you could've gotten a stack overflow here (if that's what the error was?). A single empty chunk should lead to at most 256 recursions (Chunksize^2), And empty chunks existing in the first place is a bug that should already be fixed with the next engine version.

metalgearsloth:
Is the infinite loop you described just the debug asserts due to corvax's engine version? I checked the code against rider's formatting suggestion and it seems to be identical which suggests nothing actually changed? Also it still spams the PVS debug errors when I try running the map too.

JerryImMouse:
> Changing this method to avoid excessive recursion is probably a good idea, but I'd still like to know what caused this first. I can't seem to reproduce it. E.g., running these commands seems to load fine in release mode without any stack overflow:
> 
> * `/mapping 100 /Maps/aspid.yml`
> * `/planet 100 Grasslands`
> * `/savemap 100 aspid.yml`
> * move the file and overwrite the original aspid.yml
> * `/forcemap Aspid`
> * `/restartroundnow`
> 
> Was this being caused by grids having many empty chunks? Because otherwise I'm not sure how you could've gotten a stack overflow here (if that's what the error was?). A single empty chunk should lead to at most 256 recursions (Chunksize^2), And empty chunks existing in the first place is a bug that should already be fixed with the next engine version.

I'll try this
We get into infinite loop because of planets which have a lot of tiles, and recursion just gets into inifinte loop

JerryImMouse:
> Is the infinite loop you described just the debug asserts due to corvax's engine version? I checked the code against rider's formatting suggestion and it seems to be identical which suggests nothing actually changed? Also it still spams the PVS debug errors when I try running the map too.

This infinite loop causes NavMapSystem(RefreshGrid func.) to enumerate all tiles on planet what leads to Stack overflow.
And it changes, station planet is loadable now and navmapsystem also works correctly

JerryImMouse:
Oh it seems it is really only corvax build problem. Sorry for this PR

Vonsant:
> Was this being caused by grids having many empty chunks? Because otherwise I'm not sure how you could've gotten a stack overflow here (if that's what the error was?). A single empty chunk should lead to at most 256 recursions (Chunksize^2), And empty chunks existing in the first place is a bug that should already be fixed with the next engine version.


Hello,
I would like to request a reconsideration of this PR. Despite aligning with the official build specifications, I'm encountering the same stack overflow error that was present in Corvax's build. The map indeed has numerous empty chunks, deliberately placed to separate "stations" from "points of interest" by thousands of coordinates to avoid accidental encounters. If the forthcoming engine update resolves these issues, allowing the server to launch without crashing, it would validate the extensive effort invested in the map's creation. Nevertheless, I hope this method can still be evaluated, as the precise engine modifications are yet to be seen. I'm attaching the map file and an adapted prototype for your perusal. I apologize for bursting in like this, but it's of great importance to me.
[corvax_mirage2.zip](https://github.com/space-wizards/RobustToolbox/files/13325507/corvax_mirage2.zip)

Redict:
I approve changes. The main modification involves replacing the call, with a loop. This change is advantageous because it helps prevent stack overflow problems when dealing with recursion particularly if there are numerous empty tiles and the "_ignoreEmpty" condition is true. Utilizing a loop enhances the clarity and predictability of the control flow, which is generally considered practice in processes, like this one. The remaining logic remains mostly unchanged indicating that the methods functionality is preserved while achieving efficiency and safety.

ElectroJr:
I'm not opposed to the changes, I just wanted to figure out why it was happening instead of just making changes that don't actually fix the root cause of the issue.

> We get into infinite loop because of planets which have a lot of tiles, and recursion just gets into inifinte loop

@JerryImMouse Are you actually getting an **infinite** loop, or just a large (but finite) number of empty chunks. Because an infinite loop means there's some other issue and this PR wouldn't fix it.

> The map indeed has numerous empty chunks, deliberately placed to separate "stations" from "points of interest" by thousands of coordinates to avoid accidental encounters.

@Vonsant Having places far apart on a planet shouldn't require actually having any empty chunks, empty chunks shouldn't exist. There was a bug that was fixed in #4529 that meant that planet maps wouldn't delete their empty chunks. There's now a separate issue where the map can't be loaded because empty chunks were serialized to yaml but I'll fix that in another PR (#4565). With those fixes I seem to be able to load the map fine even without this PR.

JerryImMouse:
> Are you actually getting an infinite loop, or just a large (but finite) number of empty chunks. Because an infinite loop means there's some other issue and this PR wouldn't fix it.

@ElectroJr I cant say for sure, Vonsant just made me to open this PR(and to reopen it too). All i can say is that this error caused by empty chunks and this change fixes it, and all works correctly. Tbh, this is not my change at all, I just opened this PR for Vonsant

PR #4573, opened 2023-11-16
Minor PVS stuff

metalgearsloth:
The array one is probably most risky in terms of re-using old data but all the array entries are getting cleared every tick (PVSData is only used inside of PVS)

### DIFF
--- a/Robust.Server/GameStates/PvsSystem.cs
+++ b/Robust.Server/GameStates/PvsSystem.cs
@@ -112,7 +112,6 @@ internal sealed partial class PvsSystem : EntitySystem
         RobustTree<NetEntity> tree)?> _previousTrees = new();
 
     private readonly HashSet<(int visMask, IChunkIndexLocation location)> _reusedTrees = new();
-    private readonly List<(int, IChunkIndexLocation)> _previousIndices = new();
 
     private EntityQuery<EyeComponent> _eyeQuery;
     private EntityQuery<MetaDataComponent> _metaQuery;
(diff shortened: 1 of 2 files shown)


Redict:
The changes seem to be aimed at optimization, and the overall approach looks reasonable. I approve.

Robust.Server/GameStates/PvsSystem.cs
```diff
@@ -569,17 +577,28 @@ private void OnMapCreated(MapChangedEvent e)
             _reusedTrees.Add(chunks[i]);
         }
 
-        var previousIndices = _previousTrees.Keys.ToArray();
-        for (var i = 0; i < previousIndices.Length; i++)
+        _previousIndices.Clear();
+        _previousIndices.EnsureCapacity(_previousTrees.Count);
```
ElectroJr: I don't think `_previousIndices` is event required? This can just be a simple iteration over the `_previousTrees` elements as it only removes entries without adding them.

Robust.Server/GameStates/PvsSystem.cs
```diff
@@ -1327,12 +1346,14 @@ private EntityUid[] GetSessionViewers(ICommonSession session)
...
-        var viewers = new HashSet<EntityUid>();
+        var viewers = _uidSetPool.Get();
         if (session.AttachedEntity != null)
             viewers.Add(session.AttachedEntity.Value);
 
         viewers.UnionWith(session.ViewSubscriptions);
-        return viewers.ToArray();
+        var viewersArray = viewers.ToArray();
+        _uidSetPool.Return(viewers);
+        return viewersArray;
```
ElectroJr: It's probably better to just not use a hashset here at all. I.e., just using something like

```cs
if (session.AttachedEntity is not {} local)
    return session.ViewSubscriptions.ToArray();

var viewers = new EntityUid[session.ViewSubscriptions.Count + 1];
viewers[0] = local;
int i = 0;
DebugTools.Assert(!session.ViewSubscriptions.Contains(local));
foreach (var ent in session.ViewSubscriptions)
{
    viewers[i++] = ent;
}

return viewers;
```

ElectroJr:
See https://github.com/metalgearsloth/RobustToolbox/pull/5

ElectroJr (approved):

PR #4579, opened 2023-11-19
PVS pooling

metalgearsloth:
- Pool a lot more stuff with PVS under the assumption once GameState / MsgState are serialized it's safe to fuck with it. I tried using Collections.Pooled although then you need IList and it becomes a lot more scrunkly with serializing it so I just kept using ObjectPool for now. Maybe someday it'd be better to swap as the pooled data could be re-used across more systems although this PVS data allocs to shit on live.
- Move _deletionHistory out of PVSCollection to PvsSystem with the goal of nuking RobustTree someday.
- Use 1 pool for RobustTree rather than pool per tree, at least until we kill it.

API is skrunkly though Leviathan is barely edging out over 30ms so maybe this will help it just stay under for a bit.

### DIFF
--- a/Robust.Server/GameStates/ServerGameStateManager.cs
+++ b/Robust.Server/GameStates/ServerGameStateManager.cs
@@ -109,7 +109,7 @@ namespace Robust.Server.GameStates
                 .Select(x => x.Key)
                 .Select(x => $"{x.Name} ({_entityManager.ToPrettyString(x.AttachedEntity)})");
 
-            var deletedNents = new List<NetEntity>();
+            var deletedNents = _nentListPool.Get();
             _pvs.GetDeletedEntities(GameTick.First, deletedNents);
 
             shell.WriteLine($@"Current tick: {_gameTiming.CurTick}
@@ -118,6 +118,8 @@ Deletion history size: {deletedNents.Count}
 Actual oldest: {ack}
 Oldest acked clients: {string.Join(", ", players)}
 ");
+            
+            _nentListPool.Return(deletedNents);
         }
 
         private void ResetParallelism()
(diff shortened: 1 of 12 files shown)


Robust.Shared/GameObjects/EntityState.cs
```diff
@@ -33,6 +33,23 @@ public EntityState(NetEntity netEntity, NetListAsArray<ComponentChange> changedC
...
+
+        public EntityState Clone()
+        {
+            HashSet<ushort>? netComps = null;
+
+            if (NetComponents?.Count == 0)
+            {
+                netComps = new HashSet<ushort>();
+                netComps.UnionWith(NetComponents);
+            }
```
ElectroJr, line 45:
```suggestion
            if (NetComponents != null)
                netComps = new HashSet<ushort>(NetComponents);
```

Robust.Server/GameStates/ServerGameStateManager.cs
```diff
@@ -399,10 +423,33 @@ private struct PvsData
...
+                }
+
+                _nentListPool.Return(deletions);
+                _entStatePool.Return(entStates);
+                _playerStatePool.Return(playerStates);
+
+                if (leavePvsDispose)
+                {
+                    _nentListPool.Return(leftPvs);
+                }
```
ElectroJr, line 452: This should probably be moved out of the enclosing if block

Robust.Shared/Network/INetManager.cs
```diff
@@ -86,7 +86,8 @@ public interface INetManager
         /// </summary>
         /// <param name="message">Message to send.</param>
         /// <param name="recipient">Channel to send the message over.</param>
-        void ServerSendMessage(NetMessage message, INetChannel recipient);
+        /// <returns>True if the message is sent and can be disposed.</returns>
+        bool ServerSendMessage(NetMessage message, INetChannel recipient);
```
ElectroJr, line 90: Does this only exist for integration tests? If yes, then there should probably be comments in ServerGameStateManager explaining this.

ElectroJr (changes requested):

ElectroJr:
I'm killing robusttree at the moment, and there are already quite a few conflicts.

PR #4580, opened 2023-11-19
Fix some grid setting asserts

metalgearsloth:
If I add MapGrid to an existing map these trip.


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -353,8 +353,7 @@ public void SetGridId(EntityUid uid, TransformComponent xform, EntityUid? gridId
         if (!xform._gridInitialized || xform._gridUid == gridId || xform.GridUid == uid)
             return;
 
-        DebugTools.Assert(!HasComp<MapGridComponent>(uid));
-        DebugTools.Assert(gridId == null || HasComp<MapGridComponent>(gridId));
+        DebugTools.Assert(gridId == null || HasComp<MapGridComponent>(gridId) || !HasComp<MapGridComponent>(gridId));
```
ElectroJr: The HasComps in the assert do nothing now. Shouldn't it be something like
```suggestion
		DebugTools.Assert(!HasComp<MapGridComponent>(uid) || gridId == uid);
        DebugTools.Assert(gridId == null || HasComp<MapGridComponent>(gridId));
```
Or does that still fail somehow?

Also it seems like `SetGridIdRecursive` can just be merged into `SetGridId`

ElectroJr (approved):

PR #4582, opened 2023-11-19
Pool MsgState streams

metalgearsloth:
Drastically reduces heap allocations. Content PR is https://github.com/space-wizards/space-station-14/pull/21796 due to the memorystream change for reading.

Old:
image

New:
image

I'm not sure what an ideal memory solution looks like for serialization but a 3MB reduction per second on the physics tumbler is enormous and this is easy to add.


Robust.Shared/Network/Messages/MsgState.cs
```diff
@@ -33,34 +32,31 @@ public override void ReadFromBuffer(NetIncomingMessage buffer, IRobustSerializer
...
             var uncompressedLength = buffer.ReadVariableInt32();
             var compressedLength = buffer.ReadVariableInt32();
-            MemoryStream finalStream;
+            using var finalStream = RobustMemoryManager.GetMemoryStream();
 
             // State is compressed.
             if (compressedLength > 0)
             {
-                var stream = buffer.ReadAlignedMemory(compressedLength);
+                var stream = RobustMemoryManager.GetMemoryStream();
```
ElectroJr, line 40: using?

Robust.Shared/Network/Messages/MsgState.cs
```diff
@@ -33,34 +32,37 @@ public override void ReadFromBuffer(NetIncomingMessage buffer, IRobustSerializer
...
+                buffer.ReadAlignedMemory(finalStream, uncompressedLength);
             }
 
             serializer.DeserializeDirect(finalStream, out State);
-            finalStream.Dispose();
+
+            // This is just a sanity check the the pool isn't leaking.
+#if DEBUG
+            Array.Clear(finalStream.GetBuffer());
+#endif
```
ElectroJr: the sanity check isn't actually checking anything, is there meant to be an assert here or something? Same goes for the other `#if DEBUG` block

metalgearsloth: So when I tried pooling I would get weird stuttering which seemed like there were issues but I never got it with the recyclableIO so I wasn't sure if the memory was getting re-used or what.

PR #4598, opened 2023-11-23
Use spans for sprite draw-order

metalgearsloth:
Also ref var. This really only matters if you're REALLY zoomed out and realistically we probably need batching or something idk, either way this is still a significant amount of render time but at least I can open F3 now.

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.HLR.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.HLR.cs
@@ -259,7 +259,9 @@ namespace Robust.Client.Graphics.Clyde
             var worldOverlays = GetOverlaysForSpace(OverlaySpace.WorldSpaceEntities);
 
             var spriteSystem = _entityManager.System<SpriteSystem>();
-            GetSprites(mapId, viewport, eye, worldBounds, out var indexList);
+            ProcessSpriteEntities(mapId, viewport, eye, worldBounds, _drawingSpriteList);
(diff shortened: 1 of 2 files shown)


Robust.Client/Graphics/Clyde/Clyde.HLR.cs
```diff
@@ -259,7 +259,9 @@ private void DrawEntities(Viewport viewport, Box2Rotated worldBounds, Box2 world
             var worldOverlays = GetOverlaysForSpace(OverlaySpace.WorldSpaceEntities);
 
             var spriteSystem = _entityManager.System<SpriteSystem>();
-            GetSprites(mapId, viewport, eye, worldBounds, out var indexList);
+            ProcessSpriteEntities(mapId, viewport, eye, worldBounds, _drawingSpriteList);
+            Span<int> indexList = stackalloc int[_drawingSpriteList.Count];
```
ElectroJr, line 263: Is stackalloc actually safe to use here? or can it handle a ~100k int array if someone zooms very far out?

metalgearsloth: At least for me it was fineTM, maybe OD it's not? I'll just keep the ref var change as at 100k entities it might be bad.

Maybe it's better to just store arrays for each comparison type up front and just iterate them instead.

DrSmugleaf: You can also use a ternary where it uses stackalloc if the count is small enough, and an array otherwise.

PR #4612, opened 2023-11-27
Inline _world.IsAlive

metalgearsloth:
Profiles the same in microbenchmarks but we can cut a method out at least.


metalgearsloth:
1.

PR #4619, opened 2023-11-27
Only change the offset if has already started

Uriende:
If the audio start is in the future we'll get a negative value and AL didn't like that 
I assume it will fix this lovely problem I currently have

```
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
[ERRO] clyde.oal: [PlaybackPosition:295] AL error: InvalidValue
```

### DIFF
--- a/Robust.Client/Audio/AudioSystem.cs
+++ b/Robust.Client/Audio/AudioSystem.cs
@@ -163,7 +163,7 @@ public sealed partial class AudioSystem : SharedAudioSystem
         // If audio came into range then start playback at the correct position.
         var offset = (Timing.CurTime - component.AudioStart).TotalSeconds % GetAudioLength(component.FileName).TotalSeconds;
 
-        if (offset != 0)
+        if (offset > 0)
         {
             component.PlaybackPosition = (float) offset;
         }


Robust.Client/Audio/AudioSystem.cs
```diff
@@ -163,7 +163,7 @@ private void OnAudioStartup(EntityUid uid, AudioComponent component, ComponentSt
         // If audio came into range then start playback at the correct position.
         var offset = (Timing.CurTime - component.AudioStart).TotalSeconds % GetAudioLength(component.FileName).TotalSeconds;
 
-        if (offset != 0)
+        if (offset > 0)
         {
             component.PlaybackPosition = (float) offset;
         }
```
metalgearsloth, line 169: This seems to bandaid the underlying bug rather than fix it because if the offset is negative that should be fixed.

metalgearsloth: If it's really because the audio is just barely clipped in range from pvs then just not playing it should be sufficient but need to be sure.

Uriende: I'm way less versed in the codebase in general. Also noticed that even after I added this change I still cannot hear any alert sounds for example because the volume ends up 0 or negative. 

The offset could be negative because of clock differences maybe?
I assume CurTime is coming from the local machine. Is AudioStart comes from the server in any way or also local?

Uriende: Well. I see that is not the case entirely. So how can it even happen that Audiostart is in the future? Confusing

metalgearsloth: I think this might be due to rounding on timespans for ticks and state handling so this might be fine.

PR #4633, opened 2023-11-28
Add cvar for audio

metalgearsloth:
Lets us turn it off for tests or whatever.


Robust.Shared/CVars.cs
```diff
@@ -584,6 +584,8 @@ protected CVars()
         public static readonly CVarDef<string> WatchdogBaseUrl =
             CVarDef.Create("watchdog.baseUrl", "http://localhost:5000", CVar.SERVERONLY);
 
+
+
         /*
```
metalgearsloth:
```suggestion
        /*
```

PR #4639, opened 2023-11-29
Change midi volume to gain

metalgearsloth:
Easier to work with and I wanted to make the cvars consistent.


Robust.Client/Audio/Midi/IMidiManager.cs
```diff
@@ -19,7 +19,7 @@ public interface IMidiManager
     /// <summary>
     ///     Volume, in db.
     /// </summary>
```
metalgearsloth:
```suggestion
    /// <summary>
    ///     Gain of audio.
    /// </summary>
```

PR #4640, opened 2023-11-29
Possible fix to some audio issues

Kelrak:
Whenever the gain in is set here to 0 https://github.com/space-wizards/RobustToolbox/blob/f5a2a710f0b4555620674509607293602cf0685a/Robust.Client/Audio/Sources/BaseAudioSource.cs#L175-L196
(As it does on audio startup here)
https://github.com/space-wizards/RobustToolbox/blob/f5a2a710f0b4555620674509607293602cf0685a/Robust.Client/Audio/AudioSystem.cs#L168-L169

_gain would be set to 0 and any next attempt to set the gain would encounter a division by zero and leave the volume muted. This is a proposed fix to the problem but I'm not sure if it would leave the audio stronger then it should. Would appreciate if more people could test this locally.  
When I tested it on my machine it fixed announcement sounds and ambient sounds that didn't work before.
My only guess on why some sounds are working and some aren't is because parallelism? But I'm not savvy enough in the codebase to figure out if that's logical or not

### DIFF
--- a/Robust.Client/Audio/Sources/BaseAudioSource.cs
+++ b/Robust.Client/Audio/Sources/BaseAudioSource.cs
@@ -190,7 +190,7 @@ internal abstract class BaseAudioSource : IAudioSource
                 AL.GetSource(SourceHandle, ALSourcef.Gain, out var priorGain);
                 priorOcclusion = priorGain / _gain;
                 if (Double.IsNaN(priorOcclusion)) {
-                    priorOcclusion = 1f
+                    priorOcclusion = 1f;
                 }
             }
 
--- a/Robust.Client/Audio/Sources/BaseAudioSource.cs
+++ b/Robust.Client/Audio/Sources/BaseAudioSource.cs
@@ -188,7 +188,7 @@ internal abstract class BaseAudioSource : IAudioSource
             if (!IsEfxSupported)
             {
                 AL.GetSource(SourceHandle, ALSourcef.Gain, out var priorGain);
-                priorOcclusion = _gain == 0 ? 1f : priorGain / _gain;
+                priorOcclusion = _gain == 0 ? 0.5f : priorGain / _gain;
             }
 
             _gain = value;
--- a/Robust.Client/Audio/Sources/BaseAudioSource.cs
+++ b/Robust.Client/Audio/Sources/BaseAudioSource.cs
@@ -189,6 +189,9 @@ internal abstract class BaseAudioSource : IAudioSource
             {
                 AL.GetSource(SourceHandle, ALSourcef.Gain, out var priorGain);
                 priorOcclusion = priorGain / _gain;
+                if (Double.IsNaN(priorOcclusion)) {
+                    priorOcclusion = 1f
+                }
             }
 
             _gain = value;
--- a/Robust.Client/Audio/Sources/BaseAudioSource.cs
+++ b/Robust.Client/Audio/Sources/BaseAudioSource.cs
@@ -188,10 +188,7 @@ internal abstract class BaseAudioSource : IAudioSource
             if (!IsEfxSupported)
             {
                 AL.GetSource(SourceHandle, ALSourcef.Gain, out var priorGain);
-                priorOcclusion = priorGain / _gain;
-                if (Double.IsNaN(priorOcclusion)) {
-                    priorOcclusion = 1f;
-                }
+                priorOcclusion = _gain == 0 ? 1f : priorGain / _gain
             }
 
             _gain = value;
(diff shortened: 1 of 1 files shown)


Robust.Client/Audio/Sources/BaseAudioSource.cs
```diff
@@ -189,6 +189,9 @@ public float Gain
             {
                 AL.GetSource(SourceHandle, ALSourcef.Gain, out var priorGain);
                 priorOcclusion = priorGain / _gain;
+                if (Double.IsNaN(priorOcclusion)) {
+                    priorOcclusion = 1f;
+                }
```
VasilisThePikachu: Webedit pr, did you test your changes?

Also contributing from master branch, this is discouraged.

Kelrak: Yeah. Tested locally but my computer is fully configured for my work user so this is easier, 
I'll create a branch for next time. Didn't occur to me but fair enough

metalgearsloth: Just check if _gain is 0 first rather than doing NaN checks.

Robust.Client/Audio/Sources/BaseAudioSource.cs
```diff
@@ -188,7 +188,7 @@ public float Gain
             if (!IsEfxSupported)
             {
                 AL.GetSource(SourceHandle, ALSourcef.Gain, out var priorGain);
-                priorOcclusion = priorGain / _gain;
+                priorOcclusion = _gain == 0 ? 1f : priorGain / _gain;
```
metalgearsloth:
```suggestion
                priorOcclusion = _gain == 0 ? 0f : priorGain / _gain;
```

I think this is more appropriate because priorGain and _gain would be close to 0.

Kelrak: That would leave us in the same position wouldn't it? As the gain we set would remain 0. 
I don't know what happens on other OS's when you set the gain to NaN. But I guess it either remains the same or changes to something that's not 0?

metalgearsloth: This is going to go from full volume to potentially 0 volume which seems worse than going the other way, no?

Kelrak: Yeah. But even with 0 I still get the problem:
```
time: $1701817152194 hashcode 42355014 handle 2408  Old gain: 1 for file /Audio/Announcements/announce.ogg new gain 1 value recived: 0.63095737
time: $1701817152195 after 0.63095737
time: $1701817152195 hashcode 42355014 handle 2408  Old gain: 0.63095737 for file /Audio/Announcements/announce.ogg new gain 1 value recived: 0
time: $1701817152195 after 0
[DEBG] system.audio: Starting audio /Audio/Announcements/announce.ogg on entity 1418 with Volume -2 and gain 0 and pos <0, 0>
time: $1701817152198 hashcode 42355014 handle 2408  Old gain: 0 for file /Audio/Announcements/announce.ogg new gain 0 value recived: 0.63095737
time: $1701817152198 after 0
time: $1701817152208 hashcode 42355014 handle 2408  Old gain: 0 for file /Audio/Announcements/announce.ogg new gain 0 value recived: 0.63095737
time: $1701817152208 after 0
```
Here are some of the logs for announce for example. (Ignore the mess I was getting lost)
Initially gain is set according to the volume IIRC which is -2. 
after that we set it to 0.
From that point on since we take the prior gain which is 0 and multiply by what ever new value we got. We still get 0 (Or NaN)
That's at least how it is on my machine. I'm really confused on how it works on other machines. Or I'm missing the point entirely somehow

metalgearsloth: multiplying 0 by 0 is still going to be 0, this piece of code would never hit NaN as it's not dividing by 0 unless your system is deeply broken.

Kelrak: That's my point. So if we set prior occlusion to zero gain would remain zero forever since it always multiples by it when setting it

Kelrak: When I wrote NaN there that is what happens currently. Without any changes. Since after setting the gain to zero once dividing by prior gain goes to NaN and it gets stuck that way.

Kelrak: Essentially what happens on other systems when
```
            AL.Source(SourceHandle, ALSourcef.Gain, _gain * priorOcclusion);
```
Is set to NaN? 
Guessing the library handles that differently?

Kelrak: Asked my friend to check on her system (linux) And it seems that (ALC_EXT_EFX) is just not supported on mac so that is mostly the only case it even goes into that condition. I guess if we want to leave it pretty close like 0.2 or 0.5? I'll do some local tests and see how that sounds

Kelrak: I think 0.5 Sounds okay

PR #4643, opened 2023-11-30
Add MIDI semaphore

metalgearsloth:
This is bandaid A which may be a terrible idea. Realistically some of Render could be moved over to the parallel update job but I think I'd need to refactor MIDIs to be able to do it.


deltanedas:
maybe could fix midi death lag, merg

metalgearsloth:
Midi probably needs a look-over at some point and vera is long since retired so sometime next year or the year after I might make this better.

PR #4660, opened 2023-12-04
Update map physics after events

### LINKED ISSUE #21747: Gravity anomaly broke movement
## Description
On Lizard right now: after being pushed around by the gravity anomaly I could suddenly no longer move, reconnecting allowed me to move a little bit but very quickly turns into serious rubberbanding.

Any attempt to move would just rotate my character, no actual movement would happen. I could jump up on a table but not down, disassembling the table did not let me move.

It was not possible to push me around, and I couldn't push others as they would just teleport back.

Me, the smile, and a technical assistant got stuck in this state.

image
The red circles show where we were stuck

tgrkzus:
This should fix: https://github.com/space-wizards/space-station-14/issues/21747

See: https://discord.com/channels/310555209753690112/560845886263918612/1181065068557893773

TLDR is:
- Throw system wakes up physics when it's being slept
- Which can sometimes cause the map to lose track of what is actually awake

This is a bit of a bandaid fix, but it makes nested events with physics a bit more reasonable.

(My first contribution to engine, so let me know if I'm messed up something wack)

### DIFF
--- a/Robust.Shared/Physics/PhysicsWakeEvent.cs
+++ b/Robust.Shared/Physics/PhysicsWakeEvent.cs
@@ -1,11 +1,16 @@
 using Robust.Shared.GameObjects;
 using Robust.Shared.Physics.Components;
 
-namespace Robust.Shared.Physics
-{
-    [ByRefEvent]
-    public readonly record struct PhysicsWakeEvent(EntityUid Entity, PhysicsComponent Body);
+namespace Robust.Shared.Physics;
 
-    [ByRefEvent]
-    public readonly record struct PhysicsSleepEvent(EntityUid Entity, PhysicsComponent Body);
-}
+[ByRefEvent]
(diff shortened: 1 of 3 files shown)


tgrkzus:
If you want to test the fix for the linked issue, you can use this branch. https://github.com/tgrkzus/space-station-14/tree/test-horn-lol

And use the debug horn (with Z) in your hand. To throw yourself into a wall a couple times.

metalgearsloth:
Still triggering the assert for me (at least when taken in isolation).

tgrkzus:
> Still triggering the assert for me (at least when taken in isolation).

What assert? Confused what you mean by this sorry.

metalgearsloth:
> > Still triggering the assert for me (at least when taken in isolation).
> 
> What assert? Confused what you mean by this sorry.

The assert about movebuffer.

I'm going to make a tweak to this pr and should fix the throwing issue I think.

Issue #4578, opened 2023-11-19
DestroyContacts debug assert can get triggered.


ElectroJr:
Happened during an explosion. I'm not sure what caused it and how to reproduce it, but it might have to do with the fact that `DestroyContacts()` was called while already in the process of destroying contacts?
 
![rider64_7N4bqyKCAu](https://github.com/space-wizards/RobustToolbox/assets/60421075/38b3ca78-2c42-4fd2-bf22-30908928f58f)

metalgearsloth:
See https://github.com/space-wizards/space-station-14/issues/21734 for the content issue.

CollisionWake maybe needs some guard against calling it during collision contacts or the likes, not sure.

PR #4666, opened 2023-12-05
Bandaid physics contacts getting modified during collision

metalgearsloth:
This is somewhat naughty and box2d has world locks around preventing this but I never got around to implementing them so this is the bandaid we get for the next 6 months.


ElectroJr:
Does this fix #4578?

PR #4669, opened 2023-12-07
pvs5 (pfives)

metalgearsloth:
Draft but I want to work on it over xmas.

Goal is to combine current pvs and last pvs (anchoring). Roughly 3/4 of all entities in range (approx. 2k on average) are anchored and we used to have a fast path for this where we can make assumptions that let us skip iterating large numbers of entities for compstates and just for hierarchy traversal in general.

Additionally, by splitting we can potentially bump the pvs range for static bodies and fix several audio and lighting issues as well as help alleviate client stuttering (remember when the client didn't stutter on entity popin?) without also requiring the much-needed entity init refactor.

The main advantage current pvs brought was removing unnecessary dictionary.FindValues and let us hit 200 pop though the dictionary creep addressing more and more bugs has hit it particularly hard.

Anchoring was pretty fast but at the time we were still doing per-player work for entitylookups which limited it. Now with chunking players in range it should be able to get much faster.

My rough outline on discord https://discord.com/channels/310555209753690112/1153246232064569394/1181760500137197708

### DIFF
--- a/Robust.Server/GameStates/PvsSystem.Ack.cs
+++ b/Robust.Server/GameStates/PvsSystem.Ack.cs
@@ -1,4 +1,7 @@
 using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Win32.SafeHandles;
 using Robust.Shared.GameObjects;
 using Robust.Shared.Player;
 using Robust.Shared.Threading;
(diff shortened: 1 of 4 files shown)


metalgearsloth:
https://discord.com/channels/310555209753690112/1153246232064569394/1182450101004812298 should be possible without a full refactor and we can probably reduce dictionary counts anyway.

PR #4670, opened 2023-12-07
Compiler error for AutoNetworkedField fields in a component without a AutoGenerateComponentState attribute 

### LINKED ISSUE #4590: There is no compiler error for AutoNetworkedField fields in a component without a AutoGenerateComponentState attribute
There is one for the other way around, when using AutoGenerateComponentState but no fields have AutoNetworkedField.

ficcialfaint:
Closes #4590

Will need a content update once merged as it already fails (or should a content PR be merged before this one? iunno)
image


Robust.Shared.CompNetworkGenerator/ComponentNetworkGenerator.cs
```diff
@@ -325,6 +327,31 @@ public void Execute(GeneratorExecutionContext context)
...
+                            a.AttributeClass.Equals(fieldAttr, SymbolEqualityComparer.Default));
+
+                        if (attribute == null)
+                            continue;
+
+                        var msg = "Field is marked with [AutoNetworkedField], but its class has no [AutoGenerateComponentState] attribute.";
+                        context.ReportDiagnostic(
+                            Diagnostic.Create(
+                                new DiagnosticDescriptor(
+                                    "RXN0007",
```
ficcialfaint, line 346: I'm unsure that's the correct ID. I couldn't find any info on what those mean

Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Robust.Shared.GameObjects
 {
-    [NetworkedComponent, Access(typeof(SharedPointLightSystem))]
+    [NetworkedComponent, AutoGenerateComponentState, Access(typeof(SharedPointLightSystem))]
```
ficcialfaint: Seems like marking this component with `AutoGenerateComponentState` made the tests fail with this error:
`System.InvalidOperationException : Component is not a valid reference type: SharedPointLightComponent`

Robust.Shared.CompNetworkGenerator/ComponentNetworkGenerator.cs
DrSmugleaf: That's an identifier for different errors, which are made up on the spot as you create a DiagnosticDescriptor.
You can find out what they mean by checking the message that they are used with.
In this case, its only other usage is to report `Component is marked with [AutoGenerateComponentState], but has no valid members marked with [AutoNetworkedField].`
Sometimes they are reused between similar errors, so `RXN0007` as you have it here is correct.

Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
DrSmugleaf: I think that's because the component is abstract.
You can remove it and the usages of AutoNetworkedField below as those weren't doing anything before, and PointLightComponent already has methods to get and handle its own state.

DrSmugleaf (approved):

PR #4672, opened 2023-12-07
PrototypeAttribute name parameter is optional now

ficcialfaint:
With this PR you can now do the following:
```cs
[Prototype]  // No need to do [Prototype("something")]
public sealed partial class SomethingPrototype : IPrototype
{
}
```
If the name isn't specified in the attribute, it takes the class name, converts it into camelCase and gets rid of `Prototype` at the end (This is now required for classes that don't specify custom prototype name via attribute parameter; similar like comps do)


DrSmugleaf (approved):

Robust.Shared/Prototypes/PrototypeManager.cs
```diff
@@ -745,6 +745,16 @@ public void RegisterIgnore(string name)
...
         void IPrototypeManager.RegisterType(Type type) => RegisterKind(type);
 
+        static string CalculatePrototypeName(Type type)
+        {
+            const string prototype = "Prototype";
+            if (!type.Name.EndsWith(prototype))
+                throw new InvalidPrototypeNameException($"Prototype {type} must end with the word Prototype");
+
+            var name = type.Name.AsSpan();
+            return $"{char.ToLowerInvariant(name[0])}{name[1..][..^prototype.Length]}";
```
DrSmugleaf:
```suggestion
            return $"{char.ToLowerInvariant(name[0])}{name[1..^prototype.Length]}";
```

DrSmugleaf (changes requested):

Robust.Shared/Prototypes/PrototypeManager.cs
```diff
@@ -904,4 +916,12 @@ private void OnReload(PrototypesReloadedEventArgs args)
             return data;
         }
     }
+
+    [Virtual]
+    public class InvalidPrototypeNameException : Exception
```
DrSmugleaf: [Virtual] is in case you intend for this class to be inherited. I don't think there's much of a reason to let content inherit this exception type.

```suggestion
    public sealed class InvalidPrototypeNameException : Exception
```

DrSmugleaf (changes requested):

Robust.Shared/Prototypes/PrototypeManager.cs
ficcialfaint: Oh I just copied it from component. InvalidComponentName exception is virtual, perhaps it makes sense to just inherit it?

DrSmugleaf: The two should be separate though I don't know why the component one is virtual either

ficcialfaint:
thank you

PR #4680, opened 2023-12-09
Set priorGain to 0 if gain is 0

metalgearsloth:
I think silencing the audio for a frame rather than setting it to an arbitary mid is better.

### DIFF
--- a/Robust.Client/Audio/Sources/BaseAudioSource.cs
+++ b/Robust.Client/Audio/Sources/BaseAudioSource.cs
@@ -188,7 +188,8 @@ internal abstract class BaseAudioSource : IAudioSource
             if (!IsEfxSupported)
             {
                 AL.GetSource(SourceHandle, ALSourcef.Gain, out var priorGain);
-                priorOcclusion = _gain == 0 ? 0.5f : priorGain / _gain;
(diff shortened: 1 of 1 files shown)


Kelrak:
Checked locally again. This just leaves the sound muted forever. @metalgearsloth 
Because if `priorGain` is 0 it's just 0 / whatever so forever zero.
Adding the debug logs for `/Audio/Announcements/announce.ogg` for example
```
Current value: 0.63095737 Current _gain: 0.63095737 current priorOcclusion: 1
Current value: 0 Current _gain: 0 current priorOcclusion: 1
Current value: 0.63095737 Current _gain: 0.63095737 current priorOcclusion: 0
Current value: 0.63095737 Current _gain: 0.63095737 current priorOcclusion: 0
Current value: 0.63095737 Current _gain: 0.63095737 current priorOcclusion: 0
Current value: 0.63095737 Current _gain: 0.63095737 current priorOcclusion: 0
Current value: 0.63095737 Current _gain: 0.63095737 current priorOcclusion: 0
Current value: 0.63095737 Current _gain: 0.63095737 current priorOcclusion: 0
Current value: 0.63095737 Current _gain: 0.63095737 current priorOcclusion: 0
Current value: 0.63095737 Current _gain: 0.63095737 current priorOcclusion: 0
```

PR #4681, opened 2023-12-09
Fix colour space translation feedback loop(#3839)

### LINKED ISSUE #3839: Colour slider rounding errors.
Rounding errors will happen with HSV/RGB conversion, but the current colour slider code is just garbage as updating one of the sliders cause the values of the other sliders to change. Its also very prone to stack overflows.

https://user-images.githubusercontent.com/60421075/224458044-e53481ad-dd00-44ee-b1fa-c062b4b515aa.mp4

ghost:
Now colour picker wont always use RGB color value for its sliders, but the internally saved currently value of currently set colour space.

Fix #3839


ghost:
Has the Build & Test / build (macos-latest) just randomly failed, or my change fails the unit tests?

PR #4686, opened 2023-12-10
Remove non-networked comps from NetComponents dictionary

ElectroJr:
Alternative to #4685

### DIFF
--- a/Robust.Shared/Prototypes/EntityPrototype.cs
+++ b/Robust.Shared/Prototypes/EntityPrototype.cs
@@ -194,13 +194,16 @@ namespace Robust.Shared.Prototypes
         }
 
         internal static void LoadEntity(
-            EntityPrototype? prototype,
-            EntityUid entity,
+            Entity<MetaDataComponent> ent,
             IComponentFactory factory,
             IEntityManager entityManager,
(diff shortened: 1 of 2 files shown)


Robust.Shared/Prototypes/EntityPrototype.cs
```diff
@@ -209,8 +212,11 @@ void ISerializationHooks.AfterDeserialization()
                         continue;
 
                     var fullData = context != null && context.TryGetComponent(name, out var data) ? data : entry.Component;
+                    var compReg = factory.GetRegistration(name);
+                    EnsureCompExistsAndDeserialize(entity, compReg, factory, entityManager, serManager, name, fullData, ctx);
 
-                    EnsureCompExistsAndDeserialize(entity, factory, entityManager, serManager, name, fullData, context as ISerializationContext);
+                    if (!entry.Component.NetSyncEnabled && compReg.NetID is {} netId)
+                        meta.NetComponents.Remove(netId);
```
ElectroJr, line 219: This is the actual fix, which is basically the same as in #4685, except it did it inside of `EnsureCompExistsAndDeserialize`

metalgearsloth (approved):

PR #4697, opened 2023-12-11
Expose IFileDialogManager.GetSaveFileName to content

DrSmugleaf:
@PJB3005 

This is the part of SaveFile (which is already sandboxed) that gets the path from the user-selected file.
As far as I know this should be safe to sandbox as well.
You can't cast the stream returned by SaveFile to FileStream then get its name because that is not sandboxed.

### DIFF
--- a/Robust.Client/UserInterface/DummyFileDialogManager.cs
+++ b/Robust.Client/UserInterface/DummyFileDialogManager.cs
@@ -17,5 +17,10 @@ namespace Robust.Client.UserInterface
         {
             return Task.FromResult<(Stream fileStream, bool alreadyExisted)?>(null);
         }
+
+        public Task<string?> GetSaveFileName(FileDialogFilters? filters = null)
+        {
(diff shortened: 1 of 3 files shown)


PJB3005:
I don't get it, how can content do anything with this? It can't subsequently write to the file because you're not giving it a stream right?

DrSmugleaf:
> I don't get it, how can content do anything with this? It can't subsequently write to the file because you're not giving it a stream right?

This is for the mapping editor specifically since the client needs to get the path but the server is the one that needs to save the map data to that path through MapLoaderSystem (when both are running on the same computer like with prototype hot reloading).

https://github.com/space-wizards/RobustToolbox/assets/10968691/a9a809c3-1f81-4f9c-ac8e-d7638aa8a3af

DrSmugleaf:
Replaced by https://github.com/space-wizards/RobustToolbox/pull/4702

PR #4698, opened 2023-12-11
Fix TileEdgeOverlay drawing over grids

metalgearsloth:
Need to flush the render queue after every grid. This is done lazily so if the tileedgeoverlay never runs we won't flush, this is somewhat similar to what we have to do with ProcessSpriteEntities.

image

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using OpenToolkit.Graphics.OpenGL4;
+using Robust.Shared.Enums;
 using Robust.Shared.GameObjects;
 using Robust.Shared.Graphics;
 using Robust.Shared.IoC;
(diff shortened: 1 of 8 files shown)


Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
```diff
@@ -72,6 +81,26 @@ private void _drawGrids(Viewport viewport, Box2Rotated worldBounds, IEye eye)
                     GL.DrawElements(GetQuadGLPrimitiveType(), datum.TileCount * GetQuadBatchIndexCount(), DrawElementsType.UnsignedShort, 0);
                     CheckGlError();
                 }
+
+                requiresFlush = false;
+
+                foreach (var overlay in GetOverlaysForSpace(OverlaySpace.WorldSpaceGrids))
```
ElectroJr: GetOverlaysForSpace() should be moved outside of the for loop

Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
```diff
@@ -72,6 +81,26 @@ private void _drawGrids(Viewport viewport, Box2Rotated worldBounds, IEye eye)
...
+                {
+                    if (overlay is not IGridOverlay iGrid)
+                    {
+                        continue;
+                    }
```
ElectroJr: Maybe log an error in OverlayManager if someone attempts to add a non-IGridOverlay with the WorldSpaceGrids flag?

ElectroJr (changes requested):

Robust.Client/Graphics/Clyde/Clyde.GridRendering.cs
metalgearsloth: I'll probably shove it in a hashset as my worry was it spamming the error every frame for every grid.

PR #4704, opened 2023-12-12
Should resolve some error spam in non EFX systems

Kelrak:
Should resolve the illegal enum error that appears sometimes. Only saw it locally when switching maps.
Still need to check locally and maybe I could figure out some of the other errors as well


Kelrak:
Didn't manage to replicate the other errors but it fixed the illegal enum that only shows up on map reload once.

PR #4712, opened 2023-12-13
.NET 8

PJB3005:
There were two issues with .NET 8: serv3 breaks on some nullable issue, and publishing is broken.

The serv3 issue is because System.Runtime.CompilerServices.NullableAttribute was moved into the BCL. This means the code to find the attribute broke. The lucky thing about this is that this is an API-level break, and versions of the game built against .NET 7 will keep working even when the launcher updates to .NET 8!

The publishing issue seems to just be because Microsoft.NET.ILLink.Tasks was moved out of the base SDK. Pulling it in from NuGet makes it work.

### DIFF
--- a/.github/workflows/build-docfx.yml
+++ b/.github/workflows/build-docfx.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v3.2.0
       with:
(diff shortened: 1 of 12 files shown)


VasilisThePikachu (approved):
idk looks good

PJB3005:
Don't merge until the launcher is updated ofc.

PR #4713, opened 2023-12-15
Make `TryGetEntity` protected

ElectroJr:
Makes it consistent with all the other entity system proxy methods


EmoGarbage404:
is there a reason these are protected to begin with?

metalgearsloth:
> is there a reason these are protected to begin with?

External callers should be using entitymanager if we ever need to update.

ElectroJr:
> is there a reason these are protected to begin with?

In addition to what sloth said, it also makes it easier to find relevant methods when using a system dependency. E.g., without this the auto-completion suggestions for `ContainerSystem.TryGet....` would be swamped with results that are completely unrelated to the container method that you are looking for. If you need to use those methods outside of a system, you should just use IEntityManager directly

PR #4714, opened 2023-12-15
More detailed audio logging

Kelrak:
I am totally not sure if this is the best way to do this mostly because I'm less familiar with the codebase and openal in general. But maybe adding this context to the error logs would help a little to figure out what is the cause or just to see if that could help figuring out how to replicate the stuff locally.


PJB3005:
Stale

PR #4719, opened 2023-12-16
Use FrozenDictionary for prototype instances

ElectroJr:
Also adds a new method to `IPrototypeManager` that returns a `FrozenDictionary` for a specific kind. I.e., if you have to look up many entity prototypes, you can now directly retrieve a `FrozenDictionary<string, EntityPrototype>` instead of having to use the prototype manager index methods, which requires two dictionary lookups and a cast from `IPrototype`.

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.cs
@@ -1,7 +1,9 @@
 ﻿using System;
+using System.Collections.Frozen;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Reflection;
 using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


Robust.Shared/Prototypes/PrototypeManager.cs
```diff
@@ -366,6 +375,19 @@ protected void ReloadPrototypes(IEnumerable<ResPath> filePaths)
...
+        {
+            var st = new Stopwatch();
+            st.Start();
+            foreach (var kind in kinds)
+            {
+                kind.Freeze();
+            }
+
+            // fun fact: Sawmill can be null in tests????
+            Sawmill?.Info($"Freezing prototype instances took {st.Elapsed.TotalMilliseconds:f2}ms");
```
ElectroJr: Takes about 50-70 ms on startup

mirrorcult:
bwased

mirrorcult (approved):

Robust.Shared/Prototypes/PrototypeManager.cs
PJB3005: Verbose or debug log level please.

PJB3005 (changes requested):

PR #4725, opened 2023-12-16
Add `EntitySystem<TComp>` class

ElectroJr:
Adds a new abstract class that caches a `EntityQuery<TComp>` and provides some helper methods that use the query.


PJB3005:
Idk this seems kinda questionable. Seems like it'd encourage poorly organizing entity systems to use this.

I feel like if we want this, a source gen to automatically generate the boilerplate would be nicer.

PR #4726, opened 2023-12-16
Use frozen collections in component factory and `EntityManager._entTraitDict`

ElectroJr:
This is a breaking change because it changes some public methods


PJB3005 (changes requested):
Please document the breaking changes in `RELEASE-NOTES.md`

metalgearsloth:
The frozendict for _entTraitDict will conflict with arch which is probably not worth it.

ElectroJr:
> The frozendict for _entTraitDict will conflict with arch which is probably not worth it.

I'm gonna solve the conflicts, so it should be fine to merge for now

PR #4727, opened 2023-12-17
Freeze event bus dictionaries

ElectroJr:
This PR Freezes some of the event bus dictionaries. This should make both event raising and component adding/removing somewhat faster. The PR extends the existing `_subscriptionLock` restrictions and only freezes dictionaries after the lock has been enabled. The locking now happens explicitly during entity manager's startup, rather than automatically when the first component gets added. This means some of the tests had to be updated, because they previously registered components/events after entity manager had already started.


Robust.UnitTesting/Shared/GameObjects/Systems/AnchoredSystemTests.cs
```diff
@@ -105,6 +99,49 @@ public override void Initialize()
...
 
+        internal sealed class MoveEventTestSystem : EntitySystem
+        {
+            [Dependency] private readonly SharedTransformSystem _transform = default!;
+
+            public override void Initialize()
+            {
+                base.Initialize();
+
+                _transform.OnGlobalMoveEvent += OnMove;
```
metalgearsloth: Should unsubscribe on shutdown I think.

metalgearsloth (approved):

PR #4730, opened 2023-12-18
Add more PVS logs

ElectroJr:
Trying to figure out what cause a NRE for a client on grafana


Robust.Server/GameStates/PvsSystem.Helpers.cs
```diff
@@ -14,7 +14,16 @@ internal sealed partial class PvsSystem
     {
         ref var data = ref CollectionsMarshal.GetValueRefOrAddDefault(entityData, entity, out var exists);
         if (!exists)
-            data = new(GetEntityData(entity));
```
ElectroJr, line 17: I think `GetEntityData()` failed because a deleted entity somehow got into the pvs tree/chunk, resulting in the "data" having a null metadata component?

PR #4734, opened 2023-12-18
Bound methods to IComponent, Unsafe.As cast instead of direct casting

mirrorcult:
unsafe.as is faster than a (T) cast in situations where its guaranteed to be of the correct type, and with the bounds these all should be

it is possible this isnt safe and im missing something! so dont merge if you dont know either

this i think redoes some of the work from reverted PRs for arch? but either way this should be done regardless of arch

requires https://github.com/space-wizards/space-station-14/pull/22692


PJB3005:
I'm generally scared of doing opts like this due to possibility of multi-threaded racing causing data to end in the wrong place.

mirrorcult:
is that actually a concern with this code specifically? i cant think of a way we could trigger something like that off the top of my head

PJB3005:
It'd make me more comfortable if there were explicit locks around the code that sets `_entTraitDict` so it can't possibly be raced. Otherwise it's probably fine.

PR #4739, opened 2023-12-19
Add option for left or right-aligned checkboxes

metalgearsloth:
I think right-aligned is better but this is less of a breaking change.


DrSmugleaf (commented):
Can't this use HorizontalAlignment instead?

PR #4750, opened 2023-12-22
Don't defer grid traversals

metalgearsloth:
Transform is in a slightly better state so we can remove this hacky thing.

It got to the point it was doing a bunch of unnecessary checks (no need to check container if you're just checking map / grid parent assuming no map / grid has containers).

The main concern now is: Does not deferring cause any issues, which it doesn't seem to (previously this mostly happened with transform state handling). On server-side the main worry is broadphasesystem seeing as this is being run manually to check for entities entering grids.


ElectroJr (approved):
I made some changes to the checks and also obsoleted `MoveEvent.FromStateHandling` which isn't required anymore if move events are no longer being deferred.

PR #4753, opened 2023-12-22
Comment out windows soundfont loader

metalgearsloth:
AKA """fixes""" "Not a soundfont file" getting shown twice every time a midi renderer comes into range.
It doesn't work due to fluidsynth not being able to load dls files on our version so this just avoids it spamming logs for now.


PJB3005:
> It doesn't work due to fluidsynth not being able to load dls files on our version so this just avoids it spamming logs for now.

That isn't the bug. The `.dls` file Windows ships is literally the soundfont we use. The Fluidsynth bug is that it gives the "not a soundfont" error even when it successfully loads a `.dls` soundfont.

PR #4754, opened 2023-12-22
Reduce robust mapped string allocs

metalgearsloth:
image

Top entry disappears.

Was tempted to use string.Intern as the string.Concat features heavily but wasn't sure if it would cause issues due to new data being sent or whatever and have opted to leave it for now.

### DIFF
--- a/Robust.Shared/Serialization/RobustMappedStringSerializer.MappedStringDict.cs
+++ b/Robust.Shared/Serialization/RobustMappedStringSerializer.MappedStringDict.cs
@@ -259,8 +259,11 @@ namespace Robust.Shared.Serialization
                     {
                         for (var sl = 1; sl <= parts.Length - si; ++sl)
                         {
-                            var subSubStr = String.Concat(parts.Skip(si).Take(sl));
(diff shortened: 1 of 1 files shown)


Robust.Shared/Serialization/RobustMappedStringSerializer.MappedStringDict.cs
```diff
@@ -259,8 +259,11 @@ public void AddString(string str)
                     {
                         for (var sl = 1; sl <= parts.Length - si; ++sl)
                         {
-                            var subSubStr = String.Concat(parts.Skip(si).Take(sl));
-                            AddString(subSubStr);
+                            // Don't ask me what the original was doing; we just skip by si and take sl
+                            // Use string interning as this is quite intensive.
+                            var end = si + sl;
+                            var subBetter = string.Concat(parts[si..^(parts.Length - end)]);
+                            AddString(subBetter);
```
metalgearsloth: Didn't look too hard but we could maybe .AsSpan or something else to make it better, I haven't worked much with strings before.

PR #4755, opened 2023-12-22
Marginally speedup texture preloading

metalgearsloth:


PJB3005 (commented):
Using a `Channel<T>` sounds like a way better option than allocating a morbillion `WaitHandle`s

https://github.com/space-wizards/RobustToolbox/blob/1bdd82b0bf6cbd78fe0cd64e8eb1dfaefeeeb122/Robust.Shared/Prototypes/PrototypeManager.cs#L438

PR #4759, opened 2023-12-24
Remove `RobustTree` & `PVSCollection`

ElectroJr:
This is the third part of the PVS refactor.

Changes include:
- Generally reorganises a bunch of PVS code. IMO it's better organised and easier to read now, though maybe I'm just slowly going insane.
- Removes `RobustTree` and `PVSCollection` in favour of a new `PvsChunk` class that stores entities as a sorted list rather than a in a tree.
  - Removes the recursive move event processing.
  - Iterating through entities should be faster.
  - Currently requires storing the chunk indices that an entity belongs to on the entity's metadata component.
  - Visibility masks are no longer part of the chunk location, all entities are just stored on the same chunk regardless of visibility.
- PVS override changes:
  - Entities can now have more than one kind of override, and will no longer get removed from the normal PVS chunks. 
  - Global pvs overrides are now processed / cached while updating old chunks, rather than having each player navigate the transform hierarchy. 
    - This should make things like the the round-end summary window have less of a performance impact.
  - global & session overrides now always send child entities. 
- `SessionPvsData` is now called `PvsSession` and stored more per-player information.
  - In particular, various pooled lists/collections are no longer pooled and are just stored here instead. 
- Chunks are now sorted by proximity to a player's current eye (see videos)
- Adds support for a basic chunk LOD setting.
  - This is so that you can reduce pop-in without massively increasing the number of entities that need to get sent to players. 
  - It just works by having far away chunks only send entities directly attached to the grid. I.e., walls, lights, mobs, lockers, but not inventories, contents or organs.
  - Can easily be extended to have more LOD levels. E.g., a chunk that sends inventory items, but not entities inside of inventories (e.g., stuff in backpacks)
- Adds a field on `EyeComponent` to support for per-eye PVS ranges, though it doesn't add any way of configuring it.

## Media:

Old chunk sorting, shown at 30tps, with pvs entry budget limited to 20 entities per tick, to make chunk loading easily visible.

https://github.com/space-wizards/RobustToolbox/assets/60421075/522f0c77-8856-458a-b9ed-1ed44c0c56f5

New chunk sorting:

https://github.com/space-wizards/RobustToolbox/assets/60421075/0a6582e5-17f6-4569-83f2-3cfc34e2c571


mirrorcult:
christmas miracle

Robust.Shared/GameObjects/Entity.cs
```diff
@@ -41,6 +41,8 @@ public Entity(EntityUid owner, T comp)
         owner = Owner;
         comp = Comp;
     }
+
+    public override int GetHashCode() => Owner.GetHashCode();
```
mirrorcult, line 45: isnt this not necessarily accurate? eg `Entity<FooComponent>` and `Entity<BarComponent>` for the same entity would hash to the same thing, or is the expectation that people should just wrap it into one `Entity`

ElectroJr: I'm assuming that people will probably not be storing an `Entity<T1>` and an `Entity<T2>` inside of the same hashset/collection, or be trying to compare them.

mirrorcult: thats fair

PR #4761, opened 2023-12-25
Fix playing audio streams

ike709:
Attempting to play an audio stream would cause an exception due to the filename being null, completely breaking audio in OpenDream.

OpenDream's audio is now fixed (make sure to unmute the video):

https://github.com/space-wizards/RobustToolbox/assets/5714543/6b2b8c40-fe95-4cad-b92e-80abd7cb0ef2

### DIFF
--- a/Robust.Client/Audio/AudioSystem.cs
+++ b/Robust.Client/Audio/AudioSystem.cs
@@ -404,9 +404,11 @@ public sealed partial class AudioSystem : SharedAudioSystem
 
     private bool TryGetAudio(AudioStream stream, [NotNullWhen(true)] out AudioResource? audio)
     {
-        audio = new AudioResource();
-        audio.Load(stream);
-        return true;
+        if (_resourceCache.TryGetResource(stream, out audio))
+            return true;
+
+        Log.Error($"Server failed to play audio stream {stream.Title}.");
+        return false;
     }
 
     public override (EntityUid Entity, AudioComponent Component)? PlayPvs(string filename, EntityCoordinates coordinates,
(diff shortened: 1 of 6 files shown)


Robust.Client/Audio/AudioSystem.cs
```diff
@@ -560,6 +574,15 @@ public override (EntityUid Entity, AudioComponent Component)? PlayGlobal(string
...
     }
 
+    public override void LoadStream<T>(AudioComponent component, T stream)
+    {
+        if (stream is AudioStream audioStream)
+        {
+            TryGetAudio(audioStream, out var audio);
+            SetupSource(component, audio!, audioStream.Length);
+        }
+    }
```
metalgearsloth, line 586: Generic is not even used?

Robust.Shared/Audio/Systems/SharedAudioSystem.cs
```diff
@@ -149,6 +149,29 @@ protected AudioComponent SetupAudio(EntityUid uid, string fileName, AudioParams?
...
+        }
+
+        if (comp.Params.Variation != null && comp.Params.Variation.Value != 0f)
+        {
+            comp.Params.Pitch *= (float) RandMan.NextGaussian(1, comp.Params.Variation.Value);
+        }
+
+        return comp;
+    }
+
```
metalgearsloth: Re-use methods rather than copy-pasting between them please.

Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs
```diff
@@ -12,6 +12,11 @@ public sealed class AudioResource : BaseResource
 {
     public AudioStream AudioStream { get; private set; } = default!;
 
+    public void Load(AudioStream stream)
+    {
+        AudioStream = stream;
+    }
+
```
metalgearsloth, line 19: This shouldn't be public and as mentioned above probably shouldn't be here at all.

Robust.Client/Audio/AudioSystem.cs
```diff
@@ -395,6 +402,13 @@ private bool TryGetAudio(string filename, [NotNullWhen(true)] out AudioResource?
         return false;
     }
 
+    private bool TryGetAudio(AudioStream stream, [NotNullWhen(true)] out AudioResource? audio)
+    {
+        audio = new AudioResource();
+        audio.Load(stream);
+        return true;
+    }
+
```
metalgearsloth: Shouldn't be constructing resources outside of ResourceCache, and I don't think the audioresource is even being used.

metalgearsloth (changes requested):

Robust.Client/Audio/AudioSystem.cs
ike709: AudioStream only lives in Client. I didn't know how else to declare LoadStream in Shared than having it be generic. Is there some better way to do it?

Robust.Client/Audio/AudioSystem.cs
ike709: It's the out parameter.

PR #4762, opened 2023-12-25
Don't defer adding joints

ElectroJr:
I don't know why this was (or is?) necessary, but it causes some pulling mispredicts.


metalgearsloth (approved):

metalgearsloth:
It was masking some prediction issues back in the day.

PR #4767, opened 2023-12-26
Stop toolshed from generating completions for non-toolshed commands

ElectroJr:
E.g., typing `forcemap Aspid ` (with the trailing space) will no longer suddenly list all toolshed commands as completion suggestions.

### DIFF
--- a/Robust.Server/Console/ServerConsoleHost.cs
+++ b/Robust.Server/Console/ServerConsoleHost.cs
@@ -82,7 +82,7 @@ namespace Robust.Server.Console
                 var sudoShell = new SudoShell(this, localShell, shell);
 
 #pragma warning disable CA2012
-                return CalcCompletions(sudoShell, args, argStr);
+                return CalcCompletionsOrEmpty(sudoShell, args, argStr);
 #pragma warning restore CA2012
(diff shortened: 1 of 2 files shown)


Robust.Server/Console/ServerConsoleHost.cs
```diff
@@ -247,7 +247,7 @@ private async void HandleConCompletions(MsgConCompletion message)
                 }
                 var (shedRes, _) = await completions.Value;
                 shedRes ??= CompletionResult.Empty;
-                result = new CompletionResult(shedRes.Options.Concat(result.Options).ToArray(), shedRes.Hint ?? result.Hint);
+                result = new CompletionResult(shedRes.Options.Concat(result?.Options ?? Array.Empty<CompletionOption>()).ToArray(), result?.Hint ?? shedRes.Hint);
```
metalgearsloth: I heard? you wanted some null coalescing ?????

ElectroJr: I changed it slightly?? Sot it might be more readable now????

moonheart08 (approved):

PR #4788, opened 2023-12-31
Add background color setting to the richtext

c4llv07e:
Add the ability to change the background color of the text in the markup language using "[bgcolor]". This works similarly to "[color]", but it can also be null.

Highlighting words in a chat as an example:

image


Robust.Client/UserInterface/RichText/MarkupDrawingContext.cs
```diff
@@ -7,26 +7,30 @@ namespace Robust.Client.UserInterface.RichText;
...
         Color = new Stack<Color>();
+        BackgroundColor = new Stack<Color?>();
         Font = new Stack<Font>();
         Tags = new List<IMarkupTag>();
     }
 
     public MarkupDrawingContext(int capacity)
     {
         Color = new Stack<Color>(capacity);
+        BackgroundColor = new Stack<Color?>(capacity);
```
metalgearsloth, line 25: This should be nullable.

metalgearsloth (changes requested):
Please use C# style guide for variable naming i.e. camelCase.

Robust.Client/Graphics/Font.cs
```diff
@@ -131,7 +132,15 @@ public override float DrawChar(DrawingHandleScreen handle, Rune rune, Vector2 ba
                 return metrics.Value.Advance;
             }
 
+            var start = baseline;
             baseline += new Vector2(metrics.Value.BearingX, -metrics.Value.BearingY);
+            if (background_color is Color bg_color)
```
PJB3005, line 137: This is not a good way to implement this. Large spans of background-colored text will break sprite batching and ruin performance.

You should have the text drawing code do separate drawing of the background in one go.

PJB3005 (changes requested):

c4llv07e:
Okay, then I'll think how to do it. Closing for now.

PR #4790, opened 2023-12-31
Implement OSWindowStyles.* on linux

LetterN:
Title; also adds `OSWindowStyles.NoTitleBar` for tgui-like windows with their custom titlebar and such

!!HEY!! this doesn't actually completely hide the (x) button, only the resize and such

also implements ownerWindow on linux


Robust.Client/Graphics/Clyde/Windowing/Glfw.Windows.cs
```diff
@@ -481,12 +483,54 @@ private static void WinThreadWinDestroy(CmdWinDestroy cmd)
...
+                            (byte*)&newPropVal, 1);
+#pragma warning restore CA1806
+                    }
                     else
                     {
                         _sawmill.Warning("OSWindowStyles.NoTitleOptions not implemented on this platform");
                     }
                 }
 
+                if ((parameters.Styles & OSWindowStyles.NoTitleBar) != 0)
```
PJB3005: You can use `GLFW_DECORATED` instead of manually setting it like this. The problem I'm worried about with enabling this for OpenDream is that I doubt we can ever make tgui's custom dragging logic work.

Robust.Client/Graphics/Clyde/Windowing/Glfw.Windows.cs
```diff
@@ -481,12 +483,54 @@ private static void WinThreadWinDestroy(CmdWinDestroy cmd)
                             // Cast to long here to work around a bug in rider with nint bitwise operators.
                             (nint)((long)Windows.GetWindowLongPtrW(hWnd, GWL.GWL_STYLE) & ~WS.WS_SYSMENU));
                     }
+                    else if (OperatingSystem.IsLinux())
+                    {
+                        var x11Window = (TerraFX.Interop.Xlib.Window)GLFW.GetX11Window(window);
+                        var x11Display = (Display*) GLFW.GetX11Display(window);
+                        DebugTools.Assert(x11Window != Xlib.None);
+
+                        var propName = Xlib.XInternAtom(x11Display, (sbyte*)Marshal.StringToCoTaskMemUTF8("_NET_WM_WINDOW_TYPE"), Xlib.False);
```
PJB3005: Aren't you leaking this memory every time a window is made?

Robust.Client/Graphics/Clyde/Windowing/Glfw.Windows.cs
```diff
@@ -481,12 +483,54 @@ private static void WinThreadWinDestroy(CmdWinDestroy cmd)
                             // Cast to long here to work around a bug in rider with nint bitwise operators.
                             (nint)((long)Windows.GetWindowLongPtrW(hWnd, GWL.GWL_STYLE) & ~WS.WS_SYSMENU));
                     }
+                    else if (OperatingSystem.IsLinux())
+                    {
+                        var x11Window = (TerraFX.Interop.Xlib.Window)GLFW.GetX11Window(window);
+                        var x11Display = (Display*) GLFW.GetX11Display(window);
+                        DebugTools.Assert(x11Window != Xlib.None);
```
PJB3005: This is not guaranteed to be the case if the user is using Wayland GLFW.

PJB3005 (changes requested):
Add `<RobustLinkAssemblies Include="TerraFX.Interop.Xlib" />` to the csproj btw free size savings.

Robust.Client/Graphics/Clyde/Windowing/Glfw.Windows.cs
```diff
@@ -481,6 +489,34 @@ private static void WinThreadWinDestroy(CmdWinDestroy cmd)
...
+                                (sbyte*)Marshal.StringToCoTaskMemUTF8("_NET_WM_WINDOW_TYPE_DIALOG"), Xlib.False);
+
+#pragma warning disable CA1806
+                            // [display] [window] [property] [type] [format (8, 16,32)] [mode] [data] [element count]
+                            Xlib.XChangeProperty(x11Display, x11Window, propName, Xlib.XA_ATOM, 32, Xlib.PropModeReplace,
+                                (byte*)&newPropVal, 1);
+#pragma warning restore CA1806
+
+                            Marshal.FreeCoTaskMem(propName);
+                            Marshal.FreeCoTaskMem(newPropVal);
```
PJB3005: Uh, I don't think it's safe to just free these like this?

PJB3005 (changes requested):

Robust.Client/Graphics/Clyde/Windowing/Glfw.Windows.cs
LetterN: whoops; the server does delete these on exit or when x11 server restarts

PJB3005 (approved):

PR #4795, opened 2024-01-02
Replace PVS dictionaries with an array

ElectroJr:
## Description

Currently PVS stores data about what entities have been seen by a player in a per-player `Dictionary<NetEntity, PvsData>`, where `PvsData` is a class. Every tick PVS iterates over all entities visible to a player and has to fetch or add an entry to this dictionary, which (AFAIK) is what PVS spends most of its time doing.

This PR attempts to make that faster by turning the `PvsData` class into a struct, replacing all of the dictionaries with a single array on the pinned object heap, and replacing stored `PvsData` references with an `IntPtr` (i.e., `List<PvsData>` -> `List<IntPtr>`).

Its hard to know exactly what impact this will have on real PVS performance. I've tried running several different benchmarks (see the benchmarking section). In the best case, the new approach is ~3x faster (`2.580ms` down to `0.893ms`), while in the worst case is only 10% faster (`6.123ms` -> `5.499ms`), though as I'll mention in that section, I think arrays are actually still significantly faster in that case as the benchmarks aren't making use of some parallelism.

Assuming this does actually significantly help with performance, and there isn't some major flaw in my implementation, I think its worth doing this despite the additional complexity & unsafe code? The major downside might be that it can use more memory than using dictionaries.

### Memory

The pinned array needs to be pre-allocated and takes 12 bytes per-player per-entity. In SS14, most long-lived entities will get seen (enter PVS range) by most players at least once, so I don't think it even increases memory usage relative to using dictionaries, though other games with lower player densities might require more memory.

For typical SS14 servers the array should take around 60mb (65k ents, 80 players). For OpenDream, paradise station apparently has ~150k entities, which with ~256 players would be ~440 MB. If this PR makes PVS significantly faster, I think it would still be worth it? Though I have no idea if the S14 benchmarks are going to be representative of how it'd affect OD or any game with low player densities.

I'm also not sure if this is actually significant, but using the struct array probably helps with some GC pressure? Previously deleting an entity would orphan each player's PvsData object for that entity (which also requires many dictionary lookups). Now it just has to clear the data in the array.

## Some Details

All of the unsafe/pointer code is confined to PvsSystem, though there is a new internal `IntPtr` field on the metadata component. `PvsSystem` allocates an array on the pinned object heap with some initial player & entity capacity, and it automatically grows the array if the entity or player cap is exceeded. Every player gets assigned a contiguous "row" of data in the array, which is just specified by some offset. Every spawned entity stores an `IntPtr` on their MetaData component pointing to a "column" in the array. If an entity is deleted or a player disconnects, their assigned row/columns are returned to a pool of available rows/columns. The first row / zeroth-player is actually used to store metadata about each entity (`NetEntitty`, vis mask and last modified tick).

Previously, getting data from the dictionary would look like this:

```cs
private bool ProcessEntity(PvsSession session, Entity<MetaDataComponent> entity)
{
    ref var data = ref CollectionsMarshal.GetValueRefOrAddDefault(session.Entities, entity.Comp.NetEntity, out var exists);
    if (!exists)
        data = new(nuid);
    // do stuff
}
```
now the equivalent code uses:
```cs
private unsafe bool ProcessEntity(PvsSession session, Entity<MetaDataComponent> entity)
{
    var ptr = ((PvsData*) entity.Comp.PvsData) + session.Offset;
    ref var data = ref Unsafe.AsRef<PvsData>(ptr);
    // do stuff
}
```

The main complication is that PvsSystem needs to ensure that it doesn't accidentally use old pointers from deleted entities while processing late game-state ack messages. It currently does this by making use of the existing `ForceAckThreshold` by just temporarily storing pointers from deleted entities until we can know for sure that all references to those pointers are gone. Even if that fails somehow, it should only ever modify data within the pinned array, and nothing outside of it. It would just mean that PVS thinks that a player has already seen/acked an entity, which isn't a critical error.

## Benchmarks

In this section I've used the two benchmarks added in space-wizards/space-station-14/pull/23166. One benchmark measures PVS with static viewers & unmodified entities, and one where the players are teleporting around. Only the latter of these actually has entities entering & leaving a player's PVS range, though its also hugely exaggerated because the visible entities will completely change every tick. I've further divided these benchmarks into sub-categories to try and deal with two reasons that the benchmarks might not be representative of "real" results.

Firstly, both the dictionary and arrays seem to benefit from some (possibly) unrealistic cache locality. E.g., for the arrays, the order in which entities are spawned and added to the array is correlated to the order in which entities are initially added to PVS chunks, and thus the order in which they get sent to players. In a real round, entities would move around and some of this correlation is lost, though static entities or grouped entities (mob + internal organs) would continue to benefit from it. So there are two kinds of benchmarks: shuffled and unshuffled, where the latter is basically an extreme case where any correlation / cache locality is gone. For arrays, shuffled just means we randomly assign a "column" in the array to each entity, while for for the dictionary benchmark it means that entities are assigned a random (though still unique) NetEntity. Both of these changes make the benchmarks significantly slower.

A second issues that complicates interpreting the benchmarks is that after #4787, pvs-leave messages are now no longer processed in the main PVS tick, but are instead just running in the background while the rest of the game's tick processing happens. However, for the sake of the benchmarks pvs-leave still gets run synchronously, but there is a fast and slow way of doing this. Either way isn't going to be representative of what a server will really be doing. Additionally, if many entities are leaving a player's PVS view, then the new pvs-leave processing with the array might actually be slower than it was with dictionaries, though overall PVS tick time is still faster with arrays. Given that in a real server this slower part would be running asynchronously, the non-static benchmark is probably underestimating the benefit of moving to arrays.

 
### static, unshuffled, fast pvs-leave 
|     Method | PlayerCount |     Mean |     Error |    StdDev |
|----------- |------------ |---------:|----------:|----------:|
|      Array |          80 | 0.892 ms | 0.0036 ms | 0.0032 ms |
| Dictionary |          80 | 2.580 ms | 0.0135 ms | 0.0127 ms |

### static, shuffled, fast pvs-leave
|     Method | PlayerCount |     Mean |     Error |    StdDev |
|----------- |------------ |---------:|----------:|----------:|
|      Array |          80 | 2.006 ms | 0.0383 ms | 0.0358 ms |
| Dictionary |          80 | 3.465 ms | 0.0103 ms | 0.0086 ms |

### static, shuffled, slow pvs-leave
|     Method | PlayerCount |     Mean |     Error |    StdDev |
|----------- |------------ |---------:|----------:|----------:|
|      Array |          80 | 2.410 ms | 0.0389 ms | 0.0364 ms |
| Dictionary |          80 | 3.805 ms | 0.0127 ms | 0.0119 ms |

###  teleporting, unshuffled, fast pvs-leave
|     Method | PlayerCount |     Mean |     Error |    StdDev |
|----------- |------------ |---------:|----------:|----------:|
|      Array |          80 | 4.483 ms | 0.0798 ms | 0.0746 ms |
| Dictionary |          80 | 5.472 ms | 0.0280 ms | 0.0262 ms |

### array, shuffled, slow pvs-leave
|     Method | PlayerCount |     Mean |     Error |    StdDev |
|----------- |------------ |---------:|----------:|----------:|
|      Array |          80 | 5.499 ms | 0.0237 ms | 0.0198 ms |
| Dictionary |          80 | 6.123 ms | 0.0258 ms | 0.0241 ms |


mirrorcult:
we getting pfives pvsix AND pvseven for christmas

PJB3005:
The link to the benchmarks PR is broken.

Robust.Server/GameStates/PvsSystem.Pointers.cs
```diff
@@ -0,0 +1,454 @@
...
+
+    /// <summary>
+    /// Clear all of this entity' PvsData entries. This effectively means that PVS will act as if no player
+    /// had never been sent information about this entity. Used when returning the entity's pointer back to the pool.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private unsafe void ClearEntityPvsData(IntPtr intPtr)
+    {
+        // TODO PVS is there a faster way to do this?
+        // We just need to clear 12-bytes with some fixed stride.
```
moonheart08: If you make this 16 bytes instead, you can use Vector128's set operation to set entity data in one clock. This will additionally avoid cache line overlaps (where a PVS data entry is within two cache lines, incurring additional overhead every few entries.)

moonheart08: I'd doublecheck it actually generates a vectorized store, mind you.

PJB3005: (I'm doing this)

ElectroJr:
I don't fully understand how some of the new memory stuff works, but I trust PJB. Benchmark wise its about the same speed as my pointer stuff. Benchmark results have changed relative to the original PR, but that's probably mainly because ~2 months have gone by and the benchmarks depend on Content maps & code.


Master / Dictionaries :
| Method     | PlayerCount | Mean     | Error     | StdDev    |
|----------- |------------ |---------:|----------:|----------:|
| Static   | 80          | 3.393 ms | 0.0172 ms | 0.0160 ms |
| Teleporting | 80          | 5.096 ms | 0.0275 ms | 0.0257 ms |

Old:
| Method     | PlayerCount | Mean     | Error     | StdDev    |
|----------- |------------ |---------:|----------:|----------:|
| Static   | 80          | 1.283 ms | 0.0058 ms | 0.0051 ms |
| Teleporting | 80          | 4.327 ms | 0.0162 ms | 0.0126 ms |


New memory managment:
| Method     | PlayerCount | Mean     | Error     | StdDev    |
|----------- |------------ |---------:|----------:|----------:|
| Static   | 80          | 1.313 ms | 0.0053 ms | 0.0044 ms |
| Teleporting | 80          | 4.331 ms | 0.0397 ms | 0.0372 ms |

PR #4809, opened 2024-01-05
Revert some grid/map related changes

ElectroJr:
This reverts some of the changes in #4806, in case they are causing some of the recent broadphase errors


Robust.Shared/GameObjects/Systems/EntityLookupSystem.cs
```diff
@@ -195,6 +196,12 @@ private void OnMapChange(MapChangedEvent ev)
         }
     }
 
+    private void OnGridAdd(GridAddEvent ev)
+    {
+        // Must be done before initialization as that's when broadphase data starts getting set.
+        EnsureComp<BroadphaseComponent>(ev.EntityUid);
```
ElectroJr, line 202: I have a sneaking suspicion this old comment was onto something

DrSmugleaf (approved):

PR #4815, opened 2024-01-06
Fix volume setting values being over or under 1 making the client unable to connect to servers

### LINKED ISSUE #23491: Invalid audio settings cause clients to be stuck in connecting forver
## Description

**Reproduction**

1. In your client_config.toml. set ambience.volume greater or negative to 1
2. Try to connect

**Screenshots**

**Additional context**

There's been a bunch of people getting this when trying to connect

The fix by faint is by modifying the file but it should not be the fix and we should be handling this by resetting the audio settings

https://discord.com/channels/310555209753690112/790656972801572905/1191790144638951454

VasilisThePikachu:
Fixes https://github.com/space-wizards/space-station-14/issues/23491

I wanted to make this do logging. However i cant get Log.Error to work on static context and Im not sure how much of a challenge or how much it will break if i make it non static so i just left it

This will shrimply turn the value to 0.

Maybe update engine after merging on content since its kinda is a hotfix

### DIFF
--- a/Robust.Shared/Audio/Systems/SharedAudioSystem.cs
+++ b/Robust.Shared/Audio/Systems/SharedAudioSystem.cs
@@ -154,7 +154,7 @@ public abstract partial class SharedAudioSystem : EntitySystem
     {
         if (value < 0f)
         {
-            throw new InvalidOperationException($"Tried to get volume calculation for gain of {value}.");
+            value = 0f;
         }
 
         return 10f * MathF.Log10(value);


Robust.Shared/Audio/Systems/SharedAudioSystem.cs
```diff
@@ -154,7 +154,7 @@ public static float GainToVolume(float value)
     {
         if (value < 0f)
         {
-            throw new InvalidOperationException($"Tried to get volume calculation for gain of {value}.");
+            value = 0f;
         }
```
metalgearsloth, line 158: This should still be logging an error.

VasilisThePikachu: Yeah like I mentioned. I can't get Log.Error to work since it's a static function

Can't just use the throw since it's gonna stop execution and just lead us back to the beginning of it not letting you connect

PR #4821, opened 2024-01-07
Add random pick / take methods for sets

metalgearsloth:
I want it don't at me.


Robust.Shared/Random/IRobustRandom.cs
```diff
@@ -119,6 +120,35 @@ public static bool Prob(this System.Random random, double chance)
...
+
+        foreach (var tile in set)
+        {
+            if (idx == index)
+                return tile;
+
+            idx++;
+        }
+
+        throw new InvalidOperationException();
```
PJB3005: This should be an `UnreachableException()` instead.

Robust.Shared/Random/IRobustRandom.cs
```diff
@@ -119,6 +120,35 @@ public static bool Prob(this System.Random random, double chance)
         return random.NextDouble() < chance;
     }
 
+    /// <summary>
+    /// Picks a random element from a set and returns it.
+    /// </summary>
```
PJB3005: This doc comment should make it clear this is O(n).

PJB3005 (changes requested):
Engine changelog pls.

PR #4829, opened 2024-01-11
Add support for forcing prototypes in a file or directory to be parsed as abstract

DrSmugleaf:
For downstreams that want to force some prototypes to be abstract.
This is specially useful in cases where content calls EnumeratePrototypes and loops over them.
You can now for example ignore all Guidebook entries from upstream without modifying them or any code that enumerates them.
The alternative would be either modifying those prototypes (likely to get conflicts) or the code that enumerates those prototypes (also likely to get conflicts), this allows for example for a call to be placed in EntryPoint Init (less likely to get conflicts), or in the future an event listener to do this (0 conflicts).
The reason to parse them as abstract instead of not at all is that this lets you inherit those prototypes.
Not used upstream.
Haven't profiled it but hopefully the single boolean check when not used makes it fast enough™️.
Callers must handle how to determine which files or directories to ignore on their own, there is intentionally no engine API for this to keep the API surface small.

@EmoGarbage404

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
@@ -14,6 +14,23 @@ namespace Robust.Shared.Prototypes;
 
 public partial class PrototypeManager
 {
+    /// <summary>
+    ///     Which files to force all prototypes within to be abstract.
+    /// </summary>
+    private readonly HashSet<ResPath> _abstractFiles = new();
+
+    /// <summary>
+    ///     Which directories to force all prototypes recursively within to be abstract.
+    /// </summary>
+    private readonly HashSet<ResPath> _abstractDirectories = new();
+
+    /// <summary>
+    ///     Whether or not any prototypes may be potentially forced to be abstract.
+    ///     This is true if either <see cref="_abstractFiles"/> or <see cref="_abstractDirectories"/> contain elements,
+    ///     and false otherwise.
+    /// </summary>
+    private bool _hasForcedAbstractPrototypes;
+
     public event Action<DataNodeDocument>? LoadedData;
 
     /// <inheritdoc />
@@ -35,6 +52,7 @@ public partial class PrototypeManager
             {
                 try
                 {
+                    var ignored = IsFileAbstract(file);
                     using var reader = ReadFile(file, !overwrite);
 
                     if (reader == null)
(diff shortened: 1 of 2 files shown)


Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
```diff
@@ -14,6 +14,23 @@ namespace Robust.Shared.Prototypes;
...
 {
+    /// <summary>
+    ///     Which files to force all prototypes within to be abstract.
+    /// </summary>
+    private readonly HashSet<ResPath> _abstractFiles = new();
+
+    /// <summary>
+    ///     Which directories to force all prototypes recursively within to be abstract.
+    /// </summary>
+    private readonly HashSet<ResPath> _abstractDirectories = new();
```
PJB3005: Making these `HashSet<T>` doesn't make much sense since you're only ever iterating over them. Unless we care strongly about content registering duplicates it'd be better to just make this a `List<T>` or `ValueList<T>`.

Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
```diff
@@ -14,6 +14,23 @@ namespace Robust.Shared.Prototypes;
...
+
+    /// <summary>
+    ///     Whether or not any prototypes may be potentially forced to be abstract.
+    ///     This is true if either <see cref="_abstractFiles"/> or <see cref="_abstractDirectories"/> contain elements,
+    ///     and false otherwise.
+    /// </summary>
+    private bool _hasForcedAbstractPrototypes;
```
PJB3005: This honestly doesn't seem worth storing separately, a couple count checks per prototype isn't going to be noticeable.

Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
DrSmugleaf: Done in [`c11e546` (#4829)](https://github.com/space-wizards/RobustToolbox/pull/4829/commits/c11e546ea0027fea04710ae42c76f7a440f7921b) (as a List<T>)

Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
DrSmugleaf: Done in [`c11e546` (#4829)](https://github.com/space-wizards/RobustToolbox/pull/4829/commits/c11e546ea0027fea04710ae42c76f7a440f7921b)

Issue #4832, opened 2024-01-12
Full state update breaks the game badly


koteq:
## Description

The full state update, which the engine triggers under [some](https://github.com/space-wizards/RobustToolbox/blob/dfb5369664fe1b146a9cbb9f15d90d08d11db55e/Robust.Client/GameStates/ClientGameStateManager.cs#L392) [rare](https://github.com/space-wizards/RobustToolbox/blob/dfb5369664fe1b146a9cbb9f15d90d08d11db55e/Robust.Client/GameStates/GameStateProcessor.cs#L163) [conditions](https://github.com/space-wizards/RobustToolbox/blob/dfb5369664fe1b146a9cbb9f15d90d08d11db55e/Robust.Client/GameStates/ClientGameStateManager.cs#L288) and the client [requests](https://github.com/space-wizards/RobustToolbox/blob/9e6d0aa44ad25b20269e533db7bd874f0e65c96e/Robust.Client/GameStates/ClientGameStateManager.cs#L389-L394) [sometimes](https://github.com/space-wizards/RobustToolbox/blob/270326e188f373a97029858ebccd7730dbf14f6c/Robust.Client/GameStates/GameStateProcessor.cs#L140-L163), breaks the game badly.

**Reproduction**

Force a full state update with the `fullstatereset` command.

I've been running into the issue myself on live servers without the command, just by playing normally. As well as I heard complaints from players who were providing screenshots with seemingly broken game state.

**Screenshots**

https://github.com/space-wizards/RobustToolbox/assets/1192090/5bf8d555-f56b-457e-b0c4-98e81ce69cc9

Note that the jittery motion isn't a recording bug. The game runs at 120 fps and the recording is at 60 fps.

**Additional context**

I'm not exactly sure what causes the bug, but it seems to be coming from PVS logic. I tried to debug it a little bit, and the server-side full state looks good to me, however the client-side state for the tick after the full update contains `EntityStates` with nullish `ComponentChanges` triggering the [MissingMetadataException](https://github.com/space-wizards/RobustToolbox/blob/dfb5369664fe1b146a9cbb9f15d90d08d11db55e/Robust.Client/Replays/Loading/ReplayLoadManager.Start.cs#L71).

The error from the in-game console, for convenience. The entity ID changes when I move across the map. I tried to `vv` several IDs, and they were belonged to LV power cables. The view variable window showed null in the metadata component field client side, but the server side worked just fine.
```
[...]

PR #4833, opened 2024-01-12
Fix full state updates

koteq:
## About the PR

Hopefully, address the weird issue with broken full state updates https://github.com/space-wizards/RobustToolbox/issues/4832

## Technical details

I noticed that `session.Entities` isn't cleared properly, causing the engine to mistakenly think that the client already seen/ack entities. Also, I think that the full update should skip PVS budgets, as it helps to address UI glitches during the update even though it causes a CPU and network utilization spike.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -18,7 +18,7 @@ Don't change the format without looking at the script!
 
 ### Bugfixes
 
-*None yet*
+* Fixed broken full state updates and excluded it from PVS budged.
 
 ### Other
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -18,7 +18,7 @@ Don't change the format without looking at the script!
 
 ### Bugfixes
 
-* Fixed broken full state updates and excluded it from PVS budged.
+*None yet*
 
 ### Other
 
@@ -43,7 +43,7 @@ END TEMPLATE-->
 
 ### Bugfixes
 
-*None yet*
+* Fixe broken full state updates and exclude it from PVS budged.
 
 ### Other
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -43,7 +43,7 @@ END TEMPLATE-->
 
 ### Bugfixes
 
-* Fixe broken full state updates and exclude it from PVS budged.
+* Fix broken full state updates and exclude it from the PVS budget.
 
 ### Other
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -43,7 +43,7 @@ END TEMPLATE-->
 
 ### Bugfixes
 
-* Fix broken full state updates and exclude it from the PVS budget.
+* Fix broken full state updates.
 
 ### Other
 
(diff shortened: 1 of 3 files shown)


RELEASE-NOTES.md
```diff
@@ -43,7 +43,7 @@ END TEMPLATE-->
 
 ### Bugfixes
 
-*None yet*
+* Fixe broken full state updates and exclude it from PVS budged.
```
```suggestion
* Fix broken full state updates and exclude it from the PVS budget.
```

Robust.Server/GameStates/PvsSystem.cs
```diff
@@ -294,15 +294,19 @@ private void GetEntityStates(PvsSession session)
...
+            {
+                session.Budget.NewLimit = _netConfigManager.GetClientCVar(session.Channel, CVars.NetPVSEntityBudget);
+                session.Budget.EnterLimit =
+                    _netConfigManager.GetClientCVar(session.Channel, CVars.NetPVSEntityEnterBudget);
+            }
+            else
+            {
+                session.Budget.NewLimit = CVars.NetPVSEntityBudget.DefaultValue;
+                session.Budget.EnterLimit = CVars.NetPVSEntityEnterBudget.DefaultValue;
+            }
```
metalgearsloth: Doesn't this always need updating as the budget still matters even with a fullstateupdate?

koteq: @metalgearsloth, the `NewLimit` budged serves a simple purpose, it prevents clients from having stutters / micro freezes when many new entities spawn-in _outside the client's view range_, e.g. when you first visit some parts of a station that are heavily populated with entities. The `NewLimit` budget kind of expects clients to have plenty of previous states, which isn't the case for the full state update. The full state update brings clients into a void with no entities around them, and the `NewLimit`, most of the time, prevents the full state update from having enough information to even have basic structures such as floor, walls, and lights spawned around the player.

Here is a short demo where I first walk the Box station's corridor pressing Num9 doing full state resets ignoring all the limits, and then the same corridor, but with Num0 doing full state resets respecting the default PVS budget limits.

https://github.com/space-wizards/RobustToolbox/assets/1192090/ebd3c310-9489-489a-8f90-29cb21a053b6

Notice that both kinds of updates cause lags, however the update that has no limits, doesn't introduce any visual glitches.

metalgearsloth: The streaming is for server purposes as much as it is for client, not obeying the streaming limit would cause dropped ticks.

koteq: Given full state updates being rare and the server-side making a simple entity iteration plus data serialization, I wouldn't expect it to cause too much trouble or even cause ticks to run out of the 1/60 interval.

Anyway, I removed the limit changes to prevent the PR from being blocked for too long. I guess visual glitches during a full state update deserve their own individual PR. Maybe someone would find a way of applying full state updates without reducing the game state to a complete void. Or perhaps the sheer existence of full state updates can be somehow removed from the game, e.g. by changing the way how replication detects entities being out of sync and smartly re-replicating only those entities.

ElectroJr: IMO temporary visual glitched caused by a full state / reset aren't an important issue, because full states / resets shouldn't ever ever be happening unless there is some other bug/issue. They're really just meant to be a fallback for when something else goes wrong.

PR #4840, opened 2024-01-16
Prevent memory leak on MsgState deserialization fail

Morb0:
`finalStream` now disposes even if `DeserializeDirect` throw exception, this prevent `StreamFinalized` recyclable memory stream error.


Robust.Shared/Network/Messages/MsgState.cs
```diff
@@ -53,9 +53,16 @@ public override void ReadFromBuffer(NetIncomingMessage buffer, IRobustSerializer
...
-            serializer.DeserializeDirect(finalStream, out State);
+            try
+            {
+                serializer.DeserializeDirect(finalStream, out State);
+            }
+            finally
+            {
+                finalStream.Dispose();
+            }
+
```
metalgearsloth, line 70: Doesn't this double-dispose it with zstd.

Morb0: zstd stream is push data into finalStream
not sure wdym

PJB3005:
This isn't really a memory leak, the GC will still clean up the pooled objects even if they don't get cleaned up properly. 

Generally however, I believe it's recommended against returning pooled buffers to stream in exception circumstances, which this certainly is. That said the reason why isn't applicable here.

I feel like this isn't an important issue regardless? Does this exception even regularly happen?

LordCarve:
Why not just slap a `using var finalStream = RobustMemoryManager.GetMemoryStream(uncompressedLength);` and be done with it for good? No need to worry about any of the wheres and whens this way.

If you're concerned about early memalloc in case decompression fails, then you can wiggle initialization to after decompression but we're being super-nitpicky in an already-nitpicky case here that I honestly don't think it matters. JIT is likely pushing that memalloc to the latest possible moment anyway.

>I feel like this isn't an important issue regardless?

Important - no. But I see it as a point of improvement. Not a memory leak as you yourself noticed, but it's always a good idea to ensure all `IDisposable`s get disposed. OP found a rare[^1] circumstance where it might not get disposed and patched it up. An improvement with no downside if you ask me. A tiny thing, but an improvement nonetheless.

[^1]: An argument could be made that frequency at which the exception occurs here doesn't play a role. If it can fail, it _sometimes will_ fail. The code is better if those failures are more graceful.

Robust.Shared/Network/Messages/MsgState.cs
LordCarve: Only `decompressStream` is disposed in the handle-compressed block. It's a different stream.
No double-disposing happening here.

PR #4858, opened 2024-01-24
Update SDL Bindings & SDL to 3

LetterN:
title, using `Alimer.Bindings.SDL`


PJB3005:
The library you're using links SDL3 incorrectly on Linux, it needs to be linked with `-Bsymbolic` to avoid crapping out with host libraries.

PR #4862, opened 2024-01-25
Exposed "Bottom Margin" to itemlist

jamessimo:
Very simple, wanted to expose the "BottomMargin" in items lists so we can have natural padding between items.

Defaults at 0 so doesn't impact existing UI

### DIFF
--- a/Robust.Client/UserInterface/Controls/ItemList.cs
+++ b/Robust.Client/UserInterface/Controls/ItemList.cs
@@ -9,6 +9,9 @@ using Robust.Shared.Input;
 using Robust.Shared.Maths;
 using Timer = Robust.Shared.Timing.Timer;
 
+/// <summary>
+/// Represents a scrollable list of items in a user interface.
+/// </summary>
 namespace Robust.Client.UserInterface.Controls
 {
     [Virtual]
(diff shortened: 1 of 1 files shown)


PJB3005 (commented):
`ItemSeparation` as name makes more sense IMO.

jamessimo:
> `ItemSeparation` as name makes more sense IMO.

+1 for this, I fixed it already.

Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -29,6 +32,10 @@ public class ItemList : Control, IList<ItemList.Item>
         public const string StylePropertySelectedItemBackground = "selected-item-background";
         public const string StylePropertyDisabledItemBackground = "disabled-item-background";
 
+        /// <summary>
+        /// Gets or sets the bottom margin of individual list items
```
PJB3005: Update comment too?

Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -68,6 +75,7 @@ private void RecalculateContentHeight()
                 itemHeight += ActualItemBackground.MinimumSize.Y * UIScale;
 
                 _totalContentHeight += (int)Math.Ceiling(itemHeight);
+                _totalContentHeight += ItemSeparation;
```
PJB3005, line 78: If the intent is this should be a separation then there probably shouldn't be a space after the very last item.

Robust.Client/UserInterface/Controls/ItemList.cs
jamessimo: Done

Robust.Client/UserInterface/Controls/ItemList.cs
jamessimo: Fixed

PR #4863, opened 2024-01-26
Localization support for any prototypes via a YAML tag

koteq:
## About the PR

[Yet another](https://github.com/space-wizards/RobustToolbox/pull/4335) attempt to have localization for those pesky hard-coded YAML strings.

So, my idea is to add `!Loc` YAML tag, passing tagged strings through the localization manager.

## Example

Here is a prototype for a jug that contains aluminium. The `name` and the `currentLabel` fields are hard-coded player facing strings that don't allow localization. While the `name` field uses [a smart trick](https://github.com/space-wizards/RobustToolbox/blob/20a411e6cee4b82b0fda1c1ba2070a2a2d60f90e/Robust.Shared/Prototypes/EntityPrototype.cs#L53-L54), replacing it [with a localized version later](https://github.com/space-wizards/RobustToolbox/blob/20a411e6cee4b82b0fda1c1ba2070a2a2d60f90e/Robust.Shared/Prototypes/EntityPrototype.cs#L73). The `currentLabel` can't use a similar trick, as it holds a free-form string, allowing players to change it with the Hand labeler any time.

```yaml
# Resources/Prototypes/Entities/Objects/Specific/chemical-containers.yml
- type: entity
  parent: Jug
  name: jug (aluminium)
  id: JugAluminium
  noSpawn: true
  components:
    - type: Label
      currentLabel: aluminium
    - type: SolutionContainerManager
      solutions:
        beaker:
          reagents:
            - ReagentId: Aluminium
              Quantity: 200
```

My idea is to allow tagging strings in prototypes so that they can get swapped with localized versions right at the moment when prototypes are being loaded.

```yaml
      currentLabel: !Loc 'reagent-name-aluminium'
```

This approach seems to be working fine, here is a reagents dispenser screenshot of the Corvax SS14 build, that uses the jug prototype

image

### DIFF
--- a/Robust.Shared/Serialization/Manager/SerializationManager.Reading.cs
+++ b/Robust.Shared/Serialization/Manager/SerializationManager.Reading.cs
@@ -3,6 +3,7 @@ using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Linq.Expressions;
 using Robust.Shared.GameObjects;
+using Robust.Shared.Localization;
 using Robust.Shared.Serialization.Manager.Definition;
 using Robust.Shared.Serialization.Manager.Exceptions;
 using Robust.Shared.Serialization.Markdown;
@@ -77,6 +78,12 @@ namespace Robust.Shared.Serialization.Manager
                     (typeof(T), type, node.GetType()!, notNullableOverride),
                     static (tuple, manager) => ReadDelegateValueFactory(tuple.baseType, tuple.actualType, tuple.node, tuple.notNullableOverride, manager),
                     this))(node, hookCtx, context, instanceProvider);
+            }
+
+            if (node.Tag == "!Loc" && node is ValueDataNode valueNode && typeof(T) == typeof(string))
+            {
+                var loc = DependencyCollection.Resolve<ILocalizationManager>();
+                return (T)(object)loc.GetString(valueNode.Value);
             }
 
             return ((ReadGenericDelegate<T>)_readGenericDelegates.GetOrAdd((typeof(T), node.GetType()!, notNullableOverride),


mirrorcult:
interesting, kind of surprised it seems that easy to impl in serialization

PJB3005:
I just realized this wouldn't ever be compatible with any attempt to have client-switchable languages which would be a problem.

koteq:
Huh? Clients could easily reload prototypes upon a language change.

What really isn't compatible is the fact a jug's initial label live server-side, though it has nothing to do with the translation approach.

In other situations, e.g., for translating a prototype's name, it should work just fine.

PJB3005:
> In other situations, e.g., for translating a prototype's name, it should work just fine.

No, because half the time those will be coming from the server and it'll be impossible to tell what the original localization string is.

With a model where the client can choose its own language, the server should *never* be sending a localized string, always sending the identifiers instead. This makes that impossible.

koteq:
> because half the time those will be coming from the server 

The server can be sending raw prototypes to clients keeping the tags in place, right?

PJB3005:
> The server can be sending raw prototypes to clients keeping the tags in place, right?

No, because if the server ever has to do any processing on that string (such as including it in another string, like "X inserts the Y into Z" it'll fall apart.

deltanedas:
serialize the args object list :trollface:

koteq:
> if the server ever has to do any processing

> the server should never be sending a localized string

In the statements, a contradiction I notice.

---

If the server has to do processing, then regardless of the implementation it'd be sending a string, not a message ID.

PJB3005:
> If the server has to do processing, then regardless of the implementation it'd be sending a string, not a message ID.

That's the trick, our plan (if it ever gets implemented) is to **never send a localized string**, by instead sending something akin to Unreal's [FText](https://docs.unrealengine.com/5.3/en-US/ftext-in-unreal-engine/). i.e. instead of "you insert the cell into the flashlight" it sends `storage-slot-insert-popup(object=ent-power-cell, target=ent-flashlight)`. Or something like that. Then the client can resolve the localization strings itself based on the local language.

koteq:
It should be sending `!Loc 'storage-slot-insert-popup(object=ent-power-cell, target=ent-flashlight)'`, if it's in a prototype's YAML isn't it this simple :trollface:

PR #4864, opened 2024-01-28
network grammar + api

deltanedas:
the networkedcomponent part works from examining urists

moved the evil vv Gender and ProperNoun into proper public api for the rest of engine and content to use

since changing 90% of the file also moved it into GameObjects namespace like almost? everything else is so no random GameObjects.Components.Localization ops


PJB3005:
Do we need any dirty calls for this?

deltanedas:
i didnt see anything modify it so probably not

from going back and forth it definitely fixes it, but if its a concern add access and make anything that modifies it use an api, then no missing dirty calls

metalgearsloth:
> i didnt see anything modify it so probably not
> 
> from going back and forth it definitely fixes it, but if its a concern add access and make anything that modifies it use an api, then no missing dirty calls

That would be better.

deltanedas:
ok this will need content pr too for setting Gender and stuff (if i understand how Access works with setter) so may as well move that out of the component and into the api system

Robust.Shared/Localization/LocalizationManager.cs
```diff
@@ -28,6 +28,8 @@ internal sealed partial class LocalizationManager : ILocalizationManagerInternal
         [Dependency] private readonly IPrototypeManager _prototype = default!;
         [Dependency] private readonly IEntityManager _entMan = default!;
 
+        private GrammarSystem _grammar = default!;
```
deltanedas: how do i dependency here

both this way and [Dependency] fail miserably when running game

RELEASE-NOTES.md
```diff
@@ -35,7 +35,8 @@ END TEMPLATE-->
 
 ### Breaking changes
 
-*None yet*
+* `GrammarComponent` now has access so you must use `GrammarSystem`.
```
deltanedas, line 39: never touched release notes will this create 50 conflicts with other prs

Robust.Shared/Localization/LocalizationManager.cs
deltanedas: @PJB3005 any idea

PJB3005: Just dynamically fetch it every time the function is called, you cannot safely cache this value from a manager.

It would probably be better if the localization functions were registered by an entity system instead, but idk if there's issues with loc functions being loaded after localization strings are.

deltanedas:
how the fuck did i make tests loc manager not have a default culture

PR #4867, opened 2024-01-28
Fixes parsing of ShaderBlendMode in ShaderPrototype

collinlunn:
Fixes an issue with ShaderPrototype - it's not currently possible to set the ShaderBlendMode by YAML due to an issue with how it was parsed.

Currently, it takes the blend mode in YAML and uses ToUpper() which converts the string to ALL upper case, when it should only be capitalizing the first letter, because:

It uses Enum.TryParse<ShaderBlendMode>() to convert the string to a ShaderBlendMode enum value. However, Enum.TryParse() is case sensitive. "SUBTRACT" does not get converted to ShaderBlendMode.Subtract, only "Subtract" would, so the method always fails to get an enum value.

To resolve this, I made only the first letter get capitalized.

### DIFF
--- a/Robust.Client/Graphics/Shaders/ShaderPrototype.cs
+++ b/Robust.Client/Graphics/Shaders/ShaderPrototype.cs
@@ -67,10 +67,9 @@ namespace Robust.Client.Graphics
 
                     var hasLight = _rawMode != "unshaded";
                     ShaderBlendMode? blend = null;
-                    if (!string.IsNullOrEmpty(_rawBlendMode))
+                    if (_rawBlendMode != null)
                     {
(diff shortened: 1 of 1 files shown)


mirrorcult:
would be better to call the overload of `Enum.TryParse` that takes in a bool for case insensitivity imo 

`public static bool TryParse<TEnum> (string? value, bool ignoreCase, out TEnum result) where TEnum : struct;`

collinlunn:
> would be better to call the overload of `Enum.TryParse` that takes in a bool for case insensitivity imo
> 
> `public static bool TryParse<TEnum> (string? value, bool ignoreCase, out TEnum result) where TEnum : struct;`

Changed to use that overload instead, seems cleaner.

mirrorcult (approved):

PR #4875, opened 2024-01-29
Smarter EntitySystem sawmill name conversion

mirrorcult:
this was bothering me far more than it should have

b4

rider64_2024-01-29_04-17-41

afta

rider64_2024-01-29_04-18-33

### DIFF
--- a/Robust.Shared/GameObjects/EntitySystem.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
 using System.Linq;
 using JetBrains.Annotations;
 using Robust.Shared.IoC;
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/EntitySystem.cs
```diff
@@ -41,8 +42,16 @@ protected virtual string SawmillName
                     name = name.Substring(0, name.Length - "System".Length);
 
                 // Convert CamelCase to snake_case
-                name = string.Concat(name.Select(x => char.IsUpper(x) ? $"_{char.ToLower(x)}" : x.ToString()));
-                name = name.Trim('_');
+                // Ignore if all uppercase, assume acronym (e.g. NPC or HTN)
+                if (name.All(char.IsUpper))
```
PJB3005, line 46: Ok but what about systems like `RCDAmmoSystem` tho

mirrorcult: w accept the L for now

PR #4877, opened 2024-01-30
Fix EndCollideEvent being raised when a contact is destroyed but the two entities are still colliding

DrSmugleaf:
I hope there is a better way to do this than looping the contacts dictionary.
@metalgearsloth 

Stopping climbing will make step trigger system think that you are no longer colliding with something that you still are colliding with, so climbing destroys the climbing fixture which destroys any contacts, if one of those contacts happens to be something that step trigger system is tracking it will remove it, if it leaves StepTriggerComponent.Colliding without elements it removes StepTriggerActiveComponent and prevents step trigger system from registering that you are stepping on something until you stop colliding and then start colliding again.
This only happens if the entity you were and still are colliding with shares a collision group with the climbing fixture.


DrSmugleaf:
Replaced by https://github.com/space-wizards/space-station-14/pull/24756

PR #4878, opened 2024-01-31
Replay int overflows, fixed again

### LINKED ISSUE #4801: long cvars are not handled correctly
So my tests are failing locally due to replay.max_compressed_size.

The issue is this is a long cvar, but TypeConvert converts this to an int as it's lower than the max value. The issue is that this tries to get cast from int32 to int64 which isn't valid.

PJB3005:
Fixed the issue with the config system that caused test failure.

Fixes https://github.com/space-wizards/RobustToolbox/issues/4801

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -35,7 +35,7 @@ END TEMPLATE-->
 
 ### Breaking changes
 
-*None yet*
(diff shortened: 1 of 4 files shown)


metalgearsloth:
Fixes https://github.com/space-wizards/RobustToolbox/issues/4801

PR #4883, opened 2024-02-01
Making possible to QueueDeleteEntity on EndCollideEvent

YuriyKiss:
Required by https://github.com/space-wizards/space-station-14/pull/22962

If contact that raises EndCollideEvent causes its own entity to be queued for deletion, it cannot be properly processed by OnContactEntityQueueDel because the contact already has the flag "Deleting", which causes assertion error.


moonheart08:
Just use QueueDelete?

YuriyKiss:
> Just use QueueDelete?

Dear god, I knew there was something to delay/queue deletion. I'll review that but I reaaaaly didn't want to get into explosion system in that PR. Does this change seem drastic enough to not be merged?

YuriyKiss:
Wait does it already do the queue-ing? 

image

metalgearsloth:
Would be better to be an assert because you shouldn't ever be deleting entities during eventbus callbacks.

YuriyKiss:
> Just use QueueDelete?

You're right, it should've been using it and already did. As I figured out now, assert error was caused by EntityManager.EntityQueueDeleted action. Updated PR comment to properly explain it

graevy:
> Just use QueueDelete?

Morgot — 02/01/2024 3:54 AM
hmm, valid point
gravy — 02/01/2024 3:54 AM
honestly
submitting an engine PR? best way to get one of the engine people to answer your question

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
```diff
@@ -186,17 +186,25 @@ public void DestroyContacts(PhysicsComponent body)
...
+            if (!DestroyContact(contact))
+            {
+                contactsFlaggedDeleting++;
+            }
         }
 
-        DebugTools.Assert(body.Contacts.Count == 0);
+        // This contact will be deleted before SimulateWorld runs since it is already set to be Deleted
+        DebugTools.Assert(body.Contacts.Count == contactsFlaggedDeleting);
     }
```
metalgearsloth, line 209: This doesn't make a whole lot of sense to me, by the time the method returns no contacts should be left over.

YuriyKiss: When you call QueueDeleteEntity in EndCollideEvent, one contact is always left in this method since this contact has the flag "Deleting" and cannot be removed

IMO that's not the best approach but the rest of the call stack is as robust as this part is and I couldn't figure out a better place to put this "edge case" check.

In any case, without this or a similar solution - the problem will persist. You will not be able to call QueueDeleteEntity on EndCollideEvent without errors.

metalgearsloth: Is there a way to repro this easily so I can take a better look?

YuriyKiss: My other PR https://github.com/space-wizards/space-station-14/pull/22962 which requires this change has all the prerequisites to reproduce this problem persistently, it happens within 3-4 explosions of a landmine. 

Arteben made a brief comment explaining how it happens and attached the error log: 
> When a regular landmine is triggered, after the message, when you step off this mine, an error occurs.
https://github.com/space-wizards/space-station-14/pull/22962#issuecomment-2005497769

All you have to do is switch onto the branch of space-station for landmines that I referenced and check the landmine behavior using the base engine branch without these changes here.

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs
```diff
@@ -335,13 +334,14 @@ public void DestroyContact(Contact contact)
...
+                SetAwake((aUid, bodyA), true);
 
             if (bodyB.CanCollide)
-                SetAwake(bUid, bodyB, true);
+                SetAwake((bUid, bodyB), true);
         }
 
         // Remove from the world
+        DebugTools.Assert(_activeContacts.Contains(contact));
         _activeContacts.Remove(contact.MapNode);
```
metalgearsloth: Contains is really expensive because it has to iterate the entire linkedlist looking for the contact.

PR #4887, opened 2024-02-03
Tweak RichText

Ertanic:
Small tweak to allow tags to get object state from the game.


Robust.Client/UserInterface/RichText/MarkupTagManager.cs
```diff
@@ -24,7 +24,7 @@ public sealed class MarkupTagManager
         new CommandLinkTag(),
         new FontTag(),
         new HeadingTag(),
-        new ItalicTag()
+        new ItalicTag(),
```
juliangiebel, line 27: why add a comma here?

Robust.Client/UserInterface/RichText/MarkupTagManager.cs
```diff
@@ -39,7 +39,7 @@ public sealed class MarkupTagManager
         typeof(CommandLinkTag),
         typeof(FontTag),
         typeof(HeadingTag),
-        typeof(ItalicTag)
+        typeof(ItalicTag),
```
juliangiebel, line 42: same as above

Robust.Client/UserInterface/RichTextEntry.cs
```diff
@@ -37,18 +37,37 @@ internal struct RichTextEntry
...
+        public RichTextEntry(FormattedMessage message, object tagsState, Control parent, MarkupTagManager tagManager, Type[]? tagsAllowed = null, Color? defaultColor = null)
+        {
+            _tagsState = tagsState;
+            Message = message;
+            _defaultColor = defaultColor ?? new(200, 200, 200);
+            _tagManager = tagManager;
+            _tagsAllowed = tagsAllowed;
+
+            Init(parent);
+        }
```
juliangiebel, line 63: Don't create a private init function and copy paste the constructor. Just add tagsState as another additional parameter

metalgearsloth (commented):
I don't think storing any object on what is essentially text markup is a particularly good idea and see no good reason why paper isn't handling this.

Robust.Client/UserInterface/RichTextEntry.cs
```diff
@@ -57,6 +76,10 @@ public RichTextEntry(FormattedMessage message, Control parent, MarkupTagManager
                 if (node.Name == null)
                     continue;
 
+                if (!node.Attributes.ContainsKey("state")
+                    && _tagsState != null)
+                    node.Attributes.Add("state", new(_tagsState));
+
```
juliangiebel, line 82: Don't save state like that inside the attributes. Pass it through using `MarkupDrawingContext`

Robust.Shared/Utility/MarkupNode.cs
juliangiebel, line 1: This is purely for the serialized data from the markup. I.e. `[testTag testattribute="example"]`

juliangiebel (changes requested):
~Putting the state into the attributes like that isn't good.
Also `tagsState` shouldn't be a single object but rather something like a string object dictionary imo.~

### From the paper PR in content:
The way you're trying to use rich text here makes me think just using string replace <name> with the name, <job> with the job on finishing editing the paper would make much more sense.
RichText is at most supposed to show dynamic data that gets updated when "opening" the text (see the keybind tag)

PR #4892, opened 2024-02-09
Add GetEncoding to Sandbox

FairlySadPanda:
Required for using encoding to filter out unwanted characters from a string.

eg:
```
    /// <summary>
    /// Returns a string with all characters not in ISO-8851-1 replaced with question marks
    /// </summary>
    private static string RemoveNonLatin(string input)
    {
        if (string.IsNullOrEmpty(input))
            return "";

        // I think there's only three valid ones in dotnet core; ascii, iso-8859-1 and utf-8
        // We can extend but it requires annoying whitelisting
        var bytes = Encoding.GetEncoding("iso-8859-1").GetBytes(input);
        return Encoding.UTF8.GetString(bytes);
    }
```

### DIFF
--- a/Robust.Shared/ContentPack/Sandbox.yml
+++ b/Robust.Shared/ContentPack/Sandbox.yml
@@ -626,6 +626,7 @@ Types:
       - "System.Text.Encoding get_UTF7()"
       - "System.Text.Encoding get_UTF8()"
       - "System.Text.Encoding get_UTF32()"
+      - "System.Text.Encoding GetEncoding(string)"
     NormalizationForm: { } # Enum
     Rune: { All: True }
     StringBuilder:


moonheart08:
This feels like a pretty weird way to approach this. Would it not be better to use the unicode tables?

FairlySadPanda:
It's actually the recommended solution for formatting to ISO. 

Staying in UTF-8 land means using ranges, which I was too lazy to look up :D

moonheart08:
> It's actually the recommended solution for formatting to ISO.
> 
> Staying in UTF-8 land means using ranges, which I was too lazy to look up :D

Okay but our goal isn't formatting to iso, it just tangentally happens doing so kindasorta solves the problem you're trying to solve.

PR #4894, opened 2024-02-10
Add UnicodeRange to sandbox

FairlySadPanda:
Need this struct allowlisted to for nice unicode sanitization.


PJB3005 (commented):
Aren't you also going to need [`UnicodeRanges`](https://learn.microsoft.com/en-us/dotnet/api/system.text.unicode.unicoderanges?view=net-8.0)?

PR #4899, opened 2024-02-15
Optimise ChunkEnumerator

metalgearsloth:
It never unioned the AABB passed in with the grid's AABB so it might inadvertantly iterate a lot more dummy chunks than it needs to.

This helps speedup FindGridsIntersecting.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -51,7 +51,7 @@ END TEMPLATE-->
 
 ### Internal
 
-*None yet*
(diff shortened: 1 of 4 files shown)


RELEASE-NOTES.md
```diff
@@ -51,7 +51,7 @@ END TEMPLATE-->
 
 ### Internal
 
-*None yet*
+* Significantly optimise ChunkEnumerator / FindGridsIntersecting in certain use cases by unionising the grid's AABB with the local AABB to avoid iterating dummy chunks.
```
```suggestion
* Significantly optimise ChunkEnumerator / FindGridsIntersecting in certain use cases by intersecting the grid's AABB with the local AABB to avoid iterating dummy chunks.
```

PR #4900, opened 2024-02-15
Add DrawPrimitives overload for List<Vector2>

metalgearsloth:
Storing ValueList in a field seems sussy so this is the next best thing + we don't have to expose it to content.


PJB3005:
> Storing ValueList in a field seems sussy

What's sussy about it? That's perfectly fine as long as you don't pass around the `ValueList` instance itself.

metalgearsloth:
> > Storing ValueList in a field seems sussy
> 
> What's sussy about it? That's perfectly fine as long as you don't pass around the `ValueList` instance itself.

Really more like if it's stored in like a dictionary.

PR #4902, opened 2024-02-15
Invalid keybinds will no longer mess up your game.

### LINKED ISSUE #4898: Keybinds will break your game if you have old loadout (invalid) values
## Description

This was an issue for a while but now became more apparant as of recently.

space-wizards/space-station-14#21352 replaced the bind from "Loadout0-9" to "Hotbar0-9". Players having remendants of "Loadout" in their keybinds expirience issues. I swear i remember having a debug assert months ago about this too

**Reproduction**

https://discord.com/channels/310555209753690112/1176862211482849384/1207242116170780672

1. Have loadout in your keybinds.yml
2. Go to keybindings menu
3. First you will notice all the reset buttons have light up,... this is not normal from what know
4. Press reset
5. Experience issues. if you are on release you wont crash but your settings menu will bug out and potentially other pr's. If you are on debug you will get debug asserted

**Screenshots**

**Additional context**

Fix is to remove the loadout entries or reset your keybinds

space-wizards/space-station-14#25210 releated
https://github.com/space-wizards/RobustToolbox/pull/4841 potencially a culprit for why its worse now. But this always existed apparently since the loadout pr.

image

keybinds to replicate 
[keybinds.yml.txt](https://github.com/space-wizards/space-station-14/files/14281257/keybinds.yml.txt)

nikthechampiongr:
fixes #4898

When a key wasn't valid there would be an exception when attempting to save keybinds. This is because they would be skipped, and not entered into _bindingsByFunction which would cause an exception whenever that key was looked up. 

This pr makes it so those keys are entered into that dictionary but not shown to players. It also means they won't be obliterated when the config saves in case those keys are valid another server.

### DIFF
--- a/Robust.Client/Input/InputManager.cs
+++ b/Robust.Client/Input/InputManager.cs
@@ -517,13 +517,16 @@ namespace Robust.Client.Input
             {
                 var baseKeyRegs = _serialization.Read<KeyBindingRegistration[]>(BaseKeyRegsNode, notNullableOverride: true);
 
+
                 foreach (var reg in baseKeyRegs)
                 {
+                    var invalid = false;
+
                     if (reg.Type != KeyBindingType.Command && !NetworkBindMap.FunctionExists(reg.Function.FunctionName))
                     {
                         Logger.ErrorS("input", "Key function in {0} does not exist: '{1}'", file,
                             reg.Function);
-                        continue;
+                        invalid = true;
                     }
 
                     if (!userData)
@@ -538,7 +541,7 @@ namespace Robust.Client.Input
                         }
                     }
 
-                    RegisterBinding(reg, markModified: userData);
+                    RegisterBinding(reg, markModified: userData, invalid);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


Robust.Client/Input/InputManager.cs
```diff
@@ -523,13 +523,16 @@ public void LoadKeyFile(ResPath file, bool defaultRegistration, bool userData =
...
 
+
                 foreach (var reg in baseKeyRegs)
                 {
+                    var invalid = false;
+
                     if (reg.Type != KeyBindingType.Command && !NetworkBindMap.FunctionExists(reg.Function.FunctionName))
                     {
                         Logger.ErrorS("input", "Key function in {0} does not exist: '{1}'", file,
                             reg.Function);
```
PJB3005: This is an expected condition going forward so shouldn't be reported as an error.
```suggestion
                        Logger.DebugS("input", "Key function in {0} does not exist: '{1}'.", file,
                            reg.Function);
```

PR #4903, opened 2024-02-16
fix default keybinds being applied on top of user set keybinds

### LINKED ISSUE #25309: The Other Keybinds Issue
## Description
As of ~~a few days ago~~ [Robust Toolbox #4898](https://github.com/space-wizards/RobustToolbox/pull/4841), modifying the keybinds from the default causes a glitch that overwrites your control scheme. 
It's possible that the cause of this issue is related to [Robust Toolbox issue #4898](https://github.com/space-wizards/RobustToolbox/issues/4898) as they popped up around the same time
The bug still exists as of commit 0a64e086c

 - Upon any changes to keybinds, every control function that was not assigned a Secondary input will display the default value as it's secondary input. 
 - EVERY control function, even the ones that were assigned a secondary bind and do not show the previous "symptom", will react to their default key being pressed, on top of everything else that is bound. 
 - **This additional set of default keys does not exist in the keybinds file, the game appears to load the defaults on top of what it reads from the keybinds**
 
When the issue manifests, this means that every control input you did not overwrite or remove will also trigger twice on every keypress, once from the Primary Input and once from the "Bug Input".
 - Can't Ready at roundstart! (Every click is doubled, cancelling itself)
 - Esc will no longer open the menu, because the second press immediately closes it. 
 - F11 will toggle fullscreen mode back and forth, putting you where you started. 
 - Crafting menu GUI button won't work for some reason, while the other GUI buttons around still do
 
# Workaround
See my post below!

# Examples
You change WASD to ESDF. After you restart the game, pressing S will move the character both down and to the left. This will be shown as your keybinds:
Screenshot from 2024-02-16 09-46-35

You change WASD to ESDF, and set 1234 as secondary binds. After you restart the game, pressing S will STILL move the character both down and to the left. This will be shown as your keybinds:
Screenshot from 2024-02-16 09-47-39

Errant-4:
Fixes https://github.com/space-wizards/space-station-14/issues/25309

`LoadKeyFile.defaultRegistration `
Whether or not this is a "default" keybind set. If it is, then it won't override the current configuration, only the defaults."

`RegisterBinding.markModified`
Whether or not to register the bind as having been modified

### Current behavior:
if a user keybinds file is detected, it will load changed entries. Every entry will pass "is this a default value?" FALSE, which is then interpreted mark this as modified? FALSE
The bound inputs are not marked as modified.

Then, the default values are loaded. It checks every bind to see if it has been marked as modified. Whether or not a user keybinds file was loaded, nothing will be marked as modified. Thus, every default value will be loaded. If no user keybinds were loaded, this will work as a proper default configuration. If they were, the defaults are loaded in addition to the user configured binds.

Either way, every binding registration call will also pass "is this a default value?" TRUE, which is then interpreted "mark this as modified"? TRUE 

You can confirm this without even looking at the code, by observing that no matter what happened, even if there was no user keybinds file and its all defaults, every entry will have the RESET button next to it active and pressable

**The problem: defaultRegistration is passed directly to markModified, when it should be inverted**

Inverting this one value pass seems to fix the keybinds behavior, and I have not been able to find any issues caused by this. 
I think this is how the whole thing was designed to work, and simply one of the two booleans was assigned a logically opposite "phrasing" than the code was written by mistake

# Please check my logic I am not to be trusted with engine code

### DIFF
--- a/Robust.Client/Input/InputManager.cs
+++ b/Robust.Client/Input/InputManager.cs
@@ -544,7 +544,7 @@ namespace Robust.Client.Input
                         }
                     }
 
-                    RegisterBinding(reg, markModified: defaultRegistration);
+                    RegisterBinding(reg, markModified: !defaultRegistration);
                 }
             }
 


moonheart08:
oops

Issue #4803, opened 2024-01-03
Sprite rendering should be based around layers not sprites


metalgearsloth:
Consider that sprite rendering, at least from the cpu end, is a similar problem space to physics + dynamictrees:

Ideally:
- Each drawdepth is on its own dynamictree. We have far fewer of those than collision layers so we would probably only have 20 or 30 and not 100.
- When we get sprites to draw we can sort per dynamictree and not globally which is much faster and means we can dispatch sprites for drawing much faster.
- Cleaner separation when we want to do overlays in-between drawdepths.
- Per-layer drawdepths is useful in some rare cases where you really don't want 2 entities.

AndrewL97:
Interested in this - I've been playing around with this concept, but struggling to get dynamictrees to click in my head.

@metalgearsloth  is there a good resource/example that I can look at to get started on this?
My super basic implementation still uses the spriteComponentTree, and just adds handling to spin up spriteData for each drawDepth override on a layer in [here](https://github.com/space-wizards/RobustToolbox/blob/ac60567583aec7edddd75425bf4f36f69a7b6439/Robust.Client/Graphics/Clyde/Clyde.Sprite.cs#L81-L89), and then a dodgy drawDepth check in the sprite's renderInternal

PR #4905, opened 2024-02-17
Add support for drawdepths per sprite layer

EmoGarbage404:
scrungles some sprite code to enable individual layers to specify drawdepths.

Main change is that ProcessSpriteEntities now gets and sorts individual sprite layers rather than entire sprites.


ElectroJr:
Why do we need to have per-layer draw depths? I know its mentioned in #4803, but why is it actually useful, and how would it work with shaders or eventual support for advanced sprite rendering options (layer groups / byond style "draw together", or  drawing an entity's sprite within another entity's layers, etc)? And would it be worthwhile given it would probably appreciably slow down the sorting step while rendering?

I do think layers should have a depth/sorting (e.g. #2970), but I'm not sure about having them sort across different entities

moonheart08:
> Why do we need to have per-layer draw depths? I know its mentioned in #4803, but why is it actually useful, and how would it work with shaders or eventual support for advanced sprite rendering options (layer groups / byond style "draw together", or drawing an entity's sprite within another entity's layers, etc)? And would it be worthwhile given it would probably appreciably slow down the sorting step while rendering?
> 
> I do think layers should have a depth/sorting (e.g. #2970), but I'm not sure about having them sort across different entities

Chairs are an easy example, currently they're missing an overlay they're supposed to have which would make it so the armrest draws over the player.

ElectroJr:
> Chairs are an easy example, currently they're missing an overlay they're supposed to have which would make it so the armrest draws over the player.

I think that should be done by allowing a sprite to get drawn as a layer of some other sprite, as mentioned in my previous comment. I.e., the whole mob should just become a layer within the chair's sprite. If you just changed the armrest's drawdepth to get drawn over mobs, it would be drawn over any mob that moves near the chair, not just the mob that is currently in the chair.

My vague wishlist/todolist for sprites is:
- Move logic out of sprite component and into the system
- Move clickmaps from content to engine
- Rework `ISpriteLayer` and add layer groups (collections of other layers or layer groups), and allow each group to have its own shader & color modulation 
- Change sprites component to basically just be a kind of layer group, and allow sprites to use the sprites of other entities as a layer.
  - Would requires somehow updating clickmaps so that you can still click on an entity that is in a chair.
- Add support for multiple "PostShader"s
- Add direction dependent drawdepths for 4/8-dir sprites to fix things like inhands being rendered incorrectly

AndrewL97:
@ElectroJr / @moonheart08 This is my implementation of it -> https://github.com/space-wizards/RobustToolbox/pull/4917

Essentially.

- Rather than gathering all of the sprites in a single CompontentTree, we have a "layered" dictionary of ComponentTrees
- Layers in the SpriteComponents track their own draw depth, inheriting the previous layer's drawdepth if no override is set
- When we add sprites to the SpriteComponentTree, it is split into the layer specific ComponentTrees


super keen for feedback if you have any :)

Doctor-Cpu:
Is there anything significant blocking this?
Would be useful on downstreams.

EmoGarbage404:
not good code and not a good approach to be worth really considering

PR #4910, opened 2024-02-21
Add more CollectionExtension methods

metalgearsloth:
Maybe array.resize is bad for sandbox coin, in which case I'd also settle for changing it to a list instead.

### DIFF
--- a/Robust.Shared/Utility/CollectionExtensions.cs
+++ b/Robust.Shared/Utility/CollectionExtensions.cs
@@ -8,9 +8,27 @@ namespace Robust.Shared.Utility
 {
     public static class Extensions
     {
+        /// <summary>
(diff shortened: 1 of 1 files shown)


Robust.Shared/Utility/CollectionExtensions.cs
```diff
@@ -8,9 +8,27 @@ namespace Robust.Shared.Utility
 {
     public static class Extensions
     {
+        /// <summary>
+        /// Ensures that the specified array has the specified length.
+        /// </summary>
+        public static void EnsureLength<T>(ref T[] array, int length)
```
PJB3005, line 14: This function is in an awkward spot because the `Extensions` class it's in is poorly named, and this isn't an extension method (no `this`). Should probably just make it a `this` function.

PR #4911, opened 2024-02-21
Add more ValueList conveniences

metalgearsloth:
Should also speedup a couple things on contentTM.


Robust.Shared/Collections/ValueList.cs
```diff
@@ -50,6 +52,44 @@ public ValueList(int capacity)
...
+    {
+    }
+
+    /// <summary>
+    /// Creates a list by copying the contents of the source list.
+    /// </summary>
+    public ValueList(List<T> list, int start, int count)
+    {
+        Array.Resize(ref _items, count);
+        Capacity = count;
```
PJB3005: `Capacity` is directly the size of the underlying array, so setting it after the array resize doesn't matter.

That said, I'd just replace this with a `_items = new T[count];` to clearly specify intent.

Robust.Shared/Collections/ValueList.cs
```diff
@@ -50,6 +52,44 @@ public ValueList(int capacity)
...
+
+        var liSpan = CollectionsMarshal.AsSpan(list);
+        var target = new Span<T>(_items, start, count);
+        liSpan.CopyTo(target);
```
PJB3005: This doesn't make much sense to me. Usually when you have a function taking `List<T> list, int start, int count` I would expect the `start` and `count` to be in the source list, not the destination (which is what this code does).

Assuming that's intentional, this code is bugged because the array capacity would need to be allocated to `start + count`, not `count`.

Robust.Shared/Collections/ValueList.cs
```diff
@@ -522,4 +568,49 @@ public T RemoveSwap(int index)
...
+        if (Count > newCount)
+            return;
+
+        EnsureCapacity(newCount);
+
+        for (var i = Count; i < newCount; i++)
+        {
+            var size = Count;
+            AddNoResize(default!, size);
+        }
```
PJB3005: You could get the span of the new region instead and `.Clear()` it. Much faster.

PJB3005 (changes requested):

Robust.Shared/Collections/ValueList.cs
metalgearsloth: oop

PR #4915, opened 2024-02-21
Rework SetWorldPosition

metalgearsloth:
How it currently works:
- Tries to set the entity's LocalPosition to the specified spot, transforming the new WorldPosition to something relative to their current parent. Problem with this is is it doesn't work with containers or if an entity gets moved off grid for grid traversal so content can very easily break this.

How I've changed it to work:
- It will de-parent and function as if the caller called AttachToGridOrMap immediately after which better aligns with expectations.

If someone wants the old method then either add a new method called "TrySetWorldPosition" that will drop it if they're in a container or do the matrix transform yourself and set localposition.

I have also added a test to verify this works going forward and doesn't break again.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
@@ -1004,9 +1004,29 @@ public abstract partial class SharedTransformSystem
             return;
         }
 
-        var (curWorldPos, curWorldRot) = GetWorldPositionRotation(component, xformQuery);
+        var (curWorldPos, curWorldRot) = GetWorldPositionRotation(component);
         var negativeParentWorldRot = component._localRotation - curWorldRot;
         var newLocalPos = component._localPosition + negativeParentWorldRot.RotateVec(worldPos - curWorldPos);
+
+        // Need to remove them from a container if they were in one directly.
+        if (_container.TryGetContainingContainer(component.Owner, out var container, transform: component))
+        {
+            EntityCoordinates destination;
+
+            if (component.MapUid != null && _mapManager.TryFindGridAt(component.MapUid.Value, worldPos, out var targetGrid, out _))
+            {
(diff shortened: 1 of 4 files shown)


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -1097,20 +1086,22 @@ public void SetWorldPositionRotation(EntityUid uid, Vector2 worldPos, Angle worl
...
 
-        var (curWorldPos, curWorldRot) = GetWorldPositionRotation(component);
-
-        var negativeParentWorldRot = component.LocalRotation - curWorldRot;
-
-        var newLocalPos = component.LocalPosition + negativeParentWorldRot.RotateVec(worldPos - curWorldPos);
-        var newLocalRot = component.LocalRotation + worldRot - curWorldRot;
-
-        SetLocalPositionRotation(uid, newLocalPos, newLocalRot, component);
+        if (!_gridQuery.HasComponent(uid) && _mapManager.TryFindGridAt(component.MapUid.Value, worldPos, out var targetGrid, out _))
```
ElectroJr:
```suggestion
        if (component.GridUid != uid && _mapManager.TryFindGridAt(component.MapUid.Value, worldPos, out var targetGrid, out _))
```

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -1097,20 +1086,22 @@ public void SetWorldPositionRotation(EntityUid uid, Vector2 worldPos, Angle worl
...
+        {
+            var (_, gridRot, invWorldMatrix) = GetWorldPositionRotationInvMatrix(targetGrid);
```
ElectroJr, line 1097: Unless we want to support grid stacking/nesting, you can just use local instead of world

PR #4917, opened 2024-02-22
Handle Layered Component Trees for per sprite-layer drawdepths

### LINKED ISSUE #4803: Sprite rendering should be based around layers not sprites

AndrewL97:
About the PR
Create a new LayeredComponentTreeComponent that allows LayeredComponentTreeSystems to handle a dictionary/layers of binary trees, rather than a single flat one.

Why / Balance
Fixes #4803
Allows us to have sprites that draw layers on different drawdepths. So entities can pass "between" another entity's layers. Think gates, doorways, etc.

Technical details
Added a LayeredComponentTreeSystem that lets us handle splitting Component Trees into Layers.

Changed SpriteTreeSystem to use the new layered system.

Added handling to the Sprite component and rendering pipeline to fetch and draw based on drawdepth/layer.
The sprite and index lists are now dictionaries, so the individual sorts don't need to handle drawDepth anymore.

the Layers config on sprite components now lets you define a drawDepth. If we don't define one, we default to the previous layer's drawDepth, or parent Sprite's if no other layers exist.
No changes have been made to the existing entities, so there shouldn't actually be any visible changes from this PR.

Changelog

tweak: Added handling for drawing sprite layers on a drawdepth separate to their parent sprite.


AndrewL97:
@metalgearsloth this is my first foray into rendering _and_ SS14, so please let me know if I have botched standards, etc.
Very much open to feedback <3

Robust.Shared/Utility/CollectionExtensions.cs
```diff
@@ -40,7 +40,8 @@ public static List<T> ShallowClone<T>(this List<T> self)
 
         public static bool TryGetValue<T>(this IList<T> list, int index, out T value)
         {
-            if (list.Count > index)
+
+            if (list.Count > index && index >= 0)
             {
                 value = list[index];
```
metalgearsloth, line 63: You should avoid autoformatting stuff like this as it just makes the review harder to read.

AndrewL97: Fixed!

metalgearsloth:
I think I like this more than https://github.com/space-wizards/RobustToolbox/pull/4905 and would rather pull emo's ideas in but I have a content backlog of stuff I really need to go through first.

AndrewL97:
Hey @metalgearsloth  - is this something you guys are still keen for? There are conflicts for me to fix now, but I'll hold off unless this is something we want :)

metalgearsloth:
Electro left some comments here on emo's PR https://github.com/space-wizards/RobustToolbox/pull/4905#issuecomment-2066861749

Moving sprite rendering to another entity's responsibility would probably greatly simplify the logic involved here.

AndrewL97:
Easy peasy, I'll close this one and have a squizz

PJB3005:
FWIW I've been meaning to review this for ages, I just haven't gotten around to it yet.

AndrewL97:
@PJB3005  
> FWIW I've been meaning to review this for ages, I just haven't gotten around to it yet.

All G - I haven't looked at this since my last commit anyway, so I'll restart with Electro's comment in mind
Not 100% sure how the entity sprite layer combination stuff will work in practice, but I'll annoy people in Discord about it

PR #4918, opened 2024-02-22
Hopefully fix sudo

### LINKED ISSUE #4828: "sudo" command no longer autocompletes
repro
1. Launch ss14
2. open console
3. write a command starting with "sudo" you wont get autocomplete result like usual
Content Client_AkRnOyAFsv

DebugOk:
Oh god no template

I have no idea what I'm really supposed to put here, or if this is the correct approach, but the lack of sudo autocompletion is kinda annoying.

This works, but I'm not sure if its the way to go? At least it was the easiest way for me to do it

Fixes #4828

### DIFF
--- a/Robust.Server/Console/ServerConsoleHost.cs
+++ b/Robust.Server/Console/ServerConsoleHost.cs
@@ -260,11 +260,9 @@ namespace Robust.Server.Console
 
             done:
 
-            result ??= CompletionResult.Empty;
-
             var msg = new MsgConCompletionResp
             {
-                Result = result,
+                Result = result ?? CompletionResult.Empty,
                 Seq = message.Seq
             };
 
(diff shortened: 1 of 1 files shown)


DebugOk:
Another thing to stripe off my PR something to all space wizards repos challenge /s

PJB3005:
Yeah so this definitely is not the correct way to fix this. There already was command completion code for sudo, but it broke somewhere. I just pushed the correct fix: a77eee56583ef77373f3d34a1d37841a53d158ea

This PR did get me off my ass to look into it though.

DebugOk:
> Yeah so this definitely is not the correct way to fix this. There already was command completion code for sudo, but it broke somewhere. I just pushed the correct fix: [a77eee5](https://github.com/space-wizards/RobustToolbox/commit/a77eee56583ef77373f3d34a1d37841a53d158ea)
> 
> This PR did get me off my ass to look into it though.

Thank you. I figured as much, but I'm not very familiar with the engine side of thing

PR #4921, opened 2024-02-24
Add basic audio limits

metalgearsloth:
Rudimentary but stops mapinit lights killing my debug game.


PJB3005:
This doesn't really seem like a good solution to the core issue, but I'm not gonna let perfect be the enemy of good here.

metalgearsloth:
> This doesn't really seem like a good solution to the core issue, but I'm not gonna let perfect be the enemy of good here.

I don't mind something else but I can't even load in maps without my game triggering debug asserts due to lights activating.

PR #4925, opened 2024-02-25
Increase replay compressed size default cvar

VasilisThePikachu:
They tend to get cut off (or well did on wizden before pjb changed it to this exact value bigger along with another patch). 

Before you got around 2ish hours in a replay before it stopped suddenly. 

I doubt most servers will reach 6ish hours before this takes effect. But those servers can increase this value if needed.


PJB3005 (commented):
This isn't the only reason they were getting cut off by replay.max_uncompressed_size too.

VasilisThePikachu:
> This isn't the only reason they were getting cut off by replay.max_uncompressed_size too.

The value is the same as the one you provided in https://discord.com/channels/310555209753690112/1010977074741252196/1209260935085166652

PR #4927, opened 2024-02-26
chore: Remove some typos

Brandon-Huu:
Removed some typos


Resources/Locale/en-US/toolshed-commands.ftl
```diff
@@ -418,6 +418,6 @@ command-description-tee =
     This essentially lets you have a branch in your code to do multiple operations on one value.
 command-description-cmd-info =
     Returns a CommandSpec for the given command.
-    On it's own, this means it'll print the comamnd's help message.
+    On it's own, this means it'll print the command's help message.
```
PJB3005:
```suggestion
    On its own, this means it'll print the command's help message.
```

PR #4931, opened 2024-02-29
Implementing random selection helpers in robust.random

Jezithyr:
Added SelectRandom to robust random to allow for random selection from arrays.
SelectRandom will throw if the array is empty, for safe selection use SelectRandomOrDefault

### DIFF
--- a/Robust.Shared/Random/RobustRandom.cs
+++ b/Robust.Shared/Random/RobustRandom.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using Robust.Shared.Utility;
 
 namespace Robust.Shared.Random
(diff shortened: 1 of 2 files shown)


Robust.Shared/Random/RobustRandom.cs
```diff
@@ -54,5 +55,19 @@ public void NextBytes(byte[] buffer)
...
+
+        public T SelectRandom<T>(List<T> list) where T : new()
+        {
+            if (list.Count == 0)
+            {
+                throw new ArgumentException("Specified List does not have any elements");
+            }
+            return list[Next(list.Count)];
+        }
     }
```
metalgearsloth, line 72: Isn't this just Pick?

mirrorcult:
this is just `Pick()`

PR #4939, opened 2024-03-03
Clip labels by default

metalgearsloth:
I imagine situations where you don't want to clip these are rare. I mainly noticed this on FancyWindow and Button but if we don't want it globally I'll just update it as I see it.


PJB3005:
This would break layout everywhere. Disabling clipping means changing labels to not demand space, and there's really no way to fix that.

PR #4941, opened 2024-03-04
Make CVar light.resolution_scale only changeable by server

notquitehadouken:
remove `CVar.CLIENTONLY` and add the `CVar.SERVER | CVar.REPLICATED`

reason:
`cvar light.resolution_scale 0.0001`
before:
image
after:
image

the rest of the light cvars don't do anything to warrant making them server-changeable only

### DIFF
--- a/Robust.Shared/CVars.cs
+++ b/Robust.Shared/CVars.cs
@@ -702,7 +702,7 @@ namespace Robust.Shared
         /// relative to the viewport framebuffer size.
         /// </summary>
         public static readonly CVarDef<float> LightResolutionScale =
-            CVarDef.Create("light.resolution_scale", 0.5f, CVar.CLIENTONLY | CVar.ARCHIVE);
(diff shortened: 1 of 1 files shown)


moonheart08 (changes requested):
This is stupid. Implement content-side sanity checks, do not break the ability for low-end users to down-res lighting.

In the future, actually review how a CVar is used and what it does before kneejerk breaking the game.

PJB3005 (changes requested):
Change the code that reads this CVar to instead clamp it to a minimal value, so it can't go below 0.05 or something.  This setting controls the game's graphics settings so it has to be settable by clients.

notquitehadouken:
> Change the code that reads this CVar to instead clamp it to a minimal value, so it can't go below 0.05 or something. This setting controls the game's graphics settings so it has to be settable by clients.

got it (this was my second idea)

PR #4948, opened 2024-03-10
Add a Vertical tab container

metalgearsloth:
Draft because it still looks skrunkly and want to align it with the horizontal one a bit. Also features none of the fake-button handling of TabContainer.


Robust.Client/UserInterface/Controls/VerticalTabContainer.xaml.cs
```diff
@@ -0,0 +1,97 @@
...
+            else
+            {
+                _currentControl = null;
+            }
+        }
+
+        base.ChildRemoved(child);
+    }
+
+    private void SetTab(Control control)
```
ShadowCommander:
```suggestion
    private void SelectTab(Control control)
```

I suggest renaming it to select for consistency with other controls.

Robust.Client/UserInterface/Controls/VerticalTabContainer.xaml.cs
```diff
@@ -0,0 +1,97 @@
...
+        _controls.Remove(child);
+        if (_tabs.Remove(child, out var button))
+        {
+            button.Dispose();
+        }
+
+        // Set the current tab to a different control
+        if (_currentControl == child)
+        {
+            var previous = _controls.IndexOf(child);
```
ShadowCommander: Doesn't this always return `-1` since `_controls.Remove(child);` was called earlier?

ShadowCommander (changes requested):

PR #4949, opened 2024-03-10
Add ToValueList

metalgearsloth:
Better than .ToArray getting used in most cases. Didn't use the ctor for it so this is more of a drop-in replacement for the existing calls and makes the changeover easier.


metalgearsloth:
Will rethink this.

PR #4954, opened 2024-03-12
Add decimal variable to Range Control rounding

ShadowCommander:


Robust.Client/UserInterface/Controls/Range.cs
```diff
@@ -86,6 +87,16 @@ public virtual bool Rounded
             }
         }
 
+        public virtual int RoundingDecimals
```
PJB3005: Does this need to be virtual?

ShadowCommander: No it doesn't need to be virtual. It's from copy pasting from the above.

PR #4959, opened 2024-03-15
Add collection parsing to the dev window for UI

ShadowCommander:
Added conversions from string and control collections to strings.
Added popup when clicking a value to show value.

Before:
Content Client_PkWsZip1Pr
After:
Content Client_lmuECDZxcP
Content Client_fVL3wGEn29

Before:
Content Client_vhOycJ2nSn
After:
Content Client_z14m147VvN


Robust.Client/UserInterface/DevWindow/DevWindowTabUIPopup.xaml.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Robust.Client.AutoGenerated;
+using Robust.Client.UserInterface.Controls;
+using Robust.Client.UserInterface.XAML;
+
+namespace Robust.Client.UserInterface;
+
+[GenerateTypedNameReferences]
+public sealed partial class DevWindowTabUIPopup : Popup
```
PJB3005: Make this `internal`.

Robust.Client/UserInterface/Controls/PopupContainer.cs
```diff
@@ -30,6 +30,7 @@ public sealed class PopupContainer : Control
         public PopupContainer()
         {
             RectClipContent = true;
+            MouseFilter = MouseFilterMode.Ignore;
```
PJB3005: The default for `MouseFilter` is already `Ignore`. Changing this shouldn't do anything?

Robust.Client/UserInterface/DevWindow/DevWindowTabUI.xaml.cs
```diff
@@ -208,26 +213,58 @@ private void Refresh()
...
+                        }
                     }
-                });
+                };
+                button.OnPressed += _ =>
+                {
+                    _popup ??= new DevWindowTabUIPopup();
+                    _popup.Text = GuiDumpCommand.PropertyValuesString(SelectedControl, prop);
+                    if (_popup.Parent != PopupContainer)
+                        PopupContainer.AddChild(_popup);
```
PJB3005: Why not just create `_popup` from XAML so it's already in `PopupContainer`?

Robust.Client/UserInterface/DevWindow/DevWindowTabUI.xaml
```diff
@@ -20,10 +20,11 @@
...
                     </PanelContainer>
                 </ScrollContainer>
-                <ScrollContainer HScrollEnabled="False">
+                <ScrollContainer>
                     <BoxContainer Name="ControlProperties" Orientation="Vertical" MouseFilter="Stop" />
                 </ScrollContainer>
             </SplitContainer>
         </BoxContainer>
     </PanelContainer>
+    <PopupContainer Name="PopupContainer" Access="Public" />
```
PJB3005: I don't think I want to see this pattern of "making a custom `PopupContainer` become commonplace. I do understand it's because secondary OS windows don't have one themselves, but I'd rather see that resolved in `CreateWindowRoot`. You'd probably need to give `WindowRoot` a `PopupContainer` and then also some form of "`ContentRoot`" that further code is advised to use instead of directly adding to the `WindowRoot`.

Robust.Client/UserInterface/DevWindow/DevWindowTabUI.xaml.cs
```diff
@@ -208,26 +213,58 @@ private void Refresh()
...
+        if (!disposing)
+            return;
+
+        if (_popup != null)
+        {
+            PopupContainer.RemoveChild(_popup);
+            _popup.Dispose();
+            _popup = null;
+        }
+    }
```
PJB3005: `Dispose` on controls is soft-deprecated. Use `ExitedTree()` instead for closing the popup.

Robust.Client/UserInterface/DevWindow/DevWindowTabUI.xaml.cs
```diff
@@ -208,26 +213,58 @@ private void Refresh()
...
                             }
-                        },
-                        new Label { Text = $"{value}" },
+                        }
                     }
-                });
+                };
+                button.OnPressed += _ =>
+                {
+                    _popup ??= new DevWindowTabUIPopup();
```
PJB3005: The same line in `InitializeComponent()` will always have happened first.

PJB3005 (changes requested):

Robust.Client/UserInterface/DevWindow/DevWindowTabUI.xaml.cs
ShadowCommander: That'll work for now. I'll see if I can make a PopupContainer on WindowRoot in another PR.

Robust.Client/UserInterface/DevWindow/DevWindowTabUI.xaml
ShadowCommander: I'll see if I can add a PopupContainer to WindowRoot in another PR.

PJB3005 (approved):

PR #4966, opened 2024-03-16
Implement sound VV

metalgearsloth:
No params yet because I'm lazy but paths and collections work.

https://github.com/space-wizards/RobustToolbox/assets/31366439/60fa676f-0d46-43ab-94da-ccf48c823f4d

Video was the WIP version but gives you the idea.


deltanedas:
vv if rust enum existed :sob:

Robust.Client/ViewVariables/ClientViewVariablesManager.cs
```diff
@@ -222,6 +223,13 @@ public VVPropEditor PropertyFor(Type? type)
                 return new VVPropEditorTimeSpan();
             }
 
+            if (typeof(SoundSpecifier).IsAssignableFrom(type))
+            {
+                var control = new VVPropEditorSoundSpecifier();
+                IoCManager.InjectDependencies(control);
```
deltanedas: wouldnt it be better to have this in the constructor of the control

PJB3005: Too late other code already code it.

Robust.Server/Audio/AudioSystem.cs
```diff
@@ -80,6 +80,9 @@ public override (EntityUid Entity, AudioComponent Component)? PlayGlobal(string
     /// <inheritdoc />
     public override (EntityUid Entity, AudioComponent Component)? PlayEntity(string filename, Filter playerFilter, EntityUid uid, bool recordReplay, AudioParams? audioParams = null)
     {
+        if (string.IsNullOrEmpty(filename))
```
PJB3005, line 83: Why these changes?

Robust.Client/ViewVariables/Editors/VVPropEditorSoundSpecifier.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        var editBox = new LineEdit()
+        {
+            HorizontalExpand = true,
+            Editable = !ReadOnly,
+        };
+
+        var controls = new BoxContainer()
+        {
+            Orientation = BoxContainer.LayoutOrientation.Vertical,
```
PJB3005: Wouldn't this make way more sense horizontally?

Robust.Server/Audio/AudioSystem.cs
metalgearsloth: SoundSpecifier being passed in as null already does it and this is the functional equivalent. It's mainly if somehow an empty string gets passed in (e.g. datafield or admin input) it doesn't unnecessarily create an audio source that will never be valid.

I could just make it an exception instead?

Robust.Client/ViewVariables/Editors/VVPropEditorSoundSpecifier.cs
metalgearsloth: Whenever audioparams readwrite gets added I would've assumed it'd be below but I guess it can probably squeeze in?

PJB3005: I was asking because you have `HorizontalExpand` on the input box which doesn't do anything if it's set to vertical. Whatever though.

Robust.Server/Audio/AudioSystem.cs
PJB3005: If that's the case this API should be made nullable to match the sound specifier API.

PR #4968, opened 2024-03-16
Add audio filepath completion helper

metalgearsloth:
Due to how audio is packaged server doesn't have most audio files.


Robust.Shared/ContentPack/ResourceManager.cs
```diff
@@ -177,14 +177,16 @@ public bool TryContentFileRead(string path, [NotNullWhen(true)] out Stream? file
...
             if (path == null)
             {
-                throw new ArgumentNullException(nameof(path));
+                return false;
             }
 
             if (!path.Value.IsRooted)
             {
-                throw new ArgumentException($"Path '{path}' must be rooted", nameof(path));
+                return false;
```
PJB3005: This really shouldn't be tolerated. It turns error on invalid input into silent failure.

I don't know why the hell this function's marked as nullable though.

Robust.Shared/Console/CompletionHelper.cs
```diff
@@ -23,6 +25,28 @@ public static class CompletionHelper
     public static IEnumerable<CompletionOption> Booleans => new[]
         { new CompletionOption(bool.FalseString), new CompletionOption(bool.TrueString) };
 
+    /// <summary>
+    /// Special-cased file handler for audio that accounts for serverside completion.
+    /// </summary>
+    public static IEnumerable<CompletionOption> AudioFilePath(string arg, IPrototypeManager protoManager,
```
PJB3005, line 32: These helper functions are all designed so that they allow directory-level navigation, 
image
This doesn't do this anymore, it would just list all the files in one big list.

Robust.Shared/ContentPack/ResourceManager.cs
metalgearsloth: Should callers just handle user input then? Because admins will be able to write to it directly.

PJB3005: Yes.

PJB3005:
image
does this actually work now? I checked out the PR and this happens.

metalgearsloth:
> does this actually work now? I checked out the PR and this happens.

Dahell, let me check again coz it was autocompleting both for me.

PJB3005:
I am an idiot that might have forgotten to test the content PR

PR #4974, opened 2024-03-17
Replay exception tolerance tweaks

ElectroJr:
- Makes the hash tolerance added in 86ecfaa use the existing error tolerance cvar
- Makes partial/incomplete replays not error, now it just logs a warning.
- Makes the error-tolerance cvar not archived

See also: https://github.com/space-wizards/space-station-14/pull/26212


PJB3005:
Yeah ok that's a better way to do it huh.

PR #4975, opened 2024-03-17
Feature/refine robust random and cleanup

Fildrance:
Refined xml-doc for **`IRobustRandom`**
Changed **`RobustRandom`** namespace to be file-scoped
Added new methods to **`RandomExtensions`** for getting list of random items from collections:
```
    IReadOnlyCollection<T> GetItems<T>(this IRobustRandom random, IList<T> collection, int count, bool allowDuplicates = true);

    ValueList<T> GetItems<T>(this IRobustRandom random, ValueList<T> collection, int count, bool allowDuplicates = true);

    Span<T> GetItems<T>(this IRobustRandom random, Span<T> collection, int count, bool allowDuplicates = true);
```
Main scenarios for new methods:

- Need to pick item random from list of items
- Need to pick several random items from list, only unique
- Need to pick items from list several times, duplicates allowed.

Separated refactoring and adding code into separate commits for comfortable code-review.


Robust.Shared/Random/RandomExtensions.cs
```diff
@@ -133,6 +134,147 @@ public static bool Prob(this IRobustRandom random, float chance)
...
+        }
+
+        if (allowDuplicates == false && count >= collection.Count)
+        {
+            var arr = collection.ToArray();
+            random.Shuffle(arr);
+            return arr;
+        }
+
+        var maxCollectionIndex = collection.Count - 1;
```
ElectroJr: IIRC the random.Next() upper bound is exclusive, so you don't need the -1?

Robust.Shared/Random/RandomExtensions.cs
```diff
@@ -133,6 +134,147 @@ public static bool Prob(this IRobustRandom random, float chance)
...
+        {
+            var index = random.Next(maxCollectionIndex);
+            if (!allowDuplicates)
+            {
+                while (!selectedIndexes.Add(index))
+                {
+                    index = index == collection.Count
+                        ? 0
+                        : index + 1;
+                }
```
ElectroJr: I'm fairly sure this won't lead to a random selection if allowDuplicates is false, because it will preferentially select items that are adjacent to an already selected item. E.g., if you have a list {a, b, c, d}, and the first item picked is a, then for the next item there's a 50% chance to pick b instead of c or d.

It's probably better to do something like:
```cs
var tmp = collection.ToArray();
        
for (int i = 0; i < count; i++)
{
	var j = random.Next(tmp.Length - i);
	rolled[i] = tmp[j];
	tmp[j] = tmp[tmp.Length - i - 1]
}
```
I.e., choosing randomly from a list, then swapping with the item at the back of the list and pretending the list is 1 shorter.


Also, however it ends up being implemented, it should avoid creating a list/hashset unless required by the `allowDuplicates` option,

Robust.Shared/Random/RandomExtensions.cs
Fildrance: True, i missed that point. Fixed that, gonna double-check documentation for IRobustRandom that i wrote, i feel it needs fixing on that part too...

ElectroJr (changes requested):

Robust.Shared/Random/RandomExtensions.cs
Fildrance: Ah, solution is simple! i just need to roll for 1 less each time - this way there won't be 'bonus-chance' for first element after already picked block, it'll be more like virtual mapping!

Fildrance: fixed

Fildrance:
@ElectroJr can u pls close comments if they are fixed correctly?

metalgearsloth:
Just bulldoze https://github.com/space-wizards/RobustToolbox/pull/4980

I woulda merged yours first but wasn't sure on some changes.

metalgearsloth (changes requested):
Needs release notes.

Fildrance:
> Needs release notes.

Done (question mark)

Fildrance:
> Just bulldoze #4980
> 
> I woulda merged yours first but wasn't sure on some changes.

merged as well as i could

ElectroJr (approved):
I pushed some changes including:
- Reducing some code duplication by having the ValueList implementation reuse the Span  one
- Changing the return type to `T[]` instead of things like `IReadOnlyCollection<T>`.
- Removing some of the "This is O(n)." comments, because AFAIK they were incorrect
- Removing an old internal shuffle extension that is no longer required after dotnet 8

PR #4978, opened 2024-03-18
Add debug assert to Dirty(uid, comp)

ElectroJr:


Tayrtahn (approved):
I used this while doing code cleanup, and it's very helpful!

PR #4980, opened 2024-03-19
Add 2 random methods

metalgearsloth:
One for valuelist shuffle and one for NextAngle range.


PJB3005 (commented):
That's a Pick, not a Shuffle 🤔.

PR #4981, opened 2024-03-20
Code cleanup: Purge obsolete MapManager methods

Tayrtahn:
Removes every use of obsolete MapManager methods (GetGrid, GridExists, and TryGetGrid), replacing them with appropriate EntityManager methods or their corresponding EntitySystem extension methods.

This is definitely low -hanging fruit on the code cleanup tree, but clearing out compiler warnings is always nice.


mirrorcult:
blessed

PR #4988, opened 2024-03-23
Add a method in SharedTransformSystem for swapping the position of two entities

EmoGarbage404:
Common content pattern that breaks frequently due to oversights (using entity coordinates, not respecting containers, etc)

The method simply takes two entities and swaps their world positions, moving them across maps. This hopefully will reduce issues in content where bad transform calls result in unintended parenting to various entities.


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -1511,4 +1511,39 @@ public void PlaceNextTo(Entity<TransformComponent?> entity, Entity<TransformComp
                 PlaceNextTo((entity, xform), targetXform.ParentUid);
         }
     }
+
+    public bool SwapPositions(Entity<TransformComponent?> entity1, Entity<TransformComponent?> entity2)
+    {
+        if (!Resolve(entity1, ref entity1.Comp) || !Resolve(entity2, ref entity2.Comp))
+            return false;
+
```
metalgearsloth: Use the xformquery for the resolve.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -1511,4 +1511,39 @@ public void PlaceNextTo(Entity<TransformComponent?> entity, Entity<TransformComp
...
+        SetParent(entity1, _mapManager.GetMapEntityId(pos2.MapId));
+        SetWorldPosition(entity1.Comp, pos2.Position);
+        if (container2 != null)
+            _container.Insert(entity1, container2);
+
+        SetParent(entity2, _mapManager.GetMapEntityId(pos1.MapId));
+        SetWorldPosition(entity2.Comp, pos1.Position);
+        if (container1 != null)
+            _container.Insert(entity2, container1);
+
```
metalgearsloth: As is this is going to send out up to 3 moveevents for each entity when you can cut it down to 1 with branching: either setmapcoordinates (see https://github.com/space-wizards/RobustToolbox/pull/4915 for idea) or container insertion.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -1511,4 +1511,39 @@ public void PlaceNextTo(Entity<TransformComponent?> entity, Entity<TransformComp
...
+        if (!Resolve(entity1, ref entity1.Comp) || !Resolve(entity2, ref entity2.Comp))
+            return false;
+
+        // save ourselves the hassle and just don't move anything.
+        if (entity1 == entity2)
+            return true;
+
+        var pos1 = GetMapCoordinates(entity1.Comp);
+        var pos2 = GetMapCoordinates(entity2.Comp);
+
```
metalgearsloth: You're getting this much earlier than you need to, you should be doing this if there's no containers involved.

metalgearsloth (changes requested):
What happens if Entity A is parented to Entity B further up the hierarchy? This also really needs a test.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -1511,4 +1511,90 @@ public void PlaceNextTo(Entity<TransformComponent?> entity, Entity<TransformComp
...
+        if (_container.TryGetContainingContainer(entity1, out var container1))
+            _container.TryRemoveFromContainer(entity1, true);
+        else
+            pos1 = GetMapCoordinates(entity1.Comp);
+
+        if (_container.TryGetContainingContainer(entity2, out var container2))
+            _container.TryRemoveFromContainer(entity2, true);
+        else
+            pos2 = GetMapCoordinates(entity2.Comp);
```
metalgearsloth: Use the Remove method that takes in a container otherwise you're resolving it twice up the hierarchy.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -1511,4 +1511,90 @@ public void PlaceNextTo(Entity<TransformComponent?> entity, Entity<TransformComp
...
+        {
+            _container.Insert(entity1, container2);
+        }
+        else if (pos2 != null)
+        {
+            var mapUid = _mapManager.GetMapEntityId(pos2.Value.MapId);
+
+            if (!_gridQuery.HasComponent(entity1) && _mapManager.TryFindGridAt(mapUid, pos2.Value.Position, out var targetGrid, out _))
+            {
+                var (_, _, invWorldMatrix) = GetWorldPositionRotationInvMatrix(targetGrid);
```
metalgearsloth: Just use GetInvWorldMatrix

metalgearsloth (changes requested):
Not sure if you care but the rotations will be based on their old local-rotations.

PR #4990, opened 2024-03-23
Misc Toolshed tweaks

ElectroJr:
- Added an `ICommonSession` parser
- Toolshed commands now only generate completions for the trailing arguments. Previously it generated completions whenever an argument failed to parse, resulting in it generating nonsensical completions.
- Fixed the `CommandDiscriminator` equality override not actually overriding the default record struct equality. This was causing it to constantly add new entries to the `_concreteImplementations` dictionary.
- ~~Removes an obsolete & broken not-implemented exception~~ actually it appears to still be required
- Adds `ILocalizationManager` to parsers & commands.


Robust.Shared/Toolshed/Syntax/Expression.cs
```diff
@@ -46,6 +46,10 @@ public sealed class CommandRun
 
             if (parserContext.EatTerminator())
                 break;
+
+            // Prevent auto completions from dumping a list of all commands at the end of any complete command.
+            if (parserContext.Index == parserContext.MaxIndex + 1)
```
moonheart08:
```suggestion
            if (parserContext.Index > parserContext.MaxIndex)
```

Robust.Shared/Toolshed/ToolshedCommandImplementor.cs
```diff
@@ -124,19 +125,29 @@ public ToolshedCommandImplementor()
             {
                 error?.Contextualize(parserContext.Input, (start, parserContext.Index));
                 args = null;
-                autocomplete = null;
-                if (doAutocomplete)
+
+                // Only generate auto-completions if the parsing error happened for the last argument.
+                if (doAutocomplete && parserContext.Index == parserContext.MaxIndex + 1)
```
moonheart08:
```suggestion
                if (doAutocomplete && parserContext.Index > parserContext.MaxIndex)
```

Robust.Shared/Toolshed/ToolshedCommandImplementor.cs
```diff
@@ -124,19 +125,29 @@ public ToolshedCommandImplementor()
...
                 {
                     parserContext.Restore(chkpoint);
                     autocomplete = _toolshedManager.TryAutocomplete(parserContext, argument.ParameterType, null);
                 }
                 return false;
             }
             args[argument.Name!] = parsed;
+
+            if (!doAutocomplete || parserContext.Index != parserContext.MaxIndex + 1)
```
moonheart08:
```suggestion
            if (!doAutocomplete || parserContext.Index <= parserContext.MaxIndex)
```

Robust.Shared/Toolshed/TypeParsers/SessionTypeParser.cs
```diff
@@ -0,0 +1,63 @@
...
+            return false;
+        }
+
+        if (_player.TryGetSessionByUsername(word, out var session))
+        {
+            result = session;
+            return true;
+        }
+
+        error = new InvalidUsername(Loc.GetString("cmd-parse-failure-session", ("username", word)));
```
moonheart08: Message should be provided by the error's description function, not here.

Robust.Shared/Toolshed/TypeParsers/SessionTypeParser.cs
```diff
@@ -0,0 +1,63 @@
...
+        error.Contextualize(parser.Input, (start, parser.Index));
+        return false;
+    }
+
+    public override async ValueTask<(CompletionResult? result, IConError? error)> TryAutocomplete(ParserContext parserContext,
+        string? argName)
+    {
+        var opts = CompletionHelper.SessionNames(true, _player);
+        return (CompletionResult.FromHintOptions(opts, "<Session>"), null);
```
moonheart08:
```suggestion
        return (CompletionResult.FromHintOptions(opts, "<player session>"), null);
```

moonheart08 (changes requested):
Once these are addressed this should be good to go.

Robust.Shared/Toolshed/ToolshedCommand.cs
```diff
@@ -117,27 +119,26 @@ protected ToolshedCommand()
...
-                        continue;
-
-                    myParams.Add(param.Name!, param.ParameterType);
-                }
-            }
+                    myParams.TryAdd(param.Name!, param.ParameterType);
 
-            if (parameters.TryGetValue(subCmd ?? "", out var existing))
-            {
-                if (!existing.SequenceEqual(existing))
```
ElectroJr, line 133: This was previously bugged because `existing.SequenceEqual(existing)` is always true. After I fixed it, many commands started erroring so I though the feature was implemented, but I misunderstood it. 

If I understand correctly now, it works as long as the (sub)command name & piped type are unique, so I changed the `parameters` dictionary key to a (string, Type?) tuple

moonheart08 (approved):

PR #4995, opened 2024-03-24
Make ValueList<T> implement IList<T>

ElectroJr:
Mainly to avoid unnecessary code duplication (e.g., #4975)


PJB3005 (changes requested):
This allocates and causes non-obvious copying semantics. This is not safe to do because people _will_ fuck it up.

PR #5009, opened 2024-03-27
Fix buffered audio disposals

### LINKED ISSUE #5007: Audio player/manager runs out of memory when running for a few hours or too many audio streams are played
There seems to be a problem with the Audio player/manager that causes it to run out of memory after some hours of playtime or when too many sounds are played at once. For instance, when 20-30 players are stunned by the behonker at once 5 times in a row after playing for like an hour, it will cause partial audio loss as AL runs out of memory. This will also happen after a few hours too 5 or 6 hours depending on how many audio sources are played during gameplay hinting at a slow memory leak as previous audio sources that have been played are not cleaned up correctly. I have run into this on Wizden servers, Delta-V servers and Frontier server. The fastest way I have found to reproduce it during live gameplay was by jumping into the Delta-V server during peak hours because of the number of sounds played at once. I do not have the error logs but someone has already posted error logs regarding the same issue on the ss14 issue tracker linked here https://github.com/space-wizards/space-station-14/issues/25254#issue-2135502227
The issue pops up in the error log on line 91044 and just repeats until the client is closed.
image

metalgearsloth:
weh

Probably fixes https://github.com/space-wizards/RobustToolbox/issues/5007

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -43,7 +43,7 @@ END TEMPLATE-->
 
 ### Bugfixes
 
-*None yet*
+* Fix buffered audio sources not being disposed.
 
 ### Other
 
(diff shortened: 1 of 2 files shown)


deltanedas:
1 line fix to big bug real

PR #5010, opened 2024-03-27
Color.Lerp

TheShuEd:
Linear interpolation function between two colors.

### DIFF
--- a/Robust.Shared.Maths/Color.cs
+++ b/Robust.Shared.Maths/Color.cs
@@ -699,10 +699,10 @@ namespace Robust.Shared.Maths
         {
             t = MathHelper.Clamp01(t);
 
-            float lerpedR = MathHelper.Lerp(startColor.R, endColor.R, t);
-            float lerpedG = MathHelper.Lerp(startColor.G, endColor.G, t);
-            float lerpedB = MathHelper.Lerp(startColor.B, endColor.B, t);
-            float lerpedA = MathHelper.Lerp(startColor.A, endColor.A, t);
+            float lerpedR = MathHelper.Lerp(start.R, end.R, t);
+            float lerpedG = MathHelper.Lerp(start.G, end.G, t);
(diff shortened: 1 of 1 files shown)


chromiumboy:
You have to define 't', looks like you named it 'blend' in the input parameters

TheShuEd:
yeah, typo

chromiumboy:
If this is all you need, I approve of the code

However, if you want to get a bit more fancy and avoid the transition to grey in-between the two colors, I've linked [an article on color interpolation](https://www.alanzucconi.com/2016/01/06/colour-interpolation/)

The TLDR is to convert to colors to Hue, Chroma and Lightness before lerpping (the HCY color format in our code, by the looks)

TheShuEd:
Apparently, the correct implementation would be to create ColorHSV, and functions to convert RGB color to HSV, and vice versa.

TheShuEd:
ok, i did it

TheShuEd:
wait, no

PJB3005:
> However, if you want to get a bit more fancy and avoid the transition to grey in-between the two colors, I've linked an article on color interpolation.

Actually this is why you **shouldn't do color math in sRGB space**. There are many ways to interpolate colors depending on the situation but interpolation in sRGB space (99% chance that's what you're doing) is wrong. The article seems to make this exact same mistake given the graphs use sRGB interpolated colors.

PJB3005 (changes requested):
This function really should not be called "lerp" now that it's an HSV-space interpolation. Regardless this is so specific of an operation I'm not sure it makes sense in the engine.

TheShuEd:
check this out. The name is LerpHSV, and it now works correctly.

TheShuEd:
No, it doesn't work. Okay, if it's specific enough, I won't add it here.

PJB3005:
Also we have Color.InterpolateBetween

PR #5015, opened 2024-03-30
Enable lidgren discovery protocol

VasilisThePikachu:
Could enable some nice stuff like for ss14 web to check if the server udp is open to the internet before advertising or for the launcher to display latency for servers in favourite.

### DIFF
--- a/Robust.Shared/Network/NetManager.cs
+++ b/Robust.Shared/Network/NetManager.cs
@@ -502,7 +502,7 @@ namespace Robust.Shared.Network
                         case NetIncomingMessageType.DiscoveryRequest:
                             // Create a response and write the server name to it
                             NetOutgoingMessage response = peer.Peer.CreateMessage();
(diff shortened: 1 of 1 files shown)


Robust.Shared/Network/NetManager.cs
```diff
@@ -496,6 +499,15 @@ public void ProcessPackets()
                             HandleStatusChanged(peer, msg);
                             break;
 
+                        case NetIncomingMessageType.DiscoveryRequest:
+                            // Create a response and write the server name to it
+                            NetOutgoingMessage response = peer.Peer.CreateMessage();
+                            response.Write(_config.GetCVar(CVars.GameHostName));
```
PJB3005: This should probably be some structured response like JSON instead. If you don't wanna do that right now, make it not send anything at all for now.

PJB3005 (changes requested):

PJB3005:
Stale.

PR #5019, opened 2024-03-31
Make various ValueList enumerators use spans

ElectroJr:
Apparently its ~3-4 times faster than using the ValueList enumerator. Main benefit is making broadcast events ever so slightly faster.


metalgearsloth:
guh I'll fix master, this is what I get for just checking build

PR #5022, opened 2024-04-03
add CopyData to appearance system

deltanedas:
it copies the apperance of src entity to dest entity yep

needed for prop hunt gaming


PJB3005:
This isn't nearly enough, many entities use more complex data than just shader parameters that *somehow* needs to be replicated. God knows how.

deltanedas:
current process is:
1. try copying visualizers
2. blacklist if that fails

going good so far

metalgearsloth:
Considering photos would also use this wouldn't you also need to network the spritecomp data too?

deltanedas:
no i just copy it clientside in content :trollface:

deltanedas:
merg

also photos would need a hell of a lot more than AppearanceComponent

PR #5026, opened 2024-04-05
Rewrite the entire lighting architecture on accident (to add normalmaps)

notquitehadouken:
## Lighting be like.

- [ ] Hope that I suddenly learn how to OpenGL
- [x] Weep
- [ ] Do the bare minimum
- [ ] Hope I don't have to

## Normals be like.

- [ ] Put a place for normal maps in RSIs
- [x] Allow loading from file into RSIs
- [x] Have a default normal map (0, 0, 1)
- [ ] Try not to rework the entire lighting engine
- [ ] Do that thing where I painstakingly change every single thing that requires a texture
- [ ] Come up with a better idea than a tuple
- [ ] Make everything use a tuple for textures
- [ ] Make normals actually do stuff with lighting
- [ ] Make sure it actually runs good (probably gonna be fine)

### DIFF
--- a/Robust.Client/Graphics/AtlasTexture.cs
+++ b/Robust.Client/Graphics/AtlasTexture.cs
@@ -1,7 +1,9 @@
+using System.Numerics;
 using JetBrains.Annotations;
 using Robust.Shared.Graphics;
 using Robust.Shared.Maths;
 using Robust.Shared.Utility;
+using Vector3 = Robust.Shared.Maths.Vector3;
 
 namespace Robust.Client.Graphics
 {
@@ -33,7 +35,7 @@ namespace Robust.Client.Graphics
         /// </summary>
         public UIBox2 SubRegion { get; }
 
-        public override Color GetPixel(int x, int y)
+        public override (Color, Vector3) GetPixel(int x, int y)
         {
             DebugTools.Assert(x < SubRegion.Right);
             DebugTools.Assert(y < SubRegion.Top);
--- a/Robust.Client/Graphics/Clyde/Clyde.Textures.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.Textures.cs
@@ -4,9 +4,11 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Data;
 using System.IO;
+using System.Numerics;
 using System.Runtime.InteropServices;
 using System.Threading;
 using OpenToolkit.Graphics.OpenGL4;
+using OpenToolkit.Mathematics;
 using Robust.Client.Utility;
 using Robust.Shared.Graphics;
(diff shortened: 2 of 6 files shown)


MACMAN2003:
does this mean it'd be possible to make an unusually spherical circle....

PJB3005:
> Make normals actually do stuff with lighting

Good luck rewriting the entire rendering architecture.

PJB3005:
(I am sarcastic, sorry)

PJB3005:
More productively:

Implementing this will require a complete rewrite of the rendering engine to make it render normals into a separate g-buffer, much like 3D engines. I do not think it is even productive to attempt this on the current shitpile of OpenGL code that is our renderer.

moonheart08:
> More productively:
> 
> Implementing this will require a complete rewrite of the rendering engine to make it render normals into a separate g-buffer, much like 3D engines. I do not think it is even productive to attempt this on the current shitpile of OpenGL code that is our renderer.

Adding atop this, our current rendering approach (Deferred Lighting) is incompatible with normal maps, the lighting engine cannot be used as is. In order to support normals we'd need either a Forward+ (probably the ideal approach) or *actual* Deferred renderer. Our current approach is good due to its simplicity for the number of lights it can handle but that simplicity comes at the cost of advanced lighting effects.

notquitehadouken:
> > More productively:
> > Implementing this will require a complete rewrite of the rendering engine to make it render normals into a separate g-buffer, much like 3D engines. I do not think it is even productive to attempt this on the current shitpile of OpenGL code that is our renderer.
> 
> Adding atop this, our current rendering approach (Deferred Lighting) is incompatible with normal maps, the lighting engine cannot be used as is. In order to support normals we'd need either a Forward+ (probably the ideal approach) or _actual_ Deferred renderer. Our current approach is good due to its simplicity for the number of lights it can handle but that simplicity comes at the cost of advanced lighting effects.

i'll give up once i run out of bullshit to pull

deathride58:
It might genuinely be feasible to fake normal maps by having them offset where exactly in UV space the default shader pulls from the lighting framebuffer, though this could end up being quite janky. But doing it proper is out of the question without some massive changes to rendering as a whole.

notquitehadouken:
image
highlight

MACMAN2003:
if you need any assistance with learning opengl (the hard part) then i highly recommend [https://learnopengl.com](url)

notquitehadouken:
after running in circles repeatedly with putting ( , ) on everything i have given up and will try again once i come up with a better idea

notquitehadouken:
image
i made pi two

MACMAN2003:
pi two?? that's just tau :trollface:

deltanedas:
Activate Windows
Go to Settings to activate Windows.

PR #5027, opened 2024-04-06
Compress large networked entity events

ElectroJr:
This PR makes networked `EntityEventArgs` get automatically compressed if they are over 1024 bytes (game states currently get compressed above 256 bytes). This is probably useful for systems that consistently send easily compressible data (e.g., gas tile overlay), or just send large messages (e.g., some admin/debug overlays). 

I'm somewhat unsure about what to do in the case of multi-threaded message sending. `RaiseNetworkEvent` now takes in an optional `ZStdCompressionContext`, so callers could use something like the `PvsThreadResources` pool, but it defaults to using a `ThreadLocal<ZStdCompressionContext>`.

This PR also removes the `EntityMessageType` enum, seeing as it was unused, and I'm not sure what it was event meant to be used for.

### DIFF
--- a/Robust.Shared/Network/CompressedNetMessage.cs
+++ b/Robust.Shared/Network/CompressedNetMessage.cs
@@ -1,0 +1,84 @@
+using System;
+using System.Buffers;
+using Lidgren.Network;
+using Robust.Shared.GameObjects;
+using Robust.Shared.Serialization;
+using Robust.Shared.Utility;
+
+namespace Robust.Shared.Network;
+
+public abstract class CompressedNetMessage : NetMessage
+{
+    protected int ReadCompressed<T>(NetIncomingMessage buffer, IRobustSerializer serializer, out T obj, bool direct)
+    {
+        var uncompressedLength = buffer.ReadVariableInt32();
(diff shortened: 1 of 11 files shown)


PJB3005:
holy fucking shit sensor monitoring console let's fricking go.

Robust.Shared/Network/CompressedNetMessage.cs
```diff
@@ -0,0 +1,84 @@
+using System;
+using System.Buffers;
+using Lidgren.Network;
+using Robust.Shared.GameObjects;
+using Robust.Shared.Serialization;
+using Robust.Shared.Utility;
+
+namespace Robust.Shared.Network;
+
+public abstract class CompressedNetMessage : NetMessage
```
PJB3005, line 10: This shouldn't be an abstract class at all, it only has some helper functions.

PJB3005 (changes requested):
> `RaiseNetworkEvent` now takes in an optional `ZStdCompressionContext`

Leaks way too many implementation details and is a weird API. The handle should be some opaque object that can be requested in advance instead.

PJB3005:
Stale

PR #5032, opened 2024-04-08
Add Type tracking to FieldNotFoundErrorNode

Tayrtahn:
FieldNotFoundErrorNode now stores the Type of the datadef that the error relates to (because the type does not have the field in question). This information was already being used as part of the error description; this change just makes the node keep a record of it for future use.

This is intended for use in a Content PR to improve the YAML linter.

### DIFF
--- a/Robust.Shared/Serialization/Markdown/Validation/FieldNotFoundErrorNode.cs
+++ b/Robust.Shared/Serialization/Markdown/Validation/FieldNotFoundErrorNode.cs
@@ -5,7 +5,9 @@ namespace Robust.Shared.Serialization.Markdown.Validation;
 
 public sealed class FieldNotFoundErrorNode : ErrorNode
 {
+    public Type Type;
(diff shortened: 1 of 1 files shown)


Robust.Shared/Serialization/Markdown/Validation/FieldNotFoundErrorNode.cs
```diff
@@ -5,7 +5,9 @@ namespace Robust.Shared.Serialization.Markdown.Validation;
 
 public sealed class FieldNotFoundErrorNode : ErrorNode
 {
+    public Type Type;
     public FieldNotFoundErrorNode(ValueDataNode key, Type type) : base(key, $"Field \"{key.Value}\" not found in \"{type}\".", false)
     {
+        Type = type;
```
PJB3005:
```suggestion
    public Type FieldType { get; }
    public FieldNotFoundErrorNode(ValueDataNode key, Type type) : base(key, $"Field \"{key.Value}\" not found in \"{type}\".", false)
    {
        FieldType = type;
```

1. make it get-only
2. `FieldType` is a more clear name.

PR #5041, opened 2024-04-15
Add FullscreenUI

notquitehadouken:
adds a simple thing for making XAML containers that are fullscreen (as per suggestion of Jezithyr)


Robust.Client/UserInterface/CustomControls/DefaultFullscreen.xaml.cs
```diff
@@ -0,0 +1,119 @@
...
+            MouseFilter = MouseFilterMode.Stop;
+
+            Contents = ContentsContainer;
+
+            XamlChildren = new SS14ContentCollection(this);
+        }
+
+        public Control Contents { get; private set; }
+
+        public sealed class SS14ContentCollection : ICollection<Control>, IReadOnlyCollection<Control>
```
moonheart08: Please don't copypaste these things. Generalize the original type by making it internal and generic, then create an empty inheritor, maybe?

moonheart08 (changes requested):

notquitehadouken:
if you ignore the part where it doesn't work, it works.

PJB3005:
Stale.

PR #5042, opened 2024-04-16
Partial MapManager refactor

ElectroJr:
This PR partially refactors map manager and moves a lot of the logic into the map system. Half of the changes are just updates to unit tests to avoid obsolete warnings and to get rid of explicit map Ids.

IMO all the map code should eventually be moved to the system as I don't know of any reason not to, its nice being able to use all the entity system convenience methods, and having the functionality split between a manager & system is just kind of awkward. 

Requires space-wizards/space-station-14/pull/26994

Changes:
- Moves map creation, pausing and initialization logic from `IMapManager` to `MapSystem`
  - Various IMapManager methods are now marked as obsolete, and just redirect to the system.
  - The new map creation methods can return both the MapId and EntityUid.
- Removes
  - `IMapManagerInternal.CreateMap()`
  - `IMapManager.RemoveMapId()`
  - `IMapManager.NextMapId()`
  - `IMapManager.GetGridComp()`
  - `IMapManager.GetAllGrids()`
- Adds a new `IEntityManager.InitializeAndStartEntity` override that allows you to specify whether the entity should get map-initialized
  - Previously this was always automatically inferred from the current MapId.
  - Used to fix a bug introduced in #5030, where spawning an entity into a container on a pre-init map would map-initialize the spawned entity.
- Adds "client-side" mapIds.
  - These are just MapIds with negative values.
  - Required to ensure that a client-side map entity doesn't get assigned a MapId that ends up conflicting with a server-side map.
  - I don't know if a game will ever actually need this functionality, but it was pretty trivial to add.


metalgearsloth:
Lord forgive me for conflicts but I need this so fucking bad.

PR #5043, opened 2024-04-16
Expose worldPosition in SpriteComponent.Render

DrSmugleaf:
Can't specify the position when rendering an entity through a world drawing handle otherwise.

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -8,7 +8,6 @@ using System.Text;
 using Robust.Client.Graphics;
 using Robust.Client.ResourceManagement;
 using Robust.Client.Utility;
-using Robust.Shared;
 using Robust.Shared.Animations;
 using Robust.Shared.ComponentTrees;
 using Robust.Shared.GameObjects;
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1237,9 +1236,9 @@ public RSI.StateId LayerGetState(int layer)
         public IEnumerable<ISpriteLayer> AllLayers => Layers;
 
         // Lobby SpriteView rendering path
-        public void Render(DrawingHandleWorld drawingHandle, Angle eyeRotation, Angle worldRotation, Direction? overrideDirection = null)
+        public void Render(DrawingHandleWorld drawingHandle, Angle eyeRotation, Angle worldRotation, Direction? overrideDirection = null, Vector2? worldPosition = null)
         {
-            RenderInternal(drawingHandle, eyeRotation, worldRotation, Vector2.Zero, overrideDirection);
+            RenderInternal(drawingHandle, eyeRotation, worldRotation, worldPosition ?? Vector2.Zero, overrideDirection);
```
ShadowCommander: Maybe put the Vector2.Zero as default parameter instead of null?

ShadowCommander (changes requested):

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
DrSmugleaf: Done in [`0dcf57f` (#5043)](https://github.com/space-wizards/RobustToolbox/pull/5043/commits/0dcf57f03a04207560861b99c1b24e1d9d2afbaa)

PR #5047, opened 2024-04-18
Adds rotation to Map Position spawns

keronshb:
This is mostly for the Polymorph system so you can pass the parent polymorph's rotation to the child. Can be used in other applications at least.

Tested with polymorphing into a rod and using the rods linear velocity. I also tested the Tesla and lightning (beam system) behaves normally.

https://github.com/space-wizards/RobustToolbox/assets/54602815/39a7d227-7924-43f4-b5a5-e8e787966c5a

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.Spawn.cs
+++ b/Robust.Shared/GameObjects/EntityManager.Spawn.cs
@@ -87,7 +87,7 @@ public partial class EntityManager
     public EntityUid Spawn(string? protoName = null, ComponentRegistry? overrides = null)
         => Spawn(protoName, MapCoordinates.Nullspace, overrides);
 
-    public virtual EntityUid Spawn(string? protoName, MapCoordinates coordinates, ComponentRegistry? overrides = null, Angle rotation = new())
+    public virtual EntityUid Spawn(string? protoName, MapCoordinates coordinates, ComponentRegistry? overrides = null, Angle rotation = default!)
     {
         var entity = CreateEntityUninitialized(protoName, coordinates, overrides, rotation);
         InitializeAndStartEntity(entity, coordinates.MapId);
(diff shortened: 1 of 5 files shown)


Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -324,6 +325,7 @@ public virtual EntityUid CreateEntityUninitialized(string? prototypeName, MapCoo
             {
                 coords = new EntityCoordinates(mapEnt, coordinates.Position);
                 _xforms.SetCoordinates(newEntity, transform, coords, null, newParent: mapXform);
+                _xforms.SetWorldRotation(newEntity, rotation);
```
metalgearsloth: SetCoordinates has an overload that takes in rotation that means only 1 moveevent goes out instead of 2 (movement updates are the most expensive thing in the game).

Robust.Shared/GameObjects/EntityManager.Spawn.cs
```diff
@@ -86,9 +87,9 @@ public virtual EntityUid SpawnAttachedTo(string? protoName, EntityCoordinates co
     public EntityUid Spawn(string? protoName = null, ComponentRegistry? overrides = null)
         => Spawn(protoName, MapCoordinates.Nullspace, overrides);
 
-    public virtual EntityUid Spawn(string? protoName, MapCoordinates coordinates, ComponentRegistry? overrides = null)
+    public virtual EntityUid Spawn(string? protoName, MapCoordinates coordinates, ComponentRegistry? overrides = null, Angle rotation = new())
     {
```
metalgearsloth: does rotation = default / rotation = Angle.Zero work instead.

keronshb: Angle.Zero doesn't, says it needs to be compile time.
default worked

Robust.Shared/GameObjects/EntityManager.cs
keronshb: OOOPS i missed that null there

PR #5050, opened 2024-04-18
Auth is now required by default

VasilisThePikachu:
Most people are going to be using authentication anyway.

I tested and launching a local environment will still let you join as normal. This will just make sure servers who forget to change the ``auth.mode`` to required don't get a nasty surprise later on


PJB3005:
I was going to say "no it's set to required on release builds" but no turns out that's wrong so yeah this is the intended behavior.

PR #5052, opened 2024-04-19
Fix RecursiveMapInit

ElectroJr:
Also changes the mapmanager sawmill to use the same sawmill as the system, so you only need to configure one log level.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -43,7 +43,7 @@ END TEMPLATE-->
 
 ### Bugfixes
 
-*None yet*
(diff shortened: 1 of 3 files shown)


deltanedas:
is this an AddRange funny that makes it so the list cant be reused across calls or what

ElectroJr:
> is this an AddRange funny that makes it so the list cant be reused across calls or what

It was just an oversight on my end. The issue was that map initialization can be recursive / is not done one at a time. Some entities spawn & initialize new maps when they get map-initialized, which was clearing the `_toInitialize` queue while it was still initializing another map.

PR #5055, opened 2024-04-20
Interface shuffling for shared

metalgearsloth:
Bottom text

### DIFF
--- a/Robust.Client/Player/PlayerManager.cs
+++ b/Robust.Client/Player/PlayerManager.cs
@@ -65,6 +65,16 @@ namespace Robust.Client.Player
                 LocalStatusChanged?.Invoke(e);
         }
 
+        public ICommonSession CreateAndAddSession(NetUserId user, string name)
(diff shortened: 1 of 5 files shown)


Robust.Client/Player/PlayerManager.cs
```diff
@@ -65,6 +65,16 @@ private void StatusChanged(object? sender, SessionStatusEventArgs e)
...
+        public ICommonSession CreateAndAddSession(NetUserId user, string name)
+        {
+            return CreateAndAddSessionInternal(user, name);
+        }
+
+        public ICommonSession CreateAndAddSession(INetChannel channel)
+        {
+            return CreateAndAddSessionInternal(channel);
+        }
+
```
metalgearsloth, line 77: Still needed for replaysTM.

PJB3005 (changes requested):
Move them to an internal interface nerd.

PR #5056, opened 2024-04-20
Fix yaml linter & improve validation of static fields

ElectroJr:
This PR fixes an issue where the prototype validation methods used by the yaml linter would fail to properly validate prototypes with multiple parents.

It also changes how the old `ValidateFields()` method works. Previously it would look through all types known to `IReflectionManager` and look for fields with the dataField or `ValidatePrototypeId<T>` attribute to check if the default value of those fields are valid. This was both to check that the default values of data definitions were valid, and to validate hardcoded prototypes that get used by various systems.

This PR replaces this method with `ValidateStaticFields()`, which now only validates static fields, but also automatically validates `EntProtoId` and `ProtoId<T>` fields, not just strings. It also adds support for `IEnumerable<>` types. Datafields now instead just get validated directly when the yaml linter validates each prototype, though this does mean that the default values of non-prototype related datafields are no longer automatically validated.

I.e., previously the linter would only check strings like
```cs
[ValidatePrototypeId<TagPrototype>] static string Tag = "Wall";
```
However, now it will also check:
```cs
static ProtoId<TagPrototype> Tag = "AirAlarm";
static ProtoId<TagPrototype>[] Tags = { "AirAlarm", "Airlock" };
static HashSet<EntProtoId> Ents = new() { "AirAlarm", "Airlock" };
```

Requires space-wizards/space-station-14/pull/27188

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.YamlValidate.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.YamlValidate.cs
@@ -53,7 +53,7 @@ public partial class PrototypeManager
                     var mapping = node.ToDataNodeCast<MappingDataNode>();
                     var id = mapping.Get<ValueDataNode>("id").Value;
 
-                    var data = new PrototypeValidationData(mapping, resourcePath.ToString());
+                    var data = new PrototypeValidationData(id, mapping, resourcePath.ToString());
                     mapping.Remove("type");
 
                     if (prototypes.GetOrNew(type).TryAdd(id, data))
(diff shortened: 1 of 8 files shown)


Robust.Shared/Prototypes/PrototypeManager.YamlValidate.cs
```diff
@@ -176,23 +201,22 @@ public PrototypeValidationData(MappingDataNode mapping, string file)
...
+        var i = 0;
+        foreach (var parentId in data.Parents)
         {
             var parent = prototypes[parentId];
             EnsurePushed(parent, prototypes, type);
-
-            for (var i = 0; i < parents.Length; i++)
-            {
-                parentNodes[i] = parent.Mapping;
-            }
```
ElectroJr, line 190: This is where the multiple parent bug happened -  it would set all parent mappings equal to the same parent

PR #5059, opened 2024-04-21
Predicted BUIs

metalgearsloth:
Cleaned it up significantly and made it use networking properly and moved everything to shared.

Content PR is https://github.com/space-wizards/space-station-14/pull/27214

Alright notes:
- PlayerBoundUserInterface got turned into InterfaceData which is just datafield stuff, server doesn't need to deal with BUIs anymore. It's only in shared so we can put all the client code in shared.
- BUI states now get networked via ComponentStates. This also means player-specific BUI state overrides are gone.
- Callers shouldn't need to do any of the hacky stuff to get a BUI just to update a state or the likes.
- Methods like IsUiOpen now work how you'd expect them to from a shared context and be reliable.
- BUIs now properly handle themselves upon changing entity (e.g. if you aghost and return it will re-open your BUIs), lobby, disconnection and reconnection, and more.
- BUI opening should at minimum have parity with the old system (where the opening is not predicted) and best case is predicted.
- BUIs closing is now shared.
- BUIs had ICommonSession almost entirely stripped out so we can add NPC support in the future (this is especially critical for storage so we don't need to hack in NPC-specific paths).
- There's a lot of API breakage but it will be worth itTM as (IMO) it is a lot saner.
- There may be some level of BUI state data leakage to bad clients though not sure the best way around this.

The main pain point for downstreams is updating yaml so Uis are keyed by the actual UiKey and not a list that gets turned into a dictionary, same with IntrinsicUI.


Callmore:
wait a sec
predicted nt block game? 👀

mirrorcult:
blessed

metalgearsloth:
> wait a sec predicted nt block game? 👀

No because arcade code is terrible and would need to be entirely redone.

PR #5061, opened 2024-04-22
Rework entity prototype categories

ElectroJr:
- Renames engine categories to use pascal case.
  - Requires space-wizards/space-station-14/pull/27232.
- Added `IPrototypeManager.EnumerateAllParents()` for enumerating over all parents, including any that are abstract.
- Added `IPrototypeManager.Categories` for retrieving lists of prototypes that belong to a given category.
- Allows categories to provide a default value for a prototype's EditorSuffix. E.g., all prototypes in the debug category should get "[Debug]" appended.
- Each category now has a `HideSpawnMenu` field. The old `HideSpawnMenu` category is now just a category with that field set to true.
- Reworked category inheritance. Inheritance can now be disabled per category using a `Inheritable` field.
  - This required some jank for entities to properly inherit categories from abstract parents.
- Entity prototypes can now be automatically added to categories based on the components that they have. 
  - This should help avoid issues where people might forget to specify categories. E.g., it makes it trivial to ensure that all action-entities are in the actions category, and are hidden from the spawn menu.
  - There are two ways to associate a component with a category:
    - By adding the component to the category's `Components` datafield.
    - By giving the component an `EntityCategoryAttribute`.
  - Having two ways of doing this might lead to inconsistency, if it should be changed to use only one of these, then it should be done via the attribute, so that content can use engine categories.


Robust.Shared/Prototypes/EntProtoId.cs
```diff
@@ -20,6 +20,11 @@ namespace Robust.Shared.Prototypes;
         return protoId.Id;
     }
 
+    public static implicit operator EntProtoId(EntityPrototype proto)
+    {
+        return new EntProtoId(proto.ID);
+    }
+
```
Tayrtahn, line 30: Test failure seems to be caused by this implicit operator interacting with this code in Content: 

https://github.com/space-wizards/space-station-14/blob/master/Content.Shared%2FBody%2FPrototypes%2FBodyPrototypeSerializer.cs#L185

where the BodyPrototypeSlot constructor is forcing null to be cast to EntProtoId.

deltanedas: seems adding one for EntityPrototype? to EntProtoId? (if possible) and using that would fix it

ElectroJr: > seems adding one for EntityPrototype? to EntProtoId? (if possible) and using that would fix it

You can't sadly.

Also this has breaking changes anyways because of the CamelCase prototype name changes, so its probably fine to also have this.

deltanedas:
the parent enumerating doesnt check if its abstract i think? its failing test with EnergyCutlass -> invalid parent BaseItem

ElectroJr:
> the parent enumerating doesnt check if its abstract i think? its failing test with EnergyCutlass -> invalid parent BaseItem

It's intentionally including abstract prototypes, so that you can add a category to an abstract proto and then have all inheritors also be part of that category. 

AFAICT the error was caused some parallel code problems that should've been fixed by #5200

PR #5062, opened 2024-04-22
Remove IP address and HWId from ViewVariables

ShadowCommander:
Prevents viewing IP Address and HWId from ViewVariables.

Prep for space-wizards/space-station-14#6859 and space-wizards/space-station-14#23720

Content Client_81zRdmc2h7


Geekyhobo:
This is good and covers all the bases that I found. Going to reference this for my PII flag PR as well

PR #5065, opened 2024-04-24
Add SetMapCoordinates

EmoGarbage404:
wrapper function for setcoordinates that works with MapCoordinates

needed for content.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
@@ -877,7 +877,6 @@ public abstract partial class SharedTransformSystem
         SetMapCoordinates((entity, xform), coordinates);
     }
 
-    [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -870,6 +870,29 @@ public MapCoordinates GetMapCoordinates(Entity<TransformComponent> entity)
...
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public void SetMapCoordinates(EntityUid entity, MapCoordinates coordinates)
+    {
+        var xform = XformQuery.GetComponent(entity);
+        SetMapCoordinates((entity, xform), coordinates);
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public void SetMapCoordinates(Entity<TransformComponent> entity, MapCoordinates coordinates)
```
metalgearsloth:
```suggestion
    public void SetMapCoordinates(Entity<TransformComponent> entity, MapCoordinates coordinates)
```

PR #5067, opened 2024-04-25
Add margin input value order as a comment

ShadowCommander:
Makes it easier when implementing margin, since it shows up as a popup when hovered in the IDE.


PJB3005 (changes requested):
This just isn't a valid summary comment, sorry. Put it in the remarks instead.

PR #5070, opened 2024-04-26
Add method for getting type of var from serializer

amylizzle:
OpenDream is abusing the serializer to do DMF parsing, and as a result of https://github.com/OpenDreamProject/OpenDream/pull/1757 I need to know the subtype of the vars at runtime.

This PR adds a function on the `SerializationManager`  which gets the `Type` of a variable on another `Type` by it's `DataField` tag.


DrSmugleaf (approved):
@PJB3005

PR #5077, opened 2024-04-28
Add refresh BUI helpers

metalgearsloth:
Still need to manually call it in AfterAutoHandleState though not sure the easiest way to avoid boilerplate there without exchanging it with different boilerplate.


metalgearsloth:
This got rolled into some PR or other.

PR #5092, opened 2024-04-30
add SetUi to shared ui system

deltanedas:
evil refactor removed it

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
@@ -911,6 +911,19 @@ public abstract class SharedUserInterfaceSystem : EntitySystem
     }
 
     /// <summary>
+    /// Add a UI after an entity has been created.
+    /// It cannot be added already.
+    /// </summary>
+    public void AddUi(Entity<UserInterfaceComponent?> ent, Enum key, InterfaceData data)
+    {
+        if (!Resolve(ent, ref ent.Comp))
+            return;
+
+        ent.Comp.Interfaces[key] = data;
+        Dirty(ent, ent.Comp);
+    }
+
+    /// <summary>
     ///     Verify that the subscribed clients are still in range of the interface.
     /// </summary>
     private bool CheckRange(
--- a/Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
@@ -911,13 +911,12 @@ public abstract class SharedUserInterfaceSystem : EntitySystem
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


deltanedas:
@metalgearsloth

Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
```diff
@@ -910,6 +910,19 @@ public override void Update(float frameTime)
         }
     }
 
+    /// <summary>
+    /// Add a UI after an entity has been created.
+    /// It cannot be added already.
```
metalgearsloth: This comment is false because the code doesn't care if it's added already. SetUi is probably a better name for it too as Add implies it will throw if it exists already.

Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
```diff
@@ -910,6 +910,19 @@ public override void Update(float frameTime)
...
+    /// </summary>
+    public void AddUi(Entity<UserInterfaceComponent?> ent, Enum key, InterfaceData data)
+    {
+        if (!Resolve(ent, ref ent.Comp))
+            return;
```
metalgearsloth: Should this just ensure instead.

metalgearsloth (changes requested):

Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
deltanedas: yeah userinterface doesnt actually need anything funny anymore

deltanedas:
@metalgearsloth a

Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
```diff
@@ -910,6 +910,18 @@ public override void Update(float frameTime)
         }
     }
 
+    /// <summary>
+    /// Set a UI after an entity has been created.
+    /// </summary>
+    public void SetUi(Entity<UserInterfaceComponent?> ent, Enum key, InterfaceData data)
+    {
+        if (!Resolve(ent, ref ent.Comp, false))
+            ent.Comp = AddComp<UserInterfaceComponent>(ent);
```
AJCM-git, line 919: why not use the normal EnsureComp?

AJCM-git: if its cuz of the cursed Entity<> edge case, just do `var comp = ensure(ent.Owner)`  and access it that way

deltanedas: ensurecomp wont check if uid == comp.Owner i think, doing resolve does

PR #5102, opened 2024-05-04
Hidden tiles

TheShuEd:
A small addition that allows you to hide some tiles from the spawn menu. This is useful for downstrimes, disabling the ability to see standard tiles, but not causing huge problems with cutting them from everywhere they are used.

### DIFF
--- a/Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
+++ b/Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
@@ -146,6 +146,8 @@ public sealed class TileSpawningUIController : UIController
 
         foreach (var entry in _shownTiles)
         {
+            if (entry.Hidden) continue;
+            
             Texture? texture = null;
             var path = entry.Sprite?.ToString();
 
--- a/Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
+++ b/Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
@@ -135,8 +135,8 @@ public sealed class TileSpawningUIController : UIController
         if (!string.IsNullOrEmpty(searchStr))
         {
             tileDefs = tileDefs.Where(s =>
-                Loc.GetString(s.Name).Contains(searchStr, StringComparison.CurrentCultureIgnoreCase) ||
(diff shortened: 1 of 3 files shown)


TheShuEd:
I wonder how I should initialize this field in an object that is not in the engine?

TheShuEd:
requires https://github.com/space-wizards/space-station-14/pull/27666

PJB3005 (changes requested):
I'd rename the property to `EditorHidden` or something, so people don't mistake it for a rendering property or something.

TheShuEd:
Done

Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
```diff
@@ -146,6 +146,8 @@ private void BuildTileList(string? searchStr = null)
 
         foreach (var entry in _shownTiles)
         {
+            if (entry.EditorHidden) continue;
```
PJB3005: Didn't realize this on the first pass: it's probably better to do this filter *before* the name filter and sort, i.e. put it on line 133 with a `.Where()`.

TheShuEd: Ok, but tomorrow. Im sleeping now

Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
```diff
@@ -135,8 +135,8 @@ private void BuildTileList(string? searchStr = null)
         if (!string.IsNullOrEmpty(searchStr))
         {
             tileDefs = tileDefs.Where(s =>
-                Loc.GetString(s.Name).Contains(searchStr, StringComparison.CurrentCultureIgnoreCase) ||
-                s.ID.Contains(searchStr, StringComparison.OrdinalIgnoreCase));
+                (Loc.GetString(s.Name).Contains(searchStr, StringComparison.CurrentCultureIgnoreCase) ||
+                s.ID.Contains(searchStr, StringComparison.OrdinalIgnoreCase)) && !s.EditorHidden);
```
PJB3005: Doing it like this only applies the hidden when you have a search string.

PR #5103, opened 2024-05-05
Increase Uploads size

Pgriha:
Increasing the maximum size of the file uploaded by the client to the server from 3 to 15 MB.

### DIFF
--- a/Robust.Shared/CVars.cs
+++ b/Robust.Shared/CVars.cs
@@ -1705,7 +1705,7 @@ namespace Robust.Shared
         /// Set to zero or a negative value to disable limit.
         /// </summary>
         public static readonly CVarDef<float> ResourceUploadingLimitMb =
-            CVarDef.Create("netres.limit", 3f, CVar.REPLICATED | CVar.SERVER);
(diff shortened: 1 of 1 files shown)


metalgearsloth:
Can you please fill out the why section of the pr template.

Pgriha:
> Can you please fill out the why section of the pr template.

Why-why..
Convenience in downloading the same music to the server when there is no direct access to it. But it can weigh simply more than 3 megabytes. Either large maps or folders with several medium maps or audio.
(Of course, it would be possible to completely remove the limit...
But accidentally mixing up a file and starting to load some gigabyte archive, crashing the server, sounds bad.)

PJB3005:
Resource downloading is already incredibly problematic because the networking layer cannot handle high-bandwidth transfer, and this problem gets worse with increased RTT. If you actually uploaded 15 or even 10 megabyte of files, many people would need to wait *minutes* just to be able to connect. There may be other performance issues such as server memory usage if a lot of people connect at once.

Given that this is *already* a config variable that can be changed by servers, I do not think it is a good idea to change the default in the engine.

Thank you for your contribution, however I will be closing this PR.

PR #5107, opened 2024-05-05
Analyzer to ban uncached regexes

PJB3005:
Using static Regex functions that take in a pattern is bad, because they constantly have to be re-parsed. Cache the Regex instance.


metalgearsloth:
Looks like code to me.

PR #5116, opened 2024-05-08
Add physics delta states

metalgearsloth:
Significantly cuts down on data being sent + should make client state handling faster.

### DIFF
--- a/Robust.Shared/Physics/Components/PhysicsComponentState.cs
+++ b/Robust.Shared/Physics/Components/PhysicsComponentState.cs
@@ -39,7 +39,7 @@ public sealed class PhysicsDeltaState : IComponentState, IComponentDeltaState
     public Vector2 LinearVelocity;
     public float AngularVelocity;
 
-    public bool FullState => true;
+    public bool FullState => false;
 
     public void ApplyToFullState(IComponentState fullState)
     {
--- a/Robust.Shared/Physics/Components/PhysicsComponentState.cs
+++ b/Robust.Shared/Physics/Components/PhysicsComponentState.cs
@@ -1,13 +1,12 @@
 ﻿using System;
 using System.Numerics;
 using Robust.Shared.GameObjects;
-using Robust.Shared.Maths;
 using Robust.Shared.Serialization;
 
 namespace Robust.Shared.Physics.Components;
 
 [Serializable, NetSerializable]
-public readonly record struct PhysicsComponentState(
+public sealed class PhysicsComponentState(
         bool CanCollide,
         bool SleepingAllowed,
         bool FixedRotation,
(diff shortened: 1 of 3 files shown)


Robust.Shared/Physics/Components/PhysicsComponentState.cs
```diff
@@ -25,11 +24,35 @@ namespace Robust.Shared.Physics.Components;
...
 }
+
+[Serializable, NetSerializable]
+public sealed class PhysicsDeltaState : IComponentState, IComponentDeltaState
+{
+    public Vector2 LinearVelocity;
+    public float AngularVelocity;
+
+    public bool FullState => true;
+
```
metalgearsloth:
```suggestion

    public bool FullState => false;

```

Oop.

Robust.Shared/Physics/Components/PhysicsComponentState.cs
```diff
@@ -25,11 +24,35 @@ namespace Robust.Shared.Physics.Components;
...
 }
+
+[Serializable, NetSerializable]
+public sealed class PhysicsDeltaState : IComponentState, IComponentDeltaState
+{
+    public Vector2 LinearVelocity;
+    public float AngularVelocity;
+
+    public bool FullState => false;
+
```
metalgearsloth, line 43: Okay you know what I have NFI how deltas are supposed to work, I guess I mark this as FullState = true and shit's gravy?

```suggestion

    public bool FullState => true;

```

ShadowCommander (approved):

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
```diff
@@ -442,7 +471,7 @@ public void SetAwake(Entity<PhysicsComponent> ent, bool value, bool updateSleepT
 
         // Update wake system last, if sleeping but still colliding.
         if (!value && body.CanCollide)
-            _wakeSystem.UpdateCanCollide(ent, checkTerminating: false, dirty: false);
+            _wakeSystem.UpdateCanCollide(ent, checkTerminating: false);
 
         if (updateSleepTime)
             SetSleepTime(body, 0);
```
ElectroJr, line 477: Why change this and removes the dirty call later on? AFAIK you also want to dirty the ent even if body.CanCollide is false

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
```diff
@@ -69,21 +69,38 @@ private void OnPhysicsInit(EntityUid uid, PhysicsComponent component, ComponentI
...
-            component.FixedRotation,
-            component.BodyStatus,
-            component.LinearVelocity,
-            component.AngularVelocity,
-            component.BodyType,
-            component._friction,
-            component.LinearDamping,
-            component.AngularDamping);
+        if (args.FromTick <= component.FullUpdate)
+        {
```
ElectroJr, line 73: IIRC it always also needs a `args.FromTick <= component.CreationTick` check, in case the entity was newly spawned? Not 100% sure though. The other delta state stuff might've just had it in case the comp got added to an entity.

PR #5120, opened 2024-05-10
Fix checking wrong property in `TryGetVariableType()`

amylizzle:
I guess I only tested the case where the tag and variable name happened to match. Whoops.

`BackingField.Name` is the var name ie `BackgroundColor`
`Attribute.Tag` is the `DataField` tag name ie `background-color`


PJB3005:
Does this work if the data field name is inferred automatically? i.e.
```cs
[DataField]
public int Foobar { get; set; }
```

amylizzle:
It does! For some reason it automatically applied camelCase in my test, but the inferred name worked as a query.

DrSmugleaf (approved):

PR #5126, opened 2024-05-12
remove XAMLIL from assembly after use

aspiringLich:
Simple. These embedded resources (probably) aren't used after we generate the populate method and inject it into the `BaseWindow` constructors. Therefore, we can just get rid of them. 

The savings are incredible.

**`Robust.Client.dll`**: 1,833 KiB (1,876,992 B) → 1,816 KiB (1,859,584 B) 
**`Content.Client.dll`**: 3,116 KiB (3,190,784 B) → 3,105 KiB (3,180,032 B)

I'm like, pretty sure this doesn't break anything. UIs still open and everything, so, it's probably fine.

### DIFF
--- a/Robust.Client.Injectors/XamlCompiler.cs
+++ b/Robust.Client.Injectors/XamlCompiler.cs
@@ -239,6 +239,7 @@ namespace Robust.Build.Tasks
                         engine.LogErrorEvent(new BuildErrorEventArgs("XAMLIL", "", res.FilePath, 0, 0, 0, 0,
                             $"{res.FilePath}: {e.Message}", "", "CompileRobustXaml"));
                     }
+                    res.Remove();
                 }
                 return true;
             }


PJB3005:
Weird, I'd expect bigger savings. the `.xaml` files in content add up to 400 KiB alone. Did you typo the PR desc?

aspiringLich:
The XAML is crunched down into XAMLIL bytecode no?

aspiringLich:
I'd be very surprised if the XML for the UI took up any significant fraction of the size of the pack, because, you know, of all the other stuff in a pack.

PJB3005:
> The XAML is crunched down into XAMLIL bytecode no?

The `.xaml` files are parsed and turned into IL code that constructs the equivalent .NET objects, yes. That's what XAMLIL does. The thing is that with the way XAMLIL works, the `.xaml` files are *first* embedded into the compiled `.dll` file with `EmbeddedResource`, *then* the IL compilation happens in a post-build pass, looking only at the `.xaml` files already embedded into the `.dll` being processed. These `.xaml` files are not used later by the game process. Your PR removes them from the `.dll` file which yes, is fine and a good idea AFAIK.

I tested your PR and I do see the amount of savings I would expect, and can confirm the `.xaml` files are gone from the embedded resources now.

> I'd be very surprised if the XML for the UI took up any significant fraction of the size of the pack, because, you know, of all the other stuff in a pack.

You'd be surprised. XML isn't exactly an information dense format. Like I said the raw `.xaml` files are ~430 KiB for `Content.Client`. By comparison, it's about 3.5 MiB of C# just for `Content.Client`. Yeah, that's also way more than the compiled .NET code.

aspiringLich:
I see.

PR #5132, opened 2024-05-13
Add GetHashCode method to MarkupNode

notquitehadouken:
it's a warning


PJB3005 (changes requested):
This implementation is completely incorrect.

The reason I hadn't personally tried to fix this warning yet is because this entire class is buggy shit and this isn't helping anything. I'd rather have this warning reminding us how bad our code is than muting it with *another* bug.

notquitehadouken:
> This implementation is completely incorrect.
> 
> The reason I hadn't personally tried to fix this warning yet is because this entire class is buggy shit and this isn't helping anything. I'd rather have this warning reminding us how bad our code is than muting it with _another_ bug.

oH okay

PR #5133, opened 2024-05-13
Do not log wrong and correct watchdog token into info logs

VasilisThePikachu:
"Hey PJB, how bad is it to print passwords to console?"
"Quite bad"
"Ok cause you are doing it"
"What are you accusing me off"
- Contrib vc


PJB3005:
Change it to a verbose log instead.

PR #5134, opened 2024-05-13
Add analyzer and fixer for redundant DataField tag arguments

Tayrtahn:
Adds an analyzer to provide an editor hint when a DataField attribute specifies a name that matches what the automatically generated name would be. Also adds a code fixer to automate the removal of these issues.

This is the first time I've messed with code analyzers/fixers, so please **do** review harshly if I did anything wrong, so I can get better at it.

Screenshot 2024-05-13 at 7 51 27 PM
Screenshot 2024-05-13 at 7 51 38 PM
Screenshot 2024-05-13 at 7 51 48 PM

That Fix All option looks pretty tempting...


mirrorcult:
zased

Robust.Analyzers/DataDefinitionAnalyzer.cs
```diff
@@ -56,8 +56,18 @@ public sealed class DataDefinitionAnalyzer : DiagnosticAnalyzer
         "Make sure to add a setter."
     );
 
+    private static readonly DiagnosticDescriptor DataFieldRedundantTagRule = new(
```
PJB3005, line 60: The C# term is "attribute", probably best to use it?

Robust.Analyzers/DataDefinitionAnalyzer.cs
```diff
@@ -248,6 +268,30 @@ private static bool HasAttribute(ITypeSymbol type, string attributeName)
...
+        // No args, no problem
+        if (attribute.ConstructorArguments.Length == 0)
+            return false;
+
+        // If a tag is explicitly specified, it will be the first argument...
+        var tagArgument = attribute.ConstructorArguments[0];
+        // ...but the first arg could also something else, since tag is optional
+        // so we make sure that it's a string
+        if (tagArgument.Value is not string explicitName)
+            return false;
```
PJB3005, line 286: This feels like it's not a good way to test this?

Is the reason it "could be something else" if you have a named arg? Can't you just check if it's a named arg in that case?

Robust.Analyzers/DataDefinitionAnalyzer.cs
```diff
@@ -248,6 +268,30 @@ private static bool HasAttribute(ITypeSymbol type, string attributeName)
...
+
+        // Get the name that sourcegen would provide
+        var fieldName = symbol.Name.AsSpan();
+        var automaticName = $"{char.ToLowerInvariant(fieldName[0])}{fieldName.Slice(1).ToString()}";
```
PJB3005: This code needs to really be shared with the relevant engine code. Use the same MSBuild magic as we use for all the attributes to get access to the code from this project.

PJB3005 (changes requested):

Robust.Analyzers/DataDefinitionAnalyzer.cs
Tayrtahn: I went with "tag" since that's the name of the property/arg that's checked: DataField.Tag

Robust.Analyzers/DataDefinitionAnalyzer.cs
Tayrtahn: Yep, agreed.

Robust.Analyzers/DataDefinitionAnalyzer.cs
PJB3005: Makes sense.

Robust.Analyzers/DataDefinitionAnalyzer.cs
Tayrtahn: I agree that it's funky, and will totally break if we ever add another string arg to DataFieldAttribute.

Yeah, named args were the reason I had in mind.

I'm not sure how to work out if it's a named arg with the API here. `ConstructorArguments` is just an `ImmutableArray<TypedConstant>`, so I can't see a way to check for a name there. `NamedArguments` is an `ImmutableArray<KeyValuePair<String, TypedConstant>>`, but it doesn't seem to include positional args.

I wonder if it would work to iterate over the Values in NamedArguments and see if there's a match? I can give that a shot.

PJB3005: ... do positional arguments actually *go* into `ConstructorArguments`?

Robust.Analyzers/DataDefinitionAnalyzer.cs
Tayrtahn: Well, trying to include the relevant code proved to be a nightmarish mess of requirements, so I just put the tag autogeneration method into a static method on a new utility class and included that. Hope that's okay.

PR #5136, opened 2024-05-14
Move some `EntityCoordinates` methods to `TransformSystem`

ElectroJr:
This PR has several miscellaneous changes for a content PR, including:
- Marks various `EntityCoordinates` methods as obsolete and moves their logic to `TransformSystem`
- Removes various already obsoleted & `EntityCoordinates` methods
- Adds methods to UserInterfaceSystem to get the interface data & ui range.
- Adds the `BoundUserInterfaceMessage` to the `BoundUserInterfaceMessageAttempt` event, so that specific interactrions/messages can be given special treatment (e.g., opening the UI).
- Updates various methods in `ContainerSystem` to take in optional transform components.
- Changes `BoundUserInterfaceCheckRangeEvent` so that it only gets raised directed and not broadcast. This is just to try slightly improve performance, because doing both will just be slower.

Requires space-wizards/space-station-14/pull/27999


PJB3005:
Is it possible to do obsoletions? I really don't want to tank breaking changes here.

ElectroJr:
> Is it possible to do obsoletions? I really don't want to tank breaking changes here.

AFAIK I removed all the breaking changes, though content tests are still failing because of a new error log, but that was an existing error that just wasn't being caught before by tests.

PR #5137, opened 2024-05-14
Add remark about control right alignment in box containers

ShadowCommander:


PJB3005 (changes requested):
This should really be a general-purpose explanation in the docs for the `Expand` properties instead. In fact the recommendation here doesn't even make sense for a lot of use cases.

One-off comments like these feel like an extremely bad way to handle explanation of how the UI layout system works in general.

PR #5139, opened 2024-05-14
Improve markup node equals method and implement `GetHashCode`

juliangiebel:
What it says on the tin.

The equals method wasn't implemented correctly. I looked up how it and `GetHashCode` is supposed to be implemented and fixed the implementation accordingly.


PJB3005 (changes requested):
Your `GetHashCode` implementation is wrong, you're getting the reference equality has for the dictionary.

PJB3005:
Replaced by 38c227b692344ab86b58057e6cbabc8807ed7517

PR #5145, opened 2024-05-18
perf: Budget less memory for Replay Checkpoints (#28052)

tom-leys:
- Half the checkpoint RAM budget overall (1GB instead of 2GB)
- Simple change of the CVars and some stats so checkpoint creation triggers less frequently
- Based on a Lizard replay, checkpoints move from on average every 70 ticks to every 350.
- Added a minimum number of ticks between any two checkpoints (default 60 ticks)
- More likely to use a checkpoint when panning through the timeline (previously you had to jump quite far for it to look for a checkpoint)

Relates to https://github.com/space-wizards/space-station-14/issues/28052

Before (6.33GB before final load):

image

After (5.35GB total):

image

Naturally this will reduce seek performance a bit.


Robust.Shared/CVars.cs
```diff
@@ -1647,20 +1647,25 @@ protected CVars()
...
         /// Maximum number of ticks before a new checkpoint tick is generated.
         /// </summary>
-        public static readonly CVarDef<int> CheckpointInterval = CVarDef.Create("replay.checkpoint_interval", 200);
+        public static readonly CVarDef<int> CheckpointInterval = CVarDef.Create("replay.checkpoint_interval", 500);
 
         /// <summary>
         /// Maximum number of entities that can be spawned before a new checkpoint tick is generated.
         /// </summary>
-        public static readonly CVarDef<int> CheckpointEntitySpawnThreshold = CVarDef.Create("replay.checkpoint_entity_spawn_threshold", 100);
+        public static readonly CVarDef<int> CheckpointEntitySpawnThreshold = CVarDef.Create("replay.checkpoint_entity_spawn_threshold", 2000);
```
ElectroJr: Was this meant to be 200? 2000 seems like a lot. Spawning entities is pretty slow, quite a lot slower than applying states. 2k ents would make scrubbing lag quite a lot.

Actually maybe the fact that solution and audio/sounds are now ents also mean that it does need to be significantly higher than before

Robust.Client/Replays/Loading/ReplayLoadManager.Checkpoints.cs
```diff
@@ -150,13 +155,28 @@ TimeSpan GetTime(GameTick tick)
             serverTime[i] = GetTime(curState.ToSequence) - initialTime;
             ticksSinceLastCheckpoint++;
 
-            if (ticksSinceLastCheckpoint < _checkpointInterval
+            if (ticksSinceLastCheckpoint < _checkpointMinInterval ||
+                (ticksSinceLastCheckpoint < _checkpointInterval
                 && spawnedTracker < _checkpointEntitySpawnThreshold
-                && stateTracker < _checkpointEntityStateThreshold)
+                && stateTracker < _checkpointEntityStateThreshold))
             {
```
ElectroJr: Maybe split the if statement just to make it slightly easier to parse the logic while reading. I.e., 
```cs
            if (ticksSinceLastCheckpoint < _checkpointMinInterval)
                continue;

            if (ticksSinceLastCheckpoint < _checkpointInterval
                && spawnedTracker < _checkpointEntitySpawnThreshold
                && stateTracker < _checkpointEntityStateThreshold)
            {
                continue;
            }
```
I'm mostly saying that because I got confused when reading it and thought it was `( (a || b) && c && d)` instead of `( a || ( b && c && d))`

Robust.Shared/CVars.cs
tom-leys: This was just evidence based. I wanted checkpoints to occur around every 500 ticks but they were instead occurring every 70 ticks on average, triggered by the spawn threshold.

I had to lift the threshold this much to spread the checkpoints out to once every 380 ticks on average.

As for the "make scrubbing lag quite a lot" I have another solution to that I'm going to submit in a different PR where we don't try to seek directly to the user's goal in a single frame but instead spread that CPU load over several frames.

tom-leys:
New PR https://github.com/space-wizards/RobustToolbox/pull/5152 helps mitigate effects of less checkpoints by spreading the processing of later ticks over many frames.

PR #5146, opened 2024-05-18
Perf: Avoid a copy of ComponentChanges every tick within Checkpoints

tom-leys:
- Also remove temporary Dictionary created every tick * every change
- Reduces GC load, 8GB less temporary allocations on large replays.
- Makes processing step over 2x as fast on my system (though I am testing that with a memory profiler running)
- Use delta.ApplyToFullState when possible avoiding 1GB of temporaries for Gas tiles for instance.

Related to https://github.com/space-wizards/space-station-14/issues/28052 though this PR reduces memory thrashing rather than final memory usage.

Before  (showing allocations during processing replay step):
image

After: 
image

After using ApplyToFullState:
image

I tested using this replay:
https://moon.spacestation14.com/replays/lizard/2024/05/18/lizard-2024_05_18-16_22-round_53948.zip


Robust.Client/Replays/Loading/ReplayLoadManager.Checkpoints.cs
```diff
@@ -359,6 +405,63 @@ TimeSpan GetTime(GameTick tick)
...
+            DebugTools.Assert(existing.State is IComponentDeltaState fullDelta && fullDelta.FullState);
+            var internalState = existing.State;
+            if (modifyInPlace)
+            {
+                delta.ApplyToFullState(internalState);
+            }
+            else
+            {
+                internalState = delta.CreateNewFullState(existing.State);
+            }
```
ElectroJr: I think its never safe to assume `existing.State` is currently "owned" by the scratch and not a shared reference with some other state. Unless you can somehow guarantee that the state was cloned when the scratch was first created. So I think modifyInPlace just needs to be removed / always false for now? 

You could probably track whether or not a given state/netid is safe to modify to try reduce allocs but its probably better to do that in a separate PR, so ill just revert those changes for now. Otherwise AFAICT it look all good?

tom-leys: Actually yes, we can be confident that it is safe, because of the way the scratch data is only created between checkpoints.

ElectroJr: I'm not sure what you mean. Only being created between checkpoints only ensures that the component state object won't be the same as one stored in some previous checkpoint, but not that it won't be the same one stored in a normal state between the checkpoints

tom-leys: I'm not sure how much detail I need to go in to in order to explain this. I'll try a simple explanation. 

This PR introduced a scratch data object. This holds only changes since the last checkpoint.

When we first encounter a change for an entity that is unchanged, we create a scratch data object for that entity and we set modifyInPlace to false. This uses CreateNewFullState as above to create a copy which will be used in the next checkpoint.

If we encounter additional changes for the same entity, we'll find that entity in the scratch data, so this routine will be called with modifyInPlace set to false. Then ApplyToFullState will be used instead. 

When we reach the checkpoint, the copies (modified along the way) are stored into the checkpoint and the scratch data is cleared. The next tick processed will be back to copying data.

This has a huge performance impact on large memory structures with small changes, specifically gas tiles. Overall copies allocate about 2 GB of copies without this optimisation.

Neither ApplyToFullState or CreateNewFullState are applied on the tick data, they are only applied to copies of it stored in the scratch data.

In the changed code, delta is the tick data (which we should not modify) and internalState is the state we copy to / modify in place. I am trusting that ApplyToFullState and CreateNewFullState do not change delta, only internalState / existing.State (which is the scratch data)

ElectroJr: > Neither ApplyToFullState or CreateNewFullState are applied on the tick data, they are only applied to copies of it stored in the scratch data.

This is the problem I am having, as far as I can tell you are not cloning/copying the data into the scratch, you are just copying references to objects, and then modifying those objects via `ApplyToFullState`. which will also modify previous states & checkpoints. And even if you were copying that data, I think that would probably be even worse for performance.

Lets say we have an entity with two components called A and B:
- We start collecting data for a new checkpoint and at some point encounter a game-state with an ent-state for our entity, which has a new full comp-state for component A.
- As the entity hasn't been modified yet during this checkpoint generation, we use `scratch = new UpdateScratchData(oldEntState)` to create new scratch data using `oldEntState`, which is the **full** entity-state that contains all of the comp-states that the entity had when the previous checkpoint was generated. 
- When the last checkpoint was generated, the comp-states are not cloned, `oldEntState` and thus our scratch now contains the same comp-state objects that are being used in previous checkpoints or even individual game-states.
- Then we set modifyInPlace = false, and apply the game-state for the next tick (which only contains a full comp-state for comp A),
- In this case `modifyInPlace ` doesn't matter as we just update the dictionary value for the comp A state
- The comp A state is now an object shared with at most one game-state, while the comp B state is still shared with other checkpoints or another game-state.
- If later on we then get any delta state for either comp A or compB, `modifyInPlace` will now be false, so we would use `ApplyToFullState()` to modify the state for A or B currently stored in scratch
- But this then means we've now modified the comp-state that is currently also being referenced by other game-states and possibly other checkpoints.

I think the only way you can safely use `ApplyToFullState` is if you actively track whether you can know for sure that the object currently stored in the scratch is **only** stored in scratch and not in any other game-state or checkpoint, And I think the only way you can know that is if you know you have used `CreateNewFullState` at least once for that component since you created the new scratch data. So you need to track "owned" states in a `HashSet<NetId>`, and "un-own" them whenever applying a non-delta state, creating a new checkpoint, or any other case where the object might stop being unique to the scratch.

PR #5148, opened 2024-05-19
Use non-generic `TryComp()` for metadata & transform

ElectroJr:
Allows entity systems to use the cached xform & metadata queries

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/ContainerSystem.cs
@@ -242,7 +242,7 @@ namespace Robust.Client.GameObjects
 #if DEBUG
             var uid = GetEntity(netEntity);
 
-            if (TryComp<MetaDataComponent>(uid, out var meta))
+            if (TryComp(uid, out MetaDataComponent? meta))
             {
                 DebugTools.Assert((meta.Flags & ( MetaDataFlags.Detached | MetaDataFlags.InContainer) ) == MetaDataFlags.Detached,
                     $"Adding entity {ToPrettyString(uid)} to list of expected entities for container {container.ID} in {ToPrettyString(container.Owner)}, despite it already being in a container.");
--- a/Robust.Client/GameObjects/EntitySystems/EyeSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/EyeSystem.cs
@@ -56,7 +56,7 @@ public sealed class EyeSystem : SharedEyeSystem
             if (eyeComponent.Eye == null)
                 continue;
 
-            if (!TryComp<TransformComponent>(eyeComponent.Target, out var xform))
+            if (!TryComp(eyeComponent.Target, out TransformComponent? xform))
             {
                 xform = Transform(uid);
                 eyeComponent.Target = null;
--- a/Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
@@ -508,7 +508,7 @@ public abstract partial class SharedMapSystem
     private void OnGridRemove(EntityUid uid, MapGridComponent component, ComponentShutdown args)
     {
         Log.Info($"Removing grid {ToPrettyString(uid)}");
-        if (TryComp<TransformComponent>(uid, out var xform) && xform.MapUid != null)
+        if (TryComp(uid, out TransformComponent? xform) && xform.MapUid != null)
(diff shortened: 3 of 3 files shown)


moonheart08:
This doesn't change the semantics of the code, the generic is just implied now.

deltanedas:
the generic one didnt use query when the explicit one did

moonheart08:
image
I am seemingly wrong, but man this is bad API design then.

edit: in fact why does C# even allow overloading like this, ???

deltanedas:
yeah imo TryTransform and TryMeta would be better

ElectroJr:
> I am seemingly wrong, but man this is bad API design then.
> 
> edit: in fact why does C# even allow overloading like this, ???

I'd be much happier if I could just override the generic `TryComp<T>` methods with specialized variants as well and then it wouldn't be an issue, but AFAIK you can't do that. But IIRC our convention was already to prefer the `TryComp(uid, out T? comp)` over `TryComp<T>(uid, out var comp)`, though I can't seem to find it in the convention docs.

> yeah imo TryTransform and TryMeta would be better

I feel like it it would probably just confuse contributors if transform & metadata had to use their own special function to fetch the component. If `Comp<T>()` could be overridden I'd rather just remove the existing `Transform()` & `Metadata()` methods.

moonheart08:
> > I am seemingly wrong, but man this is bad API design then.
> > edit: in fact why does C# even allow overloading like this, ???
> 
> I'd be much happier if I could just override the generic `TryComp<T>` methods with specialized variants as well and then it wouldn't be an issue, but AFAIK you can't do that. But IIRC our convention was already to prefer the `TryComp(uid, out T? comp)` over `TryComp<T>(uid, out var comp)`, though I can't seem to find it in the convention docs.
> 
> > yeah imo TryTransform and TryMeta would be better
> 
> I feel like it it would probably just confuse contributors if transform & metadata had to use their own special function to fetch the component. If `Comp<T>()` could be overridden I'd rather just remove the existing `Transform()` & `Metadata()` methods.

Given the fact the type is generics, doing if/else type checks in the generic impl would be optimized out by ryujit down to what we explicitly have written here.

ElectroJr:
> Given the fact the type is generics, doing if/else type checks in the generic impl would be optimized out by ryujit down to what we explicitly have written here.

I guess I'll give that a go later today and use that if benchmarks show its just as fast

moonheart08:
> > Given the fact the type is generics, doing if/else type checks in the generic impl would be optimized out by ryujit down to what we explicitly have written here.
> 
> I guess I'll give that a go later today and use that if benchmarks show its just as fast

https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACAZQAtsoAHAGW2ADoASgFcAdhgCW+GAG4a9BsXKDREqXwDCEfB3EAbGFCYGAbuLAxcs6nIDMC0g3UMA3jQbuGAbQBS4jAHEYEQMzAAoMAE8OGAgAM1DxMQBKJLQGAF03D08AWRgMFggAEwBJbV1QvILiso5dAHkOCQgRXD4AQQBzTthcXHFjGBKRXUTEzqTM6g8FO0SMBhyAHgAVAD5QiCEFlYZjJJcsmfdxWIZwqJj4lYOAXluGSOi40KYMKHGUw+nj372GB6hG6bYAAKxgYAwSQARNgAO7w7DQqx/Y7EADsDHIKL+AF8GDBdLgYN9UTNjACGEUYLFsEJdBgcb9cUdjqyZhiGEgAJxMhgsn7uVnEOZiBgAIU222Y73GDAQB1cguOAHoVY8oNhOjAihEQAxcGxOLpeAwwNgRAByBbEknzCCPFjiXBm7C6XQMWLQA1aEmwPAtPjsjycnJShYKvkC3FAA===

would be my answer if sharplab could see into the reftype generic, but you *can* still see the opt occur on the `int` specialization of `M`.

ElectroJr:
After actually benchmarking, it seems like it the difference between using or not using the transform/metadata query is pretty negligible and the PR might be unnecessary? Which is quite surprising to me, as, I would've thought the array index would still have noticeable impact. Previous benchmarks did have a difference, but AFAICT that was actually just because of interface overhead, and the differences largely go away when using `EntityManager` instead of `IEntityManager`. It's possible that it does still matter and the benchmark just isn't reliable for some reason? 

Benchmark branches:
https://github.com/ElectroJr/RobustToolbox/tree/bench2
https://github.com/ElectroJr/space-station-14/tree/bench2

PR #5151, opened 2024-05-19
Adds API to ScrollContainers to be able to set the scroll handle position acurrately

AJCM-git:
What it says on the tin, basically adds 3 methods `SetScrollValueToPercentage` and a vertical and horizontal version that serve as wrappers for `SetScrollValue`


PJB3005:
What do you need this for?

AJCM-git:
> What do you need this for?

To access the scrollbar position methods, for example, I'm making a messenger PDA app and i want to scroll to the bottom when entering a profile or when sending/being sent a message

image


https://github.com/space-wizards/RobustToolbox/assets/60196617/9f3be7cc-5501-4570-9b8c-14b594156d7e


I assumed having access to the engine-side children of a control to change behavior was guzzy since things like AnimatedTextureRect.DisplayRect do it, is it a problem here?

moonheart08:
It would be better API design to add a scroll-to-target API to ScrollContainer instead.

AJCM-git:
> It would be better API design to add a scroll-to-target API to ScrollContainer instead.

Took a small break and did that

Robust.Client/UserInterface/Controls/ScrollContainer.cs
```diff
@@ -322,5 +312,66 @@ private void _scrollValueChanged(Range obj)
...
+        #region Scrollbat Set Methods
+
+        /// <summary>
+        /// Sets the scrollbars thumbs position to a value from 1 (the start of the scrollbar) to 100 (the end of the scrollbar)
+        /// </summary>
+        /// <param name="value">
+        /// The value from 1 to 100 that we want to put the scrollbars thumbs in.
+        /// Any value higher than 100 will just be interpreted as 100
+        /// </param>
+        public void SetScrollValueToPercentage(int? hValue, int? vValue)
```
PJB3005, line 325: These functions should be called "ratio", take float parameters and take values on the scale 0->1, not 0->100

PJB3005 (changes requested):

PR #5153, opened 2024-05-20
LineEdit placeholder tweaks

AJCM-git:
Adds a new field `HidePlaceHolderOnFocus` that dictates whether or not to hide a placeholder when we focus on the line input

### DIFF
--- a/Robust.Client/UserInterface/Controls/LineEdit.cs
+++ b/Robust.Client/UserInterface/Controls/LineEdit.cs
@@ -8,7 +8,6 @@ using Robust.Shared;
 using Robust.Shared.Configuration;
 using Robust.Shared.Input;
 using Robust.Shared.IoC;
-using Robust.Shared.Log;
 using Robust.Shared.Maths;
 using Robust.Shared.Timing;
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -186,6 +187,10 @@ public int SelectionStart
         public int SelectionLower => Math.Min(_selectionStart, _cursorPosition);
         public int SelectionUpper => Math.Max(_selectionStart, _cursorPosition);
 
+        public bool Focused;
```
PJB3005: We already have `HasKeyboardFocus()`.

Robust.Client/UserInterface/Controls/LineEdit.cs
```diff
@@ -186,6 +187,10 @@ public int SelectionStart
...
+
+        public bool HidePlaceHolderOnFocus;
```
PJB3005: This needs to be a property.

PJB3005 (changes requested):

PR #5156, opened 2024-05-21
Stream replay loading using seperate Decompression and Seserialize tasks

tom-leys:
- Improves memory locality by immediately processing replay data that just got loaded into checkpoints (no time to flush from caches & page to disc)
- Reads data in parallel to checkpoint processing so 2x faster by better using multiple CPU cores
- Improves https://github.com/space-wizards/space-station-14/issues/28052 by ensuring low memory systems don't have to do a second pass over replay data that has likely been paged out 

How it works
- A decompression task opens and decompresses replay files, handing those MemoryStreams on using a channel
- A deserialization task processes each replay file, creating states and passing those on using a channel
- Checkpoint processing remains in the main thread, reading states as they are deseralized and processing these

On my system, the deserialization task remains a bottleneck (after other Replay Prs merged). We could deserialize with two tasks instead of one to further improve performance but I didn't want to over-complicate this PR.

There's also reuse of decompressed memory stream memory.

- Respects the Job system, using wrapped async waits if there is a Loading Job


ghost:
Tom Leys, you are great

tom-leys:
> Tom Leys, you are great

Thanks Diesel :)

It's a pity I don't contribute any more but I've moved on to other things now.

PR #5158, opened 2024-05-22
msbuild option for generating xmldoc file for engine client / shared / server

mirrorcult:
need this for something secret (useful)

so i can `dotnet build --property:RobustGenerateXMLDocFile=true` and just have it work

### DIFF
--- a/MSBuild/Robust.Properties.targets
+++ b/MSBuild/Robust.Properties.targets
@@ -19,7 +19,7 @@
     <EnableClientScripting>False</EnableClientScripting>
     <!-- Client scripting is only enabled on tools builds for security and size reasons. -->
     <EnableClientScripting Condition="'$(RobustToolsBuild)' == 'true'">True</EnableClientScripting>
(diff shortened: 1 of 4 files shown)


MSBuild/Robust.Properties.targets
```diff
@@ -19,6 +19,7 @@
     <EnableClientScripting>False</EnableClientScripting>
     <!-- Client scripting is only enabled on tools builds for security and size reasons. -->
     <EnableClientScripting Condition="'$(RobustToolsBuild)' == 'true'">True</EnableClientScripting>
+    <RobustGenerateXMLDocFile>False</RobustGenerateXMLDocFile>
```
mirrorcult, line 22: this isnt strictly necessary but mostly putting it here so i can just set this to true when i want it to always generate for debugging

moonheart08 (approved):
the does does indeed do the thing

mirrorcult:
https://github.com/space-wizards/space-station-14/pull/28194#issuecomment-2132065397

PR #5159, opened 2024-05-22
make entity query constructors internal

deltanedas:
if you accidentally do `var query = new EntityQueryEnumerator<Whatever>();` it still compiles and you get a microsoft nullpointerexception *at runtime* which is bad

now that they are internal if you do this you get a compile error

nothing in content or outside of gameobjects constructs them so this should Just Work

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -35,7 +35,7 @@ END TEMPLATE-->
 
 ### Breaking changes
 
-*None yet*
+* `EntityQuery` and `EntityQueryEnumerator` constructors are now internal to avoid mistakes.
 
 ### New features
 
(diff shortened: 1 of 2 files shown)


mirrorcult:
why does that even compile

deltanedas:
no fucking idea but i have been trolled by it for the last time

moonheart08:
did you verify that this makes it not compile? because the issue seems to be the struct default constructor (which is implied and not written explicitly), not any of the other ones.

deltanedas:
> did you verify that this makes it not compile? 

no :trollface:

is it possible to remove the empty constructor by doing `private EntityQueryEnumerator()` or something so its explicit

moonheart08:
> > did you verify that this makes it not compile?
> 
> no :trollface:
> 
> is it possible to remove the empty constructor by doing `private EntityQueryEnumerator()` or something so its explicit

You explicitly declare it to prevent this, yea.

deltanedas:
> The parameterless struct constructor must be 'public'.

trolled by: bill gate

moonheart08:
No point to this PR then.

deltanedas:
shouldve non-microsoft java

PR #5161, opened 2024-05-23
Add `IConfigurationManager.OnCvarValueChanged` event

ElectroJr:
Adds an event that gets invoked whenever any cvar value changes. Required for a content PR.

This also changes some cvar setting logic & adds some debug asserts to ensure that cvars never change their type once they have been registered.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -40,7 +40,7 @@ END TEMPLATE-->
 ### New features
 
 * A new `replay.checkpoint_min_interval` cvar has been added. It can be used to limit the frequency at which checkpoints are generated when loading a replay.
-* Added `IConfigurationManager.OnCvarValueChanged`. This is a c# event that gets invoked whenever any cvar value changes.
+* Added `IConfigurationManager.OnCVarValueChanged`. This is a c# event that gets invoked whenever any cvar value changes.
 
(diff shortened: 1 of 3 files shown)


RELEASE-NOTES.md
```diff
@@ -40,6 +40,7 @@ END TEMPLATE-->
 ### New features
 
 * A new `replay.checkpoint_min_interval` cvar has been added. It can be used to limit the frequency at which checkpoints are generated when loading a replay.
+* Added `IConfigurationManager.OnCvarValueChanged`. This is a c# event that gets invoked whenever any cvar value changes.
```
PJB3005: I think `OnCVarValueChanged` would fit existing naming better?

Robust.Shared/Configuration/IConfigurationManager.cs
```diff
@@ -8,8 +8,13 @@ namespace Robust.Shared.Configuration
     /// <summary>
     /// Additional information about a CVar change.
     /// </summary>
-    public readonly struct CVarChangeInfo
+    public readonly struct CVarChangeInfo(string name, GameTick tickChanged, object newValue, object oldValue)
```
PJB3005: This makes the constructor public which I don't like, as it means adding further parameters is a breaking change.

PR #5167, opened 2024-05-24
Add a launch server bat file on build

VasilisThePikachu:
People often have issues by not having net runtime installed when trying to start a sandbox. And the exe instantly closes before they can see any errors. Some users dont know how to use CMD in order to get this error shown. So we can force them to use cmd and review any errors while keeping it user friendly.

I can make a macos/linux version too if desired.


MACMAN2003:
mmmm lunch server

VasilisThePikachu:
english skill issue

deltanedas:
lunchpad mindustry

PR #5168, opened 2024-05-24
Add FormattedMessage.RemoveMarkupPermissive

Tayrtahn:
Adds a version of FormattedMessage.RemoveMarkup that uses FromMarkupPermissive so it doesn't throw an exception on bad markup.

Intended to make space-wizards/space-station-14#28256 very slightly cleaner, and just good to have.


ElectroJr:
#5170 also contains a method for this

mirrorcult:
closing in favor of #5170

PR #5175, opened 2024-05-28
Better parallel exception logging

metalgearsloth:
weh

image

### DIFF
--- a/Robust.Shared/Threading/ParallelManager.cs
+++ b/Robust.Shared/Threading/ParallelManager.cs
@@ -3,6 +3,7 @@ using System.Threading;
 using Microsoft.Extensions.ObjectPool;
 using Robust.Shared.Configuration;
 using Robust.Shared.IoC;
+using Robust.Shared.Log;
 
 namespace Robust.Shared.Threading;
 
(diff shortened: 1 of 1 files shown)


Robust.Shared/Threading/ParallelManager.cs
```diff
@@ -234,14 +256,23 @@ public void Set(IParallelRobustJob robust, int start, int end, ParallelTracker t
...
+            }
+            catch (Exception exc)
+            {
+                _sawmill.Error($"Exception in ParallelManager: {exc.StackTrace}");
+            }
+            finally
+            {
+                // Set the event and return it to the pool for re-use.
+                _tracker.Set();
+                _parentPool.Return(this);
```
DrSmugleaf, line 274: Could there be a case where this is returning a broken internal job instance that will keep throwing when reused?

DrSmugleaf (approved):

Robust.Shared/Threading/ParallelManager.cs
metalgearsloth: Jobs get dumped back into the objectpool so I wouldn't think anything like that would happen ?At best stuff that throws every tick would keep logging but that's already an issue.

PR #5177, opened 2024-05-29
Add some System.Random methods

metalgearsloth:
a


Robust.Shared/Random/RandomExtensions.cs
```diff
@@ -128,6 +129,9 @@ public static Angle NextAngle(this System.Random random, Angle minAngle, Angle m
         return minAngle + (maxAngle - minAngle) * random.NextDouble();
     }
 
+    public static Vector2 NextVector2(this System.Random random, float minMagnitude, float maxMagnitude)
```
PJB3005: Should probably be like `NextPolarVector2` or something to make it clear how it's generated.

Issue #4976, opened 2024-03-18
Version of SpriteView for EntityPrototypes


Plykiya:
This is... kind of a request for a feature? I don't really know how to implement this myself... or if I'm really understanding this properly.

I've been working on https://github.com/space-wizards/space-station-14/pull/26178 and encountered a problem with EntityPrototypes not really having a way to support layering of textures, so I'm stuck with just the base texture appearance. A lot of the textures I'm working with are pretty much only differentiable as a result of the additional sprite layers, without it they all look the same.

My only options at the moment are to pretend it's not a problem (bad), create custom icons for each LatheRecipe (bad), or to just pray and hope someone helps me (sad but best choice)

moonheart08:
> This is... kind of a request for a feature? I don't really know how to implement this myself... or if I'm really understanding this properly.
> 
> I've been working on [space-wizards/space-station-14#26178](https://github.com/space-wizards/space-station-14/pull/26178) and encountered a problem with EntityPrototypes not really having a way to support layering of textures, so I'm stuck with just the base texture appearance. A lot of the textures I'm working with are pretty much only differentiable as a result of the additional sprite layers, without it they all look the same.
> 
> My only options at the moment are to pretend it's not a problem (bad), create custom icons for each LatheRecipe (bad), or to just pray and hope someone helps me (sad but best choice)

What you should do is spawn a copy of the entity client-side and have a view showing that client-side version.
Drawing prototypes directly won't ever be supported, but drawing client-side entities works perfectly fine. Look at guidebook's entity view for an example.

PR #5185, opened 2024-05-30
EntityPrototypeView

EmoGarbage404:
New control based off of SpriteView that lets you chuck in an entity prototype and use that instead.

The control manages all the logic of spawning and deleting the entity.

Can be used in content in places where just using the prototype causes issues with layer coloring or incorrect logic and such.


Robust.Client/UserInterface/Controls/EntityPrototypeView.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        if (_currentPrototype != null)
+        {
+            _ourEntity = EntMan.Spawn(_currentPrototype);
+            SpriteSystem.ForceUpdate(_ourEntity.Value);
+            SetEntity(_ourEntity);
+        }
+    }
+
+    protected override void Dispose(bool disposing)
```
PJB3005: Do not use control disposal, use `EnteredTree` and `ExitedTree` instead.

Plykiya:
solves #4976 🥲

PR #5191, opened 2024-06-01
Item Lists use list of textures, add methods to render layered textures

Plykiya:
Sorry for the bunch of automatic formatting changes

- Adds a DrawLayeredTextureRectRegion function that just does what DrawTextureRectRegion does but with a foreach for each texture (should I just not add this basically duplicate function and just foreach on DrawTextureTectRegion instead?)
- Changes ItemList's Icon from a Texture to a List of Textures, added an overload to AddItem() that supports a single texture and a list of textures

The queue for lathe recipes uses ItemList, ItemList doesn't support LayeredTextures so the icons appear wrong, this lets me fix it

Before:

https://github.com/space-wizards/RobustToolbox/assets/58439124/57ed2f67-3dd3-42a1-8018-ae7cd39299f7

After:

https://github.com/space-wizards/RobustToolbox/assets/58439124/7c136934-0d32-43d4-b313-9a451c057df5

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.RenderHandle.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.RenderHandle.cs
@@ -6,6 +6,7 @@ using Robust.Shared.GameObjects;
 using Robust.Shared.Maths;
 using OpenToolkit.Graphics.OpenGL4;
 using Robust.Shared.Graphics;
+using System.Collections.Generic;
 
 namespace Robust.Client.Graphics.Clyde
 {
@@ -124,13 +125,13 @@ namespace Robust.Client.Graphics.Clyde
                     sr = subRegion ?? new UIBox2(0, 0, texture.Width, texture.Height);
                 }
 
-                var clydeTexture = (ClydeTexture) texture;
+                var clydeTexture = (ClydeTexture)texture;
                 return clydeTexture;
             }
 
             public void RenderInRenderTarget(IRenderTarget target, Action a, Color? clearColor)
             {
-                _clyde.RenderInRenderTarget((RenderTargetBase) target, a, clearColor);
+                _clyde.RenderInRenderTarget((RenderTargetBase)target, a, clearColor);
             }
 
             public void SetScissor(UIBox2i? scissorBox)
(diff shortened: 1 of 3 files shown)


Plykiya:
I handled the content that throws an error in the other PR in the other repo

PJB3005:
I would seriously rather we just stop using ItemList than pile on more bad hacks on this horribly designed control.

Plykiya:
regardless of whatever control you want me to use, I'm going to need something that uses a list of textures rather than a single texture for the icon...

PJB3005:
Aren't lathe icons just pulled from the prototypes?

Plykiya:
> Aren't lathe icons just pulled from the prototypes?

yeah, originally the code just used this which returns a single texture
`icon = recipe.Icon == null
                ? _spriteSystem.GetPrototypeIcon(recipe.Result).Default 
                : _spriteSystem.Frame0(recipe.Icon) ;`
which leads to issues when there are certain textures that are blank by default and only change their appearance using layers of textures
I've been using
`textures = SpriteComponent.GetPrototypeTextures(entityProto, _resources).Select(o => o.Default).ToList();`
to get a list of all the textures the proto has, but I've been running into issues with some controls that are currently only designed to accept a Texture as opposed to a List of Textures

PJB3005:
I could've sworn we had a proper way to display entity prototypes directly...

Regardless, we already have `LayeredTextureRect` which is what the spawn panel uses. There's no need to add stuff to `DrawingHandle` directly.

EmoGarbage404:
> I could've sworn we had a proper way to display entity prototypes directly...

i made one tho

https://github.com/space-wizards/RobustToolbox/pull/5185

Plykiya:
> > I could've sworn we had a proper way to display entity prototypes directly...
> 
> i made one tho
> 
> #5185

oh, sick
pretty sure I can just use that if it does what I think it does, I'd have to migrate the code away from using the ItemList control though

Plykiya:
alright I'll close my PRs for now and wait until the other PR gets merged, I know what to do

ShadowCommander:
Look at ListContainer for doing lists of things with more customizable buttons.

PR #5192, opened 2024-06-01
Fixed incmd command

TornadoTechnology:
# About PR
Fix and clean up `incmd` command.

# About error
Command `incmd` generated `FullInputCmdMessage` with `NetCoordinates` and `NetEntity`, command handler try cast this messeg to `ClientFullInputCmdMessage` what was causing the client to crash.


Robust.Client/GameObjects/EntitySystems/InputSystem.cs
```diff
@@ -147,17 +148,47 @@ private void GenerateInputCommand(IConsoleShell shell, string argstr, string[] a
...
+
+                wOffset = new Vector2(wX, wY);
+            }
 
+            var coords = EntityCoordinates.FromMap(pent, _transform.GetMapCoordinates(pent).Offset(wOffset), _transform, EntityManager);
             var funcId = _inputManager.NetworkBindMap.KeyFunctionID(keyFunction);
 
-            var message = new FullInputCmdMessage(_timing.CurTick, _timing.TickFraction, funcId, state,
-                GetNetCoordinates(coords), new ScreenCoordinates(0, 0, default), NetEntity.Invalid);
+            var message = new ClientFullInputCmdMessage(_timing.CurTick,
```
TornadoTechnology, line 202: It's can be write without overloading constructor, but i think this better

Example:
```cs
var message = new ClientFullInputCmdMessage(_timing.CurTick, _timing.TickFraction, funcId)
{
    Coordinates = coords,
    ScreenCoordinates = new ScreenCoordinates(0, 0, default),
    State = state,
    Uid = EntityUid.Invalid
};
```

PR #5210, opened 2024-06-05
Remove obsolete anchor entity functions

Plykiya:
Requires https://github.com/space-wizards/space-station-14/pull/28613

- Remove obsolete AnchorEntity() variants in favor of Entity<T> versions
- Update TransformComponent and SharedTransformSystem to use the Entity<T> versions by tupling the Transform and Grid Entities
- Update tests to use the Entity<T> version by tupling the Transform and Grid entities
- OnInitAnchored_AddedToLookup test had all of its obsolete functions updated to preferred methods while I was trying to figure out why it was failing
- Got rid of an unnecessary debug assertion on line 117 that doesn't actually check for a problem

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
@@ -232,10 +232,12 @@ public abstract partial class SharedTransformSystem
             return;
 
         MapGridComponent? grid;
+        EntityUid gridUid;
 
         // First try find grid via parent:
         if (component.GridUid == component.ParentUid && TryComp(component.ParentUid, out MapGridComponent? gridComp))
         {
+            gridUid = component.ParentUid;
             grid = gridComp;
         }
         else
@@ -243,7 +245,7 @@ public abstract partial class SharedTransformSystem
             // Entity may not be directly parented to the grid (e.g., spawned using some relative entity coordinates)
             // in that case, we attempt to attach to a grid.
             var pos = new MapCoordinates(GetWorldPosition(component), component.MapID);
-            _mapManager.TryFindGridAt(pos, out _, out grid);
+            _mapManager.TryFindGridAt(pos, out gridUid, out grid);
         }
 
         if (grid == null)
@@ -252,7 +254,7 @@ public abstract partial class SharedTransformSystem
             return;
         }
 
-        if (!AnchorEntity(uid, component, grid))
+        if (!AnchorEntity((uid, component), (gridUid, grid)))
             component._anchored = false;
     }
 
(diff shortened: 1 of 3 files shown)


Plykiya:
hrmmmm
all the tests use .anchor which is another, different obsoleted thing... let's see if changing that out fixes the test fails

Plykiya:
Okay so I went from 9 test failures to 7 test failures, so the two tests that I properly handed over the tile indices over to the AnchorEntity() function are now succeeding...

Plykiya:
oh wow... another test fail gone
I guess I'll do the same thing for the other six then?

Plykiya:
Guh, don't know what to do about that last test
It's just not anchoring properly after initialization

Plykiya:
Rather than adding TileIndices at every possible turn to get this to work
Maybe I should just... not, add the TileIndices and deal with the Debug Assertion on line 99 of SharedTransformSystem.Component.cs as a sign that something is very wrong if the variant of AnchorEntity() without TileIndices fails to pass

Plykiya:
So my working theory is that specific debug assert is just flat out... unnecessary
So I removed that specific debug assert condition in [SharedTransformSystem.Component.cs](https://github.com/space-wizards/RobustToolbox/pull/5210/files#diff-c3a411183a1e422acc4dee6caaef3f81aaec6c9f6550fc42c5d9a7f3ae329144)[SharedTransformSystem.Component.cs](https://github.com/space-wizards/RobustToolbox/pull/5210/files#diff-c3a411183a1e422acc4dee6caaef3f81aaec6c9f6550fc42c5d9a7f3ae329144)[SharedTransformSystem.Component.cs](https://github.com/space-wizards/RobustToolbox/pull/5210/files#diff-c3a411183a1e422acc4dee6caaef3f81aaec6c9f6550fc42c5d9a7f3ae329144) and undid all the TileIndices, so they all use the AnchorEntity() version that was throwing the dumb debug assertion error

The reason I think this is the case is because the version of AnchorEntity() with tile indices has no debug assertion and it just... works
So if I remove the debug assertion for the version of AnchorEntity() without tile indices it should just work because it's running the same function in the end anyways

Plykiya:
oh hey it worked, nice

Plykiya:
so in summary
the tests weren't actually failing as a result of their test conditions throwing an assertion
the tests were "failing" because there was just an AnchorEntity() function with an unnecessary debug assertion even though the code works in the conditions it assumed would be problematic. every test passes with that debug assertion removed

Plykiya:
I'll take a look at this again and double-check there aren't any test fails

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -100,22 +89,13 @@ public abstract partial class SharedTransformSystem
...
-
     public bool AnchorEntity(EntityUid uid, TransformComponent xform)
     {
         return AnchorEntity((uid, xform));
     }
 
     public bool AnchorEntity(Entity<TransformComponent> entity, Entity<MapGridComponent>? grid = null)
     {
-        DebugTools.Assert(grid == null || grid.Value.Owner == entity.Comp.GridUid);
-
```
metalgearsloth, line 118: This assert shouldn't be removed.

The problem is when an entity is not parented to a grid but is trying to be anchored anyway.

metalgearsloth (changes requested):

PR #5217, opened 2024-06-08
Try optimize `NetEntities` console completion helper

ElectroJr:
Might help alleviate the lag seen in space-wizards/space-station-14/issues/28258, though this PR also outright removes the non-player completions for the tpto command as the helper will still typically iterate over many entities, and is thus still very slow.


metalgearsloth:
At some point we could probably wrap it in a cpujob or smth to at least time-slice it.

PR #5219, opened 2024-06-08
Minor API niceties

metalgearsloth:


Robust.Shared/Player/ActorSystem.cs
```diff
@@ -20,4 +20,14 @@ private void OnActorShutdown(EntityUid entity, ActorComponent component, Compone
     {
         _playerManager.SetAttachedEntity(component.PlayerSession, null);
     }
+
+    public ICommonSession? GetSession(EntityUid? uid)
```
Tayrtahn, line 40: Maybe also a TryGetSession helper?

PR #5223, opened 2024-06-09
Add support for overlays to draw controls

ElectroJr:
Nice feature to have, e.g. for richtext support for something like space-wizards/space-station-14/pull/28705

Overview of changes:
- Added `IUserInterfaceManager.RenderControl()` for manually drawing controls.
  - This is basically just exposing the `IUserInterfaceManagerInternal` functions.
  - Instead of having a bunch of optional arguments, I chose to use the existing public `ControlRenderArguments` struct, under the assumption that it is not meant to be internal/obsoleted/removed.
- The `OverlayDrawArgs` struct now provides an `IRenderHandle` and has an internal constructor.
  - This is currently required for control rendering. However, I could instead have changed control rendering to take in a screen handle instead.
  - I don't really understand the need for the separation of the screen/world handle, and I'd rather just allow people to use IRenderHandle which provides both, even if it is currently an unstable API that is subject to change.
- The `IClydeViewport` overlay rendering methods now take in an `IRenderHandle` instead of a world/screen handle. Required for the viewports to draw overlays with the new `OverlayDrawArgs`.
  - Requires space-wizards/space-station-14/pull/28786
- Controls can now be manually restyled via `Control.InvalidateStyleSheet()` and `Control.DoStyleUpdate()`
  - `InvalidateStyleSheet()` is just the old internal `StyleSheetUpdate()` method, but was renamed for consistency with the measure & arrange invalidation methods.


Robust.Client/UserInterface/UserInterfaceManager.cs
```diff
@@ -403,10 +427,11 @@ public void FrameUpdate(FrameEventArgs args)
 
             control.PreRenderChildren(ref args);
 
-            foreach (var child in control.Children)
+            for (var index = 0; index < control.ChildCount; index++)
             {
+                var child = control.GetChild(index);
                 var pos = position + (Vector2i)Vector2.Transform(child.PixelPosition, coordinateTransform);
-                control.RenderChildOverride(ref args, child.GetPositionInParent(), pos);
+                control.RenderChildOverride(ref args, index, pos);
```
ElectroJr, line 440: These changes are unrelated to the main PR, I just noticed that control rendering was doing a bunch of unnecessary `List.IndexOf` calls.

Conflee:
PR's like this desperately need an example for how to use them. I've read the code like 10 times but I am not super C# literate so how to actually use this feature is a total mystery. Documentation please I beg!

ElectroJr:
The feature was added to help with stuff like the linked map-text content PR, so that it could be updated to use rich text, but I don't think that ever got implemented and it never got used used anywhere else so there aren't really any great examples. And using it is somewhat janky due to having to manually intialize/measure the control's dimensions. FWIW here is a simple demo overlay that uses it to draw a rich text label near the mouse. You can hopefully figure out how to modify that to draw whatever control you want. Though as its hasn't really been used I wouldn't be surprised if you somehow run into unexpected issues.

```cs
public sealed partial class DemoOverlay : Overlay
{
    [Dependency] private IUserInterfaceManager _userInterfaceManager = default!;
    [Dependency] private IInputManager _inputManager = default!;

    private RichTextLabel _label;

    public override OverlaySpace Space => OverlaySpace.ScreenSpace;

    public DemoOverlay()
    {

        // Create the control we want to draw and manually do style & measure updates. Usually this would be done as
        // part of the normal ui/control management and rendering, but has to be invoked manually here whenever the
        // control changes as it is only getting drawn manually in our overlay.
        _label = new RichTextLabel();
        _label.SetMarkup("AAA");

        // required to load fonts & other UI stylesheet stuff.
        _label.InvalidateStyleSheet();
        _label.ForceRunStyleUpdate();

        // I think measure & arrange start invalid anyway, so this might not actually be needed?
        _label.InvalidateMeasure();
        _label.InvalidateArrange();

        // Calculate the control's desired size given the amount of space available (in this case, infinite space)
        _label.Measure(Vector2.PositiveInfinity);
        var size = _label.DesiredSize;

        // Tell the control to actually arrange itself for a given available size (i.e., position itself and its children for rendering)
        _label.Arrange(UIBox2.FromDimensions(Vector2.Zero, size));
    }

    protected override void Draw(in OverlayDrawArgs args)
    {
        // Draw the control next to the mouse. Note that this is for screen-space overlays, if you
        // wanted to use a world-space overlay, you would need to convert use world instead of screen coordinates.
        var mousePosition = (Vector2i)_inputManager.MouseScreenPosition.Position;
        _userInterfaceManager.RenderControl(args.RenderHandle, _label, mousePosition);
    }
}
```

PR #5224, opened 2024-06-10
Added multi prototype loading

TornadoTechnology:
## About PR
The server and client prototype manager now accept the directory list when loading standard prototypes

## Why
It's weird that we only have 1 single folder that we can sub prototypes from by default

### DIFF
--- a/Robust.Client/GameControllerOptions.cs
+++ b/Robust.Client/GameControllerOptions.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using Robust.Shared;
 using Robust.Shared.Utility;
 
@@ -54,9 +55,13 @@ namespace Robust.Client
         public ResPath AssemblyDirectory { get; init; } = new(@"/Assemblies/");
 
         /// <summary>
-        ///     Directory to load all prototypes from.
+        ///     Directories to load all prototypes from.
         /// </summary>
-        public ResPath PrototypeDirectory { get; init; } = new(@"/Prototypes/");
+        public HashSet<ResPath> PrototypeDirectories { get; init; } = new()
+        {
+            new("/EnginePrototypes/"),
+            new("/Prototypes")
+        };
 
         /// <summary>
         /// Directory resource path containing window icons to load.
--- a/Robust.Client/Prototypes/ClientPrototypeManager.cs
+++ b/Robust.Client/Prototypes/ClientPrototypeManager.cs
@@ -44,8 +44,11 @@ namespace Robust.Client.Prototypes
 
         public override void LoadDefaultPrototypes(Dictionary<Type, HashSet<string>>? changed = null)
         {
-            LoadDirectory(new("/EnginePrototypes/"), changed: changed);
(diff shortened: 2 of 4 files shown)


PJB3005:
What is the use case for this? Why do you need an extra prototypes directory in content?

deltanedas:
would be *mildly* useful downstream to have say /Resources/DeltaVPrototypes instead of /Resources/Prototypes/DeltaV but then its inconsistent with textures etc so idk

PJB3005:
That really doesn't seem useful enough to warrant the extra complexity this introduces IMO.

TornadoTechnology:
> That really doesn't seem useful enough to warrant the extra complexity this introduces IMO.

Now all prototypes Shared we do not have to make a separate folder with prototypes visible only to the server or client, and also do not add them to the build. Perhaps it will also be convenient to make folders for forks, with completely their own prototypes

PJB3005:
Slight problem with this implementation then at least: content can't actually add more prototype paths with this PR.

TornadoTechnology:
> Slight problem with this implementation then at least: content can't actually add more prototype paths with this PR.

Why?
```cs
using Content.Server._Harmony;
using Robust.Server;
using Robust.Shared.Utility;

namespace Content.Server
{
    internal static class Program
    {
        public static void Main(string[] args)
        {
            Patch.Load();

            var options = new ServerOptions();
            options.PrototypeDirectories.Add(new ResPath("/MyPrototypeDir"));
            
            ContentStart.StartLibrary(args, options);
        }
    }
}
```

deltanedas:
bruh

TornadoTechnology:
xd

PJB3005:
> > Slight problem with this implementation then at least: content can't actually add more prototype paths with this PR.
> 
> Why?

This is not a supported distribution method of the game for SS14 scenarios. If you actually want to be able to use this for an SS14 server, you cannot use this.

TornadoTechnology:
> This is not a supported distribution method of the game for SS14 scenarios. If you actually want to be able to use this for an SS14 server, you cannot use this.

It's not for ss14, it's for OTHER games on this cool engine

Robust.Client/GameControllerOptions.cs
```diff
@@ -54,9 +55,13 @@ public sealed class GameControllerOptions
         public ResPath AssemblyDirectory { get; init; } = new(@"/Assemblies/");
 
         /// <summary>
-        ///     Directory to load all prototypes from.
+        ///     Directories to load all prototypes from.
         /// </summary>
-        public ResPath PrototypeDirectory { get; init; } = new(@"/Prototypes/");
+        public HashSet<ResPath> PrototypeDirectories { get; init; } = new()
+        {
+            new("/EnginePrototypes/"),
```
PJB3005, line 62: Do not include `EnginePrototypes` in this list. These need to remain engine internal so content has no business touching it.

PJB3005 (changes requested):

metalgearsloth:
ping if you come back to it.

PR #5227, opened 2024-06-11
Added separate localization & clean up

TornadoTechnology:
## About PR
Now localization is separate for server and client!

## Media
https://github.com/space-wizards/RobustToolbox/assets/54727692/eecfd2b0-07b8-4178-ba8d-ad3a6b5204a4

## Why
Maybe in the context of SS14 it might be, a bit odd, but in the context of a game engine it's a basic function

## How it works
All the magic happens thanks to the ``LocalizationManager.LoadDefault`` method, it gets the current localization for the server or client from `CVars`, and loads it. Also for ``LocalizationManager.AddFunction`` to work, it returns the loaded CultureInfo.


Robust.Shared/Localization/LocalizationManager.cs
```diff
@@ -337,8 +352,27 @@ public void ReloadLocalizations()
...
 
+        public void SetCulture(CultureInfo culture)
+        {
+            if (HasCulture(culture))
+            {
+                DefaultCulture = culture;
+                return;
+            }
+
+            LoadCulture(culture);
```
ElectroJr: `LoadCulture` only sets the default culture if it is currently null, so shouldn't this method still set it after loading it?

ElectroJr:
> Now localization is separate for server and client!

> Maybe in the context of SS14 it might be, a bit odd, but in the context of a game engine it's a basic function

While this is a necessary part of a full solution and will help with client-side text, note that it doesn't fully fix the client-side localization issue. This is because a lot of the in game text gets resolved by the server into a string using the server's locale, and then that string gets sent to clients. So a lot of the non-UI text that the client will see will still be in another language. A full fix probably requires replacing all of those strings with some kind of struct with the Loc string id & arguments. 

Also, fully supporting client side languages still requires the server to provide the resources for the language, and at least for ss14 I'm not sure how any forks will bother actively maintaining multiple languages.

Robust.Shared/CVars.cs
```diff
@@ -1753,5 +1753,12 @@ protected CVars()
         /// </summary>
         public static readonly CVarDef<bool> LaunchContentBundle =
             CVarDef.Create("launch.content_bundle", false, CVar.CLIENTONLY);
+
+        /*
+         * Localization
+         */
+
+        public static readonly CVarDef<string> CultureName =
```
PJB3005: This CVar is named incorrectly, it should be `LocCultureName`.

PJB3005 (changes requested):
Remove useless formatting changes, they make review harder.

Robust.Shared/Console/Commands/LocalizationCommands.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace Robust.Shared.Console.Commands;
+
+[UsedImplicitly]
+internal sealed class LocalizationSetCulture : LocalizedCommands
+{
+    private const string Name = "localization_set_culture";
+    private const int ArgumentCount = 1;
+
+    public override string Command => Name;
+    public override string Description => "Set DefaultCulture for the client LocalizationManager";
```
PJB3005: Why does this say client? It's defined in shared.

Also this isn't how `LocalizedCommands` works. Ironic for this kind of PR.

Robust.Shared/Console/Commands/LocalizationCommands.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    public override void Execute(IConsoleShell shell, string argStr, string[] args)
+    {
+        if (args.Length != ArgumentCount)
+        {
+            shell.WriteLine(Loc.GetString($"cmd-invalid-arg-number-error"));
+            return;
+        }
+
+        var localization = IoCManager.Resolve<ILocalizationManager>();
```
PJB3005: This needs to be injected via `[Dependency]`.

Robust.Shared/Console/Commands/LocalizationCommands.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        if (localization.HasCulture(culture))
+        {
+            localization.DefaultCulture = culture;
+            shell.WriteLine($"Loaded culture {culture.Name} set as default");
+            return;
+        }
+
+        shell.WriteLine($"Loaded culture {culture.Name} not found, loading is in progress");
```
PJB3005: You really do not need to log this. Hell on the client this will not be displayed before the next message is.

Robust.Shared/CVars.cs
```diff
@@ -1753,5 +1753,12 @@ protected CVars()
...
+            CVarDef.Create("loc.culture_name", "en-US", CVar.ARCHIVE);
```
ElectroJr: This should  have a subscription that actually changes the default value when the cvar gets set, and the command should probably just change the default locale by setting the cvar. It should probably also be replicated, so the client can know what value the server is using.

Robust.Shared/Console/Commands/LocalizationCommands.cs
```diff
@@ -0,0 +1,47 @@
...
+    public override void Execute(IConsoleShell shell, string argStr, string[] args)
+    {
+        if (args.Length != ArgumentCount)
+        {
+            shell.WriteLine(Loc.GetString($"cmd-invalid-arg-number-error"));
+            return;
+        }
+
+        var localization = IoCManager.Resolve<ILocalizationManager>();
+        var culture = new CultureInfo(args[0]);
```
PJB3005: This needs to catch errors if the user provides an invalid language.

Robust.Shared/Localization/LocalizationManager.cs
```diff
@@ -41,6 +44,18 @@ void IPostInjectInit.PostInject()
             _prototype.PrototypesReloaded += OnPrototypesReloaded;
         }
 
+        public CultureInfo LoadDefault()
+        {
+            var name = _configuration.GetCVar(CVars.CultureName);
+            var culture = new CultureInfo(name);
```
PJB3005: This needs to handle an invalid culture being specified in the config.

Robust.Shared/CVars.cs
ElectroJr: Unless I'm misremembering how server->client replicated cvars work. I know for client->server replicated values the server can differentiate between its own value, and client-side values 

If that is how it works, then the content PR should probably also ensure that the server locale is used as a fallback

Robust.Shared/Console/Commands/LocalizationCommands.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        if (localization.HasCulture(culture))
+        {
+            localization.DefaultCulture = culture;
+            shell.WriteLine($"Loaded culture {culture.Name} set as default");
+            return;
+        }
+
+        shell.WriteLine($"Loaded culture {culture.Name} not found, loading is in progress");
+        localization.LoadCulture(culture); // It's also to set culture as default
```
PJB3005: Uh, `LoadCulture` explicitly does *not* set it as default if something else already is.

PJB3005 (changes requested):

PJB3005:
While this PR does add some fairly innocent convenience methods to `LocalizationManager`, this isn't going to address any of the *actual* issues blocking different languages between client and server, which Electro already covered.

This can be fine for small games on RT, I guess, but for SS14-based games this isn't helping anything.

TornadoTechnology:
> While this PR does add some fairly innocent convenience methods to `LocalizationManager`, this isn't going to address any of the _actual_ issues blocking different languages between client and server, which Electro already covered.
> 
> This can be fine for small games on RT, I guess, but for SS14-based games this isn't helping anything.

It's PR for small games, pls do something!!!!

Robust.Shared/Utility/CultureInfoExtension.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Globalization;
+
+namespace Robust.Shared.Utility;
+
+public static class CultureInfoExtension
+{
+    public static bool Equals(this CultureInfo cultureInfo, CultureInfo otherCultureInfo)
+    {
+        return cultureInfo.Name == otherCultureInfo.Name;
+    }
```
PJB3005: `CultureInfo`s are more complicated than this. Overriding a base `Equals` like this is not a correct "equals" check. This function should be called `NameEquals` or something at the very least.

Robust.Shared/Localization/Loc.cs
PJB3005, line 1: 1. Do not do large file formatting changes like this.
2. The changes to this file (to add a static variable) are absolutely not good. If you want to avoid the overhead, get a dependency to `ILocalizationManager`. Do *not* break global state like this.

Robust.Client/Console/Commands/LocalizationCommands.cs
```diff
@@ -0,0 +1,47 @@
...
+internal sealed class LocalizationSetCulture : LocalizedCommands
+{
+    [Dependency] private readonly ILocalizationManager _localization = default!;
+
+    private const string Name = "localization_set_culture";
+    private const int ArgumentCount = 1;
+
+    public override string Command => Name;
+    public override string Description => "Set DefaultCulture for the client LocalizationManager";
+    public override string Help => $"{Name} <cultureName>";
```
PJB3005: These localizations still aren't fixed.

Robust.Client/Console/Commands/LocalizationCommands.cs
```diff
@@ -0,0 +1,47 @@
...
+using Robust.Shared.IoC;
+using Robust.Shared.Localization;
+using Robust.Shared.Utility;
+
+namespace Robust.Client.Console.Commands;
+
+[UsedImplicitly]
+internal sealed class LocalizationSetCulture : LocalizedCommands
+{
+    [Dependency] private readonly ILocalizationManager _localization = default!;
```
PJB3005: You already have access to this via the `LocalizedCommands` field.

Robust.Shared/Utility/CultureInfoHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+
+public static class CultureInfoHelper
+{
+    public static readonly FrozenSet<string> CultureNames = GenerateAllCultureNames();
+
+    /// <exception cref="InvalidCastException"></exception>
+    public static CultureInfo GetCultureInfo(string code)
+    {
+        if (!HasCulture(code))
+            throw new InvalidCastException();
```
PJB3005: This is not the correct exception type to use here. Just remove the check, let the `CultureInfo` constructor throw.

Robust.Shared/Utility/CultureInfoHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Linq;
+
+namespace Robust.Shared.Utility;
+
+public static class CultureInfoHelper
+{
+    public static readonly FrozenSet<string> CultureNames = GenerateAllCultureNames();
```
PJB3005: I am not sure what use it is to cache this large set in the engine like this. This only really feels barely useful for the completions of the command you added.

Robust.Shared/Utility/CultureInfoHelper.cs
PJB3005, line 1: I am unsure about the entire purpose of this file. I would expect it would be best that we used `CultureInfo.GetCultureInfo()` with the option to accept non-existent cultures, for robustness' sake. The game's ability to load localization files on disk shouldn't really be tied to "does the operating system think this language exists."

Robust.Shared/Localization/LocalizationManager.cs
```diff
@@ -353,6 +393,20 @@ public void LoadCulture(CultureInfo culture)
             DefaultCulture ??= culture;
         }
 
+        public List<CultureInfo> GetFoundCultures()
+        {
+            var result = new List<CultureInfo>();
+            foreach (var name in _res.ContentGetDirectoryEntries(new ResPath("/Locale")))
```
PJB3005: Don't hardcode this path.

Robust.Shared/Localization/LocalizationManager.cs
```diff
@@ -337,8 +358,27 @@ public CultureInfo? DefaultCulture
...
+
+        public bool HasCulture(CultureInfo culture)
+        {
+            return _contexts.ContainsKey(culture);
+        }
+
         public void LoadCulture(CultureInfo culture)
         {
+            // Attempting to load an already loaded culture
+            Debug.Assert(!HasCulture(culture));
```
PJB3005: This is a public API, so this should not be an assert that's compiled out on debug.

PJB3005 (dismissed):
I'll fix these issues myself.

Robust.Shared/CVars.cs
PJB3005: The entire point of this API is that it's for non-client-server scenarios right now, so it doesn't (and in fact, can't) be synchronized.

Robust.Shared/Localization/LocalizationManager.cs
```diff
@@ -41,6 +44,24 @@ void IPostInjectInit.PostInject()
...
+        public CultureInfo SetDefaultCulture()
+        {
+            var defaultCode = _configuration.GetCVar(CVars.LocDefaultCultureName);
+            var code = _configuration.GetCVar(CVars.LocCultureName);
+
+            if (!CultureInfoHelper.TryGetCultureInfo(code, out var culture))
+            {
+                // He'll throw out the exception if he can't find a default culture
+                CultureInfoHelper.GetCultureInfo(defaultCode, out culture);
+            }
```
PJB3005: This isn't guarding against "we don't have localization files for this culture", it's guarding against "the OS doesn't know about this culture". This feels like a weird failure mode to account for, when there's a pretty obvious fallback mechanism that should work well enough: just use a non-predefined culture.

TornadoTechnology:
Wow

PR #5228, opened 2024-06-11
Make PvsSystem consider offset and zoom from EyeComponent

geraeumig:
Implemented changes so that `PvsSystem` takes into account `EyeComponent`'s `Offset` and `Zoom` properties when calculating which chunks and entities should be sent to the client.


Robust.Server/GameStates/PvsSystem.Session.cs
```diff
@@ -125,14 +125,17 @@ private void UpdateSession(PvsSession session)
         // Update visibility masks & viewer positions
         // TODO PVS do this before sending state.
         // I,e, we already enumerate over all eyes when computing visible chunks.
-        Span<MapCoordinates> positions = stackalloc MapCoordinates[session.Viewers.Length];
+        Span<(MapCoordinates pos, Vector2? zoom)> positions = stackalloc (MapCoordinates, Vector2?)[session.Viewers.Length];
         int i = 0;
```
metalgearsloth: Isn't it better to just get a Box2 back instead if we're doing this.

geraeumig: Would assume that Box2 is mostly used to represent actual rectangles, somehow using it makes the Chebyshev distance division feel even more awkward.
It would have to remain a span of tuples, as `MapId` still has to be stored in it.

ElectroJr:
Some of the scaling stuff is what `EyeComponent.PvsSize` was meant to be used for, but its kind of flawed and wasn't fully implemented. Do you mind if I push some changes to your branch to just improve that? and then you should just be able to set something like `eye,PvsScale = Math.Max(eye.Zoom.X, eye.Zoom.y)` somewhere in content if you want to change the pvs range.

I wanted to keep eye zoom & eye pvs range separate to avoid uneccesarily changing it & maybe causing light pop in if a player just temporarily zooms in their view for whatever reason.

geraeumig:
Sure, go ahead, would the entity draw range be also influenced by the PvsSize?

ElectroJr:
I changed it to just use a pvs scale float. In principle you should just need to set it to `Math.Max(zoom.x, zoom.y)` and then it should work basically as it did before my changes.

> Sure, go ahead, would the entity draw range be also influenced by the PvsSize?

Nah that's still just controlled by the eye's zoom.

PR #5231, opened 2024-06-11
Remove CompIdx locks

### LINKED ISSUE #5230: CompIdx thread contention
At least that's what I think this is.

image

Ideally we'd just cache it upfront I guess?

metalgearsloth:
So GetComponentState in PVS calls RaiseComponentEvent which in turn calls this. When you start getting a significant number of players it seems to run into lock contention considering every single compstate get will lock this.

Instead we'll just store a FrozenDictionary on ComponentFactory. This reference already exists everywhere we need it + now we replace both a lock and a dictionary for double-win in single-threaded and multi-threaded scenarios.

Resolves https://github.com/space-wizards/RobustToolbox/issues/5230


metalgearsloth:
I am now seeing tests running into static issues which is urgh.

ElectroJr (approved):

PR #5236, opened 2024-06-14
Avoid read lock in `GetEntityQuery`

ElectroJr:
~~Probably causing #5230? At least AFAICT this is the only place were it `Index(Type t)` gets used somewhere were it might be multithreaded.~~ nvm forgot about get state events. But probably still worthwhile even with #5231


metalgearsloth:
https://github.com/space-wizards/RobustToolbox/pull/5231 also uses a frozendictionary instead too.

PR #5237, opened 2024-06-14
Add attempt event for sessionstate

metalgearsloth:


Robust.Server/Player/PlayerManager.cs
```diff
@@ -130,15 +130,19 @@ private void HandlePlayerListReq(MsgPlayerListReq message)
             session.ConnectedTime = DateTime.UtcNow;
             SetStatus(session, SessionStatus.Connected);
 
-            var list = new List<SessionState>();
+            var list = new List<SessionState>(players.Length);
+            var ev = new GetSessionStateAttempt(session);
+            _entityManager.EventBus.RaiseEvent(EventSource.Local, ref ev);
```
PJB3005, line 135: This isn't really simulation logic so using EventBus here doesn't really make a lot of sense. Just use a regular C# event?

Robust.Server/Player/PlayerManager.cs
```diff
@@ -130,15 +130,19 @@ private void HandlePlayerListReq(MsgPlayerListReq message)
...
+
             foreach (var client in players)
             {
                 var info = new SessionState
                 {
                     UserId = client.UserId,
-                    Name = client.Name,
+                    Name = ev.Cancelled ? string.Empty : client.Name,
```
PJB3005, line 142: Isn't this just the player name? It doesn't seem very useful to try to hide this info when player names are just public from the auth API anyways.

Robust.Server/Player/PlayerManager.cs
```diff
@@ -130,15 +130,19 @@ private void HandlePlayerListReq(MsgPlayerListReq message)
...
                     Status = client.Status,
-                    Ping = client.Channel!.Ping
+                    Ping = client.Channel.Ping,
+                    ControlledEntity = EntManager.GetNetEntity(client.AttachedEntity),
```
PJB3005, line 145: Exposing the controlled entity seems very powerful, what use does this have for players?

PJB3005 (changes requested):

PR #5242, opened 2024-06-15
Reduce uses of `IComponentFactory.GetIndex(Type)`

ElectroJr:
This PR updates various `RaiseComponentEvent` and component lifestage methods to reduce the usage of `IComponentFactory.GetIndex()`. Most notably, it should make component removal and `AfterAutoHandleStateEvent` raising somewhat faster by avoiding the frozen dictionary lookup and using `CompIdx.Index<T>` instead.


Robust.Shared/GameObjects/IEntityManager.Components.cs
```diff
@@ -100,7 +100,7 @@ public partial interface IEntityManager
         /// </summary>
         /// <typeparam name="T">The component reference type to remove.</typeparam>
         /// <param name="uid">Entity UID to modify.</param>
-        bool RemoveComponent<T>(EntityUid uid, MetaDataComponent? meta = null);
```
ElectroJr, line 103: Didn't list this as a breaking change, because any code that was calling thing using something other than a component should've already been failing

PR #5244, opened 2024-06-17
FormattedMessage/DebugConsole performance improvements

PJB3005:
This should fix perf issues, especially ones caused by "we are shitting out tons of errors into the console every tick and there's no cap help".


juliangiebel (approved):
Rich text changes look good to me 🎉

PR #5248, opened 2024-06-18
Make unshaded sprite layers not require shader/texture changes.

ElectroJr:
Currently drawing an unshaded sprite layer requires switching to the "unshaded" shader, which requires changing textures and creating a new render batch, which generally slows things down. It could be changed to set a shader uniform instead, which is faster than changing textures but still somewhat slow.

This PR changes the default shader so it infers whether lighting should be enabled or disabled from a vertex attribute. Instead of adding a new attribute, it re-uses the colour modulation attribute, where negative values now imply that the lightmap shouldn't get sampled. It's kind of janky, but avoids having to send more vertex data. 

This is a revival of a year+ old branch, and when I benchmarked it back then it did help increased the fps I got on my laptop [by a bit](https://discord.com/channels/310555209753690112/900426319433728030/1102466865382568027).

This PR also renames some of the local variables used in the default shader template to try be somewhat more consistent and to help avoid accidental variable name conflicts (which happened with the `modulate` variable while working in this, which is what prompted this change). Requires [a content PR ](https://github.com/space-wizards/space-station-14/pull/29150) to update the outline shader.


PJB3005:
> This PR also renames some of the local variables used in the default shader template to try be somewhat more consistent and to help avoid accidental variable name conflicts (which happened with the modulate variable while working in this, which is what prompted this change). Requires https://github.com/space-wizards/space-station-14/pull/29150to update the outline shader.

Please no breaking changes.

PJB3005 (approved):
Revert the breaking changes but otherwise looks good

Robust.Client/Graphics/Clyde/Shaders/base-default.frag
```diff
@@ -1,21 +1,52 @@
...
+    if (VtxModulate.x < 0.0)
+    {
+        // Negative VtxModulate implies unshaded/no lighting.
+        MODULATE = -1.0 - VtxModulate;
+        LIGHT = vec4(1.0);
+    }
+    else
+    {
+        MODULATE = VtxModulate;
+        LIGHT = texture2D(LIGHTMAP, Pos);
```
PJB3005: I double checked the GLSL spec after not finding info anywhere else.

Sampling a texture inside non-uniform control flow like this is undefined. Either move the texture sample to always happen or use a function like `textureLod()` which doesn't need implicit derivatives.

PJB3005 (changes requested):
Actually, nevermind.

Robust.Client/Graphics/Clyde/Shaders/base-default.frag
ElectroJr: I've just moved it outside of the branch.

ElectroJr:
> Revert the breaking changes but otherwise looks good

I've reverted some of the changes, but there will still be a breaking change due to the fact that `VtxModulate` may now contain negative values and fragment shaders that relied on it should be using the `MODULATE` variable instead.

PJB3005 (approved):

PR #5249, opened 2024-06-18
Add an entry to RSI meta.json for controlling RSI atlas grouping

ElectroJr:
This adds an "atlasGroup" integer to the RSI meta.json that is intended to allow RSIs to try and control how they are inserted into the RSI atlases. Currently this doesn't actually ensure that each "group" has its own atlas, but ideally that's what it would do eventually (rectangle packing when).  Currently its mainly useful for trying to prevent large & niche RSIs from taking up space in the main atlas.

### DIFF
--- a/Schemas/rsi.json
+++ b/Schemas/rsi.json
@@ -50,7 +50,8 @@
                         ]
                     ]
                 }
-            ]
+            ],
+            "atlasGroup": 0
         }
     ],
     "required": [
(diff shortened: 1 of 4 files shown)


Schemas/rsi.json
```diff
@@ -184,6 +185,13 @@
                     }
                 }
             }
+        },
+        "atlasGroup": {
+            "$id": "#/properties/atlasGroup",
+            "default": "",
+            "description": "An integer that is used to group RSIs into the same texture atlas. Defaults to 0. Negative values prevent the RSI from being put into a larger atlas.",
+            "title": "atlasGroup",
+            "type": "integer"
```
ElectroJr, line 194: I'm not really sure how the RSI schema validator works, but hopefully this is right?

PJB3005 (commented):
Not sure I'm a fan of this being controlled with an int, that seems like an organizational mess.

ElectroJr:
> Not sure I'm a fan of this being controlled with an int, that seems like an organizational mess.

Yeah there's probably a better way of doing this. We probably still want something like this, but I'm just gonna close the PR for now because I'm not likely to finish it anytime soon.

PR #5251, opened 2024-06-18
SIMD Colour multiplication

ElectroJr:
This PR changes colour multiplication & interpolation so that they just use `System.Numerics.Vector4`. I ran the content colour interpolation benchmarks, and they are unchanged, while a similar benchmark for multiplication is ~3 times faster.

### DIFF
--- a/Robust.Shared.Maths/Color.cs
+++ b/Robust.Shared.Maths/Color.cs
@@ -1003,7 +1003,7 @@ namespace Robust.Shared.Maths
         /// </summary>
         /// <param name="other">The Color4 structure to compare to.</param>
         /// <returns>True if both Color4 structures contain the same components; false otherwise.</returns>
-        public readonly bool Equals(in Color other)
+        public readonly bool Equals(Color other)
         {
             // TODO COLOR why is this approximate
             // This method literally doesn't do what its docstring says it does.
--- a/Robust.Shared.Maths/Color.cs
+++ b/Robust.Shared.Maths/Color.cs
@@ -46,22 +46,22 @@ namespace Robust.Shared.Maths
     public struct Color : IEquatable<Color>, ISpanFormattable
     {
         /// <summary>
-        ///     The red component of this Color4 structure.
+        ///     The red component of this Color structure.
         /// </summary>
         public float R;
 
         /// <summary>
-        ///     The green component of this Color4 structure.
+        ///     The green component of this Color structure.
         /// </summary>
         public float G;
 
         /// <summary>
-        ///     The blue component of this Color4 structure.
+        ///     The blue component of this Color structure.
         /// </summary>
         public float B;
 
         /// <summary>
-        ///     The alpha component of this Color4 structure.
+        ///     The alpha component of this Color structure.
         /// </summary>
         public float A;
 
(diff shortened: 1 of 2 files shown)


Robust.Shared.Maths/Color.cs
```diff
@@ -309,7 +327,6 @@ public static bool TryFromName(string colorName, out Color color)
         public static Color FromSrgb(Color srgb)
         {
             float r, g, b;
-#if NETCOREAPP
```
ElectroJr, line 312: We use `MathF` all over the place. based on the copyright I'm guessing this is a relic from the Open Toolkit Library?

PJB3005: No this is just from before we were done moving to .NET Core 3.1.

Robust.Shared.Maths/Color.cs
```diff
@@ -35,40 +36,45 @@
...
         ///     The alpha component of this Color4 structure.
         /// </summary>
+        [FieldOffset(3*sizeof(float))]
         public float A;
 
+        /// <summary>
+        /// Vector representation, for easy SIMD operations.
+        /// </summary>
+        [FieldOffset(0), NonSerialized]
+        public SysVector4 RGBA;
```
PJB3005: Don't do this, use `Unsafe.BitCast<>()` instead.

PJB3005: The JIT doesn't like `FieldOffset` shenanigans if it can be avoided, IIRC.

ElectroJr: Changed it to use bitcast. 

I know constructors are inefficient with overlapping data unless you use `Unsafe.SkipInit`, but what are the other reason the JIT doesn't like it? I can't seem to find info about it.

PJB3005: Makes it harder for the JIT to do analysis on how the fields are accessed which can have all kinds of effects like enregistration and all that, AFAIK.

Robust.Shared.Maths/Color.cs
```diff
@@ -90,14 +92,30 @@ public Color(float r, float g, float b, float a = 1)
...
+        {
+            this = Unsafe.BitCast<SysVector4, Color>(vec);
+        }
+
+        /// <summary>
+        ///     Constructs a new Color structure by coping the contents of another color struct.
+        /// </summary>
+        public Color(in Color color) : this(color.RGBA)
+        {
+        }
```
PJB3005: This constructor seems useless?

Robust.Shared.Maths/Color.cs
```diff
@@ -124,32 +142,32 @@ public Color(byte r, byte g, byte b, byte a = 255)
...
 
         /// <summary>
-        ///     Compares the specified Color4 structures for equality.
+        ///     Compares the specified Color structures for equality.
         /// </summary>
         /// <param name="left">The left-hand side of the comparison.</param>
         /// <param name="right">The right-hand side of the comparison.</param>
         /// <returns>True if left is equal to right; false otherwise.</returns>
-        public static bool operator ==(Color left, Color right)
+        public static bool operator ==(in Color left, in Color right)
```
PJB3005: These are all ABI breaking changes. Do they actually improve performance?

ElectroJr: This was leftover from when I tried to add `in` to the `MathHelper.CloseToPercent()` equality checks in case it was somehow causing defensive copies, but accidentally left these behind after I removed those. IIRC it made no difference.

PR #5258, opened 2024-06-20
Entity<T> casts to Entity<T?> automatically

deltanedas:
with 1-4 components any more you are sus as hell and im lazy

doesnt work because shit language but i want to know if theres a way to do it

use case is say function `void Func(Entity<TransformComponent?> ent)` and you have `Entity<TransformComponent>` before this you had to do `(ent, ent.Comp)` which is shit, now you just pass ent and it automatically gets converted (if it worked :trollface:)


Tayrtahn:
Sure would be nice. But yeah, I don't think this is possible.

deltanedas:
shouldve used rust smh my head

PJB3005:
Re-open this PR when it uh, actually works?

PR #5262, opened 2024-06-23
Add LocalTilesIntersecting for circles

metalgearsloth:
Title, found a bug in dungeon code.

Not an efficient way to do it by any means but it's similar to the current one for world-coords.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
@@ -922,7 +922,7 @@ public abstract partial class SharedMapSystem
             var tileCenter = tile.GridIndices + grid.TileSizeHalfVector;
             var direction = tileCenter - localCircle.Position;
 
-            if (direction.LengthSquared() <= localCircle.Radius * localCircle.Radius)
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
```diff
@@ -909,6 +909,26 @@ public void SetTiles(EntityUid uid, MapGridComponent grid, List<(Vector2i GridIn
...
+            localCircle.Position.X + localCircle.Radius, localCircle.Position.Y + localCircle.Radius);
+
+        var tileEnumerator = GetLocalTilesEnumerator(uid, grid, aabb, ignoreEmpty, predicate);
+
+        while (tileEnumerator.MoveNext(out var tile))
+        {
+            var tileCenter = tile.GridIndices + grid.TileSizeHalfVector;
+            var direction = tileCenter - localCircle.Position;
+
+            if (direction.LengthSquared() <= localCircle.Radius * localCircle.Radius)
```
Tayrtahn:
```suggestion
            if (direction.IsShorterThanOrEqualTo(localCircle.Radius))
```
For improved readability?

Maybe I'm just trying to justify the existence of these helper methods 🙃

PR #5263, opened 2024-06-23
offgrid friction in island solver

EmoGarbage404:
part 2 of https://github.com/space-wizards/space-station-14/pull/29383

adds support for friction to be modified when off grid.
this is primarily to help facilitate different gameplay levels when working off-grid vs while working on a weightless grid.

pretty shrimple pr.

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Island.cs
+++ b/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Island.cs
@@ -196,6 +196,19 @@ public abstract partial class SharedPhysicsSystem
     private const int VelocityConstraintsPerThread = 16;
     private const int PositionConstraintsPerThread = 16;
 
+    [ByRefEvent]
+    public struct GetLinearDampingOverrideEvent
+    {
+        public float? LinearDampingOverride;
+    }
+
+    public delegate void LinearDampingOverrideHandler(ref GetLinearDampingOverrideEvent ev);
+
+    /// <summary>
+    ///     Invoked to allow content to override linear damping for objects.
+    /// </summary>
+    public event LinearDampingOverrideHandler? OnGetLinearDampingOverride;
+
     #region Setup
 
     private void InitializeIsland()
@@ -717,7 +730,6 @@ public abstract partial class SharedPhysicsSystem
         var angles = ArrayPool<float>.Shared.Rent(bodyCount);
         var offset = island.Offset;
         var xformQuery = GetEntityQuery<TransformComponent>();
-        var mapGridQuery = GetEntityQuery<MapGridComponent>();
 
         for (var i = 0; i < island.Bodies.Count; i++)
         {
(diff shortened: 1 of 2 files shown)


Robust.Shared/Physics/Systems/SharedPhysicsSystem.Island.cs
```diff
@@ -744,7 +747,12 @@ private bool InternalParallel(IslandData island)
 
                 angularVelocity += body.InvI * body.Torque * data.FrameTime;
 
-                linearVelocity *= Math.Clamp(1.0f - data.FrameTime * body.LinearDamping, 0.0f, 1.0f);
+                var onGrid = xform.GridUid != null || mapGridQuery.HasComp(xform.MapUid);
+                var linearDamping = onGrid
+                    ? body.LinearDamping
+                    : body.OffGridLinearDamping;
+
+                linearVelocity *= Math.Clamp(1.0f - data.FrameTime * linearDamping, 0.0f, 1.0f);
```
metalgearsloth: Few things:

1. If an entity is on a map that is also a grid then GridUid should be getting set still so the hascomp should be unnecessary.
2. This should be a C# event raised with Entity\<TransformComponent, PhysicsComponent\> to get damping override as it's very ss14 specific.

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Island.cs
```diff
@@ -744,7 +759,11 @@ private bool InternalParallel(IslandData island)
 
                 angularVelocity += body.InvI * body.Torque * data.FrameTime;
 
-                linearVelocity *= Math.Clamp(1.0f - data.FrameTime * body.LinearDamping, 0.0f, 1.0f);
+                var ev = new GetLinearDampingOverrideEvent();
```
metalgearsloth: Don't you still need the entity itself?

Also don't add another struct for it just pass the entity directly to the delegate.

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Island.cs
```diff
@@ -744,7 +752,12 @@ private bool InternalParallel(IslandData island)
 
                 angularVelocity += body.InvI * body.Torque * data.FrameTime;
 
-                linearVelocity *= Math.Clamp(1.0f - data.FrameTime * body.LinearDamping, 0.0f, 1.0f);
+                float? linearDampingOverride = null;
+                var entity = new Entity<PhysicsComponent, TransformComponent>(body.Owner, body, xform);
+                OnGetLinearDampingOverride?.Invoke(ref entity, ref linearDampingOverride);
+                var linearDamping = linearDampingOverride ?? body.LinearDamping;
+
+                linearVelocity *= Math.Clamp(1.0f - data.FrameTime * linearDamping, 0.0f, 1.0f);
```
metalgearsloth, line 760: Couple more questions:
1. Do you have the content code that will be using this PRd? As it looks like the linked content PR doesn't use it.
2. Does just subscribing to entity parent changes and using the body's lineardamping work? As this will also severely impact servers like RMC14 with the overhead.

EmoGarbage404: 1. the linked content PR uses the `OnGetLinearDampingOverride`.
2. i don't follow what you mean with the suggestion of using the parent.

PR #5265, opened 2024-06-24
Add loop support for SpriteSystem.GetFrame

metalgearsloth:
For Ftl I just want it played once.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
@@ -184,7 +184,8 @@ namespace Robust.Client.GameObjects
         /// <summary>
         /// Gets the specified frame for this sprite at the specified time.
         /// </summary>
-        public Texture GetFrame(SpriteSpecifier spriteSpec, TimeSpan curTime)
+        /// <param name="loop">Should we clamp on the last frame and not loop</param>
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
```diff
@@ -184,7 +184,8 @@ public void ForceUpdate(EntityUid uid)
         /// <summary>
         /// Gets the specified frame for this sprite at the specified time.
         /// </summary>
-        public Texture GetFrame(SpriteSpecifier spriteSpec, TimeSpan curTime)
+        /// <param name="loop">Should we clamp on the last frame and not loop</param>
+        public Texture GetFrame(SpriteSpecifier spriteSpec, TimeSpan curTime, bool loop = true)
```
PJB3005, line 188: Adding an optional arg like this is an ABI breaking change so if we want to avoid that you'd need to make an overload.

PJB3005 (commented):
Seems good otherwise

Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
metalgearsloth: If the default behavior weren't loopable I'd probably have `GetFrame` and `GetLoopedFrame` rather than `GetUnloopedFrame` or something which is also a breaking change so this might be the lesser of two evils.

PR #5269, opened 2024-06-25
Cache HasCustomAttribute

### LINKED ISSUE #5268: HasCustomAttribute is really expensive in debug
I know it's debug but when it's more expensive than initializing an entity there's somewhat of a problem.

metalgearsloth:
Makes debug a lot faster. No idea if static semaphoreslims are bad but my game didn't explodeTM.

Resolves https://github.com/space-wizards/RobustToolbox/issues/5268

image

image

### DIFF
--- a/Robust.Shared/Utility/TypeHelpers.cs
+++ b/Robust.Shared/Utility/TypeHelpers.cs
@@ -4,11 +4,16 @@ using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Threading;
 
 namespace Robust.Shared.Utility
 {
     public static class TypeHelpers
     {
+        private static SemaphoreSlim _hasSlim = new(1);
+        private static Dictionary<MemberInfo, Dictionary<Type, bool>> _hasAttribute = new();
+
         /// <summary>
         ///     Returns absolutely all fields, privates, readonlies, and ones from parents.
         /// </summary>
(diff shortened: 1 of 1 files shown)


Robust.Shared/Utility/TypeHelpers.cs
```diff
@@ -4,11 +4,16 @@
...
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Threading;
 
 namespace Robust.Shared.Utility
 {
     public static class TypeHelpers
     {
+        private static SemaphoreSlim _hasSlim = new(1);
+        private static Dictionary<MemberInfo, Dictionary<Type, bool>> _hasAttribute = new();
```
PJB3005, line 15:
```suggestion
        private static readonly Dictionary<MemberInfo, Dictionary<Type, bool>> _hasAttribute = new();
```

Robust.Shared/Utility/TypeHelpers.cs
```diff
@@ -183,7 +188,18 @@ public static bool IsBackingField(this MemberInfo memberInfo)
 
         public static bool HasCustomAttribute<T>(this MemberInfo memberInfo) where T : Attribute
         {
-            return memberInfo.GetCustomAttribute<T>() != null;
+            using (_hasSlim.WaitGuard())
```
PJB3005, line 191: Just use `lock()` on the dictionary instead.

PJB3005 (commented):
It would be way better if *usages* of this function were cached instead, but I guess I'll allow it...

metalgearsloth:
> It would be way better if _usages_ of this function were cached instead, but I guess I'll allow it...

If you want me to do it I can do it, it's primarily dirty and addcomp calls afaik.

PR #5277, opened 2024-06-30
Replaces the entity spawn window's bespoke method of object icon rendering with entityprototypeview

deathride58:
Title.

Basically, this allows the entity spawn window to have (theoretical) rendering parity with the actual game world

Examples!
dotnet_wgJV546bQw

dotnet_etmjJn1025

Very straight-forward PR, overall

### DIFF
--- a/Robust.Client/UserInterface/Controllers/Implementations/EntitySpawningUIController.cs
+++ b/Robust.Client/UserInterface/Controllers/Implementations/EntitySpawningUIController.cs
@@ -327,8 +327,7 @@ public sealed class EntitySpawningUIController : UIController
         if (_window == null || _window.Disposed)
             return;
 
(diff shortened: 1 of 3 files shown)


Willhelm53:
Seems like this will fix [c#26972](https://github.com/space-wizards/space-station-14/issues/26972) and probably a few others! Or at least it's closely related.

deathride58:
This will not fix 26972, as vendors use their own separate bespoke implementation for icon rendering (just as the entity spawn window does). The proper fix for it will be to make the vendor UI (which is within content) properly utilize `EntityPrototypeView`, but that in itself would require a fairly thorough gutting of the vendor UI due to how `ItemList` works (the vendor UI would need to be rewritten to instead use either `ListContainer` or the content-side `SearchListContainer`, as `ItemList` does not support containing arbitrary controls)

PR #5281, opened 2024-07-03
Add TryQueueDeleteEntity

metalgearsloth:
Having to check deleted by hand up front is annoying.

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.cs
+++ b/Robust.Shared/GameObjects/EntityManager.cs
@@ -468,6 +468,21 @@ namespace Robust.Shared.GameObjects
             ent.Comp4.LastModifiedTick = CurrentTick;
         }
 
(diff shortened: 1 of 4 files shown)


Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -468,6 +468,21 @@ public virtual void Dirty(EntityUid uid, IComponent component, MetaDataComponent
             ent.Comp4.LastModifiedTick = CurrentTick;
         }
 
+        public bool TryQueueDeleteEntity(EntityUid? uid)
+        {
+            if (uid == null)
```
Tayrtahn, line 473: Should this check for EntityUid.Invalid too?

metalgearsloth: Deleted already checks for it.

PR #5282, opened 2024-07-04
Add cvar to limit nearby command range

DrSmugleaf:
Admins don't necessarily know the implications of a very large range. It caused a server crash in RMC14.
Default 200.
Also I can't do this from content so here it is.

### DIFF
--- a/Robust.Shared/CVars.cs
+++ b/Robust.Shared/CVars.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Runtime.InteropServices;
 using System.Threading;
 using Lidgren.Network;
 using Robust.Shared.Audio;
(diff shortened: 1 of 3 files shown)


Robust.Shared/CVars.cs
```diff
@@ -1759,5 +1758,16 @@ protected CVars()
...
             CVarDef.Create("launch.content_bundle", false, CVar.CLIENTONLY);
+
+        /*
+         * TOOLSHED
+         */
+
+        /// <summary>
+        ///     Whether or not to show a button that opens the guidebook when a player changes their species,
+        ///     explaining the difference between each.
+        /// </summary>
```
metalgearsloth: que

DrSmugleaf: Copy paste / patch paste blunder

DrSmugleaf:
```suggestion
        /// <summary>
        ///     The max range that can be passed to the nearby toolshed command.
        ///     Any higher value will cause an exception.
        /// </summary>
```

PR #5286, opened 2024-07-06
Adds event for mass and angular inertia changes.

TemporalOroboros:
Adds `MassDataChangedEvent` which is raised when the mass or angular inertia of an entity changes.
This allows other systems to respond to such changes.


Robust.Shared/Physics/Events/MassChangedEvent.cs
```diff
@@ -0,0 +1,31 @@
...
+public readonly record struct MassDataChangedEvent(
+    Entity<PhysicsComponent, FixturesComponent> Entity,
+    float NewMass,
+    float NewInertia,
+    Vector2 NewCenter,
+    float OldMass,
+    float OldInertia,
+    Vector2 OldCenter
+)
+{
```
metalgearsloth: You can just have the mass + intertia fields be properties that point to those values on the physicscomponent itself.

PR #5288, opened 2024-07-06
Add AppearanceSystem Data dictionaries and RemoveData

SlamBamActionman:
## What?

This PR allows AppearanceSystem to send and store dictionaries in its Data property, using `SerializationManager.CopyTo` as an alternative to ICloneable.*

It also adds `SharedAppearanceSystem.RemoveData`, as there is currently no way to remove a Data entry (only set the entry value to empty).

## Why?

This change makes AppearanceSystem Data more flexible; since Data keys must be enums it becomes difficult to send dynamic data for systems that don't want to have predefined enums (e.g. ClothingSystem). Dictionary support solves this without bystepping the scoping as the dictionary itself must still be keyed to an enum.

There's also no method to remove a Data entry at the moment and instead forces systems to rely on setting an empty value for a key it no longer wants to interact with, when a more elegant solution would be to simply remove the entry. 

- - -
*Note that this also comments out a Debug.Assert for SetData, as there appears to currently not be any way to check if an object can be parsed via CopyTo without actually copying the object.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedAppearanceSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedAppearanceSystem.cs
@@ -43,9 +43,27 @@ public abstract class SharedAppearanceSystem : EntitySystem
         if (component.AppearanceData.TryGetValue(key, out var existing) && existing.Equals(value))
             return;
 
-        DebugTools.Assert(value.GetType().IsValueType || value is ICloneable, "Appearance data values must be cloneable.");
+        //Commented out until there is a suitable way to check that ISerializationManager.CopyTo works without doing the copying
+        //DebugTools.Assert(value.GetType().IsValueType || value is ICloneable, "Appearance data values must be cloneable.");
 
         component.AppearanceData[key] = value;
+        Dirty(uid, component);
+        QueueUpdate(uid, component);
+    }
+
+    public void RemoveData(EntityUid uid, Enum key, AppearanceComponent? component = null)
+    {
+        if (!Resolve(uid, ref component, false))
+            return;
+
+        // If appearance data is changing due to server state application, the server's comp state is getting applied
+        // anyways, so we can skip this.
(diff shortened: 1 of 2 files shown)


PJB3005:
> This change makes AppearanceSystem Data more flexible; since Data keys must be enums it becomes difficult to send dynamic data for systems that don't want to have predefined enums (e.g. ClothingSystem)

I was shocked to see this isn't possible since it was *literally* designed for exactly this use case. So I checked, and #3086 removed it with wrong justification 😐:

> Why:
> - ensures yaml can get linted to check for basic typos
> - Simplifies appearance logic.
> - string keys are hardly used. ~9 vs 283 uses.

SlamBamActionman:
I don't particularly mind that Appearance Data uses enums since it *does* ensure that there is a strictly enforced system to avoid two entries from two different systems having the same name. 

That being said, it sure made working out how to implement it in ClothingSystem more difficult... But at least dictionaries frees it up again while still maintaining enum enforcement.

Robust.Shared/GameObjects/Systems/SharedAppearanceSystem.cs
```diff
@@ -43,13 +43,31 @@ public void SetData(EntityUid uid, Enum key, object value, AppearanceComponent?
...
 
+    public void RemoveData(EntityUid uid, Enum key, AppearanceComponent? component = null)
+    {
+        if (!Resolve(uid, ref component, false))
+            return;
+
+        // If appearance data is changing due to server state application, the server's comp state is getting applied
+        // anyways, so we can skip this.
+        if (_timing.ApplyingState
+            && component.NetSyncEnabled) // TODO consider removing this and avoiding the component resolve altogether.
```
PJB3005: Pull this check into another function, it's shared by `SetData()`.

PJB3005 (changes requested):

PJB3005 (approved):

PR #5289, opened 2024-07-06
Don't overide the robust injectors and always use the build configuration

VasilisThePikachu:
Space station 14 has a "build client tools" batch file, as it turned out it error out instantly if you had never compiled release due to it overiding msbuild to forcibly use the release version. This pr removes that feature and just compiles it again for the target config.

### DIFF
--- a/MSBuild/XamlIL.targets
+++ b/MSBuild/XamlIL.targets
@@ -22,8 +22,6 @@
   <!-- XamlIL does not make use of special Robust configurations like DebugOpt. Convert these down. -->
   <PropertyGroup>
     <RobustInjectorsConfiguration>$(Configuration)</RobustInjectorsConfiguration>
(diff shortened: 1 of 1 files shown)


PJB3005:
Ok wait I'm taking a look at this and the reason this is broken is different than I thought.

PJB3005:
So this is fucking stupid. The solution *is* in fact configured to use the correct configuration when building for Tools:
image

The problem is that compiling the project directly with `dotnet run` **this gets completely ignored**. For fuck's sake??? This makes absolutely no fucking sense what is the .NET SDK doing.

I don't know how we would go about fixing this short of adding a Tools build target to every project, which would be stupid as hell. Or figuring out the best spot to ask for support with MSBuild like this.

VasilisThePikachu:
~~Wouldn't you ask the dotnet team? Don't they make msbuild?~~

You did

PJB3005:
Closing since we established this isn't the proper fix.

PR #5201, opened 2024-06-02
tweak: visual improvement of OptionButton

Voomra:
## About the PR
OptionButton has an opaque background.  
OptionButton is no longer ugly! :tada: 

## Media
sample

## Changelog

:cl:
- tweak: visual improvement of OptionButton

PR #5290, opened 2024-07-07
Add styleclass to OptionsButton popup background widget

eoineoineoin:
Allows customization of the background used by the OptionButton popup items, instead of always rendering transparent. I suspect that it wasn't always transparent, because the code was _almost_ setup to have a background, but probably bitrot away. 

End result is essentially the same as https://github.com/space-wizards/RobustToolbox/pull/5201, but this change is significantly smaller. Has no effect by default; will make a content PR to supply a reasonable default.

### DIFF
--- a/Robust.Client/UserInterface/Controls/OptionButton.cs
+++ b/Robust.Client/UserInterface/Controls/OptionButton.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Numerics;
 using Robust.Client.Graphics;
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/Controls/OptionButton.cs
```diff
@@ -14,6 +14,7 @@ public class OptionButton : ContainerButton
         public const string StyleClassOptionButton = "optionButton";
         public const string StyleClassPopup = "optionButtonPopup";
         public const string StyleClassOptionTriangle = "optionTriangle";
+        public const string StyleClassOptionsBackground = "optionsBackground";
```
ShadowCommander:
```suggestion
        public const string StyleClassOptionsBackground = "optionButtonBackground";
```

We have a few different controls with Option as part of the name, so it's better to differentiate more just in case.

ShadowCommander (changes requested):
Just minor naming stuff. Otherwise, everything looks good.

These controls should really be converted to XAML at some point.

ShadowCommander (approved):

PR #5292, opened 2024-07-09
Adds a new generic control that represents clickable item slot

luizwritescode:
## About this PR

Adds `ItemSlotButton` as a generic `Control` class that represents a button with an item slot.

Adds a tab in the `uitest` command for testing this feature

Also separated the tabs into their own methods inside of `UITestCommand` command to fix the Tab titles and make it more organized (it still has no dedicated .xaml file, but at least it's more readable now).

Captura de tela 2024-07-09 120013

## Why

As far as I know, the only way have sprites with a `OnPressed ` event is nesting Controls and this PR aims to make that easier. This opens a few possibilities to make UI a bit better. I can definitely see a use case replacing the odd UI design choices that work around machines that interact with entities.

Captura de tela 2024-07-08 181634

Allows inserting and ejecting objects from any UI with a single control. Could also replace most if not all "Eject" buttons from the game.

## Technical details

It simply inherits `ContainerButton` and adds a `SpriteView` and a `Label` as fields. It then manages them as Children of it's root `BoxContainer`. I found it problematic to keep empty Controls in the tree, so it will dynamically remove and add them as needed.
Had to create 4 constructors and 2 `SetEntity()` methods so it would be fully compatible with `SpriteView`.

## TODO
Drafting this for now because: 
- I still need to create an override for the click method so it could use custom sounds instead of the default UI click sound.
- Add an override of `OnMouseEntered` and `OnMouseExited` to allow for custom behavior, like changing the sprite when the mouse is over the button.
- Also, as of right now, it doesn't seem to change with the  "button open right" "button open left" style classes
- Maybe add Left and Right `LabelPositionMode` for the Label. I can only think of a few cases where this would be useful, but it would be nice to have. 

Let me know if I forgot to add anything or if there is unecessary code here

### DIFF
--- a/Robust.Client/Console/Commands/UITestCommand.cs
+++ b/Robust.Client/Console/Commands/UITestCommand.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Numerics;
 using Robust.Client.Graphics;
 using Robust.Client.UserInterface;
 using Robust.Client.UserInterface.Controls;
@@ -8,6 +9,9 @@ using Robust.Shared.IoC;
 using Robust.Shared.Localization;
 using Robust.Shared.Maths;
 using Robust.Shared.Utility;
+using Robust.Shared.GameObjects;
+using Robust.Shared.Map;
+using Robust.Shared.Timing;
 
 namespace Robust.Client.Console.Commands;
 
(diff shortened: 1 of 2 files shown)


PJB3005:
My gut feeling is this is too specific of a control to warrant being in the engine. Any reason this can't just be in content?

luizwritescode:
there are 2 reasons I preferred to submit this here instead of content:
 1. I wanted an easy way to test it without changing any already implemented UI so changing `UITestCommand` made sense.
 2. Since `SpriteView` and `BoxContainer` are in the engine itself, before I started coding it made sense to have this close to them

Granted now that I think about it I have no issues with rewriting this for content aside from needing a new name for it that won't clash with `SlotButton` already used for inventory.

If that is the case, after the rewrite, could remove the new class from this PR and just keep the cleaner code and tab title fix for `UITestCommand`

deltanedas:
`ItemSlot` is used in content so that would be annoying

PJB3005:
I don't think this is a good idea in engine because the exact language for how this kind of slot may work in content could radically change, and this control might not be able to fit the needs of that anymore.

> I wanted an easy way to test it without changing any already implemented UI so changing UITestCommand made sense.

This breaks `uitest` when not connected to a server by the way.

luizwritescode:
> This breaks uitest when not connected to a server by the way.

Sadly, yes. The tab that displays SpriteViews handles it by checking for `ClientRunLevel.InGame` and I missed that. 

> ItemSlot is used in content so that would be annoying

In fact, `ItemSlot`, `SlotButton`, `SlotControl`, `ItemSlotButtonContainer` and `ItemSlotUIContainer` are all taken.

Either way, having a generic control still feels like a better way to have clickable item slots in UI without having to nest controls in the xaml. Let me know if this feature is still wanted so I can rewrite it for content and rename it to something more abstract like `SpriteViewButton`.

luizwritescode:
Just realized that what I was trying to make with this can simply be achieved by nesting  3 or so controls in a Button and managing each child with separate logic. Closing this as it has become stale/unecessary. I might revisit this later to test if you can have a button set a custom AudioSource to play onpressd.

Issue #5259, opened 2024-06-20
edgeSpritePriority corners layering error


TheShuEd:
https://github.com/space-wizards/space-station-14/issues/29276
![image](https://github.com/space-wizards/RobustToolbox/assets/96445749/bd66d247-9572-46ac-9324-b86a7746fdb2)

The current logic has a bunch of bugs with rendering.
- Corner edges of a tile are compared only with the corner neighbor, which means that several layers can be drawn on one corner at once, overlapping each other
- Given the order in which edges are rendered (left to right from top to bottom of all tiles), side edges can overlap each other, also breaking priorities and visuals

Considering that the game uses only 2 tiles with edge priorities (grass and snow), these tiles are extremely rare in the Continental biome up close, and that the corners are extremely small, no one just didn't see this bug.

Unfortunately, trying to experiment with sprite edge sizes, strongly demonstrated this bug.

![image](https://github.com/space-wizards/RobustToolbox/assets/96445749/71bf3f96-95d0-41aa-a13a-d103a2a0bcbc)
![image](https://github.com/space-wizards/RobustToolbox/assets/96445749/865f019f-9239-4f05-8565-8093cf9c4c5c)

PR #5295, opened 2024-07-10
Refactor TileEdgeOverlay

metalgearsloth:
When trying to screenshot vgroid I noticed a not-insignificant amount of CPU time was spent on TileEdgeOverlay iterating thousands of tiles.

Not entirely sure if this is the best way to slot it into the existing tile rendering as I'm not sure how expensive stuff like GLBuffer ctors are (otherwise I'd just allocate a new one whenever it changed size and slap the edge quads into the same buffer). Alternatively I could make the one buffer and slap edge sprites onto the end which might also be better? PJB would know more.

image


TornadoTechnology:
It looks like all edges should be added to the texture atlas, and then be used in `RenderCommandDrawBatch`, which would not just save the rendered texture, but also draw all sprites with one call to `RenderCommand`, it would greatly increase productivity, even during generation, which constantly creates new tiles

TheShuEd:
Can it fix
https://github.com/space-wizards/RobustToolbox/issues/5259

?

TheShuEd:
after this PR, tiles with same edge priority looks like this

image

PR #5300, opened 2024-07-12
Update GridSplit_Tests.cs to not use Component.Owner

Plykiya:
- xform.Anchored is obsolete, replaced with TransformSystem's AnchorEntity function
- Usage of `IEnumerable<MapGridComponent> GetAllMapGrids(MapId mapId)` replaced with `IEnumerable<Entity<MapGridComponent>> GetAllGrids(MapId mapId)`, all future references to .Owner are then properly taken from the `Entity<MapGridComponent>` instead of the `<MapGridComponent>` and don't need changing

image

### DIFF
--- a/Robust.UnitTesting/Shared/Map/GridSplit_Tests.cs
+++ b/Robust.UnitTesting/Shared/Map/GridSplit_Tests.cs
@@ -143,6 +143,7 @@ public sealed class GridSplit_Tests
         var entManager = sim.Resolve<IEntityManager>();
         var mapManager = sim.Resolve<IMapManager>();
         var mapSystem = sim.Resolve<IEntityManager>().System<SharedMapSystem>();
+        var transformSystem = sim.Resolve<IEntityManager>().System<SharedTransformSystem>();
         var mapId = sim.CreateMap().MapId;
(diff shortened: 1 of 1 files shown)


Robust.UnitTesting/Shared/Map/GridSplit_Tests.cs
```diff
@@ -158,13 +159,14 @@ public void ReparentSplit()
...
-        anchoredXform.Anchored = true;
+
+        transformSystem.AnchorEntity((anchored, anchoredXform), gridEnt);
         Assert.That(anchoredXform.Anchored);
 
         mapSystem.SetTile(gridEnt, new Vector2i(2, 0), Tile.Empty);
         Assert.That(mapManager.GetAllGrids(mapId).Count(), Is.EqualTo(2));
 
-        var newGrid = mapManager.GetAllMapGrids(mapId).First(x => x != grid);
+        var newGrid = mapManager.GetAllGrids(mapId).First(x => x.Comp != grid);
```
metalgearsloth, line 169: Anything touching mapmanager you're spinning your wheels on because they should be on mapsystem instead.

PR #5308, opened 2024-07-19
Add method to TrimEnd FormattedMessage

metalgearsloth:
Sometimes you may want to deliberately add a newline at the end so better to just give this control instead of implicitly do it.

### DIFF
--- a/Robust.Shared/Utility/FormattedMessage.cs
+++ b/Robust.Shared/Utility/FormattedMessage.cs
@@ -236,6 +236,40 @@ public sealed partial class FormattedMessage : IReadOnlyList<MarkupNode>
     }
 
     /// <summary>
+    /// Removes excess whitespace at the end of markup.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +235,40 @@ public void Clear()
...
+    {
+        for (var i = _nodes.Count - 1; i >= 0; i--)
+        {
+            var node = _nodes[i];
+            var markup = node.Value;
+
+            if (markup.StringValue == null)
+                break;
+
+            var value = markup.StringValue.Trim();
```
PJB3005, line 251: Isn't this going to trim at the end of *both* sides of the last node?

PJB3005 (changes requested):
Would it be important for this to clean up formatting nodes that become empty as the result of this transformation?

PR #5310, opened 2024-07-19
tweak(HWId): Refactor code to use guard statements

Brandon-Huu:
bottom text

### DIFF
--- a/Robust.Shared/Network/HWId.cs
+++ b/Robust.Shared/Network/HWId.cs
@@ -11,24 +11,23 @@ namespace Robust.Shared.Network
 
         public static byte[] Calc()
         {
-            if (OperatingSystem.IsWindows())
-            {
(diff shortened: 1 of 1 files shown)


PJB3005:
Is this literally just swapping the if statement around?

Brandon-Huu:
> Is this literally just swapping the if statement around?

yes

PJB3005:
That is not an improvement to the code.

Conceptually, the code as it currently is is "if we are on Windows do A, on other platforms fall back to B". The intent is that future platforms could simply be chained off as an else-if.

This PR would make such a change harder, as it clearly cements the code as "this is only supported on Windows"

Leaving aside the question of what an improved HWID system in the future might entail, a *good* styling improvement would be to pull the Windows logic into its own function, and then the "large if block" would no longer be an issue. Regardless, I do not think this kind of minor stylistic change is worth a PR in any case.

Issue #5309, opened 2024-07-19
Client randomly crashes in SharedUserInterfaceSystem.Update


stalengd:
Client crashes with the following exeption:

```
Unhandled exception. System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNext()
   at Robust.Shared.GameObjects.SharedUserInterfaceSystem.Update(Single frameTime) in E:\Projects\space-station-14\RobustToolbox\Robust.Shared\GameObjects\Systems\SharedUserInterfaceSystem.cs:line 951
   at Robust.Shared.GameObjects.EntitySystemManager.TickUpdate(Single frameTime, Boolean noPredictions) in E:\Projects\space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntitySystemManager.cs:line 306
   at Robust.Client.GameStates.ClientGameStateManager.PredictTicks(GameTick predictionTarget) in E:\Projects\space-station-14\RobustToolbox\Robust.Client\GameStates\ClientGameStateManager.cs:line 521
   at Robust.Client.GameStates.ClientGameStateManager.ApplyGameState() in E:\Projects\space-station-14\RobustToolbox\Robust.Client\GameStates\ClientGameStateManager.cs:line 454
   at Robust.Client.GameController.Tick(FrameEventArgs frameEventArgs) in E:\Projects\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.cs:line 535
   at Robust.Client.GameController.<StartupContinue>b__58_0(Object sender, FrameEventArgs args) in E:\Projects\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.cs:line 224
   at Robust.Shared.Timing.GameLoop.Run() in E:\Projects\space-station-14\RobustToolbox\Robust.Shared\Timing\GameLoop.cs:line 240
   at Robust.Client.GameController.ContinueStartupAndLoop(DisplayMode mode) in E:\Projects\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.Standalone.cs:line 162
   at Robust.Client.GameController.GameThreadMain(DisplayMode mode) in E:\Projects\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.Standalone.cs:line 147
[...]

PR #5311, opened 2024-07-19
Fixed data race in ParallelTracker

stalengd:
Judging by the tests I have done, this should fix the [random client crash](https://github.com/space-wizards/RobustToolbox/issues/5309). At least it won't work any worse anyway.
Race may occur in changed piece of code because, for example, 2 last threads will do `Decrement()`, first one then enter if statement and encounder 0 of PendingTasks, setting event. This will release control on the calling side of parallel manager and new parallel job can be queued. Meanwhile, one last thread may still be right after if statement and only now decide to continue execution, setting event which may alredy be used for other parallel job.

### DIFF
--- a/Robust.Shared/Threading/ParallelManager.cs
+++ b/Robust.Shared/Threading/ParallelManager.cs
@@ -290,9 +290,9 @@ internal sealed class ParallelManager : IParallelManagerInternal
         /// </summary>
         public void Set()
         {
-            Interlocked.Decrement(ref PendingTasks);
-
-            if (PendingTasks <= 0)
+            // We should atomically get new value of PendingTasks
+            // as the result of Decrement call and use it to prevent data race.
+            if (Interlocked.Decrement(ref PendingTasks) <= 0)
                 Event.Set();
         }
     }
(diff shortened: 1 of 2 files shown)


Robust.Shared/Threading/ParallelManager.cs
```diff
@@ -290,9 +290,9 @@ private sealed class ParallelTracker
...
         public void Set()
         {
-            Interlocked.Decrement(ref PendingTasks);
-
-            if (PendingTasks <= 0)
+            // We should atomically get new value of PendingTasks
+            // as the result of Decrement call and use it to prevent data race.
+            if (Interlocked.Decrement(ref PendingTasks) <= 0)
                 Event.Set();
         }
```
metalgearsloth, line 297: Realistically it should be using a channel but good catch. Do you mind also writing a release note for it.

stalengd: and I broke everything...

stalengd: done

PR #5312, opened 2024-07-20
Add analyzer/fixer for replacing ProtoId<EntityPrototype> with EntProtoId

Tayrtahn:
Adds an analyzer that detects any use of `ProtoId<EntityPrototype>` and flags it as an error. To make life easier, also adds a code fixer that will swap it with `EntProtoId` for you.

This is implemented as a new attribute: `PreferOtherTypeAttribute`. It can be applied to a generic class or struct and takes two arguments: the type to look for as the T type argument for the generic, and the type that should be used as a replacement. This is applied to `ProtoId<T>` like this:
```c#
[PreferOtherType(typeof(EntityPrototype), typeof(EntProtoId))]
```
Hopefully that's self-explanatory enough.

If you're thinking that flagging it as an error seems excessive, I'm inclined to agree, but I want to make note of the fact that any use of `ProtoId<EntityPrototype>` will trip a [debug assert](https://github.com/space-wizards/RobustToolbox/blob/2664061993545007b7accc1e5a3b4058dfb70280/Robust.Shared/Prototypes/PrototypeManager.ValidateFields.cs#L197) when running the YAML linter. It might make sense to remove the debug assert in favor of this analyzer and possibly downgrade this from an error to a warning, so let me know if that would be preferred.

I also had to add the nuget package for NUnit code fix verifiers, since trying to develop analyzers/fixers without a functioning unit test is a miserable time. If that's not desired, let me know and I can get rid of the test and the package import.

### DIFF
--- a/Robust.Analyzers/PreferOtherTypeAnalyzer.cs
+++ b/Robust.Analyzers/PreferOtherTypeAnalyzer.cs
@@ -35,8 +35,6 @@ public sealed class PreferOtherTypeAnalyzer : DiagnosticAnalyzer
 
     private void AnalyzeField(SyntaxNodeAnalysisContext context)
     {
-        var preferOtherTypeAttribute = context.Compilation.GetTypeByMetadataName(AttributeType);
-
         if (context.Node is not VariableDeclarationSyntax node)
             return;
 
@@ -51,6 +49,8 @@ public sealed class PreferOtherTypeAnalyzer : DiagnosticAnalyzer
         var symbolInfo = context.SemanticModel.GetSymbolInfo(node.Type);
         if (symbolInfo.Symbol?.GetAttributes() is not { } attributes)
             return;
+
+        var preferOtherTypeAttribute = context.Compilation.GetTypeByMetadataName(AttributeType);
 
         foreach (var attribute in attributes)
         {
--- a/Robust.Analyzers/PreferOtherTypeAnalyzer.cs
+++ b/Robust.Analyzers/PreferOtherTypeAnalyzer.cs
@@ -1,0 +1,70 @@
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Robust.Roslyn.Shared;
+
+namespace Robust.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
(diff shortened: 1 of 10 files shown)


Robust.Shared/Analyzers/PreferOtherTypeAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+/// <summary>
+/// Marks that use of a generic Type should be replaced with a specific other Type
+/// when the type argument T is a certain Type.
+/// </summary>
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct)]
+public sealed class PreferOtherTypeAttribute(Type type, Type otherType) : Attribute
+{
+    public readonly Type Type = type;
+    public readonly Type OtherType = otherType;
+}
```
PJB3005: Maybe rename these to `genericArgument` and `replacementType`?

Robust.Analyzers/PreferOtherTypeAnalyzer.cs
```diff
@@ -0,0 +1,75 @@
...
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class PreferOtherTypeAnalyzer : DiagnosticAnalyzer
+{
+    private const string AttributeType = "Robust.Shared.Analyzers.PreferOtherTypeAttribute";
+
+    private static readonly DiagnosticDescriptor PreferOtherTypeDescriptor = new(
+        Diagnostics.IdPreferOtherType,
+        "Use the specific type",
+        "Use the specific type {0} instead of {1} when T is {2}",
```
PJB3005:
```suggestion
        "Use the specific type {0} instead of {1} when the type argument is {2}",
```

Robust.Analyzers/PreferOtherTypeAnalyzer.cs
```diff
@@ -0,0 +1,75 @@
...
+    public override void Initialize(AnalysisContext context)
+    {
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.ReportDiagnostics | GeneratedCodeAnalysisFlags.Analyze);
+        context.EnableConcurrentExecution();
+        context.RegisterSyntaxNodeAction(AnalyzeField, SyntaxKind.VariableDeclaration);
+    }
+
+    private void AnalyzeField(SyntaxNodeAnalysisContext context)
+    {
+        var preferOtherTypeAttribute = context.Compilation.GetTypeByMetadataName(AttributeType);
```
PJB3005: Best to move this lookup to later before it's used, so that it doesn't have to run for every single declaration in the project.

PR #5313, opened 2024-07-20
Fix mouse hover not updating for new controls

metalgearsloth:
It only ever updated on movement previously. The issue is for example if a new window shows up where the mouse is (or any control) it doesn't handle it. Just checking it every frame AFAIK shouldn't be that expensive. Worst case we just have some flag to check it that gets set on <mouse movement OR controls changed>.


Robust.Client/UserInterface/UserInterfaceManager.Input.cs
```diff
@@ -144,11 +144,8 @@ public void KeyBindUp(BoundKeyEventArgs args)
     public void MouseMove(MouseMoveEventArgs mouseMoveEventArgs)
     {
         _resetTooltipTimer();
-        // Update which control is considered hovered.
-        var newHovered = MouseGetControl(mouseMoveEventArgs.Position);
-        SetHovered(newHovered);
```
PJB3005: I wouldn't remove this code to avoid potential issues with input ordering.

If you have an input relying on the hovered control, then mouse move -> input could start getting incorrect results, especially if the game happens to hitch at the wrong time. This is because the frame update hover update only happens *after* input events whereas the mouse move code ensures inputs always result in consistent state.

PJB3005 (changes requested):

PR #5322, opened 2024-07-29
TryFindComponentOnEntityContainerOrParent no fail please

IProduceWidgets:
TryFindComponentOnEntityContainerOrParent
would fail trying to Resolve on parents that don't have the component its searching for. This prevents that by changing the Resolve to TryComp.

### DIFF
--- a/Robust.Shared/Containers/SharedContainerSystem.cs
+++ b/Robust.Shared/Containers/SharedContainerSystem.cs
@@ -380,9 +380,6 @@ namespace Robust.Shared.Containers
             if (entityQuery.TryComp(xform.ParentUid, out foundComponent))
                 return true;
 
-//            if (entityQuery.Resolve(xform.ParentUid, ref foundComponent, false))
-//                return true;
-
             return TryFindComponentOnEntityContainerOrParent(xform.ParentUid, entityQuery, ref foundComponent);
         }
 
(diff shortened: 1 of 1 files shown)


ElectroJr:
Isn't the trycomp just equivalent to the old resolve? Unless people are passing in components via the ref? I have no idea why this method even takes in a ref comp

deltanedas:
yeah shouldnt foundComponent be out instead of ref

IProduceWidgets:
Agreed, I switched it to out and tested that it works recursively as expected.

Fortunately I think I'm the only person who has tried to use this on content so there's nothing to fix!

IProduceWidgets:
> Isn't the trycomp just equivalent to the old resolve? Unless people are passing in components via the ref? I have no idea why this method even takes in a ref comp

I have no idea why it was using ref. I don't think there's any uses of the method on content, probably because of its scrungliness, and the method that returns all found parent components instead of the first one doesn't use ref and also uses trycomp instead. As for if they're functionally different, the resolve throws an assert if any of the parents while recursing **don't** have the component, which is no good here. Also as ref it required you to pass a ref, which is not needed imo, and unintuitive.

IProduceWidgets:
Bump.
It's just a bugfix, not new functionality.

PJB3005 (changes requested):
This is a minor breaking change which I'd rather not merge. I do get that `out` makes more sense but what's done is done, it should be possible to implement the function regardless.

IProduceWidgets:
Hokay, made it ref again to avoid the breaky.

IProduceWidgets:
Bump

PJB3005 (approved):

PR #5324, opened 2024-07-30
Add ObjectSerializer, AppearanceComponent.AppearanceDataInit, and AppearanceSystem.AppendData

SlamBamActionman:
**1. ObjectSerializer**

This PR adds ObjectSerializer, a serializer for the generic `object` type. The serializer attempts to get the type from the .yaml (use `!type:<type>` e.g. `!type:string`) via `ISerializationManager` and `IReflectionManager.TryLooseGetType`.

**2. AppearanceDataInit**

AppearanceSystem has a new property `AppearanceDataInit`. This property is used to set initial values for `AppearanceData` in yaml while also being compatible with serialization. The reason why `AppearanceData` isn't being set directly in yaml is because there are several components that, upon the entity being spawned, add entries to `AppearanceData`. Were it to be serializable this would cause excessive bloat when saving yaml prototypes and also causes the "Prototype modifies component on spawn" assert to fail. 

`AppearanceDataInit` adds the data to `AppearanceData` via the setter. The reason for it being done this way is because the addition must happen *before* the `AppearanceComponent` is initialized but *after* the yaml is parsed. This is to ensure reading the `AppearanceData` from a prototype directly still works (example: `SharedChameleonClothingSystem` reading the prototype via `EntityPrototype.TryGetComponent` to copy the `AppearanceData`). 

**3. AppendData**

AppearanceSystem also now has AppendData, which can be used to copy data from another AppearanceComponent without clearing the existing data.

**Breaking Changes**

This PR access-restricts `AppearanceComponent` to `SharedAppearanceSystem`. This breaks any outside system that uses the obsolete method `AppearanceComponent.TryGetData`. To fix this, any system outside `SharedAppearanceSystem` and its inheritors must change to using `SharedAppearanceSystem.TryGetData`.

Because of this, for SS14 https://github.com/space-wizards/space-station-14/pull/30550 must first be merged to avoid build errors.

### DIFF
--- a/Robust.Shared/GameObjects/Components/Appearance/AppearanceComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Appearance/AppearanceComponent.cs
@@ -2,20 +2,8 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using Robust.Shared.GameStates;
-using Robust.Shared.Serialization;
 using Robust.Shared.Serialization.Manager.Attributes;
 using Robust.Shared.ViewVariables;
-
-
-
-using Robust.Shared.Reflection;
-using Robust.Shared.IoC;
-using Robust.Shared.Serialization.Manager;
-using Robust.Shared.Serialization.Markdown;
-using Robust.Shared.Serialization.Markdown.Mapping;
-using Robust.Shared.Serialization.Markdown.Validation;
-using Robust.Shared.Serialization.TypeSerializers.Interfaces;
-using Robust.Shared.Utility;
 
 namespace Robust.Shared.GameObjects;
 
@@ -43,7 +31,7 @@ public sealed partial class AppearanceComponent : Component
     /// </remarks>
     [ViewVariables] internal bool UpdateQueued;
 
-    [DataField(customTypeSerializer: typeof(AppearanceSerializer))]
+    [DataField]
     public Dictionary<Enum, object> AppearanceData = new();
 
     [Obsolete("Use SharedAppearanceSystem instead")]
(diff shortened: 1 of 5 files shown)


Robust.Shared/GameObjects/Components/Appearance/AppearanceComponent.cs
```diff
@@ -30,7 +31,8 @@ public sealed partial class AppearanceComponent : Component
     /// </remarks>
     [ViewVariables] internal bool UpdateQueued;
 
-    [ViewVariables] internal Dictionary<Enum, object> AppearanceData = new();
+    [DataField]
+    public Dictionary<Enum, object> AppearanceData = new();
```
metalgearsloth: This must use access attribute.

Robust.Shared/GameObjects/Systems/SharedAppearanceSystem.cs
```diff
@@ -115,6 +115,32 @@ public void CopyData(Entity<AppearanceComponent?> src, Entity<AppearanceComponen
...
+        dest.Comp ??= EnsureComp<AppearanceComponent>(dest);
+
+        foreach (var (key, value) in srcComp.AppearanceData)
+        {
+            dest.Comp.AppearanceData[key] = value;
+        }
+
+        Dirty(dest, dest.Comp);
+        QueueUpdate(dest, dest.Comp);
+    }
```
metalgearsloth, line 143: Having one ensure and one resolve as hidden behaviors is pretty nasty, you should just be consistent between them.

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/ObjectSerializer.cs
```diff
@@ -0,0 +1,92 @@
...
+            value = serializationManager.Read(type, node, hookCtx, context);
+
+            if (value == null)
+                throw new NullReferenceException($"Found null data for {node}, expected {type}");
+        }
+
+        return value;
+    }
+    #endregion
+
```
metalgearsloth, line 64: Last time I tried something like this it broke pretty easily on primitive types so this needs accompanying tests.

metalgearsloth (changes requested):
Also obviously needs a release note.

Robust.Shared/GameObjects/Components/Appearance/AppearanceComponent.cs
deltanedas: why shouldnt the entire component have access, theres api for modifying stuff

Robust.Shared/GameObjects/Systems/SharedAppearanceSystem.cs
SlamBamActionman: The resolve is for the source, which must have the component. The EnsureComp is for the destination, which is not guaranteed to have the component, and will add it if it doesn't. This is the same behavior as in CopyData.

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/ObjectSerializer.cs
SlamBamActionman: The current tests for prototypes already go through and attempt the serialization; would they not count for this Serializer too?

If not, how would such a test look like?

PR #5329, opened 2024-08-02
Add IBoundUserInterfaceState

metalgearsloth:
Lets me use structs in some cases for BUI states. Should still be backwards compatible for forks for a bit.


PJB3005 (commented):
In what scenarios does this enable lower allocations from using structs?

PR #5336, opened 2024-08-05
Add a copy to clipboard button on alert popups.

Titian3:
This is my first RT PR, Apologies if have done any of this incorrectly.

Changes:
- Added copy button that stores UI alert message popups to clipboard. 

For fixing [this issue](https://github.com/space-wizards/space-station-14/issues/28560) about copying out ban messages but would be good for relaying connection issues and other errors shown here.

image

### DIFF
--- a/Robust.Client/UserInterface/IUserInterfaceManager.cs
+++ b/Robust.Client/UserInterface/IUserInterfaceManager.cs
@@ -64,7 +64,7 @@ namespace Robust.Client.UserInterface
 
         IDebugMonitors DebugMonitors { get; }
 
-        void Popup(string contents, string title = "Alert!", bool clipboardButton = true);
+        void Popup(string contents, string? title = null, bool clipboardButton = true);
 
         Control? MouseGetControl(ScreenCoordinates coordinates);
 
--- a/Robust.Client/UserInterface/IUserInterfaceManager.cs
+++ b/Robust.Client/UserInterface/IUserInterfaceManager.cs
@@ -64,7 +64,7 @@ namespace Robust.Client.UserInterface
 
         IDebugMonitors DebugMonitors { get; }
 
-        void Popup(string contents, string title = "Alert!", string clipboardButton = "Copy");
+        void Popup(string contents, string title = "Alert!", bool clipboardButton = true);
 
         Control? MouseGetControl(ScreenCoordinates coordinates);
 
--- a/Robust.Client/UserInterface/IUserInterfaceManager.cs
+++ b/Robust.Client/UserInterface/IUserInterfaceManager.cs
@@ -64,7 +64,7 @@ namespace Robust.Client.UserInterface
 
         IDebugMonitors DebugMonitors { get; }
 
-        void Popup(string contents, string title = "Alert!");
+        void Popup(string contents, string title = "Alert!", string clipboardButton = "Copy");
 
         Control? MouseGetControl(ScreenCoordinates coordinates);
 
--- a/Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
+++ b/Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
@@ -53,14 +53,45 @@ internal sealed partial class UserInterfaceManager
             }
         }
 
-        public void Popup(string contents, string title = "Alert!")
+        public void Popup(string contents, string title = "Alert!", string clipboardButton = "Copy")
         {
             var popup = new DefaultWindow
             {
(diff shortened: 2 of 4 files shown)


PJB3005:
Ban messages in-game aren't shown with this alert dialog, that's part of `LauncherConnecting` in content.

In general this dialog is pretty bad design wise and not used much, but I don't mind merging this as it can't really hurt.

Robust.Client/UserInterface/IUserInterfaceManager.cs
```diff
@@ -64,7 +64,7 @@ public partial interface IUserInterfaceManager
 
         IDebugMonitors DebugMonitors { get; }
 
-        void Popup(string contents, string title = "Alert!");
+        void Popup(string contents, string title = "Alert!", string clipboardButton = "Copy");
```
PJB3005: I would rather this be a boolean that indicates whether there is a copy button or. I can't imagine many reasons the button would need to be renamed, and as it stands this can't be localized practically.

Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
```diff
@@ -53,14 +53,45 @@ private void RunArrange(Control control)
...
+                Text = clipboardButton,
+                MinSize = new Vector2(100, 30),
+            };
+
+            copyButton.OnPressed += _ =>
+            {
+                _clipboard.SetText(contents);
+            };
+
+            var grid = new GridContainer
```
PJB3005: Use a `BoxContainer` instead of a grid for vertical layout like this.

Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
```diff
@@ -53,14 +53,45 @@ private void RunArrange(Control control)
...
+            {
+                Columns = 1,
+                VSeparationOverride = 10,
+            };
+
+            grid.AddChild(label);
+
+            var buttonContainer = new GridContainer
```
PJB3005: Again, `BoxContainer` but horizontal this time.

PJB3005 (changes requested):

Robust.Client/UserInterface/IUserInterfaceManager.cs
Titian3: I'm just paranoid for localization now after doing it so much in SS14.
Flag based:
False:
image
True:
image

Both with the boxContainer now.

Titian3:
> Ban messages in-game aren't shown with this alert dialog, that's part of `LauncherConnecting` in content.
> 
> In general this dialog is pretty bad design wise and not used much, but I don't mind merging this as it can't really hurt.

Thanks will go do another PR for that one too.

Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
```diff
@@ -53,14 +53,56 @@ private void RunArrange(Control control)
...
+                Orientation = BoxContainer.LayoutOrientation.Vertical
+            };
+
+            vBox.AddChild(label);
+
+            if (clipboardButton)
+            {
+                var copyButton = new Button
+                {
+                    Text = "Copy",
```
PJB3005: Localize this.

Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
```diff
@@ -53,14 +53,56 @@ private void RunArrange(Control control)
...
+                var hBox = new BoxContainer
+                {
+                    Orientation = BoxContainer.LayoutOrientation.Horizontal
+                };
+
+                var spacer = new Control
+                {
+                    MinSize = new Vector2(0, 30), // This will grow to take available space
+                    HorizontalExpand = true
+                };
```
PJB3005: You can just give the parent container `HAlignment.Right` and then you don't need to bother with a spacing control like this.

PJB3005 (changes requested):

Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
Titian3: Also localized the title as it was a hardcoded default param.

Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
Titian3: Updated this and removed all unnecessary comments etc too.

PR #5344, opened 2024-08-07
Better location reporting for DataField analyzers

Tayrtahn:
Improves the precision of reported locations for the "redundant tag" and "VV ReadWrite" analyzers. Instead of highlighting the entire member declaration, only the specific attribute is now highlighted.

### DIFF
--- a/Robust.Analyzers/DataDefinitionAnalyzer.cs
+++ b/Robust.Analyzers/DataDefinitionAnalyzer.cs
@@ -154,13 +154,13 @@ public sealed class DataDefinitionAnalyzer : DiagnosticAnalyzer
 
             if (HasRedundantTag(fieldSymbol))
             {
-                TryGetAttributeLocation(field, "DataField", out var location);
+                TryGetAttributeLocation(field, DataFieldAttributeName, out var location);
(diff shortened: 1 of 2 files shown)


Robust.Analyzers/DataDefinitionAnalyzer.cs
```diff
@@ -152,12 +154,14 @@ private void AnalyzeDataField(SyntaxNodeAnalysisContext context)
 
             if (HasRedundantTag(fieldSymbol))
             {
-                context.ReportDiagnostic(Diagnostic.Create(DataFieldRedundantTagRule, context.Node.GetLocation(), fieldSymbol.Name, type.Name));
+                TryGetAttributeLocation(field, "DataField", out var location);
```
PJB3005: You're using a string literal here but a constant on the other code.

PJB3005 (changes requested):

Robust.Analyzers/DataDefinitionAnalyzer.cs
Tayrtahn: Whoops

PR #5350, opened 2024-08-11
XAML hot reloading

nyeogmi:
This is an implementation of XAML hot reloading. After this change, you can make an arbitrary change to a UI object's XAML files, then see altered behavior the next time someone constructs that object.

Here's a video so you can see it happen:
Everything_cyyMS4dYOa

Here's a summary of the approach:

- When compiling XAML for a project, write XamlMetadataAttribute entries that contain enough info to feed a JIT compiler.
- Watch the project directory for new XAML and tell a background service about it each time any comes in.
- Use XamlX in System.Reflection.Emit mode at runtime to JIT-compile all the new XAML that arrives.
- When objects try to populate themselves, redirect that to the background service, which will call a JIT-ed version of the functionality if one exists.
- If none existed, call the old AOT-compiled version.

Here's how I dealt with various tradeoffs:

- Runtime resolution of methods is both expensive and scary: as a result, I made it so implementers of the "hot reloader" interface can always return false to have it do nothing.
- The watcher is quite a horror, so it's off to one file away from the rest of the hot loader implementation. It exports no public methods -- nobody else in the codebase should be using its approach as written.
- The XAML compiler is `internal`, so nothing should be able to bust out of the sandbox and grab it.
- At @Jezithyr 's advice, anything that could possibly break sandbox is behind conditional compilation flags. 

Here are unrelated things I fixed up in the process of this:

- One Pidgin version for the whole project. To fix this, I rewrote the parsers without parser combinators and removed it from the compile-time project.

Things I didn't fix:

- A prod-safe version of this would use Cecil to generate code. I did not do this. The XamlX+Cecil API is incredibly dependent on filenames right now and I have no idea how to get that working.
- Extensive testing: I'm a very new SS14 dev and things _appear_ to work fine, but I don't know where the fragile points are. I also don't know what SS14's attitude towards unit tests are -- I'd be happy to write some if we can agree on what they would cover, but this change also seems like it would be on the very annoying side to autotest.
- Generating huge amounts of IL at build time is kind of a cursed approach. I don't know what else would actually work though. Given that we're committed to this strategy, I'll admit I don't think our code is _that bad_.


nyeogmi:
Ack, tests are _very much_ on fire. I'll look at this soon, unless any of you wants to. I don't even see a run configuration for these -- how do I get them running locally?

(EDIT: @deathride58 sez the commands in this file should do ok: `https://github.com/space-wizards/RobustToolbox/blob/master/.github/workflows/build-test.yml`)

nyeogmi:
OK, tests are passing for now.

deathride58:
To paraphrase a mirror of our report from the Discord: we've done the testing, and are able to confirm that XAML hot reload works under a selection of increasingly cursed Linux environments, along with all the Windows environments we have. We are unable to test under MacOS.

We encourage knowledgeable contribs to give this PR a try themselves, and *really* put this through its paces! Knowledgeable contribs will very likely already know how to go about this. Those who don't can look at the edit history of this post if they'd like a spoiler/hint on how to do it.

nyeogmi:
@nok-ko proposed:

> Basic lathe test works just fine on macos! Is there a way to force a reload of a currently open GUI? Like for example reloading something like MainMenu would be a little annoying with that limitation. that said it’s very swag. also found an unrelated macos graphics bug in the process (moving windows across desktops with different DPIs messes up the scale *sigh*) but that’s the macos life

I think my thought is that (1) this should be implemented at some point (2) it's really hard to do.

Here's the proposal I wrote for how to implement this. 

> fwiw: I thought about how to implement "reload ui on hot reload" -- this is the best i can do:
>
> - when someone creates a UI object, create a weak reference to it in a global service
> - when the global service hot reloads a type, go through the list of ui objects and if they're of that type, (1) somehow clear all > - effects of Populate() (2) call Populate() again
> - once per second: clear old bad weak references out of that service
>
> i will possibly implement this as a separate change as it is an ugly implementation and kind of fraught
>
> my prediction is that it will interact badly with everything that already exists, because widgets generally assume that references they hold to subwidgets are still vaild. really, the ideal thing would be to run the constructor again. 
> 
> i can do this by moving the constructor code to a separate method during AoT compilation, but this strategy becomes increasingly lunatic. i would say it is a top-grade bad idea
>
> i predict the cost of this at 300 to 500 lines. might actually be lower than this. if it's drastically lower and i get a chance to code it soon, i'll do it in the same PR. 

Since reading that person's comment, I have discovered that it's possible to call the constructor more than once in C#. (https://stackoverflow.com/a/27380691) So, I might try that out. This won't reset any fields to default values though, so it's fraught.

I honestly don't think this is the best possible way to do it: the best possible way might involve a more radical refactor, where FindControl is replaced by something that itself returns a proxy object. (This would also deal somewhat with the intermittent crashing.) Overall, it's a hard problem -- the XAML only got touched in one place, but all the objects the XAML creates are touched in more than one place.

I would prefer to postpone this but can try to solve the problem now if someone sees it as nonnegotiable. I do suspect there may be literally no good way to do it though.

nok-ko:
> I would prefer to postpone this but can try to solve the problem now if someone sees it as nonnegotiable. I do suspect there may be literally no good way to do it though.

I think this is 100% shippable as-is, we can save the improvement for later. Good work!

KIBORG04:
If I lived in the same city as you, I'd find you, hit you over the head, lock you in the basement, chain you up and suck your dick until it falls off. It's about time.

chavonadelal:
Making user interfaces will be a little less painful. Thank you!!!

Jezithyr:
yippee-happy

This is *fantastic work*

Robust.Client.Injectors/Robust.Client.Injectors.csproj
```diff
@@ -1,17 +1,18 @@
 <Project Sdk="Microsoft.NET.Sdk">
+    <Import Project="..\MSBuild\Robust.Engine.props" />
 
     <PropertyGroup>
-        <TargetFramework>netstandard2.0</TargetFramework>
```
PJB3005, line 4: So the *reason* that `Robust.Client.Injectors` is NS2.0 is that Visual Studio still ships a .NET FX based MSBuild for some godforsaken reason. This means that when having Robust.Client.Injectors loaded directly by the main MSBuild process... that would break.

Except we don't do that anyways right now due to file locking issues, so maybe it's fine to give up on that. Whatever.

Robust.Client.Injectors/Program.cs
```diff
@@ -55,9 +55,10 @@ public void LogCustomEvent(CustomBuildEventArgs e)
...
             IDictionary targetOutputs) => throw new NotSupportedException();
 
-        public bool ContinueOnError { get; }
-        public int LineNumberOfTaskNode { get; }
-        public int ColumnNumberOfTaskNode { get; }
-        public string ProjectFileOfTaskNode { get; }
+        // PYREX NOTE: This project was extremely null-unsafe before I touched it. I'm just marking what it did already
+        public bool ContinueOnError { get; } = default!;
+        public int LineNumberOfTaskNode { get; } = default!;
+        public int ColumnNumberOfTaskNode { get; } = default!;
```
PJB3005: You don't need to `= default!;` for primitives like bool/int

Robust.Client/UserInterface/XAML/Proxy/IXamlHotReloadManager.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Robust.Client.UserInterface.XAML.Proxy;
+
+/// <summary>
+/// This service locates the SS14 source tree and watches for changes to its Xaml files.
+///
+/// It then reloads them instantly.
+///
+/// It depends on XamlProxyManager and is only available on TOOLS builds.
+/// </summary>
```
PJB3005: For XML doc comments, `<summary>` should generally be, well, a summary. Most stuff should go in `<remarks>` generally (split by `<para>` if you wanna be fancy).

Robust.Client/UserInterface/XAML/Proxy/XamlHotReloadManager.Real.cs
```diff
@@ -0,0 +1,198 @@
...
+using Robust.Shared.ContentPack;
+using Robust.Shared.IoC;
+using Robust.Shared.Log;
+
+namespace Robust.Client.UserInterface.XAML.Proxy
+{
+    /// <summary>
+    /// The real implementation of XamlHotReloadManager, whose behavior is described
+    /// in IXamlHotReloadManager.cs.
+    /// </summary>
```
PJB3005: You can just do like `<see cref="IXamlHotReloadManager" />` so that IDEs can nicely link to the interface.

Robust.Client.Injectors/Program.cs
nyeogmi: Rider/Resharper warns if I don't give these a value, so I'll just define them as `default` instead of `default!`. I'll make the string `null!` to emphasize how that's different and scary.

Robust.Client/UserInterface/XAML/Proxy/IXamlProxyManager.cs
```diff
@@ -0,0 +1,70 @@
...
+///
+/// However, it can also call into an externally-provided implementation of the Xaml
+/// widget.
+///
+/// No source of externally-provided implementations actually exists, by default --
+/// you will need to call SetImplementation with a blob of xaml source code to provide
+/// one. IXamlHotReloadManager is an example of a service that calls into that
+/// functionality.
+/// </summary>
+public interface IXamlProxyManager
```
PJB3005: Does this interface need to be public?

Robust.Client/UserInterface/XAML/Proxy/XamlHotReloadManager.Stub.cs
```diff
@@ -0,0 +1,28 @@
...
+using Robust.Shared.Log;
+
+namespace Robust.Client.UserInterface.XAML.Proxy
+{
+    /// <summary>
+    /// A stub implementation of XamlHotReloadManager.
+    ///
+    /// Its behavior is to do nothing.
+    /// </summary>
+    internal sealed class XamlHotReloadManager : IXamlHotReloadManager
```
PJB3005: For stubs like this I prefer relying on the fact we're using interfaces here, and just making this a separate type like `XamlHotReloadManagerStub`. Then register a different type in `ClientIoC.cs` based on the `#if`.

PJB3005: Also, you should probably turn the hot reloading system *off* in integration tests, so you can also make the stub implementations used in `ClientIoC.cs` when the client is running headlessly.

Robust.Client/UserInterface/XAML/Proxy/XamlJitDelegate.Real.cs
```diff
@@ -0,0 +1,17 @@
+﻿#if TOOLS
```
PJB3005: Honestly this file is like a single delegate definition, I wouldn't even bother putting this behind an `#if`.

Robust.Client/UserInterface/XAML/Proxy/XamlHotReloadManager.Stub.cs
```diff
@@ -0,0 +1,28 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Robust.Shared.Asynchronous;
+using Robust.Shared.ContentPack;
+using Robust.Shared.IoC;
+using Robust.Shared.Log;
+
+namespace Robust.Client.UserInterface.XAML.Proxy
```
PJB3005: Namespaces are pretty inconsistent between file-scoped and not in this PR, best to use file-scoped everywhere (for new files).

Robust.Client/UserInterface/XAML/Proxy/XamlImplementationStorage.Real.cs
```diff
@@ -0,0 +1,211 @@
...
+    Dictionary<string, string> _fileContent = new();
+
+    /// <summary>
+    /// For each filename, we store the list of types interested in this file.
+    ///
+    /// This may be more than one -- that is, multiple Controls can be associated
+    /// with the same XAML file.
+    ///
+    /// At time of writing, for instance, StrippingMenu has no *.xaml file, and
+    /// therefore it uses DefaultWindow.xaml.
```
PJB3005: This doesn't make much sense to me. The `XamlIlPopulateTrampoline` should only ever be called by the constructor. For cases like windows where you have multiple levels of inheritance, the base constructor automatically gets ran by C#, and each layer of constructor runs its own `XamlIlPopulateTrampoline`. So `StrippingMenu` shouldn't have an entry here, but when `StrippingMenu` calls the base constructor for `DefaultWindow`, that should just look up the entry for `DefaultWindow`.

Robust.Client/UserInterface/XAML/Proxy/XamlImplementationStorage.Real.cs
```diff
@@ -0,0 +1,211 @@
...
+/// The specific relationships are documented below.
+/// </summary>
+internal sealed class XamlImplementationStorage
+{
+    /// <summary>
+    /// For each filename, we store its last known Uri.
+    ///
+    /// When we compile the new implementation, we will use the same Uri.
+    /// </summary>
+    Dictionary<string, Uri> _fileUri = new();
```
PJB3005: All these fields should be made `private readonly`.

Robust.Client/UserInterface/XAML/Proxy/XamlProxyHelper.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using Robust.Shared.IoC;
+
+namespace Robust.Client.UserInterface.XAML.Proxy
+{
+    /// <summary>
+    /// Specializing an interface to use an argument of generic type doesn't work
+    /// super well in our sandbox, for whatever reason.
```
PJB3005: Can you show me what you tried here? I'm not sure what the problem is.

Robust.Client/UserInterface/XAML/Proxy/XamlProxyManager.Real.cs
```diff
@@ -0,0 +1,128 @@
...
+        public bool Populate(Type t, object o)
+        {
+            return _xamlImplementationStorage.Populate(t, o);
+        }
+
+        /// <summary>
+        /// Wrap XamlJitCompiler.Compile.
+        ///
+        /// Lazily initialize XamlJitCompiler -- initializing it is expensive, and initializing it on every
+        /// individual assembly load is quadratic.
```
PJB3005: A lot of these comments should probably just be line comments in the code itself instead of inside the doc comment.

RobustXaml/XamlJitCompiler.Real.cs
```diff
@@ -0,0 +1,171 @@
+﻿#if DEBUG
```
PJB3005: I wouldn't bother conditionally compiling this code, it won't get included by the client when it shouldn't be anyways.

PJB3005 (commented):
Mostly looks good to me

Robust.Client/UserInterface/XAML/Proxy/IXamlProxyManager.cs
nyeogmi: Can't think of a good reason for it to be -- fixing locally.

Robust.Client/UserInterface/XAML/Proxy/XamlProxyHelper.cs
nyeogmi: The code in LowLevelCustomizations generated effectively the same bytecode, but the generic type was IXamlProxyManager. This is back when IXamlProxyManager was public. I don't remember offhand what the error message is -- I will collect it for you after I've dealt with your other comments.

Robust.Client/UserInterface/XAML/Proxy/XamlProxyManager.Real.cs
nyeogmi: Moved several.

RobustXaml/XamlJitCompiler.Real.cs
nyeogmi: Fixed.

Robust.Client/UserInterface/XAML/Proxy/XamlImplementationStorage.Real.cs
nyeogmi: Fixed.

Robust.Client/UserInterface/XAML/Proxy/XamlImplementationStorage.Real.cs
nyeogmi: So, here's how I think you're saying it should behave, which I agree with:

* `StrippingMenu::constructor` calls `DefaultWindow::constructor`
* `DefaultWindow::constructor` calls `XamlIlPopulateTrampoline` which ultimately calls `Populate(typeof(DefaultWindow), this)`
* `StrippingMenu` does not call `XamlIlPopulateTrampoline` (it has no xaml file and no `RobustXamlLoader::Load` call that would have been replaced)

The comment was written while I was in a state of deep confusion about why StrippingMenu would have the same XamlMetadataAttribute as another class. The comment is effectively gibberish -- I'm not sure I was thinking.

Anyway, I figured out the actual reason: XamlMetadataAttribute started with the default `[AttributeUsage(Inherited = true)]`. I've fixed this and made it so there's no longer a list involved here.

Robust.Client/UserInterface/XAML/Proxy/XamlJitDelegate.Real.cs
nyeogmi: Fixed.

Robust.Client/UserInterface/XAML/Proxy/XamlHotReloadManager.Stub.cs
nyeogmi: Fixed. (in any file GitHub thought was new, we now use file-scoped namespaces)

Robust.Client/UserInterface/XAML/Proxy/XamlHotReloadManager.Stub.cs
nyeogmi: Fixed. (on both counts)

Robust.Client/UserInterface/XAML/Proxy/XamlProxyHelper.cs
nyeogmi: If I make IXamlProxyManager public and replace XamlProxyHelper references with IXamlProxyManager references, I get this series of bizarre ILVerify errors on the AoT code: 

```
[ERRO] res.typecheck: Content.Client: ILVerify: Call not allowed on abstract methods., method: void Content.Client.Launcher.LauncherConnectingGui.!XamlIlPopulateTrampoline(Content.Client.Launcher.LauncherConnectingGui)
[ERRO] res.typecheck: Content.Client: ILVerify: The 'this' parameter to the call must be the calling method's 'this' parameter., method: void Content.Client.Shuttles.UI.ShuttleNavControl.!XamlIlPopulateTrampoline(Content.Client.Shuttles.UI.ShuttleNavControl)
```

This spooked me the heck out -- feels like spooky action at a distance.

Note that those are the only two things I changed -- the changes look like this:

image
image

Do you have any take on this?

Robust.Client/UserInterface/XAML/Proxy/XamlHotReloadManager.Real.cs
nyeogmi: Fixed.

Robust.Client/UserInterface/XAML/Proxy/IXamlHotReloadManager.cs
nyeogmi: Fixed.

Robust.Client.Injectors/Robust.Client.Injectors.csproj
nyeogmi: Reverted.

Robust.Client/UserInterface/XAML/Proxy/XamlProxyHelper.cs
nyeogmi: EDIT: We worked this out on Discord. It's just the ILVerifier complaining about a `Call` that should have been a `Callvirt`

I thought the error message was unclear but actually it's very clear -- I thought it was opposed to the concept of calling the method in the error message, but it was just opposed to the use of a Call instruction _in_ the method in the error message.

Robust.Client/ClientIoC.cs
```diff
@@ -146,6 +147,16 @@ public static void RegisterIoC(GameController.DisplayMode mode, IDependencyColle
             deps.Register<IConfigurationManagerInternal, ClientNetConfigurationManager>();
             deps.Register<IClientNetConfigurationManager, ClientNetConfigurationManager>();
             deps.Register<INetConfigurationManagerInternal, ClientNetConfigurationManager>();
+
+#if DEBUG
```
PJB3005:
```suggestion
#if TOOLS
```

PR #5354, opened 2024-08-11
Added debug warnings to Appearancecomponent

TheShuEd:
## WHY
image


Robust.Shared/GameObjects/Systems/SharedAppearanceSystem.cs
```diff
@@ -79,6 +84,10 @@ public bool TryGetData<T>(EntityUid uid, Enum key, [NotNullWhen(true)] out T val
             return true;
         }
 
+#if DEBUG
+        Log.Warning($"Trying to get AppearanceData from entity({uid}) that does not have an {nameof(AppearanceComponent)}");
+#endif
+
```
metalgearsloth, line 90: This will already log an error above.

PJB3005 (commented):
If we wanted to do this, couldn't you just remove the `false` to the `Resolve` that disables `logMissing`?

Cinkafox:
how can I understand your pain🥲

PR #5355, opened 2024-08-13
Add cvar to limit entities passed into nearby command

DrSmugleaf:
Another day another one of my admins crashed rmc14
See https://github.com/space-wizards/RobustToolbox/pull/5282 for reasoning
Default 5 (you are usually doing self nearby not entities nearby which was the problem)

### DIFF
--- a/Robust.Shared/Utility/CollectionExtensions.cs
+++ b/Robust.Shared/Utility/CollectionExtensions.cs
@@ -283,24 +283,5 @@ namespace Robust.Shared.Utility
             value = default;
             return false;
         }
-
-        public static bool HasMoreThan<T>(this IEnumerable<T> enumerable, int than)
-        {
-            if (enumerable.TryGetNonEnumeratedCount(out var count))
-                return count > than;
-
-            using var enumerator = enumerable.GetEnumerator();
-            checked
-            {
-                while (enumerator.MoveNext())
-                {
-                    than--;
-                    if (than < 0)
-                        return true;
-                }
-            }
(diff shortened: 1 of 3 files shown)


Robust.Shared/Utility/CollectionExtensions.cs
```diff
@@ -283,5 +283,24 @@ public static T Pop<T>(this IList<T> list)
             value = default;
             return false;
         }
+
+        public static bool HasMoreThan<T>(this IEnumerable<T> enumerable, int than)
+        {
+            if (enumerable.TryGetNonEnumeratedCount(out var count))
+                return count > than;
+
+            using var enumerator = enumerable.GetEnumerator();
```
PJB3005: Does toolshed's architecture guarantee that double enumeration like this won't cause a problem?

DrSmugleaf: I don't think so, in testing it worked fine with self and nearby but I don't think there are any guarantees

DrSmugleaf: The other option is doing ToArray at the start and hoping that there are not hundreds of thousands of entities

DrSmugleaf: Fixed in [`6a420ae` (#5355)](https://github.com/space-wizards/RobustToolbox/pull/5355/commits/6a420ae6c887afd126a0655c57a6333d3052c153)

PR #5356, opened 2024-08-13
Make overlay manager accept multiple overlays of the same type

stalengd:
This can be really necessary for some use cases, I personally run into such limitation of the OverlayManager and it is almost impossible to completely bypass without changing the engine. This change does not introduce any breaking changes and does not spoil performance.

### DIFF
--- a/Robust.Client/Graphics/Overlays/IOverlayManager.cs
+++ b/Robust.Client/Graphics/Overlays/IOverlayManager.cs
@@ -22,6 +22,7 @@ public interface IOverlayManager
 
     bool HasOverlay(Type overlayClass);
     bool HasOverlay<T>() where T : Overlay;
+    bool HasOverlay(Overlay overlay);
 
     IEnumerable<Overlay> AllOverlays { get; }
 }
--- a/Robust.Client/Graphics/Overlays/OverlayManager.cs
+++ b/Robust.Client/Graphics/Overlays/OverlayManager.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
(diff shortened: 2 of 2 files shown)


metalgearsloth:
Do you have an example usecase?

I've written half a dozen overlays and never seen a need for it across singulo or weather so curious what needs it.

PJB3005:
The intended design is that the overlay instance itself should be able to handle N operations itself.

Retrofitting this onto the existing API is probably a bad idea IMO.

stalengd:
> Do you have an example usecase?
> 
> I've written half a dozen overlays and never seen a need for it across singulo or weather so curious what needs it.

I do have a real usecase.
I have developed an overlay effect based on the threshold filter, it is quite generic and needs only texture and threshold value as an input. 

[2024-08-11-23-08-33.webm](https://github.com/user-attachments/assets/da3a5fa2-f413-4e76-9261-6d9f60469578)

So, many clients (by client I mean the calling side here) or even one with several textures may show this overlay simultaneously. Create dummy overlay types? What if I know how many instances of this effect are only in runtime (e.g. from prototypes)? 

> The intended design is that the overlay instance itself should be able to handle N operations itself.
> 
> Retrofitting this onto the existing API is probably a bad idea IMO.

Yes, thats how I handle this for now, but I think this is not ideal:
1. You cant control ZIndex (what I care about now) and Space (what I dont really care for now) for individual "sub-overlays",
2. You should erect not really pretty system for this on content-side while we can do this once and for all here, in engine.

metalgearsloth:
>     2. You should erect not really pretty system for this on content-side while we can do this once and for all here, in engine.

We've had code to support multiple overlays before and it turned all of the callers into a giant mess as they now had to individually track their own overlay lifetimes instead of the engine being able to do it for them.

The overlay draw call is agnostic enough that you can just implement multiple passes in 1 overlay easily and I think this would be better worked through on your end.

PR #5360, opened 2024-08-15
Fixes luminance underflow in FromHCY

redmoogle:
image

image

was working on OD was confused by the blue gradient then realized that the blue value was negative and this fixes that

### DIFF
--- a/Robust.Shared.Maths/Color.cs
+++ b/Robust.Shared.Maths/Color.cs
@@ -735,7 +735,7 @@ namespace Robust.Shared.Maths
                 b = 0.0f;
             }
 
-            var m = luminance - (0.30f * r + 0.59f * g + 0.11f * b);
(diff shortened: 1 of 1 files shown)


PJB3005:
What input values reproduce this behavior?

redmoogle:
> What input values reproduce this behavior?

Should be any value between pure red and pure green ( like the picture ablve)

PJB3005:
After a bunch of digging I've come to the conclusion the "HCY" color space is ill-defined and mathematically incorrect, so I'd rather just remove it from RT entirely. I'll close this PR and commit a deprecation for them.

For OpenDream's use case, I recommend you implement your own copy of the functions instead. For what it's worth, it seems like this fix isn't correct, and instead there's some scaling factor missing. Check this I guess: https://chilliant.blogspot.com/2012/08/rgbhcy-in-hlsl.html

PR #5364, opened 2024-08-17
Completely obsolete noSpawn

Plykiya:
Relies on https://github.com/space-wizards/space-station-14/pull/31122 getting merged first, then upstream can fully purge the usage of noSpawn in the RT

### DIFF
--- a/Robust.Shared/Prototypes/EntityPrototype.cs
+++ b/Robust.Shared/Prototypes/EntityPrototype.cs
@@ -96,16 +96,9 @@ namespace Robust.Shared.Prototypes
         [DataField("localizationId")]
         public string? CustomLocalizationID { get; private set; }
 
-
         /// <summary>
         ///     If true, this object should not show up in the entity spawn panel.
         /// </summary>
-        [ViewVariables]
-        [NeverPushInheritance]
-        [DataField("noSpawn")]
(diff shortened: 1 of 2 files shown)


slarticodefast:
The energy shotgun PR added a few more uses of `noSpawn: true` [#30643](https://github.com/space-wizards/space-station-14/pull/30643)
Those will have to be fixed before merging this.

0x6273:
1. This **removes** noSpawn, it doesn't obsolete it. There's a big difference, obsoleting just adds a warning whereas removing is a breaking change.
2. Because this is a breaking change, you should state this in [RELEASE-NOTES.md](https://github.com/space-wizards/RobustToolbox/blob/master/RELEASE-NOTES.md).
3. This is probably too soon, given that content just stopped using it. Forks should be given some time to remove their own uses of it. (It's been marked as obsolete for quite a while, but since it's used in yaml I think most IDEs won't highlight it as a warning)

metalgearsloth:
>     3. This is probably too soon, given that content just stopped using it. Forks should be given some time to remove their own uses of it. (It's been marked as obsolete for quite a while, but since it's used in yaml I think most IDEs won't highlight it as a warning)

Content has stopped using it for a while, the issue is that it keeps erroneously getting added and better to just rip the bandaid off. It can be regexed fairly easily anyway.

PR #5365, opened 2024-08-17
Remove startup traversal entirely

n00kii:
when `TransformComponent` starts up, it runs an initial grid traversal check. however, you can't flip the `GridTraversal` bool of the transform component before it's already started when you spawn an entity. ~therefore, i add an argument (defaulting to true) to set the `GridTraversal` boolean as soon as possible, before the transform component is started up.~

~i've only added it to `SpawnAttachedTo` because intuitively, if you use this function you're expecting whatever you attach to whatever to STAY attached, and not silently grid-traverse detach; so now you have the option.~

now, traversal no longer subscribes to `TransformStartupEvent`, so you can modify traversal as you wish before it actually happens


Robust.Shared/GameObjects/EntityManager.Spawn.cs
```diff
@@ -73,12 +73,12 @@ public EntityUid[] SpawnEntities(MapCoordinates coordinates, List<string?> proto
...
     }
 
-    public virtual EntityUid SpawnAttachedTo(string? protoName, EntityCoordinates coordinates, ComponentRegistry? overrides = null, Angle rotation = default)
+    public virtual EntityUid SpawnAttachedTo(string? protoName, EntityCoordinates coordinates, ComponentRegistry? overrides = null, Angle rotation = default, bool traversal = true)
     {
         if (!coordinates.IsValid(this))
             throw new InvalidOperationException($"Tried to spawn entity {protoName} on invalid coordinates {coordinates}.");
 
-        var entity = CreateEntityUninitialized(protoName, coordinates, overrides, rotation);
+        var entity = CreateEntityUninitialized(protoName, coordinates, overrides, rotation, traversal);
```
metalgearsloth: Considering we're using overloads for different functionality it would be better for traversal to just not run.

n00kii: do you mean to disable startup traversal completely? wouldn't that be a breaking change?

metalgearsloth: yes, what's wrong with breaking changes?

I'd rather just rip off a bandaid instead of making the already slow entity spawning even slower.

n00kii: sg, was just confirming

Robust.Shared/GameObjects/Systems/SharedGridTraversalSystem.cs
```diff
@@ -27,12 +27,6 @@ internal sealed class SharedGridTraversalSystem : EntitySystem
     public override void Initialize()
     {
         base.Initialize();
-        SubscribeLocalEvent<TransformStartupEvent>(OnStartup);
-    }
-
-    private void OnStartup(ref TransformStartupEvent ev)
-    {
-        CheckTraverse(ev.Entity);
     }
```
metalgearsloth, line 30: This is going to be a massively breaking change and the corresponding spawn code needs to be adjusted to handle this.

n00kii: so if im understanding what each usecase for the different spawn functions are, here's how this should look like:

(1) base `Spawn`: spawn in nullspace
(2) `Spawn` with `MapCoordinates` overload: spawn on map, reparent to grid if overlap
(3) `SpawnAttachedTo`: spawn with parent and position provided by `EntitiyCoordinates`
(4) `SpawnAtPosition`: same as (2) but transforms provided `EntityCoordinates` to `MapCoordinates` first
(5) `TrySpawnInContainer`, `SpawnInContainerOrDrop`, `TrySpawnNextTo`, `SpawnNextToOrDrop`: as their names imply

ignoring (1) since trivial case, and (5) since those functions are built up from (1..4). (2) and (4) just need to make sure they check for overlapping grids and reparent if one is found. is it fine to do have it implemented with the changes i've made to the `MapCoordinates` overload of `CreateEntityUninitialized`? (3) uses the `EntityCoordinates` overload of `CreateEntityUninitialized` which doesn't check for grids so no need to worry about traversing in that case

n00kii: but now im confused because this is just making traversal implicitly built into certain spawn functions rather than relying on the system for it; is that what we want?

ElectroJr:
For the same reasons as outlined in my comment in #5551, generally you don't want to allow an entity that's not on a grid to still be parented to the grid, so I don't generally think this is a good idea. The actual content issue you're trying to fix can be fixed without engine changes by just correctly setting the world rotation. Though ensuring that the laser beam doesn't split when a grid is rotating or moving requires reworking the visual so that it uses a single entity rather than two

slarticodefast:
The PR that needed this engine change (https://github.com/space-wizards/space-station-14/pull/28797) was closed in favor of another solution, that works without this (https://github.com/space-wizards/space-station-14/pull/34515).

Not sure if this engine PR should still be merged on its own.

PR #5370, opened 2024-08-18
Fixed client crash on devwindow inspect control's property which throws exception

stalengd:
Some control properties even in engine itself (e.g. Label.Text) may throw exceptions. devwindow scans all properties and does not handle any exceptions, which leads to crash in case of one. I think this is not right for debugging purpuses, so I did that in such case exception just printed as a value of faulty property:

### DIFF
--- a/Robust.Client/Console/Commands/Debug.cs
+++ b/Robust.Client/Console/Commands/Debug.cs
@@ -579,7 +579,16 @@ namespace Robust.Client.Console.Commands
         private static string GetMemberValue(MemberInfo? member, Control control, string separator, string
                 wrap = "{0}")
         {
-            var value = member?.GetValue(control);
+            object? value = null;
+            try
+            {
+                value = member?.GetValue(control);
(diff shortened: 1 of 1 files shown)


Robust.Client/Console/Commands/Debug.cs
```diff
@@ -579,7 +579,16 @@ internal static string PropertyValuesString(Control control, string key)
...
         {
-            var value = member?.GetValue(control);
+            object? value = null;
+            try
+            {
+                value = member?.GetValue(control);
+            }
+            catch (Exception exception)
+            {
+                var exceptionToPrint = exception.InnerException ?? exception;
```
PJB3005, line 589: For completeness I would prefer if `InnerException` here is only accessed if it is indeed a `TargetInvocationException`.

stalengd: Okay, this makes sense

PR #5372, opened 2024-08-20
Make resetting contacts on the client only set is touching if it is true

DrSmugleaf:
For sloth to vibe check
Long story but making guns predicted leads to projectiles on the client not properly getting StartCollideEvent raised because resetcontacts blindly sets this to true and that makes physics never raise a start collide event because by the time it checks and it is true it has already been set to true by this


metalgearsloth:
this is crazy but I also need to throw out half of physics for prediction after we get deltas so.

PR #5373, opened 2024-08-20
add bool[] support to shaders

qwerltaz:
You can now use bool arrays in `.swsl` shader files. Can be useful.


Robust.Client/Graphics/Clyde/GLObjects/Clyde.ShaderProgram.cs
```diff
@@ -418,6 +418,38 @@ private void SetUniformDirect(int slot, Vector2[] vectors)
...
+
+            public void SetUniform(int uniformName, bool[] bools)
+            {
+                var uniformId = GetUniform(uniformName);
+                SetUniformDirect(uniformId, bools);
+            }
+
+            private void SetUniformDirect(int slot, bool[] bools)
+            {
+                var intBools = new int[bools.Length];
```
PJB3005: Don't like this allocation here, probably better to use `stackalloc`.

Robust.Client/Graphics/Clyde/GLObjects/Clyde.ShaderProgram.cs
```diff
@@ -418,6 +418,38 @@ private void SetUniformDirect(int slot, Vector2[] vectors)
...
+
+                unsafe
+                {
+                    fixed (bool* boolPtr = bools)
```
PJB3005: No need to be iterating the boolean list with a pointer.

PJB3005 (changes requested):

PJB3005 (approved):

PR #5382, opened 2024-08-22
Don't spam BUI closing in state handling

metalgearsloth:
We just do what content does and defer it until update. Saves performance + we don't have some BUIs that do special logic on open re-running it constantly (e.g. open on mouse position).


Winkarst-cpu:
- Fix containers now.
- Forgot to update the comment?
изображение

PR #5388, opened 2024-08-22
Make AudioSystem ZOffset public

### LINKED ISSUE #5352: Eye zoom should affect audio Z offset, make public audio ZOffset
This issue https://github.com/space-wizards/space-station-14/issues/21946
To make it possible, the Robust.Client AudioSystem should have ZOffset public set.
```cs
namespace Robust.Client.Audio;
public sealed partial class AudioSystem : SharedAudioSystem
...
public override float ZOffset
{
    get => _zOffset;
    set
    {
        _zOffset = value;
        _audio.SetZOffset(value);

        var query = AllEntityQuery<AudioComponent>();

        while (query.MoveNext(out var audio))
        {
            // Pythagoras back to normal then adjust.
            var maxDistance = GetAudioDistance(audio.Params.MaxDistance);
            var refDistance = GetAudioDistance(audio.Params.ReferenceDistance);

            audio.MaxDistance = maxDistance;
            audio.ReferenceDistance = refDistance;
        }
    }
}
```
Or another alternative, by making partial class `AudioSystem.Public` with SetZOffset public method
```cs
namespace Robust.Client.Audio;
public sealed partial class AudioSystem
...
public void SetZOffset(float offset)
{
    ZOffset = offset;
}
```

Winkarst-cpu:
Resolves #5352.

### DIFF
--- a/Robust.Shared/Audio/Systems/SharedAudioSystem.cs
+++ b/Robust.Shared/Audio/Systems/SharedAudioSystem.cs
@@ -47,7 +47,7 @@ public abstract partial class SharedAudioSystem : EntitySystem
     /// </summary>
     public int OcclusionCollisionMask { get; set; }
 
-    public virtual float ZOffset { get; protected set; }
+    public virtual float ZOffset { get; set; }
 
     public override void Initialize()
     {
--- a/Robust.Client/Audio/AudioSystem.cs
+++ b/Robust.Client/Audio/AudioSystem.cs
@@ -60,7 +60,7 @@ public sealed partial class AudioSystem : SharedAudioSystem
     public override float ZOffset
     {
         get => _zOffset;
-        protected set
+        set
         {
             _zOffset = value;
             _audio.SetZOffset(value);


Robust.Shared/Audio/Systems/SharedAudioSystem.cs
```diff
@@ -47,7 +47,7 @@ public abstract partial class SharedAudioSystem : EntitySystem
     /// </summary>
     public int OcclusionCollisionMask { get; set; }
 
-    public virtual float ZOffset { get; protected set; }
+    public virtual float ZOffset { get; set; }
```
metalgearsloth, line 51: Is there a reason this is needed? SetZOffset is already public and content setting this isn't handled properly at all ie.
1. cvar changes can bulldoze it.
2. Overwriting it would mean the max-length adjustments could get bulldozed.

Winkarst-cpu: Because ``AudioSystem`` overrides ``ZOffset`` that comes from its parent, ``SharedAudioSystem`` and when overriding, you can't change the modifiers since you will get [CS0507](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0507), and this is will not compile.

Winkarst-cpu: Oh... I understood what you meant now. ``IAudioListener`` already has a ``SetZOffset`` method that is public... I'm sorry for being dumb and will try it...

PR #5394, opened 2024-08-25
- add: grid priority finding

Cinkafox:
Add grid priority. Its useful for building vehicles from grids or somethink else.

https://github.com/user-attachments/assets/9375e0d9-3d5a-4702-9100-9e60d2f16c2a

By default its zero, and it doesn't affect performance.

### DIFF
--- a/Robust.Shared/Map/MapManager.Queries.cs
+++ b/Robust.Shared/Map/MapManager.Queries.cs
@@ -3,16 +3,20 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Numerics;
 using Robust.Shared.GameObjects;
+using Robust.Shared.Log;
 using Robust.Shared.Map.Components;
 using Robust.Shared.Map.Enumerators;
 using Robust.Shared.Maths;
 using Robust.Shared.Physics;
 using Robust.Shared.Physics.Collision.Shapes;
+using Robust.Shared.Utility;
 
 namespace Robust.Shared.Map;
 
 internal partial class MapManager
 {
+    private static readonly PriorityQueue<Entity<MapGridComponent>> GridPriorityQueue = new(new GridComparer());
+
     private bool IsIntersecting(
         ChunkEnumerator enumerator,
         IPhysShape shape,
@@ -109,6 +113,8 @@ internal partial class MapManager
 
         var worldAABB = shape.ComputeAABB(transform, 0);
 
+        GridPriorityQueue.Clear();
+
         var gridState = new GridQueryState(
             callback,
             worldAABB,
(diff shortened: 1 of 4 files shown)


metalgearsloth:
What's the usecase for this? Because it doesn't work if either grid has entities overlapping it breaks and if we start adding a bunch of functionality it's suddenly z-levels but less flexible.

Cinkafox:
I need a grid laying for some thinks. It is not yet clear when Z-level will appear

metalgearsloth:
> I need a grid laying for some thinks. It is not yet clear when Z-level will appear

for what things because this isn't actual grid layering.

Robust.Shared/Map/MapManager.Queries.cs
```diff
@@ -140,10 +146,19 @@ public void FindGridsIntersecting(EntityUid mapEnt, PolygonShape shape, Transfor
                 return true;
             }
 
-            state.Callback(data.Uid, data.Grid);
+            if (data.Grid.Priority > 0)
+                GridPriorityQueue.Add(data);
+            else
+                state.Callback(data.Uid, data.Grid);
```
metalgearsloth: This is one of the most expensive and called paths in the game so saying it doesn't affect performance is a bit of a lie.

Cinkafox:
> > I need a grid laying for some thinks. It is not yet clear when Z-level will appear
> 
> for what things because this isn't actual grid layering.

Some train think. I use grid for this.

Robust.Shared/Map/MapManager.Queries.cs
Cinkafox: Okay. I only have one grid in priority, namely the train. I can simply change the priority number to a boolean value

metalgearsloth:
My issue with this is:
- it makes a hot path much slower (even the shape ctor is unbearably slow but I have a bandaid pr for that).
- The code will need ripping out at some point in the next 12 months.
- It doesn't even work for what people would want to use it for (grid layering) and just supports background grids.

Robust.Shared/Map/MapManager.Queries.cs
```diff
@@ -3,16 +3,20 @@
...
 using Robust.Shared.Physics;
 using Robust.Shared.Physics.Collision.Shapes;
+using Robust.Shared.Utility;
 
 namespace Robust.Shared.Map;
 
 internal partial class MapManager
 {
+    private static readonly Queue<(EntityUid, MapGridComponent)> GridPriorityQueue = new();
+
```
metalgearsloth, line 19: Oh yeah another one this breaks the thread-safety of IMapManager.

metalgearsloth:
As it is I'm not sufficiently convinced on the worth of this PR in lieu of actual Z-levels. Additionally it breaks the thread-safety of mapmanager and in turn entitylookup which will lead to significant performance degradation on some hot paths, and am opting to close at this time.

PR #5414, opened 2024-08-31
Add AfterSaveEvent

IProduceWidgets:
Adds an AfterSaveEvent that broadcasts right after saving a map. This is to complement the BeforeSaveEvent that already exists.


PJB3005:
This feature really doesn't seem justified given the use case in SS14 seems to be to pile on spaghetti instead of doing things properly? I can't see why this is needed.

metalgearsloth:
see review

PR #5421, opened 2024-09-03
Introduce a WarningsAsErrors block to Robust.Properties.targets

Mervill:
One way to have fewer warnings is to turn some warnings into errors.

I have one suggestion to start off: `CS0105` which is the **duplicate** `using` warning. I don't think I care too much about _unused_ usings but duplicate usings should definitely be an error.

It helps take some of the load off of maintainers by turning incidental shit that causes warnings into quick failures. It's not a big ask for either us or downstreams to correct either, so I seems like a no brainer.


mirrorcult:
wasnt aware you could do this for individual warnings, good idea imo

Mervill:
The duplicate `using` from content was removed so if the tests are re-run we should be good.

PJB3005:
Could this cause undue problems for downstreams when doing merges?

slarticodefast:
Is this even needed anymore since we already enabled this in content for all non-obsoletion warnings?
https://github.com/space-wizards/space-station-14/pull/42983/

PJB3005:
Yeah I don't think this is worth investigating at this time.

PR #5422, opened 2024-09-03
Try fix invalid PVS index bug

ElectroJr:
Some code was using the default value of PvsIndex as the invalid value, even though the 0th index actually gets assigned to entities. I think zero used to be invalid previously and the checks were just never updated once that changed?


Robust.Server/GameStates/PvsSystem.ToSendSet.cs
```diff
@@ -78,7 +82,7 @@ private void AddEntity(PvsSession session, ref PvsChunk.ChunkEntity ent, ref Pvs
 
         if (meta.LifeStage >= EntityLifeStage.Terminating)
         {
-            Log.Error($"Attempted to send deleted entity: {ToPrettyString(ent.Uid)}, lifestage is {meta.LifeStage}.\n{Environment.StackTrace}");
+            Log.Error($"Attempted to send deleted entity: {ToPrettyString(ent.Uid)}, Meta lifestage: {ent.Meta.EntityLifeStage}, PVS lifestage: {meta.LifeStage}.\n{Environment.StackTrace}");
             EntityManager.QueueDeleteEntity(ent.Uid);
             return;
         }
```
Mervill: I mentioned in a conversation with PJB that the previous version of this [removed the entity from the PvsSession's working set](https://github.com/space-wizards/RobustToolbox/commit/a2d050436824cee441180ab72658054295224fe6#diff-565a71a3b03f8683dbdc3064124a2daa4fee326b5b4cf535454684ce3105b311R91-R98) instead of actually calling delete on the entity. It's not clear to me what the consequences of that are. I suspect it's incidental to the current PVS issue though.

`QueueDeleteEntity` eventually just calls `DeleteEntity` which contains:
```C#
public virtual void DeleteEntity(EntityUid? uid)
   ...
if (meta.EntityLifeStage >= EntityLifeStage.Deleted)
	return;

if (meta.EntityLifeStage == EntityLifeStage.Terminating)
{
	var msg = $"Called Delete on an entity already being deleted. Entity: {ToPrettyString(e)}";
	_sawmill.Error($"{msg}. Trace: {Environment.StackTrace}");
}
```
If you're getting the `Attempted to send deleted entity` error **and** you see entities that aren't supposed to be being deleted that _are_ being deleted do you also see this sawmill error from `DeleteEntity`? It would be further down in the logs since it's running at the end of the frame.

Might help confirm if there really is a desync between the value of `ent.Meta.EntityLifeStage` and `meta.LifeStage`.

ElectroJr: > If you're getting the Attempted to send deleted entity error and you see entities that aren't supposed to be being deleted that are being deleted do you also see this sawmill error from DeleteEntity? It would be further down in the logs since it's running at the end of the frame.

Smug is the only one that has been running into these errors so far. and those errors did appear earlier, but I don't know if they're still happening https://discord.com/channels/310555209753690112/900426319433728030/1275994976114835468

Trying to re-delete the terminating entity was just added in the hopes of maybe improving exception tolerance in case it somehow manages to actually delete the entity and fix it. Ideally that code shouldn't ever actually need to run, its a sign that something else has already gone wrong somewhere else. Though looking at the code again, it shouldn't even be doing this because this code gets run in parallel.

PR #5428, opened 2024-09-06
Fix tags with controls in RichText and OutputPanel

Lgibb18:
**Allows normal and bug-free use of tags with control like cmdlink in outputPanel and in any RichText.**
* also alliws to get FornattedMessage from RichTextLabel to get nodes
- Controls now hide normally in chat when scrolling

https://github.com/user-attachments/assets/3386a036-d0bb-4935-95de-903bbec484ec

How it was before:

https://github.com/user-attachments/assets/92a82ded-390f-47f6-a92b-861daf2c91ca

- Now the control tags work well with new lines

https://github.com/user-attachments/assets/0e65b523-6a68-4f1b-80d7-33a0e239d414

How it was before:

https://github.com/user-attachments/assets/b2fcf74e-4bf2-4cfe-bd2b-36ce62f4bcb2


JeanGrenouille010101:
Will this allow for things like fillable forms in 13?

Lgibb18:
> Will this allow for things like fillable forms in 13?


https://github.com/user-attachments/assets/660842c4-cced-46fe-badc-77687755691e

yes :)

Lgibb18:
Some of the crazy things you can make out of it
Content Client_0DjHBKMEf0
Content Client_jO4rdV76K1
Content Client_i8N7bGkyTG

Robust.Client/UserInterface/RichTextEntry.cs
```diff
@@ -67,6 +67,19 @@ public RichTextEntry(FormattedMessage message, Control parent, MarkupTagManager
             _tagControls = tagControls;
         }
 
+        internal readonly void Hide()
+        {
+            var nodeIndex = -1;
+            foreach (var node in Message)
```
ElectroJr: Constantly setting the `Visible` field of all the controls seems like a questionable way to do this.
If you just want to stop the controls from being drawn while they are out of view, AFAIK that's what the scissor box test it meant to be used for. I.e., if `Control.RectClipContent` is enabled, the check in https://github.com/space-wizards/RobustToolbox/blob/4949b34c88ef414df914d4ce59096e9cd370104e/Robust.Client/UserInterface/UserInterfaceManager.cs#L355 should stop the control from being drawn. Though I wouldn't be too surprised if its buggy or for whatever reason doesn't play nicely with scroll containers.

Lgibb18: https://github.com/user-attachments/assets/d0c91b92-187f-4101-8f91-131b3306b26b

yeah, the main problem is that it's buggy.

LevitatingTree:
wait, are the fillable forms actually in-game now? (sorry if i misunderstood)

ElectroJr:
> wait, are the fillable forms actually in-game now? (sorry if i misunderstood)

The actual forms/tags have to be implemented by content, this just fixed some engine bugs for rich text

PR #5440, opened 2024-09-15
Add shapecasts + raycasts

metalgearsloth:
Actual raycasts. Need this for AI LIDAR experiment.


eoineoineoin:
Coincidentally, I started to take a look at the raycast end of this recently (https://github.com/eoineoineoin/RobustToolbox/commits/eoin/2024.09.xx.raycasts/) - I took a slightly different approach of putting a castray() function on the shapes interface which is a bit clearer IMO (though, that strategy doesn't work for shapecasts).

If you want to run with this, that's fine, I'll leave it - I still had to finish some functions, had a bunch of todos, polish, and tests to do, but I did notice a few issues with the SharedPhysicsSystem.Queries.cs which would also be good to include.

Robust.Shared.Maths/Matrix3Helpers.cs
```diff
@@ -67,6 +67,15 @@ public static Box2 TransformBox(this Matrix3x2 refFromBox, in Box2 box)
         return Unsafe.As<Vector128<float>, Box2>(ref lbrt);
     }
 
+    /// <summary>
+    /// Gets the position of the Matrix. Will have some precision loss.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector2 Position(this Matrix3x2 t)
```
eoineoineoin: Not sure what you mean by "precision loss", but this method already exists (Matrix3x2.Translation), so no need for it (along with the tests).

metalgearsloth:
I'm porting the new C code from box2d.

TornadoTechnology:
Holy shit, it's pretty cool!

Djungelskog2:
I dont know what this means in reality but does this mean we might actually get stuff related to not being able to interact/see stuff in the dark? (not invalidating the darkness via a right click)

metalgearsloth:
So atm I wouldn't use shapecasts for a character controller til erin does his sample but the raycasts work fine as is.

kanopus952:
Could this PR have broken the hitscan? If you shoot a hitscan at an entity in space, it doesn't take any damage

ElectroJr:
> Could this PR have broken the hitscan? If you shoot a hitscan at an entity in space, it doesn't take any damage

@kanopus952  Yes it did. I don't think anyone saw your comment, if something like this happens again you should probably open an issue or ping on discord.

PR #5441, opened 2024-09-15
Bump cefglue

ike709:
Includes https://github.com/space-wizards/cefglue/pull/6 so OpenDream doesn't have a morbillion warnings about XML comments.


ike709:
I'm not sure why it's suddenly full of errors when the only change is disabling a warning?

ike709:
PJBased

PR #5442, opened 2024-09-16
Modify markup tag interfaces and fix some bugs

ElectroJr:
This PR marks the `IMarkupTag` interface and associated methods as obsolete, and adds a new `IMarkupTagHandler` interface and methods in `MarkupTagManager` that should replace the old ones.

The main reason for this change is to try prevent confusion & bugs, due to users assuming that there are multiple tag instances per rich text tag, when in reality the tag manager only ever creates one instance per tag type. This confusion has already lead to a bug in content, where `TextLinkTag` assumes that it has an instance per control.

This PR also fixes another bug, where controls that were using the `RichTextEntry` constructor to add child controls never removed those controls once the entry was removed.


juliangiebel (changes requested):
This is good. Thanks for renaming and documenting things and the bugfixes. This should make it easier for people to create richtext tag.
I'd like it if at least the tags in engine where changed to use the new interface before this gets merged otherwise this PR only needs to get its conflicts resolved and then it can get merged imo.
👍

PR #5443, opened 2024-09-16
Hot reload resources

metalgearsloth:
Turns out we already had engine support for a lot of these but no FileSystemWatcher.

Now we have:
- tiles
- shaders
- audio

.rsi + texture TBC if I finish them before undrafting.


Sh18RW:
Is there any changes? You mentioned your PR in https://github.com/space-wizards/RobustToolbox/pull/5562 but it does not solve my issue and it looks forgotten.

metalgearsloth:
> Is there any changes? You mentioned your PR in #5562 but it does not solve my issue and it looks forgotten.

I need to re-look at yours with a clear head.

Sh18RW:
If `AL.GenBuffer` fails, the line `_audioSampleBuffers.Add(buffer, new LoadedAudioSample(buffer));` in `AudioManager.Public.cs` throws an exception because the key 0 is already present in the dictionary.

Partmedia:
This appears to have broken starting the client on FreeBSD 14.2-RELEASE amd64. The client fails to launch with an error message:

```
[ERRO] clyde.win.glfw: GLFW Error: [65550] Failed to detect any supported platform
```

I tested this by checking out SS14 `db96cc78`, the first version of SS14 using RT 248.0.0 (which fails to launch), then removed this commit via interactive rebase. After removing this commit, the client launches.

Another workaround that works is launching the client with `--cvar res.prototype_reload_watch=false`

PR #5452, opened 2024-09-18
Fix ResPath CanonPath

metalgearsloth:
Apparently this is supposed to standardise to / but this isn't always the case. Alternatively we could just assert for performance reasons I'm good with either. The comment as written says this should happen. I think it's mostly DirLoader stuff can end up mixing the separators around.


Robust.Server/ServerOptions.cs
```diff
@@ -29,12 +29,12 @@ public sealed class ServerOptions
...
         ///     Directory to load all assemblies from.
         /// </summary>
-        public ResPath AssemblyDirectory { get; init; } = new(@"/Assemblies");
+        public ResPath AssemblyDirectory { get; init; } = new(@"/Assemblies/");
 
         /// <summary>
         ///     Directory to load all prototypes from.
         /// </summary>
-        public ResPath PrototypeDirectory { get; init; } = new(@"/Prototypes");
+        public ResPath PrototypeDirectory { get; init; } = new(@"/Prototypes/");
```
metalgearsloth, line 37: This is just to be consistent with client.

PJB3005 (commented):
Uh, what does this fix? You absolutely shouldn't be able to pass a `\` into a `ResPath` and expect it to work.

PJB3005:
Adding a significant amount of logic to the constructor for `ResPath` also is a bad idea. It's intentionally designed to "just" be a string wrapper for performance, this breaks that.

metalgearsloth:
> Adding a significant amount of logic to the constructor for `ResPath` also is a bad idea. It's intentionally designed to "just" be a string wrapper for performance, this breaks that.

I'll chuck an assert in? IIRC it's conversions to/from file directories but would have to double-check.

metalgearsloth:
Okay I double-checked my hot reload branch and it's just the getcontentroot iterator so I can just sanitise those inputs before they go to respath? There might be some other ones internally but nothing that seemed to trip the assert so would need to find them as they come up.

Robust.Shared/ContentPack/ResourceManager.cs
```diff
@@ -379,7 +379,10 @@ public IEnumerable<ResPath> GetContentRoots()
             {
                 if (root is DirLoader loader)
                 {
-                    yield return new ResPath(loader.GetPath(new ResPath(@"/")));
+                    // Sanitise platform specific path.
+                    var rootDir = loader.GetPath(new ResPath(@"/")).Replace(Path.DirectorySeparatorChar, '/');
```
PJB3005: This should be done by the dir loader, not the resource manager.

PJB3005 (changes requested):

Robust.Shared/ContentPack/ResourceManager.cs
metalgearsloth: yeah nfi what I was thinking

PR #5456, opened 2024-09-22
Fix TabContainer click detection when UIScale was not == 1.0

### LINKED ISSUE #5321: TabContainer's click bounds doesn't account for UI scale
Having a UI scale other than 1 leads to the tab's click bounds being off.

https://github.com/OpenDreamProject/OpenDream/assets/5714543/474f4f23-c4ff-4645-87c9-77240241579d

Ran into this in OpenDream, https://github.com/OpenDreamProject/OpenDream/issues/1871

eoineoineoin:
Fixes #5321.

Draw() and KeyBindDown() had copy-pasted-but-slightly-different code for calculating the size of each tab. The KeyBindDown() version didn't account for UIScale. Now, Draw() will just store the coordinates for each tab, and KeyBindDown() will look up the click location - removes code duplication, unnecessary calculations and fixes the bug.

### DIFF
--- a/Robust.Client/UserInterface/Controls/TabContainer.cs
+++ b/Robust.Client/UserInterface/Controls/TabContainer.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Numerics;
 using Robust.Client.Graphics;
 using Robust.Shared.Input;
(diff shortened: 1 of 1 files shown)


PJB3005 (commented):
This code is absolutely horrible but I can blame Godot so good enough for me.

PR #5459, opened 2024-09-23
Make shared compstates copy cloneable types

metalgearsloth:
Sometimes you can get trolled by stuff like autonetworked dictionaries not being copied on the client so you get mispredicts.

Downside is this allocates on server so not sure if we can just compile this out on server or what.


Robust.Shared.CompNetworkGenerator/ComponentNetworkGenerator.cs
```diff
@@ -246,10 +246,18 @@ public class ComponentNetworkGenerator : ISourceGenerator
...
+                                // get first ctor arg of the field attribute, which determines whether the field should be cloned
+                                // (like if its a dict or list)
+                                getStateInit.Append($@"
+                {name} = new(component.{name}),");
+                            }
+                            else
+                            {
+                                getStateInit.Append($@"
+                {name} = component.{name} == null ? null : new(component.{name}),");
+                            }
```
ElectroJr: Only the client should have to clone for get-state (and it would only ever have to do so once per comp).
So this should somehow check is-server/is-client to avoid allocs.

And for fixture sys, its easy to manually add the checks

metalgearsloth: I can't debug sourcegen code coz my IDE blows up but not sure if I can just get the assembly or what, I'll see if I can figure something out.

metalgearsloth: Okay I ended up using FULL_RELEASE but somehow it blows up tests. I think on an actual server it probably doesn't matter for getstate? Alternatively I use an msbuild property like pjb said.

PJB3005 (commented):
I don't think FULL_RELEASE gets communicated properly to source gens? I'd rather have this relying on a compile property dynamically isnide the source gen.

PJB3005:
Stale

metalgearsloth:
I really need to come back to this one because a bunch of content stuff is broken as a result.

PR #5466, opened 2024-09-27
EntityPrototypeView has become disposable

Ertanic:
EntityPrototypeView does not delete an entity unless done manually via `SetPrototype(null)`. This causes the test to fail in https://github.com/space-wizards/space-station-14/pull/32339.

### DIFF
--- a/Robust.Client/UserInterface/Controls/EntityPrototypeView.cs
+++ b/Robust.Client/UserInterface/Controls/EntityPrototypeView.cs
@@ -56,4 +56,15 @@ public class EntityPrototypeView : SpriteView
         base.ExitedTree();
         EntMan.TryQueueDeleteEntity(_ourEntity);
     }
+
+    protected override void Dispose(bool disposing)
+    {
+        base.Dispose(disposing);
+
+        if (!disposing)
+            return;
+
+        if (_ourEntity is not null)
+            EntMan.DeleteEntity(_ourEntity);
+    }
 }


Robust.Client/UserInterface/Controls/EntityPrototypeView.cs
```diff
@@ -56,4 +56,15 @@ protected override void ExitedTree()
...
+    protected override void Dispose(bool disposing)
+    {
+        base.Dispose(disposing);
+
+        if (!disposing)
+            return;
+
+        if (_ourEntity is not null)
+            EntMan.DeleteEntity(_ourEntity);
+    }
```
metalgearsloth, line 69: ExitedTree handles this, it seems like you're having an issue elsewhere.

Ertanic: > ExitedTree handles this, it seems like you're having an issue elsewhere.

Before adding explicit entity deletion in the `Dispose` method, [DeleteAllThenGhost](https://github.com/space-wizards/space-station-14/blob/caac678fa6e5c5a8f0602dd5b4ffbf1eea3b34ef/Content.IntegrationTests/Tests/Minds/MindTest.DeleteAllThenGhost.cs#L9) test was crashing with 100% chance. In debug mode, breakpoints and call stack I realized that the problem is in EntityPrototypeView, it generates undeletable entities. For some reason `ExitedTree` method does not work. As I [wrote](https://ptb.discord.com/channels/310555209753690112/310555209753690112/1289076513391644694) in Discord, after adding `Dispose`, the test started to pass.

metalgearsloth: Seems like ExitedTree is not being called when it should be then?

Ertanic: > Seems like ExitedTree is not being called when it should be then?

It looks like, but `Dispose` is enough for my PR to work

PJB3005 (changes requested):
This is not a good fix, find the real root cause.

Ertanic:
> This is not a good fix, find the real root cause.

I found the bug, I fixed it as far as my motivation goes. My PR will probably work. I don't really care why something doesn't work. If you are not satisfied with my solution - I can close the PR, and look for the root cause yourself. I won't do more than that. My PR is aimed at making my main PR work, not at fixing your engine.

metalgearsloth:
If you want to fix the other bug feel free to open another PR then.

PR #5473, opened 2024-09-29
Entities filtering for forks

TheShuEd:
A new entity category “ForkFiltered” has been added. Added new CVar “build.entities_category_filter”, where this category can be assigned.

Forks can assign this CVar to their server, and assign the new category to all their entities to safely hide all upstream content from spawn menu without causing conflicts.


AwareFoxy:
🙏

PR #5474, opened 2024-09-29
Add `required` keyword attributes to sandbox whitelist

ElectroJr:
Content can't currently use the `required` keyword because the associated attributes aren't whitelisted.


Robust.Shared/ContentPack/Sandbox.yml
```diff
@@ -588,7 +589,6 @@ Types:
     Vector2: { All: True }
     Vector3: { All: True }
     Vector4: { All: True }
-    Matrix3x2: { All: True }
```
ElectroJr, line 591: This is a duplicate entry

PR #5475, opened 2024-09-30
Expose GameTitle, WindowIconSet and SplashLogo to content

VasilisThePikachu:
Requirement for upstream ss14 pr


PJB3005 (changes requested):
This should probably be a general purpose API that can expose other info in the resource manifest too, like the window icon set or splash logo.

PR #5478, opened 2024-10-02
Fix RenderInRenderTarget

### LINKED ISSUE #5476: RenderInRenderTarget doesn't restore state properly
So there's render target and bound render target. RenderInRenderTarget restores both fields to the boundrendertarget (via BindTargetFull) which means if you had an existing rendertarget it would get bulldozed.

AFAIK the only area this has ever been an issue was when I added an overlay draw into the lighting drawing. I got a fix in the works unless it gets fixed on master before then.

metalgearsloth:
See the linked issue for what happens.

Fixed https://github.com/space-wizards/RobustToolbox/issues/5476

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.HLR.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.HLR.cs
@@ -452,6 +452,7 @@ namespace Robust.Client.Graphics.Clyde
 
             FenceRenderTarget(rt);
 
+            _currentRenderTarget = oldRenderTarget;
             PopRenderStateFull(state);
             _updateUniformConstants(_currentRenderTarget.Size);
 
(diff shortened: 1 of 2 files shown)


Robust.Client/Graphics/Clyde/Clyde.HLR.cs
```diff
@@ -447,11 +452,20 @@ private void RenderInRenderTarget(RenderTargetBase rt, Action a, Color? clearCol
...
             SetScissorFull(oldScissor);
             _currentMatrixModel = oldTransform;
+
+            DebugTools.Assert(_currentMatrixModel.Equals(oldTransform));
+            DebugTools.Assert(_currentScissorState.Equals(oldScissor));
+            DebugTools.Assert(_currentMatrixProj.Equals(oldMatrixProj));
+            DebugTools.Assert(oldMatrixView.Equals(_currentMatrixView));
+            DebugTools.Assert(oldRenderTarget.Equals(_currentRenderTarget));
+            DebugTools.Assert(oldBoundTarget.Equals(_currentBoundRenderTarget));
+            DebugTools.Assert(oldShader.Equals(_queuedShaderInstance));
```
amylizzle, line 468: OD breaks on this assert. Idk if that means we're doing it wrong.

PR #5479, opened 2024-10-04
feat: DrawingHandleScreen now can draw segment of bagel

Fildrance:
To create better UX for radial menu it is required to draw background for radial sectors - added feature for DrawingHandleScreen to do exactly that. To give code more usages also added parameter to draw non-filled segment - it can work as borders.

Sample of what can be done using this changes is in following attachment:

https://github.com/user-attachments/assets/5ddae535-08df-4730-b089-6131956eb2a2

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -42,6 +42,7 @@ END TEMPLATE-->
 * `RequiredMemberAttribute` and `SetsRequiredMembersAttribute` have been added to the sandbox whitelist. I.e., you can now use the `required` keyword in client/shared code.
 * Added `LineEdit.SelectAllOnFocus`.
 * ``Gametitle``,``WindowIconSet`` and ``SplashLogo`` are exposed in IGameController. These will return said information set game options or whatever is set in manifest.yml.
+* Added virtual 'Control.IsPositionInside' method for customizing logic of detecting if mouse is inside control region
 
 ### Bugfixes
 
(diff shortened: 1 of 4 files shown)


Fildrance:
Forgot about changelog, gonna add it :(

Fildrance:
Done : )

PJB3005:
Isn't this use case already covered by `HasPoint`? The `PixelRect` check is only if `RectClipContent` is set, and that's *intended* to be a rough-pass optimization only.

Fildrance:
> Isn't this use case already covered by `HasPoint`? The `PixelRect` check is only if `RectClipContent` is set, and that's _intended_ to be a rough-pass optimization only.

Ye sorry it looks like it works with HasPoint. Wierd. I'll repurpose this PR for part that renders bagle segment!

Fildrance:
Sample of how it looks with 'filled = false'
image

PJB3005:
This feels specific enough to just be in content instead.

And please don't re-use PRs like this, make a new one if you must.

Fildrance:
> This feels specific enough to just be in content instead.
> 
> And please don't re-use PRs like this, make a new one if you must.

i will not be able to use stack alloc in Content.Client

PJB3005:
Pre-allocate the buffer, it's a highly specific call it's really not that big of a deal.

Fildrance:
ok

PR #5486, opened 2024-10-07
Night Vision Light Sensetivity

Rinary1:
adds light sensetivity for night vision systems.

This not night vision, It's not about seeing people, it's about seeing everything, you can't draw everything on top.


Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -434,6 +434,12 @@ private void DrawLightsAndFov(Viewport viewport, Box2Rotated worldBounds, Box2 w
             var lastSoftness = float.NaN;
             Texture? lastMask = null;
 
+            if (_lightManager.NightVision)
+            {
+                GLClearColor(_lightManager.NightVisionColor);
+                GL.Clear(ClearBufferMask.ColorBufferBit);
+            }
+
```
metalgearsloth: GL.Clear is already being called above.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -475,7 +481,18 @@ private void DrawLightsAndFov(Viewport viewport, Box2Rotated worldBounds, Box2 w
...
+                            lastColor = new Color(component.Color.R * _lightManager.LightSensitivity,
+                                component.Color.G * _lightManager.LightSensitivity,
+                                component.Color.B * _lightManager.LightSensitivity,
+                                component.Color.A);
+                        }
+                        else
+                        {
+                            lastColor = component.Color;
+                        }
+
```
metalgearsloth: Rather than baking this in this is better done on the content side or alternatively via event.

metalgearsloth (changes requested):

metalgearsloth (changes requested):
Also another option is just drawing a texture to the LightRenderTarget in an overlay depending how you want to do it.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
Rinary1: > Rather than baking this in this is better done on the content side or alternatively via event.

wdym?

darkrell: > Rather than baking this in this is better done on the content side or alternatively via event.

Could you explain how you see this? Sending an event here would be incredibly inefficient.

darkrell:
> Also another option is just drawing a texture to the LightRenderTarget in an overlay depending how you want to do it.

Also, that solves a slightly different issue. we would still prefer to have local options for devs to adjust species sensitivity to light

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
metalgearsloth: As opposed to calling GL.ClearColor twice in a row...?

The lighting cap is incredibly small anyway. Again, there's no reason you can't just update the lighting render target from content with this.

metalgearsloth:
> Also, that solves a slightly different issue. we would still prefer to have local options for devs to adjust species sensitivity to light

I still don't see why this can't just update the lighting render target from content.

metalgearsloth:
At this stage the only useful part of this I see is adding more customisation around what clearcolor is used but none of the other changes are necessary on the engine.

StarlightHost:
@metalgearsloth 

Rewritten to use the event; is it better now?

PR #5488, opened 2024-10-08
Add Erase button for TileSpawnWindow

MilenVolf:
Small QoL for mappers. Basically, it just selects space tile on "Erase" button toggled.

Besides, I added locale for "Place Tiles" menu

## Media

https://github.com/user-attachments/assets/fa47d8b8-2450-4b7b-b713-f32669e8746c

### DIFF
--- a/Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
+++ b/Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
@@ -27,12 +27,40 @@ public sealed class TileSpawningUIController : UIController
 
     private readonly List<ITileDefinition> _shownTiles = new();
     private bool _clearingTileSelections;
+    private bool _eraseTile;
 
     public override void Initialize()
     {
         DebugTools.Assert(_init == false);
         _init = true;
         _placement.PlacementChanged += ClearTileSelection;
+    }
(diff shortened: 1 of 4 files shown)


Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
```diff
@@ -35,6 +36,33 @@ public override void Initialize()
...
+
+            var newObjInfo = new PlacementInformation
+            {
+                PlacementOption = "AlignTileAny",
+                TileType = 0,
+                Range = 400,
+                IsTile = true
+            };
+
+            _placement.BeginPlacing(newObjInfo);
```
PJB3005: Don't copy paste this from the other code.

PJB3005 (changes requested):

Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
MilenVolf: Fixed in [8a8ef23](https://github.com/space-wizards/RobustToolbox/pull/5488/commits/8a8ef234ee3edeb892e91363ea4bf16a31369101)

PR #5489, opened 2024-10-13
Separate window creation in `OSWindow.Show()` to allow creation in the background

wixoaGit:
OSWindow wouldn't create the window and its UI root until `Show()` was called. This caused issues for controls we needed to initialize & run before ever being visible. Now they are both created immediately, with `Show()` being used to make them visible.

This also required being able to set a window's size after creation, so I added the ability to do that. This is something OpenDream was going to need at some point anyway.

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.Events.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.Events.cs
@@ -109,6 +109,9 @@ namespace Robust.Client.Graphics.Clyde
 
         private void SendWindowResized(WindowReg reg, Vector2i oldSize)
         {
+            if (!reg.IsVisible) // Only send this for open windows
+                return;
+
             var loaded = RtToLoaded(reg.RenderTarget);
             loaded.Size = reg.FramebufferSize;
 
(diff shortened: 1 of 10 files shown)


Iamgoofball (approved):
░░░░░░░░░░░░▄▄                     
░░░░░░░░░░░█░░█                  
░░░░░░░░░░░█░░█                  
░░░░░░░░░░█░░░█                  
░░░░░░░░░█░░░░█                  
███████▄▄█░░░░░██████▄ 
▓▓▓▓▓▓█░░░░░░░░░░░░░░█ 
▓▓▓▓▓▓█░░░░ PJB, ░░░░░░░█ 
 ▓▓▓▓▓▓█░░░░Add This░░░░█  
▓▓▓▓▓▓█░░░░Please!░░░░░█ 
▓▓▓▓▓▓█░░░░░░░░░░░░░░█ 
▓▓▓▓▓▓█████░░░░░░░░░█   
██████▀░░░░▀▀██████▀

PJB3005 (commented):
Definitely not a fan of the breaking changes introduced here. Would it be possible to add a new function instead that creates the window without showing?

wixoaGit:
> Definitely not a fan of the breaking changes introduced here. Would it be possible to add a new function instead that creates the window without showing?

Yea, splitting `Show()` into a `Create()` and `Show()` probably would have been better. I'll redo it like that.

wixoaGit:
The tests are passing on my machine

PR #5491, opened 2024-10-13
fix: EntityPrototypeView now only creates entities when the UI is open

Fildrance:
Fix for EntityPrototypeView creating entities on setting ProtoId. This was the problem that caused https://github.com/space-wizards/space-station-14/pull/32339 to fail tests - "crafting recepie ui" is connected to controller that is updating on server init with fresh list of recepies. And this update initializes UI even without opening it - which leads to creation of entities without actually deleting them (not in master, but in this menu translation PR).
Fixed following [this discussion](https://discord.com/channels/310555209753690112/770682801607278632/1294953837559545858).


Fildrance:
Tested new version of code with lathe. But I don't know where to test call to SetPrototype when window is already opened - thats some kind of update on-the-run...

Fildrance:
@metalgearsloth can u pls look at that?

PR #5495, opened 2024-10-16
Fix map netId completions

IProduceWidgets:
Map netId's completion helpers didn't work because empty strings fail NetEntity.TryParse. I special cased it.
Also made the helper pull all maps because it would be unhelpful if it didn't.

### DIFF
--- a/Robust.Shared/Console/CompletionHelper.cs
+++ b/Robust.Shared/Console/CompletionHelper.cs
@@ -188,7 +188,7 @@ public static class CompletionHelper
     {
         IoCManager.Resolve(ref entManager);
 
-        return Components<MapComponent>(string.Empty, entManager, limit: int.MaxValue); // faster just to pass the max value than bother unlimiting it.
+        return Components<MapComponent>(string.Empty, entManager, limit: 128);
     }
 
     /// <summary>
--- a/Robust.Shared/Console/CompletionHelper.cs
+++ b/Robust.Shared/Console/CompletionHelper.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


Robust.Shared/Console/CompletionHelper.cs
```diff
@@ -186,15 +186,17 @@ public static IEnumerable<CompletionOption> MapIds(IEntityManager? entManager =
 
     public static IEnumerable<CompletionOption> MapUids(IEntityManager? entManager = null)
     {
-        return Components<MapComponent>(string.Empty, entManager);
+        IoCManager.Resolve(ref entManager);
+
+        return Components<MapComponent>(string.Empty, entManager, limit: int.MaxValue); // faster just to pass the max value than bother unlimiting it.
     }
```
metalgearsloth: At least it make it like 100 or something if you want higher than 20.

IProduceWidgets: But if you have more than 100 maps it just wont work as expected.

ElectroJr: The limit only applies to matching uids, so as you start typing the uid it would still eventually show the completions for the map you want, even if there are far more than 20 maps. It only really matters if you really want to get given the completion tooltip with 100+ maps that you want to manually read through to find the right uid.

Though TBH I'm not even sure if having the limit implemented like that is really of any use. Even with it there, the helper should still only ever be used when you know there's only going to be a handful of components. Calling with with something like `TransformComponent` could still result in it iterating over almost all entities . E.g., in my toolshed PR I just arbitrarily limited it to `EntityManager.Count<TComp>() <= 128`

metalgearsloth:
```suggestion
        return Components<MapComponent>(string.Empty, entManager, limit: 128);
    }
```

PR #5496, opened 2024-10-16
add stylebox to multiline input

LetterN:
title. only tested with OpenDream, code is from `LineEdit`

### DIFF
--- a/Robust.Client/UserInterface/Controls/TextEdit.cs
+++ b/Robust.Client/UserInterface/Controls/TextEdit.cs
@@ -40,6 +40,9 @@ public sealed class TextEdit : Control
     [Dependency] private readonly IClipboardManager _clipboard = null!;
     [Dependency] private readonly IClyde _clyde = null!;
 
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/Controls/TextEdit.cs
```diff
@@ -784,15 +787,16 @@ private void AbortIme(bool delete = true)
...
 
         var renderBoxSize = _renderBox.Size;
-
         _scrollBar.Page = renderBoxSize.Y * UIScale;
 
         UpdateLineBreaks((int)(renderBoxSize.X * UIScale));
 
-        return size;
+        return finalSize;
     }
```
PJB3005, line 800: You're no longer arranging the vertical scroll bar here.

PJB3005 (changes requested):
Can't this just be done by wrapping the TextEdit in a PanelContainer?

LetterN:
got lazy will probably try that next time

PR #5498, opened 2024-10-17
Minor tileref optimisation

metalgearsloth:
Avoids creating it on every enumeration.


PJB3005 (commented):
It's not obvious to me this is faster, this is a struct type so it's not like there's an allocation happening here, and I actually expect the JIT to be able to inline this properly anyways.

PR #5499, opened 2024-10-18
Fix some window UIScale bugs

wixoaGit:
Windows were being created with a UIScale based on the window api's scaling hints. If this contradicted the `display.uiScale` cvar, it would be replaced by that the next time the window was resized. I changed it to respect the cvar from the beginning.

OSWindow also kept its `SetSize` up-to-date using its physical pixel size. If its UIScale was anything other than 1, this led to broken positioning and sizing of controls. Now it keeps UIScale in mind when updating its `SetSize`, and also multiplies its initial size by UIScale.

I also fixed a bug with `WindowReg.IsVisible` that caused it to be incorrect on window creation.

### DIFF
--- a/Robust.Client/UserInterface/UserInterfaceManager.Roots.cs
+++ b/Robust.Client/UserInterface/UserInterfaceManager.Roots.cs
@@ -25,14 +25,14 @@ internal sealed partial class UserInterfaceManager
             throw new ArgumentException("Window already has a UI root.");
         }
 
-        var cvarScale = _configurationManager.GetCVar(CVars.DisplayUIScale);
         var newRoot = new WindowRoot(window)
         {
             MouseFilter = Control.MouseFilterMode.Ignore,
             HorizontalAlignment = Control.HAlignment.Stretch,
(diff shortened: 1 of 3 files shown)


Robust.Client/UserInterface/UserInterfaceManager.Roots.cs
```diff
@@ -24,12 +25,13 @@ public WindowRoot CreateWindowRoot(IClydeWindow window)
...
         }
 
+        var cvarScale = _configurationManager.GetCVar(CVars.DisplayUIScale);
         var newRoot = new WindowRoot(window)
         {
             MouseFilter = Control.MouseFilterMode.Ignore,
             HorizontalAlignment = Control.HAlignment.Stretch,
             VerticalAlignment = Control.VAlignment.Stretch,
-            UIScaleSet = window.ContentScale.X
+            UIScaleSet = cvarScale == 0f ? window.ContentScale.X : cvarScale
```
PJB3005: This should probably be calculated with `CalculateAutoScale()` instead?

PJB3005 (changes requested):

PR #5503, opened 2024-10-26
BUI deferral tweaks

metalgearsloth:
- Removes the storage BUI bandaid.
- Defers opens now too so we can cleanly swap out BUIs on the same tick.


Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
```diff
@@ -1039,17 +1028,46 @@ public override void Update(float frameTime)
...
         if (_timing.IsFirstTimePredicted)
         {
-            foreach (var bui in _queuedCloses)
+            foreach (var (bui, open) in _queuedBuis)
             {
-                if (UIQuery.TryComp(bui.Owner, out var uiComp))
+                if (open)
                 {
-                    uiComp.ClientOpenInterfaces.Remove(bui.UiKey);
+                    bui.Open();
```
whatston3, line 1035: Should this Open call be in the try/catch block to handle any exception inside the opening?

Might be the cause of https://github.com/space-wizards/space-station-14/issues/35542

PR #5505, opened 2024-10-28
OnDropdownHide

ImRenQ:
Adds an event that allows the external code to know about the closing of the drop-down list

### DIFF
--- a/Robust.Client/UserInterface/Controls/OptionButton.cs
+++ b/Robust.Client/UserInterface/Controls/OptionButton.cs
@@ -47,6 +47,8 @@ namespace Robust.Client.UserInterface.Controls
 
         public event Action<ItemSelectedEventArgs>? OnItemSelected;
 
+        public event Action? OnDropdownHide;
+
         public string Prefix { get; set; } = string.Empty;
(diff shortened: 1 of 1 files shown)


PJB3005:
Just curious: what's your use case for this?

ImRenQ:
> Just curious: what's your use case for this?

I need to know when a popup is closed or open

I have a ListContainer, one of the entries contains an OptionButton. When the ListContainer is updated, the open popup is closed. 
Skip ListContainer updates if the popup is open (and updating after it closes) is one solution to this problem

изображение

PJB3005:
That use case would be better served by not rebuilding the entire control tree on update, which is more efficient anyways.

ImRenQ:
ListContainer can't work any other way, as far as I understand

PJB3005:
That is an issue with `ListContainer` that should be solved in content, and even then your use case does not seem appropriate for the control.

ImRenQ:
> That is an issue with `ListContainer` that should be solved in content, and even then your use case does not seem appropriate for the control.

Thanks for explaining this, I'll try to create a new "ListContainer"

PR #5507, opened 2024-10-31
Fix for server consoles without width and length

qrtDaniil:
Everything is described here https://github.com/pelican-eggs/games-standalone/issues/19. I test this fix in pterodactyl and it's work. Funny that `if (Con.CursorTop> 0)` cause strange behavior and bugs, but `if (Con.WindowWidth > 0)` works fine. Why? I don't know.

image


PJB3005:
It'd probably still be better if we had a system to just bail and disable all of `SystemConsoleManager` for non-TTY scenarios, but since this fixes the problem and is pretty small I think this is good enough.

PR #5514, opened 2024-11-09
Hopefully fix the dreaded VV refresh blink

SaphireLattice:
Shh, it's not like it's that much memory used up by this, right? It's just a list and some references. And it's just an admin tool, so whatever, not that general user impactful, but massive QoL for server owners

Stopped the blinking for me on Debug client & Tools server. Didn't test on real network conditions but probably will work


PJB3005:
Ok GitHub is broken,  Itried to close and re-open the PR to fix, now I can't even re-open the PR. 😐

PR #5522, opened 2024-11-15
Add FreeBSD packaging target

Partmedia:
- Adds a FreeBSD engine client build packaging target that can be selected by `--platform freebsd`. This is based on the Linux packaging function.
- Add FreeBSD to the default list of targets built if no `--platform` is selected


deltanedas:
this changes linux-x64 to just linux

Partmedia:
> this changes linux-x64 to just linux

Did you test this? I just tested packaging with `--platform linux` and it produces a `Robust.Client_linux-x64.zip` build.

Perhaps you missed that it's `rid` that's passed to `publish_client()` and not `platform`?

deltanedas:
ah thats it yeah

PR #5524, opened 2024-11-18
Add FormattedMessage.TrimEnd

SpaceManiac:
Add a utility function that can be used to remove accidentally-added trailing newlines after the rest of the message has been composed. Useful to avoid the blank line pop-in when examining things in SS14.

This is a little more final than trying to track down and stamp out every source of trailing newlines forever.

This has to be here because write access to `_nodes` is private.

### DIFF
--- a/Robust.Shared/Utility/FormattedMessage.cs
+++ b/Robust.Shared/Utility/FormattedMessage.cs
@@ -183,31 +183,27 @@ public sealed partial class FormattedMessage : IReadOnlyList<MarkupNode>
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


PJB3005 (commented):
Any reason this doesn't just trim all trailing whitespace altogether? Otherwise looks good.

SpaceManiac:
Newlines were the thing causing the problem. Could extend to all whitespace if that's what you prefer.

PJB3005:
Feels like it'd be more generally useful to just handle all whitespace so sure.

PR #5532, opened 2024-11-22
Fix warnings for using async without any await

CrafterKolyan:
Fixes 4 compiler warnings CS1998 for defining async method that doesn't use any await statements


PJB3005:
Your code doesn't compile. This is why you shouldn't web edit.

CrafterKolyan:
> Your code doesn't compile. This is why you shouldn't web edit.

Agree that I shouldn't do that, but right now I was having a set up where I have only a space-station-14 repo cloned. And I just edit everything there. I guess I should stop using this approach as I start to edit more and more files. Also I guess I put extra load on verification from your side and you will become less confident in my commits.

I will just do it properly

CrafterKolyan:
Ah, sorry, I understood that I simply should've invested extra 10 minutes in understanding and getting proper set up with `git submodule`. 
I fixed the issue and was able to make a set up that is both comfortable to me and is in line with the programming standards

PR #5533, opened 2024-11-22
Make font drawing more generic

amylizzle:
gonna steal this for OD's maptext implementation

it'd be helpful if I could draw it to `DrawingWorldHandle`

### DIFF
--- a/Robust.Client/Graphics/Font.cs
+++ b/Robust.Client/Graphics/Font.cs
@@ -2,7 +2,6 @@
 using System.Numerics;
 using System.Text;
 using Robust.Client.ResourceManagement;
-using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
 using Robust.Shared.Maths;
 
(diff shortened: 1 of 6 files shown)


PJB3005:
Does this actually work? Screen and world have inverted coordinate spaces, I can't imagine how the font renderer would just work with minimal changes like this.

amylizzle:
Yup 
image

PJB3005:
Uh, what happens if you try to render text with a period? That `!` looks like it's positioned wrong...

amylizzle:
image

Yeah it's a little offset from where it ought to be. I did have to apply the character metric's offset again in OD's content-side code.

Robust.Client/Graphics/Font.cs
```diff
@@ -132,7 +132,10 @@ public override float DrawChar(DrawingHandleScreen handle, Rune rune, Vector2 ba
             }
 
             baseline += new Vector2(metrics.Value.BearingX, -metrics.Value.BearingY);
-            handle.DrawTexture(texture, baseline, color);
+            if(handle is DrawingHandleWorld worldhandle)
+                worldhandle.DrawTextureRect(texture,  Box2.FromDimensions(baseline, texture.Size));
+            else
+                handle.DrawTexture(texture, baseline, color);
```
PJB3005, line 138: Really not a fan of this type check given that this is hot path code.

amylizzle: best alternative I can think of is to split `DrawChar()` into two function signatures, one with `DrawingHandleWorld` and one with `DrawingHandleScreen`. Bit code-duplicate-y but shouldn't be a slowdown.

PR #5548, opened 2024-12-03
Replace obsolete TransformComponent.Anchored set

MilenVolf:
For [#33279](https://github.com/space-wizards/space-station-14/issues/33279)

This PR also adds new condition to the AnchorEntity for case when MapGridComponent is null.


PJB3005:
Bro got got by the exact same thing that made me skip out on fixing these warnings 
image

metalgearsloth:
AnchorEntity is not a drop-in replacement because the anchored setter does the slow-path of tryfindgridat.

MilenVolf:
> AnchorEntity is not a drop-in replacement because the anchored setter does the slow-path of tryfindgridat.

I'll add additional condition then to keep old behaviour for cases like this

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -124,8 +124,13 @@ public bool AnchorEntity(Entity<TransformComponent> entity, Entity<MapGridCompon
...
             if (!TryComp(entity.Comp.GridUid, out MapGridComponent? gridComp))
-                return false;
-            grid = (entity.Comp.GridUid.Value, gridComp);
+            {
+                if (!_mapManager.TryFindGridAt(GetMapCoordinates(entity.Comp), out var gridUid, out var gridMapComp))
+                    return false;
+                grid = (gridUid, gridMapComp);
+            }
+            else
+                grid = (entity.Comp.GridUid.Value, gridComp);
```
metalgearsloth, line 133: This one I'm a bit suss on because it will fallback to the slowpath where previously it would just return false. Ideally this shouldn't even be happening because the entity itself should already have a griduid.

MilenVolf:
Superseded by #5681

PR #5549, opened 2024-12-05
Fix memory leak in PlayGlobal when filter is empty

SolStar2:
Short circuits when the player filter is empty before the SetupAudio entity is made.

### DIFF
--- a/Robust.Server/Audio/AudioSystem.cs
+++ b/Robust.Server/Audio/AudioSystem.cs
@@ -83,7 +83,7 @@ public sealed partial class AudioSystem : SharedAudioSystem
     /// <inheritdoc />
     public override (EntityUid Entity, AudioComponent Component)? PlayGlobal(string? filename, Filter playerFilter, bool recordReplay, AudioParams? audioParams = null)
     {
-        if (string.IsNullOrEmpty(filename))
+        if (string.IsNullOrEmpty(filename) || playerFilter.Count == 0)
             return null;
 
         var entity = SetupAudio(filename, audioParams);


PJB3005 (changes requested):
Surely you should add this to the other APIs that are running `AddAudioFilter` too?

Also, can you elaborate on how this fixes a memory leak? The change seems fine to me on its own (is it possibly any code can freak out at the sudden null return from this function?), but I'm not sure what memory leak this fixes.

SolStar2:
Is was at least indicated to be the case in the entity integration tests. I didn't test if this also happens in live environment since its a bit harder to test for, it may just have to do with the integration test itself.

SolStar2:
This is the pr that caused this behavior to arise. https://github.com/DeltaV-Station/Delta-v/pull/2351 Locally running with this change fixes the test failure.

slarticodefast:
Could it be that this PR also fixes sounds like the WEWLAD, zombie infection or other antag specific ones that use PlayGlobal being heard for every player in replays? They are restricted to a certain player session but that does not seem to get filtered in replays correctly.

PJB3005:
> Could it be that this PR also fixes sounds like the WEWLAD, zombie infection or other antag specific ones that use PlayGlobal being heard for every player in replays? They are restricted to a certain player session but that does not seem to get filtered in replays correctly.

Unlikely.

PJB3005:
> This is the pr that caused this behavior to arise. [DeltaV-Station/Delta-v#2351](https://github.com/DeltaV-Station/Delta-v/pull/2351) Locally running with this change fixes the test failure.

The problem here seems to be that `SpawnAndDeleteEntityCountTest` should ignore `Audio` components. Paging @ElectroJr to see if that's valid

SolStar2:
Do you guys still want the change even if its not fixing a memory leak? Or should I close?

ElectroJr:
> Do you guys still want the change even if its not fixing a memory leak? Or should I close?

I agree with PJB that if the change is made, it should also be made to the other audio methods for consistency (e.g., the `PlayEntity()` method directly beneath the one you changed).

> The problem here seems to be that SpawnAndDeleteEntityCountTest should ignore Audio components. Paging @ElectroJr to see if that's valid

The test was created before audio-entities existed and was never updated. I guess it should be updated to ignore all audio entities. Though even with that change the test is still pretty janky and I'm not sure if it should exist. It was a crude test meant to try catch some problematic entities that accidentally dump other entities on the floor or spawn null-space ents that aren't cleaned up when the entity that spawned them gets deleted, but there are always going to be quite a few entities where the test just isn't applicable.

PJB3005:
Stale.

PR #5551, opened 2024-12-06
Changes SharedGridTraversalSystem accesibility from internal to public

MLGTASTICa:
Lets people disable/enable this system at will with this change.  Done mainly because this prevents anyone from implementing custom parenting/unparenting behaviour and limits everyone to grid-parenting preety much.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedGridTraversalSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedGridTraversalSystem.cs
@@ -21,6 +21,10 @@ public sealed class SharedGridTraversalSystem : EntitySystem
 
     /// <summary>
     /// Enables or disables changing grid / map uid upon moving.
+    /// WARNING: If you do this in a live-game. You need to make sure that the parented entity
+    /// doesn't move too far away from the grid. As it will cause Entity Lookups to not see it
+    /// (because the grid its parented to is not close enough and all parented entities are assumed
+    /// to be on the grid through the broadphase component)
     /// </summary>
     public bool Enabled = true;
 
--- a/Robust.Shared/GameObjects/Systems/SharedGridTraversalSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedGridTraversalSystem.cs
@@ -11,7 +11,7 @@ namespace Robust.Shared.GameObjects;
 /// <summary>
 ///     Handles moving entities between grids as they move around.
 /// </summary>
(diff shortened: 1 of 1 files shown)


metalgearsloth:
Are they going to manage broadphases manually too

MLGTASTICa:
> Are they going to manage broadphases manually too

Probably not.

ElectroJr:
In case it isn't clear, AFAIK the issue is that if an entity is far away from the grid, but still parented to it, then it will still be part of the entity's broadphase. Any system that tries to query entities in an area (physics, lighting, rendering, entity lookups, etc) first gets a list of grids in the area, then queries each of those grids. So if an entity is still parented to a grid despite being far enough away, from the grid, it will no longer be returned by queries and wont get rendered or whatever else is being done by the query.

So unless there's some system that handles expanding a grids "size" to account for entities that are parented to the grid despite not being on the grid (which I don't know if we'd even want to bother supporting), we kind of need any entity that moves off the grid to automatically get parented to the map.

I get that the system can be a PITA and it might be nice to have more control (e.g., the bool exists because it can be a PITA for tests). But if people just start completely disabling it they will probably just unknowingly introduce bugs. But if people are aware of the issues it will cause and still want to be able to disable it I think its fine to make it public, though there should be comments in the code explaining the issues.

MLGTASTICa:
> In case it isn't clear, AFAIK the issue is that if an entity is far away from the grid, but still parented to it, then it will still be part of the entity's broadphase. Any system that tries to query entities in an area (physics, lighting, rendering, entity lookups, etc) first gets a list of grids in the area, then queries each of those grids. So if an entity is still parented to a grid despite being far enough away, from the grid, it will no longer be returned by queries and wont get rendered or whatever else is being done by the query.
> 
> So unless there's some system that handles expanding a grids "size" to account for entities that are parented to the grid despite not being on the grid (which I don't know if we'd even want to bother supporting), we kind of need any entity that moves off the grid to automatically get parented to the map.
> 
> I get that the system can be a PITA and it might be nice to have more control (e.g., the bool exists because it can be a PITA for tests). But if people just start completely disabling it they will probably just unknowingly introduce bugs. But if people are aware of the issues it will cause and still want to be able to disable it I think its fine to make it public, though there should be comments in the code explaining the issues.

Thank you for explaining the issue. I thought that entity lookups used map chunks to check for entity positions. Ill add a comment specifying the issue.

ElectroJr (approved):

PR #5554, opened 2024-12-08
Audio mixers

### LINKED ISSUE #5462: Need audio / sound categories
RobustToolbox should have some way categorizing audio sources, and some way to modify the volume per category. I.e.,  we should have separate audio sliders.

I was sure we already had an issue for this, but couldn't find it, so this might be a duplicate?

stalengd:
## What
Brings audio mixers functionality to the engine similar how they works in unity or generally any other software I believe. Thus: 
- The audio source can be assigned to the mixer, which will modify its parameters, only gain at this moment. 
- Mixers can be also hooked up to one another to create hierarchy and modify each other's output. 
- Mixer's gain can be loaded and saved to CVar.
- Mixers can be configured from prototypes.
- Custom mixers can be created at runtime from client or server.
- Both entity-based and plain object mixers for places where entities are a no-go (e.g. user interface).

## Why
Fixes #5462 and https://github.com/space-wizards/space-station-14/issues/22710, corresponding PR to SS14 will fix some other issues there, generally a must-have tool for the engine.

### Underlying tech
This audio mixers implementation does not use OpenAL auxiliary slots (if they even can be used to implement this) or any other fancy low-level stuff. Instead, simple mechanism of storing references in mixers and sources to each other is used, and when gain of any mixer is changed, all connected mixers and sources modify their gain value (see `AudioMixer.cs`). However, I see no problems to seamlessly re-implement internal work of audio mixers to use OpenAL auxiliary later, this will need to be done to support the audio mixer effects as well.

### MixableAudioSource
I did not dare to put mixers support into AudioSource, since it is a rather thin interface with OpenAL, so the MixableAudioSource wrapper is used to support audio mixers. AudioComponent has now reference to IMixableAudioSource, not IAudioSource.

### Audio mixer prototypes
This is a way to have "audio categories", game can define any number of mixer prototypes and use their IDs to assign corresponding mixer to audio sources. This replaces dubious pattern of handling subscriptions to CVars for every source and for the most cases this is nearly impossible to do manually. "Prototyped" audio mixers can be assigned via AudioParams, so it is really easy to manage. There are also a problem of many sources not having any mixers assigned at the moment in SS14, but customizable default mixer can be handy in this case.

### Plain object mixers
AudioMixer is a plain object and AudioSource counterpart, so exists only on client side and managed by AudioMixersManager. Manager creates custom mixers and stores mixers from prototypes, so only one AudioMixer instance corresponding to any AudioMixerPrototype id can exist.

### Entity-based mixers
AudioMixerComponent is a counterpart of AudioComponent and they are used together. Mixer component on client obtains reference to plain object mixer and system will assign this mixer to internal audio source component when needed. Mixer component fully supports synchronization, so can be assigned to audio sources on the server side, Mixer components are spawned on separate global entities.

### Entity-based mixers from prototypes problem
The concept of "prototyped" mixers conflicts with entity-based mixers in terms of lifetimes. 

My first implementation of this was lazily creating mixers for prototype when requested, newly created mixer then stored in dictionary indefinitely. Unfortunately, some SS14 integration tests does not like this idea when any entity creates more other entities than deletes after got deleted.

This implementation creates all mixer entities from prototypes at start of the simulation (check out the funny way how system knows when its time to spawn entities) and when entities flushed. Set of such entities exists separately on client and server though because single-player games may exist and (probably) entities from server may not be sent to the client fast enough. The problem with this implementation is that other SS14 integration test (DeleteAllThenGhost this time) can't handle that deleting all entities on the server leaves mixer entities on client.

So this problem is relevant and PR probably should not be merged like this, either mixers should be implemented differently (but I don't know how at this point) or SS14 integration tests should be adjusted.

## Discussion
As you can see, there are plenty of debatable decisions in this PR, so maintainers may fell free to discuss and request changes to implementation here, or for quicker communication in discord (`stalengd`, DM or ping).

## PR to SS14 repo
https://github.com/space-wizards/space-station-14/pull/33816


Robust.Client/Audio/Sources/MixableAudioSource.cs
```diff
@@ -0,0 +1,172 @@
...
+{
+    [ViewVariables(VVAccess.ReadOnly)]
+    private readonly IAudioSource _innerSource;
+    [ViewVariables(VVAccess.ReadOnly)]
+    private IAudioMixer? _mixer;
+    [ViewVariables]
+    private float _selfGain = 1f;
+
+    private bool _isDisposed = false;
+
```
metalgearsloth, line 20: Better to just set a mix on the audio source itself than wrapping another one.

Robust.Client/Audio/AudioSystem.Mixers.cs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    private void DisposeMixer(IAudioMixer mixer)
+    {
+        // We don't want to dispose mixers from prototypes cos they are supposed to be re-used.
+        if (mixer.ProtoId is { })
+            return;
+        mixer.Dispose();
+    }
+}
```
metalgearsloth, line 97: This hidden behavior is unintuitive to use, would be better to just use prototypes as initialisation for mixer settings / backing CVars and nothing more.

Robust.Server/Audio/AudioSystem.Mixers.cs
```diff
@@ -0,0 +1,22 @@
...
+        return mixerEntity;
+    }
+
+    public override Entity<AudioMixerComponent> CreateMixer(Entity<AudioMixerComponent>? outMixer)
+    {
+        var mixerEntity = base.CreateMixer(outMixer);
+        _pvs.AddGlobalOverride(mixerEntity);
+        return mixerEntity;
+    }
+}
```
metalgearsloth, line 22: Aren't the mixer CVar values clientside? I don't think we need to be networking them as global entities which add overhead to PVS.

Robust.Shared/Audio/Components/AudioComponent.cs
```diff
@@ -70,14 +71,20 @@ public sealed partial class AudioComponent : Component, IAudioSource
...
     /// </summary>
     [DataField, AutoNetworkedField]
     public EntityUid? Auxiliary;
 
+    /// <summary>
+    /// Audio mixer entity to pass audio to.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public EntityUid? Mixer;
+
```
metalgearsloth, line 87: The existing API for effects was not good and better off just making it its own non-networked class.

Robust.Shared/Audio/Components/AudioMixerComponent.cs
```diff
@@ -0,0 +1,81 @@
...
+using Robust.Shared.Serialization;
+using Robust.Shared.ViewVariables;
+
+namespace Robust.Shared.Audio.Components;
+
+[RegisterComponent, NetworkedComponent, Access(typeof(SharedAudioSystem))]
+public sealed partial class AudioMixerComponent : Component, IAudioMixer
+{
+    public IAudioMixer? Out => Mixer.Out;
+
```
metalgearsloth, line 16: None of this code is documented and as above would be better as its own class than trying to hook it in awkwardly.

Robust.Shared/Audio/Mixers/AudioMixerPrototype.cs
```diff
@@ -0,0 +1,43 @@
...
+using Robust.Shared.Prototypes;
+using Robust.Shared.Serialization.Manager.Attributes;
+
+namespace Robust.Shared.Audio.Mixers;
+
+/// <summary>
+/// Preset for creating <see cref="IAudioMixer"/>s from.
+/// </summary>
+[Prototype("audioMixer")]
+public sealed class AudioMixerPrototype : IPrototype
```
metalgearsloth, line 11: Prototype strings are obsolete.

Robust.Shared/Audio/Systems/SharedAudioSystem.Mixers.cs
```diff
@@ -0,0 +1,234 @@
...
+    private readonly Dictionary<ProtoId<AudioMixerPrototype>, Entity<AudioMixerComponent>> _audioMixers = new();
+    private bool _isMixersStarted;
+
+    protected virtual void InitializeMixers()
+    {
+        SubscribeLocalEvent<AudioMixerComponent, ComponentShutdown>(OnMixerShutdown);
+        SubscribeLocalEvent<AudioMixerComponent, ComponentGetState>(OnGetState);
+        SubscribeLocalEvent<AudioMixerComponent, ComponentHandleState>(OnHandleState);
+        EntityManager.AfterEntityFlush += LoadPrototypedMixers;
+    }
```
metalgearsloth, line 28: Making them entities really makes a lot of this handling awkward, would be better to just make them their own class.

metalgearsloth (changes requested):
I've taken another look at it.

1. I don't think we should store mixers on audioparams because if we have 20 different audio params feeding into say, footstep sounds, then it's impossible for callers to track that. Wherever is playing the sound should be managing it (and setting it on the audio entity directly).
2. All of the new code needs documenting.
3. We will get multiple gain sources at some point (e.g. fades, occlusion at some point) so mixer should be another property on the source to make it easier to handle mixing.

I get the idea of trying to make mixers entities to allow for hot reloading / prototypes dynamically doing it but there's no easy to combine audioparams on soundpaths with audioparams when the sound is actually being played, meaning the mix needs to be manually specified on every prototype that may be relevant. It's better to just manually specify it on the audiocomponent (which in turn feeds the IAudioSource) or add code to allow combining audioparams cleanly.

Snorkcom:
> I've taken another look at it.
> 
> 1. I don't think we should store mixers on audioparams because if we have 20 different audio params feeding into say, footstep sounds, then it's impossible for callers to track that. Wherever is playing the sound should be managing it (and setting it on the audio entity directly).
> 2. All of the new code needs documenting.
> 3. We will get multiple gain sources at some point (e.g. fades, occlusion at some point) so mixer should be another property on the source to make it easier to handle mixing.
> 
> I get the idea of trying to make mixers entities to allow for hot reloading / prototypes dynamically doing it but there's no easy to combine audioparams on soundpaths with audioparams when the sound is actually being played, meaning the mix needs to be manually specified on every prototype that may be relevant. It's better to just manually specify it on the audiocomponent (which in turn feeds the IAudioSource) or add code to allow combining audioparams cleanly.

I really wish we had a volume setting for sound effects already. Our ears are dying from the chief engineer’s tools and stuff like that.

metalgearsloth:
Re-open if you come back to it.

PR #5556, opened 2024-12-11
Monospace Support for Rich Text

southbridge-fur:
This adds the new `[mono]` tag to the RichText system allowing for monospaced text in player-created paper documents.

## Technical Details 

Created the new [MonoTag.cs](Robust.Client/UserInterface/RichText/MonoTag.cs) class which handles the [mono] tag when it appears in rich text documents. This loads the Monospace font described in the [Prototypes/fonts.yml](https://github.com/space-wizards/space-station-14/tree/67d444feb0b47b63a54221a7124e2277a214972f/Resources/Prototypes/fonts.yml) file and pushes it onto the MarkupDrawingContext object. 

Also added the `MonoTag` entry to the [MarkupTagManager.cs](Robust.Client/UserInterface/RichText/MarkupTagManager.cs) list of valid tags. Forks will need to also need to add this tag to the list of their [PaperWindow.xaml.cs](https://github.com/space-wizards/space-station-14/blob/67d444feb0b47b63a54221a7124e2277a214972f/Content.Client/Paper/UI/PaperWindow.xaml.cs#L40) `_allowedTags` attribute. However, I will be opening a PR on Wizden to account for this so it should propagate neatly. 

## Breaking Changes

If a branch does not have a monospace font defined in its `Resources/Prototypes/fonts.yml` and if someone tries to use the `[mono]` tag in a text document it will likely cause a huge issue. That being said this isn't something that seems to be handled in the similar BoldTag.cs or ItalicTag.cs so I'm gonna assume it's fine. 

## Demonstration
Using the Wizden SS14 debug build.

https://github.com/user-attachments/assets/e01cbc4c-219e-4dec-a6d0-9f9ef9bf3f62

### DIFF
--- a/Robust.Client/UserInterface/RichText/MarkupTagManager.cs
+++ b/Robust.Client/UserInterface/RichText/MarkupTagManager.cs
@@ -24,7 +24,8 @@ public sealed class MarkupTagManager
         new CommandLinkTag(),
         new FontTag(),
         new HeadingTag(),
-        new ItalicTag()
+        new ItalicTag(),
+        new MonoTag()
     }.ToDictionary(x => x.Name.ToLower(), x => x);
 
     /// <summary>
@@ -39,7 +40,8 @@ public sealed class MarkupTagManager
         typeof(CommandLinkTag),
         typeof(FontTag),
         typeof(HeadingTag),
-        typeof(ItalicTag)
+        typeof(ItalicTag),
+        typeof(MonoTag)
     };
 
     public void Initialize()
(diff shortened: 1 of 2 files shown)


PJB3005 (changes requested):
Can't this be implemented in SS14? I don't think this needs to be part of Robust.

southbridge-fur:
This is where all the other elements of rich text are implemented, I assumed this is where it had to go.

southbridge-fur:
ahh it can be implemented there, then I'll do that.a

PR #5559, opened 2024-12-13
Add InterfaceData constructor

poklj:
### What
Add an InterfaceData constructor to UserInterfaceComponent's InterfaceData to allow for dynamic setting using setUi
### Why
Allows Component Systems to create BoundUi's dynamically (By ensuring a `UserInterfaceComponent` onto the entity and then utilizing `SharedUserInterfaceSystem.setUi()` to set the interface)  and be able to receive messages from them without needing a separate entity and a relay.


slarticodefast:
To clarify why this is needed:
At the moment a BUI can only be added to a prototype before spawning by having the serializer create the `InterfaceData` object in the `UserInterfaceComponent`:
grafik

Without the constructor it is not possible to add a BUI to an entity after spawning.

Issue #5306, opened 2024-07-16
MacOS M1 OpenAL error with generating buffer


Sh18RW:
# Description
I already have been written about that on forum.

While gaming after some time (I think buffers stack overflow) `AL.GenBuffer` starts generate InvalidValue error and returns 0 new buffer's index, it causing new errors because engine can't upload audio.
After debugging I noticed that nowhere calls `AL.DelBuffer`, but source freeing calling. I think, it is a clue.
Maybe TTSless servers devoid of this problem because there are fewer audio upload calls. But I think, it needs changes in engine.

https://forum.spacestation14.com/t/macos-open-al-al-genbuffer-returns-0-null-after-some-time/9840

Sh18RW:
I tried to find a cause. My guess turned out to be true: default MacOS's realisation of OpenAL has the limit is 1023 buffers. Engine doesn't free up buffers, because of it OpenAL after a lot of created buffers can't continue to do it.
How I got this? I made a test. I wrote a program with C++, firstly I used default OpenAL
![image](https://github.com/user-attachments/assets/f0cdc7f3-04e2-4b1b-9d18-9c3134b7d136)
After I used OpenAL Soft (Installed with brew):
![image](https://github.com/user-attachments/assets/580c9931-412a-421a-8011-93ccf551b404)
There is the limit as you can see.
The problem has two ways to solve it: freeing up buffers or use OpenAL Soft solution. But both solutions have their own problems. It needs rewriting anyway. I didn't find how to make OpenTK.OpenAL to use OpenAL Soft.

Sh18RW:
but as an alternative, use Silk.NET.OpenAL.Soft instead OpenTK

PJB3005:
> but as an alternative, use Silk.NET.OpenAL.Soft instead OpenTK

This is not that easy

Blocking this on #4997

Sh18RW:
In this case, I suggested a fix. It will be the simplest solution at the moment.

HoofedEar:
@Sh18RW Just curious, does the PR you made (#5735) significantly reduce the number of crashes on M1 macOS or is it not at all relevant to this macOS issue?

Sh18RW:
> Just curious, does the PR you made ([#5735](https://github.com/space-wizards/RobustToolbox/pull/5735)) significantly reduce the number of crashes on M1 macOS or is it not at all relevant to this macOS issue?

[...]

PR #5562, opened 2024-12-17
Audio Buffers Overflowing Fix

Sh18RW:
This PR solves https://github.com/space-wizards/RobustToolbox/issues/5306 issue.
I'm not sure what the implementation should be.

The main problem is that we cannot be sure that the buffer is never used anywhere else. For this reason, in order to ensure the safety of the buffer, I have decided to save a values array and upload it to the buffer when it is used. But there is another problem. We can upload audio files multiple times, and each time we save the audio to the memory. Maybe we should have another dictionary to save "string key as name and IClydeHandle as value" and audio files with the same name should be saved only once. However, if the name is the same, it does not tell us that the content is the same. In this case, we can replace older content by new and give a work to GC.

Maybe it should be modified, but I need your advice.
This solution has been tested with Space Station 14.


AverageNotDoingAnythingEnjoyer:
There is a high possibility this will fix all the `OpenAL: invalid value` errors which tend to happen whenever any in-game sound plays on Mac and tend to increase over time in their amount.
Also, there's a chance this might fix constant crashes, but I'm not sure about that.
Blessed PR.

metalgearsloth:
Similar to https://github.com/space-wizards/RobustToolbox/pull/5443/files though this one has significantly more structural changes.

Sh18RW:
> Similar to https://github.com/space-wizards/RobustToolbox/pull/5443/files though this one has significantly more structural changes.

Does your solution free up buffers? I have tried to do using the `Dispose` method in the `AudioStream`, but it has not worked. But your solution looks common so If it works then I will close my PR.

Edited: I have tested with Space Station 14 your changes, `Dispose` method never calls on `AudioStream`...

Sh18RW:
I am still waiting...

AverageNotDoingAnythingEnjoyer:
> Similar to https://github.com/space-wizards/RobustToolbox/pull/5443/files though this one has significantly more structural changes.

Tbf this one passes all the checks as of now, so maybe better use this one first?

Sh18RW:
Dumb PR.
#5735

PR #5567, opened 2024-12-22
Augment audio system with the ability to have captions 

sowelipililimute:
Mainly manages `CaptionComponent`s attached to the audio entities whenever they should display a visible caption. Can't use a `Caption` field on the `AudioComponent` due to the `ComponentTreeSystem` wanting to listen to the `ComponentStartup`/`ComponentShutdown` events that are already being listened to by the client `AudioSystem`.

### DIFF
--- a/Robust.Client/Audio/AudioSystem.cs
+++ b/Robust.Client/Audio/AudioSystem.cs
@@ -44,7 +44,6 @@ public sealed partial class AudioSystem : SharedAudioSystem
     [Dependency] private readonly SharedMapSystem _maps = default!;
     [Dependency] private readonly SharedTransformSystem _xformSys = default!;
     [Dependency] private readonly SharedPhysicsSystem _physics = default!;
-    [Dependency] private readonly IEntityManager _entityManager = default!;
 
     public event Action<Entity<CaptionComponent, TransformComponent>>? OnSubtitledAudioStart;
     public event Action<Entity<CaptionComponent, TransformComponent>>? OnSubtitledAudioEnd;
@@ -59,6 +58,7 @@ public sealed partial class AudioSystem : SharedAudioSystem
     private float _audioFrameTimeRemaining;
 
     private EntityQuery<PhysicsComponent> _physicsQuery;
+    private EntityQuery<CaptionComponent> _captionQuery;
 
     private float _maxRayLength;
     private float _zOffset;
@@ -102,6 +102,7 @@ public sealed partial class AudioSystem : SharedAudioSystem
         UpdatesAfter.Add(typeof(EyeSystem));
 
         _physicsQuery = GetEntityQuery<PhysicsComponent>();
+        _captionQuery = GetEntityQuery<CaptionComponent>();
 
         SubscribeLocalEvent<AudioComponent, ComponentStartup>(OnAudioStartup);
         SubscribeLocalEvent<AudioComponent, ComponentShutdown>(OnAudioShutdown);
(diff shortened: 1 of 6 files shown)


Robust.Client/Audio/AudioSystem.cs
```diff
@@ -44,6 +44,10 @@ public sealed partial class AudioSystem : SharedAudioSystem
     [Dependency] private readonly SharedMapSystem _maps = default!;
     [Dependency] private readonly SharedTransformSystem _xformSys = default!;
     [Dependency] private readonly SharedPhysicsSystem _physics = default!;
+    [Dependency] private readonly IEntityManager _entityManager = default!;
+
```
metalgearsloth: Already a dependency on systems.

Robust.Client/Audio/AudioSystem.cs
```diff
@@ -192,17 +196,23 @@ public void SetMasterVolume(float value)
 
     private void OnAudioPaused(EntityUid uid, AudioComponent component, ref EntityPausedEvent args)
     {
+        if (_entityManager.TryGetComponent(uid, out CaptionComponent? caption) && _entityManager.TryGetComponent<TransformComponent>(uid, out var xform))
```
metalgearsloth: Use the other trygetcomponent as it has a direct xformcomp overload.
Also add EntityQuery<CaptionComponent> as well.

Robust.Client/ComponentTrees/CaptionTreeSystem.cs
```diff
@@ -0,0 +1,46 @@
...
+using Robust.Shared.GameObjects;
+using Robust.Shared.Maths;
+using Robust.Shared.Physics;
+using Robust.Shared.Audio.Components;
+using Robust.Shared.IoC;
+
+namespace Robust.Client.ComponentTrees;
+
+public sealed class CaptionTreeSystem : ComponentTreeSystem<CaptionTreeComponent, CaptionComponent>
+{
```
metalgearsloth: Needs doc comments explaining please.

Robust.Client/ComponentTrees/CaptionTreeSystem.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        if (_entityManager.TryGetComponent(entry.Uid, out AudioComponent? audio))
+        {
+            var radius = audio.MaxDistance;
+            var radiusVec = new Vector2(radius, radius);
+            return new Box2(pos - radiusVec, pos + radiusVec);
+        }
+        return default;
+    }
+
```
metalgearsloth: Same here on entmanager + trygetcomp.

metalgearsloth (changes requested):

Robust.Client/Audio/AudioSystem.cs
```diff
@@ -192,17 +197,23 @@ public void SetMasterVolume(float value)
 
     private void OnAudioPaused(EntityUid uid, AudioComponent component, ref EntityPausedEvent args)
     {
+        if (_captionQuery.TryGetComponent(uid, out var caption) && EntityManager.TryGetComponent(uid, out TransformComponent? xform))
```
slarticodefast, line 200: No need to do a TryComp for the TransformComponent, all entities have that component.
And if it was deleted the first one would fail.
So you can just use the `Transform(uid)` shorthand, which uses a query

Robust.Shared/Audio/Components/CaptionComponent.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Robust.Shared.Audio.Components;
+
+/// <summary>
+/// Stores the caption data for an audio entity.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState(true)]
+public sealed partial class CaptionComponent : Component, IComponentTreeEntry<CaptionComponent>
+{
+    [DataField, AutoNetworkedField]
+    public LocId? Caption { get; set; }
```
slarticodefast, line 18: add documentation to all of these

Robust.Shared/Audio/SoundCollectionPrototype.cs
```diff
@@ -13,6 +14,9 @@ public sealed partial class SoundCollectionPrototype : IPrototype
     [IdDataField]
     public string ID { get; private set; } = default!;
 
+    [DataField]
+    public LocId? Caption { get; private set; }
```
slarticodefast, line 18: documentation

PR #5571, opened 2024-12-23
Try fix entity deserialization component composition

ElectroJr:
This PR removes complex inheritance pushing when deserializing entities. This is meant to try and fix a bug where deserialization of entities with components/data-fields with the AlwaysPushInheritance would insert unwanted prototype data into loaded entities. 

E.g., this was causing the action grant component to gain an ever growing number of actions in the content test map anytime it was loaded & saved.
chrome_QQ1PR8hIDS

This also tries to prevent some possible issues where inheritance pushing wasn't cloning some fields, which might've lead to accidentally modifying a parent's data.


ElectroJr:
I just merged this into #5572

PR #5574, opened 2024-12-23
Illumination measuring system to determine how "lit up" entities are

Chubbygummibear:
Please bear with me since this is my first large PR and like 2nd PR ever for SS14/RT, so I expect I made a lot of mistakes that will need to be rectified upon review.

# Overview
This is an entirely new system that is designed to serve as a tool for content systems that would like to know the illumination level of a given entity. This is achieved by querying a PointLightComponent Tree on the server to determine which lights are nearby, then uses a raycast query of the Occluder Tree System to check for occlusion. Any lights after that which are both in range and unoccluded have their light power multiplied by an attenuation function to get the final light value applied to the entity. Only the most powerful light value on the entity is kept. That value is cached in the corresponding component and returned upon calling the TryGetLightLevel function.

The system does nothing on its own and needs to be triggered by other systems calling TryGetLightLevel which will cause the light calculations to occur. There is a cooldown of 0.5 Seconds on getting the light level re-calculated unless a boolean to force the update is called. If many calls are made to the function in a short time span, the cached value will be returned without rerunning the calculation. This way, if numerous systems call for the light level in the same tick, and every tick. The most expensive operation will not be called every time.

# Demonstration

https://github.com/user-attachments/assets/f04bca65-624b-4e4f-8698-29f814449d6c

It does work with masked lights as well, although i had to refactor the way light mask prototypes work to accomplish this.

# Limitations and concerns
- Naturally performance is a concern, as this is using raycasting frequently to check for occlusion. I did my best to mitigate the impact by implementing a cooldown for how often the system should be allowed to do so, as well as implementing the ability to define cooldowns on a per system that calls this system basis.
- Animations is a big one. With animations like spinning lights, flickering lights, and dimming areas being handled entirely on the client, there's no way to access this effectively on the server to know that the lights have changed in some way.
- Client and Server PointLightTrees function independently of one another. This is due to the fact that the `SharedPointLightComponent` is abstract, and ComponentTrees do not like abstract or virtual classes at all. To make these trees synced, all the Client and Server PointLightComponent attributes would need to be moved up into the Shared space. Which in and of itself is a large refactor.
- Small gripe but I'm not sure if `LightSensitive` makes sense anymore, because initially this component/system was going to send out networked events on regular intervals, but i later opted to just put the onus on other systems to call this one. Maybe it should be `EntityIllumination` or something

Anwyay, I did my best to bring back this functionality that was pretty core to old SS13 and hope we can work together to get this in and further develop with the new design space


ghost:
I would love this feature,

Robust.Shared/GameObjects/Systems/SharedLightSensitiveSystem.cs
```diff
@@ -0,0 +1,129 @@
...
+                {
+                    continue;
+                }
+
+                return false;
+            }
+
+            return true;
+        }
+
```
metalgearsloth: Please don't just copy-paste unperformant code from content to engine. This raycast doesn't even work properly and should be using the new raycastsystem and avoid allocating so much.

Robust.Server/ComponentTrees/LightTreeSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+    protected override int InitialCapacity => 128;
+    protected override bool Recursive => true;
+
+    protected override void OnCompStartup(EntityUid uid, PointLightComponent component, ComponentStartup args)
+    {
+        base.OnCompStartup(uid, component, args);
+        /// This feels like a poor workaround to the fact that Entites can't have duplicate subscriptions to the same event, but it works. 
+        _pointLightSystem.OnLightStartup(uid, component, args);
+    }
+
```
metalgearsloth: This is just duplicating the sub anyway? This method is pointless as is.

Robust.Server/ComponentTrees/LightTreeSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+    protected override Box2 ExtractAabb(in ComponentTreeEntry<PointLightComponent> entry)
+    {
+        if (entry.Component.TreeUid == null)
+            return default;
+
+        var pos = XformSystem.GetRelativePosition(
+            entry.Transform,
+            entry.Component.TreeUid.Value,
+            GetEntityQuery<TransformComponent>());
+
```
metalgearsloth, line 15: Resolving getentityquery every time isn't cheap, is this supposed to be on the system.

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,146 @@
...
+    private void ProcessNearbyLights(EntityUid uid, LightSensitiveComponent component, TransformComponent entityXform)
+    {
+        var illumination = 0f;
+
+        var ourPosition = _transform.GetMapCoordinates(uid);
+        var bounds = _physics.GetWorldAABB(uid);
+        var queryResult = _lightTreeSystem.QueryAabb(ourPosition.MapId, bounds);
+
+        foreach (var val in queryResult)
+        {
```
metalgearsloth: You're resolving worldposition twice for no reason, between the AABB and MapCoordinates.

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,146 @@
...
+            var (lightComp, lightXform) = val;
+            if (!lightComp.Enabled)
+                continue;
+
```
metalgearsloth: Disabled lights shouldn't even be stored in the lighttree, it's a massive waste of performance to be moving hundreds of these (PDAs for example) per tick.

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,146 @@
...
+            if (!InRangeUnOccluded(lightPosition, ourPosition, lightComp.Radius, null))
+                continue;
+
+            entityXform.Coordinates.TryDistance(EntityManager, lightXform.Coordinates, out var dist);
+
+            var denom = dist / lightComp.Radius;
+            /// A slight modification of standard inverse square falloff that incorperates the radius of the light to make it more forgiving
+            var attenuation = 1 - (denom * denom);
+
+            var calculatedLight = 0f;
```
metalgearsloth: Raycasting is significantly more expensive than checking illumination contribution if we're just going to take max and not accumulate them (though I suspect not accumulating will cause other issues) so you're better off doing that first and earlying out.

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            if (!InRangeUnOccluded(lightPosition, ourPosition, lightComp.Radius, null))
+                continue;
+
+            entityXform.Coordinates.TryDistance(EntityManager, lightXform.Coordinates, out var dist);
+
+            var denom = dist / lightComp.Radius;
+            /// A slight modification of standard inverse square falloff that incorperates the radius of the light to make it more forgiving
+            var attenuation = 1 - (denom * denom);
+
```
metalgearsloth: TryDistance is also too expensive to use when you already have mapcoordinates for both above. It's doing a dozen+ trycomps when you can just do a vector subtraction.

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// <param name="entityXform">The TransformComponent of the entity</param>
+    /// <returns>If the component existed in the entity.</returns>
+    private void ProcessNearbyLights(EntityUid uid, LightSensitiveComponent component, TransformComponent entityXform)
+    {
+        var illumination = 0f;
+
+        var ourPosition = _transform.GetMapCoordinates(uid);
+        var bounds = _physics.GetWorldAABB(uid);
+        var queryResult = _lightTreeSystem.QueryAabb(ourPosition.MapId, bounds);
+
```
metalgearsloth: The query is making no distinction between clientside vs shared lights so client and server will inherently mispredict values (ignoring the fact that it will be mispredicted anyway due to ping).

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,146 @@
...
+                        coneLight = lightComp.Energy * attenuation * attenuation * angleAttenuation;
+                    }
+                    // Target is inside the inner cone, so the light level will use standard falloff.
+                    else
+                    {
+                        coneLight = lightComp.Energy * attenuation * attenuation;
+                    }
+                    calculatedLight = Math.Max(calculatedLight, coneLight);
+                }
+            }
```
metalgearsloth: Lots of redundant Math.Abs here.

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            // We only care about the strongest light level being applied to the entity. 
+            // I really don't want to deal with multiplicative or additive illumination
+            illumination = Math.Max(illumination, calculatedLight);
+        }
+
+        SetIllumination(uid, illumination, component);
+        component.LastUpdate = _gameTiming.CurTime;
+    }
+
```
metalgearsloth: This should be doing NextUpdate so it applies properly on first tick.

Robust.Server/GameObjects/EntitySystems/PointLightSystem.cs
```diff
@@ -27,6 +34,16 @@ protected override void UpdatePriority(EntityUid uid, SharedPointLightComponent
...
+    private void OnInserted(EntityUid uid, PointLightComponent component, EntGotInsertedIntoContainerMessage args)
+    {
+        SetContainerOccluded(uid, args.Container.OccludesLight, component);
+    }
+
+    private void OnRemoved(EntityUid uid, PointLightComponent component, EntGotRemovedFromContainerMessage args)
+    {
+        SetContainerOccluded(uid, false, component);
+    }
+
```
metalgearsloth, line 60: Please don't do this, you're misunderstanding what containeroccluded is for. There's a reason UpdateEntity is the method updating this.

Containers are recursive so this event doesn't get fired off a lot of the time. E.g. a PDA light will never fire this event because it's a nested container.

Robust.Server/GameObjects/EntitySystems/PointLightSystem.cs
```diff
@@ -72,4 +92,35 @@ public override bool RemoveLightDeferred(EntityUid uid)
...
+    public override void SetContainerOccluded(EntityUid uid, bool occluded, SharedPointLightComponent? comp = null)
+    {
+        if (!ResolveLight(uid, ref comp) || occluded == comp.ContainerOccluded || comp is not PointLightComponent clientComp)
+            return;
+
+        base.SetContainerOccluded(uid, occluded, comp);
+        if (comp.Enabled)
+            _lightTree.QueueTreeUpdate(uid, clientComp);
+    }
+
```
metalgearsloth: Content should not be setting this as they will break contained lights so remove as mentioned above.

Robust.Shared/GameObjects/Components/Light/LightMaskPrototype.cs
```diff
@@ -0,0 +1,70 @@
...
+using Robust.Shared.Serialization.Manager;
+using Robust.Shared.Serialization.Manager.Attributes;
+using Robust.Shared.Serialization.Markdown.Mapping;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype.Array;
+using Robust.Shared.Utility;
+using Robust.Shared.ViewVariables;
+
+namespace Robust.Shared.GameObjects;
+
+[Prototype("lightMask")]
```
metalgearsloth: Redundant string.

Robust.Shared/GameObjects/Components/Light/LightMaskPrototype.cs
```diff
@@ -0,0 +1,70 @@
...
+    [ViewVariables(VVAccess.ReadWrite), DataField("maskPath", required: true)]
+    public ResPath MaskPath;
+
+    /// <summary>
+    /// Light of light cones that correspond to the light mask.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadWrite), DataField("lightCones", required: true)]
+    public List<LightConeData> LightCones = default!;
+
+}
```
metalgearsloth, line 28: Redundant datafield strings, also why is it default null suppress this will just blow up if hot reloaded.

Robust.Shared/GameObjects/Components/Light/LightMaskPrototype.cs
```diff
@@ -0,0 +1,70 @@
...
+    [DataField("innerWidth", required: true)]
+    public float InnerWidth = default!;
+
+    /// <summary>
+    ///     Angle limit of the cone to be within the reduced light. Beyond this angle, you are considered out of the light.
+    /// </summary>
+    [DataField("outerWidth", required: true)]
+    public float OuterWidth = default!;
+
+}
```
metalgearsloth, line 63: No reason not to be a struct also datafield strings redundant, and don't just do null suppress.

Robust.Shared/GameObjects/Components/Light/LightSensitiveComponent.cs
```diff
@@ -0,0 +1,31 @@
...
+    public float LightLevel = 0f;
+    
+    /// <summary>
+    ///     When this Entity last had its light level evaluated. 
+    ///To prevent multiple expensive updates per tick.
+    ////// </summary>
+    [AutoNetworkedField]
+    [ViewVariables(VVAccess.ReadOnly)]
+    public TimeSpan LastUpdate = TimeSpan.Zero;
+
```
metalgearsloth: These should really be getting saved as datafields considering they're stateful

Robust.Shared/GameObjects/Components/Light/PointLightComponentState.cs
```diff
@@ -21,4 +21,6 @@ public sealed class PointLightComponentState : ComponentState
     public float Radius;
 
     public Vector2 Offset;
+
+    public bool ContainerOccluded;
 }
```
metalgearsloth, line 30: I think it's sufficient enough that this gets derived on UpdateEntity and doesn't need networking.

Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
```diff
@@ -88,11 +91,10 @@ public float AnimatedRadius
...
         /// <summary>
-        /// The resource path to the mask texture the light will use.
+        /// The Prototype ID of the light mask the light uses.
         /// </summary>
-        [ViewVariables(VVAccess.ReadWrite)]
-        [DataField("mask")]
-        public string? MaskPath;
+        [DataField("lightMask")]
+        public ProtoId<LightMaskPrototype>? LightMask;
     }
```
metalgearsloth, line 152: Needs engine CL as a breaking change, also datafield string redundant.

Robust.Shared/GameObjects/Systems/SharedLightSensitiveSystem.cs
```diff
@@ -0,0 +1,129 @@
...
+using Robust.Shared.Maths;
+using Robust.Shared.Physics;
+
+namespace Robust.Shared.GameObjects
+{
+
+    public abstract class SharedLightSensitiveSystem : EntitySystem
+    {
+        [Dependency] private readonly SharedTransformSystem _transform = default!;
+        [Dependency] private readonly OccluderSystem _occluder = default!;
```
metalgearsloth: Please document your systems, especially on engine.

Robust.Shared/GameObjects/Systems/SharedLightSensitiveSystem.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        protected virtual void SetIllumination(EntityUid uid, float value, LightSensitiveComponent? comp = null)
+        {
+            if (!ResolveComp(uid, ref comp))
+                return;
+
+            comp.LightLevel = value;
+            Dirty(uid, comp);
+        }
+
```
metalgearsloth: You should be checking if the value is equal before dirtying as dirtying is expensive.

Robust.Shared/GameObjects/Systems/SharedLightSensitiveSystem.cs
```diff
@@ -0,0 +1,129 @@
...
+            if (angle == double.NaN && _transform.ContainsEntity(targetUid, lightUid) || _transform.ContainsEntity(lightUid, targetUid))
+            {
+                angle = 0f;
+            }
+
+            return angle;
+        }
+
+    }
+}
```
metalgearsloth: Same here, delete all of this. Even GetAngle should not be getting called in a hot loop when you already have the data inside of the loop, just assume every transform method is 1,000 times more expensive than the basic maths operations.

Robust.Shared/GameObjects/Systems/SharedPointLightSystem.cs
```diff
@@ -37,6 +38,15 @@ public void SetColor(EntityUid uid, Color value, SharedPointLightComponent? comp
...
 
+    public virtual void SetContainerOccluded(EntityUid uid, bool occluded, SharedPointLightComponent? comp = null)
+    {
+        if (!ResolveLight(uid, ref comp) || occluded == comp.ContainerOccluded)
+            return;
+
+        comp.ContainerOccluded = occluded;
+        Dirty(uid, comp);
+    }
+
```
metalgearsloth, line 62: As above delete as content will kneecap itself with this.

metalgearsloth (changes requested):
I don't think the changes are that bad however I think you need to consider the fact content will likely need to call this inside of parallelism due to how expensive it will be even after you optimise it.

metalgearsloth:
Also I would note that content can entirely override this as it sees fit, e.g. they can draw directly to the light render target for tile emissive or rooves or whatever and it probably needs an event attached.

metalgearsloth:
Note that I don't expect you to actually do it in this PR but I do expect content PRs to be frozen from using this until it is implemented.

Chubbygummibear:
> Note that I don't expect you to actually do it in this PR but I do expect content PRs to be frozen from using this until it is implemented.

i would be willing to try fixing this up in a more optimized manner and in a way that better fits into the content/engine schema, but I would definitely need help since I'm very unfamiliar with the ECS paradigm, and content/server/shared system. I have enough experience to get things working, but when you say like "consider parallelism", or when I should or shouldn't dirty a component, it's totally new to me.

Robust.Server/ComponentTrees/LightTreeSystem.cs
Chubbygummibear: the problem i was trying to solve here is that the ComponentTreeSystem automatically subscribes to ComponentStartup for the purpose of updating the tree. However the PointLightSystem also subscribes to the Startup event because it wants to call UpdatePriority for the metadata. This caused a duplicate subscription error that i could not figure out how to get around without making a hacky call chain. I tried asking around the server but didn't find a good workaround to the duplicate subs issue

Robust.Server/ComponentTrees/LightTreeSystem.cs
Chubbygummibear: i'm afraid i don't understand what you mean by "on the system". this behavior is just copied from the AmbientSoundTreeSystem, since a while into this project, i reasoned that the local sound system that used line of sight and occlusion to determine what an entity "heard" wasn't all that different from light. However i do recognize that the cost would be different performing this on the server as opposed to the client

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
Chubbygummibear: what would be a good solution to avoiding mispredictions? also i would like to make the component trees between client and server synchronized, but i don't know the best way to do that when i tried to put everything in Shared before and it caused a large number of errors because it did not like the abstract and virtual methods touching the component trees

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
Chubbygummibear: yeah that's true. i wanted the sanity checks that come with the GetMapCoordinates or GetWorldPositionRotation, but those don't return every value i need at the same time, so i'll probably add a new function to this system that's a copy of those but returns all the values i need in one pass

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
Chubbygummibear: good point

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
Chubbygummibear: will simplify to a single var

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
Chubbygummibear: ah yeah that makes sense. i'll change it to that

Robust.Server/ComponentTrees/LightTreeSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+public sealed class LightTreeSystem : ComponentTreeSystem<LightTreeComponent, PointLightComponent>
+{
+    [Dependency] private readonly PointLightSystem _pointLightSystem = default!;
+    
+    #region Component Tree Overrides
+    protected override bool DoFrameUpdate => false;
+    protected override bool DoTickUpdate => true;
+    protected override int InitialCapacity => 128;
+    protected override bool Recursive => true;
```
ElectroJr: Recursive component trees are really expensive, and probably need to be re-worked somehow. If there's **any** recursive tree, then anytime a MoveEvent is fired it has to recursively iterate through all children of that entity and try to fetch components. Currently there are no server-side recursive trees for specifically that reason, it'd hurt server performance quite a bit. E.g,, players generally contain many (100+) entities, so just handling player movement would be relatively expensive.

I'm not sure what the best solution in the long term is. I think recursive trees should be replaced with some setup where entities with the component being being attached to other entities should add a component & metadata flag to their parent. I.e., the mover entity has some ListListComponent that points directly to any children with a light component, thus avoiding the recursive move events. But thats beyond the scope of this PR and probably something I'll work on

But at the very least for now this whole tree should be locked behind a cvar for any server that doesn't want to use the feature. Even if recursive trees are updated to be faster, it should probably stay toggleable. I.e., give the base tree system some virtual `IsEnabled` bool that skips init, shutdown, and maybe prints warnings/errors when people try to use queries. Being able to disable the trees is probably just something I should've added from the beginning. E.g., I don't even know if opendream or whatever else needs to use things like the ocluder & light trees.

Robust.Server/ComponentTrees/LightTreeSystem.cs
ElectroJr: One of the common solutions is to have some startup event that exists specifically for that component, e.g. `GridStartupEvent` or `TransformStartupEvent`. And thats probably a better solution in case other content systems want to extend light startup logic without having to somehow give all lights a new component.

Robust.Server/ComponentTrees/LightTreeSystem.cs
ElectroJr: Sloth just means having a field on the system itself that stores the query. I.e. use a `private EntityQuery<TComponent> _query` field  that gets set on system init. Regarding the other tree systems, those should be updated. I either messed that up when writing them or maybe I just wrote them system before we started caching/reusing queries.

Though in the specific where its for TransformComponent or MetadataComponent, you shouldn't even need to cache the queries at all, as metadata system already does that for you. I.e., just using `Transform(uid)` will use `EntityManager.TransformQuery.GetComponent(uid)`

Robust.Server/GameObjects/Components/PointLightComponent.cs
```diff
@@ -1,9 +1,32 @@
...
+namespace Robust.Server.GameObjects
 {
+    [RegisterComponent]
+    public sealed partial class PointLightComponent : SharedPointLightComponent, IComponentTreeEntry<PointLightComponent>
+    {
+        #region Component Tree
+
+        /// <inheritdoc />
+        [ViewVariables]
+        public EntityUid? TreeUid { get; set; }
```
ElectroJr: Except for the `DynamicTree` field, cant all of the fields be moved to shared?
Though like I said in another comment, you can probably move even that to shared if you get rid of the new constraint.

Robust.Shared/GameObjects/Components/Light/LightMaskPrototype.cs
ElectroJr: what sloth said + data fields should all default to vv read+write.

Robust.Shared/GameObjects/Components/Light/LightMaskPrototype.cs
ElectroJr: I thought hot reloading would work fine with required data-fields? Though this probably can just be null or optional.

Robust.Server/ComponentTrees/LightTreeSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+using Robust.Server.GameObjects;
+using Robust.Shared.ComponentTrees;
+using Robust.Shared.GameObjects;
+using Robust.Shared.Maths;
+using Robust.Shared.Physics;
+using Robust.Shared.IoC;
+
+namespace Robust.Server.ComponentTrees;
+
+public sealed class LightTreeSystem : ComponentTreeSystem<LightTreeComponent, PointLightComponent>
```
ElectroJr: > Client and Server PointLightTrees function independently of one another. This is due to the fact that the SharedPointLightComponent is abstract

I **think** you can just remove the `where TComp : new()` requirement in `ComponentTreeSystem<TTreeComp, TComp>`? It doesn't seem to actually be required, and then you can just point to the abstract shared component, and have a shared tree system. Thought you might still want a server/client specific version that inherits from the abstract shared system for `DoFrameUpdate` or cvar toggling.

If you do this, you might have to update some of the rendering code so it casts shared light comp to the client comp to access the mask texture, but its probably worth it to avoid duplicating the systems

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,190 @@
...
+            // I really don't want to deal with multiplicative or additive illumination
+            illumination = Math.Max(illumination, calculatedLight);
+        }
+
+        SetIllumination(uid, illumination, component);
+        
+    }
+
+    public Box2 GetWorldAABB(EntityUid uid, out Vector2 worldPos, out Angle worldRot, FixturesComponent? manager = null, PhysicsComponent? body = null, TransformComponent? xform = null)
+    {
```
ElectroJr: This just seems like a variant of physics system method that also returns the world position? If thats the case, that should just be an overload in physics system, not a method here.

Robust.Shared/GameObjects/Systems/SharedLightSensitiveSystem.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (!ResolveComp(uid, ref comp))
+                return;
+
+            comp.LightLevel = value;
+            Dirty(uid, comp);
+        }
+
+
+        public virtual bool InRangeUnOccluded<TState>(MapCoordinates origin, MapCoordinates other, float range,
+            TState state, Func<EntityUid, TState, bool> predicate, bool ignoreInsideBlocker = true, IEntityManager? entMan = null)
```
ElectroJr: I didn't read through this all that carefully, but I assume this just just a direct port of content's `ExamineSystemShared.InRangeUnOccluded` to engine? I don't know if this is the best spot for it, but I guess `OccluderSystem` is client-side. Though maybe its better in a shared point light system?

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,190 @@
...
+using Robust.Shared.Prototypes;
+using Robust.Shared.Timing;
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Numerics;
+using TerraFX.Interop.Windows;
+
+namespace Robust.Server.GameObjects;
+
+public sealed class LightSensitiveSystem : SharedLightSensitiveSystem
```
ElectroJr: I think engine should just provide a more minimalist "LightLevel" system that can be used to compute the light level at a given position, probably with the caveat in the documentation that its generally expensive. 

Then content can use that to do more complex things if it wants (e.g., cache it on comp on entities that auto-update, or create a tile map or something like that)

Robust.Server/GameObjects/EntitySystems/PointLightSystem.cs
```diff
@@ -1,19 +1,26 @@
...
+using Robust.Shared.Containers;
 using Robust.Shared.GameObjects;
 using Robust.Shared.GameStates;
 using Robust.Shared.IoC;
+using Robust.Shared.Maths;
 
 namespace Robust.Server.GameObjects;
 
-public sealed class PointLightSystem : SharedPointLightSystem
+    public sealed class PointLightSystem : SharedPointLightSystem
```
ElectroJr: indent

Robust.Server/GameObjects/EntitySystems/PointLightSystem.cs
```diff
@@ -39,6 +47,32 @@ private bool IsHighPriority(SharedPointLightComponent comp)
...
+            Enabled = component.Enabled,
+            Energy = component.Energy,
+            Offset = component.Offset,
+            Radius = component.Radius,
+            Softness = component.Softness,
+            CastShadows = component.CastShadows,
+            ContainerOccluded = component.ContainerOccluded,
+        };
+
+        _lightTree.QueueTreeUpdate(uid, component);
```
ElectroJr: This shouldn't be getting called here, it should be getting used wherever something is being modified that will require a tree update.

get state gets called once per player that needs to receive a state, not once per dirty

Robust.Server/GameObjects/EntitySystems/PointLightSystem.cs
ElectroJr: This was originally my bad, I should've made the client-side method internal a while ago. Seeing as this has breaking changes anyway may as well make it internal now. And again, hopefully the tree system can be moved to shared so that most of the code duplication can be removed and this can all be handled in the shared system

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,190 @@
...
+            var dist = ourMapPos.Position - lightPosition.Position;
+            //var deltaLength = delta.Length();
+
+            var denom = dist.Length() / lightComp.Radius;
+            /// A slight modification of standard inverse square falloff that incorperates the radius of the light to make it more forgiving
+            var attenuation = 1 - (denom * denom);
+
+            var calculatedLight = 0f;
+            if (_proto.TryIndex(lightComp.LightMask, out var mask))
+            {
```
ElectroJr: I think this if block should be extract to a separate apply mask method to make it easier to read

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,190 @@
...
+                        )
+                    {
+                        coneLight = lightComp.Energy * attenuation * attenuation * angleAttenuation;
+                    }
+                    // Target is inside the inner cone, so the light level will use standard falloff.
+                    else
+                    {
+                        coneLight = lightComp.Energy * attenuation * attenuation;
+                    }
+                    calculatedLight = Math.Max(calculatedLight, coneLight);
```
ElectroJr: `lightComp.Energy * attenuation * attenuation` appears in all branches, so you can probably just move that before the if-mask check.

Also `calculatedLight` is always zero on this branch up until the math.max, so I'm not sure why you need the max

ElectroJr (changes requested):

Robust.Server/ComponentTrees/LightTreeSystem.cs
ElectroJr: See #6211

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,190 @@
...
+            if (!InRangeUnOccluded(lightPosition, ourMapPos, lightComp.Radius, null))
+                continue;
+
+            // entityXform.Coordinates.TryDistance(EntityManager, lightXform.Coordinates, out var dist);
+            var dist = ourMapPos.Position - lightPosition.Position;
+            //var deltaLength = delta.Length();
+
+            var denom = dist.Length() / lightComp.Radius;
+            /// A slight modification of standard inverse square falloff that incorperates the radius of the light to make it more forgiving
+            var attenuation = 1 - (denom * denom);
```
ElectroJr: It's probably worth actually making this match how `float val` is calculated in `light_shared.swsl`

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,190 @@
...
+                }
+            }
+            // No mask, use normal falloff
+            else
+            {
+                calculatedLight = lightComp.Energy * attenuation * attenuation;
+            }
+
+            // We only care about the strongest light level being applied to the entity. 
+            // I really don't want to deal with multiplicative or additive illumination
```
ElectroJr: IMO its probably worth supporting this I think you'd just need to do what the light shader does. I.e., compute `val` like in `light_shared.swsl`, multiply the light's color (a "black" light should have zero illumination), add all the colors together, and then map that somehow onto an illumination value. The specifics of how the color -> illumination mapping is done can always be changed later.

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
```diff
@@ -0,0 +1,190 @@
...
+        var illumination = 0f;
+
+        //var ourPosition = _transform.GetMapCoordinates(uid);
+        var ourBounds = GetWorldAABB(uid, out var ourPos, out var ourRot);
+        var ourMapPos = new MapCoordinates(ourPos, entityXform.MapID);
+        var queryResult = _lightTreeSystem.QueryAabb(ourMapPos.MapId, ourBounds);
+
+        foreach (var val in queryResult)
+        {
+            var (lightComp, lightXform) = val;
```
ElectroJr: IMO this should default to ignoring all lights with `LightComponent.CastShadows` set to false, as they should all be quite small lights and it'd help a bit with performance

Robust.Server/GameObjects/EntitySystems/PointLightSystem.cs
```diff
@@ -39,6 +47,32 @@ private bool IsHighPriority(SharedPointLightComponent comp)
...
+
+    private void OnRemoved(EntityUid uid, PointLightComponent component, EntGotRemovedFromContainerMessage args)
+    {
+        SetContainerOccluded(uid, false, component);
+    }
+
+    private void OnLightGetState(EntityUid uid, PointLightComponent component, ref ComponentGetState args)
+    {
+        args.State = new PointLightComponentState()
+        {
```
ElectroJr, line 64: This should stay in shared code, and it also has a hidden conflict with #6160 due to the missing `Falloff` & `CurveFactor`

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
ElectroJr: I'm also not sure if this is really needed, because we're not averaging the light level over an entity's bounds, its just being computed at the entity's centre. So you should be able to just use a "point box" (`Box2(pos, pos)`) to get all lights that intersect the point.

Robust.Shared/GameObjects/Systems/SharedLightSensitiveSystem.cs
ElectroJr: Actually, there's quite a few problems with this method which also needs fixing in content. Instead of just listing them all and making you fix them I'm just going to create a separate PR that migrates the method to engine and fixes all the issues I can think of.

Lumminal:
it'd be amazing if this was finished properly, we really need it to unlock a new area of content (light-based mechanics). Author doesn't seem active on this anymore so its kinda sad

Chubbygummibear:
> it'd be amazing if this was finished properly, we really need it to unlock a new area of content (light-based mechanics). Author doesn't seem active on this anymore so its kinda sad

oh don't worry i'm still suffering away at this. it got review again because i poked different devs. i was pretty confused on what needed to be done and the best way to do it, so i needed head dev help. i've had like 2 different ss14 servers poke me recently asking if it would be possible to get a portable version of this, but not really without touching the engine, so i basically have to keep at this even though it's in a nightmare state of bandaids and workarounds

Robust.Server/GameObjects/EntitySystems/PointLightSystem.cs
Chubbygummibear: Understood, but I was also hesitant to touch the shared code version because it was static and cannot be overwridden. Do you know if it's static for a specific reason? other component states just seem to be internal, virtual, or abstract

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
Chubbygummibear: I went back and forth on this a lot and maybe the best option is a cvar for that because on one hand, calculating lights that don't cast shadows means every glowing corn cob from botany would turn up in the light query for every lightsensitive entity in the area, which i'm sure would kill performance, but on the other hand i saw many entities that don't cast shadows such as glowing floor tiles which i believe SHOULD be factored in

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
Chubbygummibear: color has now been factored in by getting the Max of the RGB values, which is how luminosity is determined for HSL. however i still don't have a good answer for getting light mask calculations to multiply against the light level, which may need to be a future separate PR where we read the pngs as a 2d matrix of rgb values

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
Chubbygummibear: mission accomplished, my attenuation has been removed in favor of the shader's formula

Robust.Server/GameObjects/EntitySystems/LightSensitiveSystem.cs
Chubbygummibear: on the todo list for if we can come up with a better way of extracting a mask value from the images of the mask itself, or we just stick with what i have now

Robust.Server/GameObjects/EntitySystems/PointLightSystem.cs
ElectroJr: No idea why it was static, theres no real reason it needs to be.

Chubbygummibear:
we are merge conflict free once again

someone please end this nightmare

Golubgik:
you won't give up 💪

Chubbygummibear:
this pr may be one of the oldest on this repo but i assure you it is still a bear trap waiting for another maintainer to acknowledge me. on god whoever does is going to be stuck with me

Chubbygummibear:
image

PR #5576, opened 2024-12-26
PlacementManager CurrentEraserMouseCoordinates fix

### LINKED ISSUE #5575: PlacementManager's Erase method is behaving strangely
When modifying maps and using the PlacementManager's entity-based rectangular erase method, the coordinates of the erasing rectangle appear to be very off.

This area in the video is near 0,0 for the map. 
https://github.com/user-attachments/assets/dbccbef7-ea15-4efc-982e-dbe9807e8248

I'm going to guess it has something to do with this change [PlacementManager.cs](https://github.com/space-wizards/RobustToolbox/commit/ac30ad18209cbcbb25b53989ed21b9b38d6c91bf#diff-0e0b858def01d1fbefe40b1fa248a50db1af4ad5b8e74eff24998336762f4fcc)

southbridge-fur:
Resolves #5575

With the recent modification to this method to remove deprecated calls, it appears a bug was introduced which messed up the coordinates being used in `PlacementManager`'s `CurrentEraserMouseCoordinates`. It appears the issue has something to do with how the call to the `ToCoordinates` method includes an entity which causes it to seemingly convert the coordinates to be relative to the entity, which isn't very conducive to the erase method here. One day I hope to understand the coordinate systems in this game so I can know if my read on this is correct, but today is not that day.

I've resolved the issue by using a different method overload for `EntityCoordinates`'s `ToCoordinates` which doesn't have an argument for the entity. 

This seems to work rather well, however during testing I noticed a new bug, which threw an endless stream of error messages if I happened to mouse over any UI element. I've added an additional fix for this, which simply checks whether the MapId we're moused over is equal to the `MapId.Nullspace` value. This bug is also present in the current version of the code, so I figured I'd resolve it as well.

My fix for this simply causes the method to return its false result, which functionally just makes the selector freeze in place while you're mousing over UI elements. It doesn't feel the best, but I fail to understand the coordinate systems in this game, so I won't be resolving that.

https://github.com/user-attachments/assets/cee375de-1a31-4ddb-9a2a-00804f059239

### DIFF
--- a/Robust.Client/Placement/PlacementManager.cs
+++ b/Robust.Client/Placement/PlacementManager.cs
@@ -536,13 +536,14 @@ namespace Robust.Client.Placement
             }
             else
             {
+                var mousePosition = EyeManager.PixelToMap(InputManager.MouseScreenPosition);
                 var map = EntityManager.GetComponent<TransformComponent>(ent).MapID;
-                if (map == MapId.Nullspace || !Eraser)
+                if (map == MapId.Nullspace || !Eraser || mousePosition.MapId == MapId.Nullspace)
                 {
                     coordinates = new EntityCoordinates();
                     return false;
                 }
-                coordinates = XformSystem.ToCoordinates(ent, EyeManager.PixelToMap(InputManager.MouseScreenPosition));
+                coordinates = XformSystem.ToCoordinates(mousePosition);
                 return true;
             }
         }


southbridge-fur:
the day I finally understand the coordinate systems will be the day I implement copy-paste (and completely overhaul PlacementManager)

ElectroJr:
> the day I finally understand the coordinate systems

It's not that complicated, but placement manager is really not the place to learn, its shitty, partially ancient code that no one has properly cleaned up in a long time.

Point in case, the bug here was only possible because placement manager is hot garbage and was and is still assuming that the entity coordinates are relative to the map.

PR #5577, opened 2024-12-27
Add WeakEntityReference

ElectroJr:
This adds a new `WeakEntityReference` struct that is intended to be used by component data-fields to refer to entities that may or may not still exist, or in the case of `WeakEntityReference<T>` that might no longer have an otherwise required `T` component. Feel free to suggest better names for the struct, maybe it should just be abbreviated?.

## Why
The current convention is that a a non-null EntityUid? (or NetEntity) should correspond to a valid non-deleted entity. Saving or loading entities that have components with invalid / deleted entity uids currently logs an error or completely fails, because the assumption is that you forgot to include an entity in the save or something has gone wrong. Similarly, trying to sent a deleted entityuid over the network will log an error when converting it to a NetEntity using the EntityManager helper methods.

Ensuring that entities delete or clear references to themselves when deleted or components are removed is a PITA (device linking, ghosts following, etc). So it'd be nice to have some kind of struct that points to an entity that may or may not still exist, that doesn't need to be manually cleaned up, and which yaml & network serialization knows it can just ignore if the entity is deleted or not being included in a yaml save file. Though it also shouldn't get overused, as it is obviously slower than directly using an EntityUid.

## How

The struct is is just a simple wrapper around `NetEntity`, and there are some new IEntityManager methods & EntitySystem proxy methods to convert back to either an `EntityUid?` or `Entity<T>?` and is meant to be used with 'is' casting (i.e., `if (Resolve(weakRef) is not {} ent)`). Previously it used an EntityUid instead of NetEntity, but I switched mostly just to avoid having to update the auto component state generator.


Tayrtahn:
This would be very helpful for solving issues like https://github.com/space-wizards/space-station-14/issues/32642, and the more general problem of tracking projectile sources.

Tayrtahn:
A few suggested additions:

`IEntityManager.GetWeakReference(EntityUid uid)` - as a shortcut for `new WeakEntityReference(EntityManager.GetNetEntity(uid))`

`IEntityManager.TryResolve(WeakEntityReference weakRef)` - expand the suite of Resolve methods for parity with the `NetEntity` resolvers.

Robust.Shared/GameObjects/EntitySystem.Proxy.cs
```diff
@@ -1510,4 +1510,36 @@ protected NetCoordinates[] GetNetCoordinatesArray(EntityCoordinates[] entities)
     }
 
     #endregion
+
+    #region WeakEntityReference
+
+    /// <inheritdoc cref="IEntityManager.Resolve(WeakEntityReference)"/>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    protected EntityUid? Resolve(ref WeakEntityReference weakRef)
```
Tayrtahn: Since these don't modify the reference, should they use `in` instead of `ref`?

Tayrtahn:
Also consider adding these:
- [Integration test](https://github.com/ElectroJr/RobustToolbox/pull/1)
- [`ToPrettyString` support](https://github.com/ElectroJr/RobustToolbox/pull/2)
- [`GetWeakReference` methods](https://github.com/ElectroJr/RobustToolbox/pull/3)

Robust.Shared/GameObjects/EntitySystem.Proxy.cs
ElectroJr: I've changed them to match the EntityManager method signature. I think this was just a holdover from an older version.

ElectroJr:
> `IEntityManager.TryResolve(WeakEntityReference weakRef)` - expand the suite of Resolve methods for parity with the NetEntity resolvers.

I've added `TryGetEntity` methods

> Also consider adding these

I've merged all those changes i

Robust.Shared/GameObjects/EntitySystem.Proxy.cs
```diff
@@ -1606,4 +1622,76 @@ protected NetCoordinates[] GetNetCoordinatesArray(EntityCoordinates[] entities)
     }
 
     #endregion
+
+    #region WeakEntityReference
+
+    /// <inheritdoc cref="IEntityManager.GetWeakReference(EntityUid, MetaDataComponent?)"/>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    protected WeakEntityReference GetWeakReference(ref EntityUid uid, MetaDataComponent? meta = null)
```
Tayrtahn:
```suggestion
    protected WeakEntityReference GetWeakReference(EntityUid uid, MetaDataComponent? meta = null)
```

Robust.Shared/GameObjects/EntitySystem.Proxy.cs
```diff
@@ -1606,4 +1622,76 @@ protected NetCoordinates[] GetNetCoordinatesArray(EntityCoordinates[] entities)
...
+    {
+        return EntityManager.GetWeakReference(uid, meta);
+    }
+
+    /// <inheritdoc cref="IEntityManager.GetWeakReference(EntityUid?, MetaDataComponent?)"/>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    protected WeakEntityReference? GetWeakReference(ref EntityUid? uid, MetaDataComponent? meta = null)
```
Tayrtahn:
```suggestion
    protected WeakEntityReference? GetWeakReference(EntityUid? uid, MetaDataComponent? meta = null)
```

Tayrtahn (changes requested):
Should remove `ref` from these too.

Tayrtahn (approved):

PR #5578, opened 2024-12-27
Log errors when encountering entity data fields

### LINKED ISSUE #4855: Serve should destroy Entity<T>
It can be added as a datafield and people have started doing it but this is going to break serialization hard afaik, it'll just be a timebomb whenever someone decides to fix persistence.

ElectroJr:
`ISerializationManager` will now log errors if it encounters `Entity<T>` data-fields, which have never been supported but have been getting used in some content code. Currently this is just done with a blacklisting attribute, because I don't think there's any easy way to automate it with the current serialization system.

This PR also makes some yaml node validation methods more robust, so that they return error nodes instead of throwing exceptions.

Fixes #4855

Requires https://github.com/space-wizards/space-station-14/pull/34083


slarticodefast:
Resolves https://github.com/space-wizards/space-station-14/issues/34030

Robust.Shared/Serialization/Manager/SerializationManager.cs
```diff
@@ -172,12 +173,66 @@ IEnumerable<Type> GetImplicitTypes(Type type)
...
+            // We want to ensure that all the fields are actually serializable.
+            // Problem is that I have NFI how to do that and all of this is such convoluted spaghetti that the best way.
+            // Why is there no "is this serializable" test.
+            // Best I could get was to try brute force this by repeatedly trying to call ValidateNode with either a value, mapping, or sequence data node and checking that at least one of them doesn't throw an exception.
+            // But that still fails, because things like EntityUid aren't actually serializable without the mapping context.
+            // TODO SERIALIZATION REFACTOR Somehow validate that data-fields are serializable.
+            // For now, I will just do a very basic blacklist check.
+
+            var forbidden = _reflectionManager.FindTypesWithAttribute<NotYamlSerializableAttribute>()
+                .Select(x => x.IsGenericType ? x.GetGenericTypeDefinition() : x )
```
PJB3005: Shouldn't the types returned here already be the generic type definition?

ElectroJr: You're correct, I've removed the unnecessary select

PR #5588, opened 2025-01-06
Push clientside transform GridUids to children

Tayrtahn:
When an entity is parented to another entity that does not currently exist on the client (such as when the new parent is outside of the client's PVS), the child entity does not have its GridUid set properly. This can cause issues with other systems that rely on the GridUid.

For example, when using the Follow button from the ghost roles panel, if the target entity is outside of PVS, the camera will be left at a strange rotation until the user moves:
Screenshot 2025-01-06 at 4 37 29 PM

This will also happen when parenting to a newly-created entity (see https://github.com/space-wizards/space-station-14/pull/33878#issuecomment-2544274891)

To fix this, a handler for `ComponentStartup` is added to the clientside TransformSystem, which calls `SetGridId` on any children of the parent. `SetGridId` handles recursing through children-of-children and aborting in the case of uninitialized transforms.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/TransformSystem.Component.cs
+++ b/Robust.Client/GameObjects/EntitySystems/TransformSystem.Component.cs
@@ -8,6 +8,21 @@ namespace Robust.Client.GameObjects;
 
 public sealed partial class TransformSystem
 {
+    protected override void OnCompStartup(EntityUid uid, TransformComponent xform, ComponentStartup args)
+    {
+        // When parenting a transform to an entity that doesn't yet exist on the client (ex: outside of PVS),
+        // the parenting will be done before the target entity's transform is initialized.
+        // The parent transforms will have their GridUids set on initialize, but this value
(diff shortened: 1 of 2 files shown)


ElectroJr:
While it does fix the specific issue with ghost following, I think this might still cause unexpected problems elsewhere. Entity/component initialization is a convoluted mess, and many components blindly expect things like TransformComponent to have already been properly set up by the time they do their startup, ~~which might be before the transform starts up and the fix that this PR adds is applied.~~
Edit: I forgot that transform is given special treatment and is actually meant to always start first, but either way its probably still better to fix it some other way

I think a better / more general fix would be to change the entity state application so that it makes sure that you never have an initialized entity getting attached to a pre-init entity. I.e., the client side entity creation logic should be more like the map-loading logic, where parents are always initialized & started before their children. I'm going to start working on a PR that tries to do that instead.

Tayrtahn:
Superseded by #5595

PR #5589, opened 2025-01-07
Fix component delta state auto-generation

ElectroJr:
Currently auto-generated component delta states have a bug in `CreateNewFullState()`, where instead of cloning an existing full state, applying the delta state, and then returning the new modified full state, it instead modifies the state that should be getting cloned. 

I.e., currently it autogenerates code like:
```cs
public sealed class AudioStart_FieldComponentState : IComponentDeltaState<AudioComponent_AutoState>
{
    public global::System.TimeSpan AudioStart = default!;

    public void ApplyToFullState(AudioComponent_AutoState fullState)
    {
                    fullState.AudioStart = AudioStart;
    }

    public AudioComponent_AutoState CreateNewFullState(AudioComponent_AutoState fullState)
    {
        var newState = new AudioComponent_AutoState
        {
            Flags = fullState.Flags,
            AudioStart = fullState.AudioStart,
            FileName = fullState.FileName,
            Params = fullState.Params,
        Auxiliary = fullState.Auxiliary,
            State = fullState.State,
            PauseTime = fullState.PauseTime,
            Global = fullState.Global,
        };

                    fullState.AudioStart = AudioStart;

        return newState;
    }
}
```
where the `fullState.AudioStart = AudioStart;` should be modifying `newState` instead.

This PR fixes that so that it now generates:
```cs
public sealed class AudioStart_FieldComponentState : IComponentDeltaState<AudioComponent_AutoState>
{
    public global::System.TimeSpan AudioStart = default!;

    public void ApplyToFullState(AudioComponent_AutoState fullState)
    {
        fullState.AudioStart = AudioStart;
    }

    public AudioComponent_AutoState CreateNewFullState(AudioComponent_AutoState fullState)
    {
        var newState = fullState.ShallowClone();
        ApplyToFullState(newState);
        return newState;
    }
}
```

This PR also makes some other changes:
- `CreateNewFullState` now only does a shallow-clone. I.e., it doesn't automatically create new collections like dictionaries/lists. AFAIK this shouldn't cause any issues and should help reduce allocations for replays.
- Applying a delta state now also results in `AfterAutoHandleStateEvent` getting raised. Currently it only gets raised when applying a full state, which I'm assuming was a bug.
- Changes some of the source generation to have nicer indentation.


aidargg32:
public sealed class AudioStart_FieldComponentState : IComponentDeltaState<AudioComponent_AutoState>
{
    public global::System.TimeSpan AudioStart = default!;

    // Applies the delta state to the full state
    public void ApplyToFullState(AudioComponent_AutoState fullState)
    {
        fullState.AudioStart = AudioStart; // Updates the full state with the delta
    }

    // Creates a new full state based on the existing one
    public AudioComponent_AutoState CreateNewFullState(AudioComponent_AutoState fullState)
    {
        var newState = fullState.ShallowClone(); // Creates a shallow copy of the full state
        ApplyToFullState(newState); // Applies the delta state to the new state
        return newState; // Returns the modified new state
    }
}

PR #5598, opened 2025-01-09
Make PVS overrides respect vismasks

ElectroJr:
Currently, if an entity is added to a session specific PVS override, it will ignore the normal visibility masks. This has lead to bugs where audio entities temporarily make entities that should be hidden visible to everyone in range of the sound (e.g., https://github.com/space-wizards/space-station-14/pull/29801). This PR makes PVS overrides respect visibility masks, except for any "force send" entities, which now ignore them.


Robust.Server/GameStates/PvsSystem.ToSendSet.cs
```diff
@@ -142,18 +142,31 @@ private bool AddEntity(PvsSession session, Entity<MetaDataComponent> entity, Gam
...
-
             // This can happen if some entity was some removed from it's parent while that parent was being deleted.
             // As a result the entity was marked for deletion but was never actually properly deleted.
-            EntityManager.QueueDeleteEntity(uid);
+
+            bool queued;
+            lock (_toDelete)
+            {
+                queued = EntityManager.IsQueuedForDeletion(uid) || _toDelete.Contains(uid);
+                _toDelete.Add(uid);
```
metalgearsloth: If it's queuedeld no need to add to _toDelete (at least I don't think server can un-queuedel entities?)

PJB3005:
> This has lead to bugs where audio entities temporarily make entities that should be hidden visible to everyone in range of the sound (e.g., https://github.com/space-wizards/space-station-14/pull/29801)

The actual bug here is that content was relying solely on PVS to avoid visibility of ghost entities, which is incorrect. Without making the root entity visible, how would audio entities know where to play with this change? Would they just not?

ElectroJr:
> The actual bug here is that content was relying solely on PVS to avoid visibility of ghost entities, which is incorrect. 

There is also a client-side system that should also be making ghosts invisible that I forgot about, which should've also been preventing that (i.e., see `GhostSystem.OnStartup`) . I guess I'll go try figure out why it wasn't working.

Edit:
Okay, for whatever fucking reason ghost visibility is set to true by default?

> Without making the root entity visible, how would audio entities know where to play with this change? Would they just not?

The audio ents wouldn't get sent to the client, so the sound wouldn't play. IMO that's what should happen, sound shouldn't suddenly override PVS visibility masks. And it also acts as an extra safety net that invisible ghosts aren't going around accidentally emitting sounds, though again that should really also be ensured by some other system. I.e., ghosts still shouldn't emit sounds when PVS is disabled.

PJB3005:
> There is also a client-side system that should also be making ghosts invisible that I forgot about, which should've also been preventing that (i.e., see `GhostSystem.OnStartup`) . I guess I'll go try figure out why it wasn't working.

The PR that added revenants broke it, IIRC.

PJB3005:
> The audio ents wouldn't get sent to the client, so the sound wouldn't play. IMO that's what should happen, sound shouldn't suddenly override PVS visibility masks. And it also acts as an extra safety net that invisible ghosts aren't going around accidentally emitting sounds, though again that should really also be ensured by some other system. I.e., ghosts still shouldn't emit sounds when PVS is disabled.

Yeah alright.

PR #5599, opened 2025-01-09
Add ChunkIndicesEnumerator override for Vector2i

metalgearsloth:
Other one is int only but need this for rectangles.


PJB3005:
Wait what? We have rectangularly-chunked grids?

metalgearsloth:
> Wait what? We have rectangularly-chunked grids?

It's possible, it's mainly if I pass in like, a room or the likes that may be rectangular and need a chunked version.

PJB3005:
> > Wait what? We have rectangularly-chunked grids?
> 
> It's possible, it's mainly if I pass in like, a room or the likes that may be rectangular and need a chunked version.

What? The rectangle should be the AABB passed to the function, *not* the chunk size. The chunk size is a property of the grid, not the area you are querying (rectangle).

PR #5603, opened 2025-01-12
Add function to check if localization culture was already loaded.

c4llv07e:
For dynamic localization change you need to know which cultures was already loaded.

This change allows to check that.


TheShuEd (approved):
Looks very good
edit:
oopsie, its robust PR, not content

PR #5604, opened 2025-01-13
Fix replay playback bugs

ElectroJr:
Fixes replays never creating implicit component states, which was causing exceptions when trying to apply physics component delta states. Also fixes a bug where entities were being initialized out of order, though I'm not sure if this actually caused issues, but its been making a debug assert added in #5343 fail.


Robust.Client/Replays/Playback/ReplayPlaybackManager.Update.cs
```diff
@@ -44,7 +44,8 @@ private void TickUpdateOverride(FrameEventArgs args)
             _gameState.UpdateFullRep(state, cloneDelta: true);
             var next = Replay.NextState;
             BeforeApplyState?.Invoke((state, next));
-            _gameState.ApplyGameState(state, next);
+            var created = _gameState.ApplyGameState(state, next);
+            _gameState.GenerateImplicitStates(created);
```
ElectroJr, line 48: This part is what fixes the replay delta state exceptions

PR #5606, opened 2025-01-14
Clear up warns in Color.cs

notquitehadouken:
This

image


PJB3005:
These aren't actual compiler warnings so I'm really not sure this is worth the hassle.

notquitehadouken:
> These aren't actual compiler warnings so I'm really not sure this is worth the hassle.

thank you Rider very cool 👍

PR #5616, opened 2025-01-18
Make GamePrototypeLoadManager send a single message with all uploaded prototypes in a single message

DrSmugleaf:
Otherwise every single individual upload from an admin causes a separate reload of the prototypes, which takes around 2-3 seconds on my machine each, and you are stuck connecting for potentially minutes on end


ElectroJr:
FWIW, #5612 should also make prototype loading faster, because ATM its always reloading **all** prototypes, instead of only the new or modified ones.

PR #5618, opened 2025-01-18
Fixing guidebook not resizable from left and right

mohamedwidar:
issue [34504](https://github.com/space-wizards/space-station-14/issues/34504) in space-wizards/space-station-14 needs this fix

It happens that the value DragMode.Move prevents the switch case from catching left and right because the code in ss14 looks like this 

```
var mode = DragMode.Move; ##note that removing this line from ss14 will leave the window unmovable##

if (Resizable)
{
    if (relativeMousePos.Y < DRAG_MARGIN_SIZE)
    {
        mode = DragMode.Top;
    }
    else if (relativeMousePos.Y > Size.Y - DRAG_MARGIN_SIZE)
    {
        mode = DragMode.Bottom;
    }

    if (relativeMousePos.X < DRAG_MARGIN_SIZE)
    {
        mode |= DragMode.Left;
    }
    else if (relativeMousePos.X > Size.X - DRAG_MARGIN_SIZE)
    {
        mode |= DragMode.Right;
    }
}

return mode;
```

so basically the equal sign removes the Move value in every other case leaving the guidebook window cursor not changing on left and right and here is a video of working solution 

https://drive.google.com/file/d/1UTgJGoV_96P2-TyJa1No1gHP9FXIL7Kd/view?usp=sharing

and i also fixed the corresponding code on ss14 repo and will make another pull request there

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -43,7 +43,7 @@ END TEMPLATE-->
 
 ### Bugfixes
 
-*None yet*
+* Fixed `BaseWindow` sometimes not properly updating the mouse cursor shape.
 
 ### Other
 
--- a/Robust.Client/UserInterface/CustomControls/BaseWindow.cs
+++ b/Robust.Client/UserInterface/CustomControls/BaseWindow.cs
@@ -102,27 +102,23 @@ namespace Robust.Client.UserInterface.CustomControls
             {
                 var cursor = CursorShape.Arrow;
                 var previewDragMode = GetDragModeFor(args.RelativePosition);
-                previewDragMode &= ~DragMode.Move;
                 switch (previewDragMode)
                 {
-                    case DragMode.Top:
-                    case DragMode.Bottom:
-                        cursor = CursorShape.VResize;
+                    case var _ when (previewDragMode & (DragMode.Bottom | DragMode.Left)) == (DragMode.Bottom | DragMode.Left):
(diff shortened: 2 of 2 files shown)


lzk228:
i also discovered that silicon laws menu also couldn't be resized on sides
can you add it to this pr so i shouldn't create a new issue?

lzk228:
oh wait, this change should apply to all windows, sorry

ElectroJr:
I haven't ever looked at window resizing code before, but from what I see IMO the current engine code is probably fine as is? The "Move" and "Resize" drag operations are always going to be mutually exclusive, and IMO it makes sense that if the move flag is set, it takes priority over resizing.

Instead I think the problem is just that space-wizards/space-station-14/pull/12681 gave `FancyWindow` a bad `GetDragModeFor()` implementation and it should probably just match the `DefaultWindow` method? And by the sounds of it you were already working on a content PR to fix that.

Though it also seems like FancyWindow doesn't have large enough margins/padding on the window contents. I.e., default window has `Margin="10"` (10 on all sides), while fancy window only has `Margin="0 2"` (i.e., 0 left/right, 2 top/bottom), which also makes it harder to drag windows.

mohamedwidar:
this doesn't really change any behavior or logic. I just made sure the cursor changes shape if it has the right combination of flags. even before the change the window was resizable just fine without the cursor changing shape.

PR #5625, opened 2025-01-21
fix some issues with ClientDisconnect

MilonPL:
fixes an issue with ClientDisconnect (and by extension the DisconnectCommand)

turns out that DisconnectCommand is sort of a fucky-wucky
- If you use it while attempting to connect to a server, you get stuck in quantum superposition state of not connecting while connecting

image

- if you use it while attempting to connect in dev env, you get stuck in a void

image

the issue was that we weren't properly handling disconnects during active connection attempts - the code assumed you were already connected when trying to disconnect (2017 code my beloved)

now you can actually disconnect while connecting without breaking the universe 👍

### DIFF
--- a/Robust.Shared/Network/NetManager.ClientConnect.cs
+++ b/Robust.Shared/Network/NetManager.ClientConnect.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
@@ -11,7 +10,6 @@ using System.Security.Cryptography;
 using System.Threading;
 using System.Threading.Tasks;
 using Lidgren.Network;
-using Robust.Shared.Log;
 using Robust.Shared.Network.Messages.Handshake;
 using Robust.Shared.Utility;
 using SpaceWizards.Sodium;
@@ -98,7 +96,7 @@ namespace Robust.Shared.Network
             {
                 await CCDoHandshake(winningPeer, winningConnection, userNameRequest, mainCancelToken);
             }
-            catch (TaskCanceledException)
+            catch (Exception e) when (e is OperationCanceledException or TaskCanceledException)
             {
                 winningPeer.Peer.Shutdown("Cancelled");
                 _toCleanNetPeers.Add(winningPeer.Peer);
@@ -120,7 +118,10 @@ namespace Robust.Shared.Network
             _logger.Debug("Handshake completed, connection established.");
         }
 
-        private async Task CCDoHandshake(NetPeerData peer, NetConnection connection, string userNameRequest,
+        private async Task CCDoHandshake(
+            NetPeerData peer,
+            NetConnection connection,
+            string userNameRequest,
             CancellationToken cancel)
         {
             var encrypt = _config.GetCVar(CVars.NetEncrypt);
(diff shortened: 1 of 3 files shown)


Robust.Shared/Network/NetManager.cs
```diff
@@ -587,6 +587,21 @@ public void ProcessPackets()
...
+            if (ClientConnectState != ClientConnectionState.NotConnecting)
+            {
+                _cancelConnectTokenSource?.Cancel();
+
+                // Clean up any pending net peers
+                foreach (var peer in _netPeers)
+                {
+                    peer.Peer.Shutdown(reason);
+                    _toCleanNetPeers.Add(peer.Peer);
+                }
```
PJB3005: Shouldn't this already be handled by `Reset`? Or does this need to happen before the `Disconnect` event fired?

MilonPL: Okay so after some testing, that bit *is* redundant, however there's still an issue with the network not resetting properly if we cancel a connection attempt to a server that hasn't started yet (for example direct connecting to `localhost` while it's still loading and then disconnecting)
I'll see if I can fix that too. Also in the NotConnecting state we should probably include a button to retry connection then? I'm not sure if it exists currently.

MilonPL:
@PJB3005 ready for re-review
Fixed it by handling cancellation in CCHappyEyeballs, turns out we weren't catching cancelled connection attempts during the handshake when connecting to non-responsive servers

PJB3005 (commented):
Adding that try catch to `AwaitNonInitStatusChange` doesn't seem like it makes a lot of sense to me. 

Reading the code, there's *already* a try catch for cancellation around most of the logic, but it seems like it's erroneously catching on `TaskCanceledException` instead of `OperationCanceledException` (definitely my fault). Does fixing that fix it?

PJB3005:
Also figured I'd mention this: we can probably replace most of the logic in `CCHappyEyeballs` with usage of the `HappyEyeballsHttp.ParallelTask<T>` that I've since created. But it's really not that important so whatever.

Robust.Shared/Network/NetManager.ClientConnect.cs
```diff
@@ -98,7 +96,7 @@ public async void ClientConnect(string host, int port, string userNameRequest)
             {
                 await CCDoHandshake(winningPeer, winningConnection, userNameRequest, mainCancelToken);
             }
-            catch (TaskCanceledException)
+            catch (Exception e) when (e is OperationCanceledException or TaskCanceledException)
```
PJB3005: `TaskCanceledException` is a subtype of `OperationCanceledException`, you don't need to check both of these and don't need a `when` expression.

Robust.Shared/Network/NetManager.ClientConnect.cs
```diff
@@ -289,171 +290,91 @@ private byte[] MakeAuthHash(byte[] sharedSecret, byte[] pkBytes)
...
-                {
-                    // Only one address to try. Pretty straight forward.
-                    firstReason = await AwaitNonInitStatusChange(firstConnection, mainCancelToken);
-                    winningPeer = firstPeer;
-                    winningConnection = firstConnection;
-                }
+                return (result.Peer, result.Connection);
             }
-            catch (TaskCanceledException)
+            catch (Exception e) when (e is OperationCanceledException or TaskCanceledException)
```
PJB3005: Ditto with the canceled exceptions.

Robust.Shared/Network/NetManager.ClientConnect.cs
```diff
@@ -536,5 +458,17 @@ private Task<NetIncomingMessage> AwaitData(NetConnection connection,
...
+        [Virtual]
+        private class ConnectionAttempt(NetPeerData peer, NetConnection connection) : IDisposable
+        {
+            public NetPeerData Peer { get; } = peer;
+            public NetConnection Connection { get; } = connection;
+
+            public void Dispose()
+            {
+                Peer.Peer.Shutdown("Disposing unused connection attempt");
+            }
```
PJB3005: Don't you need to add it to `_toCleanNetPeers` here?

Robust.Shared/Network/NetManager.ClientConnect.cs
```diff
@@ -536,5 +458,17 @@ private Task<NetIncomingMessage> AwaitData(NetConnection connection,
         }
 
         private sealed record JoinRequest(string Hash, string? Hwid);
+
+        [Virtual]
+        private class ConnectionAttempt(NetPeerData peer, NetConnection connection) : IDisposable
```
PJB3005: Make this `sealed` instead of `[Virtual]`.

PJB3005 (changes requested):

PJB3005 (approved):

PR #5629, opened 2025-01-23
Add normal maps

notquitehadouken:
title
yes i'm trying again

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -1289,7 +1289,7 @@ namespace Robust.Client.GameObjects
         [ViewVariables(VVAccess.ReadWrite)]
         public bool NoRotation { get => _screenLock; set => _screenLock = value; }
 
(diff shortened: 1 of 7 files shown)


notquitehadouken:
image

notquitehadouken:
i screamed when i saw this

image

PJB3005:
Please only open a PR when you have a basic prototype working...

K-Dynamic:
What would a normal map do and look like in SS14

notquitehadouken:
> What would a normal map do and look like in SS14

lights but cooler

Cinkafox:
Can you provide an example of this think?

PJB3005:
Normal maps in 2D are totally possible, though they may require a lot of artist work. https://theartsquirrel.com/30/digital-art-technique-using-normal-and-depth-maps-to-create-lighting-effects-in-2d-art/

The problem is that implementing them with SS14's renderer would require a complete rewrite right now. This PR is nowhere near actually working for even a prototype, so I'd rather not have it clog up the repo.

notquitehadouken:
> The problem is that implementing them with SS14's renderer would require a complete rewrite right now. 

im magic

> This PR is nowhere near actually working for even a prototype, so I'd rather not have it clog up the repo.

fair enough

notquitehadouken:
i almost have a working prototype (90% of this was figuring out how to autogenerate the normals)

PR #5632, opened 2025-01-24
Cancellable audio

### LINKED ISSUE #5611: AudioSystem should raise a cancellable event to block audio from playing
There should be some way for content to prevent entities from making sounds. E.g., admin ghosts shouldn't make noise. 
While it could just be implemented in content by ensuring that checks are performed before any audio system methods are called, it'd probably be much easier to just make the audio methods raise an attempt event. I.e., `PlayEntity` and `PlayPvs` should raise a cancellable event directed at the entity that the sound is being attached to. For audio being attached to a position rather than entity, I don't think there's an easy fix so that'd be up to content to block the method from being called.

JerryImMouse:
Added cancellable event to cancel attempts of playing audio

Resolves #5611

### DIFF
--- a/Robust.Shared/Audio/Components/AudioComponent.cs
+++ b/Robust.Shared/Audio/Components/AudioComponent.cs
@@ -275,3 +275,8 @@ public enum AudioFlags : byte
 
     NoOcclusion = 1 << 1,
 }
+
+/// <summary>
+/// Raised directed at an entity, which is emitting the audio. Raised only when audio played at the entity itself, not the coordinates.
+/// </summary>
+public sealed class AudioPlayAttempt : CancellableEntityEventArgs;
--- a/Robust.Shared/Audio/Components/AudioComponent.cs
+++ b/Robust.Shared/Audio/Components/AudioComponent.cs
@@ -279,4 +279,5 @@ public enum AudioFlags : byte
 /// <summary>
 /// Raised directed at an entity, which is emitting the audio. Raised only when audio played at the entity itself, not the coordinates.
 /// </summary>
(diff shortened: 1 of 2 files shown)


JerryImMouse:
idk if I should raise this on client-side too, let me know if its needed

Robust.Shared/Audio/Components/AudioComponent.cs
```diff
@@ -275,3 +275,8 @@ public enum AudioFlags : byte
 
     NoOcclusion = 1 << 1,
 }
+
+/// <summary>
+/// Raised directed at an entity, which is emitting the audio. Raised only when audio played at the entity itself, not the coordinates.
+/// </summary>
+public sealed class PlayAudioAttempt : CancellableEntityEventArgs;
```
metalgearsloth: Struct event plz.

JerryImMouse: done

Robust.Shared/Audio/Components/AudioComponent.cs
```diff
@@ -275,3 +275,9 @@ public enum AudioFlags : byte
 
     NoOcclusion = 1 << 1,
 }
+
+/// <summary>
+/// Raised directed at an entity, which is emitting the audio. Raised only when audio played at the entity itself, not the coordinates.
+/// </summary>
+[ByRefEvent]
+public record struct PlayAudioAttemptEvent(bool Cancelled = false);
```
metalgearsloth, line 283: The clientside equivalents to PlayPvs should also be raising this as well.

PR #5633, opened 2025-01-25
Add helper functions to FormattedMessage

SlamBamActionman:
This PR adds extra functions to FormattedMessage that allows for greater manipulation of the markup node list. All new functions should be safe, properly closing any new opened tags and dealing with text nodes appropriately.

`InsertAtIndex` - Inserts a markup node at a specific index in the list. To avoid issues with unclosed tags, any node that isn't a text node is required to also define the index of the closing node.

`InsertAroundMessage` - Wraps the message with a non-text node. Good for full-color text.

`InsertAroundText` - Wraps the first-to-last text node with a non-text node.

`InsertAroundString` - Wraps every *occurence* of a string with a non-text node. The string must be fully contained in a text node (i.e. no strings separated by other nodes). Useful for codewords.

`InsertBeforeTag`/`InsertAfterTag` -  Inserts a self-closing node in front of/behind every node with a specific `Name`.

`InsertOutsideTag`/`InsertInsideTag` - Wraps every node/the contents of every node with a specific `Name`.

`InsertBeforeMessage` - `InsertAtIndex` at index 0.

`InsertAfterMessage` - `PushTag` with self-closing set to true.

`TryGetMessageInsideTag` - Gets all nodes inside of the first tag with a specific `Name`. Useful for stuff like speech bubbles.


Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +235,245 @@ public void Clear()
...
+        }
+    }
+
+    /// <summary>
+    /// Helper function that inserts a node around other nodes of a specific type.
+    /// The new node encloses any other nodes that the target node encloses.
+    /// </summary>
+    /// <param name="markupNode">The node to be inserted; may not be a text node.</param>
+    /// <param name="tagText">The tag to search for.</param>
+    public void InsertOutsideTag(MarkupNode markupNode, string tagText)
```
juliangiebel, line 508: What happens if you, for example, try to wrap `color` when then markup looks like this:
`[color]text1[color/][color]text2[color/]`

Wouldn't the second iteration cause `InsertAtIndex` to throw?

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +235,245 @@ public void Clear()
...
+    {
+        PushTag(markupNode, true);
+    }
+
+    /// <summary>
+    /// Helper function that tries to find the first instance of a tag and returns a FormattedMessage containing the nodes inside.
+    /// </summary>
+    /// <param name="returnMessage">The node to be inserted.</param>
+    /// <param name="tagText">The tag to search for.</param>
+    public bool TryGetMessageInsideTag(out FormattedMessage? returnMessage, string tagText)
```
juliangiebel: Same question as `InsertOutsideTag`

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +235,245 @@ public void Clear()
...
+        }
+    }
+
+    /// <summary>
+    /// Helper function that inserts a node inside of other nodes of a specific type.
+    /// The new node encloses any other nodes that the target node encloses.
+    /// </summary>
+    /// <param name="markupNode">The node to be inserted; may not be a text node.</param>
+    /// <param name="tagText">The tag to search for.</param>
+    public void InsertInsideTag(MarkupNode markupNode, string tagText)
```
juliangiebel, line 459: Same question as `InsertOutsideTag`

juliangiebel (changes requested):
Looks good besides a potential issue I found with how you try to find tags in certain helper methods.

If that turns out to be an actual issue you might want to traverse the node array from start to end and for every opening tag that matches you traverse again from that index, keeping a counter that gets incremented for every additional occurrence of a matching opening tag and decremented for every closing tag. If the counter is 0 when encountering a matching closing tag you've found the closing tag that belongs to the opening tag you initially matched.

A flat tree like this isn't very good for operations like that 🤔

I somewhat mentioned this in my initial writeup about richtext: https://hackmd.io/@juliang/r1lzQwPRss

SlamBamActionman:
I've implemented changes based on the review comments.

The way the InsertInside/OutsideTag functions work is that they first ensure there's equal amounts of opening and closing target tags (there always should in any correct message, so the check is arguably redudant and could be removed!), and then it simply adds an instance of the new tag before/after every instance of the target tag. 

TryGetMessageInsideTag was indeed causing issues with multiple instances of a tag. Now it appropriately gets the contents of the first tag.

juliangiebel (approved):

SlamBamActionman:
Based on Julian giving approval, I'm going ahead to merge this. Will be necessary for the chat refactor. 

EDIT: Nevermind, don't got RT merge rights :godo:

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+    {
+        PushTag(markupNode, true);
+    }
+
+    /// <summary>
+    /// Helper function that tries to find the first instance of a tag and returns a FormattedMessage containing the nodes inside.
+    /// </summary>
+    /// <param name="returnMessage">The message inside of the chosen tag.</param>
+    /// <param name="tagText">The tag to search for.</param>
+    public bool TryGetMessageInsideTag(out FormattedMessage? returnMessage, string tagText)
```
Fildrance: usually TryXXX pattern lets 'out' argument be on second place

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+    {
+        returnMessage = null;
+
+        var openingNode = _nodes.FindIndex(x => x.Name == tagText && !x.Closing);
```
Fildrance: naming  it 'openingNodeIndex' would possibly be more correct, as could be misunderstood later in code.

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+        var nextNode = openingNode;
+        var nodeCount = 1;
+
+        if (openingNode == -1)
```
Fildrance: If you can exit early without assigning variables - then better do it. Also declaring variables closer to place they are used in would improve readability

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+    /// Helper function that inserts a node after the entire message.
+    /// </summary>
+    /// <param name="markupNode">The node to be inserted.</param>
+    public void InsertAfterMessage(MarkupNode markupNode)
+    {
+        PushTag(markupNode, true);
+    }
+
+    /// <summary>
+    /// Helper function that tries to find the first instance of a tag and returns a FormattedMessage containing the nodes inside.
```
Fildrance: Kinda hanging question - don't we want any kind of overload that will return N'th inclusion? or all of them, paired with indices? Well, maybe out of scope here :<

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+        }
+    }
+
+    /// <summary>
+    /// Helper function that inserts a node around other nodes of a specific type.
+    /// The new node encloses any other nodes that the target node encloses.
+    /// </summary>
+    /// <param name="markupNode">The node to be inserted; may not be a text node.</param>
+    /// <param name="tagText">The tag to search for.</param>
+    public void InsertOutsideTag(MarkupNode markupNode, string tagText)
```
Fildrance, line 508: wouldn't it be better to say 'wrap'? then try to stay consistent with using 'wrap' and 'nest' as put outside/put inside? just an idea

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+    {
+        ArgumentNullException.ThrowIfNull(markupNode.Name);
+
+        var openingNodeCount = _nodes.Count(x => x.Name == tagText && !x.Closing);
+        var closingNodeCount = _nodes.Count(x => x.Name == tagText && x.Closing);
+
+        if (openingNodeCount != closingNodeCount)
+            throw new Exception("Opening and Closing node count mismatch.");
```
Fildrance: Also throwing general Exception type is not nice, at least use InvalidOperationException :< or maybe we have some custom stuff fitting...

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+
+            if (i + 2 >= _nodes.Count)
+                break;
+
+            i = _nodes.FindIndex(i + 2, x => x.Name == tagText && !x.Closing);
+        }
+    }
+
+    /// <summary>
+    /// Helper function that inserts a node after tags of a specific type.
```
Fildrance: It might be better to explicitly say 'after each occurrance' or something.

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+    /// <param name="matchCase">If true, the string case must match exactly.</param>
+    public void InsertAroundString(MarkupNode markupNode, string stringText, bool matchCase = true)
+    {
+        ArgumentNullException.ThrowIfNull(markupNode.Name);
+
+        for (int i = 0; i < _nodes.Count; i++)
+        {
+            var node = _nodes[i];
+
+            if (node.Name == null && node.Value.StringValue != null)
```
Fildrance: invert if for less nesting

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+            {
+                var stringNode = node.Value.StringValue;
+                var stringLocation = matchCase ? stringNode.IndexOf(stringText, StringComparison.Ordinal) : stringNode.IndexOf(stringText, StringComparison.OrdinalIgnoreCase);
+
+                if (stringLocation != -1)
```
Fildrance: same, invert if, less nesting => better readability

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+    /// Helper method that inserts a node around a string.
+    /// </summary>
+    /// <param name="markupNode">The node to be inserted; may not be a text node.</param>
+    /// <param name="stringText">The string to look for when inserting.</param>
+    /// <param name="matchCase">If true, the string case must match exactly.</param>
+    public void InsertAroundString(MarkupNode markupNode, string stringText, bool matchCase = true)
+    {
+        ArgumentNullException.ThrowIfNull(markupNode.Name);
+
+        for (int i = 0; i < _nodes.Count; i++)
```
Fildrance: if you insert inside cycle, wouldn't this check be incorrect? like u will have more nodes then original Count at the end...

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+        {
+            var node = _nodes[i];
+
+            if (node.Name == null && node.Value.StringValue != null)
+            {
+                var stringNode = node.Value.StringValue;
+                var stringLocation = matchCase ? stringNode.IndexOf(stringText, StringComparison.Ordinal) : stringNode.IndexOf(stringText, StringComparison.OrdinalIgnoreCase);
```
Fildrance: call it 'indexOfStringNode', or 'foundIndex', we have names for that stuff <_<

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+    {
+        ArgumentNullException.ThrowIfNull(markupNode.Name);
+
+        for (int i = 0; i < _nodes.Count; i++)
+        {
+            var node = _nodes[i];
+
+            if (node.Name == null && node.Value.StringValue != null)
+            {
+                var stringNode = node.Value.StringValue;
```
Fildrance: nodeText maybe?

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+                if (stringLocation != -1)
+                {
+                    // Original node needs to be removed, to make room for the new ones.
+                    _nodes.RemoveAt(i);
+
+                    // j keeps track of the number of added nodes.
+                    var j = 0;
+
+                    var beforeText = stringNode.Substring(0, stringLocation);
+                    if (beforeText != "")
```
Fildrance: not sure about convention, but String.Empty could be more descriptive and eliminate misunderstandings (reader could think in some cases that empty quotes was a mistake - developer forgot to add text in quotes - and not the real intention). But really it should be part of code convention.

Fildrance: Other option is String.IsNullOrEmpty / IsNullOrWhitespace, but they are quite specific, and might be add useless workload

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -235,6 +236,295 @@ public void Clear()
...
+                    j += 3;
+
+                    var afterText = stringNode.Substring(stringLocation + stringText.Length);
+                    if (afterText != "")
+                    {
+                        InsertAtIndex(new MarkupNode(afterText), i + j);
+                    }
+
+                    // We make sure to move the i forwards, searching for additional instances of the string.
+                    i += j - 2;
```
Fildrance, line 341: maybe keep j outside of loop and clean its value at start, but add it in for statement? currently you have 'i' getting incremented in 2 places, this might be disorienting to reader.

Fildrance (commented):
Remarked some possible improvements in readability :<
Might have missed a ton more stuff, direct interaction with string indexes and loops with duplicates of find calls are a bit wierd, but cannot think of any clearly better fix. It will work, ofc.
Either way, i think this can work better with unit tests, which most of RT code should have <_<

Robust.Shared/Utility/FormattedMessage.cs
Fildrance: Also, simple test shows that it actually is returning message STARTIN AND ENDING with provided tag, like
getting message out of "[honk]Foo[color=red] bar[/honk] [/color] bar baz" by tag "honk" would lead to "[honk]Foo[color=red] bar[/honk]". So either description and method name or the logic of method should be fixed :<

Robust.Shared/Utility/FormattedMessage.cs
Tayrtahn: Also typically uses `[NotNullWhen(true)]`

Fildrance: oh yeah, forgot that Q_Q @SlamBamActionman it should look like 
```public bool TryGetMessageInsideTag(string tagText, [NotNullWhen(true)] out FormattedMessage? result)```
the attribute is a hint for compiler that when method returns TRUE, that out parameter will NOT be null, and so compiler will not require extra null checks.

File: Robust.Shared/Utility/FormattedMessage.cs (hunk 16 above)
Comment (Fildrance) [reply Robust.Shared/Utility/FormattedMessage.cs]: fixed

File: Robust.Shared/Utility/FormattedMessage.cs (hunk 15 above)
Comment (Fildrance) [reply Robust.Shared/Utility/FormattedMessage.cs]: fixed

File: Robust.Shared/Utility/FormattedMessage.cs (hunk 14 above)
Comment (Fildrance) [reply Robust.Shared/Utility/FormattedMessage.cs]: fixed

File: Robust.Shared/Utility/FormattedMessage.cs (hunk 5 above)
Comment (Fildrance) [reply Robust.Shared/Utility/FormattedMessage.cs]: fixed

File: Robust.Shared/Utility/FormattedMessage.cs (hunk 6 above)
Comment (Fildrance) [reply Robust.Shared/Utility/FormattedMessage.cs]: fixed

File: Robust.Shared/Utility/FormattedMessage.cs (hunk 10 above)
Comment (Fildrance) [reply Robust.Shared/Utility/FormattedMessage.cs]: fixed

File: Robust.Shared/Utility/FormattedMessage.cs (hunk 12 above)
Comment (Fildrance) [reply Robust.Shared/Utility/FormattedMessage.cs]: fixed

File: Robust.Shared/Utility/FormattedMessage.cs (hunk 11 above)
Comment (Fildrance) [reply Robust.Shared/Utility/FormattedMessage.cs]: fixed

File: Robust.Shared/Utility/FormattedMessage.cs (hunk 13 above)
Comment (Fildrance) [reply Robust.Shared/Utility/FormattedMessage.cs]: fixed

File: Robust.Shared/Utility/FormattedMessage.cs (hunk 9 above)
Comment (Fildrance) [reply Robust.Shared/Utility/FormattedMessage.cs]: fixed

File: Robust.Shared/Utility/FormattedMessage.cs [hunk 18]
```diff
@@ -260,6 +270,379 @@ public void Clear()
...
+            i += 1 + j - 2 ;
+        }
+    }
+
+    /// <summary>
+    /// Helper function that inserts a node before each opening tag of a specific type.
+    /// </summary>
+    /// <param name="markupNode">The node to be inserted.</param>
+    /// <param name="tagText">The tag to search for.</param>
+    public void InsertBeforeTag(MarkupNode markupNode, string tagText)
```
PJB3005, line 413: `tagText` should probably be called `tagName` instead here, to avoid confusion.

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -260,6 +270,379 @@ public void Clear()
...
+            i = _nodes.FindIndex(i + offset, x => x.Name == tagText && !x.Closing);
+        }
+    }
+
+    /// <summary>
+    /// Helper function that inserts a node after each closing tags of a specific type.
+    /// </summary>
+    /// <param name="markupNode">The node to be inserted.</param>
+    /// <param name="tagText">The tag to search for.</param>
+    public void InsertAfterTag(MarkupNode markupNode, string tagText)
```
PJB3005, line 435: `tagText` should probably be called `tagName` instead here, to avoid confusion.

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -260,6 +270,379 @@ public void Clear()
...
+        }
+    }
+
+    /// <summary>
+    /// Helper function that nests a node inside other nodes of a specific type.
+    /// The new node encloses any other nodes that the target node encloses.
+    /// </summary>
+    /// <param name="markupNode">The node to be inserted; may not be a text node.</param>
+    /// <param name="tagText">The tag to search for.</param>
+    public void InsertInsideTag(MarkupNode markupNode, string tagText)
```
PJB3005, line 459: `tagText` should probably be called `tagName` instead here, to avoid confusion.

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -260,6 +270,379 @@ public void Clear()
...
+        }
+
+        var indexToReplace = _nodes.IndexOf(nodeToReplace);
+        if (indexToReplace == -1)
+        {
+            return;
+        }
+
+        _nodes[indexToReplace] = replacementNode;
+    }
```
PJB3005, line 642: This feels like an arbitrarily specific API? Why can't this take non-text nodes?

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -260,6 +270,379 @@ public void Clear()
...
+    {
+        PushTag(markupNode, true);
+    }
+
+    /// <summary>
+    /// Helper function that tries to find the first instance of a tag and returns
+    /// a FormattedMessage containing the found node with all nodes, nested inside.
+    /// </summary>
+    /// <param name="result">The message with all markup nodes inside chosen tag.</param>
+    /// <param name="tagText">The tag to search for.</param>
```
PJB3005, line 575: This doesn't explain when this returns false.

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -260,6 +270,379 @@ public void Clear()
...
+    public bool TryGetMessageInsideTag(string tagText, [NotNullWhen(true)] out FormattedMessage? result)
```
PJB3005, line 576: `tagText` should probably be called `tagName` instead here, to avoid confusion.

Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -260,6 +270,379 @@ public void Clear()
...
+            i += 2;
+        }
+    }
+
+    /// <summary>
+    /// Helper function that wraps a node around other nodes of a specific type.
+    /// </summary>
+    /// <param name="markupNode">The node to be inserted; may not be a text node.</param>
+    /// <param name="tagText">The tag to search for.</param>
+    public void InsertOutsideTag(MarkupNode markupNode, string tagText)
```
PJB3005, line 508: `tagText` should probably be called `tagName` instead here, to avoid confusion.

Robust.UnitTesting/Shared/Utility/FormattedMessage_Test.cs
```diff
@@ -83,5 +102,479 @@ public static void TestEnumerateRunes(string text)
...
+        public static void TestInsertAtIndex_WithTagEndLesserThenStart_Throws()
+        {
+            var message = FormattedMessage.FromMarkupOrThrow("Foo[color=red] bar [/color] bar baz");
+            Assert.Throws<ArgumentException>(() =>
+                message.InsertAtIndex(new MarkupNode("honk", null, null), 2, 1)
+            );
+        }
+
+        [Test]
+        public static void TestInsertAroundMessage_InsertPlanText_Throws()
```
PJB3005, line 181:
```suggestion
        public static void TestInsertAroundMessage_InsertPlainText_Throws()
```

PJB3005 (changes requested):
Looks good barring some nits.

PR #5642, opened 2025-01-29
fix external profilers crashing when starting the server

MilonPL:
when launching through external profilers like ANTS or dotTrace in modes other than sampling/timeline, because of ???? we get an invalid handle error when trying to open files that don't exist, causing the entire program to crash

added a check for running under a profiler (yes i hate the nullable bool and scanning modules, but i dont know of a better way to do this) then we always check if the file exists in those cases

normal execution is still fast, we only eat the performance hit when running under these profilers (wont affect timeline though if you attach it later)

### DIFF
--- a/Robust.Shared/Utility/FileHelper.cs
+++ b/Robust.Shared/Utility/FileHelper.cs
@@ -1,6 +1,8 @@
 ﻿using System;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using Microsoft.Win32.SafeHandles;
 using TerraFX.Interop.Windows;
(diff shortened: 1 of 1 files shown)


Robust.Shared/Utility/FileHelper.cs
```diff
@@ -44,6 +49,21 @@ private static unsafe bool TryGetFileWindows(string path, [NotNullWhen(true)] ou
             return false;
         }
 
+        // When using external profilers like dotTrace in line-by-line mode, this shit breaks
+        // So we have to do the slow File.Exists() check in those cases because what else can we do?
+        _isUnderProfiler ??= Process.GetCurrentProcess()
+            .Modules
+            .Cast<ProcessModule>()
+            .Any(m => m.ModuleName.Contains("JetBrains") ||
```
FairlySadPanda, line 57: Don't hardcode these.

FairlySadPanda (commented):
Definitely a PJB one this, just a quick suggestion on avoiding hardcoding

MilonPL:
I'm not sure if there's a better way to check for a profiler, and I can't hardcode in all of them. I guess maybe a custom command line argument? But also seems unnecessarily complex

metalgearsloth:
Agreed on the review but uhh need steps to repro or smth.

MilonPL:
\>open JetBrains Rider 
\>run the project with a profiler attached in the line-by-line mode
\>it crashes

Issue #5596, opened 2025-01-08
MapChunk.Fixtures can contain non-existent fixtures


ElectroJr:
[Reported on discord](https://discord.com/channels/310555209753690112/900426319433728030/1324107945826451591)

Can cause the client to freeze due to exceptions while rendering. It'd be easy to turn the dictionary access into a try-get, but that's just hiding some other (potentially also severe) bug under the rug.

```[ERRO] runtime: Caught exception in "entsys"
System.Collections.Generic.KeyNotFoundException: The given key 'grid_chunk-32'01-0'01' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Robust.Shared.Map.MapManager.IsIntersecting(ChunkEnumerator enumerator, IPhysShape shape, Transform shapeTransform, Entity`1 grid) in /home/runner/work/RobustToolbox/RobustToolbox/Robust.Shared/Map/MapManager.Queries.cs:line 29
   at Robust.Shared.Map.MapManager.<>c__65`1.<FindGridsIntersecting>b__65_0(GridQueryState`1& state, Proxy proxy) in /home/runner/work/RobustToolbox/RobustToolbox/Robust.Shared/Map/MapManager.Queries.cs:line 174
   at Robust.Shared.Physics.B2DynamicTree`1.Query[TState](TState& state, QueryCallback`1 callback, Box2& aabb) in /home/runner/work/RobustToolbox/RobustToolbox/Robust.Shared/Physics/B2DynamicTree.cs:line 894
   at Robust.Shared.Map.MapManager.FindGridsIntersecting[TState](EntityUid mapEnt, IPhysShape shape, Box2 worldAABB, Transform transform, TState& state, GridCallback`1 callback, Boolean approx, Boolean includeMap) in /home/runner/work/RobustToolbox/RobustToolbox/Robust.Shared/Map/MapManager.Queries.cs:line 139
   at Robust.Shared.Map.MapManager.FindGridsIntersecting[TState](EntityUid mapEnt, Box2 worldAABB, TState& state, GridCallback`1 callback, Boolean approx, Boolean includeMap) in /home/runner/work/RobustToolbox/RobustToolbox/Robust.Shared/Map/MapManager.Queries.cs:line 227
[...]

PR #5644, opened 2025-01-31
Tweak grid state slightly

metalgearsloth:
Doubt it affects anything but if tile changes throws it at least ensures the fixtures are updated.


ElectroJr:
Possibly fixes #5596? Though obviously wouldn't fix whatever is actually causing the exceptions, which might be some other engine bug. And I don't think the fixture related logs I've seen have ever included some other relevant exception that might've caused the errors in #5596 to start.

PR #5654, opened 2025-02-04
add a method for copying components

MilonPL:
title, the result of me being sleep deprived

adds EntityManager methods to either copy a component from one entity to another or copy multiple components from params

the code itself is pretty simple, the EntitySystem proxy methods are `CopyComp` and `CopyComps`

thanks to Sloth, FairlySloggersPanda, Kickguy and everyone else who helped make this code


keronshb:
Looks good to me but would want another engine maintainer to take a look.

keronshb (approved):

Robust.Shared/GameObjects/EntityManager.Components.cs
```diff
@@ -1028,6 +1028,59 @@ public bool TryGetComponent([NotNullWhen(true)] EntityUid? uid, ushort netId,
...
+        {
+            component = default;
+            if (CopyComponent(source, target, typeof(T), out var baseComponent, metadataTarget))
+            {
+                component = (T)baseComponent;
+                return true;
+            }
+            return false;
+        }
+
```
metalgearsloth: Generic copy is going to be faster than using typeof.

Robust.Shared/GameObjects/EntityManager.Components.cs
```diff
@@ -1028,6 +1028,59 @@ public bool TryGetComponent([NotNullWhen(true)] EntityUid? uid, ushort netId,
...
+        public bool CopyComponent(EntityUid source, EntityUid target, Type type, [NotNullWhen(true)] out IComponent? component, MetaDataComponent? metadataTarget = null)
+        {
+            component = null;
+
+            if (!MetaQuery.Resolve(target, ref metadataTarget, false))
+                throw new ArgumentException($"Entity {target} is not valid.", nameof(target));
+
+            if (!typeof(IComponent).IsAssignableFrom(type))
+                return false;
+
```
metalgearsloth: This should be a debug assert.

Robust.Shared/GameObjects/EntityManager.Components.cs
```diff
@@ -1028,6 +1028,59 @@ public bool TryGetComponent([NotNullWhen(true)] EntityUid? uid, ushort netId,
             return TryGetComponent(uid.Value, netId, out component, meta);
         }
 
+        /// <inheritdoc/>
+        public bool CopyComponent(EntityUid source, EntityUid target, Type type, [NotNullWhen(true)] out IComponent? component, MetaDataComponent? metadataTarget = null)
+        {
+            component = null;
+
```
metalgearsloth: If you use the IComponent interface you could probably have the generic method call this one as well.

metalgearsloth (changes requested):

Robust.Shared/GameObjects/EntityManager.Components.cs
```diff
@@ -1028,6 +1028,54 @@ public bool TryGetComponent([NotNullWhen(true)] EntityUid? uid, ushort netId,
...
 
+        /// <inheritdoc/>
+        public bool CopyComponent<T>(EntityUid source, EntityUid target, [NotNullWhen(true)] out T? component, MetaDataComponent? metadataTarget = null) where T : IComponent
+        {
+            component = default;
+
+            if (!MetaQuery.Resolve(target, ref metadataTarget, false))
+                throw new ArgumentException($"Entity {target} is not valid.", nameof(target));
+
+            if (!HasComponent<T>(source))
```
poklj: utilizing something like `CopyComps(source, ent, null,  typeof(VocalComponent), typeof(SpeechComponent));` causes a `Unknown component: IComponent` error

image

Robust.Shared/GameObjects/IEntityManager.Components.cs
```diff
@@ -341,6 +341,38 @@ public partial interface IEntityManager
...
+        /// <summary>
+        /// Copy a single component from source to target entity and return a reference to the copied component.
+        /// </summary>
+        /// <param name="source">The source entity to copy the component from</param>
+        /// <param name="target">The target entity to copy the component to</param>
+        /// <param name="type">The type of component to copy</param>
+        /// <param name="component">The copied component if successful</param>
+        /// <param name="meta">Optional metadata of the target entity</param>
+        /// <returns>Whether the component was successfully copied</returns>
+        bool CopyComponent(EntityUid source, EntityUid target, IComponent type, [NotNullWhen(true)] out IComponent? component, MetaDataComponent? meta = null);
```
Fildrance: shouldn't it all be TryCopyComponent, as it follow overall TryXXX Microsoft lovely pattern

Robust.Shared/GameObjects/IEntityManager.Components.cs
```diff
@@ -341,6 +341,38 @@ public partial interface IEntityManager
         /// <returns>If the component existed in the entity.</returns>
         bool TryGetComponent([NotNullWhen(true)] EntityUid? uid, ushort netId, [NotNullWhen(true)] out IComponent? component, MetaDataComponent? meta = null);
 
+        /// <summary>
+        /// Copy a single component from source to target entity and return a reference to the copied component.
```
Fildrance: 'Tries to copy', and 'if source entity have it' should be here, probably

Robust.Shared/GameObjects/IEntityManager.Components.cs
MilonPL: yes it should, i will fix it one day i swear

Robust.Shared/GameObjects/IEntityManager.Components.cs
```diff
@@ -341,6 +341,38 @@ public partial interface IEntityManager
...
+        /// </summary>
+        /// <param name="source">The source entity to copy the component from</param>
```
Fildrance: Other xml-docs in file have dot at the end so this prolly all should too :3
And i'm not sure 'The' is nessesary here?... sorry if i'm wrong <_<

Fildrance: Well, thats fair for all of the comments here

Robust.Shared/GameObjects/IEntityManager.Components.cs
Fildrance: Method description also leaves to interpretation the question of 'what will it do if target entity already have that component'. The closest thing we can imagine is - it will return false and null, but it is also possible it will
1 - return true and reference to existing one, without overwritng (it exists already, there it is, take it)
2 - return false and reference (no i didnt copied it, take the existing one you jerk)
3 - return true and null (its already there! but not copied. wierd one, yeah).

this can be avoided by making more overloads and mimics for TryComp and Resolve methods but i wouldn't advice that until someone else asks. Clear description will suffice i think :3 (and no, u better have at least some more description in summary, as people will read summary first, and could read description on property only by doing extra work)

Robust.Shared/GameObjects/EntityManager.Components.cs
```diff
@@ -1028,6 +1028,58 @@ public bool TryGetComponent([NotNullWhen(true)] EntityUid? uid, ushort netId,
             return TryGetComponent(uid.Value, netId, out component, meta);
         }
 
+        /// <inheritdoc/>
+        public bool CopyComponent(EntityUid source, EntityUid target, IComponent sourceComponent, [NotNullWhen(true)] out IComponent? component, MetaDataComponent? meta = null)
+        {
+            // Use the concrete type of the sourceComponent to call the generic version
+            var type = sourceComponent.GetType();
+            var method = GetType().GetMethod(nameof(CopyComponent), 1, [typeof(EntityUid), typeof(EntityUid), type.MakeByRefType(), typeof(MetaDataComponent)
```
Fildrance: better use GetType result to duplicate generic version of code. Reflection of this type could be used if method that was generated dynamically would be cached for each requested type, but sadly it still will make code poorly readable. maybe there is an even better idea how to do it, but using dynamic method creation here is a bad fit :(

Robust.Shared/GameObjects/EntityManager.Components.cs
```diff
@@ -1028,6 +1028,58 @@ public bool TryGetComponent([NotNullWhen(true)] EntityUid? uid, ushort netId,
...
+        public bool CopyComponents(EntityUid source, EntityUid target, MetaDataComponent? meta = null, params IComponent[] types)
+        {
+            if (!MetaQuery.Resolve(target, ref meta))
+                return false;
+
+            var allSuccessful = true;
+
+            foreach (var type in types)
+            {
+                if (!CopyComponent(source, target, type, out _, meta))
```
Fildrance: each call will lead to dirty internally, i wonder if such bulk operations could be performed in a more optimal manner if dirty-related code would be called once?..

Fildrance (commented):
neat, but where are the tests, mate :(

Robust.Shared/GameObjects/EntityManager.Components.cs
metalgearsloth: The overhead of this will blow using generics out of the water, I'll try and re-write something if I get a chance.

Robust.Shared/GameObjects/EntityManager.Components.cs
metalgearsloth: If metadatacomp is passed through it is just a field check for lastmodified which is not super bad for now.

PR #5656, opened 2025-02-04
Add ICloneable support to ComponentNetworkGenerator

Tayrtahn:
ComponentNetworkGenerator now clones component fields that implement ICloneable.

As suggested here: https://github.com/space-wizards/space-station-14/pull/34838#discussion_r1938672739

### DIFF
--- a/Robust.Shared.CompNetworkGenerator/ComponentNetworkGenerator.cs
+++ b/Robust.Shared.CompNetworkGenerator/ComponentNetworkGenerator.cs
@@ -723,7 +723,17 @@ public partial class {componentName}{deltaInterface}
 
         private static bool IsCloneType(ITypeSymbol type)
         {
-            if (type is not INamedTypeSymbol named || !named.IsGenericType)
+            if (type is not INamedTypeSymbol named)
+            {
+                return false;
+            }
+
+            if (ImplementsInterface(named, nameof(ICloneable)))
+            {
+                return true;
+            }
+
+            if (!named.IsGenericType)
             {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


Robust.Shared.CompNetworkGenerator/ComponentNetworkGenerator.cs
```diff
@@ -735,5 +745,18 @@ private static bool IsCloneType(ITypeSymbol type)
...
+            foreach (var interfaceType in type.AllInterfaces)
+            {
+                if (interfaceType.ToDisplayString().Contains(interfaceName))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
```
metalgearsloth, line 760: typeof(XYZ).IsAssignableFrom should cover this right?

Tayrtahn: Dunno, but I can find out.

Tayrtahn: Sooo... looks like a no. Getting a `System.Type` from a `ITypeSymbol` is complicated and pretty janky. See https://stackoverflow.com/questions/34747260/getting-type-from-a-symbol-in-roslyn.

I assume that's why it's done this way elsewhere in Robust. The function is just copied from here anyway:
https://github.com/space-wizards/RobustToolbox/blob/ae7725aafe684a4866601d3d96082951f21c06a9/Robust.Serialization.Generator/Types.cs#L237-L248

The code duplication is a bit yucky, but it doesn't seem worth trying to break out an internal method of an internal class in a separate assembly to avoid it.

PR #5659, opened 2025-02-05
Cap two lighting CVars to prevent cheating 

notquitehadouken:
Resolution Scale: Capped at 1 because setting it above 1 does nothing except lag
Blur Factor: Capped at 0.003f because above that lights go way further than they should (#4942 part 3)


notquitehadouken:
Blur off

image

Factor of 0.003

image

Factor of 0.01

image

PJB3005 (changes requested):
You should not be modifying CVars in value changed handlers.

PR #5660, opened 2025-02-06
Remove unused obsolete TryGetContainingContainer override

TemporalOroboros:
There are a bunch of existing calls to this method that are compatible with the `Entity<T>` version which still throw `Obsolete` warnings because they use an `EntityUid` and apparently the conversion to an `Entity<T>` is lower priority.
As a result this fixes ~a dozen obsolete warnings.

Requires https://github.com/space-wizards/space-station-14/pull/34912

## Breaking Changes
Any downstreams still using the obsolete version and passing the components separately will need to pack them with the uid now.
Otherwise, if they were relying on the implicit conversion from a different `Entity<T>` or similar they'll need to use `.Owner`
If they were just passing an `EntityUid` they should be fine.


Tayrtahn (approved):
Please kill it, I hate it so.

slarticodefast (approved):
We had three separate maintainer PRs trying to kill it so far.

PR #5661, opened 2025-02-06
A compilation of simple one-line fixes

TemporalOroboros:
Mass-purge of obsolete method calls.


Robust.Server/Physics/GridFixtureSystem.cs
```diff
@@ -268,8 +268,8 @@ private void CheckSplits(EntityUid uid, HashSet<ChunkSplitNode> dirtyNodes)
                     newGrids[i] = newGridUid;
 
                     // Keep same origin / velocity etc; this makes updating a lot faster and easier.
-                    _xformSystem.SetWorldPosition(newGridXform, gridPos);
-                    _xformSystem.SetWorldPositionRotation(newGridUid, gridPos, gridRot, newGridXform);
+                    _xformSystem.SetWorldPosition((newGridUid, newGridXform), gridPos);
+                    _xformSystem.SetWorldPositionRotation(newGridUid, gridPos, gridRot, newGridXform); // Why do we set this twice???
```
TemporalOroboros: Is there a reason for this or should I kill the first one?

VerinSenpai: looks like SetWorldPosition calls SetWorldPositionRotationInternal with grid rotation as null. I don't know  enough about C# or the codebase to say for certain but this seems redundant.

TemporalOroboros: Alright, I'm yeeting it. If grid splitting explodes yell at me and I'll put it back.

VerinSenpai: Yeetus Deletus!

PJB3005:
CI fail, can that be avoided?

TemporalOroboros:
> CI fail, can that be avoided?

Yes

Robust.Client/GameObjects/Components/Renderable/SpriteBoundsOverlay.cs
```diff
@@ -16,13 +16,15 @@ public sealed class ShowSpriteBBCommand : LocalizedCommands
 
         public override void Execute(IConsoleShell shell, string argStr, string[] args)
         {
-            EntitySystem.Get<SpriteBoundsSystem>().Enabled ^= true;
+            IoCManager.Resolve<IEntityManager>()
+                .System<SpriteBoundsSystem>()
+                .Enabled ^= true;
```
ElectroJr: Eventually static `IoCManager.Resolve()` might also just get obsoleted, so this change is just kind of kicking the can down the road.

Ideally, it'd just inherit from `LocalizedEntityCommands` instead of `LocalizedCommands` and use a `[Depenency] SpriteBoundsSystem` field.

same goes for `LightBBCommand.cs`

TemporalOroboros: Done.

PR #5672, opened 2025-02-12
Fix MultiRootInheritanceGraph not detecting circular inheritance

beck-thompson:
`MultiRootInheritanceGraph` failed to properly detect the case of you being your own parent! This caused an infinite loop on start up of SS14 if you had a prototype whose parent was itself.

Example:
```
- type: entity
  parent: CoolEntity
  id: CoolEntity
```

I think the fix is just to add yourself to the data structure before checking for circular inheritance but I could be wrong! You could also just manually check that you don't have yourself as a parent but that seemed hacky. This makes sense and does appear to work.

### DIFF
--- a/Robust.Shared/Prototypes/MultiRootInheritanceGraph.cs
+++ b/Robust.Shared/Prototypes/MultiRootInheritanceGraph.cs
@@ -35,21 +35,12 @@ public sealed class MultiRootInheritanceGraph<T> where T : notnull
 
     public void Add(T id, params T[] parents)
     {
-        _rootNodes.Remove(id);
-
-        foreach (var parent in parents)
-        {
-            var edges = _edges.GetOrNew(parent);
-            edges.Add(id);
-            _parents[id] = parents;
-
-            if (!_parents.ContainsKey(parent))
-                _rootNodes.Add(parent);
-        }
-
         //check for circular inheritance
         foreach (var parent in parents)
         {
+            if (EqualityComparer<T>.Default.Equals(parent,id))
+                throw new InvalidOperationException($"Circular Inheritance detected for id \"{id}\" and parent \"{parent}\"");
+
             var parentsL1 = GetParents(parent);
             if(parentsL1 == null) continue;
 
@@ -68,6 +59,18 @@ public sealed class MultiRootInheritanceGraph<T> where T : notnull
                     }
                 }
             }
+        }
+
+        _rootNodes.Remove(id);
+
+        foreach (var parent in parents)
+        {
+            var edges = _edges.GetOrNew(parent);
+            edges.Add(id);
+            _parents[id] = parents;
+
+            if (!_parents.ContainsKey(parent))
+                _rootNodes.Add(parent);
         }
     }
 
--- a/Robust.Shared/Prototypes/MultiRootInheritanceGraph.cs
+++ b/Robust.Shared/Prototypes/MultiRootInheritanceGraph.cs
@@ -38,8 +38,8 @@ public sealed class MultiRootInheritanceGraph<T> where T : notnull
         //check for circular inheritance
         foreach (var parent in parents)
         {
-            if (EqualityComparer<T>.Default.Equals(parent,id))
-                throw new InvalidOperationException($"Circular Inheritance detected for id \"{id}\" and parent \"{parent}\"");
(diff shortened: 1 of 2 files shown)


beck-thompson:
I uhh think the tests might be borked... The run fine normally on my computer 😢

deltanedas:
6 hour test real

beck-thompson:
Figured out the issue with the test. The issue is that the `TestCircleException` runs and properly detects and throws exceptions. However, because the `PrototypeManager` is reused over mutliple tests, the graph is all messed up (There are now like partially inserted things that have circular inheritance).

I'm unsure how I should fix this! Here is what I think I could do:
1.) Rewrite the `PrototypeManager_Test` to not reuse the manager throughout multiple tests.
2.) Change my logic so nothing gets inserted into the graph. I can only think of kind of hard coded checks beforehand so it wont be as pretty!
3.) I had a third I idea... I forget now :laughing:

For the person reviewing this, or anyone who has an idea :laughing:, let me know what I should do :saluting_face:

Robust.Shared/Prototypes/MultiRootInheritanceGraph.cs
```diff
@@ -35,6 +35,18 @@ public bool TryGetParents(T id, [NotNullWhen(true)] out T[]? parents)
...
+
+        foreach (var parent in parents)
+        {
+            var edges = _edges.GetOrNew(parent);
+            edges.Add(id);
+            _parents[id] = parents;
+
+            if (!_parents.ContainsKey(parent))
+                _rootNodes.Add(parent);
+        }
```
Tayrtahn: Would it work to just do a comparison of `parent` and `id` at the top of this loop and throw an exception if they're equal?

beck-thompson: Yes! You'd have to loop through all the parents but it would work. The only reason I didn't like it was because it didn't seem as pretty but if you think its a good idea I can change it!

beck-thompson: I wasn't thinking and your right can just check in the for loop! I agree I think that's a better solution :partying_face:

Robust.Shared/Prototypes/MultiRootInheritanceGraph.cs
```diff
@@ -38,6 +38,9 @@ public void Add(T id, params T[] parents)
         //check for circular inheritance
         foreach (var parent in parents)
         {
+            if (EqualityComparer<T>.Default.Equals(parent,id))
+                throw new InvalidOperationException($"Circular Inheritance detected for id \"{id}\" and parent \"{parent}\"");
```
Tayrtahn:
```suggestion
                throw new InvalidOperationException($"Self Inheritance detected for id \"{id}\"");
```
Might as well make the message distinct

Robust.Shared/Prototypes/MultiRootInheritanceGraph.cs
```diff
@@ -38,6 +38,9 @@ public void Add(T id, params T[] parents)
         //check for circular inheritance
         foreach (var parent in parents)
         {
+            if (EqualityComparer<T>.Default.Equals(parent,id))
```
Tayrtahn:
```suggestion
            if (EqualityComparer<T>.Default.Equals(parent, id))
```
Formatting

Robust.UnitTesting/Shared/Prototypes/MultiRootGraphTest.cs
```diff
@@ -104,4 +104,11 @@ public void CircleCheckTest()
         Assert.Throws<InvalidOperationException>(() => graph.Add(Id3, new[] { Id1 }));
 
     }
+
+    [Test]
+    public void IsOwnParentTest()
+    {
+        var graph = new MultiRootInheritanceGraph<string>();
+        Assert.Throws<InvalidOperationException>(() => graph.Add(Id1, new []{Id1}));
```
Tayrtahn:
```suggestion
        Assert.Throws<InvalidOperationException>(() => graph.Add(Id1, new[] { Id1 }));
```
Formatting

Tayrtahn (approved):
I'm not an engine maint, but this looks good to me! Just a few suggested tweaks.

PR #5674, opened 2025-02-12
Move `ParseObject` method into a public class for content to use

Simyon264:
Uhhh, yeah. Useful in Content.


SlamBamActionman (approved):
Makes sense as a public static to me.

PR #5675, opened 2025-02-12
remove obsolete EntityUid methods in favor of Entity<T> pattern

MilonPL:
entityuid casts to entity<T> so in 99% of cases it was being marked as obsolete despite being fine

Entity<T> truly is the best thing ever since sliced bread


PJB3005 (changes requested):
Split the PR into warning cleanup and obsolete method removal. We can't remove these methods for a good while yet.

MilonPL:
> We can't remove these methods for a good while yet.

@PJB3005 I think the container ones would be fine to remove? Otherwise the new ones end up getting incorrectly marked as obsolete in new PRs because of `EntityUid` casting to `Entity<T?>`.

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -372,9 +372,11 @@ public bool Anchored
                 {
                     _anchored = value;
                 }
-                else if (value && !_anchored && _mapManager.TryFindGridAt(MapPosition, out _, out var grid))
+                else if (value && !_anchored && _mapManager.TryFindGridAt(MapPosition, out var gridUid, out var grid))
                 {
-                    _anchored = _entMan.EntitySysManager.GetEntitySystem<SharedTransformSystem>().AnchorEntity(Owner, this, grid);
+                    _anchored = _entMan.EntitySysManager.GetEntitySystem<SharedTransformSystem>().AnchorEntity(
+                        (Owner, this),
+                        (gridUid, grid));
```
ElectroJr, line 379: The new AnchorEntity method seems to have a debug assert that is not present in the other methods that seems to be failing some tests for some reason

ElectroJr:
I know that at least the container method has been marked as obsolete for a decent time now (about a year, #5136), so I think its fine being removed by now. Though all of these breaking changes are pretty similar and I'd rather just do them all at once anyways.  But I also know I'm also just generally more trigger happy with making breaking changes. So I guess its up to @PJB3005

MilonPL:
I have no idea what's causing the assert to fail, it's just not present in the non-entity variant, and even if we pass in the grid entity as null it fails to anchor it and fails at a later assert. Might want to make sure the behavior is the same before removing those, or when we finally remove transformcomponent methods. 
Needs another content PR because one of these has sneaked in, and also evil (ent, null) removed.

PR #5679, opened 2025-02-14
feat: shaders now can accept array of Color as parameter

Fildrance:
Added overload ShaderInstance.SetParameterImpl that accepts array of colors and translates it to vec4[] in shader. Corrected respective code, tested that it works (on one sample).
```
void SetParameterImpl(string name, Color[] value);
```

Was required by visual component that have list of buttons, each one of which could have own background color (radial menu).

### DIFF
--- a/Robust.Client/Graphics/Clyde/GLObjects/Clyde.ShaderProgram.cs
+++ b/Robust.Client/Graphics/Clyde/GLObjects/Clyde.ShaderProgram.cs
@@ -334,7 +334,7 @@ namespace Robust.Client.Graphics.Clyde
             }
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            private void SetUniformDirect(int slot, in Color color, bool convertToLinear=true)
+            private void SetUniformDirect(int slot, in Color color, bool convertToLinear = true)
             {
                 var converted = color;
                 if (convertToLinear)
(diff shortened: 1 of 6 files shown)


Robust.Client/Graphics/Clyde/GLObjects/Clyde.ShaderProgram.cs
```diff
@@ -349,6 +349,33 @@ private void SetUniformDirect(int slot, in Color color, bool convertToLinear=tru
                 }
             }
 
+            public void SetUniform(string uniformName, in Color[] colors)
```
PJB3005: This should not be passed via `in`.
```suggestion
            public void SetUniform(string uniformName, Color[] colors)
```

Robust.Client/Graphics/Clyde/GLObjects/Clyde.ShaderProgram.cs
```diff
@@ -349,6 +349,33 @@ private void SetUniformDirect(int slot, in Color color, bool convertToLinear=tru
...
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            private void SetUniformDirect(int slot, Color[] colors, bool convertToLinear = true)
+            {
+                unsafe
+                {
+                    if (convertToLinear)
+                    {
+                        for (int i = 0; i < colors.Length; i++)
+                        {
+                            colors[i] = Color.FromSrgb(colors[i]);
```
PJB3005: You're mutating this array.

Robust.Client/Graphics/Shaders/ParsedShader.cs
```diff
@@ -225,7 +225,8 @@ public static bool TypeSupportsArrays(this ShaderDataType type)
             return
                 (type == ShaderDataType.Float) ||
                 (type == ShaderDataType.Vec2) ||
-                (type == ShaderDataType.Bool);
+                (type == ShaderDataType.Bool) ||
+                (type == ShaderDataType.Vec4);
```
PJB3005, line 229: TODO comment above this is now outdated.

PJB3005 (changes requested):

Robust.Client/Graphics/Shaders/ParsedShader.cs
Fildrance: corrected to mention only int and vec3

Robust.Client/Graphics/Clyde/GLObjects/Clyde.ShaderProgram.cs
Fildrance: Replaced mutation with filling stack-allocated array

Robust.Client/Graphics/Clyde/GLObjects/Clyde.ShaderProgram.cs
Fildrance: :sadge: copy-paste. fixed.

PJB3005 (approved):

PR #5681, opened 2025-02-16
Expand and comment SharedTransformSystem anchoring methods

TemporalOroboros:
Biggest change here is making the default anchor setter use the entities current grid instead of calculating it from scratch.
Doesn't seem to break anything and Sloth says it shouldn't.

Otherwise, this is just expanding and adding doc comments to the TransformSystem anchoring API.

Required for https://github.com/space-wizards/space-station-14/pull/34938

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
@@ -135,12 +135,14 @@ public abstract partial class SharedTransformSystem
     }
 
     /// <inheritdoc cref="Unanchor(Entity{TransformComponent}, bool)"/>
+    [Obsolete("Use Entity<T> varients")]
     public void Unanchor(EntityUid uid)
     {
         Unanchor(uid, XformQuery.GetComponent(uid));
     }
 
     /// <inheritdoc cref="Unanchor(Entity{TransformComponent}, bool)"/>
+    [Obsolete("Use Entity<T> varients")]
     public void Unanchor(EntityUid uid, TransformComponent xform, bool setPhysics = true)
     {
         Unanchor((uid, xform), setPhysics);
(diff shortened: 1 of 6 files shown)


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -165,6 +221,67 @@ public void Unanchor(EntityUid uid, TransformComponent xform, bool setPhysics =
...
+            return true;
+
+        // This will be set again when the transform initializes, actually anchoring it.
+        if (!comp.Initialized)
+        {
+            comp._anchored = value;
+            return true;
+        }
+
+        if (value && _mapManager.TryFindGridAt(GetMapCoordinates(entity), out var gridUid, out var gridComp))
```
TemporalOroboros, line 266: @metalgearsloth So, it turns out a few tests rely on this implicitly since creating a grid/tile under an entity does not move the entity onto the grid immediately. At least one of them (https://github.com/space-wizards/RobustToolbox/blob/5c7ab43049046b0edd77f390710934a78b09df7e/Robust.UnitTesting/Shared/GameObjects/Systems/AnchoredSystemTests.cs#L183-L204) explicitly disables grid traversal before attempting to anchor the test entity. Would you rather I change the tests to explicitly anchor the entities to the test grid, or hook into tile changes to collect any entities the grid expands underneath?

metalgearsloth: Tile changes should collect entities in general but should not be anchoring them, so it sounds like the test needs to anchor them.

metalgearsloth:
I ended up rolling some of these into a new PR but the biggest one I added is preventing callers from footgunning themselves with AnchorEntity doing 0 validation if the entity is already anchored. Will copy from other repo.

PR #5683, opened 2025-02-17
Add SingletonEntitySystem

Tayrtahn:
An extension of EntitySystem intended for systems that need to hold some data that isn't logically tied to an entity in the world.

This avoids the ECS-violating temptation to store data on the system itself, instead spawning a dummy entity in nullspace and attaching a data-holder component to it, which means that data should properly persist through a full game save/load. A reference to the entity is cached in the system to avoid the expense of querying for it each time, and component lifestage handlers are set up to ensure that the reference remains valid.

This is the same logic as used in https://github.com/space-wizards/space-station-14/pull/30686. The purpose of this class is to make future cases of this pattern simpler to implement (the content PR would just use `public sealed class RoundEndStatisticsSystem : SingletonEntitySystem<RoundEndStatisticsComponent>`)


deltanedas:
should have a round entity spawned from a prototype with wire layouts statistics etc added to it imo
since this is used for per-round data it should be on a single round entity

Tayrtahn:
Hm, this doesn't work for networked components. I have a plan for fixing that.

Tayrtahn:
Alright, my plan for making this work with networked component just seems to cause nasty client crashes - *other* entities are syncing incorrectly somehow. So I'm just moving this to be server-only.

Tayrtahn:
> should have a round entity spawned from a prototype with wire layouts statistics etc added to it imo
> since this is used for per-round data it should be on a single round entity

It would be pretty easy to change this around so that all the components are on a single entity instead of spawning a new entity for each one. I suppose that would be slightly less overhead.

Tayrtahn:
Actually, my networked version was working fine, I was just testing it the wrong way.

Moved back to Shared and changed the API to account for sync delays on the client. It's a little less convenient to work with than before (you need to use `TryGetInstance` instead of just being able to use `Instance`) but it's still pretty nice.

Tayrtahn:
Added an integration test which sets up an instance on the server, checks that networked singleton components are properly synchronized to the client, and verifies that saving the entity data to a file, deleting everything, and reloading the saved data properly restores the singleton entity and component.

FairlySadPanda:
Hi, apologies for running a sanity check on this:

This system stores one and only one component on an entity. But you leak that entity whenever you try and get the component. The entity does not appear to be useful except as a device to store a component, and considering all you are doing here is creating an in-memory cache of data by sticking it on an Entity:

1. Why use more than one entity at all? One mega-entity in nullspace can store every single unique global component on itself.
2. Why leak that the entity exists at all? Why not just say "this system stores data typed to a component and you can get it when you like".
3. Why return the entity when trying to get the data from this system, and not just the component itself?
4. Can you ensure this system cannot be badly broken by poor code outside of the system? What if I use CopyComp to duplicate the singleton component? Can you ensure that one and only one copy of the component exists at any one time? If not, is this really enforcing a singleton pattern?
5. Isn't this just storing data in the system with extra religion attached? Why should an engineer use this over creating a Manager to solve their problems?

My hunch is that this is a component storage system; the fact that the specific strategy used is a global, single copy of a component is maybe not useful to leak to systems that inherit from this one, or users of systems that do? In the past I've referred to classes that store, or provide a layer of access over, data as a repository. As such this feels like a ComponentRepositorySystem, or some similar parlance.

Tayrtahn:
Those are all excellent points, thanks for the feedback! 

- I agree that it would make more sense to use a single entity that holds all the components that use this. The structure of this was based on scavenging through the chem refactor, which uses a similar setup to track reagent entities. That version has the main entity serve as a container for the reagent entities and uses container event subscriptions to do stuff with them. I wanted to preserve the possibility of using this system like that, but I didn't really think it's worth it. Consider that a planned change at this point.
- Exposing just the component instead of the entity also sounds like a good change. I think I had some concerns about future problems if we ever get struct components, but we can probably change things to pass by ref assuming that ever happens.
- The Singleton pattern is indeed not well enforced. There's a debug assert that will trip on component startup if the system already has a cached reference, but there's nothing really stopping that from happening - an admin could add a second insurance via VV and things would get weird. The content demo I tried with this already showed an example of it failing: there's an integration test that creates one of every component, and it fails based on execution order - an instance is created indirectly when another system initializes, then the test manually adds another instance which trips the assert. I'm not really sure how to improve things in this department.
- The main goal of this system is to make it easier to store data that isn't tied to a specific entity in a way that will persist through save/load actions. I'm starting to think that the concept might be a bit off though - it might make more sense to always have data tied to an entity representing *something*, even if it's abstract, like a grid, map, or game round.

Thanks for the feedback. I'm gonna think about this one some more.

K-Dynamic:
How did it go in the end

Tayrtahn:
I think it's an idea worth revisiting at some point, but not by me right now.

PR #5685, opened 2025-02-19
Make the ScrollContainer scrollbars public API

sowelipililimute:
https://github.com/space-wizards/space-station-14/blob/master/Content.Client/Power/PowerMonitoringWindow.xaml.Widgets.cs#L272

https://github.com/space-wizards/space-station-14/blob/master/Content.Client/Medical/CrewMonitoring/CrewMonitoringWindow.xaml.cs#L390

https://github.com/space-wizards/space-station-14/blob/master/Content.Client/Atmos/Consoles/AtmosMonitoringConsoleWindow.xaml.cs#L365

https://github.com/space-wizards/space-station-14/blob/master/Content.Client/Atmos/Consoles/AtmosAlertsComputerWindow.xaml.cs#L527

I don't know why this wasn't done the second or third or fourth time this exact fragment to try and search for the scrollbar within the scrollcontainer's children was copypasted

### DIFF
--- a/Robust.Client/UserInterface/Controls/ScrollContainer.cs
+++ b/Robust.Client/UserInterface/Controls/ScrollContainer.cs
@@ -15,8 +15,8 @@ namespace Robust.Client.UserInterface.Controls
         private bool _vScrollVisible;
         private bool _hScrollVisible;
 
-        private readonly VScrollBar _vScrollBar;
-        private readonly HScrollBar _hScrollBar;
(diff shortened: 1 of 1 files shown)


PJB3005:
Yeah no, the use case for this is being able to scroll to a certain UI element/position. That should be functionality on the control instead of exposing the scrollbar like that.

All that code copy pasting should never have passed review and I'm incredibly upset at content maintainers again.

PR #5687, opened 2025-02-20
Revert "Add ICloneable support to ComponentNetworkGenerator"

ElectroJr:
Reverts space-wizards/RobustToolbox#5656

Generates invalid code when using something like
```cs
[DataField, AutoNetworkedField]
public TestClass TestField = new();

public sealed class TestClass : ICloneable
{
    public object Clone() => new();
}
```

See https://discord.com/channels/310555209753690112/900426319433728030/1341968643646423210


deltanedas:
was good for rooting out uses of String at least

PR #5689, opened 2025-02-20
Expose ScrollContainer's ScrollValue/-Target

Toby222:
Facilitates implementing nicer autoscroll for certain windows and widgets.
Alternative solution to space-wizards/RobustToolbox/pull/5685


PJB3005:
I prefer #5697 so I merged that one instead.

Toby222:
Perfect!

PR #5692, opened 2025-02-20
IRobustCloneable and generator support

Tayrtahn:
A redo of #5656.

This time ditching `System.ICloneable` in favor of a custom `IRobustCloneable<T>` class. `Clone()` method returns `T` instead of `object`, which is nice. Added special handling to `ComponentNetworkGenerator` for implementers.

Includes a test which defines a class and a struct implementing `IRobustCloneable` and uses them as DataFields on a component. The component is networked and fields values are set on the server and verified on the client. This doesn't entirely make sure that the field has actually been cloned, but it does verify that the source code generated for the component is functional.


walksanatora:
wait when this is merged... can we make `Component` generate IRobustClonable and finally de-jankify cloning.

PJB3005 (approved):

PR #5696, opened 2025-02-22
Fix issue regarding Tilemaps not saving when modified

southbridge-fur:
When new tiles are added to a map, the serializer wasn't able to properly allocate new IDs for the tilemap because the list of known tile IDs wasn't getting populated. This PR makes that list get populated now.

### DIFF
--- a/Robust.Shared/EntitySerialization/EntitySerializer.cs
+++ b/Robust.Shared/EntitySerialization/EntitySerializer.cs
@@ -593,7 +593,7 @@ public sealed class EntitySerializer : ISerializationContext,
     public MappingDataNode Write()
     {
         DebugTools.AssertEqual(Maps.ToHashSet().Count, Maps.Count, "Duplicate maps?");
-        DebugTools.AssertEqual(Grids.ToHashSet().Count, Grids.Count, "Duplicate frids?");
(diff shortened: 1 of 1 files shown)


Robust.Shared/EntitySerialization/EntitySerializer.cs
```diff
@@ -593,7 +596,7 @@ private void SerializeEntityInternal(EntityUid uid)
     public MappingDataNode Write()
     {
         DebugTools.AssertEqual(Maps.ToHashSet().Count, Maps.Count, "Duplicate maps?");
-        DebugTools.AssertEqual(Grids.ToHashSet().Count, Grids.Count, "Duplicate frids?");
+        DebugTools.AssertEqual(Grids.ToHashSet().Count, Grids.Count, "Duplicate grids?");
```
southbridge-fur, line 599: Also I fixed this typo

ArtisticRoomba:
This issue is **critical**.

When the bug occurs, map saves can *fail silently* and state that things are saved successfully when in reality *they are not*, causing a mapper to lose all of their work. This happens from the moment the bug happens and autosaves won't save you.

A mapper can continue to map for hours after triggering the bug without knowing that all of their work will be wiped when the local server shuts down. Unfortunately we've had one person lose 90% of a new map they were working on because they trusted saves/autosaves.

PR #5703, opened 2025-02-24
Expand PreferOtherTypeAnalyzer/Fixer to check all uses

Tayrtahn:
Rejiggers `PreferOtherTypeAnalyzer` and `PreferOtherTypeFixer` to check all instances of the type being used, not just variable declarations. This includes uses as type arguments in other generics, method return types, property get accessors, nullables, and arrays.

In other words, this will now flag and fix stuff like:
`List<ProtoId<EntityPrototype>>` -> `List<EntProtoId>`
`Dictionary<int, ProtoId<ReagentPrototype>>` -> `Dictionary<int, EntProtoId>`
`List<HashSet<Queue<ProtoId<EntityPrototype>>>>` -> `List<HashSet<Queue<EntProtoId>>>`
`ProtoId<EntityPrototype>[]` -> `EntProtoId[]`
`ProtoId<EntityPrototype> MyMethod() => _someValue` -> `EntProtoId MyMethod() => _someValue`
`ProtoId<EntityPrototype> MyProperty => _someValue` -> `EntProtoId MyProperty => _someValue`
`ProtoId<EntityPrototype>?` -> `EntProtoId?`

This also massively simplifies the fixer logic by using the diagnostic's Properties to pass the replacement type name from the analyzer, avoiding the need to search for the attribute again in the fixer.

This required replacing some instances of `ProtoId<EntityPrototype>` that had snuck into the project, including a few that were causing debug asserts in content's YAML linter.

The tests were updated to cover these cases as well.


Robust.Shared/Prototypes/EntProtoId.cs
```diff
@@ -17,8 +17,7 @@ namespace Robust.Shared.Prototypes;
 /// </remarks>
 /// <remarks><seealso cref="ProtoId{T}"/> for a wrapper of other prototype kinds.</remarks>
 [Serializable, NetSerializable]
-public readonly record struct EntProtoId(string Id) : IEquatable<string>, IComparable<EntProtoId>, IAsType<string>,
-    IAsType<ProtoId<EntityPrototype>>
+public readonly record struct EntProtoId(string Id) : IEquatable<string>, IComparable<EntProtoId>, IAsType<string>
```
Tayrtahn, line 20: Is it okay to outright delete this, or should I suppress the warning for these cases? Everything seems to work fine without it, but I'm hesitant to trust that I'm right on this.

Robust.Shared/Prototypes/EntProtoId.cs
```diff
@@ -17,8 +17,7 @@ namespace Robust.Shared.Prototypes;
 /// </remarks>
 /// <remarks><seealso cref="ProtoId{T}"/> for a wrapper of other prototype kinds.</remarks>
 [Serializable, NetSerializable]
-public readonly record struct EntProtoId(string Id) : IEquatable<string>, IComparable<EntProtoId>, IAsType<string>,
-    IAsType<ProtoId<EntityPrototype>>
```
Tayrtahn, line 21: Is it okay to just outright remove this or should I suppress the warning for this case?

Everything seems to be fine with it removed, but I'm hesitant.

PR #5710, opened 2025-02-26
Modify TryCopyComponents metadata TryGetComponent

poklj:
fixes `TryCopyComponents` where it attempts to pull the metadata from the source rather than the target required further down inside of CopyComponent.

Without this change, both providing a `Metadata(target)` or even `null` will cause a `Entity is not the owner of the Component. Component: MetaDataComponent` error


metalgearsloth (changes requested):
Needs test coverage like the others either you can do it or I can maybe do it in a couple days

poklj:
> Needs test coverage like the others either you can do it or I can maybe do it in a couple days

Should be as simple as using a modified variant of the previous CopyMultiple test but using the TryCopyComponents instead of `CopyComponents`

PR #5712, opened 2025-02-27
CVar serialization fix

### LINKED ISSUE #5711: net.max_connections is saved on the client
It shouldn't be in client_config.toml, client doesn't care.

Ideally if a cvar is Server + archive only the server archives it, not the client.

metalgearsloth:
Fixes https://github.com/space-wizards/RobustToolbox/issues/5711

On my local client it removes net.max_connections, game.hostname, and the admin cvars.

### DIFF
--- a/Robust.Shared/Configuration/ConfigurationManager.cs
+++ b/Robust.Shared/Configuration/ConfigurationManager.cs
@@ -301,12 +301,31 @@ namespace Robust.Shared.Configuration
 
             try
             {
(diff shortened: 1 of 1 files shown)


metalgearsloth:
Might stew on this a bit more as we have a lot of server only vars that are being cross-referenced, hmm...

metalgearsloth:
Actually I think this is fine, most of the ones I checked did need to be replicated but the client shouldn't care about saving them (e.g. hostname).

PR #5714, opened 2025-02-28
Fix deceptive "successfully saved" messages for mappers

DeepwaterCreations:
**As things currently stand:** 
When a mapper tries to save their map, the serialization process might throw an exception, in which case nothing whatsoever is actually saved. Unfortunately, the client console will still print out "Map successfully saved!" causing the mapper to blithely go on putting hours of work into their mapping project without realizing that as soon as the server is shut down, all of their allegedly-saved work will vanish into the nether-voids.

**The fix:** 
A couple lines of code to actually check the return value of the TrySaveMap() function and print a different message if it's false. I did this for grid saving, too, since I noticed it made the same mistake. It's a pretty minor change, but if nothing else, at least mappers won't be flat-out stone-cold lied to about their saves working.

**Should this be a hotfix?** 
I'm not sure. Technically, nothing is *broken* as such. The serialization process does spit out its stack traces into the server log, so if the mapper is keeping an eye on that, they'll know what's happening. On the other hand, I have personally lost hours and hours of work on multiple occasions to this, because I was taking the "Map saved successfully" message at its word. Which, you know, I feel is pretty dang reasonable. 

I know other mappers have suffered this same fate, though the bug that was breaking serialization for the majority of folks (and me at the time) did get fixed. But as I later discovered, there's more than one way to hit this particular landmine.

If it were me, I'd say yes, hotfix; protecting people from potentially losing hours of work is better done ASAP, regardless of how it's done or how technical the actual change is or isn't. But I'm just this guy, you know?

**N.B.:**
My Portuguese translation is based entirely on Duck Duck Go's translation tool, so I can't really vouch that it's accurate. I figured it was better than nothing. I didn't add one for grids because none of the other grid stuff's strings were translated either, so probably that should be its own whole effort, not part of this PR.


Robust.Server/Console/Commands/MapCommands.cs
```diff
@@ -207,8 +214,15 @@ public override void Execute(IConsoleShell shell, string argStr, string[] args)
...
-            shell.WriteLine(Loc.GetString("cmd-savemap-success"));
+            bool saveSuccess = _system.GetEntitySystem<MapLoaderSystem>().TrySaveMap(mapId, new ResPath(args[1]));
+            if(saveSuccess)
+            {
+                    shell.WriteLine(Loc.GetString("cmd-savemap-success"));
+            }
+            else
+            {
+                    shell.WriteError(Loc.GetString("cmd-savemap-error"));
+            }
```
Tayrtahn, line 225:
```suggestion
            shell.WriteLine(Loc.GetString(saveSuccess ? "cmd-savemap-success" : "cmd-savemap-error"));
```

DeepwaterCreations: But one of them's a WriteError, not a WriteLine.

Tayrtahn: Whoops. Nevermind then!

metalgearsloth:
We have LocalizedEntityCommand now but it's whatever, more surprising people don't read the errors above it.

PR #5716, opened 2025-03-01
Validate that ProtoId values are not abstract

Tayrtahn:
Adds an extra validation step to `ProtoId<T>` and `EntProtoId` which makes sure that the prototype is not abstract. This lets YAML linter prevent attempts to spawn abstract prototypes (example: https://github.com/space-wizards/space-station-14/pull/35562)

If there's a valid reason to have a `ProtoId` field for an abstract prototype, then go ahead and close this. I can't think of one, and there don't seem to be any existing uses of it.

### DIFF
--- a/Robust.Shared/Serialization/TypeSerializers/Implementations/EntProtoIdSerializer.cs
+++ b/Robust.Shared/Serialization/TypeSerializers/Implementations/EntProtoIdSerializer.cs
@@ -25,7 +25,7 @@ public sealed class EntProtoIdSerializer : ITypeSerializer<EntProtoId, ValueData
         if (prototypes.TryGetKindFrom<EntityPrototype>(out _) && prototypes.HasMapping<EntityPrototype>(node.Value))
         {
             if (!prototypes.TryIndex<EntityPrototype>(node.Value, out _))
-                return new ErrorNode(node, $"{typeof(EntityPrototype)} {node.Value} is abstract!");
+                return new ErrorNode(node, $"{nameof(EntityPrototype)} {node.Value} is abstract!");
 
             return new ValidatedValueNode(node);
         }
--- a/Robust.Shared/Serialization/TypeSerializers/Implementations/EntProtoIdSerializer.cs
+++ b/Robust.Shared/Serialization/TypeSerializers/Implementations/EntProtoIdSerializer.cs
@@ -24,7 +24,7 @@ public sealed class EntProtoIdSerializer : ITypeSerializer<EntProtoId, ValueData
         var prototypes = dependencies.Resolve<IPrototypeManager>();
         if (prototypes.TryGetKindFrom<EntityPrototype>(out _) && prototypes.HasMapping<EntityPrototype>(node.Value))
         {
-            if (!prototypes.TryIndex<EntityPrototype>(node.Value, out _))
+            if (!prototypes.HasIndex<EntityPrototype>(node.Value))
                 return new ErrorNode(node, $"{nameof(EntityPrototype)} {node.Value} is abstract!");
 
             return new ValidatedValueNode(node);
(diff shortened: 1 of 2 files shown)


deltanedas:
does TryIndex even work for abstract, and if so why

Tayrtahn:
> does TryIndex even work for abstract, and if so why

Nope, it returns false for abstract prototypes (which is how this PR checks for them). The problem linked above snuck through because the entity spawn table stuff doesn't check for abstract prototypes (which it probably should). That would just be a runtime check though - this extra validation would let the linter catch the problem.

Robust.Shared/Serialization/TypeSerializers/Implementations/EntProtoIdSerializer.cs
```diff
@@ -23,7 +23,12 @@ public ValidationNode Validate(ISerializationManager serialization, ValueDataNod
     {
         var prototypes = dependencies.Resolve<IPrototypeManager>();
         if (prototypes.TryGetKindFrom<EntityPrototype>(out _) && prototypes.HasMapping<EntityPrototype>(node.Value))
```
PJB3005: Are these checks for `HasMapping` even necessary at this point?

Robust.Shared/Serialization/TypeSerializers/Implementations/EntProtoIdSerializer.cs
```diff
@@ -23,7 +23,12 @@ public ValidationNode Validate(ISerializationManager serialization, ValueDataNod
...
+        {
+            if (!prototypes.TryIndex<EntityPrototype>(node.Value, out _))
```
PJB3005: Use `HasIndex` instead

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/ProtoIdSerializer.cs
```diff
@@ -21,7 +21,12 @@ public ValidationNode Validate(ISerializationManager serialization, ValueDataNod
     {
         var prototypes = dependencies.Resolve<IPrototypeManager>();
         if (prototypes.TryGetKindFrom<T>(out _) && prototypes.HasMapping<T>(node.Value))
+        {
+            if (!prototypes.TryIndex<T>(node.Value, out _))
```
PJB3005: Ditto

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/ProtoIdSerializer.cs
```diff
@@ -21,7 +21,12 @@ public ValidationNode Validate(ISerializationManager serialization, ValueDataNod
     {
         var prototypes = dependencies.Resolve<IPrototypeManager>();
         if (prototypes.TryGetKindFrom<T>(out _) && prototypes.HasMapping<T>(node.Value))
```
PJB3005: Ditto

PJB3005 (changes requested):

Robust.Shared/Serialization/TypeSerializers/Implementations/EntProtoIdSerializer.cs
Tayrtahn: Apparently not.

Robust.Shared/Serialization/TypeSerializers/Implementations/EntProtoIdSerializer.cs
Tayrtahn: For some reason I thought that method didn't exist. Oops. Fixed.

PJB3005 (changes requested):
Uh, isn't this always going to say a missing prototype is abstract?

PR #5718, opened 2025-03-02
Analyzer & Fixer for redundant Prototype type strings

Tayrtahn:
Adds a code analyzer and fixer for Prototypes with manually set names that match the autogenerated names, just like the analyzer/fixer for DataFields.

The severity is currently set as warning, since there aren't as many instances of this as there are for DataFields. It can be reduced to info if preferred though.

Exceptions are made for matching names if they are not specified by string literals (i.e. using a const that resolves to a redundant name) or if the class name does not end with the word "Prototype".

Warning:
Screenshot 2025-03-01 at 7 21 14 PM
Quick fix:
Screenshot 2025-03-01 at 7 21 30 PM
After fix:
Screenshot 2025-03-01 at 7 21 43 PM

For implementation, the bulk of `CalculatePrototypeName` was moved from PrototypeManager to a new `PrototypeUtility` static class.

`InvalidPrototypeNameException` was deleted and replaced with an additional diagnostic that reports when a prototype using an autogenerated name does not end with the word "Prototype":
Screenshot 2025-03-01 at 8 58 40 PM
(shown as a warning here, but upgraded to an error in the PR, as violations cause a runtime error so we should catch it before then)


Tayrtahn:
Needs adjustment to ignore specified names that are not string literals. Will fix when I'm back from vacation.

Tayrtahn:
Fixed up and ready to go.

PR #5721, opened 2025-03-03
Cleanup: Remove redundant Prototype names

Tayrtahn:
Removes manually-specified name strings from all Prototype attributes where the autogenerated name would match.

Yes, this was just using #5718 and selecting Fix All -> Solution

~`AudioMetadataPrototype` was allowed to stay redundant since it's pulled from a const value which is used elsewhere and it's important that those values match.~ the analyzer has been updated so it permits redundancy if the attribute argument is not a string literal.

### DIFF
--- a/Robust.Shared/Audio/AudioMetadataPrototype.cs
+++ b/Robust.Shared/Audio/AudioMetadataPrototype.cs
@@ -9,7 +9,9 @@ namespace Robust.Shared.Audio;
 /// These prototypes get automatically generated when packaging the server,
 /// to allow the server to know audio lengths without shipping the large audio files themselves.
 /// </summary>
+#pragma warning disable RA0035 // Redundant Prototype Type specification
 [Prototype(ProtoName)]
+#pragma warning restore RA0035 // Redundant Prototype Type specification
 public sealed partial class AudioMetadataPrototype : IPrototype
 {
     public const string ProtoName = "audioMetadata";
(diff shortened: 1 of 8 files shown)


Robust.Shared/Audio/AudioMetadataPrototype.cs
```diff
@@ -9,7 +9,9 @@ namespace Robust.Shared.Audio;
 /// These prototypes get automatically generated when packaging the server,
 /// to allow the server to know audio lengths without shipping the large audio files themselves.
 /// </summary>
+#pragma warning disable RA0035 // Redundant Prototype Type specification
 [Prototype(ProtoName)]
+#pragma warning restore RA0035 // Redundant Prototype Type specification
```
PJB3005: Why can't this one be fixed properly?

Tayrtahn: It probably can? It uses a public const string defined in the class to set its name, and that value is used elsewhere:
https://github.com/space-wizards/RobustToolbox/blob/543088ea1feb3d45cbd106b220451cf310d748a4/Robust.Packaging/AssetProcessing/Passes/AssetPassAudioMetadata.cs#L65

I don't know why we would change the name of the prototype class, but if we ever did we would have to make sure to change the const too.

PJB3005: Then surely the correct fix is to make the analyzer not complain about this scenario?

Tayrtahn: Yeah, that would make sense. I'll see if I can get it to distinguish between a const and a literal.

PR #5725, opened 2025-03-05
Add ignore option for custom cursors

ghost:
Yep, a small tweak to ignore any cursor updates from UserInterfaceManager. It is need for custom implementation of cursors from the game code.


PJB3005 (commented):
I am not sure what the intended use case for this is. You want custom cursor shapes to not apply even when hovered over controls? What about controls that aren't just the default arrow cursor?

ghost:
> I am not sure what the intended use case for this is. You want custom cursor shapes to not apply even when hovered over controls? What about controls that aren't just the default arrow cursor?

Yes, i want to disable shape changing when you focused on control and do it by self from the game. For now i can only do it with frame update methods (every frame update) and apply custom cursor for every xaml controls. It is not good way and works some buggy from linux.

Well, it might be not so critical, if you want make your UI fully in viewport rendering, instead using RT's avalonia controls.
Because currently i working on the custom game instead of SS14, with using RT.

image

PJB3005:
> Yes, i want to disable shape changing when you focused on control and do it by self from the game. For now i can only do it with frame update methods (every frame update) and apply custom cursor for every xaml controls. It is not good way and works some buggy from linux.

Even when hovering over controls that normally need to change the cursor shape to e.g. a resize bar? Would you not want to be able to change these instead with a full set of cursor themes?

ghost:
> > Yes, i want to disable shape changing when you focused on control and do it by self from the game. For now i can only do it with frame update methods (every frame update) and apply custom cursor for every xaml controls. It is not good way and works some buggy from linux.
> 
> Even when hovering over controls that normally need to change the cursor shape to e.g. a resize bar? Would you not want to be able to change these instead with a full set of cursor themes?

Well, i just want to set up the custom cursor, like old RPG games (like The Witcher 1). 

And i don't know how to design cursor shapes, because UIManager use the enum values and _clyde. I don't want make another theme logic in the engine code, because it makes some limits for the game. Just let me set cursors by the game code, without override it by UIManager.

PJB3005:
Ok I changed my mind. This should be done with a proper theming system. Anything else is an accessibility trap that games would not want to use long-term.

PR #5735, opened 2025-03-09
Add checking for an unexpected audio state and freeing up buffers.

### LINKED ISSUE #5306: MacOS M1 OpenAL error with generating buffer

Sh18RW:
If `AL.GenBuffer` fails, it will return 0 as an id. If it fails two times, it can throw an exception by adding it to a dictionary with the id.
Also there is `Dispose` method calling from AudioStream deconstructor (fixes https://github.com/space-wizards/RobustToolbox/issues/5306).


PJB3005:
Isn't this just going to cause the game to constantly try to load audio files and crap out errors everywhere, likely nuking performance in the process? I don't see how this is much of an improvement over the previous status.

A proper fix would be to either move to an audio engine that doesn't have these limits limits (ideally) or dynamically unload audio (good regardless).

Sh18RW:
> A proper fix would be to either move to an audio engine that doesn't have these limits limits (ideally) or dynamically unload audio (good regardless).

I'm not completely sure what "dynamically unloading audio" means.

> Isn't this just going to cause the game to constantly try to load audio files and crap out errors everywhere, likely nuking performance in the process? I don't see how this is much of an improvement over the previous status.

When audio playback finishes, the sound entity (and all AudioStream references) gets removed. If I understand correctly, the AudioStream becomes unusable after this, so destructor cleanup should be safe.

That said, unloading audio could still throw exceptions due to the recurring buffer 0 issue (when adding to _audioSampleBuffers). The additional checks do help with debugging though and say that something goes wrong by throwing an exception.

I don't think this will hurt performance - at least I haven't observed the game endlessly retrying failed loads. And for the MacOS case, freeing buffers should prevent this specific error, while other OS users rarely encounter it anyway (RAM out).

PJB3005:
> If I understand correctly, the AudioStream becomes unusable after this, so destructor cleanup should be safe.

No? The audio stream contains the audio data and gets *reused* as long as the audio file is loaded. And we have no way to unload audio loaded from resources right now.

> I don't think this will hurt performance - at least I haven't observed the game endlessly retrying failed loads. And for the MacOS case, freeing buffers should prevent this specific error, while other OS users rarely encounter it anyway (RAM out).

If you throw an error on "audio failed to load", then the engine is just going to keep continually trying to re-load that file, because the resource manager has no way to block assets that failed loading (nor is this a good solution). This would utterly demolish performance.

Sh18RW:
> If you throw an error on "audio failed to load", then the engine is just going to keep continually trying to re-load that file, because the resource manager has no way to block assets that failed loading (nor is this a good solution). This would utterly demolish performance.

It will attempt to load the audio twice and drop loading. If a broken AudioStream file is stored in the cache, it may cause problems in the future, but if it causes an exception while loading, next time it will try to load the resource again instead of using broken one.

I have seen the discussion in Discord `Audio Strike Team`. Should I continue to do something worthwhile, or is some audio reworking planned?
(I think, I do not have the necessary knowledge...)

I have tried to do something like dynamic reloading in #5562, but it looks like a silly solution..

PJB3005:
> It will attempt to load the audio twice and drop loading.

No? This is not at all how the resource system works. It's going to try to load it every time the audio file is played again. I tested this by artificially incurring a sound load error once the client is running. The client becomes completely unusable with lag and error spam.

image

PJB3005:
Closing as this PR just doesn't work.

PR #5739, opened 2025-03-12
Ports GenIdStorage from Pow3r

TemporalOroboros:
Mainly exists to get it out of Pow3r and someplace its easier to re-use.

### DIFF
--- a/Robust.Shared/Collections/GenIdStorage.cs
+++ b/Robust.Shared/Collections/GenIdStorage.cs
@@ -1,7 +1,9 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using JetBrains.Annotations;
 using Robust.Shared.Utility;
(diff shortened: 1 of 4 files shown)


PJB3005:
What use case do you have in mind for this?

TemporalOroboros:
> What use case do you have in mind for this?

ECS node groups that
1. Don't rebuild the entire group whenever a single node is added/removed.
2. Don't rely on a hashset for group membership.
3. Don't need to resolve an additional node component to update list indices for every removed node.

PJB3005:
Is there anything in here that can't just be put in the sandbox in content?

TemporalOroboros:
Nothing at the moment.

PJB3005:
> Nothing at the moment.

I'd rather this be kept in content, then.

metalgearsloth:
I have thought of the usefulness for entities that probably should be recycled like audio or bullets... but also arch should fix that anyway so.

PJB3005:
> I have thought of the usefulness for entities that probably should be recycled like audio or bullets... but also arch should fix that anyway so.

I'd really rather it be internal in that case.

PJB3005:
Closing this as I really don't feel like this is something that needs to be a public API part of RT. It's too small, niche, and open to experimentation for that.

If we want it for internal use in RT later, we can add it.

PR #5741, opened 2025-03-12
Reduce EntityManager.IsDefault allocations

metalgearsloth:
We don't actually need the MappingDataNode so we can avoid allocating the entire class per entity.


metalgearsloth:
Also faster coz we can potentially early out now.

PR #5745, opened 2025-03-13
Add UpdateVisibilityMask method

metalgearsloth:
We tipped over to the point of systems stepping on each other's toes. Now we do the normal thing and just use the eventbus and it makes content a whole lot cleaner.

Non-breaking change.


SlamBamActionman (approved):
Does what it says on the tin

slarticodefast (approved):
Should be fine to merge on its own without the content PR.

PR #5750, opened 2025-03-17
Add button to jump to live chat when scrolled up

ciaran:
The intent here is to make it clear when the player is missing live incoming chat or radio comms, while also providing quick way to resume the live feed.

image

Initially I had experimented with a simple arrow icon 
image

but I’ve currently settled on having a button which matches the style of the other chat buttons around it.

There has been some feedback in Discord discussions but I’ve opened this PR to get directed, collected feedback on the UI look so I can work on whatever may need to be changed.

Also of note is that this change currently affects everywhere an `OutputPanel` is being used (e.g. the console, `csi`…). I see no reason this functionality wouldn’t be wanted anywhere there is an `OutputPanel` but it can be amended to only affect the chat feed if preferred.

### DIFF
--- a/Robust.Client/UserInterface/Controls/OutputPanel.cs
+++ b/Robust.Client/UserInterface/Controls/OutputPanel.cs
@@ -22,6 +22,8 @@ namespace Robust.Client.UserInterface.Controls
 
         public const string StylePropertyStyleBox = "stylebox";
 
+        public bool ShowScrollDownButton { get; set; } = false;
+
         private readonly RingBufferList<RichTextEntry> _entries = new();
         private bool _isAtBottom = true;
 
@@ -64,7 +66,7 @@ namespace Robust.Client.UserInterface.Controls
             _scrollBar.OnValueChanged += _ =>
             {
                 _isAtBottom = _scrollBar.IsAtEnd;
-                _scrollDownButton.Visible = !_isAtBottom;
+                _scrollDownButton.Visible = ShowScrollDownButton && !_isAtBottom;
             };
         }
 
@@ -206,7 +208,7 @@ namespace Robust.Client.UserInterface.Controls
             var styleBoxSize = _getStyleBox()?.MinimumSize.Y ?? 0;
 
             _scrollBar.Page = UIScale * (Height - styleBoxSize);
-            _scrollDownButton.Visible = !_scrollBar.IsAtEnd;
+            _scrollDownButton.Visible = ShowScrollDownButton && !_scrollBar.IsAtEnd;
             _invalidateEntries();
         }
 
--- a/Robust.Client/UserInterface/Controls/OutputPanel.cs
+++ b/Robust.Client/UserInterface/Controls/OutputPanel.cs
@@ -4,6 +4,7 @@ using Robust.Client.Graphics;
 using Robust.Client.UserInterface.RichText;
 using Robust.Shared.Collections;
 using Robust.Shared.IoC;
+using Robust.Shared.Localization;
 using Robust.Shared.Maths;
 using Robust.Shared.Utility;
 
@@ -26,6 +27,7 @@ namespace Robust.Client.UserInterface.Controls
         private bool _firstLine = true;
         private StyleBox? _styleBoxOverride;
         private VScrollBar _scrollBar;
+        private Button _scrollDownButton;
 
         public bool ScrollFollowing { get; set; } = true;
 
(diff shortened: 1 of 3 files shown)


slarticodefast:
Maybe something like this?
chat

sowelipililimute:
Screenshot_20250317-112525

i don't see any reason to diverge from the conventional floating button + badge design, but a smalltext banner works too

VerinSenpai:
wonderful

ciaran:
I’d really like to work on getting this pushed through but I’m not sure what the next steps are here, this is complicated by requiring a change in both repositories and people not agreeing on the way this should look.

I’m happy with the current appearance I settled on ([shown here](https://github.com/space-wizards/space-station-14/pull/35913#issue-2929952457)) but I’m happy to make any change if there is a concrete decision I can follow.

As for this repo, it would be good if someone could make the decision on whether we want this behaviour to be a global `OutputPanel` change or if it will be something specific to the client chat box.
If for some reason this behaviour is _not_ wanted on other `OutputPanel`s (e.g. debug console), then I could perhaps rework the feature and change this repo’s PR to be only expose some new `OutputPanel` events.

Robust.Client/UserInterface/Controls/OutputPanel.cs
```diff
@@ -43,7 +45,25 @@ public OutputPanel()
                 HorizontalAlignment = HAlignment.Right
             };
             AddChild(_scrollBar);
-            _scrollBar.OnValueChanged += _ => _isAtBottom = _scrollBar.IsAtEnd;
+
+            AddChild(_scrollDownButton = new Button()
+            {
+                Name = "scrollLiveBtn",
+                StyleClasses = { "chatScrollDownButton" },
```
PJB3005: 1. Make this a constant that is public.
2. Don't refer to it as "chat".

Robust.Client/UserInterface/Controls/OutputPanel.cs
```diff
@@ -43,7 +45,25 @@ public OutputPanel()
...
+                VerticalAlignment = VAlignment.Bottom,
+                HorizontalAlignment = HAlignment.Center,
+                Text = String.Format("⬇    {0}    ⬇", Loc.GetString("hud-output-scroll-down")),
```
PJB3005: If this text string is used by the engine, it should be included by the engine. That means "hud" is not appropriate naming and it needs to be made generic. Another option is to allow content to set the text used.

PJB3005 (changes requested):
This needs to be made opt-in per `OutputPanel` and cannot default to on.

Robust.Client/UserInterface/Controls/OutputPanel.cs
ciaran: 8c1ea57804c05d01ae60d788946dd8e87557d27a

Robust.Client/UserInterface/Controls/OutputPanel.cs
ciaran: 1bace71356bd4a9d4e4498f43282311d6b56a72a

ciaran:
> This needs to be made opt-in per `OutputPanel` and cannot default to on.

Note I’ve opted-in the `DebugConsole` to show the button for testing, this could be removed before merge until the styling is merged if needed.

Robust.Client/UserInterface/Controls/OutputPanel.cs
```diff
@@ -43,7 +49,25 @@ public OutputPanel()
...
+                MaxWidth = 300,
+                Visible = false,
+            });
+
+            _scrollDownButton.OnPressed += _ => ScrollToBottom();
+
+            _scrollBar.OnValueChanged += _ =>
+            {
+                _isAtBottom = _scrollBar.IsAtEnd;
+                _scrollDownButton.Visible = ShowScrollDownButton && !_isAtBottom;
```
PJB3005: This assignment logic is copy-pasted between the two locations.

Robust.Client/UserInterface/Controls/OutputPanel.cs
```diff
@@ -15,17 +16,22 @@ namespace Robust.Client.UserInterface.Controls
     [Virtual]
     public class OutputPanel : Control
     {
+        public const string StyleClassOutputPanelScrollDownButton = "outputPanelScrollDownButton";
+
         [Dependency] private readonly MarkupTagManager _tagManager = default!;
 
         public const string StylePropertyStyleBox = "stylebox";
 
+        public bool ShowScrollDownButton { get; set; } = false;
```
PJB3005: Update the visibility when this property changes.

PJB3005 (changes requested):

PJB3005 (approved):

PR #5755, opened 2025-03-19
Pass AnimationPlayerComponent in AnimationCompletionEvent

chromiumboy:
This PR makes it so the animation player component associated with the entity is passed when it raises the animation completion event

Required for content PR https://github.com/space-wizards/space-station-14/pull/35123

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
@@ -76,7 +76,7 @@ namespace Robust.Client.GameObjects
             foreach (var key in remie)
             {
                 component.PlayingAnimations.Remove(key);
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
```diff
@@ -210,5 +210,10 @@ public sealed class AnimationCompletedEvent : EntityEventArgs
         /// If false, it was removed prematurely via <see cref="AnimationPlayerSystem.Stop(Robust.Client.GameObjects.AnimationPlayerComponent,string)"/> or similar overloads.
         /// </summary>
         public bool Finished { get; init; }
+
+        /// <summary>
+        /// The entity animation player component
+        /// </summary>
+        public AnimationPlayerComponent? AnimationPlayer { get; init; }
```
metalgearsloth: I think all the places that use this have it as non-null so should be safe to make it non-nullable?

PR #5759, opened 2025-03-20
Fix showvelocities

metalgearsloth:
Can't use StateRoot anymore so just pretend it's a window.


slarticodefast:
Can you do the same for `showanchored`? That one is using StateRoot as well and is broken.

PR #5761, opened 2025-03-20
Add entity description as tooltip on entity spawn panel

Tayrtahn:
We have the description loaded from the prototype already, why not show it? This can help make it more obvious what an entity is before spawning it.

Screenshot 2025-03-20 at 5 49 50 PM


slarticodefast (approved):
Good idea!
An alternative to consider would be the protoId, which  can be useful as well, especially for debugging purposes.

PR #5767, opened 2025-03-23
Generic parallelmanager

metalgearsloth:
Forgor to do this after lookups and mapmanager teehee.

Haven't tested it yet but can do it later, and I forgot if CI even runs the parallel one.


PJB3005 (changes requested):
This doesn't reduce any allocs though? You're still boxing when inserting into the `InternalJob`.

metalgearsloth:
> This doesn't reduce any allocs though? You're still boxing when inserting into the `InternalJob`.

I should've drafted but I am looking at removing the pooling + the interface, just need to figure out a way around it.

PJB3005 (changes requested):
image

Now you're allocating here.

Robust.Shared/Threading/ParallelManager.cs
```diff
@@ -168,87 +184,70 @@ private ParallelTracker InternalProcess<T>(T job, int amount) where T : IParalle
         {
             var start = i * batchSize;
             var end = Math.Min(start + batchSize, amount);
-            var subJob = new InternalParallelJob<T>(job, tracker, _sawmill, start, end);
+            var subJob = new InternalParallelJobState<T>(job, tracker, _sawmill, start, end);
 
             // From what I can tell preferLocal is more of a !forceGlobal flag.
             // Also UnsafeQueue should be fine as long as we don't use async locals.
-            ThreadPool.UnsafeQueueUserWorkItem(subJob, true);
+            ThreadPool.UnsafeQueueUserWorkItem(Execute, subJob, true);
```
PJB3005: This allocates internally in the .NET runtime.

image

What you're trying to do *isn't possible*. Make the job objects classes and force consumers to pool when appropriate. Many of them already do in effect.

PJB3005 (changes requested):

PJB3005:
Seriously, this is getting old.

metalgearsloth:
> Seriously, this is getting old.

I'm sorry I am trying my best here.

PJB3005 (commented):
What does making it generic actually achieve now that they're all classes, anyways?

metalgearsloth:
> What does making it generic actually achieve now that they're all classes, anyways?

Laziness from when I had already changed it :trollface:

PR #5769, opened 2025-03-24
Cleanup warnings in WordWrap

Tayrtahn:
Stupidly simple: just replaces calls to the static `Logger` with calls to an IoC resolved `ISawmill`. The static version uses IoC internally anyway, so this change doesn't break anything.

### DIFF
--- a/Robust.Client/UserInterface/WordWrap.cs
+++ b/Robust.Client/UserInterface/WordWrap.cs
@@ -2,6 +2,7 @@
 using System.Diagnostics.Contracts;
 using System.Text;
 using Robust.Client.Graphics;
+using Robust.Shared.IoC;
 using Robust.Shared.Log;
 using Robust.Shared.Utility;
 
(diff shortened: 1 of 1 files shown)


PJB3005 (changes requested):
The point of deprecating `Logger` is to avoid doing this exact pattern. This is not a solution.

Tayrtahn:
@PJB3005 Is this better, or does it need a deeper solution? It seems to avoid the threading issues that the static logger has.

PJB3005 (changes requested):
The point is to *stop using `IoCManager`*.

PJB3005:
Also we don't even have an `ISawmill` in IoC so your previous code didn't even work.

Tayrtahn:
> The point is to *stop using `IoCManager`*.

Gotcha. Sounds like the whole word wrap struct will need to be redesigned for that, so I'm gonna drop this until some future day when I have a better idea of how to do that. Thanks for taking the time to review!

PR #5772, opened 2025-03-25
equatable FormattedMessage

Errant-4:
I need this so GroupBy works on them, for the GhostRole pick UI

Code test
image

### DIFF
--- a/Robust.Shared/Utility/FormattedMessage.cs
+++ b/Robust.Shared/Utility/FormattedMessage.cs
@@ -16,7 +16,7 @@ namespace Robust.Shared.Utility;
 /// </summary>
 [PublicAPI]
 [Serializable, NetSerializable]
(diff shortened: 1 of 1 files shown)


PJB3005 (approved):
Not a fan of this having to convert to markup (expensive allocation), but I guess it's fine.

Errant-4:
If ToString() is meaningfully cheaper, we can use that instead, it is not likely to ever make a difference in functionality. But so far the only thing that ever uses this is the ghostrole UI. It gets called once for each open ghostrole when a player opens the window

PJB3005:
> If ToString() is meaningfully cheaper, we can use that instead, it is not likely to ever make a difference in functionality. But so far the only thing that ever uses this is the ghostrole UI. It gets called once for each open ghostrole when a player opens the window

The correct, optimized thing to do is to actually do equality checks on all of the tags as C# objects, not to serialize it to a string.

PR #5773, opened 2025-03-25
Fix DirectoryNotFoundException when saving map or grid on Unix systems

JerryImMouse:
Someone forgor to create directory before saving map or grid, this caused an exception on trying to save smth anywhere else than `/`.

Probably SafeFileHandle.Unix.cs doesn't handle directory creating itself as on Windows(I haven't checked), so it worked on most local servers, but not on production servers.

This fix has been tested on local linux server.

An exception itself:
image

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -47,6 +47,7 @@ END TEMPLATE-->
 ### Bugfixes
 
 * Fix serialization source generator breaking if a class has two partial locations.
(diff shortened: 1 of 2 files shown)


JerryImMouse:
Also, this fix is present only because of such case is not covered in unit tests...

ElectroJr:
Ah thanks for fixing that, I must've accidentally dropped the createdir call in somewhere in #5572

PR #5779, opened 2025-03-26
Add OtherBody API to contacts

metalgearsloth:
Thought I had a pr for this.

### DIFF
--- a/Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
+++ b/Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
@@ -417,6 +417,17 @@ namespace Robust.Shared.Physics.Dynamics.Contacts
 
             throw new InvalidOperationException();
         }
+
+        [Pure, PublicAPI]
+        public PhysicsComponent OtherBody(EntityUid uid)
+        {
+            if (uid == EntityA)
+                return BodyB!;
+            else if (uid == EntityB)
+                return BodyA!;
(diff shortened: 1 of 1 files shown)


Robust.Shared/Physics/Dynamics/Contacts/Contact.cs
```diff
@@ -417,6 +417,17 @@ public EntityUid OtherEnt(EntityUid uid)
...
+        [Pure, PublicAPI]
+        public PhysicsComponent OtherBody(EntityUid uid)
+        {
+            if (uid == EntityA)
+                return BodyB!;
+            else if (uid == EntityB)
+                return BodyA!;
+
+            throw new InvalidOperationException();
+        }
```
slarticodefast, line 441:
```suggestion

        [Pure, PublicAPI]
        public PhysicsComponent OurBody(EntityUid uid)
        {
            if (uid == EntityA)
                return BodyA!;
            else if (uid == EntityB)
                return BodyB!;

            throw new InvalidOperationException();
        }

        [Pure, PublicAPI]
        public PhysicsComponent OtherBody(EntityUid uid)
        {
            if (uid == EntityA)
                return BodyB!;
            else if (uid == EntityB)
                return BodyA!;

            throw new InvalidOperationException();
        }
```
Might as well add this one, too. All the other fields got both versions.

slarticodefast (approved):

PR #5780, opened 2025-03-26
equatable FormattedMessage, take 2

Errant-4:
No more markup conversions. Addressing the point/request brought up on #5772


Robust.Shared/Utility/FormattedMessage.cs
```diff
@@ -281,13 +281,30 @@ IEnumerator<MarkupNode> IEnumerable<MarkupNode>.GetEnumerator()
...
-        return ToMarkup().GetHashCode();
+        var hash = 0;
+        var i = 1;
+        foreach (var node in _nodes)
+        {
+            hash += node.GetHashCode() * i;
+            i++;
+        }
+
+        return hash;
```
PJB3005: Use `System.HashCode` to combine these instead, I don't trust that this will behave correctly.

PJB3005 (changes requested):

PR #5782, opened 2025-03-26
fully remove the obsolete version of TryGetContainingContainer

slarticodefast:
All the remaining cases in the content repo will automatically cast to the `Entity<TransformComponent?, MetaDataComponent?>` variant, which also gets rid of all the warnings.

Currently what people are doing to get rid of the warning is this
grafik


slarticodefast:
Duplicate of https://github.com/space-wizards/RobustToolbox/pull/5660
Please just kill it already
this is the third PR

PR #5783, opened 2025-03-27
Make MappingDataNode use string keys

ElectroJr:
Requires space-wizards/space-station-14/pull/36111

This PR limits yaml mappings/dictionaries to using only string keys. This effectively changes MappingDataNode from a wrapper around `Dictionary<DataNode,DataNode>` to `Dictionary<string, DataNode>`. This was already brought up a few times in maintainer meetings, with the main reasons for this change including:
- Reducing allocations
- Removing the `static readonly ThreadLocal<ValueDataNode> FetchNode` jank
- Preventing issues due to using mutable objects as dictionary keys (nothing prevents you from changing the value or hashcode of a `ValueDataNode` after it has been used as a dictionary key) 
- Probably improves performance for adding or fetching entries.

This is obviously a breaking change, though I've added some helper methods to `MappingDataNode` that take in a `ValueDataNode` to make updating the engine version somewhat easier. Aside from that, two other issues came up while working on this PR

Firstly, yaml validators need to know the position of a key in the yaml file to report errors. Replacing the DataNode keys with strings removes this information. To fix this, mapping nodes loaded from a yaml file will still store the original ValueDataNode which can be retrieved via `MappingDataNode.GetKeyNode()`. However AFAIK this is only actually required when running the yaml linter, and should probably still somehow be disabled by default. And it could probably also be reworked to only store the `NodeMark` struct instead of the whole node, but its good enough for now.

Secondly, the tuple serializer currently only supports 2-item tuples and serializes them as a mapping with a single entry. I.e., `("foo", "bar")` serializes to `{ foo: bar }`. Given that mappings now require string keys, this would make it impossible to serialize tuples where the first item is null or can't be serialized as a yaml string. So this PR changes the serializer to use sequences instead (i..e, `[foo, bar]`). It can still try to read mappings, but actually reading the mapping might fail due to the other changes, which would require any old prototypes or maps to be manually migrated to using a sequence node.  It seems like tuple data fields aren't all that common, and this issue didn't seem to come up at all in ss14, so maybe its fine? I also think this is a change we should make anyways, because using mapping nodes obviously wouldn't work if we ever wanted to support tuple data fields with 3+ items, so they really should've been using sequence nodes from the very beginning.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -35,7 +35,14 @@ END TEMPLATE-->
 
 ### Breaking changes
 
-*None yet*
+* Yaml mappings/dictionaries now only support string keys instead of generic nodes
+  * Several `MappingDataNode` method arguments or return values now use strings instead of a DataNode object
+  * The MappingDataNode class has various helper methods that still accept a `ValueDataNode`, but these methods are marked as obsolete and may be removed in the future.
+  * yaml validators should use `MappingDataNode.GetKeyNode()` when validating mapping keys, so that errors can print node start & end information
+* ValueTuples yaml serialization has changed
+  * Previously they would get serialized into a single mapping with one entry (i.e., `{foo : bar }` 
+  * Now they serialize into a sequence (i.e., `[foo, bar]`
+  * The ValueTuple serializer will still try to read mappings, but may fail if the previously serialized "key" can't be read as a simple string
 
 ### New features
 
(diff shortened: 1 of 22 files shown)


metalgearsloth (approved):
Didn't see anything amiss but pjb looks to be reviewing again so.

metalgearsloth:
Shipping this if no objections will check back soon.

PR #5788, opened 2025-03-29
Fix popup text overflowing the sides of the screen

DrSmugleaf:
Most noticeable with long ban reasons such as datacenter/vpn bans

### DIFF
--- a/Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
+++ b/Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
@@ -5,7 +5,6 @@ using Robust.Client.Graphics;
 using Robust.Client.UserInterface.Controls;
 using Robust.Client.UserInterface.CustomControls;
 using Robust.Shared.Localization;
-using Robust.Shared.Log;
 using Robust.Shared.Maths;
 using Robust.Shared.Profiling;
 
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
```diff
@@ -61,7 +60,7 @@ public void Popup(string contents, string? title = null, bool clipboardButton =
                 Title = string.IsNullOrEmpty(title) ? Loc.GetString("popup-title") : title,
             };
 
-            var label = new Label { Text = contents };
+            var label = new RichTextLabel { Text = $"[color=white]{contents}[/color]" };
```
PJB3005: You're gonna need to escape the contents here to avoid any markup being interpreted.

PJB3005 (changes requested):

Robust.Client/UserInterface/UserInterfaceManager.Layout.cs
DrSmugleaf: Fixed in [`fba6e11` (#5788)](https://github.com/space-wizards/RobustToolbox/pull/5788/commits/fba6e110f8003643fb5288d04bc9c2407e0eeae9)

PR #5790, opened 2025-03-30
Add TryDirtyField

metalgearsloth:
Avoids doing the equatable checks every time and engine handles it.

I also want to look at a field version that would hopefully mean the caller doesn't need to pass in the string name.


metalgearsloth:
Will writeup an issue but had ideas on how to do it without field shenanigans and avoiding needing both the field + string fieldname.

PR #5792, opened 2025-03-31
set default pitch variation for all sounds at -0.08

Metekillot:
This adjusts the default pitch variation for all sounds to -0.08f. The current default is 'null', which (in my experience and for many other players) results in significant ear fatigue; in my case it also results in headaches.

A value of -0.08 reduces ear fatigue without significantly altering how recognizable a given sound is for the purposes of gameplay dynamics such as overhearing nefarious activities or knowing who is walking around on the other side of a wall. I intend to make other modifications to some of the sound systems for Space Station 14, but this variable here is defined within the game engine itself. Down the line it may be wiser to define sound defaults as their own .yml, but at this time I think this sufficiently addresses the problems of ear fatigue.

No variation:

https://github.com/user-attachments/assets/a81a8213-f60f-4b27-83f1-e4b7a176c7db

-0.08f variation:

https://github.com/user-attachments/assets/d149a480-f504-4704-a647-577765cd886d

### DIFF
--- a/Robust.Shared/Audio/AudioParams.cs
+++ b/Robust.Shared/Audio/AudioParams.cs
@@ -93,14 +93,14 @@ namespace Robust.Shared.Audio
         ///     If not null, this will randomly modify the pitch scale by adding a number drawn from a normal distribution with this deviation.
         /// </summary>
         [DataField]
-        public float? Variation { get; set; } = null;
+        public float? Variation { get; set; } = Default.Variation;
 
         // For the max distance value: it's 2000 in Godot, but I assume that's PIXELS due to the 2D positioning,
         // so that's divided by 32 (EyeManager.PIXELSPERMETER).
         /// <summary>
         ///     The "default" audio configuration.
         /// </summary>
-        public static readonly AudioParams Default = new(0, 1, SharedAudioSystem.DefaultSoundRange, 1, 1, false, 0f);
+        public static readonly AudioParams Default = new(0, 1, SharedAudioSystem.DefaultSoundRange, 1, 1, false, 0f, -0.08f);
 
         // explicit parameterless constructor required so that default values get set properly.
         public AudioParams() { }


PJB3005:
I don't understand, why 0.08? Is that a value specific to SS14's selection of sounds?

metalgearsloth:
I'd also make it a cvar as well I think.

Metekillot:
> I don't understand, why 0.08? Is that a value specific to SS14's selection of sounds?

It's from personal experimentation. I'm not trained in sound design but I've noticed most of the sounds trend to higher pitches already, and introducing positive pitch variation ended up making them more grating. I tried variations from .2 to -.1, and -0.08 was where I noticed a distinct lack of ear fatigue while modulating the sounds in a way that positively contributed to immersion. The specific -0.08 may owe to the sound design of SS14; I don't have a control set of repetitive sound effects to test the game sound engine with to determine what -- if any, since the audio parameters don't seem to do anything out of the ordinary -- pitch variation would otherwise be the "ideal".

Metekillot:
> I'd also make it a cvar as well I think.

Letting people fiddle with the sound too much -- beyond volume -- gives me an uneasy feeling around the nature of sound as a core gameplay mechanic, and leaving it in the runtime hands of admins seems like it would lead to constant ear annihilating "jokes" in downstream servers

metalgearsloth:
> > I'd also make it a cvar as well I think.
> 
> Letting people fiddle with the sound too much -- beyond volume -- gives me an uneasy feeling around the nature of sound as a core gameplay mechanic, and leaving it in the runtime hands of admins seems like it would lead to constant ear annihilating "jokes" in downstream servers

This is an engine PR so you should come at it from the perspective of "what would a non-ss14 game want to do" which is probably use a cvar to use their own values.

Metekillot:
Fair point

PR #5793, opened 2025-03-31
Added  a copy visuals function to the point light system

Scyarah:
# Background
While working on my pull request for a `ChameleonHardsuit`in the Wizden repository I ran into the problem, that I had to copy details from one `PointLightComponent` to another. Since only the PointLightSystem had full access to those properties, I had to add it in there. I hope that is not a problem.

# Technical details
- Added a CopyVisuals to the SharedPointLightSystem. This right now copies the relevant properties
- Added a CopyVisuals to the PointLightSystem which calls SetMask which was required to have the visual cone show up.

# PR in Wizden Repo
[Chameleon Hardsuit PR](https://github.com/space-wizards/space-station-14/pull/35798)

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedPointLightSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedPointLightSystem.cs
@@ -92,26 +92,26 @@ public abstract class SharedPointLightSystem : EntitySystem
     /// <summary>
     ///     Copy visuals from one pointlight to another.
     /// </summary>
-    public void CopyVisuals(EntityUid uid, SharedPointLightComponent otherPointLight, SharedPointLightComponent? pointlight = null, bool forceDisabled = false)
+    public void CopyVisuals(Entity<SharedPointLightComponent?> pointLightEntity, SharedPointLightComponent? otherPointLight = null, bool forceDisabled = false)
     {
-        if (!Resolve(uid, ref pointlight))
+        if (pointLightEntity.Comp is not SharedPointLightComponent pointLight || otherPointLight == null)
             return;
 
-        pointlight.Color = otherPointLight.Color;
-        pointlight.Energy = otherPointLight.Energy;
-        pointlight.Offset = otherPointLight.Offset;
-        pointlight.Radius = otherPointLight.Radius;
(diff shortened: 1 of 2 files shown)


Robust.Shared/GameObjects/Systems/SharedPointLightSystem.cs
```diff
@@ -87,4 +87,31 @@ public void SetSoftness(EntityUid uid, float value, SharedPointLightComponent? c
         comp.Softness = value;
         Dirty(uid, comp);
     }
+
+
+    /// <summary>
+    ///     Copy visuals from one pointlight to another.
+    /// </summary>
+    public void CopyVisuals(EntityUid uid, SharedPointLightComponent otherPointLight, SharedPointLightComponent? pointlight = null, bool forceDisabled = false)
```
Tayrtahn: Use the modern `Entity<T>` way of passing in a paired uid and component.
```suggestion
    public void CopyVisuals(Entity<SharedPointLightComponent?> pointlight, SharedPointLightComponent otherPointLight, bool forceDisabled = false)
```

Scyarah: Changed it, thanks for the hint! This is my first PR in both RobustToolbox and SpaceStation overall so I am glad for every suggestion.

metalgearsloth:
Isn't this just serializationmanager.copyto?

Scyarah:
> Isn't this just serializationmanager.copyto?

I guess kind of? This is my first PR for both RobustToolbox and Spacestation and I mostly just tried to keep it similiar to the other code in the ChameleonSystem after some other experiments. I do however not want to copy everything, for example I want enabled to stay as it is.

metalgearsloth:
ChameleonSystem is doing the wrong things.

Ideal world you just call ISerializationManager.CopyTo and then you get what is needed. Worst case you add a typeserializer to handle mask setting but otherwise yeah, this shouldn't be necessary for raw datafield copying.

Scyarah:
I am completely redoing my initial PR now, so I don't think this one will be needed anymore!

PR #5799, opened 2025-03-31
Add Fluent function for possessive nouns (thing's vs things')

Tayrtahn:
Adds a new function for localization that can be used to correctly format the possessive form of arbitrary strings.

For example:
`This is { POSS-NOUN($input) } house`
"Bob" -> "This is Bob's house"
"Chris" -> "This is Chris' house"

Works well in combos:
`This is { THE(POSS-NOUN($entity)) } favorite`
MobMouse -> "This is the mouse (234)'s favorite"
DrinkShotGlass -> "This is the shot glass' favorite"
BaseMobHuman-> "This is Urist McHands' favorite"

### DIFF
--- a/Robust.Shared/Localization/LocalizationManager.Functions.cs
+++ b/Robust.Shared/Localization/LocalizationManager.Functions.cs
@@ -40,6 +40,7 @@ namespace Robust.Shared.Localization
             // Misc
             AddCtxFunction(bundle, "ATTRIB", args => FuncAttrib(bundle, args));
             AddCtxFunction(bundle, "CAPITALIZE", FuncCapitalize);
+            AddCtxFunction(bundle, "APOSTROPHE-S", FuncApostropheS);
             AddCtxFunction(bundle, "INDEFINITE", FuncIndefinite);
         }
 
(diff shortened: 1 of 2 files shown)


perryprog:
Okay this is such a nitpick, but it's a thing that tends to bother me so I want to mention it for posterity.

It's a bit debatable whether the possessive pronoun of a singular noun that ends with "s" is formed by adding just `'` or adding `'s`. (That is, `Chris'` versus `Chris's`.) The rule of thumb I tell most people is you should add the extra `s` if that's how you say that—most people do say [kɹɪsɪz] (Chris-is) (I think I sometimes hear [kɹɪsʔz] (Chris-glottal stop-z) which is weird but similar), so in writing you'd (again, arguably) prefer `Chris's` as the written possessive.

ANYWAY: most US style guides (Chicago, AP, MLA, and APA are the ones I know say this—I don't know any that don't suggest this, actually) recommend using the `s's` form, so I would suggest always using that form for this feature.

Note that this only applies to singular nouns, though: it's well agreed that `54 slugcats` should have its possessive written as `54 slugcats'`.

(The one exception I know (and probably most commonly cited) is `Jesus'` since apparently some people pronounce the possessive as [dʒiːzəs], though I personally say it as [dʒiːzəsɪz] so who knows.)

Tayrtahn:
Huh, now you've got me thinking! 

As someone whose name ends with an S, I have always preferred the s' form, even in the singular case, to the point that I've occasionally praised/grumbled at a piece of software for following/not following that style. 

You make a good point about pronunciation! I absolutely would pronounce the possessive of Chris as "Chris-is", which I must admit makes a pretty compelling case for the s's form. It still looks wrong to me, but that's just my opinion. 

Unfortunately, I don't think we have a good way to distinguish between singular and plural nouns, and I'm not sure how sensible that would be anyway - an entity is (probably?) always going to be singular (a dog, a piece of paper, a swarm of bees), and while I believe Fluent can associate data with arbitrary strings, we don't really make use of that feature. 

Anyway, I think what I'll do for now is remove the use of this function from the content PR I made it for, since it's not really worth it. I'll keep this PR open for consideration.

PR #5804, opened 2025-04-02
Add an AudioParams.AddVariation() function akin to AddVolume()

JokeCam:
Required by https://github.com/space-wizards/space-station-14/pull/36259


Robust.Shared/Audio/AudioParams.cs
```diff
@@ -164,6 +164,17 @@ public readonly AudioParams WithVariation(float? variation)
...
         }
 
+        /// <summary>
+        ///     Returns a copy of this instance with a modified variation set, for easy chaining.
+        /// </summary>
+        [Pure]
+        public readonly AudioParams AddVariation(float? variation)
+        {
+            var me = this;
+            me.Variation += variation;
```
PJB3005: Given that this variation is used for a normal distribution, is a linear add like this mathematically correct?

PJB3005: Also, this addition doesn't work if `Variation` is null: 
image

PJB3005 (dismissed):

Robust.Shared/Audio/AudioParams.cs
JokeCam: Switched to scailing multiplier, and added a check for null value.
Also damn, the issue had a beginner friendly tag, didn't know I would need to learn what a normal distribution is😄

slarticodefast (changes requested):
Adding variances of normal distributions instead of multiplying them was the mathematically correct way to handle this.

Example: Having a variance of 0 would be equivalent to "no variation" and adding that to any existing value should not change anything.
Multiplying variances would however bulldoze the previous value in this case.

And the default value of 0.1f you picked is completely arbitrary, if the existing variance is null then it should just use the added value.

Robust.Shared/Audio/AudioParams.cs
```diff
@@ -164,6 +164,20 @@ public readonly AudioParams WithVariation(float? variation)
...
+        /// <summary>
+        ///     Returns a copy of this instance with a modified variation set, for easy chaining.
+        /// </summary>
+        [Pure]
+        public readonly AudioParams AddVariation(float variation)
+        {
+            var me = this;
+            if (variation == 0f)
+                return me;
+
```
JokeCam, line 176: me.Variation - "If not null, this will randomly modify the pitch scale by adding a number drawn from a normal distribution with this deviation."
Adding variation of 0f without this if statement will make variation not null, which is bad? I assumed so and that's why I added this if statement.

Robust.Shared/Audio/AudioParams.cs
```diff
@@ -164,6 +164,20 @@ public readonly AudioParams WithVariation(float? variation)
...
         }
 
+        /// <summary>
+        ///     Returns a copy of this instance with a modified variation set, for easy chaining.
+        /// </summary>
+        [Pure]
+        public readonly AudioParams AddVariation(float variation)
+        {
+            var me = this;
+            if (variation == 0f)
```
slarticodefast:
```suggestion
        public readonly AudioParams AddVariation(float? variation)
        {
            var me = this;
            if (variation == null)
```

slarticodefast (approved):
This should work now, but I'll let someone else double check.

PR #5808, opened 2025-04-03
Add ForbidLiteralAttribute and analyzer

Tayrtahn:
Adds an analyzer that warns when literal values are passed to methods for parameters marked with the `[ForbidLiteral]` attribute.

If the parameter is a collection expression  (`["foo", "bar"]`), the analyzer will check each value. This works for `params` as well.

"Forbid" is a bit misleading, since it's just a warning. I'm open to suggestions for a better name.

Created to deal with this problem: https://discord.com/channels/310555209753690112/900426319433728030/1357196400550744145

`TagSystem` in content has seen some string literal abuse with its `HasTag` method - it's being called using a string literal for the tag, which has at least two problems: there is no validation being done to make sure that these are actual tags, and relying on literal values deep within a class ("magic strings") is sloppy programming that's just asking for trouble (these values should be extracted into const/static fields at the top of the class for easier reuse and maintenance).

This can probably be used to combat the use of magic strings/numbers in other places too.

### DIFF
--- a/Robust.Analyzers/ForbidLiteralAnalyzer.cs
+++ b/Robust.Analyzers/ForbidLiteralAnalyzer.cs
@@ -32,6 +32,37 @@ public sealed class ForbidLiteralAnalyzer : DiagnosticAnalyzer
         context.RegisterOperationAction(AnalyzeOperation, OperationKind.Invocation);
     }
 
+    private void AnalyzeOperation(OperationAnalysisContext context)
+    {
+        if (context.Operation is not IInvocationOperation invocationOperation)
+            return;
+
+        // Check each parameter of the method invocation
+        foreach (var argumentOperation in invocationOperation.Arguments)
+        {
+            // Check for our attribute on the parameter
+            if (!AttributeHelper.HasAttribute(argumentOperation.Parameter, ForbidLiteralType, out _))
+                continue;
+
+            // Handle parameters using the params keyword
+            if (argumentOperation.Syntax is InvocationExpressionSyntax subExpressionSyntax)
+            {
+                // Check each param value
+                foreach (var subArgument in subExpressionSyntax.ArgumentList.Arguments)
+                {
+                    CheckArgumentSyntax(context, argumentOperation, subArgument);
+                }
+                continue;
+            }
+
+            // Not params, so just check the single parameter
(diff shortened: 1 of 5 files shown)


slarticodefast:
Can we do the same for other cases where `ProtoId` should be used instead of a string?
I found a few cases for methods like `Spawn` or `SpawnEntity` using hardcoded strings.
Another case would be `ToolSystem.HasQuality`. And I'm sure there are more.

Edit: Oops, wanted to answer on the content PR.

Tayrtahn:
Yeah, it can check for any sort of literal being used as a parameter - magic numbers begone!

If you want to try it out, check out this engine PR, slap the attribute on some method parameters and run `dotnet clean; dotnet build | grep RA0033` and see what you get!

Robust.Analyzers/ForbidLiteralAnalyzer.cs
```diff
@@ -0,0 +1,102 @@
...
+        "Pass in a validated wrapper type like ProtoId, or a const or static value."
+    );
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => [ForbidLiteralRule];
+
+    public override void Initialize(AnalysisContext context)
+    {
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+        context.EnableConcurrentExecution();
+        //context.RegisterSyntaxNodeAction(AnalyzeExpression, SyntaxKind.InvocationExpression);
```
ljm862: Unused code here!

Tayrtahn: Whoops! A remnant of an earlier attempt. Thanks for catching that!

K-Dynamic:
> Created to deal with this problem: https://discord.com/channels/310555209753690112/900426319433728030/1357196400550744145

Can you post a screenshot or describe the issue for archival purposes

Tayrtahn:
> Can you post a screenshot or describe the issue for archival purposes

Screenshot 2025-04-03 at 6 58 15 PM

(also added an explanation to the PR description)

PJB3005 (approved):

PR #5810, opened 2025-04-04
Flick implementation for AnimationPlayer

EmoGarbage404:
Adds a mirrored implementation of Byond's `Flick()` for AnimationPlayer.

This is a function you can call from server or shared code to run a state's animation on a target layer. The goal is to reduce lots of duplicated content code that exists simply to network state and run animation boilerplate. 

No breaking changes.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
@@ -1,26 +1,81 @@
 using System;
 using System.Collections.Generic;
 using Robust.Client.Animations;
+using Robust.Client.Graphics;
+using Robust.Shared.Animations;
 using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
+using Robust.Shared.Timing;
 using Robust.Shared.Utility;
 
 namespace Robust.Client.GameObjects
 {
-    public sealed class AnimationPlayerSystem : EntitySystem
+    public sealed class AnimationPlayerSystem : SharedAnimationPlayerSystem
     {
         private readonly List<Entity<AnimationPlayerComponent>> _activeAnimations = new();
 
         private EntityQuery<AnimationPlayerComponent> _playerQuery;
+        private EntityQuery<SpriteComponent> _spriteQuery;
         private EntityQuery<MetaDataComponent> _metaQuery;
 
         [Dependency] private readonly IComponentFactory _compFact = default!;
+        [Dependency] private readonly IGameTiming _timing = default!;
 
         public override void Initialize()
         {
             base.Initialize();
+
+            SubscribeNetworkEvent<AnimationFlickEvent>(OnAnimationFlick);
+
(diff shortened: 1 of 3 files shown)


metalgearsloth:
We really need animation in shared lol

Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
```diff
@@ -1,28 +1,87 @@
...
-    public sealed class AnimationPlayerSystem : EntitySystem
+    public sealed class AnimationPlayerSystem : SharedAnimationPlayerSystem
     {
         private readonly List<Entity<AnimationPlayerComponent>> _activeAnimations = new();
 
         private EntityQuery<AnimationPlayerComponent> _playerQuery;
+        private EntityQuery<SpriteComponent> _spriteQuery;
         private EntityQuery<MetaDataComponent> _metaQuery;
 
         [Dependency] private readonly IComponentFactory _compFact = default!;
```
metalgearsloth, line 21: Can you also use the entitymanager one while you're here lol.

Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
```diff
@@ -1,28 +1,87 @@
...
+        {
+            if (!TryGetEntity(ev.Entity, out var uid))
+                return;
+            Flick(uid.Value, ev.StateId, ev.LayerKey);
+        }
+
+        /// <inheritdoc/>
+        public override void Flick(EntityUid uid, string stateId, object layerKey)
+        {
+            if (!_timing.IsFirstTimePredicted)
```
slarticodefast, line 45: Does this work if you have a non-predicted server event?
Best give it a test with multiple clients and see if the second client can see a flick animation caused by the first one.

Robust.Shared/Animations/SharedAnimationPlayerSystem.cs
```diff
@@ -0,0 +1,31 @@
...
+
+public abstract class SharedAnimationPlayerSystem : EntitySystem
+{
+    /// <summary>
+    /// Plays an <see cref="AnimationTrackSpriteFlick"/> for a target entity for the full duration of the state.
+    /// </summary>
+    /// <param name="uid">The entity to play the animation on.</param>
+    /// <param name="stateId">The state we are playing.</param>
+    /// <param name="layerKey">A key for the sprite layer. Must be NetSerializable.</param>
+    public abstract void Flick(EntityUid uid, string stateId, object layerKey);
```
slarticodefast, line 15: This will likely cause prediction issues - the client first predicts it and then receives the server event causing a second flick. So you might need an API similar to popups or audio where you pass in a user the animation was predicted for, who will be exempt from receiving the server event.
I kinda hate that you need the user though and that will cause some of the same problems we have with popups - so maybe it's possible to take a different approach using appearance data which would get synced between server and client even without the user.

PR #5812, opened 2025-04-06
Remove sharing sessions clientside, add DisplayName

SlamBamActionman:
This PR removes sharing ICommonSessions clientside for other players; the session data includes information that may be desired to keep hidden from clients, such as account name and unique ID. The client's own session is still provided (necessary for state generation).

DisplayName is a new property for ICommonSessions that is meant to be used in instances where a player may wish to be anonymous.
`GetDisplayName` gets the session's `DisplayName` if set, otherwise defaults to the `Name`.
`GetTruthfulNameString` gets the session's `Name` and, if set, `DisplayName` in the format `"Name (DisplayName)"`. Useful for admin tools such as ahelp.

Since replays are used for administration purposes, they are provided the full session list.
Because the sessions aren't shared, *only* the client's own entity is given the `ActorComponent` component clientside. They still exist on the serverside for all relevant entities though.

This implementation has been made for primarily admins in mind, however if a true player-controlled anonymous mode is desired, this should lay the groundwork for it.


ScarKy0:
Peak

Robust.Shared/Player/DummySession.cs
```diff
@@ -72,6 +73,11 @@ public void SetName(string name)
         };
     }
 
+    public void SetDisplayName(string? name)
+    {
+        throw new NotSupportedException();
+    }
+
```
metalgearsloth: Any particular reason it doesn't just set it?

Robust.Client/Player/PlayerManager.cs
```diff
@@ -269,6 +269,7 @@ private bool UpdatePlayerList(IEnumerable<SessionState> remotePlayers, bool full
 
                 // Check if the data is actually different
                 if (session.Name == state.Name
+                    &&session.DisplayName == state.DisplayName
```
metalgearsloth:
```suggestion
                if (session.Name == state.Name
                    && session.DisplayName == state.DisplayName
```

metalgearsloth (changes requested):

metalgearsloth:
pjb likely has an interest in reviewing this so I would also vet it with her.

Robust.Shared/Player/DummySession.cs
SlamBamActionman: Just a holdover from a previous implementation that I scrapped. Fix'd.

Tayrtahn:
I feel like not having the Actor component on clients might break some predicted things in content, but I'm not sure. It's worth doing anyway, but we should find any instances of that and be prepared to fix them.

Robust.Server/Player/PlayerManager.cs
```diff
@@ -130,17 +130,20 @@ private void HandlePlayerListReq(MsgPlayerListReq message)
             session.ConnectedTime = DateTime.UtcNow;
             SetStatus(session, SessionStatus.Connected);
 
-            var list = new List<SessionState>();
-            foreach (var client in players)
+
+            var list = new List<SessionState>() { new() { UserId = session.UserId, Name = session.Name, DisplayName = session.DisplayName, Status = session.Status } };
+            // Commented this out if it ends up we'll need this in the future;
+            /*foreach (var client in players)
```
ElectroJr: Use a cvar

Robust.Server/GameStates/PvsSystem.Session.cs
```diff
@@ -45,7 +45,12 @@ internal void ComputeSessionState(PvsSession session)
         else
             GetAllEntityStates(session);
 
-        _playerManager.GetPlayerStates(session.FromTick, session.PlayerStates);
+        // Replay recordings get all playerstates to assist with admin investigations.
+        // Note that this may also reveal underlying session UIDs and names upon viewing the replay.
+        if (session.Session == null)
+            _playerManager.GetPlayerStates(session.FromTick, session.PlayerStates);
+        else
+            _playerManager.GetPlayerState(session.FromTick, session.Session, session.PlayerStates);
```
ElectroJr: Why not just combine the methods into one and have a null session imply its a replay.

Robust.Shared/Player/SharedPlayerManager.State.cs
```diff
@@ -28,12 +28,36 @@ public void GetPlayerStates(GameTick fromTick, List<SessionState> states)
...
+            return;
+
+        // TODO PlayerManager delta states
+        // Track last update tick/time per session, and only send sessions that actually changed.
+
+        states.EnsureCapacity(1);
+        foreach (var player in InternalSessions.Values)
+        {
+            if (session.UserId == player.UserId)
+            {
```
ElectroJr: cvar

ElectroJr (changes requested):

Robust.Shared/Player/CommonSession.cs
```diff
@@ -19,6 +19,9 @@ internal sealed class CommonSession : ICommonSessionInternal
     [ViewVariables]
     public string Name { get; set; } = "<Unknown>";
 
+    [ViewVariables(VVAccess.ReadWrite)]
+    public string DisplayName { get; set; } = string.Empty;
```
PJB3005, line 23: A lack of display name should be null, not empty.

Robust.Shared/Player/SharedPlayerManager.Sessions.cs
```diff
@@ -246,6 +247,23 @@ public void SetName(ICommonSession session, string name)
...
+    public string GetDisplayName(ICommonSession session)
+    {
+        return session.DisplayName != string.Empty ? session.DisplayName : session.Name;
+    }
+
+    public string GetTruthfulNameString(ICommonSession session)
+    {
+        return $"{session.Name}" +
+               $"{(session.DisplayName != string.Empty ? $" ({session.DisplayName})" : "")}";
+    }
```
PJB3005, line 265: This needs to be localization driven.

Robust.Shared/CVars.cs
```diff
@@ -405,6 +405,12 @@ protected CVars()
         public static readonly CVarDef<bool> NetHWId =
             CVarDef.Create("net.hwid", true, CVar.SERVERONLY);
 
+        /// <summary>
+        /// If true, the sessions of all players connected to the server will be shared with connected clients.
+        /// If false, a client will only receive session information about itself.
+        /// </summary>
+        public static readonly CVarDef<bool> NetShareAllClientSessions =
+            CVarDef.Create("net.share_all_client_sessions", false, CVar.REPLICATED | CVar.SERVER);
```
PJB3005, line 413: Since I'd really expect this to be a game choice rather than a server choice, it would probably be better if this wasn't a CVar.

PJB3005 (changes requested):

PR #5817, opened 2025-04-09
Optimization RSI preloading / atlas creation

beck-thompson:
When RSIs were preloaded they worked like this (big picture overview):
1.) Load all images and combine rsi files into mini atlases
2.) Create a new maxsize by maxsize image atlas
3.) Loop through all RSIs and load them onto the atlas. If the atlas gets filled, create a new one and lock in the current one.
4.) Create a new image that is only as big as necessary (If the final atlas didn't get filled up all the way)
5.) Fit the orignal final atlas into the new cropped one.

For ss14 specifically, there is only enough sprites to allocate about half of one max sized atlas so step 3 really never occurs. This means that we are allocating 1 maxed sized image and 1 ~ half max sized image. This isn't necessary so what I've done is basically just removed that first allocation!

My changes now make it work like this:
1.) Load all images and combine rsi files into mini atlases.
2.) Loop through all RSIs, pre determine what atlas they would go into and their offset inside of it. If we know an atlas will be filled up, allocate an image of the correct size (This is only done when we 100% know we will need the atlas). This also uses the a slightly modified [first-fit decreasing-height strip packing algorithm](https://en.wikipedia.org/w/index.php?title=Strip_packing_problem&oldid=1263496949#First-fit_decreasing-height_(FFDH)). More information about that below.
3.) Loop through all FFDH levels and actually put them in their respective atlas images.
4.) Finalize all atlases (Actually load the atlas into Clyde and give all rsis a reference to that atlas)

I did some brief testing on ss14 and got a minor but still "noticeable" speed boost to preloading textures (~11.5% or 0.3 seconds)! Texture loading has a lot of variance but its defiantly an improvement.

image
(Before and after times - I ran each 10 times.)

For the atlas itself, it's now being stored using a slightly modified FFDH algorithm that I create. The algorithm is basically exactly the same, the only difference is I use the empty spaces above placed squares as new "levels" that you can insert stuff into.

image
Those "Free" level spaces would be completely ignored in the standard FFDH algorithm, mine just makes them a normal level which is filled up like normal. 

Here is a compassion of SS14s atlas was filled:

(Before this PR `cropped utilization: 80.84%, fill percentage: 60.94%`)
Screenshot from 2025-04-15 18-30-11

(FFDH sorting only `cropped utilization: 89.22%, fill percentage: 55.21%`)
Screenshot from 2025-04-15 18-27-39

(FFDH sorting w/ my modifications `cropped utilization: 97.00%, fill percentage: 50.78%`)
Screenshot from 2025-04-15 18-25-24

So this new algorithm is able to reduce SS14s atlas size by ~10%. For other repositories like Goob, the savings are even greater (They needed to allocate two atlases, this reduces it to 1!). This new allegorist saves them ~1.3 seconds (Mostly due to not having to allocate massive images for the second atlas).

### DIFF
--- a/Robust.Client/ResourceManagement/ResourceCache.Preload.cs
+++ b/Robust.Client/ResourceManagement/ResourceCache.Preload.cs
@@ -195,7 +195,6 @@ namespace Robust.Client.ResourceManagement
             var deltaY = 0;
             Vector2i offset = default;
             int atlasRsiIndexStart = 0;
-            int atlasCount = 0;
             int filledPixels = 0;
             for (int i = 0; i < atlasList.Length; i++)
             {
@@ -223,7 +222,7 @@ namespace Robust.Client.ResourceManagement
                 }
 
                 deltaY = Math.Max(deltaY, rsi.AtlasSheet.Height);
-                rsiPositions[i] = (atlasCount, offset);
+                rsiPositions[i] = (finalMetaAtlases.Count, offset);
                 offset.X += rsi.AtlasSheet.Width;
                 filledPixels += rsi.AtlasSheet.Width * rsi.AtlasSheet.Height;
             }
@@ -308,7 +307,7 @@ namespace Robust.Client.ResourceManagement
             sawmill.Debug(
                 "Preloaded {CountLoaded} RSIs into {CountAtlas} Atlas(es?) ({CountNotAtlas} not atlassed, {CountErrored} errored) in {LoadTime}",
                 rsiList.Length,
-                atlasCount,
+                finalMetaAtlases.Count,
                 nonAtlasList.Length,
                 errors,
                 sw.Elapsed);
--- a/Robust.Client/ResourceManagement/ResourceCache.Preload.cs
+++ b/Robust.Client/ResourceManagement/ResourceCache.Preload.cs
@@ -18,10 +18,7 @@ using Robust.Shared.Log;
 using Robust.Shared.Maths;
 using Robust.Shared.Utility;
 using SixLabors.ImageSharp;
-using SixLabors.ImageSharp.Formats.Png;
 using SixLabors.ImageSharp.PixelFormats;
-using SixLabors.ImageSharp.Processing;
-using Vector3 = System.Numerics.Vector3;
 
 namespace Robust.Client.ResourceManagement
 {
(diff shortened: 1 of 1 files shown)


Robust.Client/ResourceManagement/ResourceCache.Preload.cs
```diff
@@ -180,61 +181,88 @@ private void PreloadRsis(ISawmill sawmill)
 
             // Each RSI sub atlas has a different size.
             // Even if we iterate through them once to estimate total area, I have NFI how to sanely estimate an optimal square-texture size.
-            // So fuck it, just default to letting it be as large as it needs to and crop it as needed?
+            // So fuck it, just default to adding them in from smallest to largest.
             var maxSize = Math.Min(GL.GetInteger(GetPName.MaxTextureSize), _configurationManager.GetCVar(CVars.ResRSIAtlasSize));
-            var sheet = new Image<Rgba32>(maxSize, maxSize);
 
+            // (final meta atlas index, offset)
+            var rsiPositions = new (int, Vector2i)[atlasList.Length];
```
metalgearsloth: Couldn't you stackalloc a span for this or is it potentially too large.

Robust.Client/ResourceManagement/ResourceCache.Preload.cs
```diff
@@ -180,61 +181,88 @@ private void PreloadRsis(ISawmill sawmill)
...
+
+            // (start rsi index, end rsi index, final meta atlas reference).
+            var finalMetaAtlases = new List<(int, int, Image<Rgba32>)>();
```
metalgearsloth: ValueList

Robust.Client/ResourceManagement/ResourceCache.Preload.cs
beck-thompson: For reference, the length is `2812` currently!

So I (just) tested both that and this new one:
`Span<(int, Vector2i)> rsiPositions = stackalloc (int, Vector2i)[atlasList.Length];`

From when I tested it, it *seemed* to run somewhat slower using stackalloc - I really don't know why that would be the case but I ran it a bunch and it seemed slightly slower. Also, if you want to do some other testing, run this with goobstations repo, they have some crazy atlases and it'll test having more than one atlas.

PJB3005: This really isn't something that should be stackalloc'd

beck-thompson:
Made a big update! For testing I'd recommend adding the following:

Line 300-304:
```
var random = new Random();
var color = new Rgba32((byte)random.Next(256), (byte)random.Next(256), (byte)random.Next(256));
var image = new Image<Rgba32>(rsi.AtlasSheet.Width, rsi.AtlasSheet.Height, color);

image.Blit(box, imageAtlases[level.AtlasId], rsi.AtlasOffset);
```

Line 308:
```
atlases[0].Save("atlasOut.png", new PngEncoder());
```

PR #5818, opened 2025-04-10
Entity spawn prediction

metalgearsloth:
Needs test + proxy methods to make it easier. Also might look at gun prediction to see what's needed there.

ATM I just got virtual hands fully predicted.


Robust.Client/GameObjects/ClientEntityManager.cs
```diff
@@ -291,5 +291,14 @@ public int Compare((uint seq, MsgEntity msg) x, (uint seq, MsgEntity msg) y)
...
+
+        public override void FlagPredicted(Entity<MetaDataComponent?> entity)
+        {
+            if (!MetaQuery.Resolve(entity.Owner, ref entity.Comp, false))
+                return;
+
+            DebugTools.Assert(IsClientSide(entity.Owner, entity.Comp));
+            EnsureComponent<PredictedSpawnComponent>(entity.Owner);
+        }
     }
```
metalgearsloth: Might end up redoing this so all spawns are predicted and anyone that wants the entity to persist (e.g. gun prediction) uses this.

metalgearsloth:
v2 will need a bunch of stuff when I feel like doing it.

PR #5819, opened 2025-04-11
Lower light defaults

VasilisThePikachu:
These look exactly the same, but reduce gpu usage by quite a bit from my testing. I believe this should be the default for future clients.

Needed https://github.com/space-wizards/space-station-14/pull/36471 since these defaults are  the "medium" preset

I cannot (myself) tell a difference between the previous and the proposed lighting default.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -47,7 +47,7 @@ END TEMPLATE-->
 
 ### Other
 
-*None yet*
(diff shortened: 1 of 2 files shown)


PJB3005 (changes requested):
Really not a fan of this approach to changing lighting settings. In truth, these properties shouldn't even be in RT's cvars file at all, they should instead be configured via content in all cases (and configured there).

PJB3005:
Closing, see above.

PR #5821, opened 2025-04-12
Make BoundUserInterfaceMessageAttempt broadcast again

ElectroJr:
This effectively undoes the breaking change in #5811, though the event will still get raised directed at the UI entity.


metalgearsloth (approved):

deltanedas:
couldnt you have added broadcast: true to the original raise

metalgearsloth:
> couldnt you have added broadcast: true to the original raise

Was talked about on discord fam

PR #5827, opened 2025-04-13
Allow toolshed commands to use entity system dependencies

ElectroJr:
I thought they could already do this, but apparently not. 
If any commands need to be available before entity systems initialize, they can just remain non-toolshed commands.


RELEASE-NOTES.md
```diff
@@ -35,13 +35,14 @@ END TEMPLATE-->
 
 ### Breaking changes
 
-*None yet*
+* `ToolshedEnvironment` initialization  now requires that entity systems have already been initialized first.
```
```suggestion
* `ToolshedEnvironment` initialization now requires that entity systems have already been initialized first.
```

ElectroJr:
ill just merge this into another PR

PR #5835, opened 2025-04-14
Tilestack support (tile smoothing)

murphyneko:
This PR adds support to tilestacking - a component on a grid is able to store tiles under a certain tile. This is only used in non-standard cases - for example, in context of SS14 content, placing wood floor on planet grass (currently, it is only supposed to go on plating).

Moved to engine as per request by Sloth.


ScarKy0:
Holy shit

murphyneko:
todo change some methods to use ensurecomp for tilestackmapcomp instead

PJB3005:
@metalgearsloth I'm curious why you feel like this should be in engine? This feels like it has a high risk to restrict content arbitrarily if they need to extend it in the future. For example if uncovering a tile should also do something more than just change the tile type.

murphyneko:
some of the reasoning from Sloth i got, but this might not be everything

metalgearsloth:
> @metalgearsloth I'm curious why you feel like this should be in engine? This feels like it has a high risk to restrict content arbitrarily if they need to extend it in the future. For example if uncovering a tile should also do something more than just change the tile type.

I think the engine should be storing the data itself with a simple API to interact with it and let content deal with popping or the likes. The main issue atm is we now have to manually curate every instance of content setting tile data (e.g. explosions, prying, shoveling, rcd, whatever else) which is just bound to cause maintenance problems long-term.

Robust.Shared/GameObjects/Systems/SharedMapSystem.Tilestack.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// </summary>
+    private void OnTileChanged(ref TileChangedEvent ev)
+    {
+        DeleteTilestack(ev.ChunkIndex, ev.Entity.Owner, ev.Entity.Comp);
+    }
+
+    /// <summary>
+    ///     Gets a list of tiles under the current tile, returns true if succeeds.
+    /// </summary>
+    public bool TryTilestack(Vector2i gridIndices, EntityUid gridUid, out List<Tile>? tilestack)
```
metalgearsloth: Is this meant to be notnullwhen(true)

metalgearsloth (changes requested):
1. Needs chunking because a dictionary of tiles is incredibly wasteful, look at anchoring. It also really should be nullable tilestacks because most grids aren't going to care.
2. This is going to make planets much more expensive and you might need pooling.
3. Network it.
4. Needs deltas.
5. Please use EntityQuery<T> for performance-critical codepaths.
6. Needs accompanying tests.
7. You need to serialize this or you will break maps.

ScarKy0:
:despair:

NotSoDana:
why...

PR #5836, opened 2025-04-14
Adds a parameter that lets RayCast's also intersect non-hard fixtures.

MLGTASTICa:
Adds a parameter that lets people select if a RayCast should count non-hard fixture collisions as a hit and thus return them.
Very usefull in cases where people have detector zones & want those detector zones to work with raycasting.
False by default. To not modify any expected behaviour.


MLGTASTICa:
Defeated by git..

PR #5837, opened 2025-04-14
Adds an argument to permit Raycasting to collide with non-hard fixtures.

MLGTASTICa:
Adds a parameter so that RayCasting can be used for checking non-hard fixtures too. Would be very useful for systems that employ non-hard fixtures and also employ raycasting.


metalgearsloth:
The new system has a proxy callback and should be used instead.

MLGTASTICa:
> The new system has a proxy callback and should be used instead.

The new system of what?

PR #5843, opened 2025-04-16
Add SnapToGrid to engine

metalgearsloth:
Title just some content stuff also a nice trygettileref method because doing the manual grid checks is annoying.


slarticodefast:
Can you move the other methods in `EntityCoordinatesExtensions` to the engine as well?
There is no reason for them to be in content.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Coordinates.cs
```diff
@@ -12,6 +14,69 @@ public abstract partial class SharedTransformSystem
      * For grid methods see SharedMapSystem.Coordinates
      */
 
+    /// <summary>
+    /// Tries to snap these grids to coordinate if possible.
```
slarticodefast: What does this even mean, could you make the sentence a little more clear?
Also add a summary to the overload below. It helps to have a tooltip in the IDE when you hover your mouse over the function.

Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
```diff
@@ -1527,6 +1528,23 @@ public MapCoordinates GridTileToWorld(EntityUid uid, MapGridComponent grid, Vect
...
+
+        if (!TryComp(grid, out MapGridComponent? mapGrid))
+        {
+            tileRef = TileRef.Zero;
+            return false;
+        }
+
+        return TryGetTileRef(grid.Value, mapGrid, coordinates, out tileRef);
+    }
+
```
slarticodefast, line 1562: add the summary here and to the other two overloads as well
badly documented engine code is often very problematic for newer content contributors

Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
```diff
@@ -1527,6 +1528,23 @@ public MapCoordinates GridTileToWorld(EntityUid uid, MapGridComponent grid, Vect
...
 
+    /// <summary>
+    /// Tries to resolves coordinates to a corresponding tileRef. Consider using the other methods if you already have the grid.
+    /// </summary>
+    [Pure]
+    public bool TryGetTileRef(EntityCoordinates coordinates, out TileRef tileRef)
+    {
+        var grid = _transform.GetGrid(coordinates);
+
+        if (!TryComp(grid, out MapGridComponent? mapGrid))
```
slarticodefast: `_gridQuery` should be available from the main file

slarticodefast (approved):
Thanks!

Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
slarticodefast: Oh, wait, this one can still be changed to use the query we have available

slarticodefast:
Oh, wait. The query comment is still unresolved. Minor optimization though.

PR #5849, opened 2025-04-17
RandomPredicted extension

MilonPL:
adds semi-random methods for IRobustRandom that can be safely used in Shared without the risk of misspredicts
shouldn't be used for anything sensitive that you wouldn't want the client to guess, but we already use this solution for dice anyways
cloning this code every time is bad so instead of
```cs
        var rand = new System.Random((int)_timing.CurTick.Value);

        var roll = rand.Next(1, entity.Comp.Sides + 1);
        SetCurrentSide(entity, roll);
```

we do
```cs
        var roll = _random.NextPredicted(_timing, 1, entity.Comp.Sides + 1);
        SetCurrentSide(entity, roll);
```

### DIFF
--- a/Robust.Shared/Random/RandomPredicted.cs
+++ b/Robust.Shared/Random/RandomPredicted.cs
@@ -1,0 +1,205 @@
+using System;
+using System.Collections.Generic;
+using Robust.Shared.Maths;
+using Robust.Shared.Timing;
+
+namespace Robust.Shared.Random;
+
+public static class RandomPredicted
+{
+    /// <summary>
+    /// Get a predictable Random instance seeded with the current tick.
+    /// This ensures identical random sequences on both client and server.
+    /// </summary>
+    /// <param name="random">The <see cref="IRobustRandom"/> instance.</param>
+    /// <param name="timing">The <see cref="IGameTiming"/> to use for seeding.</param>
+    /// <remarks>
+    /// NOTE: This means the client can guess what the number will be based on the current tick,
+    /// do NOT use this for sensitive applications.
+    /// </remarks>
+    public static System.Random GetPredictedRandom(this IRobustRandom random, IGameTiming timing)
+    {
+        return new System.Random((int)timing.CurTick.Value);
+    }
+
+    /// <summary>
+    /// Get predictable random <see cref="float"/> value between 0 (included) and 1 (excluded).
+    /// </summary>
+    /// <param name="random">The <see cref="IRobustRandom"/> instance.</param>
+    /// <param name="timing">The <see cref="IGameTiming"/> to use for seeding.</param>
+    /// <inheritdoc cref="GetPredictedRandom" path="/remarks"/>
+    public static float NextFloatPredicted(this IRobustRandom random, IGameTiming timing)
+    {
+        return random.GetPredictedRandom(timing).NextFloat();
+    }
+
+    /// <summary>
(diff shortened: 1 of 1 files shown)


EmoGarbage404 (changes requested):
should have some kind of seed/offset parameter that gets added to the curtick so that multiple things being done at once don't have the same outcome.

slarticodefast:
Yeah, see predicted dice which also use the net entity uid

metalgearsloth (commented):
1. Might be useful to be able to pass in the netentity for seed as otherwise every single prng result will be the exact same (e.g. if you shoot 20 guns on the same tick they all have the same accuracy).
2. eww classes, struct would be nicer.

MilonPL:
> 1. Might be useful to be able to pass in the netentity for seed as otherwise every single prng result will be the exact same (e.g. if you shoot 20 guns on the same tick they all have the same accuracy).

It already has optional seed, you can just do EntityUid.Id or whatever 
It's a lot of methods so I didn't want to make extra overloads for entities

Robust.Shared/Random/RandomPredicted.cs
```diff
@@ -0,0 +1,242 @@
...
+    /// <remarks>
+    /// NOTE: This means the client can guess what the number will be based on the current tick,
+    /// do NOT use this for sensitive applications.
+    /// </remarks>
+    [PublicAPI]
+    public static System.Random GetPredictedRandom(this IRobustRandom random, IGameTiming timing, int seed = 0)
+    {
+        var tickValue = (int) timing.CurTick.Value;
+
+        var combinedSeed = HashCode.Combine(tickValue, seed);
```
slarticodefast, line 27: So, I tried using this and it turns out that `HashCode.Combine` is intitialized with a random seed, meaning that client and server will create different results.
grafik

grafik

So this needs a deterministic hash function instead.

slarticodefast (changes requested):

perryprog:
Doesn't need to be this PR, just writing it down so there's a chance I don't forget—this should also be used for SoundCollectionSpecifiers getting rolled so predicted playback of sound collections will give everyone the same file. Ditto for pitch variation, though that would need a predictable random guassian.

Robust.Shared/Random/RandomPredicted.cs
FairlySadPanda: @slarticodefast Is there any update on this determinism topic

slarticodefast: Do you mean specifically HashCode.Combine or this PR as a whole?
For the first one I added a deterministic hash function as a workaround in Content, see ClumsySystem as an example.
For random predicted we probably will need to rewrite RobustRandom as well to be struct-based. If you are interested I already looked up a few scientific papers with licences that are open-source friendly (which is surprisingly hard to find, everyone just uses the algorithms), but I didn't find time to implement it myself yet. And milon is on a break.

FairlySadPanda: OK. Can you make this a task? This is in my wheelhouse and I'd be happy to do so to unblock randomness on `Content.Shared`.

IProduceWidgets: HashCodeCombine method exists now for this specific usecase.
Its uh, in content though I think.

PR #5851, opened 2025-04-18
Add debug assert to entity audio stop

metalgearsloth:
Catches some cases of audio ent fields being networked (they shouldn't be).


metalgearsloth:
Also of note this will probably break MovementSound / trip asserts if merged as is which is why I held off, ideally that needs fixing first on ss14.

PR #5853, opened 2025-04-18
Allow subclasses of RichTextLabel to get a copy of the parsed message

ciaran:
Needed to allow subclasses to modify the displayed message (add a colour) without having to parse the message string again.


PJB3005 (changes requested):
Controls like this aren't actually meant to be inherited.

ciaran:
I see, I'll rewrite to wrap it instead then.

PR #5854, opened 2025-04-18
add RecursiveGetAllEntitiesWithComp

SolStar2:
added functionality in container system to get all entities with a component recursively. This will help in situations in which all entities with a certain component needs to be processed before a parent container does something. Such example as a polymorph dropping all MindContainerComponet entities before polymorphing so they are not sent to null-space.

### DIFF
--- a/Robust.Shared/Containers/SharedContainerSystem.cs
+++ b/Robust.Shared/Containers/SharedContainerSystem.cs
@@ -404,18 +404,18 @@ namespace Robust.Shared.Containers
         /// <summary>
         /// Recursively get all entities with Component T in all containers on the entity
         /// </summary>
-        /// <param name="containter">The root container to search in.</param>
+        /// <param name="root">The root container to search in.</param>
         /// <param name="maxDepth">The max depth that will be searched before aborting.</param>
         /// <typeparam name="T">The component being searched for.</typeparam>
         /// <returns>All entities with component T in container and all children</returns>
         public HashSet<Entity<T?>> RecursiveGetAllEntitiesWithComp<T>(
-            Entity<ContainerManagerComponent> containter,
+            Entity<ContainerManagerComponent> root,
             int maxDepth = 1000) where T : Component
         {
             var entities = new HashSet<Entity<T?>>();
 
             var stack = new Stack<EntityUid>();
-            stack.Push(containter);
+            stack.Push(root);
 
             while (stack.Count > 0)
             {
--- a/Robust.Shared/Containers/SharedContainerSystem.cs
+++ b/Robust.Shared/Containers/SharedContainerSystem.cs
@@ -402,46 +402,55 @@ namespace Robust.Shared.Containers
         }
 
         /// <summary>
-        /// Recursively get all entities with Component T in all containers on the entity
+        /// Recursively collects all entities with Component T in all containers on the entity.
         /// </summary>
-        /// <param name="root">The root container to search in.</param>
-        /// <param name="maxDepth">The max depth that will be searched before aborting.</param>
-        /// <typeparam name="T">The component being searched for.</typeparam>
-        /// <returns>All entities with component T in container and all children</returns>
-        public HashSet<Entity<T?>> RecursiveGetAllEntitiesWithComp<T>(
+        /// <typeparam name="T">The component to search for.</typeparam>
+        /// <param name="root">The root entity to begin the search from.</param>
+        /// <param name="matchingEntities">Optional pre-allocated collection to store results.</param>
+        /// <param name="stack">Optional pre-allocated stack for traversal.</param>
+        /// <returns>
+        /// A <see cref="HashSet{T}"/> containing all found entities with component <typeparamref name="T"/>.
+        /// If <paramref name="matchingEntities"/> was provided, returns the same instance with the added entities.
+        /// </returns>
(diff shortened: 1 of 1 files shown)


Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -401,6 +401,47 @@ public bool TryFindComponentsOnEntityContainerOrParent<T>(
...
+        /// <param name="root">The root container to search in.</param>
+        /// <param name="maxDepth">The max depth that will be searched before aborting.</param>
+        /// <typeparam name="T">The component being searched for.</typeparam>
+        /// <returns>All entities with component T in container and all children</returns>
+        public HashSet<Entity<T?>> RecursiveGetAllEntitiesWithComp<T>(
+            Entity<ContainerManagerComponent> root,
+            int maxDepth = 1000) where T : Component
+        {
+            var entities = new HashSet<Entity<T?>>();
+
```
metalgearsloth: Have hashset get passed in.

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -401,6 +401,47 @@ public bool TryFindComponentsOnEntityContainerOrParent<T>(
...
+            {
+                var currentUid = stack.Pop();
+
+                if (!HasComp<ContainerManagerComponent>(currentUid))
+                    continue;
+
+                foreach (var container in GetAllContainers(currentUid).ToList())
+                {
+                    foreach (var entity in container.ContainedEntities.ToList())
+                    {
```
metalgearsloth: These aren't being mutated while iterated so you can safely just iterate these, ToList entirely meaningless.

Also do you want it to ignore any non-direct containers?

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -401,6 +401,47 @@ public bool TryFindComponentsOnEntityContainerOrParent<T>(
...
+                        if (HasComp<T>(entity))
+                        {
+                            entities.Add(entity);
+                            continue;
+                        }
+
+                        if (stack.Count < maxDepth) // Unlikely to have over 1000 nested entities unless there is an infinite loop.
+                            stack.Push(entity); // Process this entity's containers next
```
metalgearsloth: If the depth ever reached this far the game would be unplayably slow so just remove it.

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -401,6 +401,47 @@ public bool TryFindComponentsOnEntityContainerOrParent<T>(
...
+            var stack = new Stack<EntityUid>();
+            stack.Push(root);
+
+            while (stack.Count > 0)
+            {
+                var currentUid = stack.Pop();
+
+                if (!HasComp<ContainerManagerComponent>(currentUid))
+                    continue;
+
```
metalgearsloth: Please use EntityQuery<T> for performance critical code.

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -401,6 +401,47 @@ public bool TryFindComponentsOnEntityContainerOrParent<T>(
...
+        /// <returns>All entities with component T in container and all children</returns>
+        public HashSet<Entity<T?>> RecursiveGetAllEntitiesWithComp<T>(
+            Entity<ContainerManagerComponent> root,
+            int maxDepth = 1000) where T : Component
+        {
+            var entities = new HashSet<Entity<T?>>();
+
+            var stack = new Stack<EntityUid>();
+            stack.Push(root);
+
```
metalgearsloth: Same here use a field no point allocating.

metalgearsloth (changes requested):

Robust.Shared/Containers/SharedContainerSystem.cs
SolStar2: If thats the case should I refactor this to use recursion instead of iterative?

metalgearsloth: Whatever is fine.

Robust.Shared/Containers/SharedContainerSystem.cs
SolStar2: > Also do you want it to ignore any non-direct containers?

Not sure why you would since if you don't want to check recursively then you can do that fairly simply with existing methods.

Robust.Shared/Containers/SharedContainerSystem.cs
ewokswagger: Would it also be ok if I instead allocated them in the system?

Robust.Shared/Containers/SharedContainerSystem.cs
metalgearsloth: So there's no guarantee an entity in a direct container is in a container anywhere up its hierarchy so for example bagged entities can get skipped for example, or organs, or whatever else.

At the moment this relies exclusively upon how content has decided to do things but I'm not sure if you actually want this over a recursive check.

SolStar2: It would be nice to check non-direct containers as well if that wouldn't be a too drastic change. However, I am not quite sure how to change it to do that.

SolStar2:
Alright figured it out

PR #5861, opened 2025-04-19
Makes ItemList not run deselection callback on all list items

Centronias:
even when they weren't selected.
I have no idea if this is the best way to do this.

Fixes a bug introduced [here](https://github.com/space-wizards/RobustToolbox/pull/5796/files#diff-c145b320f707ad74c95d4df3254cdc2703416c65b2bc8ea0a6987bdb9b78cfddL293) by iterating over all list items rather than just selected list items.
image
(screenshot of the diff in case the link to a line in a PR doesn't work)

### DIFF
--- a/Robust.Client/UserInterface/Controls/ItemList.cs
+++ b/Robust.Client/UserInterface/Controls/ItemList.cs
@@ -344,9 +344,6 @@ namespace Robust.Client.UserInterface.Controls
                     continue;
 
                 var item = _itemList[i];
-                if (!item.Selected)
-                    continue;
-
                 item.Selected = false;
             }
         }
@@ -744,7 +741,7 @@ namespace Robust.Client.UserInterface.Controls
                 get => _selected;
                 set
                 {
-                    if (!Selectable) return;
+                    if (!Selectable || _selected == value) return;
                     _selected = value;
                     if(_selected) OnSelected?.Invoke(this);
                     else OnDeselected?.Invoke(this);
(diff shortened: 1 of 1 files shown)


Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -340,7 +340,7 @@ public void ClearSelected(int? except = null)
         {
             for (var i = 0; i < _itemList.Count; i++)
             {
-                if (i == except)
+                if (i == except || !_itemList[i].Selected)
                     continue;
 
                 var item = _itemList[i];
```
metalgearsloth: Make the array access it once via _itemList[i] and merg.

metalgearsloth (changes requested):

Robust.Client/UserInterface/Controls/ItemList.cs
```diff
@@ -340,10 +340,10 @@ public void ClearSelected(int? except = null)
         {
             for (var i = 0; i < _itemList.Count; i++)
             {
-                if (i == except)
+                var item = _itemList[i];
+                if (i == except || !item.Selected)
                     continue;
```
metalgearsloth: You should move the array access after checking the index, however I have nfi what it compiles to and too lazy to check but I'm going to assume it's dumb.

Centronias:
@metalgearsloth , is the anything holding this up?

PJB3005 (changes requested):
You should be doing a check in `Item.Selected`'s setter instead.

Centronias:
Lemme know if that's not what you mean, @PJB3005 🫡

PR #5865, opened 2025-04-19
Serialize TimeSpan from text

Errant-4:
TimeSpanSerializer can now convert strings from the compatible format
The string must start with a number and end with a single letter referring to the time unit used.
It can NOT combine multiple types (like "1h30m"), but it CAN use decimals ("1.5h").

This makes it possible to use more easily readable values on time-related datafields, most prominently role time requirements. It also continues to be compatible with the existing format of digits only, interpreted as seconds.
Screenshot_2025-05-04_090614

(For optimal reviewing experience, check the two commits separately. The first one is only formatting)

### DIFF
--- a/Robust.Shared/Serialization/TypeSerializers/Implementations/TimespanSerializer.cs
+++ b/Robust.Shared/Serialization/TypeSerializers/Implementations/TimespanSerializer.cs
@@ -89,7 +89,7 @@ public sealed class TimespanSerializer : ITypeSerializer<TimeSpan, ValueDataNode
             return false;
 
         // Check the last character of the input for time unit indicators
-        switch (node.Value.AsSpan()[^1])
+        switch (node.Value[^1])
         {
             case 's':
                 timeSpan = TimeSpan.FromSeconds(number);
--- a/Robust.Shared/Serialization/TypeSerializers/Implementations/TimespanSerializer.cs
+++ b/Robust.Shared/Serialization/TypeSerializers/Implementations/TimespanSerializer.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using JetBrains.Annotations;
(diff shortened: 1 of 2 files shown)


Robust.Shared/Serialization/TypeSerializers/Implementations/TimespanSerializer.cs
```diff
@@ -9,42 +11,94 @@
...
+    /// </summary>
+    private bool TryTimeSpan(ValueDataNode node, [NotNullWhen(true)] out TimeSpan? timeSpan)
+    {
+        timeSpan = null;
+        var units = new List<string>{"h", "m", "s"};
+
+        var input = node.Value.Replace(" ","");
+        var unit = input[^1].ToString();
+        var number = input.Substring(0, input.Length - 1);
+        var valid = false;
```
ElectroJr: This probably gets called a lot, so you should rewrite this to avoid allocating (i.e., no List new() or ToString()).

I.e., something like this (but maybe reorganized to avoid the TryParse duplication somehow)
```cs
if (input.Length <= 1)
    return false;

var number = input.AsSpan()[..^1];
switch (input[^1])
{
    case 's':
        if (!double.TryParse(number, out var d))
            return false;
        timeSpan = TimeSpan.FromSeconds(d);
        return true;
    case 'm':
        if (!double.TryParse(number, out d))
            return false;
        timeSpan = TimeSpan.FromMinutes(d);
        return true;
    case 'h':
        if (!double.TryParse(number, out d))
            return false;
        timeSpan = TimeSpan.FromHours(d);
        return true;
    default:
        return false;
}
```

I'm also not 100% if its meant to be `^1` or `^2` for the slice

ElectroJr: And this should probably have unit tests to make sure the parsing works as expected. its pretty easy to add, you can more or less just duplicate `DateTimeSerializerTest`

Errant-4: I made the change, but still working on the unit tests

Robust.Shared/Serialization/TypeSerializers/Implementations/TimespanSerializer.cs
```diff
@@ -9,42 +10,98 @@
...
+        // If there aren't even enough characters for a number and a time unit, exit
+        if (node.Value.Length <= 1)
+            return false;
+
+        // If the input without the last character is still not a valid number, exit
+        if (!double.TryParse(node.Value.AsSpan()[..^1], out var number))
+            return false;
+
+        // Check the last character of the input for time unit indicators
+        switch (node.Value.AsSpan()[^1])
```
PJB3005: You don't need to `.AsSpan()` this since you're just indexing to get a single char out.
```suggestion
        switch (node.Value[^1])
```

Errant-4:
~~Would it be okay to now go and en-masse change all role times requirements to the more readable format? If we would do it gradually as they get changed, we would be left with a mixed format being in use basically foreve~~
Topic beyond the scope of this PR

PR #5866, opened 2025-04-19
replace IsMap and IsGrid calls with HasComp

VerinSenpai:
As the title says, replaces all obsolete IsMap and IsGrid calls with their respective HasComp counterparts. I believe this is pretty straight forward and shouldn't break anything.


metalgearsloth (changes requested):
These are all in expensive codepaths and really, really should be using entityqueries but I'll just do it in a separate pr.

VerinSenpai:
> These are all in expensive codepaths and really, really should be using entityqueries but I'll just do it in a separate pr.

I'll keep an eye out for that, I'm thinking thats what slarti wanted me to use in my other PR.

PR #5867, opened 2025-04-19
Replace Logger.Error calls with sawmill in SpriteComponent

ljm862:
Simply replaces all `Logger.Error` and `Logger.ErrorS` calls with `sawmill.Error` instead.
Due to DI failing on `SpriteComponent` creation as it's on a different thread, I tried many different things and decided to instead use the `LogManager.RootSawmill` instead of trying to use a different one.
If someone that is more familiar with the IoC process has a better solution, please let me know!

Also removes 2 unused Dependencies in the `SpriteComponent`.

Depending on how #5602 goes, it's possible this will all need to be redone on that. 

Part of https://github.com/space-wizards/space-station-14/issues/33279 and removes about 30 warnings


PJB3005 (changes requested):
You are completely getting rid of the log categories in many cases. This is not a proper fix.

ElectroJr:
> Due to DI failing on SpriteComponent creation as it's on a different thread, I tried many different things and decided to instead use the LogManager.RootSawmill instead of trying to use a different one.

I know sprite comp is a PITA to work with due to initialization spaghetti and it being one of the few components that use `ISerializationHooks`, but presumably you could've tried to keep the logger/category by doing something like
```cs
_sawmill ??= logMan?.GetSawmill(LogCategory)
_sawmill?.Error()
```
Though that having been said:

> Depending on how #5602 goes, it's possible this will all need to be redone on that.

#5602 already remove a large chunk of the obsolete static `Logger` calls, and the others will get removed whenever the rest of the code gets moved over to the system. So I'd actually prefer avoiding introducing conflicts or adding more sprite component initialization spaghetti to get a logger. 

The help solving https://github.com/space-wizards/space-station-14/issues/33279 is appreciated, but maybe its better to mostly leave sprite comp alone for now.

ljm862:
> You are completely getting rid of the log categories in many cases. This is not a proper fix.

Yes, I had solutions where I kept the LogCategory but using it would fail on instantiation and cause a lot of test failures :( 
 
I was hoping someone would be able to suggest a proper solution to keep them! 

> I know sprite comp is a PITA to work with due to initialization spaghetti and it being one of the few components that use `ISerializationHooks`, but presumably you could've tried to keep the logger/category by doing something like
> 
> ```cs
> _sawmill ??= logMan?.GetSawmill(LogCategory)
> _sawmill?.Error()
> ```

I tried that out too unfortunately but it was failing the tests... 


> but maybe its better to mostly leave sprite comp alone for now.

I'm inclined to agree here. I first thought it would be a simple quick change but then noticed that it wasn't 😅 

Thanks!

PR #5868, opened 2025-04-20
Sharpening shader

B-Kirill:
Adds a sharpness shader based on the `Adaptive Unsharp Masking` algorithm. 
It was previously implemented via an overlay in [(36481)](https://github.com/space-wizards/space-station-14/pull/36481) PR, but now it has been directly transferred to `Viewport` render. Now that PR just adds a slider in the graphics settings to set the sharpness value.

It wasn't easy for me to figure out the rendering, so if I did something wrong or you have any advice, feedback is welcome.

### DIFF
--- a/Resources/Shaders/sharpening.swsl
+++ b/Resources/Shaders/sharpening.swsl
@@ -1,0 +1,26 @@
+uniform sampler2D SCREEN_TEXTURE;
+uniform mediump float Sharpness;
+uniform highp vec2 viewportSize;
+
+void fragment()
+{
+    highp vec2 uv = UV;
+    
+    highp vec2 texelSize = vec2(1.0) / viewportSize;
+    
+    highp vec3 center = texture(SCREEN_TEXTURE, uv).rgb;
+    highp vec3 up = texture(SCREEN_TEXTURE, uv + vec2(0.0, texelSize.y)).rgb;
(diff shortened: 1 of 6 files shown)


PJB3005:
Why does this need to be in engine?

B-Kirill:
> Why does this need to be in engine?

But why not add it to engine? I think it would be a good feature not only for SS14. 
Moreover, if not in the engine, then how in general, if the render is located exactly there.

PJB3005:
> if not in the engine, then how in general

Can't you just apply the shader when the viewport is copied to the main screen? That's literally handled by content.

> But why not add it to engine? I think it would be a good feature not only for SS14.

Because these features are often extremely opinionated and baking it into the engine will often just mean that they don't expose enough creative liberty for games.

B-Kirill:
> Can't you just apply the shader when the viewport is copied to the main screen? That's literally handled by content.

It's pretty logical... Okay, I'll get to that later

B-Kirill:
> Can't you just apply the shader when the viewport is copied to the main screen? That's literally handled by content.

Uh.. Can I have a little clue where it is? The closest I've come is `ScalingViewport.cs`, but I'm still not completely sure if that's what I need.

PJB3005:
https://github.com/space-wizards/space-station-14/blob/9430a0f83529403404c6dbb8462afe55922100f2/Content.Client/Viewport/ScalingViewport.cs#L173

B-Kirill:
Again moving to SS14

PR #5871, opened 2025-04-20
Use FrozenDict for style properties

metalgearsloth:
TryGetStyle gets called about 470 times a frame on separated hud, about 100 less on normal hud, and style updates typically don't run at all on a tick depending what's happening, so I think generally this will be faster?

### DIFF
--- a/Robust.Client/UserInterface/Control.Styling.cs
+++ b/Robust.Client/UserInterface/Control.Styling.cs
@@ -1,6 +1,9 @@
 ﻿using System.Collections;
+using System.Collections.Frozen;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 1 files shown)


PJB3005 (changes requested):
Is this based on actual performance profiles?

metalgearsloth:
Forgot to turn angle off but will get again.

PJB3005:
Style properties are intended to dynamically update, so I'm not sure a frozen dictionary makes sense. Furthermore, if the lookup cost is actually problematic, I've had ideas for how to optimize the storage to avoid that entirely.

metalgearsloth:
> I've had ideas for how to optimize the storage to avoid that entirely.

Do that then goober, this PR isn't worth it enough lol.

PR #5873, opened 2025-04-20
PredictedDeleteEntity fixes

ElectroJr:
Based on the "3 scenarios" comment in `PredictedDeleteEntity` I assume theres currently a bug and the method should delete "pure" client-side entities instead of moving them into nullspace. So this replaces a `HasComponent<PredictedSpawnComponent>(ent.Owner)` with a `IsClientSide()` check in `PredictedDeleteEntity`. Also removes some unnecessary metadata comp fetching.

### DIFF
--- a/Robust.Client/GameObjects/ClientEntityManager.Spawn.cs
+++ b/Robust.Client/GameObjects/ClientEntityManager.Spawn.cs
@@ -118,9 +118,6 @@ public sealed partial class ClientEntityManager
         DebugTools.Assert(IsClientSide(ent.Owner, ent.Comp));
         EnsureComponent<PredictedSpawnComponent>(ent.Owner);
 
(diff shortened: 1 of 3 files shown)


Robust.Client/GameObjects/ClientEntityManager.Spawn.cs
```diff
@@ -118,9 +118,6 @@ public override void FlagPredicted(Entity<MetaDataComponent?> ent)
         DebugTools.Assert(IsClientSide(ent.Owner, ent.Comp));
         EnsureComponent<PredictedSpawnComponent>(ent.Owner);
 
-        // Server has no knowledge of the entity we are so we generate a clientside nentity and send it to server.
-        DebugTools.Assert(ent.Comp.NetEntity.IsClientSide());
```
ElectroJr, line 122: duplicate assert

PR #5876, opened 2025-04-21
Minor respath improvements

ElectroJr:
Removes some allocations and slightly speeds up some methods. Probably pretty insignificant but it gives me dopamine nonetheless.

Also adds a helper function for checking whether a string satisfies the DebugAssert in the ResPath constructor.


Robust.Shared/Utility/ResPath.cs
```diff
@@ -504,6 +518,16 @@ public string ChangeSeparator(string newSeparator)
             ? CanonPath
             : CanonPath.Replace("/", newSeparator);
     }
+
+    /// <inheritdoc cref="ChangeSeparator(string)"/>
+    public string ChangeSeparator(char newSeparator)
```
PJB3005, line 518: Honestly, I feel like you can just make the previous overload call this and throw if there's more than 1 char. There's no reasonable use for that anyways.

ElectroJr: Tests apparently expect multi-char emoji separators to be supported..... im just gonna get rid of those tests
rider64_EIbvaPLHGo

PR #5883, opened 2025-04-23
Fix player-specific replay sounds playing globally in replays

PJB3005:


Robust.Shared/Audio/Systems/SharedAudioSystem.cs
```diff
@@ -271,6 +274,36 @@ private void OnAudioGetStateAttempt(EntityUid uid, AudioComponent component, ref
...
     }
 
+    private static void OnAudioPlayerFilterGetStateAttempt(
+        EntityUid uid,
+        AudioPlayerFilterComponent component,
+        ref ComponentGetStateAttemptEvent args)
+    {
+        // Only allow replay recording to capture this component.
+        // (or anybody if state attempts can be bypassed if PVS is disabled)
+        args.Cancelled |= args.Player == null;
```
ElectroJr, line 284:
```suggestion
        args.Cancelled |= args.Player != null;
```

The component also needs to override the `Component.SessionSpecific` bool, otherwise the event never gets raised.

And this is probably somewhat beyond the scope of this PR, but if replay-only comp states become more common, we should probably add a separate check in `PvsSystem.GetEntityState` that doesn't bother raise an event and just checks if `player==null`.  And that should probably be done by replacing the current bools with enums or bitflags set via an attribute & cached in `MetaData.NetComponents` instead of overriding virtual bool getters in the base component class.

ElectroJr (changes requested):

Robust.Shared/GameObjects/EntityManager.Network.cs
```diff
@@ -593,6 +593,28 @@ public NetEntity[] GetNetEntityArray(EntityUid[] entities)
         return netEntities;
     }
 
+    public NetEntity[] GetNetEntityArray(IReadOnlyCollection<EntityUid> entities)
+    {
+        var netEntities = new NetEntity[entities.Count];
+        var i = 0;
+        foreach (var ent in entities)
+        {
+            netEntities[i++] = GetNetEntity(ent);
```
ElectroJr, line 602: The content PR is failing tests because this logs an error when one of the entities has been deleted (can't get metadata comp). Which it should do in general, but in this case it doesn't really matter. So I guess there should be a `TryGetNetEntityArray` that uses `TryGetNetEntity`?

PR #5891, opened 2025-04-26
Cleanup warnings: CS0649

B-Kirill:
Remove 2x `CS0649` warnings

[CS0649](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0649): Field 'field' is never assigned to, and will always have its default value 'value'.

### DIFF
--- a/Robust.Client/Graphics/Clyde/Windowing/Glfw.cs
+++ b/Robust.Client/Graphics/Clyde/Windowing/Glfw.cs
@@ -23,7 +23,9 @@ namespace Robust.Client.Graphics.Clyde
             private readonly ISawmill _sawmillGlfw;
 
             private bool _glfwInitialized;
-            private bool _win32Experience = false;
+#pragma warning disable CS0649
+            private bool _win32Experience;
+#pragma warning restore CS0649
 
             public GlfwWindowingImpl(Clyde clyde, IDependencyCollection deps)
             {
--- a/Robust.Client/Graphics/Clyde/Windowing/Glfw.cs
+++ b/Robust.Client/Graphics/Clyde/Windowing/Glfw.cs
@@ -23,7 +23,7 @@ namespace Robust.Client.Graphics.Clyde
             private readonly ISawmill _sawmillGlfw;
 
             private bool _glfwInitialized;
-            private bool _win32Experience;
+            private bool _win32Experience = false;
 
             public GlfwWindowingImpl(Clyde clyde, IDependencyCollection deps)
             {
--- a/Robust.Client/Graphics/Clyde/Windowing/Glfw.cs
+++ b/Robust.Client/Graphics/Clyde/Windowing/Glfw.cs
@@ -23,9 +23,9 @@ namespace Robust.Client.Graphics.Clyde
             private readonly ISawmill _sawmillGlfw;
 
             private bool _glfwInitialized;
-#pragma warning disable CS0649
+#if DEBUG
             private bool _win32Experience;
-#pragma warning restore CS0649
+#endif
 
             public GlfwWindowingImpl(Clyde clyde, IDependencyCollection deps)
             {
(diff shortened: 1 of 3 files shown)


Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs
```diff
@@ -625,6 +625,11 @@ private record struct UpdateTreesJob : IRobustJob
     {
         public IEntityManager EntManager;
 
+        public UpdateTreesJob()
```
ElectroJr: Unless rider is lying to me, it seems like the whole struct is unused?

B-Kirill: > Unless rider is lying to me, it seems like the whole struct is unused?

You mean `public UpdateTreeJob()`? If yes - this is necessary here. In short, without it, the compiler will give an error when trying to assign "= default!", and I would not like to change something more globally than this.

And if you're talking about the struct, then I didn't understand how it used either.

PJB3005: They are telling you to delete the struct entirely since it's unused.

B-Kirill: Ah, got it

Robust.Client/Graphics/Clyde/Windowing/Glfw.cs
```diff
@@ -23,7 +23,7 @@ private sealed partial class GlfwWindowingImpl : IWindowingImpl
             private readonly ISawmill _sawmillGlfw;
 
             private bool _glfwInitialized;
-            private bool _win32Experience;
```
PJB3005, line 26: Revert this change.

B-Kirill: > Revert this change.

Is it better to use #pragma then?
(Okay, that's an obvious question)

Robust.Client/Graphics/Clyde/Windowing/Glfw.cs
```diff
@@ -23,7 +23,9 @@ private sealed partial class GlfwWindowingImpl : IWindowingImpl
             private readonly ISawmill _sawmillGlfw;
 
             private bool _glfwInitialized;
+#pragma warning disable CS0649
             private bool _win32Experience;
```
PJB3005, line 27: Put the logic behind `DEBUG` instead. Because that's what it's actually used for.

PJB3005 (changes requested):

Robust.Client/Graphics/Clyde/Windowing/Glfw.cs
B-Kirill: Okay

ElectroJr: The usage in `WinThreadWinCursorSet` also needs to be wrapped in an if block then.

And I guess this feedback somewhat conflicts with my feedback on #5892, but IMO the `[Dependency]` fields should still use pragmas, ~~but maybe double check with pjb~~.

edit: yep just use pragmas for the dependencies. ([discussed on discord](https://discord.com/channels/310555209753690112/770682801607278632/1365842378237149284))

B-Kirill:
`RobustToolbox/Robust.Client/Graphics/Clyde/Windowing/Glfw.Cursors.cs(90,21): error CS0103: The name '_win32Experience' does not exist in the current context [/home/runner/work/RobustToolbox/RobustToolbox/RobustToolbox/Robust.Client/Robust.Client.csproj]`

I guess it's still better to use #pragma here. (Or use #if DEBUG for that part of the code as well)

ElectroJr:
Just use a debug block for the other part of the code

PR #5892, opened 2025-04-26
Cleanup warnings: CS0414

B-Kirill:
Remove 8x `CS0414` warnings

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
@@ -14,7 +14,9 @@ namespace Robust.Client.GameObjects
         private EntityQuery<AnimationPlayerComponent> _playerQuery;
         private EntityQuery<MetaDataComponent> _metaQuery;
 
+#if DEBUG
         [Dependency] private readonly IComponentFactory _compFact = default!;
+#endif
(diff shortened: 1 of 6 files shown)


ElectroJr:
I'd rather find a way to just silence these warnings when not building in DEBUG & TOOLS. Or just disabling it with a `#pragma`. Putting the dependencies in the `#if` blocks just seems like its going to cause headaches later on. I can easily see myself creating a PR where I use `_timing` without realising its in an `#if` block. 

And I'm not even sure if the current github workflows test building with different configurations. I know content has the `Test Packaging` workflow that builds in release, but I'm not sure the engine workflows would catch this kind of issue, though that's a separate problem should probably be fixed.

B-Kirill:
> I'd rather find a way to just silence these warnings when not building in DEBUG & TOOLS. Or just disabling it with a `#pragma`. Putting the dependencies in the `#if` blocks just seems like its going to cause headaches later on. I can easily see myself creating a PR where I use `_timing` without realising its in an `#if` block.
> 
> And I'm not even sure if the current github workflows test building with different configurations. I know content has the `Test Packaging` workflow that builds in release, but I'm not sure the engine workflows would catch this kind of issue, though that's a separate problem should probably be fixed.

Ok, I'll redo it later using `#pragma`.

PR #5893, opened 2025-04-26
Add GridUidChangedEvent and MapUidChangedEvent

ElectroJr:
Makes transform system optionally raise events when setting the grid or map uid. IIRC we are currently not doing this to help improve performance when entities change maps/grids, seeing as it has to recursively iterate through all children to update these fields. 

This tries to mitigate that by only raising the event if a a new metadata flag is set. I.e., each entity has to opt-in to enable these events. This shouldn't noticeably impact map change perf, seeing as the metadata component already has to be fetched. But it will somewhat slow down griduid changes, as it does now also fetch the metadata comp of children. However, it's still going to be faster than just globally enabling the event. And having an event makes it much easier for content to implement some functionality that is dependent on the current grid (e.g., see [this discussion](https://discord.com/channels/310555209753690112/900426319433728030/1365494889797914624))

This PR also fixes an issue where occluders & lights weren't properly clearing & setting the `PvsPriority` metadata flag.

### DIFF
--- a/Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
+++ b/Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
@@ -6,7 +6,8 @@ namespace Robust.Shared.GameObjects;
 /// </summary>
 /// <remarks>
 /// Event handlers should not modify positions or delete the entity, because the move event that triggered this event
-/// is still being processed. This may also mean that positional entity queries are not currently reliable.
+/// is still being processed. This may also mean that the entity's current position/parent has not yet been updated,
+/// positional entity queries are not reliable.
 /// </remarks>
 [ByRefEvent]
 public readonly struct GridUidChangedEvent(Entity<TransformComponent, MetaDataComponent> ent, EntityUid? old)
--- a/Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
+++ b/Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
@@ -7,7 +7,7 @@ namespace Robust.Shared.GameObjects;
 /// <remarks>
 /// Event handlers should not modify positions or delete the entity, because the move event that triggered this event
 /// is still being processed. This may also mean that the entity's current position/parent has not yet been updated,
-/// positional entity queries are not reliable.
+/// and that positional entity queries are not reliable.
 /// </remarks>
 [ByRefEvent]
 public readonly struct GridUidChangedEvent(Entity<TransformComponent, MetaDataComponent> ent, EntityUid? old)
--- a/Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
+++ b/Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
@@ -1,0 +1,21 @@
+namespace Robust.Shared.GameObjects;
+
+/// <summary>
+/// Raised directed at an entity when <see cref="TransformComponent.GridUid"/> is modified.
+/// This event is only raised if the entity has the <see cref="MetaDataFlags.ExtraTransformEvents"/> flag set.
+/// </summary>
+/// <remarks>
+/// Event handlers should not modify positions or delete the entity, because the move event that triggered this event
+/// is still being processed. This may also mean that positional entity queries are not currently reliable.
+/// </remarks>
+[ByRefEvent]
+public readonly struct GridUidChangedEvent(Entity<TransformComponent, MetaDataComponent> ent, EntityUid? old)
+{
+    public readonly Entity<TransformComponent, MetaDataComponent> Entity = ent;
+    public readonly EntityUid? OldGrid = old;
+
+    public EntityUid? NewGrid => Entity.Comp1.GridUid;
+    public EntityUid Uid => Entity.Owner;
+    public TransformComponent Transform => Entity.Comp1;
+    public MetaDataComponent Meta => Entity.Comp2;
+}
--- a/Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
+++ b/Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
@@ -10,11 +10,10 @@ namespace Robust.Shared.GameObjects;
 /// and that positional entity queries are not reliable.
 /// </remarks>
 [ByRefEvent]
-public readonly struct GridUidChangedEvent(Entity<TransformComponent, MetaDataComponent> ent, EntityUid? old)
+public readonly record struct GridUidChangedEvent(
+    Entity<TransformComponent, MetaDataComponent> Entity,
+    EntityUid? OldGrid)
 {
-    public readonly Entity<TransformComponent, MetaDataComponent> Entity = ent;
-    public readonly EntityUid? OldGrid = old;
-
     public EntityUid? NewGrid => Entity.Comp1.GridUid;
     public EntityUid Uid => Entity.Owner;
     public TransformComponent Transform => Entity.Comp1;
--- a/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
@@ -48,7 +48,7 @@ public abstract partial class SharedTransformSystem
         xform._localRotation += rotation;
 
         var meta = MetaData(uid);
-        SetGridId(new(uid, xform, meta), newGridUid);
+        SetGridId((uid, xform, meta), newGridUid);
         RaiseMoveEvent((uid, xform, meta), oldGridUid, oldPos, oldRot, oldMap);
 
         DebugTools.Assert(XformQuery.GetComponent(oldGridUid).MapID == XformQuery.GetComponent(newGridUid).MapID);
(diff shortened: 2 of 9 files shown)


Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
```diff
@@ -0,0 +1,22 @@
...
+public readonly struct GridUidChangedEvent(Entity<TransformComponent, MetaDataComponent> ent, EntityUid? old)
+{
+    public readonly Entity<TransformComponent, MetaDataComponent> Entity = ent;
+    public readonly EntityUid? OldGrid = old;
+
+    public EntityUid? NewGrid => Entity.Comp1.GridUid;
+    public EntityUid Uid => Entity.Owner;
+    public TransformComponent Transform => Entity.Comp1;
+    public MetaDataComponent Meta => Entity.Comp2;
+}
```
MilonPL: Would be nice if these fields had docstring as well, but most of it is self explanatory.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -355,31 +355,47 @@ private void OnCompStartup(EntityUid uid, TransformComponent xform, ComponentSta
...
-        DebugTools.Assert(gridUid == null || HasComp<MapGridComponent>(gridUid));
-        xform._gridUid = gridUid;
+        SetGridId((uid,  xform, MetaData(uid)), gridId);
     }
 
-    /// <summary>
-    /// Sets the <see cref="GridId"/> for the transformcomponent. Does not Dirty it.
-    /// </summary>
-    public void SetGridId(EntityUid uid, TransformComponent xform, EntityUid? gridId, EntityQuery<TransformComponent>? xformQuery = null)
+    public void SetGridId(Entity<TransformComponent, MetaDataComponent?> ent, EntityUid? gridId)
```
MilonPL, line 367: This needs proper docs.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -355,31 +355,47 @@ private void OnCompStartup(EntityUid uid, TransformComponent xform, ComponentSta
...
-    internal void SetGridIdNoRecursive(EntityUid uid, TransformComponent xform, EntityUid? gridUid)
+    public void SetGridId(EntityUid uid, TransformComponent xform, EntityUid? gridId, EntityQuery<TransformComponent>? xformQuery = null)
     {
-        DebugTools.Assert(gridUid == uid || !HasComp<MapGridComponent>(uid));
-        if (xform._gridUid == gridUid)
-            return;
-
-        DebugTools.Assert(gridUid == null || HasComp<MapGridComponent>(gridUid));
-        xform._gridUid = gridUid;
+        SetGridId((uid,  xform, MetaData(uid)), gridId);
```
MilonPL: Seems like we're never using the query? Why even pass it if we have it on the TransformSystem anyways? Also one extra space before xform.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -355,31 +355,47 @@ private void OnCompStartup(EntityUid uid, TransformComponent xform, ComponentSta
...
+        _metaQuery.ResolveInternal(ent.Owner, ref ent.Comp2);
+        if ((ent.Comp2!.Flags & MetaDataFlags.ExtraTransformEvents) != 0)
         {
-            SetGridId(child, XformQuery.GetComponent(child), gridId);
+#if DEBUG
+            var childCount = ent.Comp1.ChildCount;
+            var oldParent = ent.Comp1.ParentUid;
+#endif
+
+            var ev = new GridUidChangedEvent(ent!, ent.Comp1._gridUid);
```
MilonPL: Isn't `(ent, ent.Comp1, ent.Comp2)` a bit cleaner than null forgiving operators? 
Less prone to programming errors and explains what's happening a bit better.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -47,8 +47,8 @@ internal void ReAnchor(
         xform._localPosition = tilePos + newGrid.TileSizeHalfVector;
         xform._localRotation += rotation;
 
-        SetGridId(uid, xform, newGridUid, XformQuery);
         var meta = MetaData(uid);
+        SetGridId(new(uid, xform, meta), newGridUid);
```
MilonPL: Do we need to create a new Entity<T>? Why not just pass it in as `(uid, xform, meta)`?

Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
```diff
@@ -0,0 +1,22 @@
...
+/// Raised directed at an entity when <see cref="TransformComponent.GridUid"/> is modified.
+/// This event is only raised if the entity has the <see cref="MetaDataFlags.ExtraTransformEvents"/> flag set.
+/// </summary>
+/// <remarks>
+/// Event handlers should not modify positions or delete the entity, because the move event that triggered this event
+/// is still being processed. This may also mean that the entity's current position/parent has not yet been updated,
+/// and that positional entity queries are not reliable.
+/// </remarks>
+[ByRefEvent]
+public readonly struct GridUidChangedEvent(Entity<TransformComponent, MetaDataComponent> ent, EntityUid? old)
```
MilonPL: Couldn't these be made record structs? I'd say record structs are better and our docs say to use them for events when possible.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
ElectroJr: Its a holdover from back shortly after entity queries were introduced. Instead of storing them as fields in the system, the code used to just `GetEntityQuery<T>()` and pass the result into the methods. Nowadays meta & xform queries aren't even really needed cause EntityManager just has them.

Eventually the functions should just be removed / replaced, but it'd be a breaking change so its easier to just leave it for now.

MilonPL: It's only used twice in Content and both are in the same system, honestly I'd argue nuking it as a B/C in favor of the `Entity<T>` variant is fine. But your call.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
ElectroJr: I'll try do a proper benchmark to see if it makes a difference, but its largely precautionary in case using the implicit tuple cast operator is worse for performance, which it might be? Some goes for the null forgiving thing, though thats also useful if ever we use Arch, as then the ent struct might contain more data. I'd love it if the nullability analyzer were just smart enough to not need the null forgiving operator.

I.e., just playing around with c# lab, both the IL & jit asm seem very different with something like
```
public int Foo(int a, int b) {
    return Sum(new(a, b));
}

public int Bar(int a, int b) {
    return Sum((a, b));
}

public int Sum(Entity<int> f)
{
    return f.Comp + f.Owner;
}
```

Robust.Shared/GameObjects/EntitySystemMessages/GridUidChangedEvent.cs
ElectroJr: In this case they're so self explanatory I don't think it helps. E.g., the docstring for `EntityUid? OldGrid` would probably just be "The old grid's EntityUid".

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
MilonPL: Huh, yeah that makes sense, it does generate the extra ValueTuple ctor + op_Implicit call compared to direct construction. The overhead might not matter much since those are value-type operations on the stack. You can probably measure the gap in tight microbenchmarks, but it shouldn't be more than a few CPU instructions.

I do think we can sacrifice a bit of perf in non-critical code like this for better code readability (compared to null forgiving operators). Should we start using the constructor for all code though? Are there any downsides? I don't think `new(uid, comp)` is any different than `(uid, comp)` from a readability perspective.

ElectroJr: >  I don't think new(uid, comp) is any different than (uid, comp) from a readability perspective.

Yeah hence I was just using the constructor to be on the safe side, even if the compiler/jit does just end up reducing it down to the same expression. 

I created a [basic benchmark](https://github.com/ElectroJr/RobustToolbox/tree/entity-bench), and it seems like its  basically equivalent, so it really doesn't matter. So I've updated the PR to use the more common implicit tuple "constructor" syntax. Though the null forgiving thing still makes sense to use in case Entity<T> gets given extra data.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -354,32 +354,49 @@ private void OnCompStartup(EntityUid uid, TransformComponent xform, ComponentSta
...
+            DebugTools.AssertEqual(ent.Comp1.ParentUid, oldParent);
+#endif
+        }
+
+        ent.Comp1._gridUid = gridId;
+
+        foreach (var child in ent.Comp1._children)
+        {
+            SetGridId((child, XformQuery.GetComponent(child), null), gridId);
         }
```
metalgearsloth, line 400: Definitely right call to try and punish callers if they modify it with the event.

metalgearsloth (approved):

PR #5894, opened 2025-04-26
Check audio file signatures instead of extensions

### LINKED ISSUE #5789: Determining the audio file container should not be dependant by its file extension
https://github.com/space-wizards/RobustToolbox/blob/c7d228c223cb8df80e0d33ada6399d5e1cdc74ed/Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs#L32-L43

This is more of an issue with the `playglobalsound` command in 14, but we can't expect the user to upload a file with the extension, and this will end up claiming that "the file is not an ogg or wav" which is clearly false. It just does not have the file extension

Instead, this should be automatically detected via checking the file signature.

dffdff2423:
Fixes #5789

Test audio files based on their magic bytes.

I am not really sure what else to put here.

### DIFF
--- a/Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs
+++ b/Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using Robust.Client.Audio;
 using Robust.Shared.Audio;
@@ -11,6 +12,13 @@ namespace Robust.Client.ResourceManagement;
 
 public sealed class AudioResource : BaseResource
 {
+    // from: https://en.wikipedia.org/wiki/List_of_file_signatures
+    private static readonly byte[] OggSignature = "OggS"u8.ToArray();
+    private static readonly byte[] RiffSignature = "RIFF"u8.ToArray();
+    private const int WavSignatureSkip = 8; // RIFF????
+    private static readonly byte[] WavSignature = "WAVE"u8.ToArray();
+    private const int MaxSignatureLength = 12; // RIFF????WAVE
+
     public AudioStream AudioStream { get; private set; } = default!;
 
     public void Load(AudioStream stream)
@@ -28,12 +36,20 @@ public sealed class AudioResource : BaseResource
         }
 
         using var fileStream = cache.ContentFileRead(path);
+        var signature = new byte[MaxSignatureLength];
+        if (fileStream.Read(signature) != signature.Length)
+        {
+            throw new IOException("Unable to read signature from Audio file.");
+        }
+        fileStream.Seek(0, SeekOrigin.Begin);
+
         var audioManager = dependencies.Resolve<IAudioInternal>();
-        if (path.Extension == "ogg")
+        if (signature.Take(OggSignature.Length).SequenceEqual(OggSignature))
         {
             AudioStream = audioManager.LoadAudioOggVorbis(fileStream, path.ToString());
         }
-        else if (path.Extension == "wav")
+        else if (signature.Take(RiffSignature.Length).SequenceEqual(RiffSignature)
+                 && signature.Skip(WavSignatureSkip).SequenceEqual(WavSignature))
         {
             AudioStream = audioManager.LoadAudioWav(fileStream, path.ToString());
         }
--- a/Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs
+++ b/Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs
@@ -3,7 +3,6 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using Robust.Client.Audio;
-using Robust.Shared.Audio;
 using Robust.Shared.ContentPack;
 using Robust.Shared.IoC;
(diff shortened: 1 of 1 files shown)


Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs
```diff
@@ -28,12 +36,20 @@ public override void Load(IDependencyCollection dependencies, ResPath path)
         }
 
         using var fileStream = cache.ContentFileRead(path);
+        var signature = new byte[MaxSignatureLength];
+        if (fileStream.Read(signature) != signature.Length)
+        {
+            throw new IOException("Unable to read signature from Audio file.");
+        }
+        fileStream.Seek(0, SeekOrigin.Begin);
```
PJB3005: We can't guarantee that these file streams are seekable.

PJB3005 (changes requested):

Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs
dffdff2423: Is there a better way of fixing this other than creating two streems?

PJB3005: > Is there a better way of fixing this other than creating two streems?

Some kind of buffered stream that can allow you to peek the first couple bytes while still allowing the full file to be passed through.

dffdff2423: > Some kind of buffered stream that can allow you to peek the first couple bytes while still allowing the full file to be passed through.

Doing a bit of research, I can't find anything that would let me do that in the stdlib, but I am not exactly a c# expert. Is there some class that RT has or I am missing which I am not aware of?

- [BufferedStream](https://learn.microsoft.com/en-us/dotnet/api/system.io.bufferedstream?view=net-9.0) is not peekable.
- [StreamReader](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.peek?view=net-9.0#system-io-streamreader-peek) exists but is intended for text data and not binary data. Additionally, only the first character is Peekable which would probably be fine in this situation but is not ideal.
- Using a MemoryStream to guarantee it is seekable seems like a massive waste of memory for this.

If I am not missing something would you like me to implement one for RT? That seems a bit overkill for such a small feature like this.

PJB3005: > Is there some class that RT has or I am missing which I am not aware of?

You'd need to make it yourself, yes.

To be honest, if you want a crappy solution for now, just check if the stream is seekable, and if it isn't copy it to a memory stream.

Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs
```diff
@@ -28,14 +35,19 @@ public override void Load(IDependencyCollection dependencies, ResPath path)
         }
 
         using var fileStream = cache.ContentFileRead(path);
+        var seekableStream = fileStream.CanSeek ? fileStream : fileStream.CopyToMemoryStream();
+        byte[] signature = seekableStream.ReadExact(MaxSignatureLength);
+        seekableStream.Seek(0, SeekOrigin.Begin);
+
         var audioManager = dependencies.Resolve<IAudioInternal>();
-        if (path.Extension == "ogg")
+        if (signature.Take(OggSignature.Length).SequenceEqual(OggSignature))
```
PJB3005: Can you use a range index instead of LINQ for this?

Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs
```diff
@@ -28,14 +35,19 @@ public override void Load(IDependencyCollection dependencies, ResPath path)
...
         {
-            AudioStream = audioManager.LoadAudioOggVorbis(fileStream, path.ToString());
+            AudioStream = audioManager.LoadAudioOggVorbis(seekableStream, path.ToString());
         }
-        else if (path.Extension == "wav")
+        else if (signature.Take(RiffSignature.Length).SequenceEqual(RiffSignature)
```
PJB3005: Ditto

PJB3005 (changes requested):

Robust.Client/ResourceManagement/ResourceTypes/AudioResource.cs
dffdff2423: Huh, I did not know c# supported that or I would have. Sorry, will fix.

Is there an array compare func other than SequenceEqual? That is under Linq.

PJB3005: You can use `.AsSpan()` to get a subslice of the array and then compare it with `.SequenceEqual()`, that'll also be the most performant way (not that that matters here).

dffdff2423: Oops- I did not see that (I was tabbed out). Is the way I did it now fine, or do you want me to change it to that?

PJB3005: It doesn't really matter.

PR #5895, opened 2025-04-27
add EntGridChangedEvent 

MilonPL:
Adds the EntGridChangedEvent that is raised whenever an entity changes grids, if they have the GridTracking flag. This is because EntParentChangedMessage doesn't work inside containers and out of all solutions, this seemed like the least sketchy one.

~~I'm not entirely sure on adding a separate flag for it since we could maybe reuse InContainer? Realistically you only need this when an entity isn't parented to the grid, so for stuff like inventory and storage, because you can use EntParentChanged for everything else. But maybe that changes in the future?~~

merge with https://github.com/space-wizards/space-station-14/pull/36972

B/C 
changed the signature of `SetGridId` from `(EntityUid uid, TransformComponent xform, EntityUid? gridId, EntityQuery<TransformComponent>? xformQuery = null)` to `(EntityUid uid, TransformComponent xform, MetaDataComponent meta, EntityUid? gridId)`
i'll write release notes later if no one else does, want a vibe check first

### DIFF
--- a/Robust.Shared/GameObjects/Components/MetaDataComponent.cs
+++ b/Robust.Shared/GameObjects/Components/MetaDataComponent.cs
@@ -238,6 +238,11 @@ namespace Robust.Shared.GameObjects
         /// a grid or map.
         /// </summary>
         PvsPriority = 1 << 4,
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


metalgearsloth:
https://github.com/space-wizards/RobustToolbox/pull/5893 ?

PR #5896, opened 2025-04-27
Replace virtual `Component.SessionSpecific` property with attributes

ElectroJr:
Currently component-type specific networking can be configured by overriding two virtual bool getters (`Component.SendOnlyToOwner` and `Component.SessionSpecific`). This PR replaces these with an argument passed to the `NetworkedComponentAttribute`.  This change isn't strictly required for anything in particular I just feel like its a better approach and the properties probably should've used attributes in the first place.

This PR also adds another networking restriction type for component states that should only be saved to a replay and not sent to any players (useful for #5883). The majority of the changes have to do with changing the `MetadataComponent.NetComponents` dictionary values from `IComponent` to an `(IComponent, StateRestriction)` .

Requires https://github.com/space-wizards/space-station-14/pull/36974


Robust.Shared/GameStates/NetworkedComponentAttribute.cs
```diff
@@ -1,10 +1,52 @@
...
+    /// <summary>
+    /// This component will raise an <see cref="ComponentGetStateAttemptEvent"/> to determine whether the component
+    /// should be networked to any specific session.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class)]
-    public sealed class NetworkedComponentAttribute : Attribute { }
+    /// <remarks>
+    /// Replays will still always receive this component.
+    /// </remarks>
+    SessionSpecific = 3
```
MilonPL, line 54: I feel like the better descriptions explained it better, especially now that it is on the attribute.
This could mean the *component* itself is not networked (as in, it's not replicated to all players outside the restriction), or the *component state* is not sent to players, but the component is still replicated with it's initial values.

ElectroJr: I've updated the comments a bit, which hopefully makes it a bit clearer that the whole component is no longer being networked, not that they're receiving an empty state.

PR #5897, opened 2025-04-28
Add documentation to the serializer interfaces and remove ITypeReaderWriter

dffdff2423:
It took me three hours to figure out how to implement a custom serializer correctly. I *think* I have figured it out. This is some docs to hopefully help anybody else who is looking at this down the line.

I have no idea what the difference between ITypeSerializer and ITypeReaderWriter is so I did not include any documentation on ITypeReaderWriter. ITypeReaderWriter is not used anywhere except as a parent of ITypeSeralizer.


Robust.Shared/Serialization/TypeSerializers/Interfaces/ITypeSerializer.cs
```diff
@@ -2,6 +2,15 @@
 
 namespace Robust.Shared.Serialization.TypeSerializers.Interfaces
 {
+    /// <summary>
+    /// A serializer for a given type that supports reading and writing.
+    /// </summary>
+    /// <p>
+    /// Types that implement this should be marked with the following annotation:
+    /// </p>
+    /// <see cref="Robust.Shared.Serialization.Manager.Attributes.TypeSerializerAttribute"/>
```
ElectroJr: The annotation is for the "default" serializers for that type, not all serializers should have the annotation.
E.g., look at any serializers specified on a datafield via the `customTypeSerializer` argument. They're serializers that implement the interface, don't have the attribute and thus aren't the "default" serializer for that type.

You should probably also add a comment to that effect on the attribute

ElectroJr:
> ITypeReaderWriter is not used anywhere except as a parent of ITypeSeralizer.

It might be leftover from an older version of serialization code? Either way its probably fine to just remove seeing as it appears to be unused.

Robust.Shared/Serialization/Manager/Attributes/TypeSerializerAttribute.cs
```diff
@@ -3,6 +3,9 @@
 
 namespace Robust.Shared.Serialization.Manager.Attributes
 {
+    /// <summary>
+    /// Registers a <see cref="Robust.Shared.Serialization.TypeSerializers.Interfaces.ITypeSerializer"/> as a default serializer for it's type
```
PJB3005:
```suggestion
    /// Registers a <see cref="Robust.Shared.Serialization.TypeSerializers.Interfaces.ITypeSerializer"/> as a default serializer for its type
```

Robust.Shared/Serialization/TypeSerializers/Interfaces/ITypeSerializer.cs
```diff
@@ -1,10 +1,17 @@
...
-}
+namespace Robust.Shared.Serialization.TypeSerializers.Interfaces;
+
+/// <summary>
+/// A serializer for a given type that supports reading and writing. Can be used as a default serializer or a custom serializer on a datafield.
+/// </summary>
+/// <p>
+/// Types that implement this may be annotated with <see cref="Robust.Shared.Serialization.Manager.Attributes.TypeSerializerAttribute"/>
+/// to register it as the default serializer for a type
+/// </p>
```
PJB3005: This needs to be in a `<remarks>` block or something. I don't think having a plain paragraph is valid.

Robust.Shared/Serialization/TypeSerializers/Interfaces/ITypeWriter.cs
```diff
@@ -6,6 +6,9 @@ namespace Robust.Shared.Serialization.TypeSerializers.Interfaces
 {
     public interface ITypeWriter<TType> : BaseSerializerInterfaces.ITypeInterface<TType>
     {
+        /// <summary>
+        /// Converts value into it's DataNode representation.
```
PJB3005:
```suggestion
        /// Converts value into its DataNode representation.
```

PJB3005 (changes requested):

Robust.Shared/Serialization/TypeSerializers/Interfaces/ITypeSerializer.cs
dffdff2423: ah okay, it showed up correctly in Rider so I assumed it was fine lol.

PR #5903, opened 2025-04-28
FixedPoint to Engine.

gusxyz:
Allows downstreams to use FixedPoint2 in projects that do not reference the Content.Shared project. 

Content PR: https://github.com/space-wizards/space-station-14/pull/37022

### DIFF
--- a/Robust.Shared/FixedPoint/FixedPoint2.cs
+++ b/Robust.Shared/FixedPoint/FixedPoint2.cs
@@ -1,0 +1,336 @@
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using Robust.Shared.Serialization;
+using Robust.Shared.Serialization.Manager.Attributes;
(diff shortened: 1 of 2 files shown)


metalgearsloth:
You're doing this for SS14 reasons not engine reasons so I don't really understand this change. If someone wants to hard fork it's what they deal with on SS14 and a typical game isn't going to care about this as a feature.

PJB3005:
Yeah no this should not be in engine.

metalgearsloth:
FWIW I have checked godot, unity, and unreal, and none of them have a simple fixed-point datatype baked in, but happy to be proven wrong. Moving BUIs, which would actually have a use for some tile-based games, out of engine while we're simultaneously moving a specific implementation of fixed maths is silly.

PJB3005:
The problem with FixedPoint2 is that it is FixedPoint*2*. I don't feel like adding variants for every conceivably useful precision to the engine.

PR #5874, opened 2025-04-20
ftl hot reloading

MilonPL:
adds ftl hot reloading with the ReloadManager

b/c for engine, LocalizationManager is now abstract with respective ClientLocalizationManager and ServerLocalizationManager classes
LocalizationManager no longer uses IPostInjectInit in favor of Initialize 

media below

https://github.com/user-attachments/assets/c6f816dc-2afc-410f-903d-e8358a05b8ed

PR #5907, opened 2025-05-04
fix yaml hotreloading

slarticodefast:
Basically reverts only the code cleanup changes done to the `ReloadManager` in https://github.com/space-wizards/RobustToolbox/pull/5874
They were removed because the IDE is telling you the `_watchers` list is never used. Turns out it was load bearing. I assume it prevents the watchers from being garbage collected.

There is probably a prettier way to do this, but this just returns it to the original state and adds a code comment to prevent it from happening again.


slarticodefast:
Also could use an integration test apparently.

VerinSenpai:
load bearing. My god. I hate it.

PR #5909, opened 2025-05-05
AnimationTrackProperty.KeyFrames to public

Princess-Cheeseballs:
I removed the protected property from the set property for the KeyFrames list.

This allows you to create methods to auto-generate and calculate entire animations as a list of keyframes and then feed that list directly into a new animation instead of having to handcraft each individual keyframe.

Allows for "real time" animation creation (you still have to create methods that make the lists but you can alter how many keyframes there are and such in real time).


metalgearsloth:
Sure...

PR #5910, opened 2025-05-05
More TimespanSerializer improvements

Errant-4:
Some code that was made redundant in the last PR by TryTimeSpan has been removed from TimeSpanSerializer.Read
A number parsing now only existed to throw if a non-number entry is input (since the numbers it was originally meant to parse are now handled in the external call). It has now been replaced with a simple throw for clarity and to more specifically communicate that particular problem.

I moved TryTimeSpan to TimeSpanExt.cs and made it public. It can now also be called with plain string input. This will be needed to enable using the new timespan inputs in console commands

There was also a potential issue with commas in the input number resulting in unintentionally large inputs. This could happen if the user mistakenly believes they count as a decimal separator. To avoid such scenarios, the code now throws if there is a comma in the input, so the command safely fails rather than behave "unexpectedly". InvariantCulture that was removed in the last update has also been re-added to all relevant places, to avoid potential cross-locale issues.


Errant-4:
Hmm maybe I should also make it accept capital s m h too

Errant-4:
Potential issue that this is hardcoded for english, but then again so is the entirety of yaml, so I guess it does not really change much

PR #5913, opened 2025-05-08
Reduce logging details in the Joint System

### LINKED ISSUE #5726: Pull Joint error revealing usernames
https://github.com/space-wizards/RobustToolbox/blob/543088ea1feb3d45cbd106b220451cf310d748a4/Robust.Shared/Physics/Systems/SharedJointSystem.cs#L164

The above mentioned system.joint error leaks the usernames of all 3 users "affected" by this error readout.

Wizard mindswap and changeling transform abilities are free to be metagamed with this information, via the use of extreme metagrudge knowledge.

B-Kirill:
When logging errors in the `SharedJointSystem`, the names of the players and some other information are displayed in the console, which in theory can be used for dishonest purposes. Of course, the chance of this is small, but it is there. 

The most common example is to find out about the presence of nuclear operatives or another antagonists by their name.

Fixes https://github.com/space-wizards/RobustToolbox/issues/5726

A good example is the log from [https://github.com/space-wizards/space-station-14/issues/11644](https://github.com/space-wizards/space-station-14/issues/11644)

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedJointSystem.cs
+++ b/Robust.Shared/Physics/Systems/SharedJointSystem.cs
@@ -160,7 +160,9 @@ public abstract partial class SharedJointSystem : EntitySystem
         {
             if (existing.BodyBUid != bUid)
             {
+#if !RELEASE
                 Log.Error($"While adding joint {joint.ID} to entity {ToPrettyString(bUid)}, the connected entity {ToPrettyString(aUid)} already had a joint with the same ID connected to another entity {ToPrettyString(existing.BodyBUid)}.");
+#endif
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


ElectroJr:
This feels like a bad solution. I agree that info and some warning logs should avoid leaking information to players, but if there are legitimate errors they should get logged. I.e., if a player runs into a joint related bug on a server and wants to report it, these errors should be in their logs. I'd rather work on fixing whatever is causing these errors to be logged. And if these errors aren't actual errors, they should be turned into info or warning logs

And even if other maints want to disable the logs, they should still get logged on the server & singleplayer games.

B-Kirill:
> This feels like a bad solution. I agree that info and some warning logs should avoid leaking information to players, but if there are legitimate errors they should get logged. I.e., if a player runs into a joint related bug on a server and wants to report it, these errors should be in their logs. I'd rather work on fixing whatever is causing these errors to be logged. And if these errors aren't actual errors, they should be turned into info or warning logs
> 
> And even if other maints want to disable the logs, they should still get logged on the server & singleplayer games.

These errors are quite common, I would make them warnings, especially since the code continues to work without problems.
Then the warning can simply output the Uid of the entity, without a name or type.

B-Kirill:
Changed the error to a warning. Also, the warning now shows the Uid of the entities, which will allow players to report this and at the same time not show excess information to the player.

TurboTrackerss14:
https://github.com/space-wizards/RobustToolbox/issues/5726 *coughs*

metalgearsloth:
> Changed the error to a warning. Also, the warning now shows the Uid of the entities, which will allow players to report this and at the same time not show excess information to the player.

Add redaction for it but that shouldn't get in the way of actual error reporting, I agree with electro on this one.

PR #5914, opened 2025-05-08
Language server

ciaran:
Proof of concept for Prototype YAML language server

Implements quite a few features but currently not well structured as it was a combination of exploring how a LS works and how RT works.

This attempts to do the minimum to get prototype loading (and the requisite parts) active, and then leans on the existing YAML Linter implementation to provide diagnostics.

Currently _only the server assembly_ is loaded, but this does still appear to be good enough to be useful for most cases (only exception I’ve run into so far when poking around prototype files is the Sprite component)

Intent is to continue to work on this but I wanted to discuss direction and collect possible minimum requirements for merging.

---

You will also need this branch for the content repo – https://github.com/ciaran/space-station-14/tree/robust-language-server
Which enables generating XML docs for the `Content.Server` and `Content.Shared` assemblies.

To use it:

* clone this repo https://github.com/ciaran/vscode-robust-toolbox
* do an `npm install`
* Then open this folder in VSCode, go to View -> Run, then press "Launch Client" at the top-left.
* It should open a new VSCode window with the extension active, open the `space-station-14` root folder in this window and then open a .yml prototype file


ElectroJr:
> but I wanted to discuss direction and collect possible minimum requirements for merging.

I've never looked into how language servers work or how to implement one, and have only just learned a bit while playing around with this PR. I also haven't paid any attention to the changes to PrototypeManager or the actual build process / how it finds assemblies.

But aside from those things, the general way this seems to work seems fine to me. Some parts probably need cleaning up (e.g., the use of all the static IoC resolves & use of `IoCManager.InitThread()` in all the lambdas). But I'd be happy to merge a pretty minimal version of this that initially only supports a couple of features / handlers (e.g., only hover support), and then hopefully features can be added over time.

The main issue I noticed is that the definition / go-to handler was trying to reconstruct the paths to C# source code from the type's full name, which generally just won't work and has to be handled some other way. I'm not sure how best to resolve that, but maybe:
- Can the LSP client handle non-URI based responses? I.e., obviously most IDEs already knows how to go a C# type when the solution is open, can that be reused?
- Can the file paths be exported alongside the xml docs for debug builds?
- Some other way to store the paths in the assemblies/pdb files? I don't actually know much about pdf files, and I know they have some paths for debugging, but I assume thats just for methods, not fields & properties?
- Use the `CallerFilePathAttribute`?
  - I have a demo of this working, where I just added it to the `DataFieldAttribute` constructor, which makes handling going to the definition of any yaml data fields pretty easy.
  - It's [available on this branch](https://github.com/ElectroJr/RobustToolbox/tree/test_lsp_PR) where I also made a few other changes and (IMO) cleaned some things up a bit.

See later comment ~~I'm also not fully sure if the way the initialization/loading is handled is ideal. It might be easier to do something like the map renderer, where it abuses integration test methods to set up a server & headless client pair while avoiding a lot of the initialization boilerplate. But on the other hand, maybe its better to "hand craft" it to avoid bloat, seeing you probably don't want a full server & client instance running in the background while developing.~~

Another minor thing is that the XML doc generation changes might conflict a bit with #5430, which seems to do some kind of custom processing of the generated xml files.

And if anyone else wants to also try play around with the language server & doesn't use VSCode or can't be bothered with plugins, I've just been using neovim to test it and its pretty simple to set up. You can more or less just throw a simple config file into `nvim/lsp/`. e.g., 
```lua
return {
  cmd = vim.lsp.rpc.connect("127.0.0.1", <port>),
  filetypes = { 'yaml'},
  root_markers = { 'SpaceStation14.sln' },
}
```

ElectroJr:
> I wanted to discuss direction and collect possible minimum requirements for merging.

I brought up this PR [on discord](https://discord.com/channels/310555209753690112/770682801607278632/1425477151540379820) when some other LSPs / plugins were being discussed (https://github.com/Ertanic/robust-lsp, https://github.com/Slava0135/vscode-robust-yaml), and it seems like one of the requirements is going to be that it needs to handle loading the game assemblies differently

Discord_64xNGsQARY

ciaran:
I don’t think that’s really necessary, it could just be restarted automatically if necessary so it won’t really interrupt the user

metalgearsloth:
Re-open if it's finished.

PR #5915, opened 2025-05-10
Always align newly created entities with the grid

EmoGarbage404:
Changes behavior in the `CreateEntityUninitialized` `MapCoordinates` overload to always parent newly created entities to the grid, if one is present.

The current behavior only parents to the grid if the entity is anchored, which creates a strange inconsistency when using map coordinates for spawning where the entity is rotated relative the map and not to the grid. Every other spawning related method treats rotation as relative to the grid, so this should be preserved here as well.

Is not strictly breaking but may change behavior for users. Lmk if anything else is needed.


metalgearsloth:
Hell yeah it's gonna break shit welcome to transform init

iaada:
I was looking into this, and both test fails seem to be because the test expects to spawn the item parented to the map, which no longer happens because we now spawn on a grid.

```
public void OnAnchored_Parent_SetToGrid()
{
    var (sim, grid, coordinates, xformSys, mapSys) = SimulationFactory();

    // can only be anchored to a tile
    mapSys.SetTile(grid, mapSys.TileIndicesFor(grid, coordinates), new Tile(1));

    var traversal = sim.System<SharedGridTraversalSystem>();
    traversal.Enabled = false;
    var ent1 = sim.SpawnEntity(null, coordinates); // this raises MoveEvent, subscribe after

    // Act
    sim.System<MoveEventTestSystem>().ResetCounters();
    sim.Transform(ent1).Anchored = true;
    Assert.That(sim.Transform(ent1).ParentUid, Is.EqualTo(grid.Owner)); // Passes
    sim.System<MoveEventTestSystem>().AssertMoved(); // Fails
    traversal.Enabled = true;
}
```
This one fails because it asserts that the parent changed, which it didn't. But it passes on the assert that it's parented to the grid, because it spawned on the grid.

```
public void Unanchored_Unanchor_Nop()
{
    var (sim, grid, coordinates, xformSys, mapSys) = SimulationFactory();

    // can only be anchored to a tile
    mapSys.SetTile(grid, mapSys.TileIndicesFor(grid, coordinates), new Tile(1));

    var traversal = sim.System<SharedGridTraversalSystem>();
    traversal.Enabled = false;
    var ent1 = sim.SpawnEntity(null, coordinates); // this raises MoveEvent, subscribe after

    // Act
    sim.System<MoveEventTestSystem>().FailOnMove = true; // Passes
    xformSys.Unanchor(ent1);
    Assert.That(sim.Transform(ent1).ParentUid, Is.EqualTo(mapSys.GetMap(coordinates.MapId))); // Fails
    sim.System<MoveEventTestSystem>().FailOnMove = false;
    traversal.Enabled = true;
}
```
This one fails because it asserts that we're parented to the map when we're parented to the grid. But it passes the check that we don't move.

EmoGarbage404:
holy shit the best solution: i just make the tests fuck off

EmoGarbage404:
@metalgearsloth updated tests to reflect new behavior so it should actually be good to merge now

not letting it perish like all my other engine PRs

Tayrtahn:
For the record, those tests are borderline broken due to changes in anchoring from when it was ECS-ified, so updating them is definitely the right move here.

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -349,7 +349,7 @@ public virtual EntityUid CreateEntityUninitialized(string? prototypeName, MapCoo
                 throw new ArgumentException($"Attempted to spawn entity on an invalid map. Coordinates: {coordinates}");
 
             EntityCoordinates coords;
-            if (transform.Anchored && _mapManager.TryFindGridAt(coordinates, out var gridUid, out var grid))
+            if (_mapManager.TryFindGridAt(coordinates, out var gridUid, out var grid))
             {
                 coords = new EntityCoordinates(gridUid, _mapSystem.WorldToLocal(gridUid, grid, coordinates.Position));
                 _xforms.SetCoordinates(newEntity, transform, coords, rotation, unanchor: false);
```
ElectroJr, line 360: The methods docs say the rotation is the "map rotation", not local rotation. I'm not opposed to changing it to local rotation, but then the docs should be updated and you might need to double check if there's anything that was relying on it being the world/map angle.

Alternatively I guess make it a `rotation==null || anchored` check, which would leave it as it was (inc leaving it incorrect for anchored ents).

ElectroJr (changes requested):

EmoGarbage404:
@ElectroJr I updated the docs to reflect that this uses local rotation. I think that's the more logical solution since some of the Spawn overrides which eventually call this method are actually based on EntityCoordinates and local rotation to begin with.

I did vet content to see if there were any breakages. Thankfully, the only thing that I noticed was a single small usage, which I can update after this PR is merged.

LMK if there's anything else I need to do.

ElectroJr:
I can try look into it later, but it looks like a content test is failing due to an observer trying to attach itself to a terminating grid?
```
2025-10-04T17:58:15.6365695Z   Failed TestGhostGridNotTerminating [410 ms]
2025-10-04T17:58:15.6366272Z   Error Message:
2025-10-04T17:58:15.6368248Z      Assert.That(code, new ThrowsNothingConstraint())
2025-10-04T17:58:15.6368719Z   Expected: No Exception to be thrown
2025-10-04T17:58:15.6369298Z   But was:  <NUnit.Framework.AssertionException: Multiple failures or warnings in test:
2025-10-04T17:58:15.6370543Z   1) SERVER: 0.404s [ERRO] system.transform: observer (197517/n197517, MobObserver) is attempting to attach itself to a terminating entity grid (197514/n197514). Trace:    at System.Environment.get_StackTrace()
```

PR #5919, opened 2025-05-10
Mark `Entity` methods as readonly

Tayrtahn:
Marks the `GetHashCode()`, `AsNullable()`, `AsType()` methods, and the `FluentOwner` property of the `Entity` variants as `readonly`. This lets the compiler know that they are pure and allows them to be called on readonly `Entity`s (like in event args) without the compiler secretly making a deep copy of the struct.

More info on readonly struct methods [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct#readonly-instance-members).

This was prompted by [a Discord conversation](https://discord.com/channels/310555209753690112/560845886263918612/1370703755464409129) about running into a compiler warning when trying to call `AsNullable` on an `Entity` event argument:

Screenshot 2025-05-10 at 10 29 39 AM

### DIFF
--- a/Robust.Shared/GameObjects/Entity.cs
+++ b/Robust.Shared/GameObjects/Entity.cs
@@ -11,7 +11,7 @@ public record struct Entity<T> : IFluentEntityUid, IAsType<EntityUid>
 {
     public EntityUid Owner;
     public T Comp;
-    EntityUid IFluentEntityUid.FluentOwner => Owner;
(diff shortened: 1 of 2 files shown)


PJB3005 (changes requested):
These are generated with `GenericEntityPrint.cs`. You need to update that instead.

PR #5920, opened 2025-05-11
Fix sprite layer bounds

ElectroJr:


Robust.Client/GameObjects/EntitySystems/SpriteSystem.Bounds.cs
```diff
@@ -53,9 +56,6 @@ public Box2 GetLocalBounds(Layer layer)
 
     internal Box2 CalculateLocalBounds(Layer layer)
     {
-        if (layer.Blank || layer.CopyToShaderParameters == null)
-            return Box2.Empty;
```
ElectroJr, line 57: The `==` here should've been a `!=`., which was causing almost all layers to have an empty bounding box. But in hindsight this can just be removed altogether.

PR #5926, opened 2025-05-12
Static Field Assert

Aidenkrz:
I'm very tired of receiving asserts that mean absolutely nothing, from now on I'm going to rewrite them everytime I see them to actually mean anything useful.

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.ValidateFields.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.ValidateFields.cs
@@ -63,7 +63,7 @@ public partial class PrototypeManager
         Dictionary<Type, HashSet<string>> prototypes)
     {
         DebugTools.Assert(field.IsStatic);
-        DebugTools.Assert(!field.HasCustomAttribute<DataFieldAttribute>(), $"Datafields should not be static. Field: {field.Name} in Type: {field.DeclaringType?.FullName}");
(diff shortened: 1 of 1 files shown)


Robust.Shared/Prototypes/PrototypeManager.ValidateFields.cs
```diff
@@ -63,7 +63,7 @@ private void ValidateStaticField(
         Dictionary<Type, HashSet<string>> prototypes)
     {
         DebugTools.Assert(field.IsStatic);
-        DebugTools.Assert(!field.HasCustomAttribute<DataFieldAttribute>(), "Datafields should not be static");
+        DebugTools.Assert(!field.HasCustomAttribute<DataFieldAttribute>(), $"Datafields should not be static. Field: {field.Name} in Type: {field.DeclaringType?.FullName}");
```
PJB3005:
```suggestion
        DebugTools.Assert(!field.HasCustomAttribute<DataFieldAttribute>(), $"Datafields should not be static. Field: {field.Name} in Type: {field.DeclaringType}");
```

Better to just let the full type be formatted, probably.

Aidenkrz: 👍

VerinSenpai:
question, wouldn't the field directly before this one prevent this one from going off and providing useful information, or is that just not a concern?

PJB3005:
No, that's an unrelated check.

VerinSenpai:
roger that 👌

PR #5931, opened 2025-05-14
Mark `NetEntity` as `NotYamlSerializable`

Tayrtahn:
Adds the `NotYamlSerializable` attribute to `NetEntity`, so the DataDefinition analyzer will prevent it being used as a DataField.

This should not be merged until violations are cleaned up in content - see https://github.com/space-wizards/space-station-14/issues/37464


slarticodefast:
Should we do the same for `NetCoordinates` since those contain a `NetEntity` or does that happen automatically through marking `NetEntity`?

PR #5932, opened 2025-05-15
Add new Vertical property to progress bars

ArtisticRoomba:
This PR adds a new Vertical property to `ProgressBar`, which allows you to specify vertical progress bars for whatever you may need. If no property is specified it simply uses the default horizontal bar.

Required for content PR https://github.com/space-wizards/space-station-14/pull/36708

Content Client_P6uiY8mQbW

All other progress bars, like in APCs and in the powernet monitor, seem to work fine.

image


metalgearsloth (approved):

metalgearsloth:
Could also maybe make it its own control.

ArtisticRoomba:
probably should + progressbars should get fields for specifying direction of fill, but tbh I'd rather get this in for the content PR than inflate scope (even though it wouldn't be that bad)

PR #5934, opened 2025-05-15
Fix `NotYamlSerializable` analyzer ignoring nullable structs

Tayrtahn:
The portion of `DataDefinitionAnalyzer` that checks that DataField types are not marked with the `NotYamlSerializableAttribute` has an issue where nullable struct types are ignored.

This is because dotnet treats `MyStruct?` as shorthand for `Nullable<MyStruct>`, and when the analyzer goes to get attributes of the field type, it winds up getting the attributes of `Nullable<T>`. Nullable classes don't have this issue because they work differently, for reasons.

To fix this, the analyzer now uses a new helper method: `TypeSymbolHelper.GetNullableUnderlyingTypeOrSelf` which checks if the passed `ITypeSymbol` is a `Nullable<T>` and returns the `ITypeSymbol` of `T` if it is. If the passed type isn't a `Nullable<T>`, it just spits the passed type back out.

This is needed to catch some violations with https://github.com/space-wizards/RobustToolbox/pull/5931 that would otherwise be ignored.


PJB3005:
Won't this also like break if you do `List<NetEntity>` though?

PJB3005 (approved):
Whatever I'm just merging this

PR #5935, opened 2025-05-15
Add `SpriteSystem` dependency to `VisualizerSystem`

Tayrtahn:
Now that `SpriteComponent` methods have been deprecated in favor of `SpriteSystem` methods, it makes sense to include a reference to `SpriteSystem` in `VisualizerSystem`, seeing as most visualizers need to manipulate sprites.

I've been fixing up warnings in content and it seems silly to add a `SpriteSystem` dependency to every single one when we could just have one in the base class.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/VisualizerSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/VisualizerSystem.cs
@@ -11,6 +11,7 @@ public abstract class VisualizerSystem<T> : EntitySystem
 {
     [Dependency] protected readonly AppearanceSystem AppearanceSystem = default!;
     [Dependency] protected readonly AnimationPlayerSystem AnimationSystem = default!;
+    [Dependency] protected readonly SpriteSystem spriteSystem = default!;
 
     public override void Initialize()
     {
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/EntitySystems/VisualizerSystem.cs
```diff
@@ -11,6 +11,7 @@ public abstract class VisualizerSystem<T> : EntitySystem
 {
     [Dependency] protected readonly AppearanceSystem AppearanceSystem = default!;
     [Dependency] protected readonly AnimationPlayerSystem AnimationSystem = default!;
+    [Dependency] protected readonly SpriteSystem spriteSystem = default!;
```
slarticodefast: What is the naming convention for `protected readonly` fields?
Should the first letter be capitalized like for the other two above?

Tayrtahn: It absolutely should be, yeah! I'm kinda bewildered that I did that wrong.

slarticodefast (approved):

PR #5944, opened 2025-05-18
Cleanup warning in `SpriteBoundsOverlay`

Tayrtahn:
Fixes a `SpriteComponent` obsolete warning in `SpriteBoundsOverlay`.

I'm doing this one as a single PR because it does some minor refactoring too.

Specifically, instead of taking in each dependency as a constructor argument, the `SpriteBoundsOverlay` class now just takes in `IEntityManager` in its constructor and retrieves the specific systems from that when it needs them.

Screenshot 2025-05-18 at 11 57 46 AM

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteBoundsOverlay.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteBoundsOverlay.cs
@@ -59,10 +59,9 @@ namespace Robust.Client.GameObjects
     {
         public override OverlaySpace Space => OverlaySpace.WorldSpace;
 
-        private readonly IEntityManager _entMan = entMan;
-        private SharedTransformSystem XformSystem => _entMan.System<SharedTransformSystem>();
-        private SpriteSystem SpriteSystem => _entMan.System<SpriteSystem>();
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/Components/Renderable/SpriteBoundsOverlay.cs
```diff
@@ -57,29 +55,26 @@ public bool Enabled
...
-
-        public SpriteBoundsOverlay(SpriteTreeSystem renderTree, SharedTransformSystem xformSystem)
-        {
-            _renderTree = renderTree;
-            _xformSystem = xformSystem;
-        }
+        private readonly IEntityManager _entMan = entMan;
+        private SharedTransformSystem XformSystem => _entMan.System<SharedTransformSystem>();
+        private SpriteSystem SpriteSystem => _entMan.System<SpriteSystem>();
+        private SpriteTreeSystem RenderTree => _entMan.System<SpriteTreeSystem>();
```
metalgearsloth: Isn't this going to resolve the System<T> call every time?

Tayrtahn: Yep, it sure is. Lemme fix that.

PR #5945, opened 2025-05-18
Cleanup remaining `SpriteComponent` warnings in engine

Tayrtahn:
Fixes the remaining 19 `SpriteComponent` obsolete warnings in RT.

Technically there's still one left:
https://github.com/space-wizards/RobustToolbox/blob/4b6b688c72fa483a0bbe6a7dad6d65428ff16426/Robust.Client/GameObjects/EntitySystems/SpriteSystem.Component.cs#L77

But it looks like that's waiting for the next phase of `SpriteSystem` ECS, so I'm not touching it.


Robust.Client/Animations/AnimationTrackSpriteFlick.cs
```diff
@@ -37,10 +37,13 @@ public override (int KeyFrameIndex, float FramePlayingTime) InitPlayback()
         public override (int KeyFrameIndex, float FramePlayingTime)
             AdvancePlayback(object context, int prevKeyFrameIndex, float prevPlayingTime, float frameTime)
         {
-            DebugTools.AssertNotNull(LayerKey);
+            var layerKey = LayerKey as Enum;
```
PJB3005, line 40: Why is this cast necessary now?

Robust.Client/GameObjects/EntitySystems/GenericVisualizerSystem.cs
```diff
@@ -32,12 +33,11 @@ protected override void OnAppearanceChange(EntityUid uid, GenericVisualizerCompo
                 if (!layerDataDict.TryGetValue(appearanceValue, out var layerData))
                     continue;
 
-                object layerKey = _refMan.TryParseEnumReference(layerKeyRaw, out var @enum)
-                    ? @enum
-                    : layerKeyRaw;
+                var layerIndex = _refMan.TryParseEnumReference(layerKeyRaw, out var @enum)
+                    ? _sprite.LayerMapReserve((uid, args.Sprite), @enum)
+                    : _sprite.LayerMapReserve((uid, args.Sprite), layerKeyRaw);
```
PJB3005, line 38: These shouldn't be separate calls.

PJB3005 (changes requested):

Robust.Client/Animations/AnimationTrackSpriteFlick.cs
Tayrtahn: The new `SpriteSystem` versions of methods like `LayerMapGet` now require that the layer key is either a string or an enum, where the `SpriteComponent` versions accepted any object.

Nothing in engine or content breaks if I change the `LayerKey` property from `object?` to `Enum?`, though I suppose it's possible that a downstream might be using a string value somewhere. Is that worth changing?

PJB3005: > The new SpriteSystem versions of methods like LayerMapGet now require that the layer key is either a string or an enum, where the SpriteComponent versions accepted any object.

These functions should be fixed, then. There's no reason for these to be separate.

Tayrtahn: Should I change them back to `object?` then?

ElectroJr: I created #6237 to make working with layer keys easier, which should fix this.

PR #5947, opened 2025-05-20
Allow multiple clients running WebView on a singular machine.

gusxyz:
Currently with WebView it is impossible to run multiple clients as its trying to use the same port for remote debugging and the same cache folder which causes the client to crash and for CEF to just open a chromium window. Yes, this isnt ideal but due to limitations on CEF's side this is what we have to do for now.

This fix allow multiple clients' to be ran on a singular machine.


Robust.Client.WebView/Cef/WebViewManagerCef.cs
```diff
@@ -30,7 +30,9 @@ internal sealed partial class WebViewManagerCef : IWebViewManagerImpl
         [Dependency] private readonly ILogManager _logManager = default!;
         [Dependency] private readonly ILocalizationManager _localization = default!;
 
+        private const int BasePort = 9222;
         private ISawmill _sawmill = default!;
+        private static readonly string ProcessName = OperatingSystem.IsWindows() ? "Robust.Client.WebView.exe" : "Robust.Client.WebView";
```
gusxyz: I really dont like this but maybe its fine due to lines 47-53? You guys tell me :(

PJB3005 (dismissed):
Yeah this is not a good approach. Just make the debug port random by default and allow people to change it via CVar if they want.

gusxyz:
Yeah its hacky, ill implement above tomorrow.

gusxyz:
> Yeah this is not a good approach. Just make the debug port random by default and allow people to change it via CVar if they want.

Do I still make it increment based on amount of web view processes'? I did this so it wouldnt spam cef_cache folders.

PJB3005:
> Do I still make it increment based on amount of web view processes'?

No, that is not a practical or reliable way to do things.

> I did this so it wouldnt spam cef_cache folders.

I am not sure what you mean with this.

gusxyz:
> > Do I still make it increment based on amount of web view processes'?
> 
> 
> 
> No, that is not a practical or reliable way to do things.
> 
> 
> 
> > I did this so it wouldnt spam cef_cache folders.
> 
> 
> 
> I am not sure what you mean with this.

If we have it completely random, each time you open a new client youll get a new 8mb cef_cache folder which isnt ideal. 

Should I just check if a cache folder already exists and if it does create cef_cache1 etc?

PJB3005:
Wait I thought this was only about debug port, CEF seriously can't re-use the cache folders? God damnit.

Put a lock file of some kind in the cache folders to detect whether they're in active use by a client. Do that to count them.

gusxyz:
> Wait I thought this was only about debug port, CEF seriously can't re-use the cache folders? God damnit.
> 
> Put a lock file of some kind in the cache folders to detect whether they're in active use by a client. Do that to count them. 

So much cleaner, Ill get to this tomorrow.

gusxyz:
finally it got the code..?

gusxyz:
This is amazing 200+ lines of code because cef is missing a --new-session flag.. 👍

PR #5948, opened 2025-05-20
Convert some physics code from `PhysicsComponent` to `Entity<PhysicsComponent>`

Tayrtahn:
Updates some parts of the physics code that use `PhysicsComponent` to use `Entity<PhysicsComponent>` instead. This allows the removal of a bunch of `.Owner` uses throughout the physics code.

This is a breaking change for `PhysicsMapComponent`, but all other changes are internal.

Requires https://github.com/space-wizards/space-station-14/pull/37647 for one change in content.

Note that while I have tested that things still work, I have not done any performance comparison to the old code. I would expect that the change from a single component to an `Entity` wouldn't be too significant, but physics code is pretty performance-sensitive, so this is something to consider.

[PZW](https://github.com/space-wizards/space-station-14/issues/33279)

## Breaking changes
`PhysicsMapComponent.AwakeBodies` is now a `HashSet<Entity<PhysicsComponent>>` instead of a `HashSet<PhysicsComponent>`.

### DIFF
--- a/Robust.Client/Physics/DebugPhysicsIslandSystem.cs
+++ b/Robust.Client/Physics/DebugPhysicsIslandSystem.cs
@@ -47,7 +47,7 @@ namespace Robust.Client.Physics
          * This will draw above every body involved in a particular island solve.
          */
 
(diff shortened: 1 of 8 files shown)


Tayrtahn:
Superceded.

PR #5950, opened 2025-05-20
fix GetMapLinearVelocity

slarticodefast:
It was missing a resolve, causing the given entity's velocity to be ignored for the calculation, but only if the component was not passed in. The integration test did not catch this because it does pass the component in everywhere.

Notably among a few other bugs this fixes the doppler effect for sounds not working:
before

https://github.com/user-attachments/assets/dcdcb294-8234-48e7-b76d-1ec9a3d3dd04

doppler_broken

after

https://github.com/user-attachments/assets/bdd3149f-9cbc-4e04-8955-54afcfbdee3b

doppler_fixed


slarticodefast:
Disabled a resolve warning.
I also adjusted the other two methods in that file to work the same in regards of a missing PhysicsComponent because it makes no sense that `GetMapVelocities` gives you a different result than `GetMapLinearVelocity` when the documentation says it is just a combination of the above two methods.

PR #5953, opened 2025-05-21
optimize ToMapCoordinates

slarticodefast:
Basically instead of 
$v_{map} = (... W_3 W_2 W_1)v_{local}$
we are doing
$v_{map} = (... W_3 (W_2 (W_1 v_{local})))$
which is mathematically equivalent.
The $W_i$ are the coordinate transformations from an entity to its parent's coordinates.
If you compare this to the previously used `GetWorldPositionRotation`
grafik
it is mostly identical, we just do the final transformation first, but we don't have to sum up the angle along the way.

This also skips this part of the previous code where the matrix was created in the end
grafik
but that optimization was a little dubious because of floating point errors.

Might make sense to benchmark this somehow before merging.

I also added a `ToWorldPosition` method with is equal to `toMapCoordinates(...).Position`, but does not create the map coordinates first. Mostly to have nicer syntax and to have an equivalent to `GetWorldPosition`.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Coordinates.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Coordinates.cs
@@ -53,8 +53,14 @@ public abstract partial class SharedTransformSystem
             return MapCoordinates.Nullspace;
         }
 
-        var worldPos = Vector2.Transform(coordinates.Position, GetWorldMatrix(xform));
-        return new MapCoordinates(worldPos, xform.MapID);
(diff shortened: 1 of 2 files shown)


metalgearsloth:
Benchmark?

slarticodefast:
> Benchmark?

How do I run one?

metalgearsloth:
> > Benchmark?
> 
> How do I run one?

Writing something lightweight in Robust.Benchmarks with robustserversim with the old implementation vs new probably.

PR #5954, opened 2025-05-21
Add WorldNormal to StartCollideEvent

metalgearsloth:
We already have the value just a matter of adding it to the event.


slarticodefast:
@beck-thompson with this you won't need to use the velocity anymore

PR #5957, opened 2025-05-23
Update SpriteSystem.Bounds.cs

DaCookieCakes:
Original issue: When (older?) maps were loaded in the Space Station 14 dev environment, and subfloors were enabled, it would crash with a DebugAssertionError pointing to these lines
image

According to an issue opened on Space Station 14 it was related to the new Pipe system
(See [SS14 issue 37448](<https://github.com/space-wizards/RobustToolbox/issues/6002>))

I went ahead and changed it to this, preventing the crash while still functioning properly
(I also cleaned up the comment)
image

Let me know if this will have any issues!

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/SpriteSystem.Bounds.cs
+++ b/Robust.Client/GameObjects/EntitySystems/SpriteSystem.Bounds.cs
@@ -10,7 +10,7 @@ using static Robust.Client.GameObjects.SpriteComponent;
 
 namespace Robust.Client.GameObjects;
 
-// This partial class contains code related to updating a sprites bounding boxes and its position in the sprite tree.
+// This partial class contains code related to updating a sprite's bounding box and its position in the sprite tree.
 public sealed partial class SpriteSystem
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/EntitySystems/SpriteSystem.Bounds.cs
```diff
@@ -19,11 +19,8 @@ public sealed partial class SpriteSystem
     /// </summary>
     public Box2 GetLocalBounds(Entity<SpriteComponent> sprite)
     {
-        if (!sprite.Comp.BoundsDirty)
-        {
-            DebugTools.Assert(sprite.Comp.Layers.All(x => !x.BoundsDirty || !x.Drawn));
+        if (!sprite.Comp.BoundsDirty && sprite.Comp.Layers.All(x => !x.BoundsDirty || !x.Drawn))
```
metalgearsloth, line 22: Isn't this just bulldozing the assert?

DaCookieCakes: > Isn't this just bulldozing the assert?

You know that's a good point, didn't really think about it too hard since it seemed to fix the issue, plus it was kinda late.
I'll look into why the pipes are crashing it, instead. Thanks for the heads up!

PR #5959, opened 2025-05-23
AudioSystem logging extension

B-Kirill:
This PR makes logging in the `AudioSystem` more detailed, which will allow players to effectively find the source of the error (now there is only the Id of the entity, which does not really help).

### DIFF
--- a/Robust.Client/Audio/AudioSystem.cs
+++ b/Robust.Client/Audio/AudioSystem.cs
@@ -582,7 +582,17 @@ public sealed partial class AudioSystem : SharedAudioSystem
     {
         if (TerminatingOrDeleted(entity))
         {
-            Log.Error($"Tried to play coordinates audio on a terminating / deleted entity {ToPrettyString(entity)}");
+            string soundInfo = "unknown sound";
+
+            if (specifier != null)
+            {
+                if (specifier is ResolvedPathSpecifier pathSpec)
+                    soundInfo = $"path: {pathSpec.Path}";
+                else if (specifier is ResolvedCollectionSpecifier collSpec)
+                    soundInfo = $"collection: {collSpec.Collection}, index: {collSpec.Index}";
+            }
+
+            Log.Error($"Tried to play coordinates audio on a terminating / deleted entity {ToPrettyString(entity)}. Sound: {soundInfo}. Trace: {Environment.StackTrace}");
             return null;
         }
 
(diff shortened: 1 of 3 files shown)


Robust.Client/Audio/AudioSystem.cs
```diff
@@ -582,7 +582,17 @@ public override (EntityUid Entity, AudioComponent Component)? PlayPredicted(Soun
...
         {
-            Log.Error($"Tried to play coordinates audio on a terminating / deleted entity {ToPrettyString(entity)}");
+            string soundInfo = "unknown sound";
+
+            if (specifier != null)
+            {
+                if (specifier is ResolvedPathSpecifier pathSpec)
+                    soundInfo = $"path: {pathSpec.Path}";
+                else if (specifier is ResolvedCollectionSpecifier collSpec)
+                    soundInfo = $"collection: {collSpec.Collection}, index: {collSpec.Index}";
```
Tayrtahn: Could this be a virtual/abstract method on `ResolvedSoundSpecifier`? `GetDebugString` or something.

Robust.Client/Audio/AudioSystem.cs
```diff
@@ -626,7 +636,17 @@ public override (EntityUid Entity, AudioComponent Component)? PlayPredicted(Soun
...
+            }
+
+            Log.Error($"Tried to play coordinates audio on a terminating / deleted entity {ToPrettyString(coordinates.EntityId)}. Sound: {soundInfo}. Trace: {Environment.StackTrace}");
```
Tayrtahn: All three of these seem to be the same? Rather than copy-pasting it, define it once in a method and call that in the three places.

B-Kirill: That's a pretty logical point

Robust.Shared/Audio/ResolvedSoundSpecifier.cs
```diff
@@ -19,6 +19,8 @@ public abstract partial class ResolvedSoundSpecifier {
     [Obsolete("String literals for sounds are deprecated, use a SoundSpecifier or ResolvedSoundSpecifier as appropriate instead")]
     public static implicit operator ResolvedSoundSpecifier(ResPath s) => new ResolvedPathSpecifier(s);
 
+    public abstract string GetDebugString();
```
PJB3005: Is just using `ToString()` not good enough here?

Robust.Client/Audio/AudioSystem.cs
```diff
@@ -753,6 +753,12 @@ protected override TimeSpan GetAudioLengthImpl(string filename)
         return _resourceCache.GetResource<AudioResource>(filename).AudioStream.Length;
     }
 
+    private void LogAudioError(ResolvedSoundSpecifier? specifier, EntityUid entityId)
```
PJB3005: This method name is not descriptive enough.

Robust.Server/Audio/AudioSystem.cs
```diff
@@ -281,4 +281,10 @@ public override void LoadStream<T>(Entity<AudioComponent> entity, T stream)
     {
         // TODO: Yeah remove this...
     }
+
+    private void LogAudioError(ResolvedSoundSpecifier? specifier, EntityUid entityId)
```
PJB3005: Ditto.

PJB3005 (changes requested):

Robust.Shared/Audio/ResolvedSoundSpecifier.cs
B-Kirill: I looked at the code again.. I think `ToString()` will be enough

PR #5960, opened 2025-05-23
Stopped recursive updates for controls that are not visible, as it vi…

Camdot:
This was causing the framerate to drop dramatically for large menus, as all controls were being updated regardless of visibility. 

I'm new to SS14 and Robust code, but this caused no issues in testing (and did fix the issue)

Here's a screenshot of a sample snapshot. I don't think it's particularly useful, but it is vaguely amusing 
image

### DIFF
--- a/Robust.Client/UserInterface/Control.cs
+++ b/Robust.Client/UserInterface/Control.cs
@@ -994,6 +994,11 @@ namespace Robust.Client.UserInterface
 
         internal int DoFrameUpdateRecursive(FrameEventArgs args)
         {
+            if (!Visible)
+            {
+                return 0;
+            }
+
             var total = 1;
             FrameUpdate(args);
 
--- a/Robust.Client/UserInterface/Control.cs
+++ b/Robust.Client/UserInterface/Control.cs
@@ -995,9 +995,7 @@ namespace Robust.Client.UserInterface
         internal int DoFrameUpdateRecursive(FrameEventArgs args)
         {
             if (!Visible)
-            {
                 return 0;
-            }
 
             var total = 1;
             FrameUpdate(args);


Robust.Client/UserInterface/Control.cs
```diff
@@ -994,6 +994,11 @@ protected virtual void Resized()
 
         internal int DoFrameUpdateRecursive(FrameEventArgs args)
         {
+            if (!Visible)
+            {
+                return 0;
+            }
```
PJB3005:
```suggestion
            if (!Visible)
                return 0;
```

PJB3005:
I wanna point out that you should probably avoid having gigantic control trees regardless if they're genuinely taking that much frame time to recurse through, as this isn't the only recursive operation we have that might affect invisible controls. That said, I don't see anything against merging this regardless.

Camdot:
@PJB3005 I agree. Thanks! Sadly, i'm working with some existing stuff, including a loadout system that has like, every item. I'll see what I can do about it in the future regardless! :)

SlamBamActionman:
Unfortunately this PR introduced a bug; I've tried to solve it myself but am kind of stumped, so commenting here to maybe get some thoughts.

This PR introduces a bug with speech bubbles. Bubbles are supposed to be hidden when the user is obscured (e.g. by a wall), and as such the chat UI code sets their visibility to false. The FrameUpdate loop then counts down the timer before the bubble should disappear. However with this PR, since the visibility is set to false, the FrameUpdate loop does not happen and the timer does not go down, resulting in the bubble appearing once the speaker is back in view. 

This might be possible to solve in Content with some sort of refactor to the bubble system, but that seems fairly difficult. As-is, bubbles seem to be a type of Control that does need to run FrameUpdate when invisible.

Tayrtahn:
I think I found a relatively simple way to fix this in the speech bubble code. Doing some testing now.

PR #5962, opened 2025-05-24
Mapping commands cleanup / QOL

Errant-4:
# Breaking Changes

The loadgrid, loadmap, savegrid and savemap commands had their parameters reordered. Path is now before MapId/GridId. Any code that calls these commands must have the two parameters swapped.

# About

Substantial code cleanup of loadmap, loadgrid, savemap and savegrid.
All console messages are now properly localized, some tweaks to wording
Some of the code was made more consistent among the different commands, and the code should be generally more readable now
Since all optional parameters must be after mandatory ones, I had to swap parameter 1 and 2 on all commands

**There is one test in Content that will have to be updated along with this engine update**
Required for https://github.com/space-wizards/space-station-14/pull/37776

### QOL
For all 4 commands, map/grid target is now an optional parameter. You could omit everything after the Y coordinate before, but now only Path to the file is strictly required.

loadgrid \<Path\> [MapID] [x] [y] [rotation] [Store Yaml UIDs]
loadmap \<Path\> [MapID] [x] [y] [rotation] [Store Yaml UIDs]
savegrid \<Path\> [Grid ID]
savemap \<Path\> [MapID] [force]

Loadgrid, savegrid and savemap automatically targets the player's map/grid, at their current coordinates. So you can just type `loadgrid /shut` and instantly get a shuttle you have previously saved

The x and y coordinates now auto-suggest the player's current coordinates, in case you wanted to spawn something with an offset. You no longer need to get them from the F3 overlay

Using loadmap without a mapId will automatically choose the next available Id. The console response tells you the number that was picked

# Media
Short version of commands

https://github.com/user-attachments/assets/025125fa-589a-48fa-a648-d70121d1c01f

Coordinate autosuggest
Screenshot_2025-05-23_203857


Errant-4:
Test fail is "normal", it's the test in Content

Also I just realized that the `mapping` command also uses the same parameter format, will it be an issue if it is now inconsistent with the other mapping commands? 

Not sure if reordering that one would be a good change, because if I go `<mapping> [path] [mapid] [grid]` there too, then it will become impossible to start mapping on a _specific_ number without loading an existing file (but on the other hand you could  always just use `mapping` without an ID and let it pick for you)

perryprog:
Could I possibly bribe you into adding a 'force' option for setting `MissingEntityBehaviour = MissingEntityBehaviour.Ignore` for saving? (There's also `LogInvalidEntities = false` for loading but I'm pretty sure that does in fact just change the log spam it gives.)

Errant-4:
> Could I possibly bribe you into adding a 'force' option for setting `MissingEntityBehaviour = MissingEntityBehaviour.Ignore` for saving? (There's also `LogInvalidEntities = false` for loading but I'm pretty sure that does in fact just change the log spam it gives.)

Uh, I guess remind me once this got merged

AJCM-git:
@Errant-4 if this is still relevant update it and ill start handling the PR

deltanedas:
`savegrid 1 /preferences.db`

- Sun Tzu, Art of War

Errant-4:
> `savegrid 1 /preferences.db`
> 
>     * Sun Tzu, Art of War

That's no different from `savegrid /preferences.db 1` that you can do on live, and I'm not sure that putting in more safeguards there is a feature that should be tacked on top of this PR

(I would also appreciate it if you did not dress up your feedback in layers of sarcasm and irony, to the point where you don't even say what you mean to say at all)

deltanedas:
yeah its very much out of scope just a funny thing

perryprog:
still crazy that sun tzu actually said that tho

Errant-4:
@AJCM-git the PR is ready for review

AJCM-git (approved):
lgtm, just filescope the namespace in mapcommands since you are already basically touching the whole file

Robust.Server/Console/Commands/MapCommands.cs
```diff
@@ -18,143 +21,266 @@ public sealed partial class SaveGridCommand : LocalizedCommands
...
+                shell.WriteError(loc.GetString("cmd-failure-no-attached-entity"));
                 return;
             }
 
-            if (!int.TryParse(args[0], out var intMapId))
+            // Read all the necessary information from the player's attached entity
+            var ent = shell.Player.AttachedEntity.Value;
+            var offset = _system.GetEntitySystem<TransformSystem>().GetWorldPosition(ent);
+            var mapId = _system.GetEntitySystem<TransformSystem>().GetMapId(ent);
+            var rot = _system.GetEntitySystem<TransformSystem>().GetWorldRotation(ent);
```
AJCM-git: almost forgot, also get transform here and give it to the sytem methods

AJCM-git (changes requested):
okay after some more testing, loadgrid coords work weird, if you dont give it coords to spawn above you it spawns the grid somewhere randomly, same if you give it coords too. Though loadmap doesnt seem to have this problem when given coords.

Errant-4:
> okay after some more testing, loadgrid coords work weird, if you dont give it coords to spawn above you it spawns the grid somewhere randomly, same if you give it coords too. Though loadmap doesnt seem to have this problem when given coords.

Strange, I could not replicate this issue? Everything spawned under me as it should
Can you reliably replicate it?

AJCM-git:
> > okay after some more testing, loadgrid coords work weird, if you dont give it coords to spawn above you it spawns the grid somewhere randomly, same if you give it coords too. Though loadmap doesnt seem to have this problem when given coords.
> 
> Strange, I could not replicate this issue? Everything spawned under me as it should Can you reliably replicate it?

Yeah i:
1. Checked out the branches (content+engine)
2. merged master on both
3. 

https://github.com/user-attachments/assets/267224c7-3f97-4d84-8a24-79155923ee8f

Errant-4:
It seems that it's specifically _newly saved_ grids that do not spawn where they should be.
The transformcomponent is also saved, and I guess those coordinates are added to where you are, when it spawns
image

If you try Maps/Shuttles/ those work fine, which is why I did not notice this. Unclear when this issue appeared, exactly

AJCM-git:
If it happens in master make an issue ill dismiss that here and merge after the other review is addressed

Errant-4:
The issue does seem to also exist on master. Presumably this has not been noticed yet because most saves are performed on maps built via mapping, and thus they would be centered near x0 y0. So even though the transformcomponent is saved and passed along when loaded, the offset would not be noticeable

Robust.Server/Console/Commands/MapCommands.cs
Errant-4: I have replaced all GetEntitySystem calls with Dependencies
Would you prefer if I waited with filescoping the namespace until you reviewed this?

Robust.Server/Console/Commands/MapCommands.cs
```diff
@@ -14,266 +16,491 @@
...
+                shell.WriteError(loc.GetString("cmd-failure-no-attached-entity"));
                 return;
             }
 
-            var mapId = new MapId(intMapId);
+            // Read all the necessary information from the player's attached entity
+            var ent = shell.Player.AttachedEntity.Value;
+            var offset = _trans.GetWorldPosition(ent);
+            var mapId = _trans.GetMapId(ent);
+            var rot = _trans.GetWorldRotation(ent);
```
AJCM-git, line 175: use/pass the transform

AJCM-git (approved):
lgtm just this and the scoping

Robust.Server/Console/Commands/MapCommands.cs
Errant-4: Okay apparently I don't understand what you want, can you please explain

AJCM-git: get the transform of ent and then pass it directly to the system methods

PR #5963, opened 2025-05-24
predict TimedDespawnComponent

kosticia:
Predicted `SharedTimedDespawnSystem `and its component.
Deleted `TimedDespawnSystem` in the server and client content folders.
Sloth and slarti said that idea of the prediction is fine.
Required for the https://github.com/space-wizards/space-station-14/pull/37257


Robust.Shared/Spawners/SharedTimedDespawnSystem.cs
```diff
@@ -5,51 +5,36 @@
...
 
         var query = EntityQueryEnumerator<TimedDespawnComponent>();
-
         while (query.MoveNext(out var uid, out var comp))
         {
             comp.Lifetime -= frameTime;
-
-            if (!CanDelete(uid))
-                continue;
+            Dirty(uid, comp);
```
slarticodefast, line 27: This will network the remaining lifetime each tick.
Instead of accumulating `frameTime` I would add a `DespawnTime` datafield, which is a `TimeSpan` serving as a timestamp. `Lifetime` should be changed into a `TimeSpan` as well, but stay constant. Then compare `CurTime` with `DespawnTime` to see if the entity should be deleted.
One thing I'm not certain about is how to intitialize `DespawnTime` for purely clientside entities. Was `MapInitEvent` raised on the client? Maybe another maintainer can answer.

kosticia: Isn't it just will network timespans instead of floats? Also Lifetime timespan will create really a lot of conflicts.

slarticodefast: No, the idea is that the timespan marks the server time at which the entity will be deleted, instead of continuously decreasing a timer. So the variable will only have to be set once on spawn rather than being updated every tick. And the yaml serializer should simply convert the float into seconds for existing prototypes without any confilcts.

kosticia: > And the yaml serializer should simply convert the float into seconds for existing prototypes without any confilcts.

I mean content and RT conflicts. Lifetime used in ton of systems for some reason

slarticodefast: Hmm, yeah, would need some fixes I guess. But dirtying this every single tick is not good in my opinion.
Maybe @metalgearsloth can comment on this.

PR #5968, opened 2025-05-25
added new smallTag

Moomoobeef:
## About
The new small tag allows for smaller text in richtext, without using [font] (which isn't generally available to players anyway)

Similar to [head], it provides 3 levels of smaller text, with the smallest being 70% the size of regular text. I could go smaller but I don't necessarily want to allow players to make tiny unreadable text.

3 levels may also be redundant, I could lower the amount of levels to 2 but change the math so the smallest is still 2/3, but I don't know if I should. I am open to feedback on that.

## Why
sometimes small fonts are useful, for disclaimers, higher detailed graphics, and surely players will find more ways to use them than I can think of.

## Demonstration
Here is an ingame demonstration of what this looks like when used:

## Problems
Right now, [bold], [italic], and [bolditalic] do not work with this tag. it will overwrite bold/italic and always be regular font.
*I don't know how to fix this.* If someone knows how, please feel free to let me know (or to do it yourself), but I am already thoroughly in over my head to begin with.

## How it works
I basically just copied the existing HeadingTag, but made it make text smaller, and used different math. Header also forces bold, but usually you don't want tiny text to also be bold (its even harder to read!) so I also changed that.
The math that calculates the size is: `(int)Math.Ceiling(FontTag.DefaultSize - ((level*0.1)*FontTag.DefaultSize))`

## related ss14 PR
Here is the related ss14 pull request that requires this one https://github.com/space-wizards/space-station-14/pull/37811

### DIFF
--- a/Robust.Client/UserInterface/RichText/MarkupTagManager.cs
+++ b/Robust.Client/UserInterface/RichText/MarkupTagManager.cs
@@ -24,6 +24,7 @@ public sealed class MarkupTagManager
         new CommandLinkTag(),
         new FontTag(),
         new HeadingTag(),
+        new SmallTag(),
         new ItalicTag()
     }.ToDictionary(x => x.Name.ToLower(), x => x);
 
@@ -39,6 +40,7 @@ public sealed class MarkupTagManager
         typeof(CommandLinkTag),
         typeof(FontTag),
         typeof(HeadingTag),
+        typeof(SmallTag),
         typeof(ItalicTag)
     };
 
--- a/Robust.Client/UserInterface/RichText/SmallTag.cs
+++ b/Robust.Client/UserInterface/RichText/SmallTag.cs
@@ -1,0 +1,36 @@
+using System;
+using Robust.Client.ResourceManagement;
+using Robust.Shared.IoC;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Utility;
+
+namespace Robust.Client.UserInterface.RichText;
+
+public sealed class SmallTag : IMarkupTagHandler
+{
+    [Dependency] private readonly IResourceCache _resourceCache = default!;
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
+
+    public string Name => "small";
+
+    /// <inheritdoc/>
+    public void PushDrawContext(MarkupNode node, MarkupDrawingContext context)
+    {
+        if (!node.Value.TryGetLong(out var levelParam))
+            return;
+
+        var level = Math.Min(Math.Max((int)levelParam, 1), 3);
+        node.Attributes["size"] = new MarkupParameter(
+            (int)Math.Ceiling(FontTag.DefaultSize - ((level*0.1)*FontTag.DefaultSize))
+        );
+
+        var font = FontTag.CreateFont(context.Font, node, _resourceCache, _prototypeManager, FontTag.DefaultFont);
+        context.Font.Push(font);
+    }
+
+    /// <inheritdoc/>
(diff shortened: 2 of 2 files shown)


slarticodefast:
Would a more general `Size` tag be too abusable in the context of paper writing? Or is the code too different from the `Heading` tag to combine them?

Moomoobeef:
> Would a more general `Size` tag be too abusable in the context of paper writing? Or is the code too different from the `Heading` tag to combine them?

I had considered just raising the max header level to 6, the way the math is set up, 6 does end up smaller than normal text, but 4 and 5 are basically useless, since they're just uglier bold text.

I could do one unified size tag that does both functions, but that would make header redundant, and if I removed header it would break a *lot* of existing work. 

So I figured out would be most simple to just have this be a separate tag

Moomoobeef:
~~was reading through docs and looked at the conventions page and saw [this](https://docs.spacestation14.com/en/general-development/codebase-info/conventions.html#dont-copy-paste-code) (the do not copy paste code section)~~

~~Guilllltyyy. I need to come back and fix this, don't merge this until I atleast try to fix this.~~

Moomoobeef:
> was reading through docs and looked at the conventions page and saw [this](https://docs.spacestation14.com/en/general-development/codebase-info/conventions.html#dont-copy-paste-code) (the do not copy paste code section)
> 
> Guilllltyyy. I need to come back and fix this, don't merge this until I atleast try to fix this.

I fucking

I'm silly, this is the wrong PR, I meant to put this on my *other* engine PR. the one with the snapping mode.

This PR is fine

Moomoobeef:
actually shit now that I think about it this PR also violates "don't copy paste a shitload of code"

Moomoobeef:
so I have looked further at this, and what I figured I should do instead of blatantly copypasting from HeadingTag and just changing a few things was to instead derive from HeadingTag and override just what I need to. Two big problems.

- HeadingTag is sealed and cannot be derived from. I *probably* should not change that
- I pretty much have to override almost everything anyway. So this is mostly pointless

Sooo... This is the part where I ask someone for a second opinion. Is this code as it is rn fine and I should just stop worrying about it, or is there some way that I should do this better. Because I was the one who initially decided that this needed to be changed, it's not like someone else told me it was bad.

Moomoobeef:
I was working on something unrelated when I came across Content.Client/UserInterface/RichText, and it looks like I probably could have implemented this there instead of having it be in the engine. Should I do that instead? looking for some advice here.

PJB3005:
Can't this be implemented in content?

slarticodefast:
> Can't this be implemented in content?

Wouldn't it be weird to have small text in content but large text in the engine?
Unless you want to move both there.

PJB3005:
Reviewing this PR gave me too many questions about the exact values in use in the PR. I came to the conclusion that this is not practical to generically implement in engine without making the entire markup system *was* more flexible.

Moomoobeef:
I would love to refactor markup but that's a bit outside the scope of my knowledge, so someone else will probably have to do that...

Or maybe I'll come back and do it after I have a few more months of C# practice under my belt

yagwog:
> I would love to refactor markup

please oh god please hahahah

I'm trying to get a tag working for a fillable [form] tag and the amount of errors markup is spitting out from markup and tag handlers is killing me

Moomoobeef:
> > I would love to refactor markup
> 
> please oh god please hahahah
> 
> I'm trying to get a tag working for a fillable [form] tag and the amount of errors markup is spitting out from markup and tag handlers is killing me

My condolences. Unfortunately I don't have time to work on a markup refactor. Best of luck on your form tag though, that's something I've been wanting for a long time.

PR #5982, opened 2025-05-28
Remove `ITileDefinition.ID`

Tayrtahn:
`ITileDefinition` implements `IPrototype`, so it already has an identical `ID` field.

Fixes a warning CS0108 for [PZW](https://github.com/space-wizards/space-station-14/issues/33279).


slarticodefast (approved):
Looks good to me!

PR #5985, opened 2025-05-29
Fix static grid center of mass

### LINKED ISSUE #5984: Station grids have center of mass 0,0
Easy to notice because the station maps use it to recenter.
If you check the `PhysicsComponent` then `LocalCenter` is `0,0`.

Image

Image

slarticodefast:
Fixes https://github.com/space-wizards/RobustToolbox/issues/5984
Happens for stations that have the station anchor enabled.
The previous calculation was exiting early, assuming that `LocalCenter` is not needed for static grids.
However, this caused a range of bugs in the content repo, for example the station maps not being centered correctly, grids not getting an atmosphere above the size threshold since `MassDataChangedEvent` was not raised, and probably some shuttle collision inaccuracies.

Since the `BodyType` now no longer affects the result of `ResetMassData` (it is only used for adjusting the velocity to conserve momentum) we can remove the `ResetMassData` call from `SetBodyType`.

### Changelog
- Fixed the LocalCenter (center of mass) being zero for static grids.
- `MassDataChangedEvent` is now always raised when `ResetMassData` is called, even for static bodies.

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
+++ b/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
@@ -312,13 +312,6 @@ public partial class SharedPhysicsSystem
             body._inertia += data.I;
         }
 
-        // Update this after re-calculating mass as content may want to use the sum of fixture masses instead.
-        if (((int) body.BodyType & (int) (BodyType.Kinematic | BodyType.Static)) != 0)
-        {
-            body._localCenter = Vector2.Zero;
-            return;
-        }
-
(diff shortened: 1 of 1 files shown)


slarticodefast:
@metalgearsloth Also doesn't the mass, inertia, local center etc need to be dirtied here?
Or is all that not networked at all?

metalgearsloth:
> @metalgearsloth Also doesn't the mass, inertia, local center etc need to be dirtied here? Or is all that not networked at all?

Everything else should be able to be derived from fixtures, the only reason you'd network it is for debugging sanity.

PR #5986, opened 2025-05-29
Fix all warnings from `Dispose` on Controls

Tayrtahn:
Replaced all uses and overrides of `Dispose` on UI controls.

- `Dispose` was replaced with `Close` for windows or `Orphan` for controls.
- `Dispose` overrides were swapped for `ExitedTree` overrides.

This felt way too easy. There's probably something wrong here, so please test and review this carefully!


Robust.Client/UserInterface/Control.cs
```diff
@@ -639,7 +639,7 @@ public void DisposeAllChildren()
             var children = new List<Control>(Children);
             foreach (var child in children)
             {
-                child.Dispose();
+                child.Orphan();
```
ElectroJr: The method should probably just be marked as obsolete instead of making `DisposeAllChildren` a duplicate of `RemoveAllChildren()`

PR #5988, opened 2025-05-30
remove ResetMassData call from SetFixedRotation

slarticodefast:
Doesn't look like it's needed there? It recalculates the fixture mass, center of mass and inertia, so we only need to call it when the fixtures change.
`SetFixedRotation` only sets the angular velocity, so no need to call it.


slarticodefast:
Hmm, `TestPointLinearImpulse` failed, might be needed after all for some weird reason. Or something is just missing a `FixturesUpdate` call. I'll investigate.

slarticodefast:
Nevermind, `ResetMassData` sets the inertia to zero when `FixedRotation` is true, so it does need to be recalculated.

PR #5989, opened 2025-06-01
Fix rich text entry measure

ThereDrD0:
Replaced `Width` with `maxSizeX` when calculating the space required for `RichTextEntry`

The problem was found when I tried to add `Control` in chat on my fork to render job icons near character name

A `Label` is placed next to the character's name, showing their job title in the color of the corresponding department.

Without fix:
The text overflows the chat boundaries because the label's size is not taken into account.

image

With fix:
Everything renders correctly.

Screenshot 2025-06-01 152029

### DIFF
--- a/Robust.Client/UserInterface/RichTextEntry.cs
+++ b/Robust.Client/UserInterface/RichTextEntry.cs
@@ -141,7 +141,7 @@ namespace Robust.Client.UserInterface
                 if (Controls == null || !Controls.TryGetValue(nodeIndex, out var control))
                     continue;
 
-                control.Measure(new Vector2(Width, Height));
+                control.Measure(new Vector2(maxSizeX, Height));
 
                 var desiredSize = control.DesiredPixelSize;
                 var controlMetrics = new CharMetrics(


iDesmond:
@DrSmugleaf when will it be possible to review this?
This is very important...

PJB3005:
> DrSmugleaf when will it be possible to review this? This is very important...

Do not ask random maintainers for reviews directly.

iDesmond:
> Do not ask random maintainers for reviews directly.

I'm sorry.. I just don't know who to ask

ThereDrD0:
Some extra context about what's going on with this fix:

I’ve created a custom `MarkupTag` that generates a `Label` (e.g. the `[Капитан]` in my screenshot). This label is inserted before the character’s name when they speak over radio, to show their job title in color.

The issue is that this `Label` is a UI `Control`, and its size wasn't being considered when measuring the space for the message inside `RichTextEntry`. As a result, the total text width could exceed the chat boundaries, which caused it to overflow horizontally (see the first screenshot).

This happens because `control.Measure()` was using `Width` (i.e. the full width of the `RichTextEntry`) instead of `maxSizeX`, which represents the actually available horizontal space for the current content block. Replacing `Width` with `maxSizeX` fixes the problem.

Let me know if I misunderstood how `maxSizeX` is supposed to be used - but in this case, it does the job.

Robust.Client/UserInterface/RichTextEntry.cs
```diff
@@ -141,7 +141,7 @@ public RichTextEntry Update(MarkupTagManager tagManager, Font defaultFont, float
                 if (Controls == null || !Controls.TryGetValue(nodeIndex, out var control))
                     continue;
 
-                control.Measure(new Vector2(Width, Height));
+                control.Measure(new Vector2(maxSizeX, Height));
```
PJB3005, line 144: This `Height` argument doesn't make sense either, does it?

ThereDrD0: Yes, the height of any inner сontrol is also not included in `Height` - it's just the height of the RichTextEntry itself.  
This should probably be replaced with something else too, but I’m not sure how to get the total height that includes all contained controls.  
My PR only fixes the incorrect width calculation

PJB3005 (approved):

ArtisticRoomba:
This may resolve https://github.com/space-wizards/space-station-14/issues/34123.

PR #5990, opened 2025-06-01
add RSI.StateId Argument

ghost:
Added new method to `RSI.StateId` `Argument` that takes anything and uses that to construct new `StateId` using the old one as a prefix.

# why

I seen that in Content.Client there is no consistency in constructing `StateId`s using prefixes and values, as sometimes they use `-` separator, and sometimes not. Also, `StateId` exists, but its rarely used, and in context of such generated ids based on prefix, `string?` is used most of the times.

So I added this method to ~~break all StateIds that dont use `-`~~ normalize the codebase.

# implementation

As of now, the new `Argument` method lives in `StateId` and returns `StateId` meaning you could use any state id prefix as standalone `StateId`. This may not be an optimal solution as putting it into some new `StateIdPrefix` might be better for type safety, but I ~~was lazy~~ decided to minimize number of required changes to implement it if PR get merged, also Im not sure if there isnt RSI where prefix alone would be valid.

### DIFF
--- a/Robust.Client/Graphics/RSI/RSI.cs
+++ b/Robust.Client/Graphics/RSI/RSI.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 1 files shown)


PJB3005:
Yeah I think this should just be extension methods in content instead.

PR #5993, opened 2025-06-02
fix the lifestage checks on predicted entity deletion

ruddygreat:
Fixes an issue where `ClientEntityManager.PredictedQueueDeleteEntity()` & `ClientEntityManager.PredictedDeleteEntity()` will log an error if the entity is already being deleted by something else, as they only check if the entity has been deleted while `SharedTransformSystem.DetachEntityInternal()` checks if the entity is terminating.

Currently only happens when the client closes so it won't cause any major issues, but can cause unrelated test fails.


metalgearsloth:
This also happens with regular deletion I think but fine with this.

PR #5987, opened 2025-05-30
added an option for snapping directional hallway signs

Moomoobeef:
## about
I added a new snapping option that allows for consistent placement of directional hallway signs
https://github.com/user-attachments/assets/5671a523-c2aa-46b5-8a46-f0601e357715

## Technical details
I basically copied AlignWallProper but changed the offsets. That's it. I also added it to the list.

There is a related PR on the ss14 repo to also update sandbox.txt https://github.com/space-wizards/space-station-14/pull/37967

PR #5994, opened 2025-06-02
Un-hardcode `EntitySpawnWindow`'s placement mode dropdown 

Tayrtahn:
https://github.com/space-wizards/RobustToolbox/pull/5987 tried to add a new placement mode for wall signs, but was rejected because it's too specific to SS14 to be added to the engine. Unfortunately, there doesn't seem to be a way for content to add placement modes! Notably, the engine class `EntitySpawnWindow` has a hardcoded array of mode names.

`PlacementManager` already has a dictionary of all Types that inherit from `PlacementMode`. This PR removes the hardcoded array from `EntitySpawnWindow` and has it use a new public property of `IPlacementManager` to get the list of modes.

One downside of switching from the hardcoded list to a reflection-based one is that the order that modes are listed in the dropdown is going to be different. I checked with the #Cartography Discord channel, and mappers seem to be okay with this. The list is now in alphabetical order.

The hardcoded array of mode names in `EntitySpawnWindow` was previously a public static field, and a few content classes established some funky coupling based around that fact - they construct their own selector by referencing the values from `EntitySpawnWindow`. These should be changed to use the new `AllModeNames` property of `IPlacementManager`. To avoid breaking existing code, `EntitySpawnWindow` was given a replacement `InitOpts` property which does an IoC resolve and returns `PlacementManager.AllModeNames`. This was marked as obsolete with a message encouraging use of `IPlacementManager.AllModeNames`.

Note there is still plenty of weirdness in this code, and `PlacementManager` could stand to be significantly rewritten. But until that day comes, this is a nice feature to have.


Moomoobeef:
I had not yet looked into this but I'm glad that you did because I definitely wouldn't have figured this out, so thank you ♥️

deltanedas:
shouldnt this just be a prototype with a field

LaCumbiaDelCoronavirus:
doesn't the RCD already do this

Tayrtahn:
> doesn't the RCD already do this

It might be able to use different placement modes (I'm not sure), but it can't add them to the entity spawn window - that uses a hardcoded list of modes defined in the engine.

> shouldnt this just be a prototype with a field

I'm not sure what you mean by this, but I'm interested in hearing more.

Edit: Oh, like a `PlacementModePrototype` that has the mode name as a field, which placement manager uses to find the appropriate class? I suppose that could work too, but it seems a bit clunkier having to define a new prototype in YAML for each class.

PJB3005 (changes requested):
This entire solution is overengineered and will need to be torn out anyways. Just make an API on `IPlacementManager` to add a new type, and add it at the bottom of the list.

Tayrtahn:
Alright, completely redone in a less over-engineered way. I'll update the PR description to reflect the new version.

This doesn't follow the suggested changes exactly because the nature of the entity spawn window doesn't really make that possible. It's considerably simpler though.

Tayrtahn:
Modified the title to reflect the fact that Content actually already can add placement modes, they're just not displayed in the Entity Spawn Window's dropdown menu and can only be accessed if set as the default placement mode for an entity - this is how atmos pipes have a custom placement mode.

This PR just un-hardcodes the list of modes shown in the dropdown.

PR #5998, opened 2025-06-05
Add ignoredcomps to IsDefault

metalgearsloth:
Useful for randomised comps we want to ignore (randomsprite).


Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -161,7 +161,7 @@ public virtual void Initialize()
         /// <summary>
         /// Returns true if the entity's data (apart from transform) is default.
```
Tayrtahn, line 162: Probably wanna update the docs

PR #5999, opened 2025-06-05
Add No Audio Displaymode/commandline argument

Aidenkrz:
Pretty simple, just a displaymode that uses the headless audio manager.


PJB3005 (changes requested):
Yeah I don't want to extend the `DisplayMode` enum further. Make it an enum.

PJB3005:
Closing given lack of response.

PR #6000, opened 2025-06-06
Adjust how OpenAL extensions are requested

perryprog:
This fixes a few things:
1. ~~Make Clyde stop trying to set a window icon when it doesn't have any. This can happen on macOS, and just leads to a random error getting logged and nothing else.~~ Moved into #6016.
2. Fixes an OpenAL error that was caused by attempting to query extensions with a null device. I _think_ this is an error that only comes on on macOS's ancient OpenAL implementation, as when I switch over to using OpenAL Soft (no comment on the pains it took to get that working, though I'll be making an issue about letting macOS builds use it in a moment) I end up without any issue and the same set of extensions registered.
3. Shouldn't ever matter, but make sure we don't put in empty strings onto _alContextExtensions.

Here's the error in question that I was getting:
```
[ERRO] clyde.oal: [LoadAudioOggVorbis:133] AL error: InvalidName
```
Note the LoadAudioOggVorbis method name is a red herring; it is caused by the GetString, there just isn't an error check until there.

Also PR number 6000 :tada:.

### DIFF
--- a/Robust.Client/Audio/AudioManager.cs
+++ b/Robust.Client/Audio/AudioManager.cs
@@ -57,8 +57,8 @@ internal sealed partial class AudioManager : IAudioInternal
         _checkAlError();
 
         // Load up AL context extensions.
-        var s = ALC.GetString(ALDevice.Null, AlcGetString.Extensions) ?? "";
-        foreach (var extension in s.Split(' '))
+        var s = ALC.GetString(_openALDevice, AlcGetString.Extensions) ?? "";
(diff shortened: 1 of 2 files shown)


metalgearsloth:
Would prefer this to be split for merge reasons + I'm unsure on the icon change but fine with the audio one.

perryprog:
Sure, can split out the icon one later today if wanted—that one is literally just to avoid the console error log, to be clear.

PR #6003, opened 2025-06-06
Add a basic loading screen!

beck-thompson:
Its kind of difficult to tell if your game is actually loading, it sometimes can take around 7-8 seconds to fully load and just having splash screen without any feedback is not super fun or informative to look at!

https://github.com/user-attachments/assets/91cad5c1-80c5-4d9d-9b98-69d28b1a8c42

Servers can also change a few settings about its location and stuff the issue is that I had to do it in a very weird way and I'm not sure its correct


Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,174 @@
...
+
+    public void EndLoadingSection()
+    {
+        var time = _sw.Elapsed;
+        Times.Add((CurrentSectionName, time));
+        _currentSection++;
+    }
+
+    /// <summary>
+    /// Start a specific
```
Tayrtahn: Well? Don't leave us hanging!

beck-thompson: Added it! I'm still not really happy with the function names - they are kind of odd but I don't know what to name em :laughing:

If you have any better ideas let me know!

beck-thompson:
OK now its "fully working" - servers can add their own custom settings. The main issue is that the code is very suspect and I'm unsure what the correct approach is :thinking:

Robust.Shared/CVars.cs
```diff
@@ -1187,6 +1187,28 @@ protected CVars()
...
+            CVarDef.Create("display.show_current_loading_section", true, CVar.CLIENTONLY);
+
+        public static readonly CVarDef<bool> DisplayShowLoadTimes =
+            CVarDef.Create("display.show_load_times", true, CVar.CLIENTONLY);
+
+        /// <summary>
+        /// The client will automatically update this to the number of sections it saw. Ths is for the loading bar on
+        /// the splash screen.
+        /// </summary>
+        public static readonly CVarDef<int> SeenNumberOfLoadingSections =
```
PJB3005: This shouldn't be a CVar.

Robust.Shared/CVars.cs
```diff
@@ -1187,6 +1187,28 @@ protected CVars()
         public static readonly CVarDef<string> DisplaySplashLogo =
             CVarDef.Create("display.splash_logo", "", CVar.CLIENTONLY);
 
+        public static readonly CVarDef<float> DisplayLoadingXLocation =
+            CVarDef.Create("display.loading_x_location", 0.5f, CVar.CLIENTONLY);
+
+        public static readonly CVarDef<float> DisplayLoadingYLocation =
+            CVarDef.Create("display.loading_y_location", 0.675f, CVar.CLIENTONLY);
```
PJB3005: These shouldn't be CVars.

Robust.Shared/ContentPack/ResourceManifestData.cs
```diff
@@ -11,12 +11,17 @@ internal sealed record ResourceManifestData(
     string? DefaultWindowTitle,
     string? WindowIconSet,
     string? SplashLogo,
+    float? LoadingXLocation,
+    float? LoadingYLocation,
```
PJB3005: This is too inflexible of a system. If we want fancy splash screen layout, rewrite the system to be more flexible.

Robust.Shared/ContentPack/ResourceManifestData.cs
```diff
@@ -11,12 +11,17 @@ internal sealed record ResourceManifestData(
...
+    bool? ShowLoadingBar,
+    bool? ShowCurrentSection,
+    bool? ShowLoadTimes,
```
PJB3005: This should be a debugging option, not one configured by content.

Robust.Shared/Configuration/IConfigurationManager.cs
```diff
@@ -58,6 +58,11 @@ internal CVarChangeInfo(string name, GameTick tickChanged, object newValue, obje
     /// </remarks>
     public interface IConfigurationManager
     {
+        /// <summary>
+        /// Checks if a configuration file has been loaded.
+        /// </summary>
+        bool HasLoadedConfiguration();
```
PJB3005: This API does not make sense with how the config system works. Remove it.

Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,214 @@
...
+        _clyde.Render();
+    }
+
+    /// <summary>
+    /// Start a loading bar "section" for the given method.
+    /// Must be ended with EndSection.
+    /// </summary>
+    public void BeginLoadingSection(object method)
+    {
+        BeginLoadingSection(method.GetType().Name);
```
PJB3005, line 143: Do not show class names to users.

Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,214 @@
...
+        _times.Add((_currentSectionName, time));
+        _currentSection++;
+    }
+
+    /// <summary>
+    /// Will run the giving function and add a custom "section" for it on the loading screen.
+    /// </summary>
+    public void LoadingStep(Action action, object method)
+    {
+        BeginLoadingSection(method.GetType().Name);
```
PJB3005: Do not show class names to users.

Robust.Client/GameController/GameController.cs
```diff
@@ -177,41 +195,47 @@ internal bool StartupContinue(DisplayMode displayMode)
                 _configurationManager,
                 _logManager.GetSawmill("res"));
 
-            _resourceCache.PreloadTextures();
+            _loadscr.LoadingStep(_resourceCache.PreloadTextures, _resourceCache);
             _networkManager.Initialize(false);
```
PJB3005: Why are some of these things in steps, but not others?

Robust.Client/Graphics/IClydeInternal.cs
```diff
@@ -16,6 +16,7 @@ internal interface IClydeInternal : IClyde, IClipboardManager
         void Render();
         void FrameProcess(FrameEventArgs eventArgs);
         void ProcessInput(FrameEventArgs frameEventArgs);
+        void ProcessInput();
```
PJB3005: Don't add this.

Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,214 @@
...
+        SeenNumberOfLoadingSections = _cfg.GetCVar(CVars.SeenNumberOfLoadingSections);
+
+        LoadingBarMaxSections = SeenNumberOfLoadingSections == 0 ? LoadingBarMaxSections : SeenNumberOfLoadingSections;
+
+        _resourceCache.TryGetResource("/EngineFonts/NotoSans/NotoSans-Regular.ttf", out _font);
+
+        _loadingBarColor = Random.Shared.Pick(Random.Shared.NextFloat() < RareColorChance ? ColorsRare : ColorsStandard);
+    }
+
+    public void BeginLoadingSection(string sectionName)
```
PJB3005: This needs to guard against entering a single section twice.

Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,214 @@
...
+
+    private FontResource? _font;
+
+    private Color _loadingBarColor;
+
+    private readonly List<(string Name, TimeSpan LoadTime)> _times = [];
+
+    private int _currentSection = 1;
+
+    private string _currentSectionName = "";
```
PJB3005: This should be nullable and default to null.

Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,214 @@
...
+using Robust.Shared.Maths;
+using Robust.Shared.Random;
+using Robust.Shared.Timing;
+
+namespace Robust.Client.Graphics;
+
+/// <summary>
+/// Manager that creates and displays a basic splash screen and loading bar.
+/// </summary>
+public sealed partial class LoadingScreenManager
```
PJB3005: All APIs on this class should be locked out once loading is complete.

Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,214 @@
...
+        ShowLoadingBar = _cfg.GetCVar(CVars.DisplayShowLoadingBar);
+        ShowCurrentLoadingSection = _cfg.GetCVar(CVars.DisplayShowCurrentLoadingSection);
+        ShowLoadTimes = _cfg.GetCVar(CVars.DisplayShowLoadTimes);
+        SeenNumberOfLoadingSections = _cfg.GetCVar(CVars.SeenNumberOfLoadingSections);
+
+        LoadingBarMaxSections = SeenNumberOfLoadingSections == 0 ? LoadingBarMaxSections : SeenNumberOfLoadingSections;
+
+        _resourceCache.TryGetResource("/EngineFonts/NotoSans/NotoSans-Regular.ttf", out _font);
+
+        _loadingBarColor = Random.Shared.Pick(Random.Shared.NextFloat() < RareColorChance ? ColorsRare : ColorsStandard);
```
PJB3005: Not a fan of the loading bar color being able to change like that. It just feels confusing.

Robust.Shared/CVars.cs
```diff
@@ -1187,6 +1187,28 @@ protected CVars()
...
+
+        public static readonly CVarDef<bool> DisplayShowLoadingBar =
+            CVarDef.Create("display.show_loading_bar", true, CVar.CLIENTONLY);
+
+        public static readonly CVarDef<bool> DisplayShowCurrentLoadingSection =
+            CVarDef.Create("display.show_current_loading_section", true, CVar.CLIENTONLY);
+
+        public static readonly CVarDef<bool> DisplayShowLoadTimes =
+            CVarDef.Create("display.show_load_times", true, CVar.CLIENTONLY);
```
PJB3005: This should be disabled by default outside dev builds.

Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        SplashLogo = _cfg.GetCVar(CVars.DisplaySplashLogo);
+        LoadingXLocation = _cfg.GetCVar(CVars.DisplayLoadingXLocation);
+        LoadingYLocation = _cfg.GetCVar(CVars.DisplayLoadingYLocation);
+        ShowLoadingBar = _cfg.GetCVar(CVars.DisplayShowLoadingBar);
+        ShowCurrentLoadingSection = _cfg.GetCVar(CVars.DisplayShowCurrentLoadingSection);
+        ShowLoadTimes = _cfg.GetCVar(CVars.DisplayShowLoadTimes);
+        SeenNumberOfLoadingSections = _cfg.GetCVar(CVars.SeenNumberOfLoadingSections);
+
+        LoadingBarMaxSections = SeenNumberOfLoadingSections == 0 ? LoadingBarMaxSections : SeenNumberOfLoadingSections;
```
PJB3005: In what scenario would `SeenNumberOfLoadingSections` ever not be zero here?

Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,214 @@
...
+        var offset = 0;
+        var x = 0;
+        _times.Sort((a, b) => b.LoadTime.CompareTo(a.LoadTime));
+        foreach (var val in _times)
+        {
+            if (x >= NumLongestLoadTimes)
+                break;
+
+            var time = val.LoadTime.ToString(@"ss\.ff");
+            var entry = $"{time} - {val.Name}";
```
PJB3005: Specify the format for `val.LoadTime` in the format string instead.

Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,214 @@
...
+            handle.DrawingHandleScreen.DrawString(new VectorFont(_font, 10), startLocation + new Vector2i(0, offset), entry);
+            offset += 13;
```
PJB3005: Remove magic number.

Robust.Client/Graphics/LoadingScreenManager.cs
```diff
@@ -0,0 +1,214 @@
...
+        var x = 0;
+        _times.Sort((a, b) => b.LoadTime.CompareTo(a.LoadTime));
+        foreach (var val in _times)
+        {
+            if (x >= NumLongestLoadTimes)
+                break;
+
+            var time = val.LoadTime.ToString(@"ss\.ff");
+            var entry = $"{time} - {val.Name}";
+            handle.DrawingHandleScreen.DrawString(new VectorFont(_font, 10), startLocation + new Vector2i(0, offset), entry);
```
PJB3005: Don't just allocate this `VectorFont` every line.

PJB3005 (changes requested):
It feels inflexible that this system can't be used by content to report its own loading status.

You need to disable vsync on Clyde while loading is happening, else the loading screen steps will slow down startup. I would also like to see a proper profile of how much CPU time is spent redrawing the screen during a load now.

The drawing code needs to be DPI-aware.

Robust.Shared/CVars.cs
beck-thompson: How would you recommend approaching this? I couldn't think of a way to easily determine the number of sections without just going through once :thinking:.

My only other idea was to allow a set number of sections and when the loading screen is finished it would verify that that many sections were completed and if they weren't give out an error.

Robust.Client/Graphics/LoadingScreenManager.cs
beck-thompson: What do you think I should do? The main issue is they need to be localized which is hard because the localization manager isn't really loaded :thinking:

Would just putting the localization high up in the boot process and writing custom names for each step seem good?

PJB3005: > What do you think I should do? The main issue is they need to be localized which is hard because the localization manager isn't really loaded 🤔

Well you either make it only shown on dev, or you refactor how the localization system works so it can be used to split loading. Take your pick.

beck-thompson:
> It feels inflexible that this system can't be used by content to report its own loading status.

I agree that it feels like it should allow more but I'm unsure exactly what that layout would look like! I pinged you on discord about this as well.

> You need to disable vsync on Clyde while loading is happening, else the loading screen steps will slow down startup. I would also like to see a proper profile of how much CPU time is spent redrawing the screen during a load now.

Tried doing this pretty sure it works! Its a little weird because I don't want to override the clients cvar.

> The drawing code needs to be DPI-aware.

I do not know how to do this at all :laughing:! After looking around it seems like `DefaultWindowScale` in clyde might be what I am looking for? The issue I have is I'm unsure how to scale stuff like fonts - they only allow an overall scale, no x y axis scale.

Updated the colors so its just randomly selects a random color from the list, if you think that still is a little odd using just white only looks fine so let me know and I can change it.

Robust.Shared/CVars.cs
beck-thompson: Update: I just did this method let me know if its better or worse

TornadoTechnology:
Soo goood! The best thing that could be added to this is providing access to the loading screen in the modloader so that projects can change it, and not just png. But as I remember it has architectural problems. But it really lacks smooth transitions, through shading or something like that.

And so you did everything awesome.

beck-thompson:
Fixed a few issues and made it a a decent amount cleaner. I'm not really sure how to do profiles correctly, but from what I can tell it doesn't really have any impact at all (I used PrefView):

image

I ran it a lot of times and it basically would only barley show up (if if showed up at all). Let me know if that's what your looking for

PJB3005 (approved):

PR #6007, opened 2025-06-07
Fully obsolete some old SharedContainerSystem methods

slarticodefast:
Needed for https://github.com/space-wizards/space-station-14/issues/33279
The remaining use cases in content will automatically cast into the `Entity<T>` overload, removing the warnings.

Also part of https://github.com/space-wizards/RobustToolbox/pull/5675 but the review requested it to be atomized.

#### Changelog
- Fully removed the obsolete overloads of IsInSameOrNoContainer, IsInSameOrParentContainer and IsInSameOrTransparentContainer. Use the `Entity<T>` variant instead.


Tayrtahn (approved):
Save us from the madness of `(ent.Owner, null)`

PR #6010, opened 2025-06-07
Change a bunch of static `Loc.GetString` calls to be properly resolved

Tayrtahn:
The static `Loc.GetString` method isn't great, since it's hiding an IoC resolve on each use - if a proper reference to the `ILocalizationManager` instance is available, it's better to use that.

`EntitySystem` already handles this nicely with its `Loc` field, which a ton of systems use (probably without even realizing that they're using it instead of the static method). `LocalizedCommands` has a reference to `ILocalizationManager`... but it's named `LocalizationManager` and it looks like there are several hundred instances of commands not being aware of it and using the static `Loc.GetString` instead.

This PR adds a `Loc` property to `LocalizedCommands` which simply returns the `LocalizationManager` field (essentially aliasing it - I'd be amazed if the compiler didn't just inline this). Thanks to the way that scoping works, this converts every instance of the static `Loc.GetString` in a `LocalizedCommands` inheritor to use the properly pre-resolved reference.

Edit: also needed to change some static methods that were using `Loc` so they pass in the reference.

Requires https://github.com/space-wizards/space-station-14/pull/38133 for a content fix.

## Breaking changes
`LocalizedCommands` now has a `Loc` property which returns the `ILocalizationManager` reference. Any classes that extend `LocalizedCommands` and explicitly need to use static methods of the `Loc` class will need to be modified to specify the full namespace, as `Loc` now refers to this new property instead.


VerinSenpai:
Hopefully I remember to come look at this later

Robust.Shared/Console/LocalizedCommands.cs
```diff
@@ -10,6 +10,8 @@ public abstract class LocalizedCommands : IConsoleCommand
 {
     [Dependency] protected readonly ILocalizationManager LocalizationManager = default!;
 
+    public ILocalizationManager Loc => LocalizationManager;
```
PJB3005, line 13: This needs to be documented as a breaking change.

Tayrtahn: Added a breaking changes section to the PR description.

PR #6021, opened 2025-06-14
Add a new toolshed command `spawn:in`

### LINKED ISSUE #6019: spawn:in command.
add a new spawn:in toolshed command that is like spawn:on but instead of spawning it ontop of where a entity is. it attempts to (via it's `ContainerManagerComponent` add it to a specific container. this can be used to make it so you can run a command and do custom outfits or put items into boxes/sheaths without having to manually do it. (I plan on doing this myself just making this issue so it is clear.)

walksanatora:
fixes #6019

while I was working in event staff on starlight I realized how usefull this would be for eg: adding items to a person for custom outfits or on the fly container fills.

https://github.com/user-attachments/assets/cdb3f194-78d1-40d4-be35-720c9848e15d

### DIFF
--- a/Resources/Locale/en-US/toolshed-commands.ftl
+++ b/Resources/Locale/en-US/toolshed-commands.ftl
@@ -195,6 +195,8 @@ command-description-spawn-at =
     Spawns an entity at the given coordinates.
 command-description-spawn-on =
     Spawns an entity on the given entity, at it's coordinates.
+command-description-spawn-in =
+    Spawns an entity in the given container on the given entity, dropping it at it's coordinates if it doesn't fit
 command-description-spawn-attached =
     Spawns an entity attached to the given entity, at (0 0) relative to it.
 command-description-mappos =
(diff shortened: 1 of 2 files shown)


walksanatora:
https://github.com/user-attachments/assets/0cd5a27b-9e82-4971-9ae2-e76431d5cec5

another video now that I added the ftl.

walksanatora:
oh wait... I just noticed this solves a point in https://github.com/space-wizards/space-station-14/issues/35493

Resources/Locale/en-US/toolshed-commands.ftl
```diff
@@ -195,6 +195,8 @@ command-description-spawn-at =
     Spawns an entity at the given coordinates.
 command-description-spawn-on =
     Spawns an entity on the given entity, at it's coordinates.
+command-description-spawn-in =
+    Spawns an entity in the given container on the given entity, dropping it at it's coordinates if it doesn't fit
```
PJB3005:
```suggestion
    Spawns an entity in the given container on the given entity, dropping it at its coordinates if it doesn't fit
```

PR #6023, opened 2025-06-15
add: VV boolean fields gray when readonly

qwerltaz:
In the view variables menu, made boolean fields gray when read-only to make them not identical to read-write boolean fields.

image


PJB3005 (changes requested):
This is not a good way to do this. This should be part of the styling for the checkbox control

PJB3005 (changes requested):
When I say "styling" I mean "stylesheet". This is still just hardcoded into the engine and disregards the content stylesheet. This isn't acceptable.

PJB3005 (changes requested):
Using modulate is not an appropriate way to change the color of the control

PR #6024, opened 2025-06-16
Changed SpriteComponent layer keys to be objects

Citrea-Lingua:
The layer/displacement system uses objects as keys throughout, except in this one place, which is causing issues when one tries to apply a displacement to a layer using a HumanoidVisualLayers-object as a key, which - as far as I can tell - is the entire reason the system got switched from String to Object in the first place.

Goes together with sister PR on the non-engine side of things: https://github.com/space-wizards/space-station-14/pull/38350


Citrea-Lingua:
Things broke, must have messed up my push, will be back when I fix that and do it properly - sorry for the mess `:D

PR #6025, opened 2025-06-16
changes SpriteComponent to use object as LayerKey

Citrea-Lingua:
The layer/displacement system uses objects as keys throughout, except in this one place, which is causing issues when one tries to apply a displacement to a layer using a HumanoidVisualLayers-object as a key, which - as far as I can tell - is the entire reason the system got switched from String to Object in the first place.

Goes together with sister PR on the non-engine side of things: 
Requires https://github.com/space-wizards/space-station-14/pull/38351

### DIFF
--- a/Robust.Shared/GameObjects/Components/Renderable/SpriteLayerData.cs
+++ b/Robust.Shared/GameObjects/Components/Renderable/SpriteLayerData.cs
@@ -60,7 +60,7 @@ public sealed partial class PrototypeCopyToShaderParameters
     /// <summary>
     /// The map key of the layer that will have its shader modified.
     /// </summary>
-    [DataField(required: true)] public string LayerKey;
+    [DataField(required: true)] public object? LayerKey;
 
     /// <summary>
     /// The name of the shader parameter that will receive the actual selected texture.
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -615,7 +615,7 @@ namespace Robust.Client.GameObjects
 
             if (layerDatum.CopyToShaderParameters is { } copyParameters)
             {
-                layer.CopyToShaderParameters = new CopyToShaderParameters(ParseKey(copyParameters.LayerKey))
+                layer.CopyToShaderParameters = new CopyToShaderParameters(copyParameters.LayerKey ?? "this is impossible")
                 {
                     ParameterTexture = copyParameters.ParameterTexture,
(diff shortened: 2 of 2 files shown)


Robust.Shared/GameObjects/Components/Renderable/SpriteLayerData.cs
```diff
@@ -60,7 +60,7 @@ public sealed partial class PrototypeCopyToShaderParameters
     /// <summary>
     /// The map key of the layer that will have its shader modified.
     /// </summary>
-    [DataField(required: true)] public string LayerKey;
+    [DataField(required: true)] public object? LayerKey;
```
PJB3005, line 63: Why are you making this nullable?

PJB3005 (changes requested):

Robust.Shared/GameObjects/Components/Renderable/SpriteLayerData.cs
Citrea-Lingua: It's required to compile. Ever assigning null here *should* be impossible, so I am pretty certain it would be fine to set as non-nullable (had it like that initially, actually, as I find setting types which don't actually expect to ever have null as nullable just... confusing, and possibly error-prone) - but the build process gets angry with me if I try that (looks like the serialisation generator *really* does not like it?) `:D
There is of course always a possibility I am missing larger implications here, which a developer more seasoned with the codebase would catch. Fully agree that this shouldn't be nullable, just haven't found a way to not have it so.
I suspect that was not an issue before, because string natively handles null as an empty string in most cases IIRC.

ElectroJr:
I'm going to look into replacing the sprite key `object` methods & fields with a simple union/ struct thats contains either an enum or string along with a yaml serializer, which should make handling keys defined via data fields easier to use.  As that will probably also involve breaking engine changes that need a content PR, I'd rather just fold this PR into that one.

Citrea-Lingua:
> I'm going to look into replacing the sprite key `object` methods & fields with a simple union/ struct thats contains either an enum or string along with a yaml serializer, which should make handling keys defined via data fields easier to use. As that will probably also involve breaking engine changes that need a content PR, I'd rather just fold this PR into that one.

No objections - as long as this bit of seemingly unfinished work is done, I don't much care how ^_^

Citrea-Lingua:
closing, as it's been superceded and I frankly do not have the time or spoons to do anything more with it. The superceding PR for this has been sitting dormant for nearly as long as this one was before it, so I have little hope this piece of unfinished work will actually get done.
May whoever runs into an issue with this next find my code at least somewhat useful, if they want to pick up the torch :sweat_smile:

PR #6035, opened 2025-06-19
Optimize `DataDefinitionAnalyzer` a bit

Tayrtahn:
`DataDefinitionAnalyzer` has been around for a while and has accumulated a bit of cruft. This PR sorts out some of the weirdness and improves its performance a bit.

- Instead of analyzing every type definition to see if it's a data definition, then analyzing every field/property to see if it's part of a datadefinition, the analyzer now checks every type definition to see if it's a data definition, then analyzes only the fields/properties within that type. Because of this, we can also skip the checks that the containing type is a data definition (because it has to be for the analysis to occur in the first place).
- `IsReadOnlyDataField`, `HasRedundantTag`, `HasVVReadWrite`, and `IsNotYamlSerializable` are called in sequence and all started by calling `IsDataField`. Now we just call `IsDataField` once before calling any of them, so we don't have to run the method just to get the same results 4 times.
- ~There was a `foreach` loop in `AnalyzeDataField` that iterated over each variable in the field declaration. However, none of the conditions actually use this information, so it's pointless. I don't believe we ever use multiple variable declarations on datafields (like `[DataField] public int foo, bar;` - would that even serialize?), so this probably doesn't help performance, but it's a bit cleaner.~

### DIFF
--- a/Robust.Analyzers/DataDefinitionAnalyzer.cs
+++ b/Robust.Analyzers/DataDefinitionAnalyzer.cs
@@ -163,13 +163,16 @@ public sealed class DataDefinitionAnalyzer : DiagnosticAnalyzer
             if (fieldSymbol == null)
                 continue;
 
+            if (!IsDataField(fieldSymbol, out _, out var datafieldAttribute))
+                continue;
+
             if (IsReadOnlyDataField(type, fieldSymbol))
             {
                 TryGetModifierLocation(field, SyntaxKind.ReadOnlyKeyword, out var location);
                 context.ReportDiagnostic(Diagnostic.Create(DataFieldWritableRule, location, fieldSymbol.Name, type.Name));
             }
 
-            if (HasRedundantTag(fieldSymbol))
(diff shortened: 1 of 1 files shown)


PJB3005 (changes requested):
> There was a foreach loop in AnalyzeDataField that iterated over each variable in the field declaration. However, none of the conditions actually use this information, so it's pointless. I don't believe we ever use multiple variable declarations on datafields (like [DataField] public int foo, bar; - would that even serialize?), so this probably doesn't help performance, but it's a bit cleaner.

It clearly was, though? The loop body uses the variable to look up the symbol to validate. And yes, that totally does serialize: 

image

Tayrtahn:
You're absolutely right - I made a change earlier that made it stop using `variable` and didn't realize that I did it myself. I've reverted that commit and fixed the earlier change so it now properly analyzes multiple declarations.

Also, good to know that serialization supports that!

PJB3005 (approved):

PR #6044, opened 2025-06-22
Make some `PlacementManager` dependency fields public

Tayrtahn:
Changes `PlacementManager` fields `MapManager`, `EyeManager`, and `EntityManager` from internal to public.

This is to allow content-defined placement modes to access these managers the same way as engine-defined placement modes, rather than having to IoC resolve them.

### DIFF
--- a/Robust.Client/Placement/IPlacementManager.cs
+++ b/Robust.Client/Placement/IPlacementManager.cs
@@ -1,5 +1,8 @@
 ﻿using System;
+using Robust.Client.Graphics;
 using Robust.Shared.Enums;
+using Robust.Shared.GameObjects;
+using Robust.Shared.Map;
 using Robust.Shared.Maths;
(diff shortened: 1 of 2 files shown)


PJB3005 (commented):
This doesn't add  them to the interface though?

Tayrtahn:
> This doesn't add  them to the interface though?

Oops, I can fix that.

Mildly interesting: it _technically_ doesn't need to be in the interface because `PlacementMode.pManager` uses the concrete `PlacementManager` and not the `IPlacementManager` interface: 

https://github.com/space-wizards/RobustToolbox/blob/da0abd253555825f371df44a41cc7fe789bce829/Robust.Client/Placement/PlacementMode.cs#L23

Is that worth changing, or meh? In any case, I'll add these fields to the interface.

PJB3005 (approved):

PR #6049, opened 2025-06-26
Add new event for os window

beck-thompson:
Needed for https://github.com/space-wizards/space-station-14/pull/36969

Right now, there isn't a way to listen for when the window is requested to close (E.g someone clicks the X button). This means that you have to listen for ALL closes. In the viewport pr, we want to actually close the window when someone presses the x, but we don't want to close it when they simply pop the window back in (that will close the OS window, but not the viewport window)


PJB3005:
So I don't mind having this as a feature (assuming you change the API so the request takes a proper cancellation opportunity), but this PR isn't necessary to implement your content feature.

beck-thompson:
> So I don't mind having this as a feature (assuming you change the API so the request takes a proper cancellation opportunity), but this PR isn't necessary to implement your content feature.

I'm not sure if I should pass through the `WindowRequestClosedEventArgs` as well as the cancellation. Let me know :saluting_face:

Robust.Client/UserInterface/Controls/OSWindow.cs
```diff
@@ -86,9 +86,9 @@ public string Title
         public event Action<CancelEventArgs>? Closing;
 
         /// <summary>
-        /// Raised when the user clicks the close button on the window.
+        /// Raised when the user clicks the close button on the window. Can be canceled!
         /// </summary>
-        public event Action<WindowRequestClosedEventArgs>? RequestClosed;
+        public event Action<WindowRequestClosedEventArgs, CancelEventArgs>? RequestClosed;
```
PJB3005: Make this a new event type instead of passing two objects. And you don't need to include the data of `WindowRequestClosedEventArgs` anyways.

PJB3005 (changes requested):

Robust.Client/UserInterface/Controls/OSWindow.cs
beck-thompson: OK good I was thinking it was not necessary!

Robust.Client/UserInterface/Controls/OSWindow.cs
```diff
@@ -85,6 +85,11 @@ public string Title
         /// </summary>
         public event Action<CancelEventArgs>? Closing;
 
+        /// <summary>
+        /// Raised when the user clicks the close button on the window. Can be canceled!
+        /// </summary>
+        public event Action<CancelEventArgs>? RequestClosed;
```
PJB3005, line 91: Make this a new event type.

PJB3005 (changes requested):

PJB3005:
Stale.

PR #6057, opened 2025-06-28
fix: resolve remaining ResolvedSoundSpecifier warnings

perryprog:
All remaining ResolvedSoundSpecifier-string-literal-obsolete-warnings were due to calls to `GetAudioLength`, but within the AudioSystem it makes sense to pass around pathnames, so I added a protected overload that takes in strings.

Progress towards space-wizards/space-station-14#33279.


perryprog:
Wondering if the overload should just be made public because there are seemingly valid uses of looking up audio length by filename in Content (Jukebox BUI), but not sure if it's something that should be encouraged.

PJB3005:
> Wondering if the overload should just be made public because there are seemingly valid uses of looking up audio length by filename in Content (Jukebox BUI), but not sure if it's something that should be encouraged.

These use cases are fine to just manually wrap it into a `ResolvedPathSpecifier`, which isn't considered obsolete.

PR #6062, opened 2025-07-04
Mark ValidatePrototypeIdAttribute as obsolete

EmoGarbage404:
These have completely fallen out of favor in exchange for the static ProtoIds so we should just mark them as obsolete so we can start to get rid of them and people can stop using them.


Tayrtahn (approved):
Yep, it's time we say goodbye to this. Using actual types through `ProtoId` is vastly better!

Tayrtahn:
Aaaaand it's gone:
Screenshot 2025-07-07 at 4 03 24 PM

PR #6065, opened 2025-07-05
feat: Add VV editor for tuples

perryprog:
Tuples good.

image

I'm still pretty bad at C#'s type system so I suspect the way I'm effectively hardcoding in the casts here could be improved...

Doesn't quite work yet because property editors only show on types that are in the serializer type dictionary and while `(TimeSpan, TimeSpan)` is (from above screenshot). In other words (and from what I understand), only types that have appeared in a serialized data type will have VV editors displayed for them, meaning the likely uniqueness of tuples makes this more common. I'm not quite sure how to fix this and I need to go sleep so opening as draft in case anyone has answers.

Relevant issue I'm running into is from:

https://github.com/space-wizards/RobustToolbox/blob/34d65a7960930f2b732ab84edc0e1482d7c2888d/Robust.Client/ViewVariables/ViewVariablesPropertyControl.cs#L80


perryprog:
I can't think of any reasonable solution besides just having it only work on tuples that have been defined in a serialized component. It means no VV editor for serverside components that have some weirdly-typed tuple, but I don't know a way around the fact that it still needs a registered serializer for the client.

PJB3005 (changes requested):
You can handle the networking the same way `KeyValuePair<K, V>` is handled.

perryprog:
Hmm. In doing that I'm not seeing any success, I think because of the reference type check that ViewVariablesPropertyControl has which makes any value types use a dummy editor (this check is skipped if the type is registered with the serializer). So this works fine for the KeyValuePair editor, but not for ValueTuples.

Edit: Ohhh I see; I wasn't paying attention to the ServerKeyValuePairToken bit.

perryprog:
Ack, this is more tricky than I thought—it turns out that even with a token for tuples you still run into the issue of the ViewVariablesPropertyEditor wanting the MemberData's type to be a reference type. You can see this if you have a component with a KeyValuePair field—it will end up using the dummy editor, and just getting ToString'd.

So I'll have to tweak the logic a bit of the property editor but I think I have an idea of what can be done.

PJB3005:
> Ack, this is more tricky than I thought—it turns out that even with a token for tuples you still run into the issue of the ViewVariablesPropertyEditor wanting the MemberData's type to be a reference type. You can see this if you have a component with a KeyValuePair field—it will end up using the dummy editor, and just getting ToString'd.

Congratulations, you found a bug in VV. Stand by while I fix my own 7 year old code.

PJB3005:
de0871d17be5d8f80f082130d0210c865c952f2a

perryprog:
Amazing, that's working great.

Only issue I'm running into now is on the writing side—I'm currently sending over a serialized ServerTupleToken, similar to how PrototypeReferenceToken works. Once that's unpacked into an (object, object), I'm not sure how to make that actually assignable for the TryModifyProperty call.

Presumably the proper way of doing this is having proper nested selectors for tuples letting an edit of a tuple field send over just that one value, which can be handled as normal, but that might be a bit out of depth for my understanding of VV.

The easy workaround here is probably just to disallow editing if the type is serverside only (and not serializable). Thoughts?

PJB3005:
> Presumably the proper way of doing this is having proper nested selectors for tuples letting an edit of a tuple field send over just that one value, which can be handled as normal, but that might be a bit out of depth for my understanding of VV.

If KeyValuePair doesn't currently allow it then you don't need to bother IMO.

perryprog:
Took a bit of elbow grease to make sure editing worked for clientside components while being disabled for serverside components, but got it there in the end. Probably overkill but I learned stuff so 🤷 

image

PJB3005 (commented):
Can this entire system be expanded to N tuple sizes? Only having it for 2-element tuples seems short-sighted.

perryprog:
I originally attempted that and didn’t have any success due to typing issues on writing values. (I can’t find a way to make the editor editable without having it be generic.) I also ended up with the awful code repetition of handling each possible tuple size since as far as I can tell there’s no way to dynamically construct tuples of a given arity (though maybe there’s some reflection black magic that’ll make it work).

I can give it another shot, though—it’s just definitely pushing the limits of my C# skills.

perryprog:
I figured out the magical reflection invocation to get higher arities working.

Tuples that are registered with the type serializer will be editable (like (TimeSpan, TimeSpan)), tuples from the server that aren't registered with the serializer are readonly (matching other value-type VV editors), and tuples on the client are writable for arities seven and below. 

Here's what it looks like for some random examples; the left is clientside, the right serverside.

image

Robust.Server/ViewVariables/ViewVariablesSession.cs
```diff
@@ -124,17 +126,34 @@ public bool TryGetRelativeObject(object[] propertyIndex, out object? value)
             for (var i = 1; i < propertyIndex.Length; i++)
             {
                 var selector = propertyIndex[i];
+
+                if (value == null || !value.GetType().IsGenericType)
```
PJB3005: What is `!value.GetType().IsGenericType` supposed to achieve?

Robust.Server/ViewVariables/ViewVariablesSession.cs
```diff
@@ -124,17 +126,34 @@ public bool TryGetRelativeObject(object[] propertyIndex, out object? value)
...
                         dynamic kv = value;
                         value = kvPair.Key ? kv.Key : kv.Value;
                         break;
+
+                    // Using this selector on an IEnumerable is not actually used for anything right now.
+                    // I was adding this for the tuple case below and it made sense to add for future use.
+                    case ViewVariablesEnumerableIndexSelector indexSelector
+                        when value is IEnumerable<object> list:
+                        value = list.ElementAtOrDefault(indexSelector.Index);
+                        break;
```
PJB3005: Do not add things like this just for future use.

Robust.Server/ViewVariables/ViewVariablesSession.cs
```diff
@@ -124,17 +126,34 @@ public bool TryGetRelativeObject(object[] propertyIndex, out object? value)
...
+
+                    case ViewVariablesEnumerableIndexSelector indexSelector
+                        when value is ITuple tuple:
+                        value = indexSelector.Index <= tuple.Length - 1 ? tuple[indexSelector.Index] : null;
+                        break;
```
PJB3005: This should be a separate selector type.

Robust.Server/ViewVariables/ViewVariablesSession.cs
```diff
@@ -124,17 +126,34 @@ public bool TryGetRelativeObject(object[] propertyIndex, out object? value)
...
+
+                    case ViewVariablesEnumerableIndexSelector:
+                        return false;
```
PJB3005: Why add this?

RELEASE-NOTES.md
```diff
@@ -40,6 +40,7 @@ END TEMPLATE-->
 ### New features
 
 * `RobustClientPackaging.WriteClientResources()` and `RobustServerPackaging.WriteServerResources()` now have an overload taking in a set of things to ignore in the content resources directory.
+* There's now a ViewVariables property editor for two-element tuples.
```
PJB3005: Update this.

Robust.Client/ViewVariables/Editors/VVPropEditorTuple.cs
```diff
@@ -0,0 +1,97 @@
...
+        return editor;
+    }
+
+    // Allow selecting, for example, dictionaries within the tuple.
+    // Wait, why do you have a field with a tuple that holds a dictionary??
+    public override void WireNetworkSelector(uint sessionId, object[] selectorChain)
+    {
+        for (var i = 0; i < _editors.Count; i++)
+        {
+            var chain = selectorChain.Append(new ViewVariablesEnumerableIndexSelector(i)).ToArray();
```
PJB3005:
```suggestion
            var chain = [..selectorChain, new ViewVariablesEnumerableIndexSelector(i)];
```

Robust.Client/ViewVariables/Editors/VVPropEditorTuple.cs
```diff
@@ -0,0 +1,97 @@
...
+            return vBoxContainer;
+
+        _actualType = value.GetType();
+
+        // We disallow editing tuples with arity more than 7 since they would
+        // a pain to construct via reflection. And no one should have tuples
+        // that large. (8 is bad because last element becomes a ValueTuple<>)
+        _readOnly = ReadOnly
+                    || tuple.Length >= 8
+                    || value is ViewVariablesBlobMembers.ServerTupleToken;
```
PJB3005: This check should be changed to checking that the type is exactly `ValueTuple`, as your `ToTuple()` code can't handle other types of tuples.

PJB3005 (changes requested):

Robust.Server/ViewVariables/ViewVariablesSession.cs
perryprog: Oops, was back from when I was trying to make the tuple editor itself generic. Turns out it's just needed for ViewVariablesSelectorKeyValuePair.

Robust.Server/ViewVariables/ViewVariablesSession.cs
perryprog: I guess it doesn't really matter since it ends up being null which gets denied anyway, but it was meant to handle the case of a selector coming in and that pointing to an object of a type that that selector isn't supposed to handle. There's a similar guard for KeyValuePair.

Robust.Client/ViewVariables/Editors/VVPropEditorTuple.cs
```diff
@@ -0,0 +1,103 @@
...
+            _tuple.Add(tuple[i]);
+            _editors.Add(editor);
+        }
+        return vBoxContainer;
+    }
+
+    private bool IsValueTuple(Type actualType)
+    {
+        // As far as I can tell, it's this or spelling out every possible arity.
+        return actualType.Name.Split('`').FirstOrDefault() == nameof(ValueTuple);
```
PJB3005: Spell out the arities instead.

PJB3005 (changes requested):

PR #6066, opened 2025-07-05
Add `ForbidLiteral` to `IPrototypeManager` methods

Tayrtahn:
Adds `ForbidLiteralAttribute` to `IPrototypeManager` methods that take a string, `EntProtoId`, or `ProtoId` parameter to lookup a prototype (`Index`, `HasIndex`, `TryIndex`).

Also fixes all violations of this in engine. Two were `ShaderPrototype`s which are now validated `static ProtoId`s. The rest were test prototypes which now use `const string`s.

Content violations [are](https://github.com/space-wizards/space-station-14/pull/38728) [already](https://github.com/space-wizards/space-station-14/pull/38745) [fixed](https://github.com/space-wizards/space-station-14/pull/38747).

### DIFF
--- a/Robust.Client/GameStates/NetInterpOverlay.cs
+++ b/Robust.Client/GameStates/NetInterpOverlay.cs
@@ -13,6 +13,8 @@ namespace Robust.Client.GameStates
 {
     internal sealed class NetInterpOverlay : Overlay
     {
+        private static readonly ProtoId<ShaderPrototype> UnshadedShader = "unshaded";
+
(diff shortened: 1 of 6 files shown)


slarticodefast:
I wonder if it is worth considering to even fully obsolete the overloads using strings.
Or are those specifically needed anywhere?

Tayrtahn:
> I wonder if it is worth considering to even fully obsolete the overloads using strings.
> Or are those specifically needed anywhere?

I think they're needed (or at least convenient) for tests, where ProtoIds can't really be validated.

PR #6069, opened 2025-07-06
Cleanup warnings: CS0414

B-Kirill:
Remove 1x `CS0414` warning

`_timing` is only used in the debug build, so the warning should be suppressed.

### DIFF
--- a/Robust.Shared/Physics/Systems/FixtureSystem.cs
+++ b/Robust.Shared/Physics/Systems/FixtureSystem.cs
@@ -21,7 +21,9 @@ namespace Robust.Shared.Physics.Systems
     /// </summary>
     public sealed partial class FixtureSystem : EntitySystem
     {
+#pragma warning disable CS0414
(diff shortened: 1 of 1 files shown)


PJB3005 (changes requested):
Can you make this `#if DEBUG` instead?

Tayrtahn:
> Can you make this `#if DEBUG` instead?

https://github.com/space-wizards/RobustToolbox/pull/5892#issuecomment-2832151219

We should decide how we want to handle this situation, since it seems to keep coming up.

PJB3005:
> > Can you make this `#if DEBUG` instead?
> 
> [#5892 (comment)](https://github.com/space-wizards/RobustToolbox/pull/5892#issuecomment-2832151219)
> 
> We should decide how we want to handle this situation, since it seems to keep coming up.

Use `#if`, if CI is a concern then make sure CI builds both debug & release.

B-Kirill:
Aww.. I just fixed my PC. Anyway, thanks for doing this for me.

PR #6070, opened 2025-07-06
Config load fail is now an error instead of a warning

VasilisThePikachu:
People keep ignoring it whenever they ask "Why is my config not working :(" till it is pointed out.

Plus this should be an error anyway in my opinion, we caught some exception, we are using defaults now cause of it. This is more serious then a warning, something is wrong.

This should make it more obvious.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -47,7 +47,7 @@ END TEMPLATE-->
 
 ### Other
 
-*None yet*
+* ConfigurationManager will now report an error instead of warning if it fails to load the config file.
 
 ### Internal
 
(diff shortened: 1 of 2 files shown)


RELEASE-NOTES.md
```diff
@@ -47,7 +47,7 @@ END TEMPLATE-->
 
 ### Other
 
-*None yet*
+* ConfigurationManager will now report an error instead of warning if it fails to load the config file.
```
PJB3005:
```suggestion
* The configuration system will now report an error instead of warning if it fails to load the config file.
```

PR #6071, opened 2025-07-07
Add mover to the container insert and remove methods

Winkarst-cpu:
This would allow us to move logic from custom ``CanInsert``/``CanRemove``/``TryInsert``/``TryRemove`` methods in SS14 into event subscriptions, since we would be able to pass the user as the mover and show popups or perform checks against them.


PJB3005:
This fundamentally does not make sense. You want to specify data about "why is this act being performed" but you're only specifying "this entity did it", which is an inflexible specific subset.

PR #6074, opened 2025-07-10
Missing await WaitIdleAsync in unit test

iaada:
I'll be honest that I'm out of my element here. I'm not at all confident with async, but this test fail gave me the run around all night on my machine and this fixed it. When scanning through the other tests, most if not all had some variant of this line after `StartServer()`, except for this test.


PJB3005:
And this will make me rip the damn bandaid off and make `.System<T>()` check idleness properly.

PR #6076, opened 2025-07-10
Makes the StartCollideEvent constructor public.

MLGTASTICa:
The Hullrot codebase manually builds and raises theis event to prevent bullets phasing through things they shouldn't without raising the physics tickrate.
It also allows people to create special effects/mechanics by forcing collisions only under certain conditions , etc.


PJB3005:
Gonna go with a big "no" here.

If you want to manually trigger "collisions", do it at a higher gameplay level. If you want to fix bullet-through-paper, then contribute proper improvements to the physics system.

Issue #5937, opened 2025-05-16
VVPropEditorEntityUid is not used for EntityUids


slarticodefast:
![Image](https://github.com/user-attachments/assets/c6441dc4-1e44-4fc1-ba47-f4d83b46e2ad)

If you open the VV window you see both the netId and uid for entities. For netIds you can easily click on the `View` button to open a new VV window for that entity.

If you compare
https://github.com/space-wizards/RobustToolbox/blob/40d869948d4485dadde60744838ee9963f229beb/Robust.Client/ViewVariables/Editors/VVPropEditorEntityUid.cs
and
https://github.com/space-wizards/RobustToolbox/blob/40d869948d4485dadde60744838ee9963f229beb/Robust.Client/ViewVariables/Editors/VVPropEditorNetEntity.cs
you see they are identical and this should work the same for EntityUids. Fixing this would be really useful for debugging as it allows you to jump from one entity to another in VV, for example from the MindContainerComponent to the mind entity

![Image](https://github.com/user-attachments/assets/80ebb6be-8d53-4085-863f-564d1b010172)

Apparently this broke in https://github.com/space-wizards/RobustToolbox/commit/3ca686298e35f2d7a3d8ce803a41be773fa13dfb

EntityUids are value types and are therefore only assigned the dummy editor here
https://github.com/space-wizards/RobustToolbox/blob/40d869948d4485dadde60744838ee9963f229beb/Robust.Client/ViewVariables/ViewVariablesPropertyControl.cs#L84-L87
instead of using `PropertyFor`
https://github.com/space-wizards/RobustToolbox/blob/40d869948d4485dadde60744838ee9963f229beb/Robust.Client/ViewVariables/ClientViewVariablesManager.cs#L65

This bug likely affects other datatypes that should use `PropertyFor` as well.

PJB3005:
Fixed by de0871d17be5d8f80f082130d0210c865c952f2a

PR #6077, opened 2025-07-10
feat: now view-variable controls can be registered from content, or even dynamically added

Fildrance:
Refactored code that creates VV-related controls so they could be registered from content. Extracted class+interface for it. Simplified related code a bit.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -14,7 +14,7 @@ Don't change the format without looking at the script!
 
 ### New features
 
-* VV control factory was extracted from  ClientViewVariablesManager to enable adding VV controls from assemblies that use RT
+* Custom VV controls can now be registered using IViewVariableControlFactory.
 
 ### Bugfixes
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -14,7 +14,7 @@ Don't change the format without looking at the script!
 
 ### New features
 
-*None yet*
+* VV control factory was extracted from  ClientViewVariablesManager to enable adding VV controls from assemblies that use RT
 
 ### Bugfixes
 
--- a/Robust.Client/ViewVariables/IViewVariableControlFactory.cs
+++ b/Robust.Client/ViewVariables/IViewVariableControlFactory.cs
@@ -1,8 +1,32 @@
-﻿using System;
+using System;
+using Robust.Client.ViewVariables.Editors;
 
 namespace Robust.Client.ViewVariables;
 
+/// <summary>
+/// Factory that creates UI controls for viewing variables based on provided property type.
+/// </summary>
 public interface IViewVariableControlFactory
 {
+    /// <summary>
+    /// Creates UI control for viewing variable of type <paramref name="type"/>.
+    /// Returns <see cref="VVPropEditorDummy"/> if fails to find proper control.
+    /// First will look for control factory by type match (<see cref="RegisterForType{T}"/>),
+    /// then will try to check each registered conditional factory (<see cref="RegisterWithCondition"/>).
+    /// </summary>
     VVPropEditor CreateFor(Type? type);
+
+    /// <summary>
+    /// Registers factory method for vv control. This factory method will be used if provided type will be exactly equal to <typeparamref name="T"/>.
+    /// </summary>
+    /// <typeparam name="T">Type of property, for which this control factory should be used. Will only be used in case of exact match.</typeparam>
+    /// <param name="factoryMethod">Factory method that creates VV control.</param>
+    void RegisterForType<T>(Func<Type, VVPropEditor> factoryMethod);
+
+    /// <summary>
+    /// Registers factory method for vv control. This factory method will be used if <paramref name="condition"/> will return true for provided type.
+    /// </summary>
+    /// <param name="condition">Condition, that will decide, if factory should be used for provided type.</param>
+    /// <param name="factory">Factory method that creates VV control.</param>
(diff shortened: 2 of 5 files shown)


Fildrance:
Forgot changelog, looked at [5937 ](https://github.com/space-wizards/RobustToolbox/issues/5937) while i am at it - bunch of checks looks pretty much redundant compared to what is inside this PR! But probably better make it in separate PR. Also we have control for 'EntityUid' but not 'EntityUid?' - so that should be added too...

image

Fildrance:
While we are at it - https://github.com/space-wizards/RobustToolbox/pull/6065/files probably can be reviewed too (or closed, as tuples in public API is an evil doing and should not exist actually).

perryprog:
Only real use I can think of for tuples is (timespan, timespan) (assuming their fields are named of course) to be honest (which is specifically what I made that for).

Fildrance:
Added methods to add conditions at the start and at the end of the list, so content projects could add VV editors with more specific checks and more broad ones too and kinda manage them. Its not as good as some explicit ordering but i think this will do. Other option is to make them named and add some kind of 'insert after' and 'insert before', but i see it being more of a pain then useful most of the time.

RELEASE-NOTES.md
```diff
@@ -14,7 +14,7 @@ Don't change the format without looking at the script!
 
 ### New features
 
-*None yet*
+* VV control factory was extracted from  ClientViewVariablesManager to enable adding VV controls from assemblies that use RT
```
PJB3005: This changelog entry is both extremely verbose and does not give useful information. It does not say *where* the new API is, and "from assemblies that use RT" is redundant information that does not need to be stated.

Robust.Client/ViewVariables/IViewVariableControlFactory.cs
PJB3005, line 1: I would merge these "`AtStart`" and "`AtEnd`" operations into a single method that takes an enum or something for the insertion position.

Robust.Client/ViewVariables/ViewVariableControlFactory.cs
```diff
@@ -0,0 +1,147 @@
...
+using Robust.Shared.IoC;
+using Robust.Shared.Map;
+using Robust.Shared.Maths;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Serialization;
+using Robust.Shared.ViewVariables;
+
+namespace Robust.Client.ViewVariables;
+
+public sealed class ViewVariableControlFactory : IViewVariableControlFactory
```
PJB3005: This should be `internal`.

Robust.Client/ViewVariables/ViewVariableControlFactory.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        RegisterWithConditionAtStart(type => type.IsEnum, _ => new VVPropEditorEnum());
+        RegisterWithConditionAtStart(
+            type => type.IsGenericType && type.GetGenericTypeDefinition() == typeof(ProtoId<>),
+            type =>
+            {
+                var typeArgumentType = type.GenericTypeArguments[0];
+                var editor = CreateGenericEditor(typeArgumentType, typeof(VVPropEditorProtoId<>));
+
+                IoCManager.InjectDependencies(editor);
```
PJB3005: Take a dependency on `IDependencyCollection` instead and use that.

PJB3005 (changes requested):

Robust.Client/ViewVariables/IViewVariableControlFactory.cs
Fildrance: I doubt enum will have any more then then 2 useful values, so bool flag it is. And as most of people would like inserting first - gonna use that as default value.

Robust.Client/ViewVariables/ViewVariableControlFactory.cs
Fildrance: done

Robust.Client/ViewVariables/ViewVariableControlFactory.cs
Fildrance: done

RELEASE-NOTES.md
Fildrance: tried to make it better with everything i got :>

Robust.Client/ViewVariables/IViewVariableControlFactory.cs
PJB3005: No, make it an enum. When using a bool flag it is not obvious whether "true" means "start" or "end".

PJB3005:
You do not need to confirm every single review comment with a reply, you're just spamming the GitHub channel on Discord.

Robust.Client/ViewVariables/IViewVariableControlFactory.cs
Fildrance: Refactored to enum. Tried to make it more of a generic thing - might be helpful later somewhere else in same situation. namespace would be awkward, but moving it should not be too harmful later.

Tayrtahn:
This seems to make `SoundSpecifier` fields use the reference editor instead of their specific editor:
Screenshot 2025-07-21 at 3 33 40 PM

Fildrance:
> This seems to make `SoundSpecifier` fields use the reference editor instead of their specific editor: 

It appears that adding conditions at the start of the list over  and over makes first condition actually become last :godo:
I actually needed to change order of condition checks in reverse compared to what we originally had. Now order of checks and options this factory presents should be exactly the same as it was previously.

Tayrtahn (approved):

perryprog:
Looks like changelog got put in wrong place.

PR #6080, opened 2025-07-12
Soft Layers Fixtures Component. 

Princess-Cheeseballs:
### Description

Adds a component that allows you to declare certain layers as "Soft" on an entity. 

This allows you to have a hard fixture that doesn't stop movement on some of its layers but will still raise start and end collide events without having to give it a second fixture, by taking advantage of the "PreventCollideEvent."

System will find the layers that the collision is happening on and cancel it if all the layers are contained within the component's bitmask. 

Need it for funny furniture shenanigans where crawling over a piece of furniture will appropriately apply a speed reduction, while also still being a hard fixture.

This component may also be useful on mobs specifically in regard to the flammable fixture which needs to be soft so it doesn't cause hard collisions, and results in two collisions happening in some cases.

#### Additional Comments

The name I chose is kind of ass in my opinion. If a better name presents itself I will change the system and component name immediately.

### DIFF
--- a/Robust.Shared/Physics/Components/FixturesSemiSoftComponent.cs
+++ b/Robust.Shared/Physics/Components/FixturesSemiSoftComponent.cs
@@ -1,0 +1,19 @@
+using Robust.Shared.GameObjects;
+using Robust.Shared.GameStates;
+using Robust.Shared.Serialization.Manager.Attributes;
+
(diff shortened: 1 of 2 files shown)


PJB3005:
This can just be implemented in content, can't it? I feel like this is too specific for being included in the engine.

PR #6082, opened 2025-07-13
feat: added generic method for getting component from ComponentRegistry

Fildrance:
Extracting data from ComponentRegistry is a bit awkward as we need to pass comp name which is generally controlled by ComponentFactory. Due to this fact, it would be messy to add generic method that gets component from ComponentRegistry  on ComponentRegistry  itself - so i added it to ComponentRegistry instead.


Fildrance:
Tried to add unit-test but interacting with ComponentRegistry in unit-tests is quite puzzling, not sure where to start and whats best to not touch, as everything is not too 'unit'-like... its basically a helper method, so not sure if we want test for it, and if we want - where is the best point to start :<

PJB3005 (changes requested):
Why is this helper function on `IComponentFactory`?

Fildrance:
> Why is this helper function on `IComponentFactory`?

Because ComponentFactory holds logic for component names, and injecting it into ComponentRegistry seems like an overkill. Also ComponentFactory is available from every EntitySystem, so it will be easily available. Its not on EntitySystem as EntitySystem does not currently provide any proxies for ComponentFactory  and seems cramped enough as is. But if you see some better place - i'm all for it, i just dont see any clear candidate myself (w/o dragging ComponentFactory  dependency).

PJB3005:
> > Why is this helper function on `IComponentFactory`?
> 
> Because ComponentFactory holds logic for component names, and injecting it into ComponentRegistry seems like an overkill. Also ComponentFactory is available from every EntitySystem, so it will be easily available. Its not on EntitySystem as EntitySystem does not currently provide any proxies for ComponentFactory and seems cramped enough as is. But if you see some better place - i'm all for it, i just dont see any clear candidate myself (w/o dragging ComponentFactory dependency).

Either make the function on `ComponentRegistry` and take in the factory as parameter, or make a an extension method against `IComponentFactory` instead.

Fildrance:
After some thinking... i guess the worst case is that people will have problems with finding that new convinience method, and best way to avoid it - is, sadly, to slap it on ComponentRegistry. I hate when objects inject some 'serice-like' type as parameter w/o following some convention or pattern, but thats lesser evil in this case.

PJB3005 (approved):

PR #6086, opened 2025-07-17
Add GetDescendant methods to SharedContainerSystem

notquitehadouken:
Title

Specifically, getting every single entity that is contained in a container/entity, including in other containers in those entities

Mostly just for convenience

### DIFF
--- a/Robust.Shared/Containers/SharedContainerSystem.cs
+++ b/Robust.Shared/Containers/SharedContainerSystem.cs
@@ -678,14 +678,9 @@ namespace Robust.Shared.Containers
         /// <summary>
         /// Enumerates every entity in an entity with a container, searching all child containers.
         /// </summary>
-        public IEnumerable<EntityUid> GetDescendantEntitiesInEntity(
-            EntityUid entity,
-            ContainerManagerComponent? containerManager = null)
+        public IEnumerable<EntityUid> GetDescendantEntitiesInEntity(Entity<ContainerManagerComponent> entity)
         {
-            if (!Resolve(entity, ref containerManager))
-                yield break;
-
-            foreach (var (_, container) in containerManager.Containers)
+            foreach (var (_, container) in entity.Comp.Containers)
             {
                 foreach (var uid in GetDescendantEntitiesInContainer(container))
                 {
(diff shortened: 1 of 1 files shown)


metalgearsloth (changes requested):
This is doing a silly amount of allocations.

notquitehadouken:
> This is doing a silly amount of allocations.

how silly.......

pheenty:
There is already a method like this but it is private iirc

pheenty:
Nvm it's for ascendant insertion not for descendant enumeration

PJB3005 (commented):
Why do these return `IEnumerator<T>` instead of `IEnumerable<T>`?

notquitehadouken:
> Why do these return `IEnumerator<T>` instead of `IEnumerable<T>`?

¯\\\_(ツ)\_/¯

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -675,6 +675,40 @@ internal bool TryGetManagerComp(EntityUid entity, [NotNullWhen(true)] out Contai
...
+
+        /// <summary>
+        /// Enumerates every entity in a specific container, searching all child containers.
+        /// </summary>
+        public IEnumerable<EntityUid> GetDescendantEntitiesInContainer(BaseContainer container)
+        {
+            foreach (var contained in container.ContainedEntities)
+            {
+                yield return contained;
+                foreach (var uid in GetDescendantEntitiesInEntity(contained))
```
PJB3005: Isn't this going to log an error for entities that don't have containers themselves? Because `GetDescendantEntitiesInEntity` uses `Resolve()`.

Robust.Shared/Containers/SharedContainerSystem.cs
```diff
@@ -675,6 +675,40 @@ internal bool TryGetManagerComp(EntityUid entity, [NotNullWhen(true)] out Contai
             return false;
         }
 
+        /// <summary>
+        /// Enumerates every entity in an entity with a container, searching all child containers.
+        /// </summary>
+        public IEnumerable<EntityUid> GetDescendantEntitiesInEntity(
+            EntityUid entity,
+            ContainerManagerComponent? containerManager = null)
```
PJB3005: Take an `Entity<T>` instead.

PJB3005 (changes requested):

PJB3005 (approved):

notquitehadouken:
...

PR #6090, opened 2025-07-20
Allow AutoNetworkedField to work with inherited datafields

slarticodefast:
#### About the PR
I know OOP is frowned upon in an ECS codebase, but hear me out:

Currently `AutoNetworkedField` is simply doing nothing when added to an abstract component and then inherited, because the source generator uses `GetMembers`, which only returns the direct members of the component, but not any inherited ones.

Abstract components are used all over the content repo, and I would argue using them makes sense in cases where it avoids a lot of copy-pasting between multiple similar components. See for example https://github.com/space-wizards/space-station-14/pull/35030 which had to rely on manual networking because the autogenerated networking wasn't working, so a lot of boilerplate code had to be added. With this PR that can be cleaned up.

#### Technical details
Added a new type symbol helper that iterates through base classes to find all AutoNetworkedField attributes in a component, including the inherited ones.
I could not find an existing integration test for autonetworking, so I added one that tests if everything works (mostly copied from ComponentStateTest).

Needs some content fixes https://github.com/space-wizards/space-station-14/pull/39086

### DIFF
--- a/Robust.UnitTesting/Shared/GameState/AutoNetworkingTest.cs
+++ b/Robust.UnitTesting/Shared/GameState/AutoNetworkingTest.cs
@@ -13,7 +13,7 @@ namespace Robust.UnitTesting.Shared.GameState;
 public sealed partial class AutoNetworkingTests : RobustIntegrationTest
 {
     /// <summary>
-    /// Does basic testing for AutoNetworkedFieldAttribute and AutoGenerateComponentAttribute
+    /// Does basic testing for AutoNetworkedFieldAttribute and AutoGenerateComponentStateAttribute
     /// to make sure the datafields are correctly networked to the client when dirtied.
     /// TODO: Add test for field deltas.
     /// </summary>
--- a/Robust.UnitTesting/Shared/GameState/AutoNetworkingTest.cs
+++ b/Robust.UnitTesting/Shared/GameState/AutoNetworkingTest.cs
@@ -79,12 +79,9 @@ public sealed partial class AutoNetworkingTests : RobustIntegrationTest
         // check client
         await client.WaitPost(() =>
         {
-            Assert.Multiple(() =>
-            {
-                // Check player got properly attached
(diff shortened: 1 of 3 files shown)


slarticodefast:
~~Huh, the RT tests were working locally. I'll look into it.~~

Fixed, I just moved something into the wrong namespace by accident.

Robust.UnitTesting/Shared/GameState/AutoNetworkingTest.cs
```diff
@@ -0,0 +1,221 @@
...
+using Robust.Shared.Map;
+using Robust.Shared.Network;
+using Robust.Shared.Serialization.Manager.Attributes;
+
+namespace Robust.UnitTesting.Shared.GameState;
+
+public sealed partial class AutoNetworkingTests : RobustIntegrationTest
+{
+    /// <summary>
+    /// Does basic testing for AutoNetworkedFieldAttribute and AutoGenerateComponentAttribute
```
slarticodefast:
```suggestion
    /// Does basic testing for AutoNetworkedFieldAttribute and AutoGenerateComponentStateAttribute
```

Robust.UnitTesting/Shared/GameState/AutoNetworkingTest.cs
```diff
@@ -0,0 +1,221 @@
...
+    /// to make sure the datafields are correctly networked to the client when dirtied.
+    /// TODO: Add test for field deltas.
+    /// </summary>
+    [Test]
+    public async Task AutoNetworkingTest()
+    {
+        var serverOpts = new ServerIntegrationOptions { Pool = false };
+        var clientOpts = new ClientIntegrationOptions { Pool = false };
+        var server = StartServer(serverOpts);
+        var client = StartClient(clientOpts);
```
PJB3005:
```suggestion
        using var server = StartServer(serverOpts);
        using var client = StartClient(clientOpts);
```

PJB3005 (approved):

PR #6091, opened 2025-07-20
Add RotationDirection to SpriteComponent

notquitehadouken:
When enabled, the sprite of anything that possesses both a rotation and an RSI with directions will have the direction rotated with the sprite

image

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -129,8 +129,6 @@ namespace Robust.Client.GameObjects
         public bool RotationDirection
         {
             get => rotationDirection;
-            [Obsolete("Use SpriteSystem.SetRotationDirection() instead.")]
-            set => Sys.SetRotationDirection((Owner, this), value);
         }
 
         [DataField("offset")] // Explicit name, in case this field ever gets renamed
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -121,14 +121,14 @@ namespace Robust.Client.GameObjects
         }
 
         [DataField]
-        internal float rotationInfluencesDirection = 0;
+        internal bool rotationDirection = false;
 
         /// <summary>
         ///     If <see cref="Rotation"/> will change the rendered RSI direction of the sprite
         /// </summary>
-        public float RotationInfluencesDirection
+        public bool RotationDirection
         {
-            get => rotationInfluencesDirection;
+            get => rotationDirection;
         }
 
         [DataField("offset")] // Explicit name, in case this field ever gets renamed
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -121,14 +121,14 @@ namespace Robust.Client.GameObjects
         }
 
         [DataField]
-        internal bool rotationDirection = false;
+        internal float rotationInfluencesDirection = 0;
 
         /// <summary>
         ///     If <see cref="Rotation"/> will change the rendered RSI direction of the sprite
         /// </summary>
-        public bool RotationDirection
+        public float RotationInfluencesDirection
         {
-            get => rotationDirection;
+            get => rotationInfluencesDirection;
         }
 
         [DataField("offset")] // Explicit name, in case this field ever gets renamed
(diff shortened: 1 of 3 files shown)


K-Dynamic:
I'm a bit confused, could you explain what this allows you to do

notquitehadouken:
> I'm a bit confused, could you explain what this allows you to do

when a sprite is rotated by changing the Rotation value it will also rotate the directional sprite in the opposite direction so it looks like it always faces the same direction even when upside down or whatever

notquitehadouken:
image
consult the graph

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -120,6 +120,19 @@ public Angle Rotation
...
+        internal bool rotationDirection = false;
+
+        /// <summary>
+        ///     If <see cref="Rotation"/> will change the rendered RSI direction of the sprite
+        /// </summary>
+        public bool RotationDirection
+        {
+            get => rotationDirection;
+            [Obsolete("Use SpriteSystem.SetRotationDirection() instead.")]
+            set => Sys.SetRotationDirection((Owner, this), value);
```
PJB3005: Why would you **add** a new obsolete property?

Robust.Client/GameObjects/EntitySystems/SpriteSystem.Setters.cs
```diff
@@ -50,6 +50,18 @@ public void SetRotation(Entity<SpriteComponent?> sprite, Angle value)
...
+    public void SetRotationDirection(Entity<SpriteComponent?> sprite, bool value)
+    {
+        if (!_query.Resolve(sprite.Owner, ref sprite.Comp))
+            return;
+
+        if (sprite.Comp.RotationDirection == value)
+            return;
+
+        sprite.Comp.rotationDirection = value;
+        _tree.QueueTreeUpdate(sprite!);
```
PJB3005: Is this necessary? The property doesn't change the bounding box of the sprite.

Robust.Client/GameObjects/EntitySystems/SpriteSystem.Render.cs
```diff
@@ -116,16 +116,22 @@ private void RenderLayer(Layer layer, DrawingHandleWorld drawingHandle, ref Matr
...
         // due to direction overrides or offsets.
         if (overrideDirection != null && state != null)
             dir = overrideDirection.Value.Convert(state.RsiDirections);
-        dir = dir.OffsetRsiDir(layer.DirOffset);
+
+        if (sprite.RotationDirection && !sprite.Rotation.EqualsApprox(Angle.Zero))
+            dir = (-sprite.Rotation).RotateDir(dir.OffsetRsiDir(layer.DirOffset).Convert())
+                .Convert(state?.RsiDirections ?? RsiDirectionType.Dir1);
+        else
+            dir = dir.OffsetRsiDir(layer.DirOffset);
```
PJB3005: This line is copied between the two if branches.

PJB3005 (changes requested):
Would `DirOffset` not be sufficient for this? With this system you still can't have a rotation and a direction rotation if they don't match.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
notquitehadouken: great question

notquitehadouken:
> Would `DirOffset` not be sufficient for this? With this system you still can't have a rotation and a direction rotation if they don't match.

i'm not entirely sure i know what this means

PJB3005:
> i'm not entirely sure i know what this means

What if I want to apply RotationDirection only 90 degrees, but have the entire sprite rotated 180 degrees?

notquitehadouken:
> > i'm not entirely sure i know what this means
> 
> What if I want to apply RotationDirection only 90 degrees, but have the entire sprite rotated 180 degrees?

oh okay one sec

notquitehadouken:
float :trollface:

PJB3005 (changes requested):
I specifically mentioned `DirOffset` as it seemed like the best solution to me. In what scenario would a float parameter make sense here?

notquitehadouken:
well if you set it to 1/2 and rotate the sprite 180 degrees it'll only rotate the direction 90 degrees

PJB3005:
> well if you set it to 1/2 and rotate the sprite 180 degrees it'll only rotate the direction 90 degrees

Doing it by fractions like this sounds extremely impractical, when the obvious use case is single step offsets.

notquitehadouken:
...

PR #6092, opened 2025-07-21
NoteOn actually being a NoteOff fix

FairlySadPanda:
NoteOn should not be suppressed if actually a NoteOff (i.e. it has velocity 0).


Robust.Client/Audio/Midi/MidiRenderer.cs
```diff
@@ -575,18 +575,28 @@ public void SendMidiEvent(RobustMidiEvent midiEvent, bool raiseEvent)
...
-                        break;
 
+                        break;
                     case RobustMidiCommand.NoteOn:
+                        // Velocity 0 *can* represent a NoteOff event.
+                        var velocity = midiEvent.Velocity;
+                        if (velocity == 0)
+                        {
+                            _rendererState.NoteVelocities.AsSpan[midiEvent.Channel].AsSpan[midiEvent.Key] = 0;
+                            _synth.NoteOn(midiEvent.Channel, midiEvent.Key, velocity);
```
FairlySadPanda, line 586: We specifically call NoteOn here because NoteOn velocity 0 is in spec, it's not some weird bug or FS-specific thing

PR #6096, opened 2025-07-22
Move ScaleVisuals to Content (and improve it)

slarticodefast:
#### About the PR
Requires content PR: https://github.com/space-wizards/space-station-14/pull/39349

~~This PR adds an [AppearanceSystem compliant](https://github.com/space-wizards/docs/pull/307) way to scale a sprite from the server. Currently in most cases the sprite scale is just set directly in the client's `SpriteComponent`, for example for plant visuals or for dwarves. This makes it impossible for changelings to dynamically transform their size to dwarf height. I know, dwarves shouldn't be squished humans in the first place, but that is a larger spriting rework and this feature is needed anyways.~~

~~Since it was a common admin request I also added the new `scale` toolshed command that allows and entity's sprite scale to be set to a certain 2d vector. This could not be extended as part of the `scale` console command, as that also adjusts the fixture, but Box2d does not have any ellipsoid fixtures, so it can only be scaled equally into both dimensions.~~

~~I kept the existing `ScaleEntityEvent` that was part of the command and moved it into the API so that it is always called upon scaling.~~

~~For testing you need to whitelist the command in the `toolshedEngineCommandPerms.yml` file.~~

Moved it to the content repository instead.
Changing the fixture scale has to be done in the fixtures system due to access permissions.

#### Changelog

- The "scale" console command has been replaced with a toolshed command in the content repository.
- `ScaleVisualsComponent` and `ScaleVisualsSystem` have been moved to `Content.Shared.Sprite`.
- Added the `SharedFixturesSystem.ScaleFixtures` method.


Simyon264:
Please make this a toolshed command

ScarKy0:
Actually this is the one command id prefer to not be toolshed

Simyon264:
Why not?

ScarKy0:
> Why not?

Mainly ease of access and cause im not a fan of toolshed syntax and if anything can always just `do "setscale $ID 2 2"`

juliangiebel:
> Mainly ease of access and cause im not a fan of toolshed syntax and if anything can always just `do "setscale $ID 2 2"`

What is your issue with toolshed? Not being a fan of it isn't really a reason to use the deprecated way of implementing commands.

ScarKy0:
> What is your issue with toolshed? Not being a fan of it isn't really a reason to use the deprecated way of implementing commands.

I just find regular commands way faster and easier to use, i mean if this is made into a toolshed command i wont mind too much, but then `scale` should be made toolshed too

Simyon264:
> > What is your issue with toolshed? Not being a fan of it isn't really a reason to use the deprecated way of implementing commands.
> 
> I just find regular commands way faster and easier to use, i mean if this is made into a toolshed command i wont mind too much, but then `scale` should be made toolshed too

you can simply mark an entity and use `> marked setscale 2 2` or implement 2 commands one which takes an entiyuid as an argument and another that takes it via pipe

juliangiebel:
Afaik it's possible to allow the command to also work with just passing the entid as a parameter in addition to the pipe etc.

PJB3005 (changes requested):
Confirming that this should be a Toolshed command

slarticodefast:
> Confirming that this should be a Toolshed command

Should I change the existing scale command into a toolshed command as well?
That way the two commands would stay consistent.

PJB3005:
> > Confirming that this should be a Toolshed command
> 
> Should I change the existing scale command into a toolshed command as well? That way the two commands would stay consistent.

Separate PR if you want to do that.

slarticodefast:
I changed it into a toolshed command. The existing scale command is functionally the same as before.
The content integration test is failing because it's missing permissions for the new command. This needs this content PR to be fixed https://github.com/space-wizards/space-station-14/pull/39157
grafik

Robust.Shared/Toolshed/Commands/Misc/ScaleCommand.cs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    [CommandImplementation("set")]
+    public EntityUid Set([PipedArgument] EntityUid input, Vector2 scale)
+    {
+        _system ??= GetSys<SharedScaleVisualsSystem>();
+
+        _system.SetSpriteScale(input, scale);
+        return input;
+    }
```
Simyon264: Toolshed automatically converts single piped T's into IEnumerable<T> for commands that expect only IEnumerable's. This command implementation is not needed. Also applies to all other commands here that take a single entityuid when they have counterparts that take in an IEnumerable

Simyon264 (changes requested):

Robust.Shared/Toolshed/Commands/Misc/ScaleCommand.cs
slarticodefast: Good to know, fixed!
A lot of the code examples I looked at were doing the same though, so we might want to clean that up.

PJB3005 (changes requested):
Is there any reason this can't be in content?

slarticodefast:
> Is there any reason this can't be in content?

Well, the console command and the component already existed in the engine, so I thought I would keep it here to avoid code duplication.
Also the SpriteComponent itself is in RT as well.
If you want me to I can completely move either all of it or only the toolshed command to content.
But the SharedScaleVisualsSystem API and the component only make sense when both are in the same repository.

PJB3005:
I believe we discussed on Discord to clarify these things. Anything else? If not I'll close this PR so it can be moved to content.

slarticodefast:
> I believe we discussed on Discord to clarify these things. Anything else? If not I'll close this PR so it can be moved to content.

I'll just update the PR, it still needs to be removed from RT to avoid code duplication.
Give me a moment.

slarticodefast:
Ok, done.
The fixture scaling has to be done using an API method due to access permissions in the component.

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Fixtures.cs
```diff
@@ -71,6 +73,45 @@ public void SetRestitution(EntityUid uid, Fixture fixture, float value, bool upd
             _fixtures.FixtureUpdate(uid, manager: manager);
     }
 
+    /// <summary>
+    /// Increases or decreases all fixtures of an entity in size by a certain factor.
+    /// </summary>
+    public void ScaleFixtures(Entity<FixturesComponent?> ent, float factor)
```
PJB3005, line 79: Yeah I don't think having this API in engine is a good idea.

PJB3005 (dismissed):

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Fixtures.cs
slarticodefast: Discussed on discord, required due to access restrictions of `EdgeShape.Vertex`

PR #6099, opened 2025-07-23
Switch buttons

Absotively:
This adds `SwitchButton`, a switch button control. This is intended mainly for power buttons in things like the APC UI and gas manipulation device UIs.

https://github.com/space-wizards/space-station-14/pull/39161 adds styles for this control to space-station-14 and changes the APC UI to use it.

## Description and technical details

I've added some screenshots that show how the control's layout works. These use the styles from https://github.com/space-wizards/space-station-14/pull/39161. That PR will not actually implement all of the UI changes shown in these screenshots. It only changes the APC menu. I plan to change the gas pump status button in a later PR, and I currently do not plan to actually change the intercom menu at all, but it makes a useful layout example.

The control has three visible components. From left to right, they are:
* an optional main label describing what the switch controls;
* a toggle switch icon;
* and a label describing the current state.

image

Clicking any of these toggles the button state.

The control has localizable default state labels of "Off" and "On"; I think having state labels improves usability in most of the cases where this control would be used. [Microsoft uses default "Off" and "On" labels and encourages developers to use them](https://learn.microsoft.com/en-us/windows/apps/design/controls/toggles); it seems likely they'd provide different guidance if these default labels caused many localization issues.

The default state labels can be overwritten by setting `OffStateText` and `OnStateText`.

The state label portion of the switch reserves overlapping space for both labels, so the button's size and layout don't change when the button is clicked.

If there is a main label, and the button is expanded horizontally, the excess space is put at the end of the main label. This means that if multiple switch buttons with the same labels and width are stacked, for instance in a vertical `BoxContainer`, they will all line up nicely.
image

If the main label is empty, the icon and state label are aligned left.
image

It is my understanding that RobustToolbox doesn't support right-to-left text; if I am wrong about that then `SwitchButton` will need some additional work to handle rtl layouts.

## CheckButton

The `CheckButton` class appears to have originally been intended to be this type of control. It inherits from `Button` and I do not believe it has ever had any actual differences from `Button`.

It is currently used in three places, all of which are have layouts based around it being regular button. As such, removing it or turning it into the actual switch button implementation would cause problems for old game versions.

I have added `[Obsolete]` to `CheckButton`. In https://github.com/space-wizards/space-station-14/pull/39161, all uses of `CheckButton` are replaced with `Button`.

### DIFF
--- a/Robust.Client/UserInterface/Controls/SwitchButton.cs
+++ b/Robust.Client/UserInterface/Controls/SwitchButton.cs
@@ -1,3 +1,4 @@
+using Robust.Client.Graphics;
 using Robust.Shared.ViewVariables;
 using static Robust.Client.UserInterface.Controls.Label;
 using Robust.Shared.Localization;
@@ -11,8 +12,9 @@ namespace Robust.Client.UserInterface.Controls
     public class SwitchButton : ContainerButton
     {
         public const string StyleClassSwitchButton = "switchButton";
-        public const string StyleClassSwitchButtonChecked = "switchButtonChecked";
 
+        public const string StylePropertyTextureUnchecked = "textureUnchecked";
+        public const string StylePropertyTextureChecked = "textureChecked";
 
         public Label Label { get; }
         public Label OffStateLabel { get; }
@@ -68,23 +70,7 @@ namespace Robust.Client.UserInterface.Controls
         protected override void DrawModeChanged()
         {
             base.DrawModeChanged();
-
-            if (Pressed)
-            {
-                TextureRect?.AddStyleClass(StyleClassSwitchButtonChecked);
-                if (OffStateLabel is not null)
-                    OffStateLabel.Visible = false;
-                if (OnStateLabel is not null)
-                    OnStateLabel.Visible = true;
-            }
-            else
-            {
-                TextureRect?.RemoveStyleClass(StyleClassSwitchButtonChecked);
-                if (OffStateLabel is not null)
-                    OffStateLabel.Visible = true;
-                if (OnStateLabel is not null)
-                    OnStateLabel.Visible = false;
-            }
+            UpdateAppearance();
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


ArtisticRoomba:
Mmmmm, good cooking.

sowelipililimute:
first impression, it's nice that the switch is always bundled with the label

but, the design of the switch itself is a bit ambiguous & relies too much on something that's not colourblind accessible 

detailed thoughts for the specifics of that texture will be left on that PR when i get the time

PJB3005:
Bikeshedding the texture for accessibility: we could put an "I" and "O" symbol on the sides. Admittedly it may be a bit abstract in some cases but it's pretty standard iconography.

Absotively:
I appreciate the feedback on the texture. I'm going to respond in https://github.com/space-wizards/space-station-14/pull/39161 since that's where the texture is.

Robust.Client/UserInterface/Controls/SwitchButton.cs
```diff
@@ -0,0 +1,192 @@
...
+                }
+                if (OffStateLabel is not null)
+                {
+                    OffStateLabel.FontColorOverride = fontColor;
+                }
+                if (OnStateLabel is not null)
+                {
+                    OnStateLabel.FontColorOverride = fontColor;
+                }
+            }
```
sowelipililimute: is there a reason we're propagating these style properties manually instead of using a `SelectorChild` in the stylesheets? i'd prefer to see this rely on that over hardcoding propagation for these properties unless there's a good reason not to

sowelipililimute (changes requested):

Robust.Client/UserInterface/Controls/SwitchButton.cs
Absotively: The main reason is that I wanted to insulate the game code from implementation details in the engine, but if that's not the preferred approach I can definitely change it.

A second reason is that with the current implementation, stylesheets would require multiple levels of `SelectorChild` to style this. The only direct child of the button is a `BoxContainer`. The main label and the icon are direct children of the `BoxContainer`, so they could be accessed with two levels of `SelectorChild`; the state labels are children of a `PanelContainer` for layout reasons, so stylesheets would need three levels of `SelectorChild` to style them.

I suppose I could refactor the control to have all the parts be direct children, though then the control would need to basically reimplement parts of `BoxContainer`.

Absotively: Oh, another option would be to put style classes on the child elements so they can be selected directly. This is what `CheckBox` does. It's probably the least bad option now that I look at it.

Absotively: Looking more closely at `CheckBox`, its approach does mean that the alternate style for it (`MonotoneCheckBox`) is implemented as a new separate control that inherits from `CheckBox` rather than just a style class that only affects stylesheets. But it still has the benefit of being the established approach, and switch buttons don't currently have any alternate styles anyways.

Absotively: The existence of `MonotoneCheckBox` convinced me that implementing a little bit of layout wouldn't be so bad. And it does get rid of two containers that were more convenient than necessary.

The new layout calculations also nicely handle switches with both state labels set to empty strings, by not reserving space for the state labels _or_ the gap between them and the icon. The previous version did not remove the gap, so the icon ended up not quite at the right edge of the control.

It might still be useful to put style classes on the labels so style rules could select specific labels, but it would also be totally unused right now, so I've left them classless.

Absotively:
The plan now is to move all this into content in https://github.com/space-wizards/space-station-14/pull/39161.

PR #6104, opened 2025-07-24
Add GridGet toolshed command

notquitehadouken:
I noticed there wasn't a toolshed command for getting the grid an entity was on or getting the name of the tile so I added them

Requires https://github.com/space-wizards/space-station-14/pull/39201

image


notquitehadouken:
...

PR #6108, opened 2025-07-26
feat: add IConfigurationManager.TrySaveToFile

perryprog:
Feels like the nicest way to deal with space-wizards/space-station-14#39196.

### DIFF
--- a/Robust.Shared/Configuration/IConfigurationManager.cs
+++ b/Robust.Shared/Configuration/IConfigurationManager.cs
@@ -64,6 +64,16 @@ namespace Robust.Shared.Configuration
         void SaveToFile();
 
         /// <summary>
+        /// Attempts to gracefully safe the configuration file to disk, failing
+        /// and returning false if no configuration file is loaded. Effectively
+        /// a wrapper around <see cref="SaveToFile" />.
+        /// </summary>
+        /// <remarks>
+        /// Will still throw if exceptions occur during saving.
+        /// </remarks>
+        bool TrySaveToFile();
+
+        /// <summary>
         /// Serializes a list of cvars to a toml.
         /// </summary>
         void SaveToTomlStream(Stream stream, IEnumerable<string> cvars);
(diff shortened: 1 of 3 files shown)


Robust.Shared/Configuration/IConfigurationManager.cs
```diff
@@ -63,6 +63,16 @@ public interface IConfigurationManager
         /// </summary>
         void SaveToFile();
 
+        /// <summary>
+        /// Attempts to gracefully safe the configuration file to disk, failing
+        /// and returning false if no configuration file is loaded. Effectively
+        /// a wrapper around <see cref="SaveToFile" />.
+        /// </summary>
+        /// <remarks>
+        /// Will still throw if exceptions occur during saving.
```
perryprog, line 72: Not sure if it's better to have this catch these exceptions and just return, but that feels weird to me.

perryprog:
Honestly this actually might be the wrong solution; it might be better to just downgrade the log from a warning to an info? Or something else?

PJB3005:
This should really be transparent to tests. Just make the configuration manager not do anything in tests, or something.

perryprog:
Hmm. I agree. Would I get in trouble if I made ClientNetConfigurationManager and SaveToFile virtual so we can have an inheritor registered for testing? I know it's Bad to make things virtual because tests require it but it seems the simplest way here.

PJB3005:
> Hmm. I agree. Would I get in trouble if I made ClientNetConfigurationManager and SaveToFile virtual so we can have an inheritor registered for testing? I know it's Bad to make things virtual because tests require it but it seems the simplest way here.

I would rather avoid more virtual nonsense on configuration manager.

perryprog:
Okay—in that case I'm not quite sure the best way to approach this then without duplicating basically everything in ClientNetConfigurationManager in order to get a test-specific implementation of NetConfigurationManager + IClientNetConfigurationManager. I guess a wrapper might work?

perryprog:
Oh word; that's a cool approach. Thanks!

PR #6113, opened 2025-07-30
Fix instrument pausing when outside PVS range

### LINKED ISSUE #6111: Midis do not play when exiting and re-entering PVS
When someone else is playing a midi, you leaving PVS and then re-entering the PVS range of the instrument will not cause the midi to resume playing.

After some debugging I discovered that MidiManager pausing the audio source is what causes this.
https://github.com/space-wizards/RobustToolbox/blob/bbbfcca303c1796167bc7dde2b59c2f8a76c38bb/Robust.Client/Audio/Midi/MidiManager.cs#L343

Removing this line of code makes MIDI instruments heard when exiting and re-entering PVS range of them.

I would PR a removal of those 2 lines, however I am unsure if that is even remotely a good solution. I tried looking for an Unpause method on the audio source but either I am blind as fuck or there is none.

FairlySadPanda:
# What

Revises MidiManager::UpdateRenderer(IMifiRenderer, MapCoordinates), tidying the code and making the handling of starting/stopping instruments work by using one method only (changing Gain).

Resolves #6111 

# Why

The code cleanup is just housekeeping.

This function clearly controls the _gain_ of a Renderer's Source. It also can control its state, but only to do one thing, and it does not handle undoing that thing. By pausing the audio source, this code assumes some other code will unpause it. That means that the state of the source has more places where it is modified. This increases complexity. In this case this broken assumption caused a bug.

# Potential Issues

Pausing the source may have been a performance choice or may have had some other meaning that was not discovered during exploration, triage and fix of this issue.

If so, this PR will need to be revised or replaced, and documentation should be added.

### DIFF
--- a/Robust.Client/Audio/Midi/MidiManager.cs
+++ b/Robust.Client/Audio/Midi/MidiManager.cs
@@ -42,7 +42,7 @@ internal sealed partial class MidiManager : IMidiManager
     [Dependency] private readonly IRuntimeLog _runtime = default!;
 
     private AudioSystem _audioSys = default!;
-    private SharedPhysicsSystem _broadPhaseSystem = default!;
+    private SharedPhysicsSystem _physics = default!;
     private SharedTransformSystem _xformSystem = default!;
 
     public IReadOnlyList<IMidiRenderer> Renderers
@@ -81,7 +81,7 @@ internal sealed partial class MidiManager : IMidiManager
     private Thread? _midiThread;
     private ISawmill _midiSawmill = default!;
     private float _gain = 0f;
-    private bool _volumeDirty = true;
+    private bool _gainDirty = true;
 
     // Not reliable until Fluidsynth is initialized!
     [ViewVariables(VVAccess.ReadWrite)]
@@ -96,7 +96,7 @@ internal sealed partial class MidiManager : IMidiManager
                 return;
 
             _cfgMan.SetCVar(CVars.MidiVolume, clamped);
-            _volumeDirty = true;
+            _gainDirty = true;
         }
     }
 
@@ -114,7 +114,8 @@ internal sealed partial class MidiManager : IMidiManager
         "/usr/share/sounds/sf2/TimGM6mb.sf2",
     };
 
-    private static readonly string WindowsSoundfont = $@"{Environment.GetEnvironmentVariable("SystemRoot")}\system32\drivers\gm.dls";
+    private static readonly string WindowsSoundfont =
+        $@"{Environment.GetEnvironmentVariable("SystemRoot")}\system32\drivers\gm.dls";
 
     private const string OsxSoundfont =
         "/System/Library/Components/CoreAudio.component/Contents/Resources/gs_instruments.dls";
@@ -145,11 +146,13 @@ internal sealed partial class MidiManager : IMidiManager
     {
         if (FluidsynthInitialized || _failedInitialize) return;
 
-        _cfgMan.OnValueChanged(CVars.MidiVolume, value =>
-        {
-            _gain = value;
-            _volumeDirty = true;
-        }, true);
+        _cfgMan.OnValueChanged(CVars.MidiVolume,
+            value =>
+            {
+                _gain = value;
+                _gainDirty = true;
+            },
+            true);
 
         _midiSawmill = _logger.GetSawmill("midi");
 #if DEBUG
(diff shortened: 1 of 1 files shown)


Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -81,7 +81,7 @@ public bool IsAvailable
     private Thread? _midiThread;
     private ISawmill _midiSawmill = default!;
     private float _gain = 0f;
-    private bool _volumeDirty = true;
+    private bool _gainDirty = true;
```
FairlySadPanda, line 84: Volume is not dirtied, Gain is.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -42,7 +42,7 @@ internal sealed partial class MidiManager : IMidiManager
     [Dependency] private readonly IRuntimeLog _runtime = default!;
 
     private AudioSystem _audioSys = default!;
-    private SharedPhysicsSystem _broadPhaseSystem = default!;
+    private SharedPhysicsSystem _physics = default!;
```
FairlySadPanda, line 45: Think this was a legacy name.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -114,7 +114,8 @@ public float Gain
         "/usr/share/sounds/sf2/TimGM6mb.sf2",
     };
 
-    private static readonly string WindowsSoundfont = $@"{Environment.GetEnvironmentVariable("SystemRoot")}\system32\drivers\gm.dls";
+    private static readonly string WindowsSoundfont =
```
FairlySadPanda, line 117: Some automatic linting has been done on this file.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -309,104 +316,80 @@ public void FrameUpdate(float frameTime)
...
+            if (!renderer.Source.Global)
+                UpdateLocalRenderer(renderer, listener);
+            else
+                UpdateGlobalRenderer(renderer);
+        }
+        catch (Exception ex)
+        {
+            _runtime.LogException(ex, _midiSawmill.Name);
+        }
+    }
```
FairlySadPanda, line 341: Drastically easier to read and means that the actually important logic isn't being sent rightward.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -309,104 +316,80 @@ public void FrameUpdate(float frameTime)
...
+        {
+            renderer.Source.Gain = 0f;
 
-            // Was previously muted maybe so try unmuting it?
-            if (renderer.Source.Gain == 0f)
-            {
-                renderer.Source.Gain = Gain;
-            }
+            return;
+        }
```
FairlySadPanda, line 350: The actual fix. "If this local renderer's entity is deleted or paused, mute the Source."

Yes, deletion being pretty fatal to the instrument means we could outright pause here. But we shouldn't extend the outcomes of this function unless we _know_ we need to - i.e. YAGNI and KISS.

Robust.Client/Audio/Midi/MidiManager.cs
```diff
@@ -309,104 +316,80 @@ public void FrameUpdate(float frameTime)
...
+            else
+                UpdateGlobalRenderer(renderer);
+        }
+        catch (Exception ex)
+        {
+            _runtime.LogException(ex, _midiSawmill.Name);
+        }
+    }
 
-            // If it's on a different map then just mute it, not pause.
```
FairlySadPanda, line 355: I've stripped most (?) of the comments out - they are all generally pretty useless towards helping the coder/reader.

VerinSenpai:
Any chance we could get a fix for containment fields disappearing outside of pvs range as well?

FairlySadPanda:
Totally different bug, sorry. That's a content issue involving a longboi entity.

VerinSenpai:
Oh I'm aware. Apologies I should've clarified, Im hoping someone more knowledgeable picks it up. One too many run-ins with the heart racing "WHERED THE GOD DAMN CONTAINMENT FIELD GO"

perryprog:
This doesn't explain why the bug doesn't occur on macOS's OpenAL implementation... 🤔 

I know it's a question that we don't need the answer to but it does bother me—if I don't forget I'll see if I can figure out why later.

PR #6115, opened 2025-07-30
Better unsubscription for multiple ConfigurationManager subscriptions

Fildrance:
# Why
Code that subscribes to a lot of configuration variables have to keep it as split methods to unsubscribe from it later. To enable using labdas while keeping unsub as an option - builder that can aggregate sub data is created. It can collected needed subs, then execute them, and return disposable object that will execute unsubscription with 1 method call.

# Technical considerations
Introduced new interface with copy-pasted methods from IConfigurationManager, sadly. It was needed to create fluent API for better UX. 

IDisposable can be replaced by custom type with 'Unsubscribe' method but at the same time we have a lot of IDE support for IDisposable and it might help.

Also fixed unsubscription (i have no idea how it worked previously, i feel like it was not).


PJB3005 (changes requested):
This should be entirely implemented with an external helper struct rather than a complicated extra interface woven into `IConfigurationManager`.

Robust.Shared/Configuration/ConfigurationManagerExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+        return this;
+    }
+
+    /// <summary>
+    /// Return disposable object that will execute unsubscription for each when disposed.
+    /// </summary>
+    public IDisposable Subscribe()
+    {
+        return new UnSubscribeActionsDelegates(_unSubscribeActions);
+    }
```
PJB3005: I do not see the point of having a separate `Subscribe` function like this, if the registrations are made immediately even without calling it. Just make this class `IDisposable`.

PJB3005 (changes requested):

Robust.Shared/Configuration/ConfigurationManagerExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        manager.OnValueChanged(name, onValueChanged, invokeImmediately);
+
+        _unSubscribeActions.Add(() => manager.UnsubValueChanged(name, onValueChanged));
+
+        return this;
+    }
+
+    /// <inheritdoc />
+    public void Dispose()
```
PJB3005, line 83: Clear the list at the end of this function.

Robust.Shared/Configuration/ConfigurationManagerExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+/// <summary>
+/// Container for batch-unsubscription of config changed events.
+/// Call Dispose() when subscriptions are not needed anymore.
+/// </summary>
+public sealed class ConfigurationMultiSubscriptionBuilder(IConfigurationManager manager) : IDisposable
+{
+    private readonly List<Action> _unSubscribeActions = [];
```
PJB3005: This should be named `_unsubscribeActions` instead. It's a single verb/noun/whatever.

Robust.UnitTesting/Shared/Configuration/ConfigurationManagerTest.cs
```diff
@@ -43,10 +43,76 @@ void ValueChanged(int value)
...
+            var lastValueBar1 = 0;
+            var lastValueBar2 = 0;
+            var lastValueBar3 = 0;
+            var lastValueBar4 = 0;
+
+            var subscription = mgr.SubscribeMultiple()
+                                  .OnValueChanged<int>("foo.bar", value => lastValueBar1 = value)
+                                  .OnValueChanged<int>("foo.bar", value => lastValueBar2 = value)
+                                  .OnValueChanged<int>("foo.bar", value => lastValueBar3 = value)
+                                  .OnValueChanged<int>("foo.bar", value => lastValueBar4 = value);
```
PJB3005: Indent, do not align.

Robust.UnitTesting/Shared/Configuration/ConfigurationManagerTest.cs
```diff
@@ -43,10 +43,76 @@ void ValueChanged(int value)
...
+
+            mgr.RegisterCVar("foo.bar", 5);
+            mgr.RegisterCVar("foo.foo", 2);
+
+            var lastValueBar = 0;
+            var lastValueFoo = 0;
+
+            var subscription = mgr.SubscribeMultiple()
+                                  .OnValueChanged<int>("foo.bar", value => lastValueBar = value)
+                                  .OnValueChanged<int>("foo.foo", value => lastValueFoo = value);
```
PJB3005: Indent, do not align.

PJB3005 (changes requested):

PJB3005 (approved):

PR #6116, opened 2025-07-31
add QueueDeleteMap to SharedMapSystem

slarticodefast:
Some content systems are creating new paused maps to store entities on. Examples are `PolymorphSystem` and `CryostorageSystem`, and the same is needed for the changeling.

For integration test and cleanup reasons these maps should be deleted again when no longer needed - that is when no more entities are stored on them. However, directly deleting the map using `DeleteMap` does not work, as we are doing this in an event subscription and deleting the map would instantly delete the subscribing entity.

So a deferred map deletion is needed.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -39,7 +39,7 @@ END TEMPLATE-->
 
 ### New features
 
-*None yet*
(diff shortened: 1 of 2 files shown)


Tayrtahn (approved):
Seems pretty straightforward to me.

PR #6117, opened 2025-07-31
feat: add analyzer for correct AfterAutoHandleStateEvent usage

### LINKED ISSUE #4846: There is no warning or error when subscribing to AfterAutoHandleStateEvent on a component without AutoGenerateComponentState(true)
If you do this, it does nothing since the event is never fired.
This shouldn't be checked on full release or outside of debug.

perryprog:
I try so hard to dislike C# but then Roslyn has to exist and it's so frickin' nice, ugh.

Fixes #4846 because this has caught me at least twice now and it's way too easy to forget to pass that silly `true` parameter.

This is hard coded as flip because I can't think of a good way to make it generic via attributes, and I'm not sure that that would make sense anyway. (It'd have to be something like `public AutoGenerateComponentStateAttribute([AllowsEventWhen(true, AfterAutoHandleStateEvent)] bool raiseAfterAutoHandleState = false, bool fieldDeltas = false)`, but this feels so hyper-specific to this one scenario that I'm not sure it's worth it.)

Requires space-wizards/space-station-14#39289 for content tests.

Bikeshedding on diagnostic strings is welcome since this is very much a new-contributor trap and it should be super clear how to fix this. I'm also a little miffed that apparently diagnostic descriptions don't actually seem to show up anywhere in at least Rider, which is where I'd like to have that info...


Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
```diff
@@ -0,0 +1,89 @@
...
+using static Microsoft.CodeAnalysis.SymbolEqualityComparer;
+
+namespace Robust.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class AfterAutoHandleStateAnalyzer : DiagnosticAnalyzer
+{
+    private const string AfterAutoHandleStateEventName = "AfterAutoHandleStateEvent";
+    private const string AutoGenStateAttribute = "Robust.Shared.Analyzers.AutoGenerateComponentStateAttribute";
+    private const string SubscribeLocalEventName = "SubscribeLocalEvent";
```
perryprog, line 15: I know this isn't done in other analyzers, but would it not be better to compile in the relevant definitions into Robust.Analyzers so we could use `nameof(AfterAutoHandleStateEvent)` and `typeof(AutoGenerateComponentStateAttribute).FullName`?

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
```diff
@@ -0,0 +1,89 @@
...
+            compilationContext.RegisterOperationAction(CheckEventSubscription, OperationKind.Invocation));
+    }
+
+    private static void CheckEventSubscription(
+        OperationAnalysisContext context)
+    {
+        if (context.Operation is not IInvocationOperation operation)
+            return;
+
+        if (!operation.TargetMethod.Name.Contains(SubscribeLocalEventName))
```
perryprog: Hm, this is going to check _any_ method that contains 'SubscribeLocalEvent' (and has at least two type arguments), not only those from e.g., EntitySystem. The pro of this is that it catches usages via Subs.SubscribeLocalEvent... which is _barely_ used.

Happy to change it (and ditto for AfterAutoHandleStateEvent) to be better qualified if that's preferred.

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
PJB3005: Yeah that doesn't make much sense and isn't practically possible. You already have a test that should cover it so this is fine.

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
PJB3005: Shrug it's probably fine.

Robust.Analyzers.Tests/AfterAutoHandleStateAnalyzerTest.cs
```diff
@@ -0,0 +1,119 @@
...
+using Microsoft.CodeAnalysis.Testing;
+using NUnit.Framework;
+using VerifyCS =
+    Microsoft.CodeAnalysis.CSharp.Testing.CSharpAnalyzerVerifier<Robust.Analyzers.AfterAutoHandleStateAnalyzer,
+        Microsoft.CodeAnalysis.Testing.DefaultVerifier>;
+
+namespace Robust.Analyzers.Tests;
+
+[Parallelizable(ParallelScope.All | ParallelScope.Fixtures)]
+[TestFixture, TestOf(typeof(ByRefEventAnalyzer))]
```
Tayrtahn: Copy-paste mistake!

perryprog: exposed 😞

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
Tayrtahn: Why `Contains` and not `==`?

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
```diff
@@ -0,0 +1,89 @@
...
+            return;
+
+        var subscriptionTypes = operation.TargetMethod.TypeArguments;
+        // Check second arg of SubscribeLocalEvent is AfterAutoHandleStateEvent
+        if (subscriptionTypes.ElementAtOrDefault(1)?.Name.Contains(AfterAutoHandleStateEventName) != true)
```
Tayrtahn: Same question

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
```diff
@@ -0,0 +1,89 @@
...
+            return;
+
+        var autoGenStateAttribute = context.Compilation.GetTypeByMetadataName(AutoGenStateAttribute);
+        if (autoGenStateAttribute == null)
+            return;
```
Tayrtahn: This could do the neat little trick that we've been using for analyzers recently where this lookup is done in a compilation start action and passed into the analyzer method. That avoids having to do the lookup with every invocation and also lets the analyzer just not run for any assemblies where the attribute isn't available.

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        // If we have a second type arg, we definitely have a first.
+        // Get the attributes for whatever that is, and then get its attributes.
+        var component = subscriptionTypes[0];
+        var autoGenAttribute = component
+            .GetAttributes()
+            .FirstOrDefault(attribute => attribute.AttributeClass?.Equals(autoGenStateAttribute, Default) ?? false);
```
Tayrtahn: This probably could/should use the helper in `AttributeHelper`

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        // First argument is raiseAfterAutoHandleState—note it shouldn't ever
+        // be null, since it has a default, but eh.
+        if (autoGenAttribute?.ConstructorArguments[0].Value is true)
```
Tayrtahn, line 78: Oo, is this using `is` instead of `==` to force pattern matching so it does both a type check and a value check? Because that's pretty cool and I'm gonna have to use it!

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
perryprog: Because for some reason I thought MethodSymbol.Name would have the same weird formatting of MemberInfo.Name with the bit with the backtick. It does not. 🙃

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
perryprog: Yup!

Tayrtahn (approved):
lgtm

Robust.Analyzers/AfterAutoHandleStateAnalyzer.cs
perryprog: Honestly now that I think about it `if (autoGenAttribute is { ConstructorArguments: [{ Value: true }, ..] })` is kinda clearer to me but I know not everyone likes pattern matching. I'm just not a fan of indexing with static values like this when you can just use a list pattern.

Issue #3768, opened 2023-02-10
No way to (de)serialize TimeSpans as time offsets if they are stored in a collection


0x6273:
There is currently no way to have data fields that store time offsets in a collection be (de)serialized as time offsets. For example:

```cs
[RegisterComponent]
public sealed class MyComponent : Component
{
    // this works
    [DataField("myTimeOffset", customTypeSerializer: typeof(TimeOffsetSerializer))]
    public TimeSpan MyTimeOffset;

    // no way to do this
    [DataField("myTimeOffsets", customTypeSerializer: typeof(?????))]
    public Dictionary<string, TimeSpan> MyTimeOffsets = new Dictionary<string, TimeSpan>();
}

```

For prototype IDs, there are a bunch of different versions for different collections, like `PrototypeIdDictionarySerializer`.

- Should there be versions of `TimeOffsetSerializer` for every collection, like there is for prototype IDs?
- Alternatively, should time offset be its own separate data type, that wraps or inherits TimeSpan?
- Could this be solved with generics somehow, like `customTypeSerializer: typeof(DictionarySerializer<StringSerializer, TimeOffsetSerializer>)`? (currently, `DictionarySerializer` takes the actual data types rather than the serializer types)

PR #6119, opened 2025-08-01
add: Dictionary<T, TimeSpan> OnUnpaused generator

ghost:
This only adds code generator. Doesn't do anything about #3768 but if fixing that is still required, I can try adding it here.


PJB3005 (changes requested):
This needs tests.

PJB3005 (approved):

PR #6114, opened 2025-07-30
Re-add `WeakEntityReference`

Tayrtahn:
A revival of https://github.com/space-wizards/RobustToolbox/pull/5577.

This version is based around an `EntityUid` rather than a `NetEntity` and includes the needed modifications to `ComponentNetworkGenerator` to make that possible (by converting the underlying `EntityUid` to a `NetEntity` only for networking purposes).

PR #6123, opened 2025-08-03
Treat deleted `EntityUid?` as a null `NetEntity?`

Tayrtahn:
I'm proposing this as an alternative to [`WeakEntityReference`](https://github.com/space-wizards/RobustToolbox/pull/6114).

While trying to convert `WeakEntityReference` from being based around a `NetEntity` to being based around an `EntityUid`, I found that I was duplicating a lot of code with the way that `EntityUid`s are converted to `NetEntity`s for component state networking. I also got confused a lot about what exactly the different types represent: what are the differences between a `WeakEntityReference`, a `WeakEntityReference?`, and an `EntityUid?`; does a `WeakEntityReference?` even make sense?

The main issue that `WeakEntityReference` is trying to solve is the massive spam of PVS errors when the auto state generator tries to get state for a component with an `EntityUid?` DataField that has been deleted.

We've been working with the idea that those DataFields should be set to null when the entity they refer to has been deleted. This is a pain to handle in many cases, as it requires the referenced entity to have its own inverse reference solely to inform the referring entity when it gets deleted. 

But why? **Isn't a nullable `EntityUid` essentially a weak reference to an entity?** In purely server-side code, we use `!Deleted(comp.NullableEnt)` or similar to check that a nullable `EntityUid` field is valid before using it. Why do we need to treat it differently for PVS states?

This PR changes the behavior of `EntityManager.GetNetEntity(EntityUid?)` so that passing in a deleted entity returns null instead of logging an error and returning `NetEntity.Invalid`. Since the related methods for getting arrays, lists, etc of `NetEntity`s call that method, they are also changed. Methods working with non-nullable `EntityUid`s are not changed and will still log an error if called on a deleted entity (which I think makes sense).

This seems to me like more logical behavior - a reference to a deleted entity is a reasonable value for an `EntityUid?` and getting component state for it shouldn't be considered an error. 

My content `WeakEntityReference` conversion PRs include tests that produce component states with references to deleted entities and check for PVS errors. This change allows those tests to pass without needing any modifications to the components they test - meaning that this change should fix all of the related errors without needing to handle them individually.

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.Network.cs
+++ b/Robust.Shared/GameObjects/EntityManager.Network.cs
@@ -195,6 +195,9 @@ public partial class EntityManager
         if (uid == null)
             return null;
 
+        if (!MetaQuery.Resolve(uid.Value, ref metadata, logMissing: false))
+            return null;
+
         return GetNetEntity(uid.Value, metadata);
     }
 
--- a/Robust.UnitTesting/Shared/GameState/ComponentStateTests.cs
+++ b/Robust.UnitTesting/Shared/GameState/ComponentStateTests.cs
@@ -264,7 +264,7 @@ public sealed partial class ComponentStateTests : RobustIntegrationTest
             Assert.That(clientEntB, Is.EqualTo(EntityUid.Invalid));
             Assert.That(client.EntMan.EntityExists(clientEntA), Is.True);
             Assert.That(client.EntMan.TryGetComponent(clientEntA, out UnknownEntityTestComponent? cmp));
-            Assert.That(cmp?.Other, Is.EqualTo(EntityUid.Invalid));
+            Assert.That(cmp?.Other, Is.Null);
         });
 
         server.Post(() => server.CfgMan.SetCVar(CVars.NetPVS, false));


slarticodefast:
I very much prefer this approach, because otherwise we will have to convert hundreds of components in the content repo to use `WeakEntityReference`, which will need a major cleanup effort.

And I would argue this PR is of critical importance given that this causes about 10000 errors each day on our servers.

slarticodefast:
And a question, how will this implementation handle something like a networked `Dictionary<EntityUid, string>`? Ideally it would automatically remove those entries when sending them to the client.

Tayrtahn:
> And a question, how will this implementation handle something like a networked `Dictionary<EntityUid, string>`? Ideally it would automatically remove those entries when sending them to the client.

I think that's actually a separate problem that can be fixed by adding special handling for `Dictionary<EntityUid?, T>` to the component state code generator. If this PR is approved, I could do that too.

PJB3005:
Nope. Not happening. An `EntityUid` should always be valid if we have one.

slarticodefast:
> Nope. Not happening. An EntityUid should always be valid if we have one.

A question about this for clarification: Wouldn't this implementation just affect nullable `EntityUid`s though?
A non-nullable `EntityUid` would still work as you are describing and cause a PVS error.
Or am I misunderstanding this?

PJB3005:
> > Nope. Not happening. An EntityUid should always be valid if we have one.
> 
> A question about this for clarification: Wouldn't this implementation just affect nullable `EntityUid`s though? A non-nullable `EntityUid` would still work as you are describing and cause a PVS error. Or am I misunderstanding this?

This changes nothing. An `EntityUid?` is an `EntityUid` in a `Nullable<T>`. The semantics of `EntityUid` should not be different juts because it's in a container.

PR #6124, opened 2025-08-03
Add `SharedOnlyEventAttribute` and analyzer

Tayrtahn:
Adds an attribute that can be applied to event structs/classes to indicate that they should only be subscribed to in Shared code. Subscriptions to the event in Server-only code will raise a warning.

Applying the attribute to `GetVerbsEvent`:
Screenshot 2025-08-03 at 6 34 31 PM

Warning on a server-only subscription:
Screenshot 2025-08-03 at 3 53 17 PM

The attribute also has an optional parameter, `AllowClientOnly` which defaults to true. If true, subscriptions are also allowed in Client-only code. In most potential use cases (verbs, examines), Client-only subscriptions are fine.

This feature was requested by @slarticodefast to help enforce https://github.com/space-wizards/space-station-14/issues/39286 - subscribing to `GetVerbsEvent` in server-only code causes the verbs menu to jump around, because it is first populated client-side with Client and Shared verbs, then gets rebuilt when the server sends its version containing new entries. This is a major annoyance for players, and has caused problems where admins accidentally trigger actions on players while trying to use examining verbs on them.


VerinSenpai:
Good freakin shit man!

Robust.Analyzers/SharedOnlyEventAnalyzer.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            ctx.RegisterOperationAction(opContext => AnalyzeInvocation(opContext, sharedSubscribeAttribute), OperationKind.Invocation);
+        });
+    }
+
+    public static void AnalyzeInvocation(OperationAnalysisContext context, INamedTypeSymbol sharedSubscribeAttribute)
+    {
+        // If the subscription is in Shared code, then we don't care about it
+        if (context.ContainingSymbol.ContainingNamespace.ToString().Contains("Shared"))
+            return;
```
PJB3005, line 48: Check this in `RegisterCompilationStartAction`.

TornadoTechnology:
This engine clearly needs more cool analyzers. Awesome.

PJB3005 (changes requested):
Would it not make more sense to have an attribute that worked with bitflags? Like `[NetworkSideOnly(Shared | Client)]`?

Robust.Analyzers/SharedOnlyEventAnalyzer.cs
Tayrtahn: That's what I wanted to do originally, but the test assembly naturally doesn't match and it winds up breaking the tests. Do you happen to know if there's a way to control the assembly name for tests?

Tayrtahn:
> Would it not make more sense to have an attribute that worked with bitflags? Like `[NetworkSideOnly(Shared | Client)]`?

Yeah, I was thinking that too. Figured I'd do it this way first and upgrade to that if desired. I'll get on it.

Robust.Analyzers/SharedOnlyEventAnalyzer.cs
PJB3005: No, because that doesn't make sense. I would just skip in test assemblies.

PR #6137, opened 2025-08-13
make SharedAudioSystem.Stop not return early when the current tick has already been predicted

### LINKED ISSUE #35081: Audio from planet/map weather persists even after the associated WeatherComponent is deleted
## Description
When there is active weather on a map, deleting the associated _(server-side)_ `WeatherComponent` doesn't stop weather audio (if you are on the map and are able to hear it in the first place), meaning you will hear it forever.
 Did not test if this effect remains after reconnecting.

**Reproduction** 
0. Have working audio..
1. Find id of the map (not grid, but map) that you're on
2. Set weather of that map to anything of your choice other than `'null'`
3. Optionally wait the full hardcoded 15 seconds for weather to fully transition from clear weather
4a. Press F7 and go to objects
4b. Select 'Maps'
4c. Right-click on the map with corresponding map id and open View Variables
5. Go to Server Components tab
6. Search for `WeatherComponent`
7. Delete `WeatherComponent` (by pressing the X on the right-hand side)
8. Notice that the audio of the weather stays and changes to full volume regardless of whether you are in an indoor space (where it is usually muffled),  *the actual weather effect is removed*

**Screenshots**
Would add media but I can't do so right now.

**Additional context**
You can also notice this issue with #34973 (steps to test are provided in the comments), by FTLing early from the expedition planet and noticing persisting audio despite having left the planet

LaCumbiaDelCoronavirus:
title; if unclear about what i'm doing, read diff

tested on latest ss14 and did not observe any difference at all so it was infact something that needed to be fixed ( as said in https://github.com/space-wizards/space-station-14/pull/37573#discussion_r2108987468 )

would fix https://github.com/space-wizards/space-station-14/issues/35081 if merged with https://github.com/space-wizards/space-station-14/pull/37573

### DIFF
--- a/Robust.Shared/Audio/Systems/SharedAudioSystem.cs
+++ b/Robust.Shared/Audio/Systems/SharedAudioSystem.cs
@@ -455,7 +455,7 @@ public abstract partial class SharedAudioSystem : EntitySystem
         if (uid == null || !Resolve(uid.Value, ref component, false))
             return null;
 
-        if (!Timing.IsFirstTimePredicted || (_netManager.IsClient && !IsClientSide(uid.Value)))
+        if (_netManager.IsClient && !IsClientSide(uid.Value))
             return null;
 
         QueueDel(uid);


LaCumbiaDelCoronavirus:
+1-1 diff and 3 months waiting 🙏

kontakt:
This change precipitated this issue:
https://github.com/space-wizards/space-station-14/issues/42153

kontakt:
I don't understand why this is the correct approach, what's the benefit of making stop call repeatedly if predicted repeatedly?
As in the linked issue - `Stop` and `PlayPredicted` are being used in a predicted instance.
`PlayPredicted` uses first time predicted and will only occur once. However, `Stop` is not first time predicted and will override a previous prediction's `PlayPredicted` and the audio just doesn't play.

PR #6140, opened 2025-08-14
feat: duplicate localization keys in downstream namespaces no longer generate error logs

little-meow-meow:
Previously the code collected occurrences of duplicate localization keys and generated log errors which induced test failure. This is good for discovering unintended duplicates, but downstream codebases sometimes intend to override localizations.

This change checks if the resource directory begins with an underscore (as is custom in downstream codebases) and does not attempt to emit log errors for duplicate keys. Consequently they will no longer be informed of duplicated keys even outside of the main namespace. Perhaps this is an acceptable tradeoff.

### DIFF
--- a/Robust.Shared/Localization/LocalizationManager.cs
+++ b/Robust.Shared/Localization/LocalizationManager.cs
@@ -467,6 +467,13 @@ namespace Robust.Shared.Localization
             {
                 var errors = resource.Errors;
                 WriteWarningForErrs(path, errors, data);
+
+                if (path.CanonPath.StartsWith($"{LocaleDirPath}/{culture.Name}/_"))
+                {
+                    context.AddResourceOverriding(resource);
+                    continue;
+                }
+
                 if (!context.InsertResourcesAndReport(resource, path, out var errs))
                 {
                     resErrors.AddRange(errs);


lzk228:
i am against this
- the main thing is it will be very hard  to spot duplicates on upstream for those who don't really change the original code and not only for them
- even if a downstream has to do some overrides, it can just comment the original keys or write needed text in them

little-meow-meow:
> the main thing is it will be very hard to spot duplicates on upstream for those who don't really change the original code and not only for them

I'm unsure what you mean by this. To be clear, this pull request causes localization keys defined in files from downstream folders not to generate duplication errors. If a downstream redefines a localization key that already exists in an upstream file, it's assumed to be intentional.

The downsides to this:
- Downstreams won't have the same localization duplication detection, either in their own files or in ported-from-downstream files.
- This is a sneaky change to drop on downstreams (who will also have no way to opt out of it if they don't want this.)

My use case is that my repository touches very, very many upstream localizations. The existing system burdens us way more than it helps.

I have zero hope that Space Wizards will ever merge this, but I would like to eventually see something done to address this. Hopefully with a more elegant solution. As for now, there's nothing else I can do.

PR #6142, opened 2025-08-15
feat: enable adding custom command keybinding for add text console commands keybindings

Fildrance:
Prerequisite to enable adding custom console commands keybinding.
Added way to save Keybinding text for command separate from FunctionName (was same field previously)
Added methods for input manager to get next custom command function, which is free from binding, and enumerate already existing custom commands.

### DIFF
--- a/Robust.Client/Input/IInputManager.cs
+++ b/Robust.Client/Input/IInputManager.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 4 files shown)


Fildrance:
Requires Changelog entry.
TECHNICALLY new methods on input manager could be transported into Content part, but i am a bit afraid of that option.

PJB3005 (changes requested):
If you want to have something like this, you should *really* just add a way to add metadata to keybinds directly.

PJB3005:
Stale.

PR #6143, opened 2025-08-15
Add helper methods for entity fixture shape radii

kosticia:
This pr adds new helper methods that makes interaction with fixture radius more comfortable and also allowing to avoid duplicating code in the future.
Required for the proper implementation of the [Make drop range same as actual interaction range #39188](https://github.com/space-wizards/space-station-14/pull/39188).

### DIFF
--- a/Robust.Shared/Physics/Systems/FixtureSystem.cs
+++ b/Robust.Shared/Physics/Systems/FixtureSystem.cs
@@ -341,11 +341,11 @@ namespace Robust.Shared.Physics.Systems
         }
 
         /// <summary>
-        /// Gets the sum of the biggests fixture radii from multiply entities
+        /// Gets the sum of the biggest fixture radii from multiple entities.
         /// </summary>
-        public float SumFixtureRadii(params EntityUid[] uids)
+        public float SumMaxFixtureRadii(params EntityUid[] uids)
         {
-            float sum = 0f;
+            var sum = 0f;
             foreach (var uid in uids)
             {
                 var fixRad = GetMaxFixtureRadius(uid);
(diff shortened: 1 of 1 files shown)


Robust.Shared/Physics/Systems/FixtureSystem.cs
```diff
@@ -340,6 +340,54 @@ private void OnHandleState(EntityUid uid, FixturesComponent component, ref Compo
             }
         }
 
+        /// <summary>
+        /// Gets the sum of the biggests fixture radii from multiply entities
```
Tayrtahn:
```suggestion
        /// Gets the sum of the biggest fixture radii from multiple entities.
```

Robust.Shared/Physics/Systems/FixtureSystem.cs
```diff
@@ -340,6 +340,54 @@ private void OnHandleState(EntityUid uid, FixturesComponent component, ref Compo
...
+        /// </summary>
+        public float SumFixtureRadii(params EntityUid[] uids)
```
Tayrtahn: Since this is the sum of each entity's max fixture radius, the name should reflect that. `SumMaxFixtureRadii` perhaps.

Robust.Shared/Physics/Systems/FixtureSystem.cs
```diff
@@ -340,6 +340,54 @@ private void OnHandleState(EntityUid uid, FixturesComponent component, ref Compo
...
+            {
+                var fixRad = GetMaxFixtureRadius(uid);
+                sum += fixRad;
+            }
+
+            return sum;
+        }
+
+        /// <summary>
+        /// Gets the biggest radius of the fixture from multiply entities.
```
Tayrtahn:
```suggestion
        /// Gets the biggest fixture radius of multiple entities.
```

Robust.Shared/Physics/Systems/FixtureSystem.cs
```diff
@@ -340,6 +340,54 @@ private void OnHandleState(EntityUid uid, FixturesComponent component, ref Compo
...
+        public float GetMaxFixtureRadius(params EntityUid[] uids)
+        {
+            List<float> radii = [];
+            foreach (var uid in uids)
+            {
+                var fixRad = GetMaxFixtureRadius(uid);
+                radii.Add(fixRad);
+            }
+
+            return radii.Max();
```
Tayrtahn: This doesn't need to allocate a list of floats. Just track the biggest radius as you iterate the entities.

Robust.Shared/Physics/Systems/FixtureSystem.cs
```diff
@@ -340,6 +340,54 @@ private void OnHandleState(EntityUid uid, FixturesComponent component, ref Compo
...
+        }
+
+        /// <summary>
+        /// Gets the biggest radius of the fixture from one entity.
```
Tayrtahn:
```suggestion
        /// Gets the biggest fixture radius from one entity.
```

Robust.Shared/Physics/Systems/FixtureSystem.cs
```diff
@@ -340,6 +340,54 @@ private void OnHandleState(EntityUid uid, FixturesComponent component, ref Compo
...
+        /// </summary>
+        public float GetMaxFixtureRadius(EntityUid uid, FixturesComponent? manager = null)
```
Tayrtahn, line 377: Use `Entity<FixturesComponent?>`

Robust.Shared/Physics/Systems/FixtureSystem.cs
```diff
@@ -340,6 +340,54 @@ private void OnHandleState(EntityUid uid, FixturesComponent component, ref Compo
...
+                return 0f;
+
+            List<float> radii = [];
+            foreach (var fixture in manager.Fixtures.Values)
+            {
+                var fixRad = fixture.Shape.Radius;
+                radii.Add(fixRad);
+            }
+
+            return radii.Max();
```
Tayrtahn: Like above, don't allocate a list of floats; track the max as you go.

Tayrtahn (changes requested):

Robust.Shared/Physics/Systems/FixtureSystem.cs
kosticia: But I don't really like use `ent.Owner` instead of just `ent` :(

metalgearsloth:
Re-open if review resolved + move it to physicssystem not fixturesystem please.

PR #6144, opened 2025-08-16
Tile border enhancements

sowelipililimute:
### implementation
- ITileDefinitions now have a BordersMode field that affects rendering
  - `Exterior8Patch`: old behaviour, up to 8 sprites on the neighbouring tiles
  - `Interior8Patch`: up to 8 sprites on the current tile based on neighbours
  - `Interior4Of16`: up to 4 sprites on the current tile out of 16 variants based on neighbours
- `Interior4Of16EdgeSprites` that holds the sprites for the 4-of-16 mode

### content build failure
- this needs new storage for the 4-of-16 sprites and i'm not about to hijack the Direction enum for that

### what it looks like
grafik

left: interior 8-patch
right: interior 4-of-16
each colour is its own image


AftrLite:
🤤 yummy.. yummers.. spriters eating good today... (i'm spriters)

Princess-Cheeseballs (approved):
Looks good, 
The massive if statement is horrific and frightening, but it's readable and optimizing it would reduce its readability for like only nanoseconds of gain.

PR #6147, opened 2025-08-17
fix(serializer): remove invalid entity uids from components

little-meow-meow:
This behavior is linked to a series of issues and fixes: 

- space-wizards/space-station-14#22845
- space-wizards/space-station-14#28274
- space-wizards/space-station-14#29035
- space-wizards/space-station-14#31416
- space-wizards/space-station-14#33359
- space-wizards/space-station-14#33664

The following PRs have cleaned up action entities that became serialized in map files:

- space-wizards/space-station-14#28589
- space-wizards/space-station-14#37457

Actions probably shouldn't be serialized, but marking `BaseAction` as `savable: false` causes entities to serialize with invalid references.

Is there ever a good reason to be serializing invalid entity UIDs? If there is and there's no way to change that then we'll probably want an attribute to mark fields for exclusion from serialization.

I tried to make the entity serializer not emit nodes for invalid entity references but couldn't figure out a nice way to do this. I assume this is why invalid references get serialized in the first place. So here's a hack that recurses serialized components and removes invalid entity references. Obviously this would be better handled some other way, but I didn't have it in me to refactor the serializer.

### DIFF
--- a/Robust.Shared/EntitySerialization/EntitySerializer.cs
+++ b/Robust.Shared/EntitySerialization/EntitySerializer.cs
@@ -509,6 +509,9 @@ public sealed class EntitySerializer : ISerializationContext,
                 // implicitly be saved as anchored.
                 compMapping = _serialization.WriteValueAs<MappingDataNode>(compType, component, alwaysWrite: true, context: this);
 
+                // Prune invalid entity references so we don't cause a zillion unforeseen problems after deserializing
+                RemoveInvalidEntityReferences(compMapping);
+
                 // This will not recursively call Except() on the values of the mapping. It will only remove
                 // key-value pairs if both the keys and values are equal.
                 compMapping = compMapping.Except(protoMapping);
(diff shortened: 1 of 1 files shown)


PJB3005:
Yeah this is not a good solution. The correct fix is WeakEntityReference and maybe something else for device networks. Not this.

little-meow-meow:
> The correct fix is WeakEntityReference and maybe something else for device networks.

Will you elaborate on this?

PJB3005:
> > The correct fix is WeakEntityReference and maybe something else for device networks.
> 
> Will you elaborate on this?

What part needs elaborating on?

little-meow-meow:
Is there a functional purpose for serializing `EntityUid.Invalid`?

PJB3005:
> Is there a functional purpose for serializing `EntityUid.Invalid`?

Invalid EntityUids and null EntityUids are fundamentally different concepts. The point is that a bug has *already* occurred when an invalid EntityUid was serialized. Trying to turn them into null or whatever will simply mean that the symptoms of that bug may then be hidden.

For cases where invalid EntityUids are not a bug, those cases *should* be moved to something like `WeakEntityReference` instead. That makes the intent clear: there is no difference between "invalid" (this was an entity, but it got severed on map save) and "no entity"

PR #6155, opened 2025-08-20
Make SpriteSpecifier.Texture fail validation if it contains ".rsi"

PJB3005:
No pointing to PNGs inside RSIs.


PJB3005:
Will cause content CI failure until https://github.com/space-wizards/space-station-14/pull/39783 is merged

PR #6130, opened 2025-08-05
Fixing the limit for networked Vars.

VigersRay:
I think it's a typo. Exception show limit for short.MaxValue (32767). But checked byte.MaxValue (255)
image

255 network vars is terribly little, RMC already has 200+.
image

PR #6157, opened 2025-08-21
Fix crash on startup with more than 255 CVars

DrSmugleaf:
Guess what overly specific dumb engine limit RMC14 just hit again
Multiplied MaxMessageSize by an arbitrary 2 because it seems slightly more sane than multiplying it by 128 at a glance


VigersRay:
https://github.com/space-wizards/RobustToolbox/pull/6130

PR #6160, opened 2025-08-23
More accurate & controllable pointlight attenuation

mirrorcult:
## what why

requires https://github.com/space-wizards/space-station-14/pull/39853 (sort of, its not strictly necessary. see below)

the current pointlight attenuation function used in the light shader is pretty bad in a couple of ways.

qualitatively, lights with essentially any radius look quite strange; the way lights attenuate at the boundary just kind of seems 'wrong' in some way, like its sharper than it should be or 'not smooth'. furthermore, small lights' intensity seems to decrease basically linearly over their radius, instead of in any physically accurate way, which makes things like APC or vendor lights look weird no matter what their energy is. obviously our light attenuation doesnt *necessarily* need to be **A) physically accurate**, it just needs to **B) be artistically controllable** and **C) look good**, but the current function unfortunately meets none of these 3 conditions (imo)

with regards to the implementation, our attenuation function has a couple significant issues. firstly, there is no way to actually control the shape of the curve for artistic reasons: the only relevant parameters to be tweaked are light radius and energy, which dont really help us at all. secondly, the actual function used has a pretty big issue... **because negative light values are clamped to 0, the function is only [c0 continuous](https://en.wikipedia.org/wiki/Smoothness#Example:_continuous_(C0)_but_not_differentiable)!!** my assumption is this is the reason that the boundary behavior 'feels wrong' in some way. it's less egregious at higher light radii, but at lower ones it's really bad.

<p align="center">
image
</p>
<p align="center">
<i>the attenuation curve currently on master<br>the function is c0 continuous, but not c1 continuous (the derivative has a discontinuity at x=1)</i>
</p>

<p align="center">
image
</p>
<p align="center">
<i>the attenuation curves in this PR<br>nice and smooth</i>
</p>

a different function could solve this problem entirely! thankfully this is not a unique problem, i tried a couple things that all felt not really correct and then googled around a bunch until i eventually stumbled upon this post: https://lisyarus.github.io/blog/posts/point-light-attenuation.html which contains two (related) pretty good pointlight attenuation functions that meet all 3 conditions above, as well as being c1 continuous!

this PR replaces the current implementation in the shader with this new function, and introduces two new parameters to pointlights that allow you to adjust the falloff parameter, as well as adjust the curve shape, which primarily affects how light falls off near its center (see below for why this is desirable), modifying it is equivalent to lerping between the two curves in the post above

**you can compare the different curves at different light radii and with different falloff values and curve factors using this desmos calculator: https://www.desmos.com/calculator/yjudaha0s6**

## uh, isn't it going to look weird in SS14 since its completely different now

yes, if it was untweaked. however, the new curves are actually not so dissimilar from the old ones in the sense that they still fill space well, so for the most part the only tweak I had to make to get them looking similar was to change the default falloff value to `6.8`, which kept "normal station lights" about the same brightness as before.

i am unconcerned with trying to "improve lighting quality" in SS14 with this PR: I went on this endeavor for a fork and I plan on using a much higher falloff value there for most lights, so the goal is to just make the uncontroversial improvements (c1 continuity and artistic control) while still looking more or less equivalent to current lights.

*however*, there is one problem. small lights. because the curve for small lights previously was so weird (it was.. basically linear, as said before, and the derivative discontinuity was extremely apparent) they looked very strange in game: like static circles of light that somehow pierce above regular station light even at very low energy levels. because of those very low energy levels, once their attenuation curves were fixed to be more reasonable, they're *veeeery* dim. 

so, this PR does have a companion content PR to increase the energies of various small point lights: https://github.com/space-wizards/space-station-14/pull/39853

it's not strictly required, but it's recommended because as I've said although I think these small lights look much better being dimmer I am not interested in trying to justify that opinion upstream and would prefer they just look about the same brightness as before. its still not perfect but I am not auditing every single light in content.

another relevant issue: the old curve actually didn't start tapering off until a little bit out from the light's center (see the desmos above). this is unrealistic but does have artistic benefits in representing "spherical" pointlights, like a glowing orb, where taking a little bit extra time to falloff makes sense (since its a sphere, not a point). this is actually the reason I decided to include `curveFactor` as an option: it lets you replicate this behavior (and anything inbetween) if you want it, so I would recommend messing with it for things of that nature.

## okay show examples

sure. here's comparisons between the old attenuation and the new attenuation + attached content PR:
https://slow.pics/c/d3RVxm4o

embedded in github (i would recommend just looking at it above though), old on the left new on the right:

<p align="center">

</p>

<p align="center">

</p>

<p align="center">

</p>

if you're having trouble spotting the difference, well, good i guess, but look at the edges of lights, and at small lights

here's a video showing the falloff for lights with curve factor  0 (left) and curve factor 1 (right) being animated from `1` to `20` and back. you can tell that the right one keeps more light around the center (the animations are desynced from eachother sorry):

https://github.com/user-attachments/assets/5d675709-3e5c-4414-9d92-6e70df4f86ff

here's a video showing curve factor being animated from 0 to 1 with a constant falloff (its kinda subtle):

https://github.com/user-attachments/assets/7db52368-7e8f-4d2c-9d04-fa12d15e143e

not going to embed this since this isn't accurate to what upstream will look like, but heres some comparisons with a higher falloff value of 20 (which is what I want to use in a fork): https://slow.pics/c/Fu2CKm54

## technical concerns

i'm pretty sure this PR is sound but there are a couple potential concerns I wasn't positive about / wanted to expound on here
- ternaries in glsl dont branch, right? they just select or whatever? im pretty sure they don't branch
- 2 more uniforms into the light shader is nontrivial i think? but for most lights (especially by default) they aren't even different so they won't change very often. i think this is worth it because I can't see an alternative really. i could remove the curve type uniform and just not let people control that since its by far the less important of the two, but i did want the option..
- the actual shader logic shouldn't realistically be any less performant, on net its probably a few more multiplications because of the mix but slightly less elsewhere idk i didnt count that hard, i can try and collapse more of the logic into one line instead of using local vars but like surely that doesnt matter right? idk how gpu optimizations work.
- the bandwidth diff from pointlights compstate being slightly beefier shouldnt matter. its two floats, and pointlights rarely change from default anyway (besides enable/disable). we can do field deltas in the future. those are a thing now right. idk im out of the loop.

### DIFF
--- a/Resources/Shaders/Internal/light_shared.swsl
+++ b/Resources/Shaders/Internal/light_shared.swsl
@@ -15,6 +15,7 @@ uniform highp float lightRange;
 uniform highp float lightPower;
 uniform highp float lightSoftness;
 uniform highp float lightFalloff;
+uniform lowp int lightCurveType;
 uniform highp float lightIndex;
 uniform sampler2D shadowMap;
 
@@ -48,11 +49,15 @@ void fragment()
         discard;
     }
 
-    highp float dist = dot(diff, diff) + LIGHTING_HEIGHT;
+    // this implementation of light attenuation primarily adapted from
+    // https://lisyarus.github.io/blog/posts/point-light-attenuation.html
+    highp float sqr_dist = dot(diff, diff) + LIGHTING_HEIGHT;
 
-    highp float s = clamp(sqrt(distance) / radius, 0.0, 1.0);
+    highp float s = clamp(sqrt(sqr_dist) / lightRange, 0.0, 1.0);
     highp float s2 = s * s;
-    highp float val = clamp(((1 - s2) * (1 - s2)) / (1 + lightFalloff * s), 0.0, 1.0);
+    // curveType nonzero = quadraticinverse, 0 = inverse
+    highp float curveFactor = lightCurveType > 0 ? s2 : s;
+    highp float val = clamp(((1 - s2) * (1 - s2)) / (1 + lightFalloff * curveFactor), 0.0, 1.0);
 
     val *= lightPower;
     val *= mask;
--- a/Resources/Shaders/Internal/light_shared.swsl
+++ b/Resources/Shaders/Internal/light_shared.swsl
@@ -16,7 +16,6 @@ uniform highp float lightPower;
 uniform highp float lightSoftness;
 uniform highp float lightFalloff;
 uniform highp float lightCurveFactor;
-uniform lowp int lightCurveType;
 uniform highp float lightIndex;
 uniform sampler2D shadowMap;
 
--- a/Resources/Shaders/Internal/light_shared.swsl
+++ b/Resources/Shaders/Internal/light_shared.swsl
@@ -14,6 +14,7 @@ uniform highp vec2 lightCenter;
 uniform highp float lightRange;
 uniform highp float lightPower;
 uniform highp float lightSoftness;
+uniform highp float lightFalloff;
 uniform highp float lightIndex;
 uniform sampler2D shadowMap;
 
@@ -48,7 +49,10 @@ void fragment()
     }
 
     highp float dist = dot(diff, diff) + LIGHTING_HEIGHT;
-    highp float val = clamp((1.0 - clamp(sqrt(dist) / lightRange, 0.0, 1.0)) * (1.0 / (sqrt(dist + 1.0))), 0.0, 1.0);
+
+    highp float s = clamp(sqrt(distance) / radius, 0.0, 1.0);
+    highp float s2 = s * s;
+    highp float val = clamp(((1 - s2) * (1 - s2)) / (1 + lightFalloff * s), 0.0, 1.0);
 
     val *= lightPower;
     val *= mask;
--- a/Resources/Shaders/Internal/light_shared.swsl
+++ b/Resources/Shaders/Internal/light_shared.swsl
@@ -15,6 +15,7 @@ uniform highp float lightRange;
 uniform highp float lightPower;
 uniform highp float lightSoftness;
 uniform highp float lightFalloff;
+uniform highp float lightCurveFactor;
 uniform lowp int lightCurveType;
 uniform highp float lightIndex;
 uniform sampler2D shadowMap;
@@ -55,8 +56,8 @@ void fragment()
 
     highp float s = clamp(sqrt(sqr_dist) / lightRange, 0.0, 1.0);
     highp float s2 = s * s;
-    // curveType nonzero = quadraticinverse, 0 = inverse
-    highp float curveFactor = lightCurveType > 0 ? s2 : s;
+    // controls curve by lerping between two variants (inverse-shape and inversequadratic-shape)
+    highp float curveFactor = mix(s, s2, clamp(lightCurveFactor, 0.0, 1.0));
     highp float val = clamp(((1 - s2) * (1 - s2)) / (1 + lightFalloff * curveFactor), 0.0, 1.0);
 
     val *= lightPower;
--- a/Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
@@ -6,6 +6,7 @@ using Robust.Shared.Serialization.Manager.Attributes;
 using Robust.Shared.ViewVariables;
 using System.Numerics;
 using Robust.Shared.IoC;
+using Robust.Shared.Serialization;
 
 namespace Robust.Shared.GameObjects
 {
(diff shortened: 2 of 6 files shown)


mirrorcult:
what are the procedures for engine changelogs. i forgot how we do those now

mirrorcult:
image
image

another comp of what you can do with falloff, the right one has a falloff of like 10 or 20 or something and energy adjusted to better match the old version

mirrorcult:
updated this PR. replaced the ability to switch curves between `Inverse` and `InverseQuadratic` with a float that allows a continuous range of curves between the two for even more control, also included a video demonstrating it.

to motivate you to look at this here are some pretty pictures with light modifications I am doing for a fork based on what is possible in this PR.

image
image

Resources/Shaders/Internal/light_shared.swsl
```diff
@@ -14,6 +14,9 @@ uniform highp vec2 lightCenter;
 uniform highp float lightRange;
 uniform highp float lightPower;
 uniform highp float lightSoftness;
+uniform highp float lightFalloff;
+uniform highp float lightCurveFactor;
+uniform lowp int lightCurveType;
```
PJB3005: This appears to be unused?

Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
```diff
@@ -33,6 +34,43 @@ public abstract partial class SharedPointLightComponent : Component
...
+        /// </summary>
+        /// <remarks>
+        ///     This does not directly control the exponent of the denominator, though it might seem that way.
+        ///     Rather, it just lerps between an inverse-shaped curve and an inverse-quadratic-shaped curve.
+        ///     Values below 0 or above 1 are nonsensical.
+        ///
+        ///     See https://www.desmos.com/calculator/yjudaha0s6 for a demonstration of how this value affects the shape of the curve
+        ///     for different light radii and falloff values.
+        /// </remarks>
+        [DataField("curveFactor"), Animatable]
```
PJB3005:
```suggestion
        [DataField, Animatable]
```

Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
```diff
@@ -33,6 +34,43 @@ public abstract partial class SharedPointLightComponent : Component
...
+        /// <remarks>
+        ///     The default value of 6.8 might seem suspect, but that's because this is a value which was introduced
+        ///     years after SS14 already standardized its light values using an older attenuation curve, and this was the value
+        ///     which, qualitatively, seemed about equivalent in brightness for the large majority of lights on the station
+        ///     compared to the old function.
+        ///
+        ///     See https://www.desmos.com/calculator/yjudaha0s6 for a demonstration of how this value affects the shape of the curve
+        ///     for different light radii and curve factors.
+        /// </remarks>
+        [DataField("falloff"), Animatable]
```
PJB3005:
```suggestion
        [DataField, Animatable]
```

Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
```diff
@@ -33,6 +34,43 @@ public abstract partial class SharedPointLightComponent : Component
...
+        /// <summary>
+        ///     Controls the shape of the curve used for point light attenuation.
+        ///     This value may vary between 0 and 1.
+        ///     A value of 0 gives a shape roughly equivalent to (1/1+distance), while
+        ///     a value of 1 gives a shape roughly equivalent to (1+distance^2).
+        ///     The former (inverse) is close to how light falls off in real life.
+        ///     The latter (inverse quadratic) is not particularly physically accurate, but is better for representing
+        ///     "spherical point lights", e.g. a glowing orb, which want some degree of consistent light in
+        ///     close to their object's center before falling off as normal.
+        /// </summary>
```
PJB3005, line 58: That's a really long summary.

PJB3005 (changes requested):
Minor nits otherwise lgtm

Resources/Shaders/Internal/light_shared.swsl
mirrorcult: oops, yes, holdover from prev implementation

Robust.Shared/GameObjects/Components/Light/SharedPointLightComponent.cs
mirrorcult: i initially kept these explicitly specified to match the rest of the file, though it does make more sense to just follow convention anyway

mirrorcult:
resolved

PJB3005 (approved):

mirrorcult:
if merged, though this isn't a strict breaking change or anything and I imagine most fork lighting will look perfectly fine even if tuned differently from wizden, there should probably be a codebase changes post at least noting that something changed and you can tune it to reduce confusion

PR #6161, opened 2025-08-23
Fix children of nullspace entities not being serialized

Roudenn:
If some entity is parented to another entity in nullspace, when saving a map it will throw an error.
This PR fixes that by adding an additional check to EntitySerializer.

**Example of what it fixes:** SS14 MindRole entities (they're parented to Mind entity which is located in nullspace)
Tested in SS14 (if you ask how that will be a long story), works

Ideally this should be added to tests, but it works already so im not sure that it's required.


Roudenn:
Actually, i think i don't understand what i'm doing, so i'll close it until i find out proper solution

correct MissingEntityBehaviour probably fixes it

PR #6164, opened 2025-08-24
Add hashing to WritableDirProvider

### LINKED ISSUE #5316: Feature request: Expose hashing libs to client & server content
OpenDream needs it for checking file hashes in the DM `browse_rsc()` cache.

amylizzle:
OpenDream needs to hash files on the client, to check versioning of DM resources. This seemed like the best way to do that.

Closes #5316


PJB3005 (changes requested):
Yeah I don't really see the point of having this here. This is a helper OpenDream could surely implement itself directly?

If access to hashing APIs needs to be added to the sandbox, that shouldn't be a big deal.

amylizzle:
Yeah sure, if sandbox is better I'll PR that

Issue #2406, opened 2022-01-09
Full server snapshot/reload


gradientvera:
As discussed on the January 8th 2022 maintainer meeting, we should support full server snapshots.
Somewhat ties into #2185 as it would make it more convenient.
Some important points brought up in the meeting:
- Have a format for gamestate saves. 
    - EntitySystems should be able to read/write into that.
- Event serialization is needed for this.
- Ideally, the snapshot workflow would work like this:
    - Press some "save state" button.
    - Restart the server.
    - Enter the server.
    - Server has the same state as before.

moonheart08:
outer rim worldgen trembling in it's boots

PaulRitter:
most notable, this would require doafters & timer.spawns (as well as any other async coroutine or deferral i am forgetting rn) to no longer support anonymous methods so we can properly (de)serialize them

ElectroJr:
Just to get some clarification on this issue: is this just meant to be properly supporting saving/loading post-init maps & ensuring that everything that needs to get saved does get saved? Or is this meant to be some other kind of snapshot. 

If its meant to be the former, I don't think its that far away from being functional? There should probably be an issue tracker that lists things that need fixing, especially because I know there are new PRs that will cause issues, but I also don't want to hold off on merging them because they in turn need other things to be properly serialized first.

moonheart08:
> Just to get some clarification on this issue: is this just meant to be properly supporting saving/loading post-init maps & ensuring that everything that needs to get saved does get saved? Or is this meant to be some other kind of snapshot.
> 
> If its meant to be the former, I don't think its that far away from being functional? There should probably be an issue tracker that lists things that need fixing, especially because I know there are new PRs that will cause issues, but I also don't want to hold off on merging them because they in turn need other things to be properly serialized first.

This is for saving the entire game state, i.e. including all round information, every map and entity, etc.

metalgearsloth:
MapLoaderSystem should also be updated to support world saving / loading as right now it only does a single map.

PR #6165, opened 2025-08-24
MapId, MapCoordinates, EntityCoordinates Type serializers

### LINKED ISSUE #6159: EntityCoordinates / MapCoordinates can't be saved in YAML
`EntityUid` can be serialized to YAML just fine, because they have their own type serializer.
But `EntityCoordinates`, `MapCoordinates` and `MapId` don't, so they can't be saved properly.

Serialization needs to be able to support writing/reading entity coordinates.
This is required for resolving #2406 , since these types often occur in components and some other places.

Roudenn:
Added 4 new type serializers.
Required for #2406
Fixes #6159 

Coordinate serializers store data in 2 nodes, one represents MapId/EntityUid, and the other the position relative to that.

MapId serializer always converts and references to EntityUid for it to be supported in YAML saves.

Tested by modifying some SS14 code, forcefully serializing to YAML and then loading back all post-init maps at once, works


slarticodefast:
Pretty sure NetEntities are not supposed to be serialized, there is a PR marking them as `NotYamlSerializable` https://github.com/space-wizards/RobustToolbox/pull/5931
Same for NetCoordinates since those contain a `NetEntity`.

Roudenn:
> Pretty sure NetEntities are not supposed to be serialized, there is a PR marking them as `NotYamlSerializable` #5931 Same for NetCoordinates since those contain a `NetEntity`.

nuked

PJB3005 (commented):
Couldn't you just make `EntityCoordinates` and `MapCoordinates` have `[DataDefinition]` and `[DataField]`s?

slarticodefast:
> Couldn't you just make `EntityCoordinates` and `MapCoordinates` have `[DataDefinition]` and `[DataField]`s?

The `MapId` in the `MapCoordinates` is just an integer though, and I think we can't assume that all maps have the same Id when deserializing, right? Or in cases where you are loading a grid onto another map this would not work.

Why do we even have `MapCoordinates` in the first place? Couldn't we just replace them with `EntityCoordinates` referencing the map entity instead?

PJB3005:
> The `MapId` in the `MapCoordinates` is just an integer though, and I think we can't assume that all maps have the same Id when deserializing, right? Or in cases where you are loading a grid onto another map this would not work.

I specifically said the two coordinates types, *not* `MapId`. `MapId` should still needs handling to do the remapping.

> Why do we even have `MapCoordinates` in the first place? Couldn't we just replace them with `EntityCoordinates` referencing the map entity instead?

Bad design from years ago, mostly. We used to have `GridCoordinates` too before we got rid of it. If we were designing the engine with hindsight, `MapId` wouldn't exist either.

Roudenn:
> Couldn't you just make `EntityCoordinates` and `MapCoordinates` have `[DataDefinition]` and `[DataField]`s?

I failed to do so since they are readonly structs, it will require to change the type of the struct to add them

Roudenn:
> > Couldn't you just make `EntityCoordinates` and `MapCoordinates` have `[DataDefinition]` and `[DataField]`s?
> 
> I failed to do so since they are readonly structs, it will require to change the type of the struct to add them

The only way around this is probably making EntityCoordinates not readonly, but i think that is a bit unsafe to do

ElectroJr:
I agree with PJB that its probably better to just create a serializer only for mapid, and not the two coordinates types.

> I failed to do so since they are readonly structs, it will require to change the type of the struct to add them

If the issue was due to the source generator/analyzers requirement that fields shouldn't be readonly, you can get around that by using records. I.e., changing the coordinates to a `readonly record struct` and giving them the `DataRecordAttribute`. Data records are basically just `DataDefinitions` where every field/property is automatically treated like it has the `DataFieldAttribute`. AFAIK that should work fine (unless this is some oversight in the source generator & analyzer). And seeing as we define all the record related functions anyway, they probably should just be records.

Though there is a currently separate bug causing issues for data records, where it tries to treat get-only properties like fields (see #6204)

Roudenn:
> I agree with PJB that its probably better to just create a serializer only for mapid, and not the two coordinates types.
> 
> > I failed to do so since they are readonly structs, it will require to change the type of the struct to add them
> 
> If the issue was due to the source generator/analyzers requirement that fields shouldn't be readonly, you can get around that by using records. I.e., changing the coordinates to a `readonly record struct` and giving them the `DataRecordAttribute`. Data records are basically just `DataDefinitions` where every field/property is automatically treated like it has the `DataFieldAttribute`. AFAIK that should work fine (unless this is some oversight in the source generator & analyzer). And seeing as we define all the record related functions anyway, they probably should just be records.
> 
> Though there is a currently separate bug causing issues for data records, where it tries to treat get-only properties like fields (see #6204)

Seems like it's an issue, when i removed serializers and added `DataRecord` attribute to `EntityCoordinates`, they just throwed an error on startup because of the `X` get-only property...
I don't know what to do now honestly

ElectroJr:
Did merge master into your branch? Because #6204 should've fixed that.

Roudenn:
Oh right probably i didn't update my branch lol.

Roudenn:
Now it should be alright

ElectroJr:
I forgot that there was still another problem, in that the map entity might not have the map component by the time that something is trying to deserialize a map id, so you can't just do a try-get component to get the map id.

There's also issues that will come up when people try to load a map with a specific id. E.g., when you use `loadmap 12 /foo.yml`, the map gets assigned the map id of 12 after entities have been loaded from yaml but before they're initialized, so the map id that gets deserialized wouldn't match the final id thats actually used. This shouldn't really matter for full game saves, because you shouldn't be trying to load a save with a forced map id anyways, but its still something that should get logged.

I've fixed the former by pre-allocating a map ids before reading from yaml, while for the latter I've just made it log an error.
I tested it by just loading & saving an entity with MapCoordinates, and it seems to work fine now.

PR #6169, opened 2025-08-28
loadgrid no fail

IProduceWidgets:
Re-adds the functionality to the loadgrid command of ensuring an empty map.


VasilisThePikachu (approved):

PJB3005 (commented):
Add a message like "MapID $id did not exist, creating without map init"

IProduceWidgets:
> MapID $id did not exist, creating without map init

you got it

IProduceWidgets:
Not sure what caused that fail, but I'm pretty sure it wasn't me.

PR #6170, opened 2025-08-28
Fix TimeSpan overflow on Read

Andrew-Fall:
Adds a check to TimeOffsetSerializer to prevent overflow.

A lot of areas use TimeSpan.MaxValue but when saved and read the current time is added which results in an overflow. 
Notably an issue for #2406


PJB3005 (approved):
Just to confirm, this is for content code using `TimeSpan.MaxValue` as an "infinite timeout", right?

Andrew-Fall:
> Just to confirm, this is for content code using `TimeSpan.MaxValue` as an "infinite timeout", right?

Yes

PR #6171, opened 2025-08-29
Remove TransformBounds(Matrix3x2 , Box2Rotated)

### LINKED ISSUE #6068: Grid method in EntityLookup.Queries never finds an entity
Edit: The culprit appears to be `Matrix3Helpers.TransformBounds(Matrix3x2, Box2Rotated)`, see the first comment.

##

`GetEntitiesIntersecting(EntityUid, Box2Rotated, LookupFlags)` in `EntityLookup.Queries.cs` always finds nothing and returns no entities. This function is currently only used by content in `TurfSystem` for `IsTileBlocked()` and `GetEntitiesInTile()`. However it's used in turn by a lot of systems.

`IsTileBlocked()` is used by `RCDSystem`, `ToolSystem`, `SpecialRespawnSystem`, `TileNotBlocked : IConstructionCondition`, `MimePowersSystem`, `GoliathTentacleSystem`, `SpawnAfterInteractSystem`. Because the function always returns empty, none of these systems are ever blocked.

`GetEntitiesInTile()` is used by `CreateEntityTileReaction` for a whitelist, `SpiderSystem` for a pseudo-whitelist, and `FloorTileSystem` to check if placing a tile is blocked by a wall. Because the lookup always returns empty, these systems never prevent what they're intending to prevent.

## Reproduction
1. Put Carpetium in a fire extinguisher
2. Spray
3. Spray again

=> The whitelist will fail, causing carpet to spawn on carpet.

##

1. Spawn as a spider
2. Place webs
3. Place webs again

=> There will be no entities to iterate through, so there will be no webs to find with `HasComp` and webs will overlap.

Image

##

1. Attempt to place a tile under a wall

=> The wall is never considered, so the tile gets placed under it.

Image

##

You can see the lack of entities directly by placing a breakpoint inside the for loops used to iterate through these function's outputs. Or put a breakpoint on `IsTileBlocked()` to see that it always returns false (i.e. not blocked).

iaada:
fixes #6068

This little method has been causing 10 separate bugs in content (see linked issue). Fixing this was not due to me knowing matrixes, but painfully tracing the code down with breakpoints and some logical deduction (re: throwing out guesses until one landed). I'm not confident as to why, I just know that it appears to work.

My understanding is that because `TransformBounds(this Matrix3x2 refFromBox, Box2Rotated box)` was transforming the `Box2Rotated` by its own transform, it resulted in the new box getting offset in unexpected ways (see comment in issue). This should have been caught by `TestBox2RotatedMatrices`, but both before and after this change that test was passing. I can only speculate as to why, but possibly because `Matrix3x2.Multiply(fooMatrix, Box2Rotated.UnitCentered)` is equivalent to `fooMatrix`.

image


Tayrtahn:
Excellent job tracking this down! Could you please add a test case to `TestBox2RotatedMatrices` that fails without this change?

iaada:
I added the new test case as requested. I realized part of the problem with this test is that it wasn't testing for a `Box2Rotated` that was already rotated. I can't judge the quality of the test case (because I don't really know matrixes), but it passes on the fixed version and fails on the old version.

PJB3005:
What the hell even is this API supposed to do?

iaada:
I can't really answer what it's supposed to do. I can tell you it was added in #5400, and I can tell you where it's used.

It's used in two places in the project. Inside this method, which is unused:
```
private bool AnyEntitiesIntersecting(EntityUid lookupUid,
                                      Box2Rotated worldBounds,
                                      LookupFlags flags,
                                      EntityUid? ignored = null)
{
    var broadphaseInv = _transform.GetInvWorldMatrix(lookupUid);

    var localBounds = broadphaseInv.TransformBounds(worldBounds);
    var polygon = new SlimPolygon(localBounds);
    var result = AnyEntitiesIntersecting(lookupUid,
                                          polygon,
                                          localBounds.CalcBoundingBox(),
                                          Physics.Transform.Empty,
                                          flags,
                                          ignored);

    return result;
}
```

And in this method, which is used twice by `TurfSystem` for `IsTileBlocked()` and `GetEntitiesInTile()`:
```
public void GetEntitiesIntersecting(EntityUid gridId, Box2Rotated worldBounds, HashSet<EntityUid> intersecting, LookupFlags flags = DefaultFlags)
{
    if (!_broadQuery.TryGetComponent(gridId, out var lookup))
        return;

    var localBounds = _transform.GetInvWorldMatrix(gridId).TransformBounds(worldBounds);
    var polygon = new SlimPolygon(localBounds);

    AddEntitiesIntersecting(gridId, intersecting, polygon, localBounds.CalcBoundingBox(), Physics.Transform.Empty, flags, lookup);
    AddContained(intersecting, flags);
}
```

PJB3005:
A `Box2Rotated` has both a box and a rotation component. Your new method implementation does not use the rotation component at all. This can't be right?

iaada:
I went back to rerun some breakpoints. I don't know matrixes, so all I can really offer is data and speculation. It doesn't seem to rotate the box, but maybe that's happening implicitly in the `Vector2.Transform`?

image


In the above image I have the status of `GetEntitiesIntersecting` both before and after the change. In both instances the rotation is zero, which I believe is because `TransformBounds` is returning a new `Box2Rotated` using vectors for the constructor, which will always result in no rotation. The test seems to be aware of this, because the test case for a rotation matrix expects a result with no rotation.

I tried this code:
```
var matty = Matrix3x2.Multiply(refFromBox, box.Transform);
var boxxy = new Box2(Vector2.Transform(box.BottomLeft, refFromBox), Vector2.Transform(box.TopRight, refFromBox));
return new Box2Rotated(boxxy, matty.Rotation());
```

image

It gets the local bounds right, but it doesn't find the entity.

Right now I'm resigned to add more test cases and say it works.

iaada:
Not sure about the test fails, they don't seem to be related to this. `TestAllClientPrototypesAreSerializable` and `TestAllServerPrototypesAreSerializable` both failed due to Xenoborg textures.

iaada:
I hacked at it more and I think I'm satisfied with this result. I went absolutely overkill with the test cases to prove to myself it works and that I didn't miss an edge case. I can trim down the test cases if needed.

PJB was half right that the rotation was just ignored in favor of finding the correct locations and returning an unrotated box. `Box2Rotated.BottomLeft` is built using a box's rotation, so it appeared to find the right positions for the new box. But it would never ever return a rotated box which isn't the expected result. Now we can actually return a rotated box, and I realized we should probably transform the origin, too.

The below image shows inside `EntityLookupSystem.GetEntitiesIntersecting()`, shows it finding the intersecting entity, and `localBounds` returning what we expect to see. Note that this function is calling `GetInvWorldMatrix()` so the 0 rotation on `localBounds` is the expected result.

image

##

The only thing I'm unhappy about is `Box2Rotated.Origin`. It claims, "the point about which the rotation occurs," but that's not what I was experiencing. Test cases I wrote expecting to rotate around an offset origin were still rotating around (0,0). Returning whatever junk origin from `TransformBounds()` into `GetEntitiesIntersecting()` didn't affect `localBounds` at all, nor did it affect `localBounds.Transform`, nor did it cause the entity to fail to be found. Granted `localBounds` has 0 rotation so maybe that's just a fluke.

This function might need more work to make rotating around origin work. Origin is used in creating the matrix from `Box2Rotated.Transform` which is used for the returned rotation, so maybe it already works? This was the test case, but it was always returning `(0.5, 0.5, -0.5, -0.5)` (the angle was fine). I'm at a loss here but maybe I'm worried over nothing.

```
new TestCaseData(Matrix3x2.CreateRotation(MathF.PI),
    new Box2Rotated(Box2.UnitCentered, Angle.Zero, new Vector2(0.5f, 0.5f)),
    new Box2Rotated(new Box2(new Vector2(1.5f, 1.5f), new Vector2(0.5f, 0.5f)), Angle.FromDegrees(180), new Vector2(0.5f, 0.5f))),
```

iaada:
I keep coming back to this because I'm not satisfied, and I just can't quite understand it. I was wrong about origin not applying, origin is applying after the rotation from the matrix transformation. In other words I'm rotating it twice. It's spinning around (0,0) from the matrix, and then spinning again around origin. To visualize take this test case:
```
new TestCaseData(Matrix3x2.CreateRotation(MathF.PI),
                new Box2Rotated(Box2.UnitCentered, Angle.Zero, new Vector2(0.5f, 0.5f)),
```
Start:
image

image
image

But there's a related problem when origin is (0,0). For instance this test case:
```
new TestCaseData(Matrix3x2.CreateRotation(MathF.PI),
    new Box2Rotated(new Vector2(0f, 0f), new Vector2(1f, 1f)),
    new Box2Rotated(new Box2(new Vector2(0f, 0f), new Vector2(1f, 1f)), Angle.FromDegrees(180))),
```

image

The box is being rotated around its center when a rotation is applied by a matrix, but around origin when rotated as a part of Box2Rotated. But maybe this is simply the wrong expectation to have. Clearly you can't rotate around origin using a matrix. I changed the assert trying to get closer to understanding things, but maybe that was also a mistake. I feel like the solution is to always return an angle of zero.

# Is this whole method a mistake??
After basically giving up, I want to ask if this method is even needed. Both instances of it being used could be replaced by `public static Box2 TransformBox(this Matrix3x2 refFromBox, Box2Rotated box)`, and with better results, too. Look at these two results in `GetEntitiesIntersecting`, the first using the method as it is now in the PR, and the second with `TransfromBox`. (Possibly) because the `Box2Rotated` version calls `CalcBoundingBox()` the local bounds is peaking out of the actual tile it's trying to capture, where the `Box2` version doesn't.

image
image

I just can't think about this on my own anymore. I can't come up with something that satisfies me, and I'm lost fumbling in the dark.

ElectroJr:
> Is this whole method a mistake??

I'm pretty sure the answer is yes.

If the goal of the method is to start with a Box2Rotated, arbitrarily transform it, and then end with a Box2Rotated, you can't do that. For example, if your box is initially rotated slightly, and you just apply a scaling transformation that stretches the x-axis, its no longer shaped like a box, the angles aren't 90 degrees, and there is no way you can describe it accurately using only a Box2Rotated. The next best thing you can do is compute the bounding box of the new shape, which is what `Box2 TransformBox(this Matrix3x2 refFromBox, Box2Rotated box)` does. So in that sense, this method should just be removed and replaced with the `Box2` variant.

I think it might be possible to do this if the transformation only involves rotation & translation, which should be the case for all the transformation we're actually interested in. I.e., we could porbably define a `Box2Rotated  Transform(this Box2Rotated box, Angle rotation, Vector2 translation)`, but I'm not immediately sure how to do that.

 And even if its not possible, for how this method is actually being used by lookups, we can actually do better than bounding boxes by just directly transforming the 4 corners and passing them to a new `SlimPolygon` constructor, which would accomplish what this method was probably trying to do. If you want to do that, feel free to work on it, but I'm fine with just removing the current method and maybe adding a `// TODO` or opening an issue.

ElectroJr:
> I think it might be possible to do this if the transformation only involves rotation & translation

I just checked and Yes, it can be done, but the actual result is somewhat complicated, involves a few trig functions, and is probably not worth it when we can just pass the transformed box's vertices directly to `SlimPolygon`

iaada:
On Electro's advice, I pulled out the method and replaced the two uses of it with `TransformBox`. Shown below is `GetEntitiesIntersecting`, working as intended and finding the entity while staying inside the tile its trying to capture.

image image

ElectroJr:
For the sake of avoiding breaking changes, its probably better to keep the method for now and mark it as obsolete, but still fix it to return a proper bounding box. I.e., 
```cs
public static Box2Rotated TransformBounds(this Matrix3x2 refFromBox, Box2Rotated box)
{
    return new Box2Rotated(TransformBox(refFromBox, Box2Rotated box));
}
```

PR #6182, opened 2025-09-04
Refactor gender localization with support for customizable pronouns

mqole:
required for https://github.com/space-wizards/space-station-14/pull/40104

Adds a new prototype, `PronounTensePrototype`, which is used to dynamically generate FTL functions. Moved all existing pronoun tenses to this format. This means that grammar is no longer hardcoded, and users seeking to localize in languages with more grammatical tenses than RT defines can add their own grammatical tenses in yml and ftl.

FTL functions are regenerated when prototypes are refreshed. They must be generated AFTER prototypes are initialized so they cannot be in the locManager init.

`GrammarComponent` now contains a dictionary which can store non-standard pronouns for an entity for each grammatical tense. If a tense is called in FTL which a pronoun exists for, that pronoun will be used. Otherwise it will default to the enum gender of the entity.


mqole:
those test fails are really scary. i'm not sure how to address them.

iaada:
`TestAllServerPrototypesAreSerializable` and `TestAllClientPrototypesAreSerializable` seems to be consistently failing due to xenoborg actions on all robust PRs, so ignoring those all tests pass.

Resources/Locale/en-US/_engine_lib.ftl
PJB3005, line 1: Since these no longer operate on entities directly, you should rename `$ent` to be `$gender` instead.

Resources/Locale/pt-BR/_engine_lib.ftl
PJB3005, line 1: Ditto.

Robust.Shared/Localization/LocalizationManager.Functions.cs
PJB3005, line 1: Definitely use constants to avoid copy pasting the gender strings here. Where are these coming from anyways?

PJB3005 (changes requested):

VoidMeticulous:
5,000 LoC and substantial complexity across 3 PRs for custom pronouns? Come on.

We're talking about more code than entire systems here. The disguise bug already shows this is touching stuff in weird ways we didn't expect. What else is going to break? And the localization stuff... Russian needs like 15 different forms? That's insane to maintain.

Every time someone touches chat, examine, identity, or basically any UI, they now have to think about custom pronouns. That's a lot of cognitive overhead for something most servers probably won't even enable.

Here's what we should actually do: just let servers add their own pronouns through config files. Want xe/xir? Add it to your YAML. Want neopronouns in Russian? Go for it, it's your config. This is literally how every moddable game handles stuff like this - you don't shove everything into core, you give people hooks to add what they need.

The PR author put in a ton of work and I respect that, but this is the wrong approach. We could spend a fraction of this effort making pronoun sets configurable and everyone wins: servers get whatever pronouns they want, core stays maintainable, and we don't have to deal with database migrations from hell.

Seriously, 5,000 lines? For something that could be a config file?

I think many others are likely remaining silent due to the sensitivity of the topic, but I think this is an extremely important decision to consider. This seems hard to justify.

PJB3005:
@VoidMeticulous You have been permanently banned from GitHub.

1. Stop being a knee-jerk bigot
2. Don't go to an unrelated PR just to bitch when the other two PRs were clearly locked for good reason. In fact ***the very thing you are suggesting instead is not possible without this PR you are complaining on.***
3. Don't harass contributors
4. If there are issues with the code complexity, this is something maintainers should decide

Robust.Shared/Localization/LocalizationManager.Functions.cs
mqole: added these constructors to make it easier to modify individual pronouns in the ui for content-side `HumanoidCharacterProfileEditor` - same logic as `WithSex`, `WithAge`, `WithSpecies` etc. just gotta do one for each datafield so they can be edited individually.

mqole: oh if you meant the below section, then each gender string has a possibility of returning a different value. not sure how to go about using constants instead. `Gender.Neuter` is a fallback value for any entity that doesnt possess a grammatical or loc gender, and this method also accepts an input thats just a string of the relevant gender, for purposes of flexibility (and also i needed it to make the ui cooperate).

bicks-newd1848:
> @VoidMeticulous You have been permanently banned from GitHub.
> 
> 1. Stop being a knee-jerk bigot
> 2. Don't go to an unrelated PR just to bitch when the other two PRs were clearly locked for good reason. In fact _**the very thing you are suggesting instead is not possible without this PR you are complaining on.**_
> 3. Don't harass contributors
> 4. If there are issues with the code complexity, this is something maintainers should decide

imagine banning someone for suggesting that you shouldn't add more bloat to the game than there already is
were they being a "bigot" or did they hurt your very fragile ego, mister princess?
do better

mqole:
still need to test things on the content / db side, but optimistic given that migrations built ok. opening this up to review again in the meantime just to make sure i havent royally fucked it

edit: nvm i freeking hate unit test. not sure how to fix that one.

arimah:
I'm going to be a real stickler for terminology here, please forgive me. This is in the interest of keeping the code comprehensible and correct, so that it communicates the correct intention :)

- ***tense*** deals with how a verb relates to time: past, present, future, non-past, or whatever the language has.
- pronouns do not typically have a tense, hence `pronounTense` makes little sense.
- ***conjugation*** is specifically all the forms a _verb_ takes. A pronoun cannot conjugate (usually).
- Similarly, though I haven't seen the word used here, ***declension*** is for nouns and other noun-like things.
- Grammatical ***case*** is what we call the I/me/my, he/him/his, they/them/their forms. Case encodes the _role_ of a noun phrase within its clause (subject, object, etc.).
- The most neutral term for "grammatical word change" is ***inflection***.

Examples:

> He is a young reptilian.
> They are a young reptilian.

The change in verb form here is called _personal conjugation_. The FTL function `CONJUGATE-BE()` is correctly named.

> You inject the syringe into her. 
> You inject the syringe into them.

The choice of her/them involves finding the right _case_. For English, it's the oblique case. In German, it might be accusative or dative, depending on how one phrases "inject into X".

> You feel sick.
> You felt sick.

This is a change in tense. :)

I think it would be more appropriate to call it `pronounInflection`, or perhaps even more generically `pronounGrammar`. Moreover, I would encourage you to ensure comments use appropriate terminology for the task they're trying to accomplish: don't say you're conjugating a pronoun! :D I haven't looked deeply at the code yet and can't comment more beyond semantic nitpickery at this time.

mqole:
@arimah no need for forgiveness, i appreciate the nitpicks immensely!! i'm definitely not a grammar freak so i do sincerely thank you for the clarification. i'll give all the comments and documentation a tidying with your feedback in mind.

PR #6184, opened 2025-09-04
Move SDL3 binding to NuGet package

PJB3005:
I'm worried about the IDE performance overhead of the 20k lines of LibraryImport it generates into Robust.Client.

Also, this allows me to trim the binding, which saves a tiny amount of space from publishes. Always nice to have.


FairlySadPanda:
chat she's deleted the codebase it's joever 
image

PR #6185, opened 2025-09-06
feat: `GetString`, `TryGetString`, `HasString` culture aware variants

Kyoth25f:
I'm trying to make search bars in content search by localization in the default culture and other arbitrary culture (the fallback would also be enough). Then I noticed there's no way to get other localization other than your default culture (correct me if I'm wrong), except by using the fallback.

This PR add variants for the `GetString`, `TryGetString` and `HasString` methods that accepts the culture as a parameter. I considered changing the signature to accept a nullable culture, but I don't think that's good. The function behaves differently depending it's null or not. With a culture selected, it searches just that in that culture bundle. When there's no culture selected, it looks in the default and, if it didn't found, looks in the fallbacks.

### DIFF
--- a/Robust.Shared/Localization/ILocalizationManager.cs
+++ b/Robust.Shared/Localization/ILocalizationManager.cs
@@ -32,11 +32,29 @@ namespace Robust.Shared.Localization
         string GetString(string messageId);
 
         /// <summary>
+        ///     Gets a language appropriate string represented by the supplied messageId.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


PJB3005 (changes requested):
Do not copy paste the implementations.

Kyoth25f:
> Do not copy paste the implementations.

Ok, I'll see what I can do

PJB3005:
Stale.

PR #6187, opened 2025-09-08
exposes most of IResourceManagerInternal

LaCumbiaDelCoronavirus:
moved everything, except for `Initialize(...)`, from `IResourceManagerInternal` up to `IResourceManager` so that it's publicly accessible

initialize method was kept internal because resourcemanager probably shouldn't be re-initialized ever
there's a few really small formatting changes made by VSC when i did the find-and-replace

## Why

i need some way to mount a real directory to the VFS but literally everything i could use was either internal/private
i couldn't use AddRoot because it needed an implementation of IContentRoot but the useful ones are internal, yet i can't write my own implementation because the types needed for it arent whitelisted in sandbox, for good reason

### Alternatives

so incase this is a bad approach here's two things i would be willing to do instead
1. make a public implementation of `IWritableDirProvider` that only does read operations (no write), but not virtual (like `WritableDirProvider`)
2. alternatively, create two separate `I...DirProvider` interfaces for reading and writing each, with the latter being internal and former public. `IResourceManager` would use the read-only form as it doesn't do any writing iirc other than creating a userdata directory

therefore you can have people actually make their own `IContentRoot` implementations that would work for mounting content directories, since vanilla `IResourceManager` publicly exposes AddRoot
would need to circumvent using RootDir someway because i think that has PII


PJB3005:
I'm sorry, but these APIs are sandboxed for a reason. This PR might as well be titled "give content the ability to read your browser passwords" because you're directly allowing content to just mount `C:` or whatever.

PR #6188, opened 2025-09-09
Improve map serialization error logging & exception tolerance

ElectroJr:
Wraps the code that serializes an entity's components in a try-catch block that logs an error with information about the problematic entity & component. Also adds an option to continue serializing the map while skipping over the entity. The default behaviour is still to abort the serialization (rethrow the exception)

### DIFF
--- a/Robust.Shared/EntitySerialization/SerializationEnums.cs
+++ b/Robust.Shared/EntitySerialization/SerializationEnums.cs
@@ -86,3 +86,27 @@ public enum MissingEntityBehaviour
     /// </summary>
     AutoInclude,
 }
+
+
+public enum EntityExceptionBehaviour
+{
+    /// <summary>
+    /// Re-throw the exception, interrupting the serialization.
+    /// </summary>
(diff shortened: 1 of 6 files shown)


Robust.Shared/EntitySerialization/SerializationEnums.cs
```diff
@@ -86,3 +86,33 @@ public enum MissingEntityBehaviour
...
+    Rethrow,
+
+    /// <summary>
+    /// Continue serializing and simply skip/ignore this entity. May result in broken maps that log errors or simply
+    /// fail to load.
+    /// </summary>
+    IgnoreEntity,
+
+    /// <summary>
+    /// Continue serializing and simply skip/ignore this entity and all ofits children.
```
PJB3005:
```suggestion
    /// Continue serializing and simply skip/ignore this entity and all of its children.
```

PJB3005 (commented):
Incredibly valuable review

PR #6189, opened 2025-09-09
Add support for including map/game saves in replays.

ElectroJr:
This PR adds some new methods to make it easier to include map/game saves in replays. Specifically it:
- Adds `IReplayFileWriter.WriteYaml()` to make it easier to write yaml documents. This could already be done with `WriteBytes`, but this is more convenient.
- Adds `MapLoaderSystem.TrySerializeAllEntities()`, which will attempt to serialize all entities to a single yaml file.
  - This overlaps with part of #6166, but I think this implementation is better.
- Marks `ActorComponent` as an "UnsavedComponent"
  - This was never actually supported, and would cause NREs when trying to access `ActorComponent.PlayerSession`, so I consider this more of a bugfix and not a breaking change.


Roudenn:
That one's very good, but some parts are heavily conflicted with #6166. Probably would be better to take your version tho, so I encourage you to keep developing this!

Also we probably should somehow cooperate on who does what changes and how

Roudenn:
#6166 I updated my PR to have more info in Technical Details part. I plan to create a file structure in which all systems can save their own files to, so it's a .zip archive with a bunch of files and `versions.yml`.
Not sure about in which order we should merge these 2 PRs, but probably would be better for this one to be first. Then i will just replace saving of all entities with a full game save and it should work fine.

ElectroJr:
> That one's very good, but some parts are heavily conflicted with https://github.com/space-wizards/RobustToolbox/pull/6166. 

Ah my bad, I should have had a proper look at the already open PRs before working on this. I've been rather inactive and hadn't noticed that PR had been created. I'll take a look it tomorrow, for now I need sleep.

deltanedas:
shouldve looked harder at the thousand open prs yeah shame on you :trollface:

ElectroJr:
> shouldve looked harder at the thousand open prs yeah shame on you :trollface:

Alas, not sure I can use that excuse for engine PRs, I was just too focused on implementing my replay idea to bother checking open PRs

PR #6192, opened 2025-09-12
Delay Entity Initialization for Spawn Behavior, add addition Container Spawner Methods

### LINKED ISSUE #6258: Entities don't get properly woken when being moved from nullspace.
Title.

To replicate: Spawn an entity, move it from nullspace to a map (Using PredictedSpawnNextToOrDrop for example, or using AntagSelectionSystem).

On content you can do this by creating soap with the create soap chemical reaction, and ensuring it's not colliding with any hard fixtures when it spawns. Then try to slip on it (you can't). Or by spawning as an antag (such as a nuclear operative) and then immediately ghosting without moving, then trying to shoot your body (bullets pass through cause no collisions). 

Entity will not be properly woken upon being moved from nullspace to its new parent, resulting in it not colliding with anything until it moves.
This is problematic particularly for static entities which can't move but may still need to collide. 

Using spawn methods which pass MapCoordinates seem to work but the engine should be able to properly wake moved entities without having to put a _physics.WakeBody after every single spawn in content.

Princess-Cheeseballs:
## Description 
Exactly as the title says, I delayed the initialization of entities spawned until they have been properly re parented or inserted into a container during spawning behavior.

- Fixes entities raising MapInit and ComponentInit in nullspace causing Transform data to be inaccurate when spawning. This resulted in some content bugs where entities would just not spawn and also a heisentest. 
- This also should fix #6258 since this was partially caused by entities initializing in nullspace during predicted spawning. Some content spawning methods will need to be altered but this fixes the majority of them.
- This would also allow us to delete Jez's workaround in SolutionsSystem which delays initialization for solutions until they're inserted into their container. 

## Why?
This was causing bugs on upstream where for example: I spawn an entityA which spawns a random entityB from a table due to its component. EntityA spawns and initializes in nullspace, so when it tries to spawn entityB, entityB fails to spawn. 

This was often worked around on content by just writing their own spawn behavior to purposefully delay initialization in very specific edge cases, and does not work with prediction. 

This specific example is the cause of a bug with the Crystals Reaction on content, and almost caused a heisentest with the surplus crate, but luckily it was caught before that PR was merged. 

The changes to initialization were going to be their own PR, but I ended up recreating one of the methods in this PR in my efforts to de-duplicate code, so I just merged it into this one. Similarly, on content people often just rewrite parts of TrySpawnEntityInContainer for performance reasons so we don't have to keep trying to get the same container over and over again so moving that boilerplate to engine is probably for the best. 

## Breaking Changes
When spawning entities MapInit is not run until the entity has been reparented to its final parent.


slarticodefast:
Please add documentation to all the new methods, container code is already confusing enough 😄

Princess-Cheeseballs:
Updated PR description since it now makes some decently sizable changes to spawning behavior in an effort to fix spawning bugs on content.

Robust.Shared/GameObjects/EntityManager.Spawn.cs
```diff
@@ -218,7 +239,7 @@ public EntityUid SpawnInContainerOrDrop(
         inserted = true;
         xform ??= TransformQuery.GetComponent(containerUid);
         var doMapInit = _mapSystem.IsInitialized(xform.MapUid);
-        var uid = Spawn(protoName, overrides, doMapInit);
+        var uid = CreateEntityUninitialized(protoName, overrides);
```
moonheart08: Does this not mean container events will be raised on an uninitialized entity?

moonheart08 (changes requested):
This makes uninitialized entities observable in content via the container events. Containers do a *lot* of things to entities as well so for example if you used these methods on a newly spawned wall, containers themselves would observe the uninitialized state by failing to unanchor it. 

I'm not sure this approach is workable, deferring mapinit alone would need reviewing most mapinit implementations in content to make sure they're ready for the change.

Princess-Cheeseballs:
> This makes uninitialized entities observable in content via the container events. Containers do a _lot_ of things to entities as well so for example if you used these methods on a newly spawned wall, containers themselves would observe the uninitialized state by failing to unanchor it.
> 
> I'm not sure this approach is workable, deferring mapinit alone would need reviewing most mapinit implementations in content to make sure they're ready for the change.

It's been a while but I did test this on content and it worked fine, but yeah we shouldn't be inserting things into containers until the component has initialized. I can also adjust comments on the Init functions to clarify when they're called and what should and shouldn't be relied upon during them. 

Content wise, I did testing to make sure nothing threw when I made this PR, that being said many of the errors this PR was intended to fix were silent so that's not exactly a saving grace.

moonheart08:
> > This makes uninitialized entities observable in content via the container events. Containers do a _lot_ of things to entities as well so for example if you used these methods on a newly spawned wall, containers themselves would observe the uninitialized state by failing to unanchor it.
> > I'm not sure this approach is workable, deferring mapinit alone would need reviewing most mapinit implementations in content to make sure they're ready for the change.
> 
> It's been a while but I did test this on content and it worked fine, but yeah we shouldn't be inserting things into containers until the component has initialized. I can also adjust comments on the Init functions to clarify when they're called and what should and shouldn't be relied upon during them.
> 
> Content wise, I did testing to make sure nothing threw when I made this PR, that being said many of the errors this PR was intended to fix were silent so that's not exactly a saving grace.

Yea the errornous behavior this casues would *also* be widely silent, with components made but not fully initialized.

Princess-Cheeseballs:
@moonheart08 made it so it only delays MapInit now.

moonheart08 (approved):
This is a breaking change, ensure it is documented as such.

Robust.Shared/GameObjects/EntityManager.Spawn.cs
```diff
@@ -172,15 +164,45 @@ public bool TrySpawnInContainer(
...
+    /// <returns>True if entity was spawned and inserted successfully</returns>
+    public bool TrySpawnInContainer(
+        string? protoName,
+        BaseContainer container,
+        [NotNullWhen(true)] out EntityUid? uid,
+        EntityUid? mapUid,
+        ComponentRegistry? overrides = null)
+    {
+        uid = CreateEntityUninitialized(protoName, overrides);
+        var meta = MetaQuery.GetComponent(uid.Value);
```
SnappingOpossum: You should be able to do this instead for these, it's otherwise discarded by CreateEntityUninitialized anyway.
```suggestion
        uid = CreateEntityUninitialized(protoName, out var meta, overrides);
```

Princess-Cheeseballs: Cheers

Robust.Shared/GameObjects/EntityManager.Spawn.cs
```diff
@@ -172,15 +164,44 @@ public bool TrySpawnInContainer(
...
 
-        if (_containers.Insert(uid.Value, container))
-            return true;
+    public bool TrySpawnInContainer(
+        string? protoName,
+        BaseContainer container,
+        [NotNullWhen(true)] out EntityUid? uid,
+        ComponentRegistry? overrides = null)
+    {
+        return TrySpawnInContainer(protoName, container, out uid, TransformQuery.GetComponent(container.Owner).MapUid, overrides);
```
whatston3:
```suggestion
        return TrySpawnInContainer(protoName, container, out uid, TransformQuery.Comp(container.Owner).MapUid, overrides);
```
Slightly shorter, probably good if using inline like this.

Robust.Shared/GameObjects/EntityManager.Spawn.cs
```diff
@@ -217,8 +242,8 @@ public EntityUid SpawnInContainerOrDrop(
     {
         inserted = true;
         xform ??= TransformQuery.GetComponent(containerUid);
```
whatston3:
```suggestion
        xform ??= TransformQuery.Comp(containerUid);
```
Ditto - shorter proxy?

Robust.Shared/GameObjects/EntityManager.Spawn.cs
```diff
@@ -229,6 +254,45 @@ public EntityUid SpawnInContainerOrDrop(
...
+    public EntityUid SpawnInContainerOrDrop(
+        string? protoName,
+        BaseContainer container,
+        out bool inserted,
+        TransformComponent? xform = null,
+        ContainerManagerComponent? containerComp = null,
+        ComponentRegistry? overrides = null)
+    {
+        inserted = true;
+        xform ??= TransformQuery.GetComponent(container.Owner);
```
whatston3:
```suggestion
        xform ??= TransformQuery.Comp(container.Owner);
```
Ditto, shorter proxy.

Robust.Shared/GameObjects/IEntityManager.Spawn.cs
```diff
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using Linguini.Syntax.Ast;
 using Robust.Shared.Collections;
```
whatston3:
```suggestion
```
Unused, please remove.

Robust.Shared/GameObjects/EntityManager.Spawn.cs
```diff
@@ -217,8 +242,8 @@ public EntityUid SpawnInContainerOrDrop(
     {
         inserted = true;
         xform ??= TransformQuery.GetComponent(containerUid);
-        var doMapInit = _mapSystem.IsInitialized(xform.MapUid);
-        var uid = Spawn(protoName, overrides, doMapInit);
+        var uid = CreateEntityUninitialized(protoName, out var meta, overrides);
+        InitializeAndStartEntity((uid, meta), doMapInit: false);
 
         if ((containerComp == null && !TryGetComponent(containerUid, out containerComp))
```
whatston3: Since you're making reasonably significant changes to these methods anyway, between this and `TrySpawnInContainer`, should `TryGetComponent<ContainerComponent>` be moved into a `EntityQuery<ContainerComponent>`?

whatston3 (commented):
Looks alright.  Found a few possible nitpicks, otherwise seems fine to me.

Depending on the appetite for the new API methods, #6784 is a minimal alternative.

Wish I'd have seen this before.

Robust.Shared/GameObjects/EntityManager.Spawn.cs
SnappingOpossum: Could also use _containers.TryGetContainer and leave the work to the container system.

Princess-Cheeseballs: Updated the PR to include this cleanup.

Robust.Shared/Containers/SharedContainerSystem.Insert.cs
```diff
@@ -136,6 +136,16 @@ public bool Insert(Entity<TransformComponent?, MetaDataComponent?, PhysicsCompon
         return true;
     }
 
+    /// <inheritdoc cref="InsertOrDrop(Entity{TransformComponent?,MetaDataComponent?,PhysicsComponent?},BaseContainer,TransformComponent?)"/>
+    public bool InsertOrDrop(Entity<ContainerManagerComponent?, TransformComponent?> container, Entity<TransformComponent?, MetaDataComponent?, PhysicsComponent?> toInsert,
+        string containerId)
+    {
+        if (!TryGetContainer(container, containerId, out var baseContainer, container))
+            return false;
```
SnappingOpossum: Test fail is because this doesn't do a DropNextTo on fail ("Trying to spawning inside a non-existent container just drops the entity"), this solution worked locally.
```suggestion
        {
            _transform.DropNextTo(toInsert, (container.Owner, container.Comp2));
            return false;
        }
```

Princess-Cheeseballs: whoops

Robust.Shared/Containers/SharedContainerSystem.Insert.cs
```diff
@@ -136,6 +136,18 @@ public bool Insert(Entity<TransformComponent?, MetaDataComponent?, PhysicsCompon
...
+    public bool InsertOrDrop(Entity<ContainerManagerComponent?, TransformComponent?> container, Entity<TransformComponent?, MetaDataComponent?, PhysicsComponent?> toInsert,
+        string containerId)
+    {
+        if (TryGetContainer(container, containerId, out var baseContainer, container))
+            return InsertOrDrop(toInsert, baseContainer, container.Comp2);
+
+        _transform.DropNextTo(toInsert, (container.Owner, container.Comp2));
+        return false;
+
+    }
```
SnappingOpossum, line 148:
```suggestion
        return false;
    }
```

Robust.Shared/GameObjects/IEntityManager.Spawn.cs
```diff
@@ -82,6 +96,23 @@ EntityUid SpawnInContainerOrDrop(
...
+        ComponentRegistry? overrides = null);
+
+    /// <inheritdoc cref="SpawnInContainerOrDrop(string?,Robust.Shared.GameObjects.EntityUid,string,Robust.Shared.GameObjects.TransformComponent?,Robust.Shared.Containers.ContainerManagerComponent?,Robust.Shared.Prototypes.ComponentRegistry?)"/>
+    EntityUid SpawnInContainerOrDrop(
+        string? protoName,
+        BaseContainer container,
+        out bool inserted,
+        TransformComponent? xform = null,
+        ContainerManagerComponent? containerComp = null,
+        ComponentRegistry? overrides = null);
```
SnappingOpossum, line 114: This and its predicted version seem to lack a proxy.

metalgearsloth (approved):

Robust.Shared/GameObjects/IEntityManager.Spawn.cs
Princess-Cheeseballs: done

deltanedas:
merge lol i had this prevent a serverside fixture getting proxies added with SpawnNextToOrDrop @metalgearsloth

PR #6193, opened 2025-09-13
Box Simd

ElectroJr:
This PR rewrites some Box2, Box2Rotated, other related methods to make more/better use of SIMD. I have no idea if the actual performance gains are worth the effort, but the methods are used quite a lot by physics & rendering, and I had fun either way.

## Min vs MinNative

One of the notable change is that this PR replaces uses of `Vector128.Min` with `Vector128.MinNative` and likewise for Max. Prior to #3584 we were explicitly using `Sse.Min`, and prior to dotnet 9 `Vector128.Min` was effectively equivalent. However to resolve issues involving nans & -0 that apparently changed in dotnet 9 (e.g., see dotnet/runtime/issues/108168), resulting in significantly slower code. Given that we just use this for computing bounding boxes, I don't think we need to worry about differentiating between -0 & 0? The results for the `GetAABBBenchmark` on my machine with Min, MinNative, and a new AVX specific version gets:
| Method    | Mean      | Error     | StdDev    | Ratio | Code Size |
|---------- |----------:|----------:|----------:|------:|----------:|
| Min    | 5.7976 ns | 0.0091 ns | 0.0085 ns |  1.00 |     491 B |
| MinNative | 0.9110 ns | 0.0013 ns | 0.0012 ns |  0.16 |     117 B |
| Avx       | 0.4925 ns | 0.0006 ns | 0.0005 ns |  0.08 |      84 B |

I don't trust that I did the benchmark correctly seeing how short the mean is. Though maybe the way BenchmarkDotNet performs multiple invocations per loop iteration happens to allow the CPU to perform instructions in parallel? But the point here is that `Min` is significantly slower than using `MinNative`, even if the difference may be somewhat exaggerated here.

## Other main Changes:
- Added `SimdHelper.GetAABB()`
  - The various box methods now use this instead of duplicating the same code.
- Added `Box2Rotated.GetVertices()`
- Added new `SlimPolygon` constructors that take in a box & transform
- Simdified `SlimPolygon.GetAABB`
- Added a new simd  method for physics' `Transform`
- Added new `Matrix3Helpers.TransformBox()` overload that returns the box's vertices instead of a bounding box. 
  - This, or the new SlimPolygon constructors should be used to replace the  old `TransformBounds()`, which AFAIK was meant to avoid enlarging a lookup but simply wasn't working as intended (see #6171).


Robust.Shared/Physics/Shapes/SlimPolygon.cs
```diff
@@ -45,27 +48,74 @@ public SlimPolygon(Box2 box)
         _normals._01 = new Vector2(1.0f, 0.0f);
         _normals._02 = new Vector2(0.0f, 1.0f);
         _normals._03 = new Vector2(-1.0f, 0.0f);
+        Centroid = box.Center;
```
ElectroJr, line 50: Centroid was never being set and this is uses `Unsafe.SkipInit`. I don't know if this was just never actually being used by anything and didn't really matter, or if this was actually a significant bug.

Robust.Client/Graphics/Clyde/Clyde.Sprite.cs
```diff
@@ -161,47 +161,32 @@ private void ProcessSprites(
...
-            var lrlr = SimdHelpers.MinMaxHorizontalSse(modX);
-            var btbt = SimdHelpers.MaxMinHorizontalSse(modY);
-            lbrt = Sse.UnpackLow(lrlr, btbt);
-        }
-        else
-        {
-            var l = SimdHelpers.MinHorizontal128(allX);
-            var b = SimdHelpers.MaxHorizontal128(allY);
-            var r = SimdHelpers.MaxHorizontal128(allX);
-            var t = SimdHelpers.MinHorizontal128(allY);
```
ElectroJr, line 198: Turns out sprite bounds have been bugged for non-Sse systems this whole time (should've been `modX` not `allX`)

Alwayswannahunt:
@ElectroJr Hi! Is it possible to make beta robust release based on v267.3.0 tag with this PR merged? Id like to test performance gain on real workload (before that Ill test for any bugs, so I'm responsible for any issues during that test)
Ill report result here if you don't mind.
Planned workload is average ss14 round and I'm looking forward to test in dozens grid interaction for like 100 players.

Robust.Shared.Maths/Box2Rotated.cs
```diff
@@ -33,7 +33,31 @@ public struct Box2Rotated : IEquatable<Box2Rotated>, ISpanFormattable
...
+            get
+            {
+                // Equivalent to  Matrix3Helpers.CreateTransform(Origin - Rotation.RotateVec(Origin), Rotation)
+                // but ~20% faster
+                var angle = (float) Rotation;
+                var sin = MathF.Sin(angle);
+                var cos = MathF.Cos(angle);
+                var cos1 = 1 - cos;
+                var dx = cos1 * Origin.X + sin * Origin.Y;
+                var dy = - sin * Origin.X + cos1 * Origin.Y;
```
metalgearsloth, line 48: Should this use Math.Sin instead to keep the angle as high-precision as possible until dx? Often I had issues with matrices and having angle imprecision issues causing headaches.

metalgearsloth (approved):

Robust.Shared.Maths/Box2Rotated.cs
ElectroJr: I don't really have a good feel for when double is necessary, if you think we need it I'm happy to change it. It'd be nice if we had some kind of test or example precision requirements that can be used as a standard to help decide

ElectroJr:
@Alwayswannahunt Sorry for the late reply, but I guess you can compare with the new engine version when it gets tagged.
Also, FWIW you can set up the server to use a custom engine version if you wanted to, only the clients actually need to use official tagged versions. Though I'm not sure how much of a hassle it is to set up, but its probably not hard.

Alwayswannahunt:
@ElectroJr Oh, you're right that actually server can run a custom version if client's part of the custom and tagged version are the same. Besides after moving to 267.4.0 version less performance issues were noticed and this PR was merged, so... I think that I achieved what I wanted! 😄 
P.S. no problem in late reply

PR #6195, opened 2025-09-14
SharedAppearanceSystem option to change appearance data without dirty

Fildrance:
SharedAppearanceSystem methods that change appearance data currently dirty Appearance component on each call. This leads to component state reset during ResetPredictedEntities. This basically blocks an ability **to have some client-only appearance code**. To 'fix' that i added option to pass flag so system wont dirty component. This does not seem like a perfect soution, but it mends problem somehow.
Not 100% sure if it would be better to create separate overload to keep binary compatibility, so just added default parameter to related methods - i can do the overloads if that seems appropriate


Fildrance:
Would like that API change for https://github.com/space-wizards/space-station-14/pull/38313

PJB3005:
This literally does not work properly, and will just lead to future issues if you attempt to use it.

PR #6196, opened 2025-09-15
Make net.interp and net.buffer_size CVars CLIENT, REPLICATED instead of CLIENTONLY

DrSmugleaf:
Requires https://github.com/space-wizards/space-station-14/pull/40364 to not fail tests
Lets the server know how much to offset for when lag compensating a client
Trying to test this vs just sending the tick an interaction was done from the client


PJB3005:
Test fail appears legit

DrSmugleaf:
> Test fail appears legit

@PJB3005 Fixed in https://github.com/space-wizards/space-station-14/pull/40364

PJB3005:
> > Test fail appears legit
> 
> @PJB3005 Fixed in [space-wizards/space-station-14#40364](https://github.com/space-wizards/space-station-14/pull/40364)

It seems the actual bug is that `CLIENT | REPLICATED` cvars get sent to a server even if not connected? This should work fine without adjusting content.

PJB3005:
Tests should be fixed with 0c691b061db44ca0d1a5bc56a55f1e3f39d47d83

PR #6197, opened 2025-09-15
Better color API

vgskye:
Currently the colorspace-related APIs in `Color` is a mess, putting Linear sRGB values in `Color` and all other colorspaces just using `Vector4` without any safeguards against confusing, say, HSL for HSV. This PR adds the `Colors` namespace, which holds actually typed color classes with methods for converting between them as appropriate.

Opened as Draft because I want to add more features(mainly Oklab gamut clipping) and have colorspace parity with `Color`, and then add deprecations to the existing `Color` colorspace API while making them be wrappers over the `Colors` API.

### DIFF
--- a/Robust.Shared.Maths/Colors/SyccColor.cs
+++ b/Robust.Shared.Maths/Colors/SyccColor.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Numerics;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using Robust.Shared.Utility;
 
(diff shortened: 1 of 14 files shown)


Robust.Shared.Maths/Colors/SyccColor.cs
```diff
@@ -0,0 +1,100 @@
+using System;
+using System.Numerics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using Robust.Shared.Utility;
+
+namespace Robust.Shared.Maths.Colors;
+
+/// <summary>
+///     Represents a color with alpha in the sYCC colorspace.
```
ZeWaka, line 10: What's the purpose of including sycc?

vgskye: parity with existing Color API, mostly.

vgskye:
I have now added `ToColor()` methods to all Colors API structs and `ToXXX()` methods to Color so it's much more ergonomic to use for the common case of "convert to some other colorspace, do work, convert back to sRGB". The `Color` methods also use the same method names as the old API, just as instance methods instead of static.

ZeWaka:
👀

vgskye:
I no longer have enough interest in ss14 to work on this pr. if someone wants this they can resubmit this idrc

PR #6200, opened 2025-09-16
Include last applied tick in `EntitySessionEventArgs`

ElectroJr:
Having the server know about the client's last applies server state at the time that a networked event was raised makes implementing useful lag compensation easier. The client could still intentionally lie about the last applied state and maybe use that to try abuse lag compensation, so the client's ping should maybe still be used as a sanity check.

Though I'm now having second thoughts about this, as this makes all entity-events one int larger. While I feel like this is more convenient, maybe this is information that should just be sent in specific events that actually need to use lag compensation instead of all networked events.


mirrorcult:
imo it makes more sense to have this information available, but specific events that need it should probably just include a field with `_gameTiming.LastRealTick` instead of all events having it?

ElectroJr:
Yeah I'll just do that instead

PR #6204, opened 2025-09-17
Fix serialization of data-records with get-only properties

ElectroJr:
This PR changes `GatherFieldData()` so that it returns false when given a property with no setter and no backing field. This was preventing record structs like this from being used:
```cs
[DataRecord]
public record struct Foo
{
    public Vector2 Position;
    public float X => Position.X;
}
```

The actual change is relatively small, but this PR also inverts some if the if statements to reduce nesting and hopefully make the method easier to parse, because it left me quite confused.

### DIFF
--- a/Robust.Shared/Serialization/Manager/Definition/DataDefinition.cs
+++ b/Robust.Shared/Serialization/Manager/Definition/DataDefinition.cs
@@ -332,13 +332,9 @@ namespace Robust.Shared.Serialization.Manager.Definition
             inheritanceBehavior ??= InheritanceBehavior.Default;
 
             if (fieldInfo.HasAttribute<AlwaysPushInheritanceAttribute>(true))
-            {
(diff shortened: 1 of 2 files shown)


Robust.Shared/Serialization/Manager/Definition/DataDefinition.cs
```diff
@@ -355,36 +349,54 @@ private bool GatherFieldData(AbstractFieldInfo fieldInfo, out DataFieldBaseAttri
...
+            // DataRecordAttribute will cause all fields to be interpreted as data fields, so we still handle them
 
-                    backingField = backingFieldInfo;
-                }
+            if (fieldInfo is not SpecificPropertyInfo)
+                return true;
+
+            var potentialBackingField = fieldInfo.GetBackingField();
+            if (potentialBackingField == null)
+                return false;
```
ElectroJr, line 370: This return is the behaviour that's actually changed

PR #6209, opened 2025-09-18
Fix compat mode floats in light attenuation + release notes

mirrorcult:
ok glad i remembered that existed but would have been better to remember at some point in the last few weeks

@PJB3005


slarticodefast:
Can we somehow run an integration test in compability mode to catch these shader problems?
This is at least the 4th time it has happened.

mirrorcult:
at the very least it'd be nice to have a linter for it tbh

PJB3005:
> Can we somehow run an integration test in compability mode to catch these shader problems?
> This is at least the 4th time it has happened.

Not until renderer rewrite

FairlySadPanda:
Think there has been a suggestion before for a bot action that pops up "hey this modifies a shader, this must be tested in compat and non-compat mode" on the PR as a reminder. It's a testing chore that gets skipped a lot.

PR #6210, opened 2025-09-19
SpriteComponent: allow animations to optionally stop instead of looping

OnsenCapy:
This PR introduces a new loop field to SpriteComponent, allowing sprite animations to be configured to play once instead of looping indefinitely.

Not very experienced with Robust Toolbox and I really just created this in a pinch so there may be a way to do this that I wasn't aware of. But as far as this implementation goes, to use this, you simply add to your YAML entity inside -type: Sprite a simple "loop: false" to make it so the animation only plays once and never again.


OnsenCapy:
When searching for a way to do this, I did see that Audio does allow for you to input loop: false, but it didn't seem to apply to sprite animations. Hopefully I'm not just blind.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1676,14 +1682,16 @@ internal void AdvanceFrameAnimation(RSI.State state)
                         // Animation finished, do we cycle back to positive or reset.
                         if (AnimationFrame < 0)
                         {
-                            if (Cycle)
```
ElectroJr, line 1677: you still need the cycle check, otherwise this breaks cycling/auto-reversing animations. You should probably just do `if(!loop){ ....; return}`

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -296,6 +296,12 @@ void ISerializationHooks.AfterDeserialization()
             LocalMatrix = Matrix3Helpers.CreateTransform(in offset, in rotation, in scale);
         }
 
+        /// <summary>
+        /// Used to allow one to decide if they wish to loop the animation or not.
+        /// </summary>
+        [DataField("loop")]
+        public bool Loop { get; set; } = true; // default true
```
ElectroJr:
```suggestion
        [DataField]
        public bool Loop { get; set; } = true;
```
I'm also not sure if this should be a layer or a sprite property. I.e., what if you want a complicated sprite with some looping animations, and others that only play once?

ElectroJr (changes requested):

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
LaCumbiaDelCoronavirus: shouldnt it be a property of the animation itself then, or a parameter in `Animation???System.Play` that specify whether the animation loops

or am i misunderstanding what this pr is for

ElectroJr: sprite animations aren't handled by the animation system, that's something else. The closest thing to the "animation itself" would be putting it on `RSI.State` and getting it from the RSI's meta.json, and then making the RSI's value the default behavior for a sprite layer

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
OnsenCapy: Fixed

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
OnsenCapy: It would be cool if it was per each individual layer, but I tried and I couldn't make it work.

chromiumboy:
I've been wanting this for a long time. Thank you!

PR #6213, opened 2025-09-20
Add abstract tile debug overlay & command

ElectroJr:
Adding a debug overlay generally involves quite a lot of boilerplate, especially if you want to handle somewhat nicely representing tile based data (e.g., content's atmos or explosion overlays). So it often just isn't worth the effort.

This PR tries to help with that by adding an abstract `TileDebugOverlay` class that other overlays can inherit from to make implementing this kind of simple debug overlays easier. Overlays that use this don't even necessarily need to be actually added to the game, I think its worth having the abstract class on its own just as a development tool for testing some PRs. As an example, adding a simple illumination level overlay for something like #5574 should become as simple as something like:
```cs
public sealed class LightLevelOverlay : TileFloatDebugOverlay
{
    private LightSensitiveSystem _system = default!;

    protected override void Init()
    {
        _system = Entity.System<LightSensitiveSystem>();
    }

    protected override float? GetData(Vector2i indices, Entity<MapGridComponent> grid)
    {
        var pos = Map.GridTileToWorld(grid, grid, indices);
        return _system.GetIllumination(pos);
    }
```

Resulting in something like this:

https://github.com/user-attachments/assets/a4dd6724-35f6-48c2-9a43-6aac585210e5

This PR also defines a new client side toolshed command that can be used to arbitrarily toggle any overlay that has a parameter-less constructor, so you don't need to create a new toggle command for each overlay. Note that the command can't actually be executed without #5862


Chubbygummibear:
holy shit i just now saw this and love it. this would be super helpful

PR #6215, opened 2025-09-20
Enable the ` (grave) key

amylizzle:
For some reason, this was mapped to tilde. 

OpenDream (specifically goonstation) needs it not to be.

I realise this will probably mess with people's flow on SS14 forks when opening the console, so it might not be a bad idea to create a keybind in the yaml for that


amylizzle:
Closing due to dumb

PR #6216, opened 2025-09-20
Make IReplayRecordingManager.StopRecording return the file path of the saved replay file

slarticodefast:
I am currently trying to make integration tests be able to record a replay, and I need the full file path to be able to attach it to the test results.

The file path is constructed in `TryStartRecording` from a cvar and time stamp, so there is currently no way to know which name was picked.

### DIFF
--- a/Robust.Shared/Replays/IReplayRecordingManager.cs
+++ b/Robust.Shared/Replays/IReplayRecordingManager.cs
@@ -120,7 +120,7 @@ public interface IReplayRecordingManager
     /// <summary>
     /// Stops an ongoing replay recording.
     /// </summary>
-    void StopRecording();
+    ResPath? StopRecording();
 
     /// <summary>
     /// Returns information about the currently ongoing replay recording.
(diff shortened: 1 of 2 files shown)


ElectroJr (changes requested):
`_recState!.DestPath` is already sent via the c# event `SharedReplayRecordingManager.RecordingFinished` (`ReplayRecordingFinished.Path`), and AFAIK won't event necessarily be where the replay is by the time the method finishes. E.g., `GameTicker.ReplaysOnRecordingFinished` will possibly have moved it.

If you want to find out where the replay really is in content, you probably want to either change `GameTicker.ReplaysOnRecordingFinished` (e.g., making it raise an event at the end), or change how the `moveToPath` is obtained in `GameTicker.ReplayStartRound` (again, maybe by raising an event and allowing other systems to override it)

slarticodefast:
Thanks for the feedback!
I'll see if I can find a clean way to implement this and will make a new PR if I do.
The whole way the final path is handles via `ReplayRecordState` seems a little overcomplicated with it being stored as an `Object` type. Is it fine to completely replace that?

ElectroJr:
> The whole way the final path is handles via ReplayRecordState seems a little overcomplicated with it being stored as an Object type. Is it fine to completely replace that?

I guess its meant to be there so that content store any arbitrary data related to the replay recording without any restrictions. You could probably replace it with some other class/struct, but that should itself probably still have an object field so that content could store custom data

PR #6219, opened 2025-09-22
List localization

sowelipililimute:
- pull list data from unicode CLDR JSON
- implement CLDR list formatting and gender
- new localization functions: `COUNT()`, `LIST()`, and `RECIPROCAL()`
	- `COUNT()` is the list count
	- `LIST()` formats the list according to the message locale, provided `type`, `width`
		- `wrapper` allows items to be wrapped in extra formatting without losing list gender data and is also just syntax sugar for a common pattern
	- `RECIPROCAL()` formats a reciprocal/reflexive pronoun for the provided entity or list thereof, i.e. `Alice hugged herself` and `Alice and Maria hugged each other`
- this PR is 50% tests by weight
- also https://github.com/Ygg01/Linguini/pull/60 is needed for `COUNT` to work properly with selectors, an xfail test is provided for this


Robust.Shared/Localization/CldrResource.cs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+[JsonConverter(typeof(CldrListPatternKeyJsonConverter))]
+internal record struct CldrListPatternKey(ListType Type, ListWidth Width);
+
+internal class CldrListPatternKeyJsonConverter : JsonConverter<CldrListPatternKey>
+{
+    public override CldrListPatternKey Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        return reader.GetString()! switch {
```
PJB3005, line 65: Fix compiler warning here.

Robust.Shared/Localization/CldrResource.cs
```diff
@@ -0,0 +1,202 @@
...
+            "listPattern-type-unit-short" => new(ListType.Unit, ListWidth.Short),
+        };
+    }
+    public override CldrListPatternKey ReadAsPropertyName(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        return Read(ref reader, typeToConvert, options);
+    }
+    public override void Write(Utf8JsonWriter writer, CldrListPatternKey key, JsonSerializerOptions options)
+    {
+        writer.WriteStringValue(key.Type switch {
```
PJB3005, line 84: Ditto.

Robust.Shared/Localization/CldrResource.cs
```diff
@@ -0,0 +1,202 @@
...
+            ListType.Or => "listPattern-type-or",
+            ListType.And => "listPattern-type-standard",
+            ListType.Unit => "listPattern-type-unit",
+        } + key.Width switch {
```
PJB3005, line 89: Ditto.

Robust.Shared/Localization/CldrResource.cs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+[JsonConverter(typeof(CldrCultureKeyJsonConverter))]
+internal record struct CldrCultureKey(CultureInfo Culture);
+
+internal class CldrCultureKeyJsonConverter : JsonConverter<CldrCultureKey>
+{
+    public override CldrCultureKey Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        return new(CultureInfo.GetCultureInfo(reader.GetString()!));
```
PJB3005: I don't know if this is a real concern in practice, but my understanding is that this method will throw if the host OS doesn't recognize this culture. It might be best to use the overload which takes a `bool predefinedOnly` and set it to false so a dummy value can be used instead?

Robust.Shared/Localization/CldrResource.cs
```diff
@@ -0,0 +1,202 @@
...
+    Dictionary<CldrCultureKey, CldrPersonListGender> PersonList
+)
+{
+    public static CldrSupplementalGenderData Merge(CldrSupplementalGenderData parent, CldrSupplementalGenderData child)
+    {
+        var personList = new Dictionary<CldrCultureKey, CldrPersonListGender>(child.PersonList);
+        foreach (var kvp in parent.PersonList)
+        {
+            if (!personList.ContainsKey(kvp.Key))
+                personList.Add(kvp.Key, kvp.Value);
```
PJB3005:
```suggestion
            personList.TryAdd(kvp.Key, kvp.Value);
```

Robust.Shared/Localization/ILocalizationManager.cs
PJB3005, line 1: Doc comments would be appreciated here.

Robust.Shared.IntegrationTests/Localization/ListLocalizationTests.cs
PJB3005, line 1: Does much of the CLDR data & Fluent code *need* to be duplicated here? Why not just load the engine resources in this test?

Robust.Shared/Localization/LocalizationManager.cs
```diff
@@ -385,13 +399,103 @@ public void LoadCulture(CultureInfo culture)
...
+            var parent = info;
+            while (parent != parent.Parent)
+            {
+                parent = parent.Parent;
+                if (_contexts.TryGetValue(parent, out data) && data.Item2.ListPatterns is not null)
+                    return data.Item2.ListPatterns;
+            }
+
+            return null;
+        }
```
PJB3005, line 464: Maybe unify the iteration logic here into a `T? ResolveCldrData<T>(CultureInfo, Func<CultureInfo, T>)` helper?

PJB3005: Also, I know in this case the code works fine, but it might be nice to do explicit reference reference equals here:
image

PJB3005 (changes requested):

PJB3005:
> also https://github.com/Ygg01/Linguini/pull/60 is needed for COUNT to work properly with selectors, an xfail test is provided for this

It's currently impossible to update Linguini until https://github.com/Ygg01/Linguini/issues/62 is fixed, fyi.

Robust.Shared.IntegrationTests/Localization/ListLocalizationTests.cs
sowelipililimute: it doesn't strictly _need_ to be duplicated here, but i tend to write tests as supplying their own data instead of relying on something that might be changed elsewhere

Robust.Shared/Localization/CldrResource.cs
sowelipililimute: might as well play it safe

Robust.Shared/Localization/LocalizationManager.cs
sowelipililimute: the main issue w/ unifying iteration logic here is that these are different data structures, one being supplemental data (list gender), and the other being resource data (list patterns). these are simple enough that their differences are more apparent than their similarities; the only thing that'd end up in a shared function would be the loop of `while parent != parent.Parent { parent = parent.Parent }`, and it'd just obfuscate the logic of both

sowelipililimute:
tests seem to be broken on master atm

Ygg01:
Issue https://github.com/Ygg01/Linguini/issues/62 has been closed but verify if v0.10.1 works for you. 

I say that  because, while https://github.com/Ygg01/Linguini/issues/60 did fix the issue for one type of expression (functions) it didn't solve issue for the rest (inline values, ref terms, variable references). 

This lead to making other changes to make more evaluation bit more eager. Hopefully there won't be any visible changes. 

But while Linguini isn't my main focus, I'm open for suggestions and PRs.

Robust.Shared.IntegrationTests/Localization/ListLocalizationTests.cs
```diff
@@ -0,0 +1,454 @@
...
+        var loc = IoCManager.Resolve<ILocalizationManager>();
+        loc.SetCulture(new CultureInfo(locale, false));
+
+        var list = new List<EntityUid>();
+        foreach (var person in people)
+        {
+            list.Add(entMan.CreateEntityUninitialized(person));
+        }
+
+        var listConjugation = loc.GetString("list-gender", ("item", list));
```
Ygg01, line 194: This should be `list-conjugation` right?

Robust.Shared.IntegrationTests/Localization/ListLocalizationTests.cs
```diff
@@ -0,0 +1,454 @@
...
+        [female] roja
+       *[other] roje
+    }
+   *[other] { GENDER($ent) ->
+        [male] rojos
+        [female] rojas
+       *[other] rojes
+    }
+}
+list-conjugation = { LIST($item) } { CONJUGATE-BE($item) } { -conjugate-red(ent: $item) }.
```
Ygg01, line 295: This shouldn't work in Fluent.  **Passing non-literals to a term argument is not-allowed**. See [FluentPlayground](https://projectfluent.org/play/?s=eyJtZXNzYWdlcyI6Imxpc3QtY29uanVnYXRpb24td3JvbmcgPSB7IC1jb25qdWdhdGUtcmVkKGNhc2U6ICRpdGVtKSB9LlxubGlzdC1jb25qdWdhdGlvbi1yaWdodCA9IHsgLWNvbmp1Z2F0ZS1yZWQoY2FzZTogMykgfS5cblxuLWNvbmp1Z2F0ZS1yZWQgPSBcbiAgICB7ICRjYXNlIC0-XG4gICAgICAgKltvbmVdIEZpcmVmb3hcbiAgICAgICAgW290aGVyXSBGaXJlZm9rc2llXG4gICAgfVxuIiwidmFyaWFibGVzIjp7Iml0ZW0iOjF9LCJzZXR1cCI6eyJ2aXNpYmxlIjpbIm1lc3NhZ2VzIiwib3V0cHV0IiwiY29uZmlnIl0sImxvY2FsZSI6ImVuLVVTIiwiZGlyIjoibHRyIn19).

I'm honestly surprised it passed parsing. ~~Might need to investigate on Linguini side as well.~~ Raised https://github.com/Ygg01/Linguini/issues/64 it was probably caused by introducing experimental features in the first place.

This is why I added experimental features to Linguini. With those features, you can use the TermRef as variables. But caveat emptor, they are experimental for a reason. I've tested them, but they haven't been battle tested.

Ygg01 (commented):
@sowelipililimute Hello, I went into your PR for a dive, was looking for reasons for errors.

Here is my summary of improvements:
1.  In `LocalizationManager.cs` https://github.com/space-wizards/RobustToolbox/blob/41ad95919258f9ea1fbd54e46714317dd09e5d41/Robust.Shared/Localization/LocalizationManager.cs#L505 it should be `var parser = LinguiniParser.FromFragment(contents)`
2. In `LocalizationManager.Entity.cs` lines https://github.com/space-wizards/RobustToolbox/blob/41ad95919258f9ea1fbd54e46714317dd09e5d41/Robust.Shared/Localization/LocalizationManager.Entity.cs#L72C25-L72C86 
and https://github.com/space-wizards/RobustToolbox/blob/41ad95919258f9ea1fbd54e46714317dd09e5d41/Robust.Shared/Localization/LocalizationManager.Entity.cs#L88 should be something like `((IReadBundle)bundle).FormatPattern(msg.Value, null, out var fmtErr);`. `FormatPattern` was moved out into `IReadBundle` 

EDIT: Regarding #2, you can also try switching to v0.10.2 (it should bring this method back).

PR #6221, opened 2025-09-22
Make the server keep track of the client's last real tick

DrSmugleaf:
So for lag compensation, right now it offsets by ping which is not reliable
Much more reliable way is to keep track of the client's last real tick to know how much to rewind positions on the server
For melee and (rmc) gun prediction, you can just send that in the request shoot event or melee event
Problem is, interaction events for example are handled through input events, which are engine-only so you cannot send that data inside them
This makes the engine do the last real tick book-keeping for you without bringing the entire lag compensation system into the engine

Tiny problem with this is the engine is ahead by 1 tick compared to doing it by hand in an event on content
Also other tiny problem is it spams late msgentity (see tests)

### DIFF
--- a/Robust.Client/GameController/GameController.cs
+++ b/Robust.Client/GameController/GameController.cs
@@ -26,7 +26,6 @@ using Robust.Client.WebViewHook;
 using Robust.LoaderApi;
 using Robust.Shared;
 using Robust.Shared.Asynchronous;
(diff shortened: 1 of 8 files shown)


metalgearsloth:
Going to move all of lag comp over and add testsTM trust.

PR #6226, opened 2025-09-23
Add `Crop()` and `Resize()` ImageSharp calls to Sandbox.yml

amylizzle:
Needed to do some clientside image processing on OD

### DIFF
--- a/Robust.Shared/ContentPack/Sandbox.yml
+++ b/Robust.Shared/ContentPack/Sandbox.yml
@@ -305,6 +305,32 @@ Types:
   SixLabors.ImageSharp.Formats:
     IImageEncoder: { All: True }
     PixelTypeInfo: { All: True }
+  SixLabors.ImageSharp.Processing:
+    ResizeExtensions:
+      Methods:
+      - "SixLabors.ImageSharp.Processing.IImageProcessingContext Resize(SixLabors.ImageSharp.Processing.IImageProcessingContext, SixLabors.ImageSharp.Size)"
+      - "SixLabors.ImageSharp.Processing.IImageProcessingContext Resize(SixLabors.ImageSharp.Processing.IImageProcessingContext, SixLabors.ImageSharp.Size, bool)"
+      - "SixLabors.ImageSharp.Processing.IImageProcessingContext Resize(SixLabors.ImageSharp.Processing.IImageProcessingContext, int, int)"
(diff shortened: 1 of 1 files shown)


Robust.Shared/ContentPack/Sandbox.yml
```diff
@@ -305,6 +305,32 @@ Types:
...
     PixelTypeInfo: { All: True }
+  SixLabors.ImageSharp.Processing:
+    ResizeExtensions:
+      Methods:
+      - "SixLabors.ImageSharp.Processing.IImageProcessingContext Resize(SixLabors.ImageSharp.Processing.IImageProcessingContext, SixLabors.ImageSharp.Size)"
+      - "SixLabors.ImageSharp.Processing.IImageProcessingContext Resize(SixLabors.ImageSharp.Processing.IImageProcessingContext, SixLabors.ImageSharp.Size, bool)"
+      - "SixLabors.ImageSharp.Processing.IImageProcessingContext Resize(SixLabors.ImageSharp.Processing.IImageProcessingContext, int, int)"
+      - "SixLabors.ImageSharp.Processing.IImageProcessingContext Resize(SixLabors.ImageSharp.Processing.IImageProcessingContext, int, int, bool)"
+      - "SixLabors.ImageSharp.Processing.IImageProcessingContext Resize(SixLabors.ImageSharp.Processing.IImageProcessingContext, int, int, SixLabors.ImageSharp.Processing.Processors.Transforms.IResampler)"
+      - "SixLabors.ImageSharp.Processing.IImageProcessingContext Resize(SixLabors.ImageSharp.Processing.IImageProcessingContext, SixLabors.ImageSharp.Size, SixLabors.ImageSharp.Processing.Processors.Transforms.IResampler, bool)"
```
ElectroJr, line 316: I didn't look too carefully, but I don't think `IResampler` and the implementations in `KnownResamplers` are whitelisted, so I don't think these overloads can currently be used. If you don't need them I guess its fine, but I'm just pointing it out in case it was an oversight.

ElectroJr (commented):
Seems like its fine to whitelist all those methods, but I don't actually know a lot about imagesharp and its probably better if @PJB3005 has a look

Robust.Shared/ContentPack/Sandbox.yml
amylizzle: honestly I just threw them all in because they basically all call each other and it was very frustrating trying to work out which subset I needed for my single usecase. I have grown to despise the sandbox.

PR #6228, opened 2025-09-25
Swap toolshed.nearby_* cvars

ElectroJr:
Swaps the default values of two cvars for toolshed's nearby command. I assume they were accidentally swapped, because an entity limit of 5 with a range limit of 200 doesn't really make sense.


ElectroJr:
I misunderstood the cvar

PR #6235, opened 2025-09-27
setAudioParams

jailedfish:
This PR is just a little method to put Params in component without recreating it.


Robust.Shared/Audio/Systems/SharedAudioSystem.cs
```diff
@@ -61,6 +61,14 @@ public override void Initialize()
...
     }
 
+    /// <summary>
+    ///  Set AudioParams for entity.
+    /// </summary>
+    public void SetAudioParams(Entity<AudioComponent> entity, AudioParams audioParams)
+    {
+        entity.Comp.Params = audioParams;
+    }
+
```
metalgearsloth, line 71: Won't this need dirtying so it networks properly?

PR #6237, opened 2025-09-28
Replace sprite layer keys with a LayerKey struct

ElectroJr:
This adds a new LayerKey struct that is meant to be used when working with sprite layer keys instead of working with `string`, `enum`, or `object` keys. This should generally make it easier to work with layer key data fields. Previously that always involved a decent bit of boilerplate where you would have to manually use IReflectionManager to try parse a string key as a possible enum key. Now you should be able to just have a `[DataField] LayerKey Key` and use that directly.

The new struct supports implicit casts from strings & enums. So while this is a breaking change, it doesn't seem like it actually needs a content PR.

Other changes:
- Separated various "TryGet" sprite system methods into a non-logging "TryGet" and an error logging "Resolve" method
- This also removes support for object keys, which were already disincentivized in #5602 by just not having overloads for methods that take arbitrary object keys. Now they will start throwing exceptions.

Supersedes #6025

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -39,7 +39,9 @@ END TEMPLATE-->
 
 ### New features
 
-*None yet*
+* Sprite layer keys now use a new LayerKey struct instead of string, enum, or object keys.
+  * The struct is backed by either a string or enum key, and supports implicit casts from those types.
+  * Arbitrary object keys are no longer supported.
 
 ### Bugfixes
 
--- a/Robust.Client/GameObjects/Components/Appearance/GenericVisualizerComponent.cs
+++ b/Robust.Client/GameObjects/Components/Appearance/GenericVisualizerComponent.cs
@@ -2,6 +2,7 @@ using Robust.Shared.GameObjects;
 using Robust.Shared.Serialization.Manager.Attributes;
 using System;
 using System.Collections.Generic;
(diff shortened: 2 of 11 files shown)


PJB3005:
These keys are not exclusively used by SpriteComponent. It seems like we should have a solution that covers other solutions too?

Citrea-Lingua:
Hence why I originally used Objects to cover this. So long as PrototypeCopyToShaderParameters.LayerKey can HumanoidVisualLayer data, I am happy :woman_shrugging: 
Any way I can help with this? It's been sitting for quite a while now.

PJB3005:
> Any way I can help with this? It's been sitting for quite a while now.

You can make a new PR that supersedes this one.

Citrea-Lingua:
Honestly, after looking at my changes in the superceded PR over the last few hours... I'm not actually sure I could do this any better than it is done here. The use cases for layer keys, as far as I can tell, are exactly two: Strings and Enums. Besides doing exactly this, with a proper struct to handle that data in expected ways..? The only other option *is* object, which can handle arbitrary data - but seems to not be desired here.
Mind you, while I am not new to this, I am also not the most experienced programmer, and it's very possible I am missing something - so I could very well be wrong. :woman_shrugging: 
Will give it another look, but... I don't see much room for other approaches '^_^

Issue #6152, opened 2025-08-18
We need WeakEntityReference or another solution to solve PVS error spam


slarticodefast:
## The problem:
Example error message:
```
Can't resolve "Robust.Shared.GameObjects.MetaDataComponent" on entity 5233063!
   at System.Environment.get_StackTrace()
   at Robust.Shared.GameObjects.EntityManager.GetNetEntity(EntityUid uid, MetaDataComponent metadata) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityManager.Network.cs:line 186
   at Content.Shared.Audio.Jukebox.JukeboxComponent.JukeboxComponent_AutoNetworkSystem.OnGetState(EntityUid uid, JukeboxComponent component, ComponentGetState& args) in /home/runner/work/space-station-14/space-station-14/Content.Shared/obj/Release/Robust.Shared.CompNetworkGenerator/Robust.Shared.CompNetworkGenerator.ComponentNetworkGenerator/JukeboxComponent_CompNetwork.g.cs:line 45
   at Robust.Shared.GameObjects.EntityManager.GetComponentState(IEventBus eventBus, IComponent component, ICommonSession session, GameTick fromTick) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityManager.Components.cs:line 1704
   at Robust.Server.GameStates.PvsSystem.GetEntityState(ICommonSession player, EntityUid entityUid, GameTick fromTick, MetaDataComponent meta) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Server/GameStates/PvsSystem.GetStates.cs:line 52
   at Robust.Server.GameStates.PvsSystem.AddPvsChunk(PvsChunk chunk, Single distance, PvsSession session) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Server/GameStates/PvsSystem.ToSendSet.cs:line 62
   at Robust.Server.GameStates.PvsSystem.GetEntityStates(PvsSession session) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Server/GameStates/PvsSystem.cs:line 311
   at Robust.Server.GameStates.PvsSystem.SerializeSessionState(PvsSession data) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Server/GameStates/PvsSystem.Serialize.cs:line 59
[...]

PR #6245, opened 2025-10-07
Weak Entity Reference

EmoGarbage404:
Basic weak entity reference implementation based on ElectroSr and Tayrtahn's implementation from #5577 combined with PJB's review from #6114 

Apologies for not preserving git history as my PR process was a bit scattershot and not exactly the cleanest. The majority of this PR is the work of others which I've done my best to correct according to review and my own beliefs.

Weak Entity Reference is simply an EntityUid wrapper that requires the TryGetEntity function to be used to get the contained entity. The check ensures that the referenced entity always exists, hopefully preventing content errors due to deleted references. Additionally, the struct networks as invalid if the referenced entity is no longer valid, preventing errors from attempting to get the metadata component.


Robust.Shared/GameObjects/WeakEntityReference.cs
```diff
@@ -0,0 +1,79 @@
...
+[CopyByRef, Serializable, NetSerializable]
+public readonly struct WeakEntityReference(EntityUid uid) : IEquatable<WeakEntityReference>, IComparable<WeakEntityReference>
+{
+    internal readonly int Id = uid.Id;
+
+    /// <summary>
+    /// The underlying entity this reference is pointing to.
+    /// You should not be accessing this directly!
+    /// </summary>
+    internal EntityUid Entity => new(Id);
```
EmoGarbage404, line 34: I would prefer to just have these be private members and then have the `TryGetEntity` function be on the struct. I think being internal is a good compromise that at least prevents content from accessing the entityUid directly.

Robust.Shared.IntegrationTests/EntitySerialization/WeakEntityReferenceTest.cs
```diff
@@ -0,0 +1,105 @@
...
+        // Delete the referenced entity on the server
+        // Make sure that the sending the new state does not cause errors
+        await server.WaitAssertion(() =>
+        {
+            var entA = sEntMan.GetEntity(netEntA);
+            var comp = sEntMan.GetComponent<WeakEntityReferenceTestComponent>(entA);
+            var entB = sEntMan.GetEntity(netEntB);
+            sEntMan.DeleteEntity(entB);
+            sEntMan.Dirty(entA, comp);
+        });
```
EmoGarbage404, line 93: I'm not sure if this actually checks if nothing caused errors on the client or server. If there's a better way to do this, let me know.

EmoGarbage404:
Got around to fixing the tests and adding in some missing API stuff (ToPrettyString, resolves, etc.)

Should be good to go now

PJB3005 (changes requested):
Why do `WeakEntityReference`s still have a concept of invalid? Especially in the serialization code, from my reading, deserialization of a broken weak entity reference would cause an error.

EmoGarbage404:
> Why do `WeakEntityReference`s still have a concept of invalid? 

Invalid is just meant to serve as the default/uninitialized state of the `WeakEntityReference`. In the prior PR this was being done with both null as well as invalid, which was kinda unnecessarily confusing. I opted to use Invalid for this value since, in the context of usage of a `WeakEntityReference`, there really shouldn't be a meaningful distinction between an entity being missing or being unable to be retrieved (deleted).

> Especially in the serialization code, from my reading, deserialization of a broken weak entity reference would cause an error.

I've added proper reading for invalid values now. I must've missed it, since the values were properly validated in other parts of the serialization.

Also as an aside I think the github tests exploded for some unrelated reason to the PR.

FairlySadPanda:
I was interested in this as a comparison so I compared this to Unreal Engine's concept of a TWeakRef<T>, and there their default state is to wrap a nullptr of type `Object*`, and then you check `IsValid()` to check if not null and then that it hooks up to a valid object via their implementation's weak reference manager. So a TWeakRef can be "invalid" whilst also not being nullptr.

I took a look and I do think that `public bool TryGetEntity(WeakEntityReference weakRef, [NotNullWhen(true)] out EntityUid? entity)` looks off. If the internal value can either be `EntityUid.Invalid` or a reference, then maybe it should be signatured as `public bool TryGetEntity(WeakEntityReference weakRef, out EntityUid entity)`, because otherwise it's introducing nullability rather than orienting consistently around `EntityUid.Invalid` being empty/invalid.

Thoughts? Happy to be wrong here.

EmoGarbage404:
> I took a look and I do think that `public bool TryGetEntity(WeakEntityReference weakRef, [NotNullWhen(true)] out EntityUid? entity)` looks off. If the internal value can either be `EntityUid.Invalid` or a reference, then maybe it should be signatured as `public bool TryGetEntity(WeakEntityReference weakRef, out EntityUid entity)`, because otherwise it's introducing nullability rather than orienting consistently around `EntityUid.Invalid` being empty/invalid.
> 
> Thoughts? Happy to be wrong here.

The `TryGetEntity` method uses a nullable out parameter to just mirror the general convention of "TryGet" functions which return null when failing. I think this makes using them more intuitive, as it prevents a situation where the `TryGetEntity` function is handled incorrectly and the invalid `EntityUid` is passed into some function. Nullables additionally have significantly better linting in C# than our bespoke invalid default value, so I think orienting the API like this results in a much better end user experience.

FairlySadPanda:
That is true, thanks for the response.

Robust.Shared.IntegrationTests/EntitySerialization/WeakEntityReferenceTest.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        // Delete the referenced entity on the server
+        // Make sure that the sending the new state does not cause errors
+        await server.WaitAssertion(() =>
+        {
+            var entA = sEntMan.GetEntity(netEntA);
+            var comp = sEntMan.GetComponent<WeakEntityReferenceTestComponent>(entA);
+            var entB = sEntMan.GetEntity(netEntB);
+            sEntMan.DeleteEntity(entB);
```
Tayrtahn, line 91: You need some asserts after deleting the entity that makes sure that calling `TryGetEntity` on the reference now returns false.

Robust.Shared/GameObjects/WeakEntityReference.cs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public bool Equals(WeakEntityReference other)
+    {
+        return Id == other.Id;
+    }
+
+    public override bool Equals([NotNullWhen(true)] object? obj)
+    {
+        if (ReferenceEquals(null, obj)) return false;
```
Tayrtahn, line 61:
```suggestion
        if (obj is null) return false;
```
IDE suggested change.

Robust.Shared/GameObjects/WeakEntityReference.cs
```diff
@@ -0,0 +1,79 @@
...
+/// should continue to be valid, which also means you do not need to clean up any references upon deletion or
+/// component removal.
+/// </remarks>
+/// <remarks>
+/// When saving a map, any weak references to entities that are not being included in the save file are automatically
+/// ignored.
+/// </remarks>
+[CopyByRef, Serializable, NetSerializable]
+public readonly struct WeakEntityReference(EntityUid uid) : IEquatable<WeakEntityReference>, IComparable<WeakEntityReference>
+{
```
Tayrtahn, line 27:
```suggestion
{
    public WeakEntityReference(EntityUid? uid) : this(uid != null ? uid.Value : EntityUid.Invalid) { }
```
A constructor able to take in null is very helpful for working with the the `Audio` API, or anything that returns an `EntityUid?`.

Robust.Shared/GameObjects/EntityUid.cs
```diff
@@ -133,6 +133,12 @@ public static explicit operator int(EntityUid self)
             return self.Id;
         }
 
+        /// <summary>
+        ///     Implicit conversion to WeakEntityReference for easy assignment.
+        ///     Note that conversion from WeakEntityReference -> EntityUid is explicit through <see cref="IEntityManager.TryGetEntity(WeakEntityReference, out EntityUid?)"/>
+        /// </summary>
+        public static implicit operator WeakEntityReference(EntityUid self) => new(self);
```
Tayrtahn, line 140:
```suggestion
        public static implicit operator WeakEntityReference(EntityUid? self) => new(self);
```
See previous comment.

Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -1053,6 +1053,17 @@ public void RunMapInit(EntityUid entity, MetaDataComponent meta)
...
+        {
+            entity = null;
+            if (weakRef.Entity == EntityUid.Invalid
+                || !EntityExists(weakRef.Entity))
+                return false;
+
+            entity = weakRef.Entity;
+            return true;
+        }
+
```
Tayrtahn, line 1066: It would be nice to also have `EntityUid? GetEntity(WeakEntityReference weakRef)` to use with null coalesce and APIs that accept nullable uids.

Otherwise there's a lot of `TryGetEntity(weakRef, out var ent) ? ent : null)`.

Tayrtahn (changes requested):
Requesting a few API tweaks based on my experience updating my old Jukebox fix PR to use this version of WeakEntityReference.

Tayrtahn:
Hm, this seems to have serialization errors when used as a component datafield:

```
Failed TestAllServerPrototypesAreSerializable [1 m 39 s]
  Error Message:
   Error when validating server prototype (EntityPrototype, Jukebox). Errors: 
 - Caught exception while writing. Exception: System.InvalidOperationException: No data definition found for type Robust.Shared.GameObjects.WeakEntityReference when writing
   at Robust.Shared.Serialization.Manager.SerializationManager.WriteValueInternal[T](T value, DataDefinition`1 definition, Boolean alwaysWrite, ISerializationContext context) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/Serialization/Manager/SerializationManager.Writing.cs:line 246
   at lambda_method278083(Closure, WeakEntityReference, Boolean, ISerializationContext)
   at lambda_method230500(Closure, JukeboxComponent, ISerializationContext, Boolean)
   at lambda_method278045(Closure, JukeboxComponent, Boolean, ISerializationContext)
   at lambda_method277981(Closure, Object, Boolean, ISerializationContext)
   at Robust.Shared.Serialization.TypeSerializers.Implementations.ComponentRegistrySerializer.Write(ISerializationManager serializationManager, ComponentRegistry value, IDependencyCollection dependencies, Boolean alwaysWrite, ISerializationContext context) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/Serialization/TypeSerializers/Implementations/ComponentRegistrySerializer.cs:line 163
   at lambda_method247714(Closure, ComponentRegistry, Boolean, ISerializationContext)
   at lambda_method219023(Closure, EntityPrototype, ISerializationContext, Boolean)
   at lambda_method247713(Closure, EntityPrototype, Boolean, ISerializationContext)
   at lambda_method247712(Closure, Object, Boolean, ISerializationContext)
   at Robust.Shared.Prototypes.PrototypeManager.ValidateProto(Type type, IPrototype instance, ISerializationContext ctx, Boolean& caughtException) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/Prototypes/PrototypeManager.YamlValidate.cs:line 168

  Stack Trace:
     at Content.IntegrationTests.Tests.PrototypeTests.PrototypeTests.SaveThenValidatePrototype(IntegrationInstance instance, String instanceId, TestEntityUidContext ctx) in /home/runner/work/space-station-14/space-station-14/Content.IntegrationTests/Tests/PrototypeTests/PrototypeTests.cs:line 49
   at Content.IntegrationTests.Tests.PrototypeTests.PrototypeTests.TestAllServerPrototypesAreSerializable() in /home/runner/work/space-station-14/space-station-14/Content.IntegrationTests/Tests/PrototypeTests/PrototypeTests.cs:line 22
   at Content.IntegrationTests.Tests.PrototypeTests.PrototypeTests.TestAllServerPrototypesAreSerializable() in /home/runner/work/space-station-14/space-station-14/Content.IntegrationTests/Tests/PrototypeTests/PrototypeTests.cs:line 23
   at NUnit.Framework.Internal.TaskAwaitAdapter.GenericAdapter`1.BlockUntilCompleted()
   at NUnit.Framework.Internal.Commands.TestMethodCommand.Execute(TestExecutionContext context)
   at NUnit.Framework.Internal.ContextUtils.<>c__DisplayClass1_0`1.<DoIsolated>b__0(Object _)

1)    at Content.IntegrationTests.Tests.PrototypeTests.PrototypeTests.<>c__DisplayClass2_0.<SaveThenValidatePrototype>b__1() in /home/runner/work/space-station-14/space-station-14/Content.IntegrationTests/Tests/PrototypeTests/PrototypeTests.cs:line 60
   at NUnit.Framework.Assert.Multiple(TestDelegate testDelegate)
   at Content.IntegrationTests.Tests.PrototypeTests.PrototypeTests.SaveThenValidatePrototype(IntegrationInstance instance, String instanceId, TestEntityUidContext ctx) in /home/runner/work/space-station-14/space-station-14/Content.IntegrationTests/Tests/PrototypeTests/PrototypeTests.cs:line 49
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.ExecutionContextCallback(Object s)

2)    at Robust.UnitTesting.Pool.TestPair`2.OnDirtyDispose() in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.UnitTesting/Pool/TestPair.Recycle.cs:line 28
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at Robust.UnitTesting.Pool.TestPair`2.OnDirtyDispose()
   at Robust.UnitTesting.Pool.TestPair`2.DisposeAsync() in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.UnitTesting/Pool/TestPair.Recycle.cs:line 108
   at Robust.UnitTesting.Pool.TestPair`2.DisposeAsync()
   at Content.IntegrationTests.Tests.PrototypeTests.PrototypeTests.TestAllServerPrototypesAreSerializable() in /home/runner/work/space-station-14/space-station-14/Content.IntegrationTests/Tests/PrototypeTests/PrototypeTests.cs:line 23
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.ExecutionContextCallback(Object s)
```

See https://github.com/space-wizards/space-station-14/pull/38827

EmoGarbage404:
# I'm going to close this PR for a handful of reasons and I would like to put WER to rest

After a lot of thought and discussions with others about technical design in ECS, I think that adding WER in the current state of the repo is a mistake. I think that it's addition into the codebase would inevitably degrade the quality of code and introduce novel issues to the detriment of a ton of systems.

## Weak Entity Reference's Purpose

Weak Entity Reference and weak references in general serve to allow code to hold onto ephemeral actors and gameobjects without really caring if they exist or not. This is admittedly really niche and not really something that you see super commonly in the codebase. Most things _do_ explicitly care about making sure their references still exist, because the deletion of these entities is typically core to their functionality and is something that needs to be accounted for.

Unreal engine has a similar concept to WER called "weak pointers." I think comparing the two is useful because it highlights a realistic usage. Weak pointers in UE are basically explicitly for garbage collection purposes, as a typical reference like a shared pointer explicitly prevents the actor from getting garbage collected. In RT, your entity can be deleted and junked at any point, regardless of if you are holding a reference to it or not. In this sense, RT's WER basically is just the coder going "i really don't care about the lifestage of this entity" which is something that i seriously doubt is being done for legitimate technical reasons.

## Relations

The proper thing that a lot of content code needs is _relations._ A proper relations system would allow systems to automatically update component fields when entities get deleted, creating automatic hooks and preventing invalid entities from existing or serializing at all.

Relations are already informally used all the time in the content codebase. Any time an entity gets a component that subs to deletion and sets a field on another ent to null, you're seeing a heavy-boilerplate relation happening. This has been the standard for managing this kinda thing for as long as i've been on the project, and it's a good system, barring the obvious issues around boilerplate.

The fact of the matter is, we need to actually care about the lifestage of our entities because just letting them vanish has obvious bad implications. First of all, we don't even know _when_ an entity is getting deleted or cleaned up. Entities getting deleted when they shouldn't be should _100% throw errors._ If code isn't accounting for things getting deleted properly that is _100% a bug that needs to be fixed._ These are things that a proper relation system could handle well and would greatly improve the code quality and reduce the amount of boilerplate needed for entity state management.

## Abuses of WER

Part of why i do not want this feature to be merged is that the vast majority of uses that people have been excitedly waiting to apply this functionality to are explicitly bad. 

The average application of WER that i see looks something like this:
- Entity A has a Component X holds a reference to Entity B
- Entity B gets deleted
- Component X's field never gets updated
- On usage, Component X's field throws an error
- To fix, we replace Component X's field with a WER

**This is 100% an incorrect usage of a WER.** Pasting over bad lifestage handling and requiring every usage of the entity to recheck its existence is not a correct way to solve this.

Issue #6152 identifies the correct solution but erroneously labels it as a "workaround" for some reason:
> Add a marker component to the referenced entity the resets the reference to null when the marker component shuts down.

It additionally has this statement on the usage of relations with regards to security:
> Also this solution is problematic regarding cheat clients, as such a marker component has to be networked as well and can be used to obtain secret information about the round, for example when marking the target player of a kill objective.

I think this is taking an overly straightforward approach to a problem that is significantly more complex. Networking data to the client is not a requirement and is certainly not necessary for all applications, especially something like an objective. In fact, my [own rewrite of objective code](https://github.com/EphemeralSpace/ephemeral-space/pull/382) uses marker components to track secret targets, simply opting to store the information on the server  and only network what limited information is needed via an intermediary entity and session overrides.

I do not think that WER has real applications in any of the scenarios i've seen listed. In the vast majority of them, it would be better served by a more robust relations system that can manage entity lifespan and allow code to update as needed, rather than forcing it into a state of ambivalence.

In any case, apologies to maintainers for wasting time with this PR. It was a good learning experience for myself but obviously it's annoying to have things dangle.

ike709:
I would put a copy of that rationale in #6152 as well where it's more likely to be seen by people working on this problem.

PR #6249, opened 2025-10-09
Add linux wayland support to claudia

VasilisThePikachu:
Could not do x11 because quote "oh my god x11 support might be a hecking nightmare" -pjb

but the skeleton is there i guess

### DIFF
--- a/Robust.Client.Graphics.Rhi/WebGpu/RhiWebGpu.Window.cs
+++ b/Robust.Client.Graphics.Rhi/WebGpu/RhiWebGpu.Window.cs
@@ -39,26 +39,36 @@ internal sealed unsafe partial class RhiWebGpu
         surfaceDesc.nextInChain = (WGPUChainedStruct*)(&surfaceDescMetal);
 
 #elif LINUX
-        // TODO: Linux surface creation
-        /*
-        WGPUSurfaceSourceXlibWindow surfaceDescX11;
-        var xDisplay = _clyde._windowing.WindowGetX11Display(window);
-        var xWindow = _clyde._windowing.WindowGetX11Id(window);
-
-        if (xDisplay != null && xWindow != null)
+        // Prioritise wayland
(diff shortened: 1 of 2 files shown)


Robust.Client.Graphics.Rhi/WebGpu/RhiWebGpu.Window.cs
```diff
@@ -39,26 +39,36 @@ private WindowData CreateSurfaceForWindow(in RhiWindowSurfaceParams surfaceParam
         surfaceDesc.nextInChain = (WGPUChainedStruct*)(&surfaceDescMetal);
 
 #elif LINUX
-        // TODO: Linux surface creation
-        /*
-        WGPUSurfaceSourceXlibWindow surfaceDescX11;
-        var xDisplay = _clyde._windowing.WindowGetX11Display(window);
-        var xWindow = _clyde._windowing.WindowGetX11Id(window);
+        // Prioritise wayland
```
PJB3005: This comment doesn't make sense. There's no prioritization happening here.

Robust.Client.Graphics.Rhi/WebGpu/RhiWebGpu.Window.cs
```diff
@@ -39,26 +39,36 @@ private WindowData CreateSurfaceForWindow(in RhiWindowSurfaceParams surfaceParam
...
+        if (surfaceParams.Wayland)
+        {
+            var surfaceDescWayland = new WGPUSurfaceSourceWaylandSurface
```
PJB3005: Move these variables outside the if/else block. I'm not sure it's really an issue but the lifetimes concern me.

PJB3005 (changes requested):

PJB3005 (changes requested):
Move only the variables. Do not initialize them outside the if/else block.

PJB3005 (approved):

PR #6251, opened 2025-10-13
Toolshed accept and return type commands for searching through toolshed commands

juliangiebel:
This is a patch from @moonheart08

This PR adds four new commands:
cmd:accepts = Filters commands based on what type they accept as a piped argument
cmd:accepts_generic = Filters commands based on what type they accept as a piped argument, allowing commands that broadly accept most types

cmd:returns = Filters commands based on what type they return as a result
cmd:returns_generic = Filters commands based on what type they return as a result, allowing commands that broadly accept most types

### DIFF
--- a/Robust.Shared/Toolshed/Commands/Misc/CmdCommand.cs
+++ b/Robust.Shared/Toolshed/Commands/Misc/CmdCommand.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using Robust.Shared.Toolshed.Syntax;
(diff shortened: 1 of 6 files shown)


Robust.Shared/Toolshed/Commands/Misc/CmdCommand.cs
```diff
@@ -22,6 +23,103 @@ public string Moo()
     [CommandImplementation("info")]
     public CommandSpec Info(CommandSpec cmd) => cmd;
 
+    [CommandImplementation("accepts")]
+    public IEnumerable<CommandSpec> Accepts([PipedArgument] IEnumerable<CommandSpec> specs, Type pipedArgument)
+        => Accepts_Impl(specs, pipedArgument, false);
+
+    // TODO: This should be an optional parameter, --generic, instead.
+    [CommandImplementation("accepts_generic")]
```
ElectroJr, line 31: Optional parameters are supported, though not with the `--<argName>` syntax, so you're forced to either specify them or use an explicit pipe `|` to chain the output into another command.

If we want the `--` syntax, I guess its fine to keep the TODO, but there should probably be an issue for the missing feature.

Robust.Shared/Toolshed/Commands/Misc/CmdCommand.cs
```diff
@@ -22,6 +23,103 @@ public string Moo()
...
+        // This did not get made into methods on ToolshedCommand because of the fitness thing.
+        // For user-friendliness reasons this always orders the most exactly matching commands first.
+        return specs.Select(cmd =>
+        {
+            var fitness = -1; // unfit, ditch.
+
+            var impls = cmd.Cmd.CommandImplementors[cmd.SubCommand ?? string.Empty]
+                .MethodsByPipedType(pipedArgument, allowGeneric);
+
+            var anyNonGeneric = false;
```
ElectroJr, line 46: AFAICT anyNonGeneric is equivalent to `fitness >= 1`. So IMO it'd be clearer if the bool were just removed, or simply defined using `fitness ` after the `foreach` block.

Its not really that important, but IMO its a bit more readable.

Robust.Shared/Toolshed/Commands/Misc/CmdCommand.cs
```diff
@@ -22,6 +23,103 @@ public string Moo()
...
+                // if any commands at all, raise it to 0, but don't modify higher fitness.
+                fitness = int.Max(fitness, 0);
+
+                // We treat object like a generic param because that's "probably what the user wants"
+                if (!impl.PipeGeneric && (impl.PipeArg!.ParameterType != typeof(object)))
+                {
+                    if (impl.PipeArg!.ParameterType == pipedArgument)
+                        fitness = int.Max(fitness, 2); // Exact match
+                    else
+                        fitness = int.Max(fitness, 1);
```
ElectroJr, line 59: why not just
```suggestion
                        return (cmd, fitness: 2); // Exact match
                    fitness = 1;
```

Robust.Shared/Toolshed/ToolshedCommandImplementor.cs
```diff
@@ -411,8 +411,28 @@ internal CommandArgument GetCommandArgument(ParameterInfo arg)
                 if (pipedType == null)
                     return false;
 
-                return x.Generic || _toolshed.IsTransformableTo(pipedType, param.ParameterType);
```
ElectroJr, line 414: This gets rid of the `x.Generic` check, which is still needed. `IsTransformableTo` just doesn't support generics atm and currently still needs to bypass it. IIRC generics are basically handled implicitly further down in the try-catch block with 
```cs
var t = GetGenericTypeFromPiped(pipedType!, x.PipeArg!.ParameterType);
return (x, x.Info.MakeGenericMethod(typeArguments?.Append(t).ToArray() ?? [t]));
```

The way toolshed handles generics is just pretty janky overall, it doesn't really handle generic type constraints at all, it just checks it implicitly by seeing if `MakeGenericMethod` throws. So I think you either need to shuffle logic around to get the a concrete method & parameter types before calling IsTransformableTo, or you just can't quite do this (its currently causing test fails)

ElectroJr (changes requested):

PR #6257, opened 2025-10-19
Fix ContactEnumerator.Empty causing NullReferenceException

ToastEnjoyer:
This caused a weird error that showed up on grafana. this should fix this error for good.
Very simple code


Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs
```diff
@@ -856,7 +856,7 @@ public ContactEnumerator(FixturesComponent? fixtures, bool includeDeleting = fal
 
         if (fixtures == null || fixtures.Fixtures.Count == 0)
         {
-            this = Empty;
+            _empty = true;
```
metalgearsloth, line 875: Would it be better to just add 2 static dictionaries to use for the enumerators (I have no idea if movenext is thread-safe on an empty dictionary).

ElectroJr: AFAIK it should be thread safe as long as nothing modifies the dictionaries, so we could do that if you think it matters? I guess it might be slightly better for performance but I don't know if that matters here.

ElectroJr:
A fix for this exception is still needed, I think this PR was closed unintentionally while several other PRs were being closed.

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Contacts.cs
ElectroJr: Creating 2 private empty static dictionaries just to access their enumerators seems unnecessarily convoluted just to slightly speed up enumeration so I'd rather keep it as is

PR #6259, opened 2025-10-21
Fixture bug fix

### LINKED ISSUE #6258: Entities don't get properly woken when being moved from nullspace.

### LINKED ISSUE #40606: Fixtures don't work on entities spawned by gamerules until said entity moves

last commit on the master branch: 27f281add5a608c3a90abbbba218fb63075278fa

### Description

When a gamerule spawns entities. they have fixtures turned off until they move. they can still be attacked via melee. but projectiles don't work. 

Spawning entities via the F5 menu doesn't cause this issue.

Image

### Reproduction

1. Add a gamerule like Nukeops
2. Take a ghostrole (but once you get it, DON'T MOVE)
3. Aghost and try to shoot it with any gun

### Additional Context

For some reason this bug didn't happen with a reinforcement monkey the nukeops rule spawned

ToastEnjoyer:
Fixes https://github.com/space-wizards/space-station-14/issues/40606
Fixes https://github.com/space-wizards/RobustToolbox/issues/6258

Added a check in `SharedTransformSystem` to detect when an entity moves from `nullspace` to a map and automatically ensure it has `PhysicsComponent` and `FixturesComponent` before calling `OnParentChange`, so the physics system can register fixtures and enable collisions immediately.

https://github.com/user-attachments/assets/d0edce6b-35d9-4c8d-9a76-be44f9e399c5


Robust.Shared/Physics/Systems/SharedPhysicsSystem.cs
```diff
@@ -162,64 +162,80 @@ private void UpdateSubsteps(int obj)
...
         internal void OnParentChange(Entity<TransformComponent, MetaDataComponent> ent, EntityUid oldParent, EntityUid? oldMap)
-        {
-            // We do not have a directed/body subscription, because the entity changing parents may not have a physics component, but one of its children might.
-            var (uid, xform, meta) = ent;
-
-            // If this entity has yet to be initialized, then we can skip this as equivalent code will get run during
-            // init anyways. HOWEVER: it is possible that one of the children of this entity are already post-init, in
-            // which case they still need to handle map changes. This frequently happens when clients receives a server
-            // state where a known/old entity gets attached to a new, previously unknown, entity. The new entity will be
-            // uninitialized but have an initialized child.
```
Princess-Cheeseballs, line 173: Probably shouldn't delete the code comments.

Princess-Cheeseballs (changes requested):

Robust.Shared/Physics/Systems/SharedPhysicsSystem.cs
ToastEnjoyer: when did I do that? ah fixing it right now

Robust.Shared/Physics/Systems/SharedPhysicsSystem.cs
```diff
@@ -162,64 +162,86 @@ private void UpdateSubsteps(int obj)
...
+{
+    DebugTools.Assert(!Deleted(uid));
+    _joints.ClearJoints(uid);
+
+    if (enteringMap)
+    {
+        if (body != null)
         {
-            RecursiveMapUpdate(uid, xform, body);
+            var fixtures = _fixturesQuery.CompOrNull(uid);
```
metalgearsloth: You haven't described what's happening or what the fix is so I'm hesitant to even merge it without looking at it personally.

metalgearsloth: Also I'd significantly prefer to not start introducing physicsmap handling back into the codebase after just ripping it out due to the hellbugs it caused.

ToastEnjoyer: > You haven't described what's happening or what the fix is so I'm hesitant to even merge it without looking at it personally.

The system forces entities with the fixture component to update immediately on map change. I didn't word the PR very well, sorry about that

ToastEnjoyer: > Also I'd significantly prefer to not start introducing physicsmap handling back into the codebase after just ripping it out due to the hellbugs it caused.

What do you suggest I do instead?

metalgearsloth: > > You haven't described what's happening or what the fix is so I'm hesitant to even merge it without looking at it personally.
> 
> The system forces entities with the fixture component to update immediately on map change. I didn't word the PR very well, sorry about that

Why do they need to be doing it a second time there, the broadphase should already be getting handled elsewhere? That should be fixed instead.

Robust.Shared/Physics/Systems/SharedPhysicsSystem.cs
```diff
@@ -162,64 +162,92 @@ private void UpdateSubsteps(int obj)
...
-                {
-                    PhysicsQuery.TryGetComponent(child, out var childBody);
-                    RecursiveMapUpdate(child, childXform, childBody);
-                }
+                foreach (var fixture in fixtures.Fixtures.Values)
+                    _broadphase.TouchProxies(fixture);
             }
+
         }
+    }
```
Pok27: !!!!!
```suggestion
            }
        }
    }
```

ToastEnjoyer:
I'll fix the build errors in a few hours, I got so sidetracked yesterday

Robust.Shared/GameObjects/Systems/SharedTransformSystem.cs
```diff
@@ -283,9 +284,32 @@ internal void RaiseMoveEvent(
...
+                    _physics.SetCanCollide(uid, true, manager: null, body: phys);
+                    _physics.WakeBody(uid, body: phys);
+                }
+
+                _physics.OnParentChange(ent, oldParent, oldMap);
+
+                var entParentChangedMessage = new EntParentChangedMessage(
+                    ev.Sender, oldParent, oldMap, ev.Component);
+
                 RaiseLocalEvent(ev.Sender, ref entParentChangedMessage, true);
```
metalgearsloth, line 313: 1. We can't just add physicscomp to every entity in the game.
2. This needs test coverage to assert what is happening (and fails) and with an associated fix.

When I just did nullspace > broadphase it worked fine so you'll need to make one that better reflects what's happening.

PR #6260, opened 2025-10-22
Fix PredictedQueueDeleteEntity mispredicts

ElectroJr:
Fixes client-side predicted entity deletion ( `EntityManager.PredictedQueueDeleteEntity`) so that it behaves more like it does on the server. In particular, entities will be deleted on the same tick after all system have been updated. Previously, it would process deletions at the beginning of the next tick, and it wouldn't process deletions while re-predicting ticks at all

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -43,7 +43,7 @@ END TEMPLATE-->
 
 ### Bugfixes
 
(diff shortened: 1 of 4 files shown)


ToastEnjoyer (approved):
Code looks good to me

PR #6264, opened 2025-10-22
Update child controls style when a control is restyled

### LINKED ISSUE #40923: Inconsistent guidebook collapsible UI
This is relatively minor, but the little triangles that show whether an element can be expanded are inconsistent. In the navigation tree, they rotate when expanded, while within the guidebook they always point down, which might lead players to think the element is already expanded and doesn't contain any information.

https://github.com/user-attachments/assets/55bfa276-d49e-4cd1-9d8f-89d4cc78f4ce

eoineoineoin:
Mildly-controversial PR incoming.

I was looking to fix https://github.com/space-wizards/space-station-14/issues/40923 - in that issue, the problematic element is a `CollapsibleHeading`. The `CollapsibleHeading` is a `ContainerButton` with a child `TextureRect`, which is what is used to display the triangles.

So, the following style rules _should_ do the trick. Use `collapsedTexture` normally, but `expandedTexture` when the heading has been pressed:

```
            E<CollapsibleHeading>()
                .ParentOf(E<BoxContainer>())
                .ParentOf(E<TextureRect>().Class(OptionButton.StyleClassOptionTriangle))
                .Prop(TextureRect.StylePropertyTexture, collapsedTexture),

            E<CollapsibleHeading>().PseudoPressed()
                .ParentOf(E<BoxContainer>())
                .ParentOf(E<TextureRect>().Class(OptionButton.StyleClassOptionTriangle))
                .Prop(TextureRect.StylePropertyTexture, expandedTexture),
```

However, this only works once, and can't be changed once the `CollapsibleHeading` is visible. Pressing the `CollapsibleHeading` will toggle it's pressed pseudo state, which causes the `CollapsibleHeading` to be restyled. However, the style rules are targetting the child `TextureRect`, which never gets queued to refresh it's style.

This PR enqueues a style update for any child `Control`s whenever a parent `Control`'s style is updated. Potentially, this is a bunch more style re-evalulations, but not sure there's a cleverer way to do this, since `Selector.Matches` could be doing anything, so no real way to know if child styles are affected.

### DIFF
--- a/Robust.Client/UserInterface/Controls/Button.cs
+++ b/Robust.Client/UserInterface/Controls/Button.cs
@@ -21,11 +21,6 @@ namespace Robust.Client.UserInterface.Controls
             AddChild(Label);
         }
 
-        protected override void StylePropertiesChanged()
-        {
-            base.StylePropertiesChanged();
-        }
-
         /// <summary>
         ///     How to align the text inside the button.
         /// </summary>
--- a/Robust.Client/UserInterface/Controls/Button.cs
+++ b/Robust.Client/UserInterface/Controls/Button.cs
@@ -1,4 +1,4 @@
-﻿using Robust.Shared.ViewVariables;
+using Robust.Shared.ViewVariables;
 using static Robust.Client.UserInterface.Controls.Label;
 
 namespace Robust.Client.UserInterface.Controls
@@ -24,11 +24,6 @@ namespace Robust.Client.UserInterface.Controls
         protected override void StylePropertiesChanged()
         {
             base.StylePropertiesChanged();
-
-			// Temporary workaround to fix button styles in content thanks to a lack of correct style updating for child selectors.
-			// The changing of the style class on the parent control (button in this case) doesn't cause a style update in the label correctly currently.
-			// This works around that.
-            Label.Restyle();
         }
 
         /// <summary>
--- a/Robust.Client/UserInterface/Stylesheet.cs
+++ b/Robust.Client/UserInterface/Stylesheet.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using JetBrains.Annotations;
@@ -427,7 +427,6 @@ namespace Robust.Client.UserInterface
 
     // Temporarily hidden due to performance concerns.
     // Like seriously this thing is awful performance wise.
-    // Also, you can't just enable it. The code is here but AddChild etc only do restyles one level deep.
     internal sealed class SelectorDescendant : Selector
     {
         public SelectorDescendant([NotNull] Selector ascendant, [NotNull] Selector descendant)
@@ -468,7 +467,10 @@ namespace Robust.Client.UserInterface
         }
     }
 
-    // Temporarily hidden due to performance concerns.
+    /// <summary>
+    ///     Given a pair of selectors, will match the element described by the child selector
+    ///     when the child is an immediate child of parent.
+    /// </summary>
     public sealed class SelectorChild : Selector
     {
         public SelectorChild(Selector parent, Selector child)
--- a/Robust.Client/UserInterface/Stylesheet.cs
+++ b/Robust.Client/UserInterface/Stylesheet.cs
@@ -314,7 +314,24 @@ namespace Robust.Client.UserInterface
 
     public abstract class Selector
     {
+        /// <summary>
+        ///     Returns true if this selector should apply to input control
+        /// </summary>
         public abstract bool Matches(Control control);
+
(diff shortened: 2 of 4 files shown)


PJB3005:
This should really check if the control is possibly involved in a child selector, and only do the recalculation in that case.

Absotively:
`Button.StylePropertiesChanged()` contains a workaround for this; it manually calls `Restyle()` on one of its children. It still doesn't check whether the calculation is actually needed, but it's applied only to a case where it is often needed.

Also, while `CollapsibleHeading` could use this workaround, controls in content can't, as `Restyle()` is not available to them.

I admit I'm pointing this out due to self-interest, as it affects [switch buttons](<https://github.com/space-wizards/space-station-14/pull/39161>). But I do think having children update automatically is likely to be more generally useful. And the comment on the workaround in `Button` makes clear that the workaround was only meant to be temporary until a fix such as this is available.

eoineoineoin:
> This should really check if the control is possibly involved in a child selector, and only do the recalculation in that case.

I agree with the sentiment - in [2e1c831](https://github.com/space-wizards/RobustToolbox/pull/6264/commits/2e1c831a55f0453298d6b80f27d91273bf6d6255), I added this; doesn't _feel_ great, but would also worry that doing something more sophisticated would add a _lot_ of complexity.

> Button.StylePropertiesChanged() contains a workaround for this; it manually calls Restyle() on one of its children. It still doesn't check whether the calculation is actually needed, but it's applied only to a case where it is often needed.

Nice spot! I've removed the workaround since whatever solution we land on should obsolete that. The only other thing in Content (right now, at least) which uses parent selectors is hud tooltips, which wouldn't ever have their styles changed after construction.

Absotively:
I think I have worked out how to work around the lack of this for switch buttons; I can apply relevant pseudoclasses to the buttons' children and change the style rules accordingly. I'm going to put off actually implementing that for the moment, though, in hopes that this gets merged and saves me from having to.

Absotively:
> I think I have worked out how to work around the lack of this for switch buttons; I can apply relevant pseudoclasses to the buttons' children and change the style rules accordingly. I'm going to put off actually implementing that for the moment, though, in hopes that this gets merged and saves me from having to.

I have implemented a workaround in switch buttons. It ended up being a bit messier than I'd hoped, so I'm definitely still hoping this change will some day remove the need for workarounds.

Robust.Client/UserInterface/Controls/Button.cs
```diff
@@ -24,11 +24,6 @@ public Button()
         protected override void StylePropertiesChanged()
```
PJB3005: Remove this overload.

PJB3005 (changes requested):

Robust.Client/UserInterface/Stylesheet.cs
```diff
@@ -314,7 +314,24 @@ public StyleProperty(string name, object value)
...
         public abstract bool Matches(Control control);
+
+        /// <summary>
+        ///     Returns true if this selector might apply to children of this
+        ///     control. This should assume variable properties such as style
+        ///     classes or pseudo classes match, and only return false if this
+        ///     selector does not match invariable properties such as the element
+        ///     type.
+        /// </summary>
+        public abstract bool PotentialChildMatch(Control control);
```
PJB3005, line 329: Why should this only check invariable properties? Wouldn't it get re-checked if stuff like style classes change anyways? The purpose of this check should be to check whether this control is the parent of a child selector.

PJB3005 (changes requested):

Robust.Client/UserInterface/Stylesheet.cs
eoineoineoin: > The purpose of this check should be to check whether this control is the parent of a child selector.

Right; that's exactly what this is doing.

It needs to assume that any variable properties like styleclasses or pseudo classes do match because they can change in the future - the reason it's only checking invariable properties is _because_ those can't change. As an example, suppose we have a style rule:

`E<Button>.Pseudo("hovered").ParentOf<BoxContainer>().ParentOf<Label>().Prop(...)`

And we have a control hierarchy like:

* Button
  * BoxContainer
    * Label

The style rule is applying some properties to the `Label` when the `Button` is hovered, so to propagate that down, the `Button` needs to restyle it's children when it's style/pseudo classes change.

However, if we had a control hierarchy like:

* Button
  * CenterContainer
    * Label

That style rule would never apply - the `CenterContainer` can never be changed to the `BoxContainer`, so changing any style/pseudo classes on the `Button` *won't* attempt to restyle it's children.

That's what it means to only check invariable properties here.

PR #6267, opened 2025-10-24
Fix word wrapping on non-runed inline metrics like links

Jewelots:
## About the PR
Previously non-runed inline metrics such as links would never wrap if they went partially offscreen, leading to some ugly cutoff

## Technical details
TextLinkTags were only ever using NextMetrics which wasn't inserting any breakLine's after a metric went over the bounds like ProcessRune does. I also had to modify RichTextEntry to avoid some edge cases where it would accidentally display a space as the first character on a newline after processing a non-runed metric.

I'm not super happy with the RichTextEntry changes but the entire Rich Text system seems to be barely held together and in need of a large overhaul (there's some other problem areas in the way rich text is rendered that are unchanged by my fix, you can see a period completely intersecting a "Click here to see a list of technologies" link in the "Science" job entry), so I'd rather a messy fix than none at all. If anyone can do it in a cleaner way then feel free to touch it up.

## Media
Static showcase before fix
image

Static showcase after fix
image

Animated showcase before fix
link_wrapping_broken

Animated showcase after fix
link_wrapping_fixed

## Requirements
- [X] I have added media to this PR or it does not require an ingame showcase.

### DIFF
--- a/Robust.Client/UserInterface/RichTextEntry.cs
+++ b/Robust.Client/UserInterface/RichTextEntry.cs
@@ -219,6 +219,8 @@ namespace Robust.Client.UserInterface
             var baseLine = drawBox.TopLeft + new Vector2(0, defaultFont.GetAscent(uiScale) + verticalOffset);
             var controlYAdvance = 0f;
 
+            var spaceRune = new Rune(' ');
+
             var nodeIndex = -1;
             foreach (var node in Message)
             {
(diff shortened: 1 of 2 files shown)


lzk228:
i remember there was an issue for that, could you search for it?

lzk228:
well ok then

PR #6268, opened 2025-10-25
Allow content to override ProcessStream and GetOcclusion in AudioSystem

EmoGarbage404:
On my server, we want the ability to be able to handle audio based on some more particular content-specific elements and decisions. However, this is largely impossible due to the audio processing being largely relegated to the engine. Rather than trying to create a web of CVAR options for different audio processes in the engine, I think it's better to just expose more parts of audio to content.

This PR adds the ability for content to define custom handlers for the `ProcessStream` and `GetOcclusion` methods (the two main areas where audio values are determined) that are used in place of the typical function behavior. This is done with static C# events which cause early returns in their corresponding functions. In a more traditional setup these would simply be virtual method calls, but I think this is a fair-enough compromise to achieve that same behavior in the given framework.

Two small implementation notes:
- The `ProcessStream` function is entirely overridden, despite some of the top-most behavior arguably being something that should stay the same no matter what. I think it's better, though, to allow people to have the specificity at the cost of (at worst) minor duplication being needed at the content-level to preserve that behavior.
- Both events are limited to a single invocation target via the use of debug asserts. This obviously makes sense for the `GetOcclusion` function but is more dubious for `ProcessStream`. I do think it's worthwhile considering what the expected use case of this is, however, as I don't really see why someone would want to split up the work of audio processing across multiple functions and I think preventing someone from processing the audio twice is a more likely issue that would want to be avoided.

### DIFF
--- a/Robust.Client/Audio/AudioSystem.cs
+++ b/Robust.Client/Audio/AudioSystem.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Numerics;
+using JetBrains.Annotations;
 using OpenTK.Audio.OpenAL;
 using Robust.Client.GameObjects;
 using Robust.Client.Graphics;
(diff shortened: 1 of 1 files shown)


Robust.Client/Audio/AudioSystem.cs
PJB3005, line 1: These should not be static.

PJB3005 (changes requested):

PJB3005 (approved):

PR #6269, opened 2025-10-25
Fix text links in the guidebook being clickable if the cursor is anywhere to the right of the link, outside it's bounds

Jewelots:
This was originally [a pull request for the main repo](https://github.com/space-wizards/space-station-14/pull/41076) but the better fix happened to be within RobustToolbox instead.

## About the PR
Text links in the guidebook had a bug where they were were clickable/hoverable when the mouse wasn't anywhere close to them (anywhere to the right of the link).

## Technical details
TextLinkTags in the guidebook were set to the default "stretch" horizontal layout, so during layout they were all being stretched to the right side of the panel they're in. Despite not causing any visual changes, this made their clickbox huge. More intricate technical details to follow in the comment below (copypasted from my writeup in the original issue).

## Media
bad_links

## Requirements
- [X] I have added media to this PR or it does not require an ingame showcase.

### DIFF
--- a/Robust.Client/UserInterface/RichTextEntry.cs
+++ b/Robust.Client/UserInterface/RichTextEntry.cs
@@ -66,6 +66,10 @@ namespace Robust.Client.UserInterface
                 // store state information and return the same control for each rich text entry.
                 DebugTools.Assert(handler.TryCreateControl(node, out var other) && other != control);
 
+                // Avoid children stretching to fill space after manual layout
+                control.HorizontalAlignment = Control.HAlignment.Left;
+                control.VerticalAlignment = Control.VAlignment.Top;
+
                 parent.Children.Add(control);
                 tagControls ??= new Dictionary<int, Control>();
                 tagControls.Add(nodeIndex, control);


Jewelots:
Details of what was wrong/why the fix exists copied from [my original post](https://github.com/space-wizards/space-station-14/pull/41076#issuecomment-3445905262)

> All rich text is inside a `Document`, which is basically just a BoxContainer.
> 
> On Arrange, the `Document`'s children are laid out vertically, and each of it's children is essentially each "paragraph" of the document. The first title, a seperator bar, the first paragraph, a block of images, etc etc. All things that are vertically stacked and don't interact with eachother at all. These are basically all `Label` (text), `RichTextLabel` (text, colored text, links) and `Box` (images, other things).
> 
> The kicker is `RichTextLabel`, which is essentially an entire paragraph. Here's what one `RichTextLabel` contains, in the engineering Document:
> 
> > [color=white][color=#F39F27FF]Engineering[/color] is a combination of [textlink="construction"link="Construction"] work, [textlink="repair"link="ExpandingRepairingStation"] work, maintaining a death machine that happens to produce [textlink="power"link="Power"], and making sure the station contains [textlink="breathable air"link="Atmospherics"].[/color]
> 
> The `RichTextLabel` has a few children (only the links) and then all the rest of the text is bespoke manual rendering with a Draw function (via `RichTextEntry`). This iterates over the nodes of the message, changing colors, drawing individual font characters, and then, importantly, taking any of the embedded `Control`s and laying them out where the next character would be written in the text, offsetting the advance and linebreaks and whatnot while doing so. This manually positions the control where it's supposed to be, and the control is sized to the minimum bounds that fit the containing elements with a Measure.
> 
> Because it's all laid out manually, that's the last time it ever gets positioned or resized, ideally. However, during arrange, the `Document` calls `ArrangeCore` on the `RichTextLabel`, which in turn calls the default `ArrangeOverride` for a `Control`, which in turn loops over every child and calls `ArrangeCore` on those, which sees they're marked as Stretch and stretches them to the full size they can accomodate. There's absolutely no checking for collisions, boundaries, other elements blocking the stretching, etc. It's all completely naive because those elements can't possibly know about all the manually rendered font characters and etc, all they know is where they were positioned, and now they're being told to stretch, so they stretch.
> 
> So, finally, knowing all that, you can do one of two fixes:
> 
> - Manually set the children inside a RichTextLabel to not stretch, so they keep their desiredSize.
> - Mark the children as "already laid out" and manually set the size based on desiredSize.
> 
> I'm not really comfortable pretending it's already laid out and would rather it go through the normal channel, so marking them as not stretch seems like the most sane fix to me. But instead of doing it on TextLinkTag, I'm going to set all children of a `RichTextEntry` to be that way instead, because the root of the issue is the `RichTextEntry` laying out its children manually, rather than the `TextLink` itself being problematic.

This fix of telling the chilren inside the `RichTextEntry` not to stretch seems to work exactly how I'd expect, and as far as I can tell, this hasn't caused any other issues to crop up elsewhere, as nothing was actually relying on this behavior.

Robust.Client/UserInterface/RichTextEntry.cs
```diff
@@ -66,6 +66,10 @@ public RichTextEntry(FormattedMessage message, Control parent, MarkupTagManager
                 // store state information and return the same control for each rich text entry.
                 DebugTools.Assert(handler.TryCreateControl(node, out var other) && other != control);
 
+                // Avoid children stretching to fill space after manual layout
+                control.HorizontalAlignment = Control.HAlignment.Left;
+                control.VerticalAlignment = Control.VAlignment.Top;
```
PJB3005, line 71: This is not a correct fix. The layout code should not be giving these children all that space in the first place.

PJB3005 (changes requested):

Robust.Client/UserInterface/RichTextEntry.cs
Jewelots: > This is not a correct fix. The layout code should not be giving these children all that space in the first place.

I wrote up the writeup Specifically to explain that, yes, the layout _should_ be giving them all that space if you ask it to stretch; the layout has no knowledge of anything else that's being rendered to the RichTextLabel because the text doesn't exist, it's purely being drawn, there's no Controls associated with it, nor does the layout know that any space is taken up by it at all, nor is there any way to Tell the layout that there was space taken up by it, nor does the layout code ever check for collisions between controls in any other part of the entire codebase.

If you feel it's best to rewrite the entire rich text And layout system to support this because you feel it should be done in a different way then absolutely feel free, but that's completely outside the scope of this fix.

PR #6270, opened 2025-10-26
added getItemText

Samuka-C:
Added a simple method to get the Item Text from a OptionButton


PJB3005:
This is what the metadata field is for.

PR #6278, opened 2025-10-30
Add Account Creation time from auth to the userdata

VasilisThePikachu:


Robust.Shared/Network/NetUserData.cs
```diff
@@ -18,6 +19,9 @@ public sealed record NetUserData
         [ViewVariables]
         public string? PatronTier { get; init; }
 
+        [ViewVariables]
+        public DateTime? CreatedTime {  get; init; }
```
PJB3005, line 23: In what scenario is this null?

PJB3005 (changes requested):

Robust.Shared/Network/NetUserData.cs
VasilisThePikachu: Unauthenticated accounts in my testing

PJB3005: Ah, right. Makes sense.

PJB3005 (approved):

PR #6284, opened 2025-11-01
Fix PrototypedCommand exception

### LINKED ISSUE #6283: Toolshed "delete" command has a problem with seemingly unrelated entities

_No response_

### Description

`do "delete $ID"` seems to have issues with entities that have... subordinate? entities, and will error out after the first deletion almost every time because a (presumably) subordinate entity gets its MetaDataComponent taken away or otherwise misplaced, which toolshed does not seem to like.

I will note I am making a great amount of estimation here. I'm not smart enough to figure out what is actually happening.

### Reproduction

-2. (Control 1) spawn a group of entities that aren't... "complex" (for example, TableFancyOrange) and then do `> entities prototyped TableFancyOrange do "delete $ID"`. This should go off without a problem.
Image

-1. (Control 2) use the command `delete entityID` on a "complex" entity directly. This should also occur without issue.
Image

1. Spawn a group of entities that are more complex, such as a gun or a mob. These should all be the same entity. I recommend doing `> entities prototyped Prototype` and making note of the entity IDs that are returned.
Image

2. Attempt to do `> entities prototyped Prototype do "delete $ID"`. It will delete precisely one of these entities, seemingly picked at random, and then it will error out, citing an entity that was not listed in the earlier `> entities prototyped Prototype` return.

Image

Image

(pardon the two screenshots for the same thing, I dunno how to increase the size of the console, if it can even be done, and would like to include the full stack)

### Additional Context

This is a potential, albeit minor, problem for administration regarding entity spam (and the deletion thereof).

ElectroJr:
Fixes #6283

Completely unrelated, but this PR also adds some `ViewVariables` attributes to some map system fields because it helps with debugging stuff and I CBF creating a separate PR for that.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -59,6 +59,7 @@ END TEMPLATE-->
 * Fix modifying `Label.FontOverride` not causing a layout update.
 * Controls created by rich-text tags now get arranged to a proper size.
 * Fix `OutputPanel` scrollbar breaking if a style update changes the font size.
+* Fix `PrototypedCommand` toolshed command throwing exceptions when encountering deleted entities.
 
 ### Other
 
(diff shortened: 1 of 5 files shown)


PJB3005 (changes requested):
Shouldn't `entities` just not be returning deleted entities?

PR #6285, opened 2025-11-02
Move SnapToGrid to engine

### LINKED ISSUE #6274: SnapgridHelper.SnapToGrid errors when used within ParallelManager
SnapgridHelper.SnapToGrid(EntityCoordinates coordinates, IEntityManager entMan, IMapManager mapManager) specifically
Resolves two dependencies which even if you pass them in it will still cause an error
See also https://github.com/space-wizards/RobustToolbox/issues/6273

ElectroJr:
This adds content's `SnapToGrid` extension methods to `SharedTransformSystem`.

This PR also changes the static `IoCManager.Resolve` methods so that they don't throw an NRE in uninitialized threads even when you pass non-null values (fixes #6274)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -51,6 +51,7 @@ END TEMPLATE-->
   * At the moment only available on Windows.
   * See `ISystemFontManager` for API.
 * The client now display a loading screen during startup.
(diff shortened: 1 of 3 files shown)


PJB3005:
What is the use case for this? This kind of feels like a high-level helper function and I'm not sure it's appropriate to have in engine.

PR #6287, opened 2025-11-03
Add subscription to cvars change by client

Alwayswannahunt:
# This PR
Adds subscription with same structure as `OnValueChanged` on same build (server -> sever or client -> client) but for client -> server. Also adds smooth implementation for shared code. 
Highly inspired by `IConfigurationManager` code, so I cant be sure about doing things right.

# Why
Some interaction prediction needs it for better UX. Also this goes as "event" way of getting client cvar without every time calling `GetClientCVar`. Usage example: 

https://github.com/user-attachments/assets/0a021665-f121-4fb5-ad03-281e54f0f792

# TODOs
- [x] Working 
- [x] Event on disconnect
- [x] Tests


Robust.Shared/Configuration/NetConfigurationManager.cs
PJB3005, line 1: What is the point of these `onDisconnect` parameters? Can't consumers just use `IPlayerManager.PlayerStatusChanged` for that themselves?

PJB3005, line 1: I would like to see overloads that have `CVarChangeInfo` available.

Robust.Server/Configuration/ServerNetConfigurationManager.cs
PJB3005, line 1: You should probably check in these subscription managers that the CVar is actually `CLIENT | REPLICATED`.

PJB3005 (changes requested):

Robust.Shared/Configuration/NetConfigurationManager.cs
Alwayswannahunt: Primarily to put everything in one place. It can be used to store client preference and if you do, also need to clear collection on disconnect.
I agree with you, it seems redundant, so I don't mind of removing it. I'll wait your reply after this clarification before doing so.

Alwayswannahunt: Can you show what method overload you want or example in codebase? Not sure that I can understand correct what needed

PJB3005: There are versions of `OnValueChanged` that give the callback a `CVarChangeInfo` instead of just the CVar value. That is what I mean.

PJB3005: Yes, please remove it.

Robust.Client/Configuration/ClientNetConfigurationManager.cs
```diff
@@ -136,4 +139,51 @@ private void ApplyClientNetVarChange(List<(string name, object value)> networked
...
+    /// <inheritdoc />
+    public override void UnsubClientCVarChanges<T>(string name, ClientCVarChanged<T> onChanged)
+    {
+        if (_player.LocalSession is not { } localSession)
+        {
+            Sawmill.Error("Got null local session for client!");
+            return;
+        }
+
+        UnsubValueChanged<T>(name, (T newValue, in CVarChangeInfo info) => onChanged(localSession, newValue, in info));
```
Alwayswannahunt, line 187: Idk if this will work. I'll write down in this thread after testing

metalgearsloth (changes requested):
Fix merge conflicts please.

PR #6290, opened 2025-11-09
Nuke collisionwake + change fixture storage

metalgearsloth:
Fixture storage is aligned more with box2d I just want to piecemeal the changes a bit in case stuff blows up.

As for CollisionWake I don't think it's necessary anymore especially with the recent broadphase change and it causes headaches on content having to deal with handling it. When I originally added it the game would get like 8fps on my system and we desperately needed it but nowadays I still get a crispy 200fps walking around so.

BC entry:
- CollisionWakeComponent has been nuked as broadphase performance is sufficient as to not require it anymore.
- Sundries tree has been nuked and now all physics bodies are stored on the dynamic / kinematic / static trees regardless of whether they have collision. StaticSundries will contain everything that has no physics (as it did before).
- Sundries flag will now query for the StaticSundries tree and StaticSundries will do all static entities (i.e. Sundries + Static physics bodies). This was the least breaking change of the 2 flags.
- There is a new EntityLookupFlag Enabled to only return collidable physics bodies.
- Entities being moved to and from nullspace don't have CanCollide altered as it's no longer relevant.

Porting:
- Remove CollisionWake and possibly set canCollide to true for your physics bodies (which is also the default).
- For prototypes you will likely just want to delete all references to canCollide: false unless you have reason for it, however maps are likely set to that for contained entities so you will want to verify those before touching them.
- EntityLookupFlags delete Sundries from your queries. If you had a query with only sundries bodies then this will need to be checked manually via CanCollide on the returned entities.

Internal:
- Fixture pairs are now stored in a central hashset.


metalgearsloth:
Cooked this branch and made a new one as it was easier.

PR #6292, opened 2025-11-10
Replace OnValueChanged with SubValueChanged

sleepyyapril:
UnsubValueChanged was added later, but their names should still be consistent with each other and use the same naming scheme to prevent confusion.

This does add a lot of warnings, but I'm willing to handle them myself.


PJB3005:
This is not worth the breaking change.

PR #6294, opened 2025-11-12
Make GetRegistration a generic

metalgearsloth:
Small pr from arch.


slarticodefast:
Out of interest, what is the advantage of doing it this way?

metalgearsloth:
> Out of interest, what is the advantage of doing it this way?

Turns out we can use the CompIdx index here which uses the array lookup instead of the frozendictionary lookup.

image

The interface version was slightly faster than generic so would've been fine for now but the additional change made it better for all cases.

PR #6309, opened 2025-11-25
Cleanup warnings: CS0162, CS0618, CS0649

B-Kirill:
This PR removes 2 small and unrelated warnings

1x [CS0162](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0162): Unreachable code detected.
`Robust.Shared/GameObjects/EntityManager.cs`

1x [CS0649](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0649): Field 'EntityEventBus.EventData.ComponentEvent' is never assigned to, and will always have its default value 'false'.
`Robust.Shared/GameObjects/EntityEventBus.Common.cs`

### DIFF
--- a/Robust.Client/UserInterface/RichText/MarkupDrawingContext.cs
+++ b/Robust.Client/UserInterface/RichText/MarkupDrawingContext.cs
@@ -8,20 +8,20 @@ public sealed class MarkupDrawingContext
 {
     public readonly Stack<Color> Color;
     public readonly Stack<Font> Font;
-    public readonly List<IMarkupTag> Tags;
+    public readonly List<IMarkupTagHandler> Tags;
 
     public MarkupDrawingContext()
     {
         Color = new Stack<Color>();
         Font = new Stack<Font>();
-        Tags = new List<IMarkupTag>();
+        Tags = new List<IMarkupTagHandler>();
     }
 
     public MarkupDrawingContext(int capacity)
     {
         Color = new Stack<Color>(capacity);
         Font = new Stack<Font>(capacity);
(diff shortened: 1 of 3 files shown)


Robust.Shared/GameObjects/EntityEventBus.Common.cs
```diff
@@ -97,7 +97,7 @@ private sealed class EventData
         /// <summary>
         /// <see cref="ComponentEventAttribute"/> set?
         /// </summary>
-        public bool ComponentEvent;
```
PJB3005, line 100: This is not an appropriate fix for this warning.

Robust.Client/UserInterface/RichText/MarkupDrawingContext.cs
PJB3005, line 1: This is a breaking change. It's probably better to wait for this type to be removed entirely instead.

PJB3005 (changes requested):

Robust.Shared/GameObjects/EntityEventBus.Common.cs
B-Kirill: > This is not an appropriate fix for this warning.

Yeah, it was worth a closer look..

This bool was added by you in https://github.com/space-wizards/RobustToolbox/pull/2911. However, it is still not used anywhere where it could be (considering that `EventData` is `private sealed class`). So it's worth deleting, I guess. It can be added back if really needed.

PJB3005 (approved):

PR #6311, opened 2025-11-26
Add BUI helper overloads supporting constructor parameters

B-Kirill:
This PR adds BUI helper overloads with support for passing constructor parameters.  It also makes a few stylistic changes.

Now there is a problem that the old method of creating a window (example: `_window = new IdCardConsoleWindow(this, _prototypeManager, accessLevels)`) supports passing constructor parameters, but helpers do not. It's very inconvenient.. so this PR fixes it.

As an example, this allows you to pass `IDependencyCollection` as a parameter to replace `IoCManager.InjectDependencies(this)` with `dependencies.InjectDependencies(this)`.

All new overloads have been tested on SS14 content and everything is working as intended.

### DIFF
--- a/Robust.Client/UserInterface/BoundUserInterfaceExt.cs
+++ b/Robust.Client/UserInterface/BoundUserInterfaceExt.cs
@@ -7,12 +7,33 @@ namespace Robust.Client.UserInterface;
 
 public static class BoundUserInterfaceExt
 {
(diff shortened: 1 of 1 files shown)


B-Kirill:
This change may not be necessary anymore, given that another PR has resolved the dependency injection issue in BUI.
I will check later to see if this PR should be closed.

PR #6312, opened 2025-11-26
Add noSpawn property to audio entity

Peptide90:
Stops audio entity showing in spawn menu.

This was a very simple thing that I noticed when wiping the spawn menu.

### DIFF
--- a/Resources/EnginePrototypes/Audio/audio_entities.yml
+++ b/Resources/EnginePrototypes/Audio/audio_entities.yml
@@ -3,7 +3,7 @@
   name: Audio
   description: Audio entity used by engine
   save: false # TODO PERSISTENCE what about looping or long sounds?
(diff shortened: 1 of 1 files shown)


slarticodefast (changes requested):
NoSpawn has been obsoleted over a year ago #5364
Use `categories: [ HideSpawnMenu ]`.
Also weird that this does not cause a linter error.

Peptide90:
> NoSpawn has been obsoleted over a year ago #5364 Use `categories: [ HideSpawnMenu ]`. Also weird that this does not cause a linter error.

I saw usage of both so thought it was fine to use that for single files and the category for multiple files, although the category isn't inherited which caused a bit of repetiton on my end. Will fix.

slarticodefast (approved):
Looks good!

PR #6320, opened 2025-11-29
make EntityQuery.Resolve error not useless

deltanedas:
use ToPrettyString instead of failing tests etc with a meaningless uid you have no way to get any kind of context for

EntityManager replaces the ISawmill since the sawmill is just stored on it anyway

# Breaking Change

EntityQueryEnumerator constructor is no longer public


Robust.Shared/GameObjects/EntityManager.Components.cs
```diff
@@ -1862,7 +1863,8 @@ public bool Resolve(EntityUid uid, [NotNullWhen(true)] ref TComp1? component, bo
 
             if (logMissing)
             {
-                _sawmill.Error($"Can't resolve \"{typeof(TComp1)}\" on entity {uid}!\n{Environment.StackTrace}");
+                var entMan = IoCManager.Resolve<IEntityManager>();
```
PJB3005: This is not an appropriate way to get the entity manager.

PJB3005 (changes requested):

Robust.Shared/GameObjects/EntityManager.Components.cs
deltanedas: as said in the pr this is a very cold path and i dont think its worth putting entitymanager in every query

PJB3005: Make it only in there on `TOOLS` or something then. This is not acceptable.

deltanedas: use ioc for release??? or not use toprettystring at all for release which just defeats the purpose of having usable error logs

PJB3005: > or not use toprettystring at all for release which just defeats the purpose of having usable error logs

> use ToPrettyString instead of failing tests

You stated your use case.

deltanedas: failing tests is fine, its just that only logging the uid means the error is useless and you have to go adding debug prints to find out what the resolve is failing on

PJB3005: Okay? That doesn't change the fact that `IoCManager` is literally not usable here? This shit needs to work from other threads.

PJB3005 (commented):
I just realized this has a public constructor, so this change is a breaking change 💀 

Could you fix it to avoid that with some obsolete tags?

deltanedas:
isnt it a straight up runtime error to do `new EntityQueryEnumerator`, wouldnt it be good to make it internal?

PJB3005:
> wouldnt it be good to make it internal?

Yes, and that would be a breaking change.

deltanedas:
yes if its breaking might as well remove it from the api entirely

PJB3005 (approved):

PR #6325, opened 2025-12-04
Fixed looped audio playback position calculation

ThereDrD0:
I encountered an odd bug: if I go far away from a looped audio source and return, the audio stops playing on the client. Disconnecting and reconnecting restores playback.

I figured out that it was a bug with the playback position calculation. It didn't account for the Loop parameter correctly and stopped playing because the position exceeded the audio length.

Here is a fix. + clean up redundant code and unused usings

Tests

image

# Example

Before(master)

https://github.com/user-attachments/assets/1231d32f-5c42-45a6-884f-5089d1b5d953

After(my branch)

https://github.com/user-attachments/assets/87ba9525-75dc-45e4-b224-956b5b0a9521

Example code used to reproduce bug

```csharp
using Robust.Server.Audio;

namespace Content.Server.AAA;

public sealed class TestSoundSystem : EntitySystem
{
    [Dependency] private readonly AudioSystem _audio = default!;

    {
        base.Initialize();

        SubscribeLocalEvent<TestSoundComponent, MapInitEvent>(OnMapInit);
    }

    private void OnMapInit(Entity<TestSoundComponent> ent, ref MapInitEvent args)
    {
        _audio.PlayPvs(ent.Comp.Sound, ent);
    }
}
```

```csharp
using Robust.Shared.Audio;

namespace Content.Server.AAA;

[RegisterComponent]
public sealed partial class TestSoundComponent : Component
{
    [DataField]
    public SoundSpecifier Sound = new SoundPathSpecifier("/Audio/Effects/glass_break1.ogg",
        AudioParams.Default.WithMaxDistance(30f).WithVolume(10f).WithLoop(true));
}
```


PJB3005 (changes requested):
Do not copy paste this code around everywhere

ThereDrD0:
Hey, just wondering if this is good to go?

PJB3005 (approved):

PR #6326, opened 2025-12-04
Fixed a typo in LocalizationManager

GlupiyBrush:
In _ILocalizationManager_ and _LocalizationManager_:
_SetFallbackCluture_ -> _SetFallbackCulture_
_SetFallbackCluture_ now obsolete

That's all.


PJB3005 (changes requested):
As this is a public API, you would need to add a compatibility override to avoid breaking existing code.

Also, I'm definitely going to put this PR on hold to avoid conflicts with #6219

PR #6327, opened 2025-12-05
fix(tools): Fix packaging libs and zips on linux.

CaasGit:
Both OSX and Linux builds were not excluding two libs from their zip file due to a missing comma. You can see that they both exist within the live client manifests.

Also fix the MS dlls having basically no date set on the file which is throwing an error when trying to zip them.


PJB3005:
Guh. Good catch.

PJB3005 (approved):

PR #6333, opened 2025-12-09
Add some default Sanitization to RichTextEntry

Princess-Cheeseballs:
This has been causing a bit of trouble upstream and Julian asked someone to just give the RichText a default set of tags so that's what this PR does. 

This makes having no tags an opt in rather than an opt out. 

Also did a bit of cleanup. Nothing breaking though.

### DIFF
--- a/Robust.Client/UserInterface/RichTextEntry.cs
+++ b/Robust.Client/UserInterface/RichTextEntry.cs
@@ -9,317 +9,316 @@ using Robust.Shared.Collections;
 using Robust.Shared.Maths;
 using Robust.Shared.Utility;
 
-namespace Robust.Client.UserInterface
+namespace Robust.Client.UserInterface;
+
+/// <summary>
+///     Used by <see cref="OutputPanel"/> and <see cref="RichTextLabel"/> to handle rich text layout.
+///     Note that if this text is ever removed or modified without removing the owning control,
+///     then <see cref="RemoveControls"/> should be called to ensure that any controls that were added by this
+///     entry are also removed.
+/// </summary>
+internal struct RichTextEntry
 {
+    public static readonly Type[] DefaultTags =
+    [
+        typeof(BoldItalicTag),
+        typeof(BoldTag),
+        typeof(BulletTag),
+        typeof(ColorTag),
+        typeof(HeadingTag),
+        typeof(ItalicTag)
+    ];
+
+    private readonly Color _defaultColor;
+    private readonly Type[]? _tagsAllowed = DefaultTags;
+
+    public readonly FormattedMessage Message;
+
     /// <summary>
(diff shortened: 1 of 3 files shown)


PJB3005 (changes requested):
Do not reformat the entire file

Princess-Cheeseballs:
> Do not reformat the entire file

Done

ruddygreat:
would make the PR way bigger, but wouldn't it be better to add a blacklist to RichTextEntry & MarkupTagManager methods, then just stick commandLink in that blacklist by default?

Robust.Client/UserInterface/RichTextEntry.cs
```diff
@@ -41,6 +51,20 @@ internal struct RichTextEntry
...
+            Message = message;
+            Height = 0;
+            Width = 0;
+            LineBreaks = default;
+            _defaultColor = defaultColor ?? new(200, 200, 200);
+            Controls = GetControls(parent, tagManager);
+
+            // Default sane tags
+            _tagsAllowed = DefaultTags;
+        }
```
PJB3005: Reuse the constructor, do not copy paste it.

PJB3005 (changes requested):
Do not reformat the entire file

Robust.Client/UserInterface/RichTextEntry.cs
```diff
@@ -19,8 +19,18 @@ namespace Robust.Client.UserInterface
...
+            typeof(BoldTag),
+            typeof(BulletTag),
+            typeof(ColorTag),
+            typeof(HeadingTag),
+            typeof(ItalicTag)
+        ];
+
         private readonly Color _defaultColor;
-        private readonly Type[]? _tagsAllowed;
+        private readonly Type[]? _tagsAllowed = DefaultTags;
```
PJB3005: Why does this need to be default initialized here?

PJB3005 (changes requested):
Isn't this PR not enough to actually apply these limits to either consumer of `RichTextEntry`?

Robust.Client/UserInterface/RichTextEntry.cs
Princess-Cheeseballs: Doesn't need to be I suppose

Princess-Cheeseballs:
> Isn't this PR not enough to actually apply these limits to either consumer of `RichTextEntry`?

Nope, I updated that just now though. I checked and nothing in content actually used the constructor directly instead using an API method in RichTextLabel or OutputPanel to create a RichTextEntry. Updated their API methods so that if they don't define their tags it'll use the default tags provided.

PJB3005:
> I checked and nothing in content actually used the constructor directly

Yeah, because it's an internal type.

PJB3005 (approved):

Princess-Cheeseballs:
> Yeah, because it's an internal type.

Top 10 reasons to never code on an empty stomach.

PR #6337, opened 2025-12-11
Add and update a lot of documentation

### LINKED ISSUE #6073: DataDefinitionAttribute needs documentation
That thing is a total mystery with not a single code comment or mention in the docs, so contributors have no clue when to use it.
Does it need to be to be applied to all classes and structs that are used as datafields? And when is `Serializable` enough? Are there cases where a `DataDefinition` does not need to be `Serializable`?

Same for `MeansDataDefinition`.

slarticodefast:
All credit goes to @moonheart08 who kindly took time to write this all.
I just made the PR.

Resolves #6073


slarticodefast:
Test fails seem to also affect other PRs.

Robust.Shared/GameObjects/Component.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using Robust.Shared.GameStates;
```
Pok27, line 2: ~~no need~~
need

PR #6341, opened 2025-12-17
Add sprite layer render order override and parent/child layer rendering

SlamBamActionman:
This PR adds two main features to `SpriteComponent`:

### Sprite layer rendering order override

By default, a sprite renders the layers in the order they are in the `List<Layer> Layers` property.

There is now a new property added `Dictionary<RsiDirection, List<int>>? LayersOrderOverride`. Its contents are meant to represent indexes of layers in the main `Layers` list for a given direction. When `LayersOrderOverride` is set, instead of rendering the layers in the order of the `Layers` list, the sprite layers are rendered in the order their indexes are listed in the override. E.g. `2, 0, 1` will render the layers in that order instead of `0, 1, 2`.

Of note is that this order *is only followed up until the highest index in the override*, and then it continues rendering any layers indexed after that. This means if an override's highest index is e.g. `2`, any layer in the `Layers` list with an index of 3 or higher will *also* render after the override (in the order listed in the `Layers` list). This is because these types of layers are often intended added during the runtime to be rendered on top of the existing sprite and may be difficult to account for in the override. 

The `LayersOrderOverride` does *not* update automatically based on changes in the `Layers` list order. It is expected that whatever function sets it also accounts for changes throughout the codebase, updating it if new layers are added/removed.
<sup>- Is this reasonable? Arguably one could also update the property if a layer is added/removed, but that might be a lot of overhead work. Let me know if we want this instead of updating via Content.</sup>

### Parent/Child layers

A child layer is a layer that will always follow its parent in the layer order, regardless of its index (override or not). This is set in the child via the property `int? ParentLayer` and in the parent via `List<int> ChildLayers`.

These properties are inaccessible outside the API to (hopefully) avoid accidental endless recursion. When a layer is added, it can optionally include a `parentIndex` which refers to the index of a layer in the `Layers` list. The parent layer gets the new layer added as a child, and when processing the rendering order any child layer gets selected for rendering after the parent layer has been rendered.

Adding/removing layers will shift the parent/child indexes accordingly, similar to how the `Index` property is shifted. If a layer with children is removed, any children are attached to the removed layer's parent (if there is one).

As a parent can have multiple children, each child will render in the order they are added to the parent. The function runs recursively, so a child can also be a parent. 

There is also the `SetToParent` method which assigns an existing layer as a parent to another existing layer, using the indexes. Of note is that if the child already has a parent, the new parent is given the child's old parent as its parent. This means `SetToParent` can be used to insert a layer in the middle of a chain of parent-children relationship.
<sup>- This is necessary for displacement maps, as they require the layer sprite they're modifying to exist when being created, but has to be before said layer in the rendering order.</sup>
<sup>- SetToParent is potentially risky as I think it would be possible to create a Parent/Child recursive loop. There is a debug assert that should be able to catch this. I am not sure how to fool-proof this otherwise however.</sup>

---

### Issues

Sprite rendering is an area that is meant for heavy optimization, and that is unfortunately not my forté. I've done my best to preserve what I hope are performant methods but I there are areas where I'm uncertain if it has been made worse. 

**LayerRenderingStrategy**

My understanding is that generating the matrices is a bit of a heavy operation. Previously this was easily avoided since the layers rendered in a simple foreach loop, but now that the behavior is more complex I needed to find a way to pass these on to the actual `RenderLayer` function. I've opted to, in case `GranularLayersRendering` is false, include the calculated matrices in a Dictionary and then retrieve them later on when `RenderLayer` is called. I'm not sure the dict is performant here though.

**Parent/Child architecture**

I think there is a case one could make that defining either the parent or the child layer isn't necessary, or that making a parent able to have multiple children is a bad idea. It seems to me that including both allows for the most flexibility but they also aren't hard-coupled and there could theoretically be a situation where there's recursion or indexes point incorrectly. Let me know if this is wrong.

**LayersOrderOverride doesn't update**

As mentioned earlier, adding/removing layers doesn't update the override. In my implementation I update these in Content so that isn't an issue, but maybe they should here as well, similar to how Parent/Children/Index indexes are?

---

### Why?

Being able to override the sprite layer order means we can make much nicer-looking content via directional sprites. There will be a Content PR up soon after this one that attempts this using the features here. The override ensures "bookmark" layers can layer sprites in the correct order depending on which direction a character is facing, while a parent/child relationship allows there to be further customization of the rendering order by not just moving the "bookmark" layers.


Robust.Client/GameObjects/EntitySystems/SpriteSystem.Render.cs
```diff
@@ -60,55 +61,127 @@ public void RenderSprite(
         var entityMatrix = Matrix3Helpers.CreateTransform(worldPosition, sprite.Comp.NoRotation ? -eyeRotation : worldRotation - cardinal);
         var spriteMatrix = Matrix3x2.Multiply(sprite.Comp.LocalMatrix, entityMatrix);
 
+        Dictionary<LayerRenderingStrategy, Matrix3x2> renderingStrategies = new() { { LayerRenderingStrategy.UseSpriteStrategy, spriteMatrix } };
```
moonheart08: Not a huge fan of allocating in this fast path. This is going to happen up to 1k+ times a frame. Make a struct for these matrices and pass it in by ref.

moonheart08 (changes requested):

hollowarc:
i beg, this would let us fix so many sprite bugs

Robust.Client/GameObjects/EntitySystems/SpriteSystem.Render.cs
SlamBamActionman: I've changed the Dict to a struct and passing it on as a ref to the subsequent functions, but it's still generating the matrices in the same location - is this sufficient? I am not sure how one would move it outside of `RenderSprite` since it relies on the angle/position to generate the matrices.

SlamBamActionman:
Test fail appears to be heisentest 🤔

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -231,6 +231,13 @@ public ShaderInstance? PostShader
         [DataField]
         public bool RaiseShaderEvent;
 
+        /// <summary>
+        ///     If set, overrides the default start-to-end rendering order of the <see cref="Layers"/> List with the provided index order.
+        ///     Layers not included in the override will only be rendered if their index in the <see cref="Layers"/> List is positioned after the highest index in the override.
+        /// </summary>
+        [ViewVariables]
+        public LinkedList<int>? LayersOrderOverride;
```
moonheart08: This will never be mutated enough to warrant using a linked list.

Robust.Client/GameObjects/EntitySystems/SpriteSystem.Render.cs
```diff
@@ -192,3 +267,14 @@ private void HandleShaderLayer(Layer layer, Texture texture, CopyToShaderParamet
...
+/// <summary>
+/// Stores the rendering matrices used for a sprite's layers.
+/// </summary>
+public struct LayerRenderingStrategyMatrices
+{
+    public Matrix3x2 UseSpriteStrategy;
+    public Matrix3x2 Default;
+    public Matrix3x2 SnapToCardinals;
+    public Matrix3x2 NoRotation;
+}
```
moonheart08: Make this `readonly` to throw the VM a bone.

Robust.Client/GameObjects/EntitySystems/SpriteSystem.Render.cs
```diff
@@ -60,55 +61,129 @@ public void RenderSprite(
...
+
+            RenderLayerWithChildren(sprite, sprite.Comp.Layers[childIndex], drawingHandle, ref matrices, angle, overrideDirection); // Recursion scares me.
+        }
     }
 
     /// <summary>
     /// Render a layer. This assumes that the input angle is between 0 and 2pi.
     /// </summary>
-    private void RenderLayer(Layer layer, DrawingHandleWorld drawingHandle, ref Matrix3x2 spriteMatrix, Angle angle, Direction? overrideDirection)
+    private void RenderLayer(Layer layer, DrawingHandleWorld drawingHandle, Matrix3x2 spriteMatrix, Angle angle, Direction? overrideDirection)
```
moonheart08: This should continue taking the matrix by `ref readonly`, please.

Robust.Client/GameObjects/EntitySystems/SpriteSystem.Render.cs
```diff
@@ -60,55 +61,129 @@ public void RenderSprite(
...
+    /// <summary>
+    /// Pre-step before rendering a sprite's layer, first rendering the given layer using the set rendering strategy and then recursively rendering any child layers after.
+    /// </summary>
+    private void RenderLayerWithChildren(Entity<SpriteComponent> sprite, Layer layer, DrawingHandleWorld drawingHandle, ref LayerRenderingStrategyMatrices matrices, Angle angle, Direction? overrideDirection)
+    {
+        if (sprite.Comp.GranularLayersRendering || layer.RenderingStrategy == LayerRenderingStrategy.UseSpriteStrategy)
+        {
+            RenderLayer(layer, drawingHandle, matrices.UseSpriteStrategy, angle, overrideDirection);
+        }
+        else
```
moonheart08: Roll this back into the switch statement, use `case _ when sprite.Comp.GranularLayersRendering:` as the second case for the same block.

Robust.Client/GameObjects/EntitySystems/SpriteSystem.Render.cs
```diff
@@ -60,55 +61,129 @@ public void RenderSprite(
...
+    }
 
-        foreach (var layer in sprite.Comp.Layers)
+    /// <summary>
+    /// Pre-step before rendering a sprite's layers, ensuring the order is overriden if <see cref="SpriteComponent.LayersOrderOverride"/> is set.
+    /// Layers with parents are skipped here, as they are assumed to be rendered as child layers instead.
+    /// </summary>
+    private void RenderLayersWithOverride(Entity<SpriteComponent> sprite, DrawingHandleWorld drawingHandle, ref LayerRenderingStrategyMatrices matrices, Angle angle, Direction? overrideDirection)
+    {
+        if (sprite.Comp.LayersOrderOverride != null)
```
moonheart08: The way this behaves if it for example contains duplicates needs very clearly documented, and it should probably not be *allowed* to contain duplicates to begin with. The way you figure out the "remaining layers" also doesn't really make sense and allows this list to practically prevent layers from rendering entirely by accident.

moonheart08 (changes requested):
This needs tests to ensure layer parenting works consistently and all edge cases (like reparenting, reordering parents, etc) are accounted for and supported.

Robust.Client/GameObjects/EntitySystems/SpriteSystem.Render.cs
SlamBamActionman: > it should probably not be *allowed* to contain duplicates to begin with

Yeah, that's fair. What would be the most performant way to track this? Since we can't be sure that every sprite layer will be in the LayersOrderOverride list (like child layers), it would need to track individual layer indexes.

> The way you figure out the "remaining layers" also doesn't really make sense and allows this list to practically prevent layers from rendering entirely by accident.

This is noted in the `LayersOrderOverride` summary; it only renders any non-indexed (non-child) layers appended to the end of the layers list. This is for two reasons:
- Certain systems that add a layer on top of the sprite do not have a corresponding mapping, they just slap them on at the end. Since they don't have a mapping they become difficult to account for when setting the `LayersOrderOverride` list, so rendering anything appended acts as a fallback to when the mapping's missing.
- Layers inserted in the middle of the layer list can't really be accounted for in the order override; the appended ones work because we know they are at the end, but inserted in the middle? With the reordering there's no info on what it should be between. So we end up intentionally skipping them. 

If these two things are undesirable I don't mind changing them, but I'd have to know what behavior would be better.

SlamBamActionman:
PR has been updated and is ready for review. Not sure the test fail is related hrm...

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -231,6 +231,20 @@ public ShaderInstance? PostShader
...
         public bool RaiseShaderEvent;
 
+        /// <summary>
+        ///     If set, overrides the default start-to-end rendering order of the <see cref="Layers"/> List with the provided index order for the given directions.
+        ///     Layers may not be skipped over when rendered. If layers have a parent/child relationship, only one layer in the chain must be included in the override.
+        ///     Any layers after the highest provided override index will be rendered on top, to account for overlay layers added during runtime.
+        /// </summary>
+        [ViewVariables]
+        public Dictionary<RsiDirection, List<int>>? LayersOrderOverride;
+
```
metalgearsloth: Considering RsiDirection only goes up to 8 I think just using an array will be significantly faster for marginal memory usage (if any, I forget dictionary overhead).
Same with using an array for the order override unless we expect content to be adjusting it every frame.

PR #6342, opened 2025-12-17
Expand SharedUserInterfaceSystem API

ArtisticRoomba:
Adds `IsUiOpen` (takes a List of UI keys to check) and `IsAnyUiOpen`.

It is sometimes useful to know if any UI is open, as well as if certain UIs are open. I don't want to write boilerplate every single time I want to check a list and I'd like to expose a simple API to check if a UI is open in general.

Required by https://github.com/space-wizards/space-station-14/pull/41885

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
@@ -802,6 +802,41 @@ public abstract class SharedUserInterfaceSystem : EntitySystem
     }
 
     /// <summary>
+    /// Returns true if any of the specified UI keys are open for this entity by anyone.
+    /// </summary>
+    /// <param name="entity">The entity to check.</param>
+    /// <param name="uiKeys">The UI keys to check.</param>
+    /// <returns>True if any UI is open, false otherwise.</returns>
+    [PublicAPI]
+    public bool IsUiOpen(Entity<UserInterfaceComponent?> entity, List<Enum> uiKeys)
+    {
+        if (!UIQuery.Resolve(entity.Owner, ref entity.Comp, false))
+            return false;
+
+        foreach (var key in uiKeys)
+        {
+            if (entity.Comp.Actors.TryGetValue(key, out var actors) && actors.Count > 0)
+                return true;
+        }
+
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
```diff
@@ -801,6 +801,41 @@ public bool IsUiOpen(Entity<UserInterfaceComponent?> entity, Enum uiKey, EntityU
...
     }
 
+    /// <summary>
+    /// Returns true if any of the specified UI keys are open for this entity by anyone.
+    /// </summary>
+    /// <param name="entity">The entity to check.</param>
+    /// <param name="uiKeys">The UI keys to check.</param>
+    /// <returns>True if any UI is open, false otherwise.</returns>
+    [PublicAPI]
+    public bool IsUiOpen(Entity<UserInterfaceComponent?> entity, List<Enum> uiKeys)
```
PJB3005: This should probably take an `IEnumerable<Enum>` instead.

Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
```diff
@@ -801,6 +801,41 @@ public bool IsUiOpen(Entity<UserInterfaceComponent?> entity, Enum uiKey, EntityU
...
+    {
+        if (!UIQuery.Resolve(entity.Owner, ref entity.Comp, false))
+            return false;
+
+        foreach (var key in uiKeys)
+        {
+            if (entity.Comp.Actors.TryGetValue(key, out var actors) && actors.Count > 0)
+                return true;
```
PJB3005: Why are you making sure the inner list is not empty, but not doing that below?

PJB3005 (changes requested):

Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
ArtisticRoomba: Ah, yeah. Unnecessary.

PJB3005 (approved):

PR #6345, opened 2025-12-20
Add a check for alwaysrender in UserInterfaceManager.cs

neomoth:
I found myself trying to render text to the right of a rendered image on a paper, and found that the only clean way to do this without any serious modifications was to set its height to zero, which works perfectly! Well, except for the fact that it'll get culled when scrolling past the top of it since it has a height of zero and it's being culled for optimization reasons.

This PR adds a single check to `UserInterfaceManager` in order to allow preventing this behavior via setting `AlwaysRender` to true. I am unsure if this is the best way to do this or if it would have been better to add another parameter for this, but I tried looking for what allows parameters to be defined when initializing a new control and couldn't find it, maybe I'm just blind but yeah.

Below are a video of showing before the fix and after the fix.

https://github.com/user-attachments/assets/0d988c41-7c1b-42d9-bcc2-637ee868fcbd

https://github.com/user-attachments/assets/5bdd36e7-03c5-45d5-ba12-8b99d20a0a04


PJB3005:
Using `AlwaysRender` for this is not at all appropriate.

Hacking the layout system with invalid heights is not supported, figure out something proper.

neomoth:
sighhhhhhhhh
lame but okay

PR #6347, opened 2025-12-21
Add support for layered configs

Penguinwizzard:
By repeatedly specifying --config-file, you can layer multiple config files on top of each other, to share base configuration for multiple servers or to handle files with different secrecy requirements.


PJB3005:
Is there a use case here that isn't handled fine by the config preset system in SS14?

PR #6349, opened 2025-12-23
Fix EntProtoId<T>.TryGet throwing an error on invalid prototype ids

DrSmugleaf:


PJB3005:
Should `ProtoId<T>` instances even *have* invalid values?

DrSmugleaf:
> Should `ProtoId<T>` instances even _have_ invalid values?

No, though I have seen it happen sometimes with default component values, and commands I think
This is mostly just so the TryGet command never throws, and only returns false if it fails on those
Could add an argument to the method too if you think its worth it, but Get throws either way

PR #6358, opened 2025-12-29
allow intentionally desyncing components on client

deltanedas:
this pr has 2 atomic commits:
- add SetComponentNetSync to EntityManager, the evil thing i need
- make TryGetAudio not throw on empty file path, and rewrite it to not copy paste an entire function

IgnoredNetComponents was added to MetaData to track these, currently i have only found 1 thing that needs it to avoid an incorrect debug assert

works good for my one use case, client muting networked audio and unmuting it.

goidamerge please


MLGTASTICa:
Very interested to see this merged. Lots of use-cases where  item containers could use desyncing for stuff that is done both server and client-side at the same time.

deltanedas:
its explicitly only usable by client

deltanedas:
@PJB3005 please merge

moonheart08 (approved):
While I think the main uses for this are all "system xyz is oblivious to usecases where the client has its own opinion", which really should be handled by fixing those systems, I understand why this is useful in the interim.

PR #6359, opened 2025-12-29
Add a HTTP client for client/server

DrSmugleaf:
Do not merge without PJB's approval I will blow up

My RMC GMs yearn for custom audio playing


VerinSenpai:
Hey that's pretty heckin cool

PJB3005 (changes requested):
This needs to block private IP address ranges

Also no compile

DrSmugleaf:
@PJB3005 Fixed in [`ce60c7a` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6359/commits/ce60c7afb52095a2f205dc4c570b99e313cea4c3), [`a682f5c` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6359/commits/a682f5cd418819a20531124f6371deed3a9ffb91), [`ab46563` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6359/commits/ab465637ad8e593418e3332fc497e0b30e461751), and [`3ef98ab` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6359/commits/3ef98abd6fa75ed5f77f8cf66d1012706d7074d9)

Robust.Shared/Network/HttpManager.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Runtime.Serialization;
+using System.Threading;
+using System.Threading.Tasks;
+using Robust.Shared.Utility;
+
+namespace Robust.Shared.Network;
+
+public sealed class HttpManager : IHttpManagerInternal
+{
+    private readonly HttpClient _client = new();
```
PJB3005: This needs to use `HappyEyeballsHttp` to avoid IPv6 issues.
Also add the user agent similar to `HttpClientHolder`.

Robust.Shared/Network/HttpManager.cs
```diff
@@ -0,0 +1,114 @@
...
+        await ThrowIfLocalUri(uri);
+        return await _client.GetFromJsonAsync<T>(uri, cancel);
+    }
+
+    private async Task ThrowIfLocalUri(Uri uri)
+    {
+        if (IPAddress.TryParse(uri.Host, out var ip))
+            ThrowIfLocalIP(ip);
+
+        var addresses = await Dns.GetHostAddressesAsync(uri.Host);
```
PJB3005, line 180: I'd prefer if these were checked from the connection handler to avoid any TOCTOU shit.

Robust.Shared/Network/HttpManager.cs
```diff
@@ -0,0 +1,114 @@
...
+        foreach (var dnsIP in addresses)
+        {
+            ThrowIfLocalIP(dnsIP);
+        }
+    }
+
+    private void ThrowIfLocalIP(IPAddress ip)
```
PJB3005, line 187: We have code for doing this in content that's probably more Robust, I'd use it instead:

https://github.com/space-wizards/space-station-14/blob/master/Content.Server/Connection/IPIntel/IPIntel.cs

PJB3005 (changes requested):

Robust.Shared/Network/HttpManager.cs
DrSmugleaf: Done in [`607dca3` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6359/commits/607dca337b22daeb14f22dc2033cfdae6f2dcb0c)

Robust.Shared/Network/HttpManager.cs
DrSmugleaf: Done in [`607dca3` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6359/commits/607dca337b22daeb14f22dc2033cfdae6f2dcb0c)

Robust.Shared/Network/HttpManager.cs
DrSmugleaf: Done in [`607dca3` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6359/commits/607dca337b22daeb14f22dc2033cfdae6f2dcb0c) (assuming I did it properly)

Unrelated, not sure if the new JSON method is safe to expose to content sandbox-wise or not

Robust.Shared/Network/HttpManager.cs
```diff
@@ -0,0 +1,217 @@
...
+using System.Net.Http.Json;
+using System.Net.Sockets;
+using System.Runtime.Serialization;
+using System.Threading;
+using System.Threading.Tasks;
+using Robust.Shared.Utility;
+
+namespace Robust.Shared.Network;
+
+public sealed class HttpManager : IHttpManagerInternal
```
moonheart08: Should be internal, content doesn't need access to this class and it can be relaxed later.


```suggestion
internal sealed class HttpManager : IHttpManagerInternal
```

Robust.Shared/Network/HttpManagerExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                return await http.GetFromJsonAsync<T>(uri, cancel);
+            }
+            catch
+            {
+                return default;
+            }
+        }
+
+        public async Task TryCopyToAsync(Uri uri, Stream stream, CancellationToken cancel)
```
moonheart08: This feels.. weirdly low utility? Not having any indicator your stream was simply never filled feels like a recipe for broken code. May be better to just not for this case.

Robust.Shared/SharedIoC.cs
```diff
@@ -50,6 +47,8 @@ public static void RegisterIoC(IDependencyCollection deps)
             deps.Register<HttpClientHolder>();
             deps.Register<RobustMemoryManager>();
             deps.Register<EntityConsoleHost>();
+            deps.Register<IHttpManager, HttpManager>();
+            deps.Register<IHttpManagerInternal, HttpManager>();
```
moonheart08, line 51: This isn't being initialized twice here, right?

Robust.Shared.Tests/Networking/HttpManagerTest.cs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    [Test]
+    [TestCase("http://192.168.0.0")]
+    [TestCase("http://192.168.0.1")]
+    [TestCase("http://192.168.255.255")]
+    [TestCase("http://10.0.0.0")]
+    [TestCase("http://10.255.255.255")]
+    [TestCase("http://172.16.0.0")]
+    [TestCase("http://172.31.255.255")]
```
moonheart08, line 33:
```suggestion
    [TestCase("http://172.31.255.255")]
    [TestCase("http://localhost")]
    [TestCase("http://[::1]")]
```

moonheart08 (changes requested):

SaphireLattice:
I admit I'm rather confused as to utility of this vs the WebSocket Transfer API?

moonheart08:
> I admit I'm rather confused as to utility of this vs the WebSocket Transfer API?

There's some valid reasons for the client to simply make HTTP requests on its own, like interacting with user-specified equipment (....see the pishock PR) and locally managed asset downloads (i.e. the server has no business being in charge of it, the walkmans in CM come to mind)

Robust.Shared/Network/HttpManagerExtensions.cs
DrSmugleaf: Removed TryCopyToAsync in [`f7ae393` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6359/commits/f7ae3938b55c201b6de9103ac3cedfde3d7fa375)

Robust.Shared/SharedIoC.cs
DrSmugleaf: TaskManager and ParallelManager do the same so hopefully not
Breakpointing the constructor it's only called once on client at least

Robust.Shared/Network/HttpManager.cs
```diff
@@ -0,0 +1,217 @@
...
+    {
+        var response = await GetAsync(uri, cancel);
+        return await response.Content.ReadAsByteArrayAsync(cancel);
+    }
+
+    public async Task<T?> GetFromJsonAsync<T>(Uri uri, CancellationToken cancel = default)
+    {
+        var response = await GetAsync(uri, cancel);
+        return await response.Content.ReadFromJsonAsync<T>(cancel);
+    }
```
moonheart08, line 106: Sus on this method. Use SerializationManager to parse JSON instead.

moonheart08 (changes requested):
One last little thing and this then looks good to me.

Robust.Shared/Network/HttpManagerExtensions.cs
PJB3005, line 1: Not a fan of these helpers. Too catch-all for something that can have many reasons for failure.

Robust.Shared/Network/HttpManager.cs
PJB3005: Yeah this isn't sandbox safe.

Robust.Shared/Network/HttpManager.cs
PJB3005: Yeah no this isn't solved. You need to do the check inside the http request handler, which in this case is `HappyEyeballsHttp.CreateHttpHandler()`

PJB3005 (changes requested):

PR #6361, opened 2025-12-31
increase Loc.GetString log level to warning

slarticodefast:
These break all the time in content when someone changes the message ids while refactoring something and we currently only notice when a player makes a bug report that a popup or UI element is not showing this correctly.

By increasing the log level to warning we can catch these if they happen during integration tests and easily find any missing ones by looking at warnings on grafana. For cases where someone expects the loc id to not exist they can use `TryGetString` instead. The documentation for that already mentions that it does not cause a warning, so I assume `GetString` was always intended to do that.

Ideally the message ids would not be hardcoded strings in the first place, but validated similar to proto ids, but some of them are generated via string formatting, so we can't use LocIds in all cases.


slarticodefast:
Test fails seem unrelated. Not sure what is going on there.

VerinSenpai:
TestSpawnTraversal is complaining about a resources directory missing.

Also good shit 👌

PR #6362, opened 2026-01-02
SpriteView updates its size whenever the entity is set

TGRCdev:
This is a fairly simple change to call ~~`UpdateSize()`~~ `InvalidateMeasure()` when a SpriteView's entity changes. This resolves an issue we were having downstream with non-standard-sized species having their heads and feet truncated. 
An Anomalocarid person with the top and bottom of their sprite truncated


PJB3005 (changes requested):
You should be calling `InvalidateMeasure()` instead.

PJB3005 (approved):

PR #6375, opened 2026-01-09
Revert "make SharedAudioSystem.Stop not return early when the current tick has already been predicted"

PJB3005:
Reverts space-wizards/RobustToolbox#6137

### DIFF
--- a/Robust.Shared/Audio/Systems/SharedAudioSystem.cs
+++ b/Robust.Shared/Audio/Systems/SharedAudioSystem.cs
@@ -455,7 +455,7 @@ public abstract partial class SharedAudioSystem : EntitySystem
         if (uid == null || !Resolve(uid.Value, ref component, false))
             return null;
 
-        if (_netManager.IsClient && !IsClientSide(uid.Value))
+        if (!Timing.IsFirstTimePredicted || (_netManager.IsClient && !IsClientSide(uid.Value)))
             return null;
 
         QueueDel(uid);


Princess-Cheeseballs:
I'm 99% sure this is a content side issue. Per the PR's comment:

> I don't understand why this is the correct approach, what's the benefit of making stop call repeatedly if predicted repeatedly?
As in the linked issue - Stop and PlayPredicted are being used in a predicted instance.
PlayPredicted uses first time predicted and will only occur once. However, Stop is not first time predicted and will override a previous prediction's PlayPredicted and the audio just doesn't play.

PlayPredicted returns early so we aren't spawning a trillion entities during prediction, that makes sense. With deletion we don't really have to worry about that? The client can only delete clientside entities here and we're gonna want to delete them if we're calling this method. 

The linked issue was content sided and looking into this I found another bug on content that I need to resolve regarding predicted audio.

Princess-Cheeseballs:
Coming back to say I think that again, this is a content bug and reverting this change in engine was a mistake. This method returns null to clear any saved audio streams by intention, but because it returns early without stopping audio if it's not first time predicted, we need to add an additional `if (_timing.IsFirstTimePredicted)` check to every single shared or client-side usage of this method so it doesn't clear the saved stream without stopping the audio.

deathride58:
We disagree that this PR is what makes `IsFirstTimePredicted` a necessity for stopping managed looping audio; space-wizards/space-station-14#42409 was developed in the context of an engine version containing 6137, and when we lacked a `IsFirstTimePredicted` check then, this lead to mispredictions causing the looping audio to simply never play (as mispredictions just stop the audio outright). Bolting on a `IsFirstTimePredicted` check fixed this.

Similarly, space-wizards/space-station-14#42588 was developed against an engine version containing this PR. Now, when lacking `IsFirstTimePredicted`, mispredictions instead cause the exact opposite issue: mispredictions without `IsFirstTimePrediction` lead to the variable keeping track of the audio stream getting prematurely nulled out due to early return, and thus making that looping audio permanent.

With or without this PR, `IsFirstTimePredicted` checks are a hard necessity when keeping track of looping audio entities in the same way that grappling hooks do

Princess-Cheeseballs:
> With or without this PR, `IsFirstTimePredicted` checks are a hard necessity when keeping track of looping audio entities in the same way that grappling hooks do

That's true but I think that the audio stream should only get overwritten as null in the scenario where the audio is actually being stopped. Having the extra check within the method, and especially having it return null is not ideal.

Although, ideally we should be passing in a reference to the current audio stream anyways and only overriding it if the stream is stopped/doesn't exist and then returning early if it's not first time predicted.

PR #6378, opened 2026-01-15
Makes RemoveJoint(uid, id) recursive by default

deathride58:
This is a companion PR to space-wizards/space-station-14#42409

We had spent several hours straight trying to figure out a way to get PortalSystem to break a joint through any means possible before we decided to just chuck some code directly into the engine.

Currently, `RemoveJoint()` does nothing if you're trying to remove a relayed joint. If you wanna remove a relayed joint, you gotta first look for said relayed joint. This is a massive headache.

This PR is a massive improvement to the `RemoveJoint()` API because it allows `RemoveJoint()` to work a lot more like one would intuitively expect when relays are involved. Coders shouldn't have to add an entire for loop just to get rid of a single joint when it could very easily just be a single function call.

With the exception of the companion PR (which is written with this PR in mind; portals won't properly sever grappling hook joints without this PR), this PR won't affect anything on upstream. Below section should be copy-pasted as a notice to downstreams that may potentially expect `RemoveJoint()` to *not* delete relayed joints.

## Breaking Changes
`RemoveJoint(uid, id)` now accepts an optional recursive argument, which is true by default. This recursive argument controls whether or not relayed joints attached to the uid are also removed. Code that relies on relayed joints *not* being deleted should set the recursive argument to false.

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedJointSystem.cs
+++ b/Robust.Shared/Physics/Systems/SharedJointSystem.cs
@@ -504,17 +504,35 @@ public abstract partial class SharedJointSystem : EntitySystem
         }
     }
 
-    public void RemoveJoint(EntityUid uid, string id)
+    /// <summary>
+    /// Removes any joints attached to this object with a specified ID.
+    /// </summary>
+    /// <param name="uid">Entity UID of the object to remove the joint from</param>
+    /// <param name="id">ID of the joint to remove</param>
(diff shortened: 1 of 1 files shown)


deathride58:
We're unable to reproduce the test fail on local here; it needs a restart

moonheart08 (approved):

Tayrtahn (approved):

PR #6379, opened 2026-01-16
Let viewport clear when eye is missing 

Ruzihm:
Currently when a viewport's eye is null or in nullspace, the render target won't be cleared, which means the view stays constant until the eye is no longer null or in nullspace.

This PR's goal is to allow viewports to be configured to clear the render target instead with the clear color.

Example use case: Rendering the viewport as black when eye is null/in nullspace in OD:

https://github.com/OpenDreamProject/OpenDream/pull/2477


Ruzihm:
Actually, since this PR might not be helpful for that particular use case. Going to leave it open in case this feature is wanted anyway.

PJB3005 (approved):

PR #6382, opened 2026-01-17
Add events on animation starts

Ataman:
Adds an `AnimationStartedEvent` fired by the `AnimationPlayerSystem`. It's set to broadcast in order to be consistent with the existing `AnimationCompletedEvent`.
Also gives `Control` an `AnimationStarted` event for consistency's sake as well.

The reason for this is to give content systems more awareness about animations and let them react to them if desired. Also, it feels right to have this given `AnimationCompletedEvent` is already available. Finally, it would be useful for my proposed [Animation State Machine](https://github.com/space-wizards/docs/pull/560).

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
@@ -154,6 +154,9 @@ namespace Robust.Client.GameObjects
             }
 
             ent.Comp.PlayingAnimations.Add(key, playback);
+
+            var startedEvent = new AnimationStartedEvent(ent.Owner, ent.Comp, key);
+            EntityManager.EventBus.RaiseLocalEvent(ent.Owner, startedEvent, true);
         }
 
         public bool HasRunningAnimation(EntityUid uid, string key)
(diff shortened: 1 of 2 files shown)


Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
```diff
@@ -154,6 +154,9 @@ public void Play(Entity<AnimationPlayerComponent> ent, Animation animation, stri
             }
 
             ent.Comp.PlayingAnimations.Add(key, playback);
+
+            var startedEvent = new AnimationStartedEvent(ent.Owner, ent.Comp, key);
+            EntityManager.EventBus.RaiseLocalEvent(ent.Owner, startedEvent, true);
```
PJB3005: Use the proxy methods instead.

Robust.Client/GameObjects/EntitySystems/AnimationPlayerSystem.cs
```diff
@@ -199,6 +202,34 @@ public void Stop(EntityUid uid, AnimationPlayerComponent? component, string key)
...
+        /// The animation player component associated with the entity this event was raised on.
+        /// </summary>
+        public AnimationPlayerComponent AnimationPlayer { get; init; }
+
+        /// <summary>
+        /// The key associated with the animation that was started.
+        /// </summary>
+        public string Key { get; init; } = string.Empty;
+
+        public AnimationStartedEvent(EntityUid uid, AnimationPlayerComponent animationPlayer, string key)
```
PJB3005: Make this constructor internal.

PJB3005 (changes requested):

Ataman:
Thanks for reviewing, requested changes are in.

PJB3005 (approved):

PR #6391, opened 2026-01-20
Add new TryGetLayer implementation

neomoth:
Adds a new implementation for `SpriteSystem.TryGetLayer` that does not require the passed component to be attached to an entity.

The main reason for this is to avoid using obsolete code from `SpriteComponent` should a case come up where you have access to the component but either can't get the entity, or the entity does not exist.

Currently in my fork I am implementing something onto encryption keys that allows you to alter its base sprite layer and icon sprite layer via ViewVariables, and I needed to make sure that when these values are null, it restores the original sprite from that prototype. I found the best way to do this was to literally get the sprite component from the prototype, but this component instance isn't attached to anything, leaving the only way to get the layer data without this implementation of `SpriteSystem.TryGetLayer` is to use `SpriteComponent.TryGetLayer`, which is obsolete.

The PR in question can be found [here](https://github.com/ss14Starlight/space-station-14/pull/2635) if you wish to examine how I am doing this and why I feel that this is a needed implementation, since again, the only other way to get the layer data as far as I know is to use obsolete code.

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/SpriteSystem.Layer.cs
+++ b/Robust.Client/GameObjects/EntitySystems/SpriteSystem.Layer.cs
@@ -45,6 +45,28 @@ public sealed partial class SpriteSystem
         return false;
     }
 
+    /// <summary>
+    /// Alternate implementation of TryGetLayer that does not require a valid entity to be attached to the sprite component.
+    /// </summary>
+    public bool TryGetLayer(
+        SpriteComponent sprite,
+        int index,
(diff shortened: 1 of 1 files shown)


neomoth:
Gonna be real IDK why this says to change some shit in a submodule, unsure if thats like important or not cuz all i know i touched was `SpriteSystem.Layer.cs`

neomoth:
closing cuz this clearly isnt getting looked at and also i need to remake my fork anyway so idk what deleting the fork would even do to the PR anyway so meh

PR #6393, opened 2026-01-22
Fix RichText height calculation for mixed fonts and inline controls

chavonadelal:
## Problem
RichTextEntry used to calculate its height while walking markup nodes, instead of calculating it from the actual content that ended up on each rendered line.

This caused several related layout bugs:
- Lines with smaller fonts were still clamped to at least the default font height.
- Lines with larger fonts could be under-measured, especially on the last line or after a newline inside the same text node.
- Lines with mixed font metrics could be under-measured when one font had a larger ascent and another had a larger descent.
- Wrapped text could assign the height of a large word to the wrong line.
- Inline controls affected horizontal layout, but were not treated as normal layout items for line breaks and line height.
- Update() and Draw() had separate vertical layout logic, which could make measured Height disagree with the rendered layout.

Because these bugs depended on where wrapping happened and which node/font/control was current at that exact point, the issue appeared intermittent rather than constant.

image

## Fix
RichTextEntry now uses a per-line layout metrics model.

Update() builds a unified stream of layout items from the rich text content. Both text runes and inline controls participate in the same layout flow:
- text runes contribute font metrics;
- inline controls contribute their measured size;
- inline controls occupy one layout index;
- LineBreaks are now based on layout item indices, not only text rune indices.

WordWrap now has explicit support for inline objects. An inline control is treated as one unsplittable item and as a word boundary, so wrapping can happen before or after it in the same way rendering expects.

After wrapping is known, RichTextEntry groups layout items by their actual lines and calculates per-line metrics:
- ContentHeight: the actual maximum height of the items on that line;
- Advance: the vertical step to the next line;
- Ascent: the baseline offset used for drawing mixed fonts;
- Descent: the space needed below the baseline for mixed fonts.

Font items use font height, font line height, font ascent, and font descent. Inline controls use their measured height. Empty explicit lines and fully empty messages still fall back to the default font metrics, preserving existing empty-line behavior such as `echo`.

Draw() no longer has a separate ad-hoc line advance calculation. It now consumes the same per-line metrics produced during Update(), so measured Height and rendered vertical layout stay consistent.

image


Robust.Client/UserInterface/RichTextEntry.cs
```diff
@@ -128,14 +128,17 @@ public RichTextEntry Update(MarkupTagManager tagManager, Font defaultFont, float
...
             LineBreaks.Clear();
 
             int? breakLine;
             var wordWrap = new WordWrap(maxSizeX);
             var context = new MarkupDrawingContext();
             context.Font.Push(defaultFont);
             context.Color.Push(_defaultColor);
+            
+            // Track maximum height for current line to handle non-standard font sizes
+            var maxCurrentLineHeight = defaultFont.GetHeight(uiScale);
```
moonheart08: Shouldn't this be initialized to 0 so that text being smaller than this is not broken? Yes, existing bug ofc.

chavonadelal: > Shouldn't this be initialized to 0 so that text being smaller than this is not broken? Yes, existing bug ofc.

Yep, you're right. Starting from `defaultFont.GetHeight()` was incorrectly clamping non-empty lines to at least the default font height.

I addressed that as part of a broader cleanup of the line height calculation: non-empty lines now derive their height from the actual layout items on that line, while empty explicit lines still use the default font fallback.

Robust.Client/UserInterface/RichTextEntry.cs
```diff
@@ -328,5 +419,67 @@ private static int GetLineHeight(Font font, float uiScale, float lineHeightScale
...
+            {
+                return new LayoutItemMetrics(height, height, 0, false);
+            }
+        }
+
+        private readonly struct LineMetrics
+        {
+            public readonly int ContentHeight;
+            public readonly int Advance;
+            public readonly int Ascent;
```
Aeshus, line 499: Shouldn't there also be something for descent?

Aeshus (changes requested):
This breaks the current behavior of the `echo` command in the console as it used to show an extra empty line but it doesn't anymore. Please document this as a breaking change if this is intentional.

Also, you have Advance and Ascent, but no Descent; why? Couldn't that be needed in case there's a font with large descents but not similarly large ascents?

Somewhat unrelated (as it's also a bug on master) but: I get a debug assert crash on shrinking the guidebook and that is still reproducible with this PR... It also crashes when I make external `uitest2` window with the RichTextEntry have a very thin width.

chavonadelal:
> This breaks the current behavior of the `echo` command in the console as it used to show an extra empty line but it doesn't anymore. Please document this as a breaking change if this is intentional.
> 
> Also, you have Advance and Ascent, but no Descent; why? Couldn't that be needed in case there's a font with large descents but not similarly large ascents?
> 
> Somewhat unrelated (as it's also a bug on master) but: I get a debug assert crash on shrinking the guidebook and that is still reproducible with this PR... It also crashes when I make external `uitest2` window with the RichTextEntry have a very thin width.

Thanks, good catch.

I restored the previous empty RichTextEntry behavior, so `echo` should keep showing an empty output line and this should not be a breaking change anymore.

I also added Descent to the line metrics and changed the line height calculation to account for max ascent + max descent. This should handle the case where a line mixes fonts with different vertical extents, including fonts with large descents.

Regarding the guidebook / `uitest2` crash: I looked into it briefly. It appears to be an existing WordWrap edge case where the available width is smaller than a single glyph or inline object. WordWrap assumes that if the current line exceeds maxSizeX, there is a valid word/glyph break point, but if the item starts at the beginning of the line and still does not fit, that invariant does not hold and the debug assert can fire.

I have not run into this issue myself before, and I am not familiar enough with that crash to confidently include it in this PR. Since it reproduces on master and seems unrelated to the mixed-font / inline-control height fix, I left it untouched here to avoid mixing in a separate fix. As a possible fix, we could replace the debug assert with a defensive fallback that keeps an unbreakable glyph/control as an oversized line item, but I would rather not include that here unless explicitly approved because it is a separate master bug.

eoineoineoin:
Nice work; gave this a bit of a bash and it seems to be working nicely.

There does seem to be one wrapping bug - it was introduced in #6267, rather than your PR, so up to you if you want to tackle it.

In this video, I've just written a paper with the contents:

> isdofiasdfuisdf seemstobeaprobleminhowthiswordiswrapped dsio [head=3]DSF[/head] [head=2]ASDF[/head]

https://github.com/user-attachments/assets/0f2c9db8-11ce-4a1e-b119-fc7b59279702

When the wrapping has to split "seemstobeaprobleminhowthiswordiswrapped", it breaks the word at a strange midpoint of the word, rather than at the end of the word, where you'd expect. Then, as the window shrinks, as soon as that weird-midpoint hits the start of the word, the entire wrapping algorithm fails and nothing is wrapped.

PR #6394, opened 2026-01-23
Make EntitySystemManager.DependencyCollection inject EntityQuery, make BUIs inject systems and entity queries

DrSmugleaf:
The technology is insane
No more 100 lines of boilerplate code in systems and BUIs
With this you can do the following in systems and bound user interfaces without having to assign any fields manually
```csharp
[Dependency] private readonly TransformSystem _transformSystem = default!;
[Dependency] private readonly IConfigurationManager _config = default!;
[Dependency] private readonly EntityQuery<PhysicsComponent> _physicsQuery = default!;
```

### DIFF
--- a/Robust.Shared/GameObjects/Components/UserInterface/BoundUserInterface.cs
+++ b/Robust.Shared/GameObjects/Components/UserInterface/BoundUserInterface.cs
@@ -32,6 +32,7 @@ namespace Robust.Shared.GameObjects
 
         protected BoundUserInterface(EntityUid owner, Enum uiKey)
         {
+            IoCManager.Resolve(ref EntMan);
             EntMan.EntitySysManager.DependencyCollection.InjectDependencies(this);
             UiSystem = EntMan.System<SharedUserInterfaceSystem>();
 
--- a/Robust.Shared/GameObjects/Components/UserInterface/BoundUserInterface.cs
+++ b/Robust.Shared/GameObjects/Components/UserInterface/BoundUserInterface.cs
@@ -32,7 +32,7 @@ namespace Robust.Shared.GameObjects
 
         protected BoundUserInterface(EntityUid owner, Enum uiKey)
         {
-            IoCManager.InjectDependencies(this);
+            EntMan.EntitySysManager.DependencyCollection.InjectDependencies(this);
             UiSystem = EntMan.System<SharedUserInterfaceSystem>();
 
             Owner = owner;
--- a/Robust.Shared/GameObjects/EntitySystemManager.cs
+++ b/Robust.Shared/GameObjects/EntitySystemManager.cs
@@ -23,11 +23,11 @@ namespace Robust.Shared.GameObjects
 {
     public sealed class EntitySystemManager : IEntitySystemManager, IPostInjectInit
     {
-        [IoC.Dependency] private readonly IReflectionManager _reflectionManager = default!;
-        [IoC.Dependency] private readonly IEntityManager _entityManager = default!;
-        [IoC.Dependency] private readonly ProfManager _profManager = default!;
-        [IoC.Dependency] private readonly IDependencyCollection _dependencyCollection = default!;
-        [IoC.Dependency] private readonly ILogManager _logManager = default!;
+        [Dependency] private readonly IReflectionManager _reflectionManager = default!;
+        [Dependency] private readonly IEntityManager _entityManager = default!;
+        [Dependency] private readonly ProfManager _profManager = default!;
+        [Dependency] private readonly IDependencyCollection _dependencyCollection = default!;
+        [Dependency] private readonly ILogManager _logManager = default!;
 
 #if EXCEPTION_TOLERANCE
         [Dependency] private readonly IRuntimeLog _runtimeLog = default!;
--- a/Robust.Shared/GameObjects/EntitySystemManager.cs
+++ b/Robust.Shared/GameObjects/EntitySystemManager.cs
@@ -193,30 +193,13 @@ namespace Robust.Shared.GameObjects
                 _systemTypes.Remove(baseType);
             }
 
-            var stopwatch = new Stopwatch();
-            stopwatch.Start();
-            var queryMethod = typeof(EntityManager).GetMethod(nameof(EntityManager.GetEntityQuery), 1, []);
-            var registeredComponentTypes = _entityManager.ComponentFactory.AllRegisteredTypes.ToArray();
-            var queryArray = new (Type Type, object Instance)[registeredComponentTypes.Length];
-            Parallel.For(
-                0,
-                registeredComponentTypes.Length,
-                i =>
-                {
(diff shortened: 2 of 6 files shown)


deltanedas:
blessed pr godspeed

Robust.Shared/GameObjects/EntitySystemManager.cs
```diff
@@ -192,6 +193,27 @@ public void Initialize(bool discover = true)
...
+            var registeredComponentTypes = _entityManager.ComponentFactory.AllRegisteredTypes.ToArray();
+            var queryList = new List<(Type Type, object Instance)>(registeredComponentTypes.Length);
+            Parallel.ForEach(registeredComponentTypes,
+                type =>
+                {
+                    var queryType = typeof(EntityQuery<>).MakeGenericType(type);
+                    var query = queryMethod!.MakeGenericMethod(type).Invoke(_entityManager, null)!;
+                    queryList.Add((queryType, query));
+                }
+            );
```
Tayrtahn: I'm guessing the test failure is due to the parallel instances finishing out of order, causing the results to be inserted into `queryList` in the wrong order.

Maybe use `Parallel.For` so you have the right index?

deltanedas: can preallocate the list as well with the right count / use an array

DrSmugleaf: Engine test fixed in [`d70e6a9` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6394/commits/d70e6a9adbcd09175388e2ffbc82789bf611da45), content one still running

Robust.Shared/GameObjects/Components/UserInterface/BoundUserInterface.cs
```diff
@@ -32,7 +32,7 @@ public abstract class BoundUserInterface : IDisposable
 
         protected BoundUserInterface(EntityUid owner, Enum uiKey)
         {
-            IoCManager.InjectDependencies(this);
+            EntMan.EntitySysManager.DependencyCollection.InjectDependencies(this);
```
Tayrtahn, line 36: I'm not sure, but I think the issue here is that `EntMan` is itself a dependency that needs to be injected, so it's still null here. Might need to manually resolve it.

Robust.Shared/GameObjects/EntitySystemManager.cs
```diff
@@ -22,11 +23,11 @@ namespace Robust.Shared.GameObjects
...
-        [Dependency] private readonly IReflectionManager _reflectionManager = default!;
-        [Dependency] private readonly IEntityManager _entityManager = default!;
-        [Dependency] private readonly ProfManager _profManager = default!;
-        [Dependency] private readonly IDependencyCollection _dependencyCollection = default!;
-        [Dependency] private readonly ILogManager _logManager = default!;
+        [IoC.Dependency] private readonly IReflectionManager _reflectionManager = default!;
+        [IoC.Dependency] private readonly IEntityManager _entityManager = default!;
+        [IoC.Dependency] private readonly ProfManager _profManager = default!;
+        [IoC.Dependency] private readonly IDependencyCollection _dependencyCollection = default!;
+        [IoC.Dependency] private readonly ILogManager _logManager = default!;
```
Tayrtahn: Also, Rider trolled you again.

Robust.Shared/GameObjects/EntitySystemManager.cs
```diff
@@ -192,6 +193,31 @@ public void Initialize(bool discover = true)
...
+                }
+            );
+
+            foreach (var (type, instance) in queryArray)
+            {
+                // This runs a lock so probably no point in parallelizing it, if this changes just move it up
+                SystemDependencyCollection.RegisterInstance(type, instance);
+            }
+
+            _sawmill.Debug($"Added {nameof(EntityQuery<>)} for all component types to IoC in {stopwatch.ElapsedMilliseconds:F0} ms");
```
PJB3005: Instead of doing this, maybe make it so `IDependencyCollection` can take a callback to dynamically construct instances for unresolved types? Then you don't need to add queries to it for *every* component. I think that's also like closer to what the solution in asp.net and such does.

Robust.Shared/GameObjects/EntitySystemManager.cs
DrSmugleaf: Fixed in [`ef4873f` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6394/commits/ef4873f872d06f6e2627bfcfc99ab41287b5312a)

Robust.Shared/GameObjects/Components/UserInterface/BoundUserInterface.cs
DrSmugleaf: Fixed in [`4c66804` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6394/commits/4c6680432f7d217aaaa10516ff5971000aa9eb40)

Robust.Shared/GameObjects/EntitySystemManager.cs
DrSmugleaf: Thought about doing that at first but wasn't sure if the added complexity inside dependency collection was worth it or not
Done in [`1e3e398` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6394/commits/1e3e398591b3d29c220ede64b33a0361efb51cf1) though I'm not convinced I did it correctly

Robust.Shared/IoC/DependencyCollection.cs
```diff
@@ -128,7 +139,19 @@ private bool TryResolveType(
...
         {
             if (!services.TryGetValue(objectType, out instance))
+            {
+                if (_lazyServices.TryGetValue(objectType, out instance))
+                    return true;
+
+                if (objectType.IsGenericType && _baseGenericLazyFactories.TryGetValue(objectType.GetGenericTypeDefinition(), out var factory))
+                {
+                    instance = factory(objectType, this);
+                    _lazyServices[objectType] = instance;
```
PJB3005: Probably best to use `GetOrAdd` on the `ConcurrentDictionary` here to avoid race conditions returning multiple instances.

PJB3005 (changes requested):

Robust.Shared/IoC/DependencyCollection.cs
DrSmugleaf: Done in [`432c2a7` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6394/commits/432c2a714a1b62ce27ff5f733a862a907fb6b353), it means it uses one more dict lookup if the type already exists the way I wrote it

Tayrtahn (approved):
Looks good to me, but I'd want to have @PJB3005 sign off on that last change.

PJB3005 (approved):

PR #6395, opened 2026-01-23
add debug info to static protoid in generic class error

deltanedas:
for this example pseudocode:
```
public sealed class EpicSystem<TComp> : EntitySystem
{
    public static readonly ProtoId<ShaderPrototype> SomeShader = "Hello";
}
```

currently when validating protoids you get a useless exception with 0 information whatsoever
this pr makes it so you have the protoid type and the field name, which makes fixing the issue fairly trivial
```
Field Robust.Shared.Prototypes.ProtoId`1[Robust.Client.Graphics.ShaderPrototype] SomeShader cannot be a static field inside a generic class
```

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.ValidateFields.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.ValidateFields.cs
@@ -94,18 +94,16 @@ public partial class PrototypeManager
     private bool TryGetIds(FieldInfo field, [NotNullWhen(true)] out string[]? ids)
     {
         ids = null;
-        object? value;
-        try
+        if (field.DeclaringType.IsGenericTypeDefinition)
         {
-            value = field.GetValue(null);
-            if (value == null)
-                return false;
-        }
-        catch (InvalidOperationException e)
-        {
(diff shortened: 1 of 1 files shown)


deltanedas:
@Tayrtahn kept you waiting, huh?

Robust.Shared/Prototypes/PrototypeManager.ValidateFields.cs
```diff
@@ -94,9 +94,18 @@ private void ValidateStaticField(
...
-        if (value == null)
-            return false;
+        object? value;
+        try
+        {
+            value = field.GetValue(null);
+            if (value == null)
+                return false;
+        }
+        catch (InvalidOperationException e)
```
PJB3005: You should really check if the type has unbound generics instead of relying on the exception being thrown.

PJB3005 (changes requested):

PJB3005:
Compile fail real

deltanedas:
im phoneposting what do you expect

PJB3005 (approved):

PR #6396, opened 2026-01-24
Cleanup warnings: CS0414

B-Kirill:
Cleanup 1x `CS0414` warning.
The suppression of some warnings has also been replaced with a more correct approach.


Tayrtahn (approved):
For reference, regarding the use of `#if` vs `#pragma`: https://github.com/space-wizards/RobustToolbox/pull/6069#issuecomment-3050607114

CI builds all configurations these days, so this is fine.

PR #6400, opened 2026-01-28
Add ProfManager.Value guard, write first command argument as a ProfManager value in ExecuteInShell

DrSmugleaf:
Value works the same as Group, but instead of writing GroupStart and GroupEnd it just writes a value when disposed
Also made EntitySystemManager use this
The command writing won't work properly with toolshed commands, but after having a quick look at the code for those I decided that I do not want to figure out how to split those up and put them in the ProfManager


PJB3005 (approved):
Systems probably should be groups instead but there's no harm in this feature ig.

PR #6406, opened 2026-01-31
Correct updating of modifier key states

### LINKED ISSUE #6369: Swapping caps lock and ctrl on Linux breaks using ctrl modifier as a keybind
## What version did the issue occur in?

wizards/78189bb

## Description

I use the "Swap Ctrl and Caps Lock" option from gnome-tweaks on my Linux machine to make Caps Lock on my keyboard behave as a Ctrl key for convenience. In the Options menu in SS14, I am able to set keybinds using the Ctrl modifier as expected (the keybinds show up in the Options screen), but in game the keybind doesn't function - neither the Caps Lock nor Ctrl key work. This is annoying since I've had to find replacement mappings for pulling & throwing objects and some text functions.

I will try to experiment with other ways to swap caps lock & ctrl keys in Linux to see if that resolves this issue, but this is the simplest way to do it and all other games/applications on my computer respect this setting.

## Reproduction

1. Run on Linux using the Gnome desktop (default on Ubuntu LTS 24.04)
2. Install the gnome-tweaks package
3. In gnome-tweaks, go to Keyboard > Additional Layout Options and turn on the "Swap Ctrl and Caps Lock" option:

Image

4. Run SS14, connect to a server, and try to Pull an object with the default keybind of Ctrl + Left Click. Nothing happens.
5. Type something in the chat box then try to use one of the commands (like Ctrl + Left Click) to select text
6. Open the Options menu, change the keybind for one of your menus to use the Ctrl key, then confirm that no longer works in game:

Image

pxc1984:
fix https://github.com/space-wizards/RobustToolbox/issues/6369

### DIFF
--- a/Robust.Client.Tests/Input/ModifierAliasKeybindTest.cs
+++ b/Robust.Client.Tests/Input/ModifierAliasKeybindTest.cs
@@ -1,0 +1,342 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Numerics;
+using NUnit.Framework;
+using Robust.Client.Input;
+using Robust.Client.Graphics;
+using Robust.Client.UserInterface;
+using Robust.Client.UserInterface.Controls;
+using Robust.Client.UserInterface.CustomControls;
+using Robust.Shared.Console;
+using Robust.Shared.ContentPack;
+using Robust.Shared.Input;
+using Robust.Shared.IoC;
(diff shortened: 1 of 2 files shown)


PJB3005:
So wait, enabling this swap option causes us to not receive key events for control explicitly?

PJB3005:
Stale.

PR #6407, opened 2026-02-02
Serializer for shorter YAML tags when using abstract base types

Ataman:
This PR adds a custom serializer named `CustomBaseTypeSerializer` that lets coders shorten the whole `!type:BaseTypeSomething` into just `!Something` given that the base type ends in `Base` and their concrete implementations replace `Base` (and only `Base`) with `Something`. The serializer only applies when specified using the `DataField` attribute and includes support for most common collections.

Includes a DiagnosticAnalyzer with pre-compile time check (to prevent the serializer being applied on incorrectly named types) and unit tests for serializer and analyzer. It adds RA00** to Diagnostics.cs for the analyzer. I couldn't find any conventions or process for requesting these and just added it for now, correct me if I got it wrong.

Rider comparison of hypothetical YAML before and after applying the serializer:
image

Finally, I'm not that proficient with YAML, feel free to reject this and suggest a better solution (or a better name instead of `CustomBaseTypeSerializer`). I'm happy to write something different if it's useful. This was also my first time writing a DiagnosticAnalyzer; it looks for `typeof()` expressions and checks the type param when `CustomBaseTypeSerializer` is in use. If there is a better approach, let me now.


Ataman:
PR #6227 appears to use RA0045 already. Will check and update before merge if it comes to that.

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/CustomBaseTypeSerializer.cs
```diff
@@ -0,0 +1,279 @@
...
+    ITypeSerializer<TBase, MappingDataNode>,
+    ITypeSerializer<TBase, ValueDataNode>,
+    ITypeSerializer<TBase[], SequenceDataNode>,
+    ITypeSerializer<List<TBase>, SequenceDataNode>,
+    ITypeSerializer<HashSet<TBase>, SequenceDataNode>,
+    ITypeSerializer<Queue<TBase>, SequenceDataNode>
+    where TBase : notnull
+{
+    // CustomType|Base -> CustomType
+    private static readonly string  BaseNameWithoutBase = typeof(TBase).Name.Replace("Base", string.Empty);
```
moonheart08: PlayerBaseTypeBase -> PlayerType


Needs to be smarter.

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/CustomBaseTypeSerializer.cs
```diff
@@ -0,0 +1,279 @@
...
+    {
+        return typeof(TBase).Name.EndsWith("Base");
+    }
+
+    private static string ExpandName(string name)
+    {
+        // !ConcreteThing -> !type:CustomTypeConcreteThing
+        return name.StartsWith("!type:")
+            ? name
+            : name.Replace("!", $"!type:{BaseNameWithoutBase}");
```
moonheart08: Should be future-proofed against further processing directives being added later. In other words, any tag with a `:` in it should be ignored, not just type.

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/CustomBaseTypeSerializer.cs
```diff
@@ -0,0 +1,279 @@
...
+    private static DataNode WriteDataNode(
+        ISerializationManager serializationManager,
+        TBase value,
+        bool alwaysWrite,
+        ISerializationContext? context = null)
+    {
+        DebugTools.Assert(EndsWithBase());
+        var node = serializationManager.WriteValue(value.GetType(), value, alwaysWrite, context);
+        DebugTools.Assert(node.Tag == null);
+        node.Tag = '!' + value.GetType().Name.Replace(BaseNameWithoutBase, string.Empty);
```
moonheart08: This too is wrong if the type name contains BaseNameWithoutBase. I don't believe that's *possible*, but this should still be prefix based.

Additionally, an analysis that enforces descendants of the base type are named properly is necessary for this to be correct.

moonheart08 (changes requested):
Needs release notes.

All the string based stuff needs reinforced + needs to support using an attribute to specify the mapping instead.

```cs
[ChildTypeTag<ChildA>("A")]
[ChildTypeTag<ChildB>("B")]
[ChildTypeTag<ChildC>("C")]
public class BaseType();
```

Ataman:
@moonheart08 Thank you for the review! Just to confirm, the custom mapping attribute should allow using the passed string as a short form in YAML even if the serializer has not been specified inside the DataField attribute, correct?

moonheart08:
> @moonheart08 Thank you for the review! Just to confirm, the custom mapping attribute should allow using the passed string as a short form in YAML even if the serializer has not been specified inside the DataField attribute, correct?

If the attribute is used then yes, this is the behavior I'd expect. Can just sourcegen a serializer definition on the spot.

Down the line the same general idea will apply to C# Unions when those land in .NET 11-12.

Ataman:
Alright, makes sense to me. Thanks for confirming that this wasn't just a dumb idea on my part. It will take a bit more time though while I read up on source generators.

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/CustomBaseTypeSerializer.cs
Ataman: Should be smarter now.

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/CustomBaseTypeSerializer.cs
Ataman: Now ignores all strings containing `:`

Ataman:
Quick update; I managed to implement the source generator and went a step further by adding another attribute for specifying the serializer itself.

Example:
```csharp
[YamlTagShortener]
[CustomChildTag<ALegacyThing>("A")]
public abstract class SomethingBase;

public class ALegacyThing : SomethingBase;
```

Would auto-generate a default serializer which works automatically on correctly named types and also convert `!A` into `ALegacyThing`.

By doing it this way, the custom type serializer can still be used and existing DataFields don't require any changes.

I still need to adapt the analyzer however. Commits will follow once that works as well.

Ataman:
Sorry for the wait, we got a bit of a crunch going at work which is kind of draining.

I managed to finish the analyzer now but another issue popped up that I initially didn't notice after switching to the `[YamlTagShortener]` attribute: It seems the engine doesn't recognize a type serializer decorated with `[TypeSerializer]` for writing when abstract classes are involved. It does recognize it for reading however and after trying track down the issue for a while without success, I decided to downgrade the serializer into a reader. I'll be happy to re-add the writer parts again if somebody can figure out what's wrong.

Robust.Shared/Serialization/TypeSerializers/Implementations/Generic/CustomBaseTypeSerializer.cs
Ataman: Analyzer now checks for naming conventions on child classes.

Ataman:
Will update the diagnostics codes and release notes once the PR has been reviewed.

PR #6409, opened 2026-02-04
Add CVar attribute to register on value changed automatically for fields, properties, and methods on systems, services, and controllers

DrSmugleaf:
Example usage:
```csharp
[CVar<CVars>(nameof(CVars.NetMaxConnections))]
public int Value;

[CVar<CVars>(nameof(CVars.NetMaxConnections))]
public void OnMaxConnectionsChanged(int val)
{
    ...
}
```

Now I don't need to remind every contributor to pass true to OnValueChanged forget 1 time and be hit by the curse of Ra

### DIFF
--- a/Robust.Shared/Configuration/ConfigurationManager.cs
+++ b/Robust.Shared/Configuration/ConfigurationManager.cs
@@ -10,6 +10,7 @@ using Nett;
 using Robust.Shared.Collections;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
+using Robust.Shared.Reflection;
 using Robust.Shared.Timing;
 using Robust.Shared.Utility;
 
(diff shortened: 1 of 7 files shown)


Robust.Shared/Configuration/ConfigurationManager.cs
```diff
@@ -1062,6 +1119,15 @@ public override string ToString()
                 return "<VIRTUAL>";
             }
         }
+
+        void IPostInjectInit.PostInject()
```
moonheart08, line 1123: I'm not entirely convinced w/ this as a solution, especially as this makes having these attributes available a "did your manager get registered before or after the point BuildGraph() is called for ConfigurationManager" question. 

At the minute this is only gonna effect stuff very early in engine and nothing else, because it runs too early.

Should this perhaps be some lower level ioc hook instead (like the ability to register something that runs for all newly registered dependencies)? But I think the less insane alternative change would be to just.. call RegisterCVarAttributes on an as-needed basis for managers that want to use them.

moonheart08 (changes requested):

PJB3005 (changes requested):
Not a fan of this relying on reflection.

Also, this has no mechanism for unsubscribing, which leaks it on entity systems.

PR #6415, opened 2026-02-08
GridFixtureSystem Optimizations

Aidenkrz:
## About the PR

Optimizes `GridFixtureSystem` in a few hot paths that were doing more work (and allocating more garbage) than they really needed to.

Ended up doing a bunch more fuckery to SharedGridFixtureSystem as well after looking through it as well, refer to technical details for info on that.

## Why / Balance

One of my maintainers pointed out that there was some evil LINQ garbage in this system that called `.Sum()` a bajillion times whenever `CheckSplits` ran. Not the end of the world, but also easily avoidable. This cleans that up and makes the whole thing scale way better without changing behavior. Also noticed system was allocating new collections constantly, so i changed that as well.

## Technical details

- The BFS scratch structures (`_splitFrontier` queue and `_splitGrids` list) are now reused as class fields instead of being allocated every `CheckSplits` call.
- The sort comparator was doing repeated LINQ `.Sum()` calls, resulting in something like
  $`O(N \times M \times \log N)`$ work.
  This is now replaced with a precomputed `gridSizes` dictionary, bringing it down to
  $`O(N \times M + N \times \log N)`$.

  As a rough example:
  - For N = 50 grids with M = 100 tiles each, this cuts the comparator work by ~90%.
  - For N = 200 grids with M = 250 tiles each, it’s closer to a ~97% reduction.

  I didn't do any real benchmarking as I don't have dottrace, but I mean, the math checks out.

- In `CreateNodes`, the tile `HashSet` now only tracks non-empty tiles (capacity set to `chunk.FilledTiles`) instead of blindly allocating for all 256 chunk positions.
- Cross-chunk neighbor detection now bails early for interior tiles, since only edge tiles can possibly have neighbors in adjacent chunks.
- `MapGridComponent` is now passed through the full call chain as a nullable parameter with TryComp/Comp fallback.
- Grid fixture diffing in `UpdateFixture` now cross-references old and new fixtures by ID instead of using a full `Fixture.Equals` comparison. The old approach always failed because content (e.g. `ShuttleSystem.OnGridFixtureChange`) modifies fixture density after creation, causing every fixture to be destroyed and recreated on every tile change. The new approach compares shapes only via `EqualsApprox`, preserving content-set properties like density. This also eliminated a redundant shape check that existed solely as a workaround for the broken equality comparison, and fixed a latent bug where `chunk.Fixtures` (a `HashSet<string>`) was being modified during `foreach` enumeration.
- During grid splits, entity reparenting now does a single broadphase query per node using the union bounding box of all its tiles, instead of one query per tile. Entities are filtered by converting their position to a tile index and checking membership in a `HashSet<Vector2i>`. This reduces broadphase queries from O(tiles) to O(1) per node.


Aidenkrz:
accidentally left the PR title as the commit name and cant change it now :-1:

ArtisticRoomba:
github stays winning

Tayrtahn:
Apparently you can still change it from the mobile app. You can also just tell me what title you'd like and I'll change it for you.

Aidenkrz:
> Apparently you can still change it from the mobile app. You can also just tell me what title you'd like and I'll change it for you.

Thanks I used GitHub mobile

moonheart08 (approved):
Looking for footage of this being tested, ideally, just as a "this obviously works". The changes lgtm though.

DrSmugleaf (approved):
Outside of DungeonJob it seems like the only other caller to CheckSplits is SharedMapSystem.RegenerateCollision, which seem fine at a glance

Aidenkrz:
https://github.com/user-attachments/assets/87943c16-8550-4661-82d7-117b1e665302

Robust.Shared/Map/Events/RegenerateGridBoundsEvent.cs
```diff
@@ -11,11 +12,13 @@ namespace Robust.Shared.Map.Events;
...
+internal readonly record struct RegenerateGridBoundsEvent(EntityUid Entity, Dictionary<MapChunk, List<Box2i>> ChunkRectangles, List<MapChunk> RemovedChunks, MapGridComponent? Grid = null)
 {
     public readonly EntityUid Entity = Entity;
 
     public readonly Dictionary<MapChunk, List<Box2i>> ChunkRectangles = ChunkRectangles;
 
     public readonly List<MapChunk> RemovedChunks = RemovedChunks;
+
+    public readonly MapGridComponent? Grid = Grid;
 }
```
Tayrtahn:
```suggestion
internal readonly record struct RegenerateGridBoundsEvent(EntityUid Entity, Dictionary<MapChunk, List<Box2i>> ChunkRectangles, List<MapChunk> RemovedChunks, MapGridComponent? Grid = null);
```

Robust.Shared/GameObjects/Systems/SharedGridFixtureSystem.cs
```diff
@@ -150,66 +147,49 @@ private bool UpdateFixture(EntityUid uid, MapChunk chunk, List<Box2i> rectangles
...
-                if (existing)
+                    newFixtures.Remove(oldId);
                     continue;
+                }
 
-                // Doesn't align with any new fixtures so delete
-                chunk.Fixtures.Remove(oldId);
-                _fixtures.DestroyFixture(uid, oldId, oldFixture, false, body: body, manager: manager, xform: xform);
-                updated = true;
+                toRemove.Add(oldId);
```
Tayrtahn: I've tracked down the fix for https://github.com/space-wizards/space-station-14/issues/42329 to this change. Reverting just this section of code (and the Type change for `newFixtures`) causes the bug to reappear, so ~the cause must be somewhere within this~. My current hypothesis is that it's something to do with `ValueList`, but I need to do more investigating.

Edit: Alternatively, the bug lies somewhere in the fixture replacement code, and the fix to fixture reuse means that it's being avoided.

Edit2: Indeed it must have something to do with the fixture replacement code. Modifying this PR so it never reuses old fixtures (by commenting out lines 160-167) causes the bug to reappear.

Tayrtahn (approved):

Tayrtahn:
Pinging @metalgearsloth to take a look at this, since it's largely his code being modified here.

Robust.Shared/GameObjects/Systems/SharedGridFixtureSystem.cs
```diff
@@ -150,66 +147,49 @@ private bool UpdateFixture(EntityUid uid, MapChunk chunk, List<Box2i> rectangles
 #pragma warning restore CS0618
 
                 var key = string.Create(CultureInfo.InvariantCulture, $"grid_chunk-{bounds.Left}-{bounds.Bottom}");
-                newFixtures.Add((key, newFixture));
+                newFixtures[key] = newFixture;
```
metalgearsloth: There should never be duplicates so this should be adding.

Robust.Shared/GameObjects/Systems/SharedGridFixtureSystem.cs
```diff
@@ -102,27 +103,23 @@ internal void RegenerateCollision(
...
-            // So we store a reference to the fixture on the chunk because it's easier to cross-reference it.
-            // This is because when we get multiple fixtures per chunk there's no easy way to tell which the old one
-            // corresponds with.
             // We also ideally want to avoid re-creating the fixture every time a tile changes and pushing that data
             // to the client hence we diff it.
-
             // Additionally, we need to handle map deserialization where content may have stored its own data
             // on the grid (e.g. mass) which we want to preserve.
-            var newFixtures = new ValueList<(string Id, Fixture Fixture)>();
+            var newFixtures = new Dictionary<string, Fixture>(rectangles.Count);
```
metalgearsloth: Others got promoted to fields but this got promoted to a local reftype?

Robust.Shared/GameObjects/Systems/SharedMapSystem.cs
```diff
@@ -154,10 +154,12 @@ public GridRemovalEvent(EntityUid uid)
     public sealed class GridInitializeEvent : EntityEventArgs
     {
         public EntityUid EntityUid { get; }
+        public MapGridComponent? Grid { get; }
 
-        public GridInitializeEvent(EntityUid uid)
+        public GridInitializeEvent(EntityUid uid, MapGridComponent? grid = null)
```
metalgearsloth: Why is it nullable?

Robust.Server/Physics/GridFixtureSystem.cs
```diff
@@ -248,14 +252,20 @@ private void CheckSplits(EntityUid uid, HashSet<ChunkSplitNode> dirtyNodes)
...
-                        .CompareTo(y.Sum(o => o.Indices.Count)));
+                var gridSizes = new Dictionary<HashSet<ChunkSplitNode>, int>(grids.Count);
+                foreach (var sizeGroup in grids)
+                {
+                    var tileCount = 0;
+                    foreach (var sizeNode in sizeGroup)
+                        tileCount += sizeNode.Indices.Count;
+                    gridSizes[sizeGroup] = tileCount;
+                }
+                grids.Sort((x, y) => gridSizes[x].CompareTo(gridSizes[y]));
```
metalgearsloth: Ref types again.

Robust.Server/Physics/GridFixtureSystem.cs
```diff
@@ -320,25 +330,38 @@ private void CheckSplits(EntityUid uid, HashSet<ChunkSplitNode> dirtyNodes)
 
                         // Update lookup ents
                         // Needs to be done before setting old tiles as they will be re-parented to the map.
-                        // TODO: Combine tiles into larger rectangles or something; this is gonna be the killer bit.
+                        // Build tile positions and union bounds so we can query once per node.
+                        var tilePositions = new HashSet<Vector2i>(node.Indices.Count);
```
metalgearsloth: Also here.

Robust.Server/Physics/GridFixtureSystem.cs
```diff
@@ -320,25 +330,38 @@ private void CheckSplits(EntityUid uid, HashSet<ChunkSplitNode> dirtyNodes)
...
-                                    !bounds.Contains(entXform.LocalPosition)) continue;
+                            if (entXform.ParentUid != oldGridUid)
+                                continue;
 
-                                _xformSystem.SetParent(ent, entXform, newGridUid, _xformQuery, newGridXform);
-                            }
+                            var entTile = new Vector2i(
+                                (int) Math.Floor(entXform.LocalPosition.X / oldGrid.TileSize),
+                                (int) Math.Floor(entXform.LocalPosition.Y / oldGrid.TileSize));
+
```
metalgearsloth, line 366: Shouldn't it be using localtotile / some helper method instead of inline maths.

Robust.Server/Physics/GridFixtureSystem.cs
```diff
@@ -404,13 +427,14 @@ private ChunkNodeGroup CreateNodes(EntityUid gridEuid, MapGridComponent grid, Ma
...
-            var tiles = new HashSet<Vector2i>(chunk.ChunkSize * chunk.ChunkSize);
+            var tiles = new HashSet<Vector2i>(chunk.FilledTiles);
 
             for (var x = 0; x < chunk.ChunkSize; x++)
             {
                 for (var y = 0; y < chunk.ChunkSize; y++)
                 {
-                    tiles.Add(new Vector2i(x, y));
+                    if (!chunk.GetTile((ushort) x, (ushort) y).IsEmpty)
+                        tiles.Add(new Vector2i(x, y));
```
metalgearsloth, line 443: Realistically you could just bitmask a ulong to avoid the array access entirely but I'll do that in my pr.

metalgearsloth (changes requested):

Robust.Server/Physics/GridFixtureSystem.cs
```diff
@@ -49,6 +50,17 @@ public sealed partial class GridFixtureSystem : SharedGridFixtureSystem
 
         private HashSet<EntityUid> _entSet = new();
 
+        private readonly Queue<ChunkSplitNode> _splitFrontier = new(4);
+        private readonly List<HashSet<ChunkSplitNode>> _splitGrids = new(1);
+        private readonly Dictionary<HashSet<ChunkSplitNode>, int> _splitGridSizes = new();
+        private readonly HashSet<Vector2i> _splitTilePositions = new();
+        private readonly Comparison<HashSet<ChunkSplitNode>> _splitGridSizeComparison;
+
+        public GridFixtureSystem()
```
Aidenkrz, line 59: can just be a lambda in the `Sort` if preferable, ts a lil weird but is better

Roudenn:
Profiled on Tools configuration (with grid splitting enabled), spawned some big grids out of lattice and cut them, and then divided the Dev map in halfs

Before:
image

After:
image

Robust.Server/Physics/GridFixtureSystem.cs
metalgearsloth: Your thing made a dummy ctor but I'll fix it.

PR #6416, opened 2026-02-08
Send server URL on hasJoined

VasilisThePikachu:
This pull request makes the engine send out the servers current server URL when advertising.

This is intended for future additions to auth. Which include hub moderation and additions to helping with account recovery.

I am setting `hub.serverurl` in the advertisement flow so that we can be sure we have some value set to use later. Also we probably should not query the ip service every time we try to advertise and should just "cache" it.


Robust.Server/ServerHub/HubManager.cs
```diff
@@ -51,7 +51,7 @@ public async void Start()
             _sawmill.Info("hub.server_url unset. Trying to determine IP address automatically...");
             try
             {
-                url = await GuessAddress();
+                _cfg.SetCVar(CVars.HubServerUrl, url);
```
PJB3005, line 55: Uhm, what?

PJB3005 (changes requested):

Robust.Server/ServerHub/HubManager.cs
VasilisThePikachu: Please read my PR description

PJB3005: And... you're removing the guessing logic?

VasilisThePikachu: Ok I am an idiot I see what this is about . Please be more verbose next time

PJB3005 (approved):
I fixed it.

PR #6417, opened 2026-02-09
Fix: `OccluderSystem` ignored entity rotation when calculating AABB

ThereDrD0:
## Issue
Occluder system translated box2 but did not take into account entity rotation. This logic only works if the box is a square.
But if someone creates non-square box2 - any rotation breaks the logic.

I found this bug while debugging why people could see through my double airlocks.

## Media

Before - a rotated double airlock with `boundingBox: "-0.5, -0.5, 1.5, 0.5"` in Occluder is half "transparent"

https://github.com/user-attachments/assets/8f564fd9-cda9-4669-ae5e-4ce5117336c7

After - you can't see through it anymore. Great!

https://github.com/user-attachments/assets/20962083-97ee-4ac3-8189-82c1ab3497c1

Double airlock prototype i used when recording videos.
```yml
- type: entity  
  id: DoubleAirlock  
  parent: Airlock  
  name: double airlock  
  components:  
  - type: Sprite  
    sprite: double_basic.rsi  
    offset: 0.5,0  
    snapCardinals: false  
  - type: Fixtures  
    fixtures:  
      fix1:  
        shape:  
          !type:PhysShapeAabb  
          bounds: "-0.49,-0.49,1.49,0.49"  
        density: 100  
        mask:  
        - FullTileMask  
        layer:  
        - AirlockLayer  
  - type: Occluder  
    boundingBox: "-0.5, -0.5, 1.5, 0.5"
```

### DIFF
--- a/Robust.Shared/GameObjects/Systems/OccluderSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/OccluderSystem.cs
@@ -43,7 +43,9 @@ public abstract class OccluderSystem : ComponentTreeSystem<OccluderTreeComponent
     protected override Box2 ExtractAabb(in ComponentTreeEntry<OccluderComponent> entry)
     {
         DebugTools.Assert(entry.Transform.ParentUid == entry.Component.TreeUid);
-        return entry.Component.BoundingBox.Translated(entry.Transform.LocalPosition);
+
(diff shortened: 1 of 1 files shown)


PJB3005:
Multi-tile occluders cannot work with the current occlusion system. This PR doesn't work.

PR #6420, opened 2026-02-15
Fix OnClientRequestFull throwing an error when trying to log data about a deleted entity

DrSmugleaf:
Slightly suboptimal for the logging to be throwing an error for an error we are handling
The client also seemed to just get stuck forever waiting for better days


Robust.Server/GameStates/PvsSystem.cs
```diff
@@ -241,10 +241,9 @@ private void OnClientRequestFull(ICommonSession session, GameTick tick, NetEntit
         var sb = new StringBuilder();
         sb.Append($"Client {session} requested full state on tick {tick}. Last Acked: {lastAcked}. Curtick: {_gameTiming.CurTick}.");
 
-        if (missingEntity != null)
+        if (missingEntity != null && TryGetEntityData(missingEntity.Value, out var uid, out _))
```
moonheart08: Should this not still log and just remark the entity doesn't exist? It likely existed previously.

PJB3005 (changes requested):
Kaylie's review comment is correct.

PR #6426, opened 2026-02-22
Make UserInterfaceComponent#Interfaces AlwaysPushInheritance

sowelipililimute:
We have a lot of places in content where we inherit user interfaces but have to manually compose the fields; this makes us not need to do that anymore


moonheart08 (approved):
Needs changelog.

Tayrtahn (approved):

PR #6431, opened 2026-02-28
Improved collision filter test

eoineoineoin:
Minor improvements to collision filter test:

SharedPhysicsSystem was doing an O(n*m) test for every possible fixture pair when testing two FixturesComponents for collision. But because it's not doing any narrowphase collisions, the same result can be calculated in O(n+m) steps, just by summing up the collision layers/masks.

Added an interface to check IsHardCollidable() using just the fixtures components. I wanted this when checking collision with an EntityPrototype, so didn't have an Entity yet.

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Fixtures.cs
+++ b/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Fixtures.cs
@@ -149,7 +149,7 @@ public abstract partial class SharedPhysicsSystem
             return false;
         }
 
-        // Fast check
+        // Check the collision filter on the rigid bodies
         if (!bodyA.Comp2.Hard ||
             !bodyB.Comp2.Hard ||
             ((bodyA.Comp2.CollisionLayer & bodyB.Comp2.CollisionMask) == 0x0 &&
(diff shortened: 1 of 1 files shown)


moonheart08 (approved):
Tentative approval, I'm iffy on the idea of encouraging people to do physics checks without a real entity, but the logic *looks* fine. Please double-check your logic in-game if this code path doesn't have any tests.

eoineoineoin:
> Tentative approval, I'm iffy on the idea of encouraging people to do physics checks without a real entity, but the logic _looks_ fine. Please double-check your logic in-game if this code path doesn't have any tests.

Pretty confident that it's equivalent - have written a few physics engines in my time :). Robust.Shared.Tests/Physics is pretty light on tests, but happy to write some if you like?

The code before was effectively doing:
```
for every fixtureA in A
     for every fixtureB in B
           if fixtureA.Mask & fixtureB.Layer != 0 || fixtureA.Layer & fixtureB.Mask != 0
                        return true
return false
```

So I'm just transforming this to get rid of the n^2 loop, by combining all the layers/masks and testing them once:
```
allLayersA = A[0].Layer | A[1].Layer | A[2].Layer | ....
allMasksA = A[0].Mask | A[1].Mask | A[2].Mask | ....
allLayersB = B[0].Layer | B[1].Layer | B[2].Layer | ....
allMasksB = B[0].Mask | B[1].Mask | B[2].Mask | ....

if allLayersA & allMasksB != 0 || allMasksA & allLayersB != 0
      return true;
return false;
````

PJB3005 (approved):

PR #6433, opened 2026-03-01
Serialize EyeComponent.DrawLight

Golubgik:
I can't think of any reason why this would be a bad thing. But it adds more possibilities for administration and testing


moonheart08 (changes requested):
This field is managed by SetDrawLight, which is responsible for keeping these fields in sync. They cannot "just" be made VV-able.

Golubgik:
> This field is managed by SetDrawLight, which is responsible for keeping these fields in sync. They cannot "just" be made VV-able.

DrawFov is not managed in the same way, but still has VV?

moonheart08:
> DrawFov

This means whoever tagged DrawFov as vvwritable shouldn't have, not that this change makes sense.

PR #6434, opened 2026-03-01
Center UI bugfix

insoPL:
Center property is not pointing at the center in UIBox2i. That property is not used in ss14 codebase at all.

### DIFF
--- a/Robust.Shared.Maths/UIBox2i.cs
+++ b/Robust.Shared.Maths/UIBox2i.cs
@@ -23,8 +23,7 @@ namespace Robust.Shared.Maths
         public readonly int Width => Math.Abs(Right - Left);
         public readonly int Height => Math.Abs(Top - Bottom);
         public readonly Vector2i Size => new(Width, Height);
-        public readonly Vector2 Center => (TopRight + BottomLeft) / 2f;
-
+        public readonly Vector2 Center => new Vector2(Left + Right, Top + Bottom) / 2f;
         public UIBox2i(Vector2i topLeft, Vector2i bottomRight)
         {
             Unsafe.SkipInit(out this);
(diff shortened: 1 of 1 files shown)


moonheart08 (approved):
ough.

Robust.Shared.Maths/UIBox2i.cs
```diff
@@ -23,7 +23,7 @@ public struct UIBox2i : IEquatable<UIBox2i>, ISpanFormattable
         public readonly int Width => Math.Abs(Right - Left);
         public readonly int Height => Math.Abs(Top - Bottom);
         public readonly Vector2i Size => new(Width, Height);
-        public readonly Vector2 Center => TopRight - BottomLeft / 2;
+        public readonly Vector2 Center => (TopRight + BottomLeft) / 2f;
```
Tayrtahn:
```suggestion
        public readonly Vector2 Center => (TopLeft + BottomRight) / 2f;
```
It's not all that important, but since `TopRight` and `BottomLeft` are properties that each allocate a new vector, I think we can save a couple allocations by using the other corners.

Tayrtahn (approved):

Robust.Shared.Maths/UIBox2i.cs
insoPL: Damn, those  suggested changes do not show up on GH mobile app at all. It already happened several times that I implemented changes myself instead of just pressing big "commit suggestion" button instead xD

moonheart08: > It's not all that important, but since `TopRight` and `BottomLeft` are properties that each allocate a new vector, I think we can save a couple allocations by using the other corners.

Vector creation doesn't allocate, they're valuetype structs.

PR #6438, opened 2026-03-06
[Proposal] Add support for using a specific sprite for the placement overlay

Menshin:
## Rationale

The current Placement system is handling the placement overlay pretty straightforward : when entering placement mode the global placement overlay entity's sprite is generated from the associated prototype.

Content systems that needs more (construction, pipe layering, etc) can hijack the engine placement system to change its default behavior, but there's no way to change the Sandbox/Mapping native placement behavior.

While this is pretty often enough, situations arise when that limitation becomes problematic : for example, turnstiles can be a pain to place correctly because the default icon/sprite is used (while there are specific sprites showing the turnstill orientation that could be better used in placement mode).

## What's proposed here

This PR propose a basic system to allows using a specific sprite on placement mode for a given prototype :

* a new optional `placementOverlaySprite` field is added to `SpriteComponent` (could be its own component too), specifying a RSI/State texture.
* if that field is set, the placement system is slightly modified to use this specific texture when building the placement overlay.

## Showcase

Placing turnstills on Sandbox mode, using forementioned sprites for placement overlay :

https://github.com/user-attachments/assets/dcf27efe-9c59-4a4a-89fa-4bbab6f3cbf3

Adapting the (content) construction system to use this feature :

https://github.com/user-attachments/assets/f776a126-a8e5-48a8-a78a-bf833b08d60a

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -167,7 +167,7 @@ namespace Robust.Client.GameObjects
         }
 
         [DataField("sprite", readOnly: true)] private string? rsi;
-        [DataField("placementOverlaySprite", readOnly: true)] public SpriteSpecifier? placementOverlaySprite;
+        [DataField(readOnly: true)] public SpriteSpecifier? placementOverlaySprite;
         [DataField("layers", readOnly: true)] private List<PrototypeLayerData> layerDatums = new();
 
         [DataField(readOnly: true)] private string? state;
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -167,6 +167,7 @@ namespace Robust.Client.GameObjects
         }
 
         [DataField("sprite", readOnly: true)] private string? rsi;
+        [DataField("placementOverlaySprite", readOnly: true)] public SpriteSpecifier? placementOverlaySprite;
         [DataField("layers", readOnly: true)] private List<PrototypeLayerData> layerDatums = new();
 
         [DataField(readOnly: true)] private string? state;
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -167,7 +167,6 @@ namespace Robust.Client.GameObjects
         }
 
         [DataField("sprite", readOnly: true)] private string? rsi;
-        [DataField(readOnly: true)] public SpriteSpecifier? placementOverlaySprite;
         [DataField("layers", readOnly: true)] private List<PrototypeLayerData> layerDatums = new();
 
         [DataField(readOnly: true)] private string? state;
--- a/Robust.Client/Placement/PlacementManager.cs
+++ b/Robust.Client/Placement/PlacementManager.cs
@@ -743,7 +743,18 @@ namespace Robust.Client.Placement
             CurrentPrototype = prototype;
             IsActive = true;
 
-            CurrentPlacementOverlayEntity = EntityManager.SpawnEntity(templateName, MapCoordinates.Nullspace);
+            //does prototype uses a specific placement overlay sprite?
+            if (prototype.TryGetComponent<SpriteComponent>(out var prototypeSprite, IoCManager.Resolve<IComponentFactory>())
+                && prototypeSprite.placementOverlaySprite != null)
+            {
+                CurrentPlacementOverlayEntity = EntityManager.SpawnEntity(null, MapCoordinates.Nullspace);
+                SetPlacementOverlaySprite(prototypeSprite, Sprite.RsiStateLike(prototypeSprite.placementOverlaySprite));
+            }
+            else
+            {
+                CurrentPlacementOverlayEntity = EntityManager.SpawnEntity(templateName, MapCoordinates.Nullspace);
+            }
+
             EntityManager.RunMapInit(
                 CurrentPlacementOverlayEntity.Value,
                 EntityManager.GetComponent<MetaDataComponent>(CurrentPlacementOverlayEntity.Value));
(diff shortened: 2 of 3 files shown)


Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -167,6 +167,7 @@ public RSI? BaseRSI
         }
 
         [DataField("sprite", readOnly: true)] private string? rsi;
+        [DataField("placementOverlaySprite", readOnly: true)] public SpriteSpecifier? placementOverlaySprite;
```
B-Kirill: A little tip:
You don't need to specify a key name in `[DataField]` if it matches a field name (in `camelCase`). So you can remove `"placementOverlaySprite"`.

Menshin:
@B-Kirill remark applied.

~~CI weirdly failed on the install dependencies step...~~

B-Kirill:
> CI weirdly failed on the install dependencies step...

Things happen. Just create an empty commit to restart the tests.

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -167,6 +167,7 @@ public RSI? BaseRSI
         }
 
         [DataField("sprite", readOnly: true)] private string? rsi;
+        [DataField(readOnly: true)] public SpriteSpecifier? placementOverlaySprite;
```
Tayrtahn: I'm definitely not a fan of adding a new field to `SpriteComponent` for this specific use. A new component would be more appropriate.

Tayrtahn (changes requested):

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
Menshin: I was wondering what was the best course.

Will do.

Menshin:
As requested, moved placementOverlay to its own component.

Robust.Shared/Placement/PlacementOverlayComponent.cs
```diff
@@ -0,0 +1,14 @@
+using Robust.Shared.GameObjects;
+using Robust.Shared.Serialization.Manager.Attributes;
+using Robust.Shared.Utility;
+
+namespace Robust.Shared.Placement
```
Tayrtahn: Please use file-scoped namespaces.

Robust.Shared/Placement/PlacementOverlayComponent.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    [RegisterComponent]
+    public sealed partial class PlacementOverlayComponent : Component
```
Tayrtahn: Add xmldoc summary to the component.

Robust.Shared/Placement/PlacementOverlayComponent.cs
```diff
@@ -0,0 +1,14 @@
...
+    {        /// <summary>
+             ///     A SpriteSpecifier that will be used while in placement mode for this prototype
+             /// </summary>
```
Tayrtahn: Start the comment block on a new line.

Robust.Shared/Placement/PlacementOverlayComponent.cs
```diff
@@ -0,0 +1,14 @@
...
+        [DataField(readOnly: true, required: true)] public SpriteSpecifier sprite;
```
Tayrtahn:
```suggestion
        [DataField(readOnly: true, required: true)] public SpriteSpecifier Sprite;
```

Menshin:
Thanks for the feedback and review, the formatting should be ok now.

Robust.Client/Placement/PlacementManager.cs
```diff
@@ -860,6 +870,28 @@ private void RequestPlacement(EntityCoordinates coordinates)
...
+            {
+                //Fallback
+                Sprite.AddTextureLayer(CurrentPlacementOverlayEntity.Value, overlayTexture.Default);
+                return;
+            }
+
+            EntityManager.EnsureComponent<SpriteComponent>(CurrentPlacementOverlayEntity.Value, out var overlaySprite);
+            Sprite.AddRsiLayer(CurrentPlacementOverlayEntity.Value, overlayRSI.StateId, overlayRSI.RSI);
+
+            if(prototype.TryGetComponent<SpriteComponent>(out var prototypeSprite, IoCManager.Resolve<IComponentFactory>()))
```
moonheart08: No freestanding resolves. Use a dependency.

Robust.Client/Placement/PlacementManager.cs
```diff
@@ -743,7 +744,16 @@ private void PreparePlacement(string templateName)
             CurrentPrototype = prototype;
             IsActive = true;
 
-            CurrentPlacementOverlayEntity = EntityManager.SpawnEntity(templateName, MapCoordinates.Nullspace);
+            if (prototype.TryGetComponent<PlacementOverlayComponent>(out var placementOverlay, IoCManager.Resolve<IComponentFactory>()))
```
moonheart08: Ditto.

moonheart08 (changes requested):

Robust.Client/Placement/PlacementManager.cs
Menshin: Done

Menshin:
Realized there was no reason for the overlay sprite `noRotation` and `scale` to be set according to the prototype sprite ones, so decoupled them.

Minus the potential reviews/requests, the PR should be in its final form.

PJB3005:
I do not think it is worth the effort to continue investing in placement as a system like this.

Menshin:
Fair enough.
What would be a better approach ? Or is this an non-issue (from the engine pov) ?

PJB3005:
Entirely new mapping system. I've been personally chipping away at it on a branch but it's nowhere near production ready.

Menshin:
> Entirely new mapping system. I've been personally chipping away at it on a branch but it's nowhere near production ready.

I see, exciting news nonetheless 👍 
Thanks for the feedback.

PR #6441, opened 2026-03-09
Fix VRAM leak in MapChunk edge rendering

B-Kirill:
When deleting a map chunk, the `DeleteChunk` method in `Clyde.GridRendering.cs` frees the main `VAO`/`VBO`/`EBO` but does not free `EdgeVAO`/ `EdgeVBO`/`EdgeEBO` which leads to a VRAM leak.
Just added the deletion of these objects.


moonheart08 (changes requested):
Keep the GL error checks, so we actually know which operation errored out.

B-Kirill:
> Keep the GL error checks, so we actually know which operation errored out.

Oh, I didn't think about it in that context. Then yes, it should stay.

PJB3005 (approved):

PR #6444, opened 2026-03-09
Fix ApplyLinearImpulse treating the impulse as being both in world and local space

SlamBamActionman:
Short and simple fix for a function used in literally one place (soon two) in Content. 

In short, `SetLinearVelocity` uses impulse as it's in world space, while `SetAngularVelocity` treats it as if it is in local space. Cue incorrect angular velocity being applied. This PR changes the local positions into being in world space.


slarticodefast:
Here a PR that showed the bug in action (see Ed's video below) and can be used for testing.
https://github.com/space-wizards/space-station-14/pull/32415

PR #6451, opened 2026-03-13
allow hiding scroll bars in the scrollbox 

JohnJJohn:
Allows the H and V scrollbars to be hidden in a scrollbox container but still allow scrolling to happen

Allows for peak like this
https://github.com/user-attachments/assets/98eee545-ed6f-4c8f-924a-f7342b9c9b49

### DIFF
--- a/Robust.Client/UserInterface/Controls/ScrollContainer.cs
+++ b/Robust.Client/UserInterface/Controls/ScrollContainer.cs
@@ -12,8 +12,8 @@ namespace Robust.Client.UserInterface.Controls
         private bool _vScrollEnabled = true;
         private bool _hScrollEnabled = true;
 
-        private bool _vScrollBarHidden = false;
-        private bool _hScrollBarHidden = false;
+        private bool? _vScrollBarHidden = null;
+        private bool? _hScrollBarHidden = null;
 
         private bool _vScrollVisible;
         private bool _hScrollVisible;
@@ -120,7 +120,7 @@ namespace Robust.Client.UserInterface.Controls
             }
         }
 
-        public bool VScrollBarHidden
+        public bool? VScrollBarHidden
         {
             get => _vScrollBarHidden;
             set
@@ -130,7 +130,7 @@ namespace Robust.Client.UserInterface.Controls
             }
         }
 
-        public bool HScrollBarHidden
+        public bool? HScrollBarHidden
         {
             get => _hScrollBarHidden;
             set
@@ -139,6 +139,12 @@ namespace Robust.Client.UserInterface.Controls
                 InvalidateArrange();
             }
         }
+
+        private bool StyleVScrollBarHidden =>
+            _vScrollBarHidden ?? TryGetStyleProperty<bool>(nameof(VScrollBarHidden), out var v) && v;
+
+        private bool StyleHScrollBarHidden =>
+            _hScrollBarHidden ?? TryGetStyleProperty<bool>(nameof(HScrollBarHidden), out var v) && v;
 
         protected override Vector2 MeasureOverride(Vector2 availableSize)
         {
@@ -222,7 +228,7 @@ namespace Robust.Client.UserInterface.Controls
 
                 if (sWidth < cWidth && _hScrollEnabled && !MathHelper.CloseTo(sWidth, cWidth, 1e-3))
                 {
-                    _hScrollBar.Visible = _hScrollVisible = !_hScrollBarHidden;
+                    _hScrollBar.Visible = _hScrollVisible = !StyleHScrollBarHidden;
                     _hScrollBar.Page = sWidth;
                     _hScrollBar.MaxValue = cWidth;
                 }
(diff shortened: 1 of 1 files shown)


moonheart08 (changes requested):
Needs to support being set via stylesheet. Make the backing fields nullable (and null by default) and if they're null look up a corresponding stylesheet property.

The property names should be `nameof(VScrollBarHidden)` and `nameof(HScrollBarHidden)`

Robust.Client/UserInterface/Controls/ScrollContainer.cs
```diff
@@ -120,7 +120,7 @@ public bool HScrollEnabled
             }
         }
 
-        public bool VScrollBarHidden
+        public bool? VScrollBarHidden
```
moonheart08, line 134: Doc comments.

Robust.Client/UserInterface/Controls/ScrollContainer.cs
```diff
@@ -130,7 +130,7 @@ public bool VScrollBarHidden
             }
         }
 
-        public bool HScrollBarHidden
+        public bool? HScrollBarHidden
```
moonheart08, line 155: Ditto.

moonheart08 (changes requested):

Robust.Client/UserInterface/Controls/ScrollContainer.cs
moonheart08: As in, [xml doc comments](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/).

moonheart08 (changes requested):
[Descriptive XMLDoc comments, please.](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/)

juliangiebel:
Don't react like that when people tell you to use the correct doc comment format

Robust.Client/UserInterface/Controls/ScrollContainer.cs
```diff
@@ -117,6 +120,54 @@ public bool HScrollEnabled
...
+        /// Controls the visibility of the vertical scroll bar, overriding the default visibility behavior.
+        /// When <see langword="true"/>, the vertical scroll bar will never be shown, even if the content
+        /// overflows vertically. When <see langword="false"/>, the scroll bar is always shown if vertical
+        /// scrolling is enabled. When <see langword="null"/>, visibility falls back to the style property
+        /// <c>VScrollBarHidden</c>, or defaults to automatic behavior based on content size.
+        /// </summary>
+        /// <remarks>
+        /// This does not disable vertical scrolling — see <see cref="VScrollEnabled"/> for that.
+        /// Changing this value will trigger a layout re-arrange via <see cref="Control.InvalidateArrange"/>.
+        /// </remarks>
```
moonheart08:
```suggestion
        /// <summary>
        /// <para>
        ///     Prevents the vertical scrollbar from being visible when <see langword="true"/>, and allows it to be visible when <see langword="false"/>.
        ///     When <see langword="null"/>, falls back to the value of the style property instead.
        /// </para>
        /// <para>
        ///    This is an override for the <see langword="bool"/> style property of the same name. Said style property mirrors the behavior of this property, and defaults to <see langword="false"/>.
        /// </para>
        /// </summary>
        /// <seealso cref="VScrollEnabled"/>
        /// <seealso cref="HScrollBarHidden"/>
```

Robust.Client/UserInterface/Controls/ScrollContainer.cs
```diff
@@ -117,6 +120,54 @@ public bool HScrollEnabled
...
+        /// Controls the visibility of the horizontal scroll bar, overriding the default visibility behavior.
+        /// When <see langword="true"/>, the horizontal scroll bar will never be shown, even if the content
+        /// overflows horizontally. When <see langword="false"/>, the scroll bar is always shown if horizontal
+        /// scrolling is enabled. When <see langword="null"/>, visibility falls back to the style property
+        /// <c>HScrollBarHidden</c>, or defaults to automatic behavior based on content size.
+        /// </summary>
+        /// <remarks>
+        /// This does not disable horizontal scrolling — see <see cref="HScrollEnabled"/> for that.
+        /// Changing this value will trigger a layout re-arrange via <see cref="Control.InvalidateArrange"/>.
+        /// </remarks>
```
moonheart08:
```suggestion
        /// <summary>
        /// <para>
        ///     Prevents the horizontal scrollbar from being visible when <see langword="true"/>, and allows it to be visible when <see langword="false"/>.
        ///     When <see langword="null"/>, falls back to the value of the style property instead.
        /// </para>
        /// <para>
        ///    This is an override for the <see langword="bool"/> style property of the same name. Said style property mirrors the behavior of this property, and defaults to <see langword="false"/>.
        /// </para>
        /// </summary>
        /// <seealso cref="HScrollEnabled"/>
        /// <seealso cref="VScrollBarHidden"/>
```

moonheart08 (changes requested):

moonheart08 (approved):

juliangiebel (approved):

PR #6452, opened 2026-03-13
improve ClientGameStateManager error logs

mastermiller01:
Added the exception to the error message, so if integration tests fail the exception isn't lost to the boundless sea.

### DIFF
--- a/Robust.Client/GameStates/ClientGameStateManager.cs
+++ b/Robust.Client/GameStates/ClientGameStateManager.cs
@@ -942,7 +942,7 @@ namespace Robust.Client.GameStates
                 }
                 catch (Exception e)
                 {
-                    _sawmill.Error($"Caught exception while updating entity broadphases");
+                    _sawmill.Error($"Caught exception while updating entity broadphases: {e}");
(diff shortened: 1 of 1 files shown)


moonheart08 (changes requested):
How does this help, exactly?
- Please do not web-edit the game engine.
- This already logs the error to IRuntimeLog, which shows up in test logs as it itself logs an error to the console.

mastermiller01:
It does not show up as the Log.Error call throws an exception in the test's log handler, so the actual exception never gets stored anywhere.

moonheart08:
> It does not show up as the Log.Error call throws an exception in the test's log handler, so the actual exception never gets stored anywhere.

Remove the log entirely and allow IRuntimeLog to do the printing, then. Change the catcher string to mention this is while deleting entities.

mastermiller01:
I'm not sure what to change exactly, should I remove the Log.Error lines altogether?

moonheart08:
> I'm not sure what to change exactly, should I remove the Log.Error lines altogether?

Correct, change the IRuntimeLog invocation to mention the origin more clearly instead.

moonheart08 (approved):

Tayrtahn (approved):

mastermiller01:
@Tayrtahn please rerun the failed test, it looks unrelated.

PR #6456, opened 2026-03-15
fix error spam for AutoNetworkedField with deleted entities

mastermiller01:
Don't log missing MetaDataComponent for GetNetEntity as 1000 things in content spam errors from it, sometimes being unfeasible to track deletions properly everywhere.
AutoGenerateComponentState and AutoNetworkedField seem too difficult to refactor to support nullable NetEntity from TryGetNetEntity which already fixed this problem, so this solution is more feasible.


moonheart08:
This change was already denied in https://github.com/space-wizards/RobustToolbox/pull/6413/

Content is broken, fix content.

Simyon264:
See above.

PR #6473, opened 2026-03-19
Add .ftl file upload support

Velken:
Added support for uploading .ftl files, by adding /Upload as a path on localization manager and making the game reload them.

Admins uploading custom .ftl must do the following format:
`/<your-uploaded-folder>/<language-code>.ftl` 
Example: `/TestUpload/en-US.ftl` 

You can have multiple .ftl files for same culture, so long they are in different folder paths (Example: `/Uploaded/1/en-US.ftl` and `/Uploaded/2/en-US.ftl` would both be loaded correctly into the game )

Solves [this](https://github.com/space-wizards/space-station-14/issues/36417) and [this](https://github.com/space-wizards/space-station-14/issues/36344) issues from content.


moonheart08:
Doesn't this mean you can only upload one ftl file per language..?

Velken:
> Doesn't this mean you can only upload one ftl file per language..?

yes, but would you need to upload multiple for admin uploads?

If you know a non-ugly way of having `/Uploaded/**/<language-code>/*` folder path, I'd prefer to do that, but I couldn't think of a non-ugly way.

moonheart08:
> > Doesn't this mean you can only upload one ftl file per language..?
> 
> yes, but would you need to upload multiple for admin uploads?
> 
> If you know a non-ugly way of having `/Uploaded/**/<language-code>/*` folder path, I'd prefer to do that, but I couldn't think of a non-ugly way.

You can upload more than one thing at a time, hence my objection.

Velken:
> > > Doesn't this mean you can only upload one ftl file per language..?
> > 
> > 
> > yes, but would you need to upload multiple for admin uploads?
> > If you know a non-ugly way of having `/Uploaded/**/<language-code>/*` folder path, I'd prefer to do that, but I couldn't think of a non-ugly way.
> 
> You can upload more than one thing at a time, hence my objection.

well, you would include all the fluent strings in the same file, also, technically, you can have multiple .ftl files, if you put each in a different subfolder `/Uploaded/1/en-US.ftl` and `/Uploaded/2/en-US.ftl` would both be loaded correctly into the game

PR #6474, opened 2026-03-20
Defer MsgPlayerList until InGame to prevent bots from scraping playerlists

Geekyhobo:
Defers sending the player list until after entering InGame in order to prevent bots from scraping player lists from servers to track players movements across servers.


moonheart08 (changes requested):
Sending the player list at all should be controllable.

Geekyhobo:
closing as better options are out there

Issue #6468, opened 2026-03-18
All [Dependency] in components must be removed.


moonheart08:
Easy to say, harder to do. Newer code *supports* these for now but long-term I'd like to simplify the inner workings of ComponentFactory and no longer inject dependencies into components under any circumstances.

All the culprits in engine:
- [ ] TransformComponent
- [ ] SpriteComponent
- [x] TimerComponent
- [x] ContainerManagerComponent
- [ ] MapGridComponent

slarticodefast:
Also `MapGridComponent`

PR #6476, opened 2026-03-22
fully remove obsolete methods in mapgridcomponent

slarticodefast:
It's been obsolete for more than 2 years, time to remove it.
Content does not use it anymore.
If someone still uses it it's an easy fix since it's just a direct replacement with the corresponding MapSystem API method.

Partially addresses #6468

### DIFF
--- a/Robust.Shared/Map/Components/MapGridComponent.cs
+++ b/Robust.Shared/Map/Components/MapGridComponent.cs
@@ -1,12 +1,9 @@
 using System;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Contracts;
 using System.Numerics;
 using Robust.Shared.GameObjects;
 using Robust.Shared.GameStates;
-using Robust.Shared.IoC;
-using Robust.Shared.Map.Enumerators;
 using Robust.Shared.Maths;
 using Robust.Shared.Physics;
 using Robust.Shared.Serialization;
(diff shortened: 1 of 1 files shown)


Robust.Shared/Map/Components/MapGridComponent.cs
```diff
@@ -15,332 +12,152 @@
...
-        [Obsolete("Use the MapSystem method")]
-        public IEnumerable<EntityUid> GetCellsInSquareArea(EntityCoordinates coords, int n)
-        {
-            return MapSystem.GetCellsInSquareArea(Owner, this, coords, n);
-        }
+/// <summary>
+/// Component marking a grid entity, which are the "floor" in space other entities can be placed on.
+/// Grids can have their own physics and move around in space while players or other entities move on them.
+/// Examples for grids are the station, shuttles or asteroids.
+/// </summary>
```
moonheart08, line 32:
```suggestion
/// <summary>
/// <para>
///     Provides a tile grid as part of an entity, with APIs to manipulate it provided on <see cref="MapSystem"/>.
///     Grids function as the optional second level in the physics hierarchy, and implicitly provide physics and the ability to collide with one-another.
/// </para>
/// <para>
///     Map grids are an unbounded 2D grid composed of <see cref="Tile"/>s, 
///     which have their properties defined by <see cref="ITileDefinition"/> inheriting prototypes.
///     A tile with the id 0 is automatically assumed to be the "empty" tile and is the default state of any tile that has not been assigned to.
/// </para>
/// <para>
///     Collision is automatically generated from set grid tiles (excluding tiles with id 0, i.e. empty).
///     Map grids have an implied additional level of physics acceleration structures,
///     functioning as an optimization for games that rely on moving grids regularly by avoiding the need to update the grids' children.
/// </para>
/// </summary>
```

This isn't content, so no content-specific documentation.

moonheart08 (changes requested):

Robust.Shared/Map/Components/MapGridComponent.cs
slarticodefast: Thanks for writing this up.
But could we at least keep some kind of simple introductory sentence to explain what a grid is and what the component is actually doing?
Because while this kind of documentation is intended for engine contributors, it is also completely useless for new contributors to content development, who won't understand what most of these words mean without diving into tons of engine code.

moonheart08: If you can think of a good introduction, sure, but it cannot be SS14-specific.

slarticodefast: Is the word "space" SS14 specific? :godo:
Because that is kinda the best way to describe it

moonheart08: > Is the word "space" SS14 specific? :godo: Because that is kinda the best way to describe it

Space for describing the wider map is fine.

PR #6483, opened 2026-03-25
Add deterministic disposal for textures in IResourceCache

B-Kirill:
Title.

This makes it much easier when you need to create a texture and dispose of it from memory at a certain point.

I didn't make the ability to dispose for other types of resources because it requires more careful implementation. Hence I added `CanBeRemoved`.

### DIFF
--- a/Robust.Client/ResourceManagement/BaseResource.cs
+++ b/Robust.Client/ResourceManagement/BaseResource.cs
@@ -8,18 +8,8 @@ namespace Robust.Client.ResourceManagement;
 /// <summary>
 ///     Base resource for the cache.
 /// </summary>
-public abstract class BaseResource : IDisposable
+public abstract class BaseResource : IBaseResource, IDisposable
 {
-    /// <summary>
-    ///     Fallback resource path if this one does not exist.
-    /// </summary>
-    public virtual ResPath? Fallback => null;
-
-    /// <summary>
-    /// Whether this resource type can be deterministically removed from the cache.
-    /// </summary>
-    public virtual bool CanBeRemoved => false;
-
     /// <summary>
     ///     Disposes this resource.
     /// </summary>
--- a/Robust.Client/ResourceManagement/BaseResource.cs
+++ b/Robust.Client/ResourceManagement/BaseResource.cs
@@ -8,8 +8,13 @@ namespace Robust.Client.ResourceManagement;
 /// <summary>
 ///     Base resource for the cache.
 /// </summary>
-public abstract class BaseResource : IBaseResource, IDisposable
+public abstract class BaseResource : IDisposable
 {
+    /// <summary>
+    ///     Fallback resource path if this one does not exist.
+    /// </summary>
+    public virtual ResPath? Fallback => null;
+
     /// <summary>
     ///     Disposes this resource.
     /// </summary>
--- a/Robust.Client/ResourceManagement/BaseResource.cs
+++ b/Robust.Client/ResourceManagement/BaseResource.cs
@@ -16,6 +16,11 @@ public abstract class BaseResource : IDisposable
     public virtual ResPath? Fallback => null;
 
     /// <summary>
+    /// Whether this resource type can be deterministically removed from the cache.
+    /// </summary>
+    public virtual bool CanBeRemoved => false;
+
+    /// <summary>
     ///     Disposes this resource.
     /// </summary>
     public virtual void Dispose()
--- a/Robust.Client/ResourceManagement/IResourceCache.cs
+++ b/Robust.Client/ResourceManagement/IResourceCache.cs
@@ -28,6 +28,10 @@ public interface IResourceCache : IResourceManager
 
     bool TryGetResource(AudioStream stream, [NotNullWhen(true)] out AudioResource? resource);
 
+    bool TryRemoveResource<T>(string path) where T : BaseResource, new();
+
+    bool TryRemoveResource<T>(ResPath path) where T : BaseResource, new();
+
     void ReloadResource<T>(string path)
         where T : BaseResource, new();
 
--- a/Robust.Client/ResourceManagement/IResourceCache.cs
+++ b/Robust.Client/ResourceManagement/IResourceCache.cs
@@ -15,10 +15,10 @@ namespace Robust.Client.ResourceManagement;
 public interface IResourceCache : IResourceManager
 {
     T GetResource<T>(string path, bool useFallback = true)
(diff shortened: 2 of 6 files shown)


Robust.Client/ResourceManagement/ResourceTypes/TextureResource.cs
```diff
@@ -127,6 +140,16 @@ public override void Reload(IDependencyCollection dependencies, ResPath path, Ca
...
 
+        public override void Dispose()
+        {
+            if (_disposed)
+                return;
+
+            _texture?.Dispose();
+            _disposed = true;
+            base.Dispose();
+        }
```
moonheart08, line 150: possible race, set disposed to true *first* then dispose the texture. Ideally also lock a `Lock` to avoid potentially racing on disposal itself.

Robust.Client/ResourceManagement/ResourceCache.cs
```diff
@@ -107,6 +106,29 @@ public bool TryGetResource(AudioStream stream, [NotNullWhen(true)] out AudioReso
...
+    {
+        var cache = GetTypeData<T>();
+
+        if (!cache.Resources.TryGetValue(path, out var resource))
+            return false;
+
+        if (!resource.CanBeRemoved)
+            return false;
+
+        if (resource.Fallback == path || (_fallbacks.TryGetValue(typeof(T), out var fallbackRes) && ReferenceEquals(fallbackRes, resource)))
```
moonheart08: *This* is extremely jank. Please try to find a better solution.

Robust.Client/ResourceManagement/BaseResource.cs
```diff
@@ -15,6 +15,11 @@ public abstract class BaseResource : IDisposable
     /// </summary>
     public virtual ResPath? Fallback => null;
 
+    /// <summary>
+    /// Whether this resource type can be deterministically removed from the cache.
+    /// </summary>
+    public virtual bool CanBeRemoved => false;
```
moonheart08: Pull this out as a static virtual property on an interface, IBaseResource.

Robust.Client/ResourceManagement/ResourceCache.cs
```diff
@@ -107,6 +106,29 @@ public bool TryGetResource(AudioStream stream, [NotNullWhen(true)] out AudioReso
...
+        => TryRemoveResource<T>(new ResPath(path));
+
+    public bool TryRemoveResource<T>(ResPath path) where T : BaseResource, new()
+    {
+        var cache = GetTypeData<T>();
+
+        if (!cache.Resources.TryGetValue(path, out var resource))
+            return false;
+
+        if (!resource.CanBeRemoved)
```
moonheart08: This is plain erroneous and should throw if encountered. See my remark about using a static virtual instead, so this can throw early and fast if you try to dispose a resource kind that doesn't support it.

moonheart08 (changes requested):

Robust.Client/ResourceManagement/ResourceTypes/TextureResource.cs
B-Kirill: > possible race, set disposed to true _first_ then dispose the texture. Ideally also lock a `Lock` to avoid potentially racing on disposal itself.

As for `true` first, it will be done.
But `Lock` seems to be unnecessary, as `TextureResource.Dispose` is only called from the main thread (`TryRemoveResource`), and there's no finalizer on `TextureResource`.

moonheart08: *We* only call it from the main thread, anyone can call it from another thread if they want to.

Robust.Client/ResourceManagement/BaseResource.cs
```diff
@@ -8,13 +8,8 @@ namespace Robust.Client.ResourceManagement;
...
 ///     Base resource for the cache.
 /// </summary>
-public abstract class BaseResource : IDisposable
+public abstract class BaseResource : IBaseResource, IDisposable
 {
-    /// <summary>
-    ///     Fallback resource path if this one does not exist.
-    /// </summary>
-    public virtual ResPath? Fallback => null;
-
```
PJB3005, line 17: This is a breaking change I'd really rather avoid.

Robust.Client/ResourceManagement/ResourceTypes/TextureResource.cs
```diff
@@ -13,12 +14,23 @@
 
 namespace Robust.Client.ResourceManagement
 {
-    public sealed class TextureResource : BaseResource
+    public sealed class TextureResource : BaseResource, IBaseResource
     {
         private OwnedTexture _texture = default!;
-        public override ResPath? Fallback => new("/Textures/noSprite.png");
+        private bool _disposed;
+        private readonly Lock _lock = new();
```
PJB3005: What is this lock for?

Robust.Client/ResourceManagement/IResourceCache.cs
```diff
@@ -15,10 +15,10 @@ namespace Robust.Client.ResourceManagement;
 public interface IResourceCache : IResourceManager
 {
     T GetResource<T>(string path, bool useFallback = true)
-        where T : BaseResource, new();
+        where T : BaseResource, IBaseResource, new();
```
PJB3005: Is specifying this additional constraint necessary for C# to allow you to access the static properties?

Robust.Client/ResourceManagement/ResourceCache.cs
```diff
@@ -107,6 +106,29 @@ public bool TryGetResource(AudioStream stream, [NotNullWhen(true)] out AudioReso
...
+
+        if (T.FallbackPath == path)
+            return false;
+
+        var cache = GetTypeData<T>();
+
+        if (!cache.Resources.TryGetValue(path, out var resource))
+            return false;
+
+        cache.Resources.Remove(path);
```
PJB3005:
```suggestion
        if (!cache.Resources.Remove(path, out var resource))
            return false;

```

PJB3005 (changes requested):

Robust.Client/ResourceManagement/IResourceCache.cs
B-Kirill: > Is specifying this additional constraint necessary for C# to allow you to access the static properties?

Yes

PJB3005 (approved):

PR #6484, opened 2026-03-25
fix ScrollContainer nullable bool logic

mastermiller01:
The logic was wrong, it mishandled the null case (default) by using `== false`. I made it use `!= true` and added comments explaining the logic, which the previous contributor forgot to do.

### DIFF
--- a/Robust.Client/UserInterface/Controls/ScrollContainer.cs
+++ b/Robust.Client/UserInterface/Controls/ScrollContainer.cs
@@ -364,15 +364,13 @@ namespace Robust.Client.UserInterface.Controls
 
             var h = _hScrollBar.Value;
             var v = _vScrollBar.Value;
-            if (!_hScrollVisible && _hScrollBarHidden == false)
-            {
+            // reset the values if everything fits, this is usually the case when the scrollbar is hidden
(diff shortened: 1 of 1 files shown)


moonheart08 (changes requested):
im tired boss. never letting people change ui without pictures of *every* usecase ever again.

Speaking of, please give evidence of this working in all scenarios (ingame)

mastermiller01:
Sorry I have more important things to do. The logic is very simple, you can just read it and think of a truth table and compare it to the comment to verify it is correct.

GeneRickUsername:
Damn. Pretty sure the fix is right though. Passing a null into an if statement does not do stuff right.

moonheart08:
I'm busy doing other engine work so proving to myself this works in-game with all cases (*visually*, not just the truth tables) is gonna be on the backburner. Hence asking.

So this just is stalled in the meantime because they didn't feel like giving me a hand and I have no reason to prioritize them.

JohnJJohn:
https://github.com/user-attachments/assets/ec6bc948-d822-41fa-875c-34d5163c0250

@moonheart08 kept you waiting huh?

moonheart08 (approved):
aight lgtm.

Tayrtahn (approved):
Thanks for adding a video!

PR #6485, opened 2026-03-26
Allow hex colour input

TriviaSolari:
Adds a text box for entering and copying hex codes for colour selectors.

## Why

In my experience this is a fairly standard feature for colour selectors to have. This also makes it less of a hassle to change colours when editing things like markings that allow for multiple colours, since now you can just copy and paste the hex code instead of having to do so for each channel.

## Details

**Resources/Locale/en-US/controls.ftl**

- New entry for the input label.

**Robust.Client/UserInterface/Controls/ColorSelectorSliders.cs**

- New private members for the text box and corresponding label
  - LineEdit horizontally expands and is centred like the sliders are. `IsValid` is purposely omitted in favour of using the current colour as a fallback, as this triggers as soon as any character in the text box is changed, which prevents you from typing in a code manually if you wanted to do that for whatever reason
  - Colour is updated when the user either presses the Enter key within the text box, or when it loses focus
- New private function `OnHexInputValueChanged`
  - Handles colour change logic, modified from `OnSliderValueChanged`
- `UpdateSliderVisuals` now also sets the new text box to the hex representation of the current colour

## Media

https://github.com/user-attachments/assets/968a64e0-96b7-4bee-bb43-df20a7940e0c

And of course, this doesn't bypass colour restrictions for species like Vox or Vulpkanin:

https://github.com/user-attachments/assets/d184dd0c-5932-46db-b3da-1fcf27272d86

And the field only appears when the HSV/RGB sliders are enabled (IIRC this is only relevant for human skin colour?)

image

### DIFF
--- a/Resources/Locale/en-US/controls.ftl
+++ b/Resources/Locale/en-US/controls.ftl
@@ -5,6 +5,7 @@ color-selector-sliders-hue = H
 color-selector-sliders-saturation = S
 color-selector-sliders-value = V
 color-selector-sliders-alpha = A
+color-selector-input-hex = Hex
 
 color-selector-sliders-rgb = RGB
 color-selector-sliders-hsv = HSV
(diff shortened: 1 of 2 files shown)


TriviaSolari:
Test fails appear to be a cargo crate sell value one, as well as something to do with stamina damage on a cleanbot, neither of which I touched in this PR.

TriviaSolari:
Realized that you could change the alpha value when you weren't supposed to when using hex codes, so I fixed that.

<details>
<summary>New videos showing fix + alpha values still working when they're allowed to be edited:</summary>

https://github.com/user-attachments/assets/93a1b3d6-8774-4e72-bd25-1ee5f97debab

https://github.com/user-attachments/assets/2ce4f1f1-b9da-4632-a384-a35526c24f1c
</details>

PJB3005 (approved):

PR #6486, opened 2026-03-26
Check that `[Virtual]` is not used on sealed/abstract/static classes

### LINKED ISSUE #6412: using [Virtual] on a sealed/uninstantiable class should be an error
found a gem in shitmed that did this, why was this not an error

it makes no sense whatsoever

sealed, abstract and static classes should error with it

Tayrtahn:
Expands the `ExplicitVirtualAnalyzer` to also check that the `[Virtual]` attribute is not used in combination with `sealed`, `abstract`, or `static`. Doing so is now reported as an error.

Resolves https://github.com/space-wizards/RobustToolbox/issues/6412.

Also slightly tweaked the analyzer to more efficiently lookup the attribute symbol and reduce code duplication by using an existing helper method to check for the attribute.

Also includes tests to verify both the new and original diagnostic rules.


Robust.Analyzers/Robust.Analyzers.csproj
```diff
@@ -38,6 +38,11 @@
     <Compile Include="..\Robust.Shared\Prototypes\PrototypeUtility.cs" LinkBase="Implementations" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!-- Needed for ExplicitVirtualAnalyzer. -->
+    <Compile Include="..\Robust.Shared\Analyzers\VirtualAttribute.cs" LinkBase="Implementations" />
+  </ItemGroup>
```
PJB3005: Am I blind? I don't see what would require this.

PJB3005 (changes requested):

PJB3005 (approved):
I fixed it.

PR #6497, opened 2026-03-29
Fix race condition in HappyEyeballs leaving extra connections alive

B-Kirill:
Title.

The issue is that after calling `successCts.CancelAsync()` in `HappyEyeballsHttp.cs`, the current code does not wait for the losing tasks to actually complete. As a result, in the cleanup loop a task may not yet have `IsCompletedSuccessfully` set. This causes `Dispose()` to be skipped and the extra connection to remain alive.
In debug build, the leftover connection triggers an assert that corrupts the connection state (which happens to me about every 10th time I launch SS14, and it's really annoying).

### DIFF
--- a/Robust.Shared/Network/HappyEyeballsHttp.cs
+++ b/Robust.Shared/Network/HappyEyeballsHttp.cs
@@ -236,13 +236,15 @@ internal static class HappyEyeballsHttp
         cancel.ThrowIfCancellationRequested();
         await successCts.CancelAsync().ConfigureAwait(false);
 
-        try
-        {
-            await Task.WhenAll(allTasks.Where(t => t != successTask)).ConfigureAwait(false);
-        }
-        catch
-        {
-        }
+        await Task.WhenAll(  
+            allTasks  
+                .Where(t => t != successTask)  
+                .Select(static t => t.ContinueWith(  
+                    static _ => { },  
+                    CancellationToken.None,  
+                    TaskContinuationOptions.ExecuteSynchronously,  
+                    TaskScheduler.Default)))  
+            .ConfigureAwait(false);
 
         if (successTask == null)
         {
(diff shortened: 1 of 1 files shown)


moonheart08 (changes requested):
Not a fan of using a try catch for this, is there not a better solution (i.e. logical cleanup)?

Robust.Shared/Network/HappyEyeballsHttp.cs
```diff
@@ -236,6 +236,16 @@ private static IPAddress[] SortInterleaved(IPAddress[] addresses)
...
         await successCts.CancelAsync().ConfigureAwait(false);
 
+        await Task.WhenAll(  
+            allTasks  
+                .Where(t => t != successTask)  
+                .Select(static t => t.ContinueWith(  
+                    static _ => { },  
+                    CancellationToken.None,  
+                    TaskContinuationOptions.ExecuteSynchronously,  
+                    TaskScheduler.Default)))
```
PJB3005: I have no idea what this is supposed to be doing.

PJB3005 (changes requested):

Robust.Shared/Network/HappyEyeballsHttp.cs
B-Kirill: > I have no idea what this is supposed to be doing.

It waits for all the losing tasks to actually complete before the cleanup loop runs. Used `ContinueWith` with an empty body so failed tasks don't throw (just in case).  
If we don't wait, `IsCompletedSuccessfully` won't be set yet and `Dispose` gets skipped:
  
```
foreach (var task in allTasks)  
{  
    if (task.IsCompletedSuccessfully && task != successTask)  
        task.Result.Dispose();  
}  
```

PJB3005: > Used `ContinueWith` with an empty body so failed tasks don't throw (just in case).

This needs to be clearly stated with a comment. And a bunch of the extra arguments you're passing to that function are (AFAICT) unecessary defaults.

Is there no more obvious way to achieve this?

B-Kirill: > Is there no more obvious way to achieve this?

Like try-catch?

B-Kirill:
Welp, tests went crazy

PJB3005 (approved):

PR #6503, opened 2026-03-30
Report all error logs on `TestPair` dispose

Tayrtahn:
Adjusts `TestPair.ReportErrorLogs` to assert each failing log from a test run when there are multiple, instead of only the first (with a generic "consult the logs" message for multiple).

### DIFF
--- a/Robust.UnitTesting/Pool/TestPair.Recycle.cs
+++ b/Robust.UnitTesting/Pool/TestPair.Recycle.cs
@@ -22,23 +22,11 @@ public partial class TestPair<TServer, TClient>
     {
         using (Assert.EnterMultipleScope())
         {
-            if (ServerLogHandler.FailingLogs.Count == 1)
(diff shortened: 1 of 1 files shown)


Robust.UnitTesting/Pool/TestPair.Recycle.cs
```diff
@@ -25,12 +25,20 @@ private void ReportErrorLogs()
             if (ServerLogHandler.FailingLogs.Count == 1)
                 Assert.Fail(ServerLogHandler.FailingLogs[0]);
             else if (ServerLogHandler.FailingLogs.Count > 1)
+            {
                 Assert.Fail("Server had multiple failing logs reported, consult the game log.");
+                foreach (var log in ServerLogHandler.FailingLogs)
+                    Assert.Fail(log);
+            }
```
moonheart08: Simplify the whole thing to just one foreach instead of any special messages.

PR #6505, opened 2026-03-31
Add a way to log toolshed commands, and vv

Geekyhobo:
Adds a way to log toolshed commands that are run on the server, and vv


moonheart08 (changes requested):
Shouldn't this be higher level i.e. at the IConsoleHost layer? The user can, in other circumstances (i.e. custom toolshed executors) just put their own stuff around their toolshed invocations themselves.

Also don't mention admin tooling specific uses in engine, engine doesn't know what that is.

moonheart08 (approved):
bodge-y but so is the entire ServerConsoleHost. rewrite eventually.

Geekyhobo:
This now also support vv logging

PJB3005 (changes requested):
I am not sure what all these mentions of `AdminLog` are supposed to do but they certainly do not belong in engine.

PR #6513, opened 2026-04-05
pishock integration

Derisionist:
relatively barebones innit

12 volt car battery adaptation soon

### DIFF
--- a/Imports/PiShock.props
+++ b/Imports/PiShock.props
@@ -1,0 +1,7 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- If you are using Robust.Client.PiShock, import this to depend on it. -->
+
+  <ItemGroup>
+    <ProjectReference Include="$(MSBuildThisFileDirectory)\..\Robust.Client.PiShock\Robust.Client.PiShock.csproj" />
+  </ItemGroup>
+</Project>
--- a/Robust.Client.PiShock/IPiShockManager.cs
+++ b/Robust.Client.PiShock/IPiShockManager.cs
@@ -1,12 +1,11 @@
 using System;
+using Robust.Shared.Analyzers;
 
-namespace Robust.Client.Utility;
+namespace Robust.Client.PiShock;
 
 [NotContentImplementable]
 public interface IPiShockManager
 {
-    void Initialize();
-
     /// <remarks>
     /// op, intensity (1–100), duration (1–15 seconds).
     /// </remarks>
--- a/Robust.Client.PiShock/PCVars.cs
+++ b/Robust.Client.PiShock/PCVars.cs
@@ -1,0 +1,35 @@
+using Robust.Shared.Configuration;
+
+namespace Robust.Client.PiShock;
+
+// ReSharper disable once InconsistentNaming
+[CVarDefs]
+public static class PCVars
+{
+    /// <summary>
+    /// Enable FUN
+    /// </summary>
+    public static readonly CVarDef<bool> PiShockEnabled =
+        CVarDef.Create("pishock.enabled", false, CVar.CLIENTONLY | CVar.ARCHIVE);
+
+    public static readonly CVarDef<string> PiShockUsername =
+        CVarDef.Create("pishock.username", string.Empty, CVar.CLIENTONLY | CVar.ARCHIVE);
+
+    public static readonly CVarDef<string> PiShockApiKey =
+        CVarDef.Create("pishock.api_key", string.Empty, CVar.CLIENTONLY | CVar.ARCHIVE | CVar.CONFIDENTIAL);
+
+    public static readonly CVarDef<string> PiShockShareCode =
+        CVarDef.Create("pishock.share_code", string.Empty, CVar.CLIENTONLY | CVar.ARCHIVE | CVar.CONFIDENTIAL);
+
+    // Clamped onto every operation as a safety cap.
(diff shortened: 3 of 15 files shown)


moonheart08:
While I get where you're coming from, this is not a feature engine can support, especially not on client. The only feasible way to integrate a device like this client-side (which falls firmly in the 'sensitive device' category your webcam, microphone, and storage go into) would be to do full launcher integration to ensure the device is never misused by a malicious server to harm the user. (which the cooldowns/etc have no effect for because the server can simply fish the client for the full auth details.)

We can't exactly *do* that right now, and there's a significantly easier route that does not require you to trust the client to securely hold authentication details for the device: just have individual servers implement the feature like they're capable of doing already.

Derisionist:
> While I get where you're coming from, this is not a feature engine can support, especially not on client. The only feasible way to integrate a device like this client-side (which falls firmly in the 'sensitive device' category your webcam, microphone, and storage go into) would be to do full launcher integration to ensure the device is never misused by a malicious server to harm the user. (which the cooldowns/etc have no effect for because the server can simply fish the client for the full auth details.)
> 
> We can't exactly _do_ that right now, and there's a significantly easier route that does not require you to trust the client to securely hold authentication details for the device: just have individual servers implement the feature like they're capable of doing already.

I'm willing to dig my heels in. What about controlling what cvars content can read/write via, say, `CONTENT_UNWRITABLE`/`CONTENT_UNREADABLE`/`ENGINE_ONLY` flags? Going the launcher direction does seem like a lot of unnecessary work (unless fully committing to allowing third-party integrations) and using the server as a de-facto proxy still gives the middleman (server operator) on-demand access.

moonheart08:
> > While I get where you're coming from, this is not a feature engine can support, especially not on client. The only feasible way to integrate a device like this client-side (which falls firmly in the 'sensitive device' category your webcam, microphone, and storage go into) would be to do full launcher integration to ensure the device is never misused by a malicious server to harm the user. (which the cooldowns/etc have no effect for because the server can simply fish the client for the full auth details.)
> > We can't exactly _do_ that right now, and there's a significantly easier route that does not require you to trust the client to securely hold authentication details for the device: just have individual servers implement the feature like they're capable of doing already.
> 
> I'm willing to dig my heels in. What about controlling what cvars content can read/write via, say, `CONTENT_UNWRITABLE`/`CONTENT_UNREADABLE`/`ENGINE_ONLY` flags? Going the launcher direction does seem like a lot of unnecessary work (unless fully committing to allowing third-party integrations) and using the server as a de-facto proxy still gives the middleman (server operator) on-demand access.

We don't want sensitive data to be in the same *process* as the game unless necessary. The user's config/etc that isn't launcher-managed all lives in content accessible places, you can't do this properly without significant launcher integration.

You're still better off doing this server-side in the interim until a full launcher-managed solution is available.

0x6273:
Couldn't this just be done in a separate program that reads data from the game client somehow? Like reading the log file maybe?

Derisionist:
> until a full launcher-managed solution is available.

The launcher is not a proprietary black box so this can be done. At bare minimum just provide an "options" field similar to how properties are set for steam games. The `--cvar` flag is already there. That's what I more or less see as necessary because how do you set the cvar other than manually editing the config file.

Doing something closer to how CEF is implemented seems to be overkill so I am not considering that direction yet.

> Couldn't this just be done in a separate program that reads data from the game client somehow? Like reading the log file maybe?

> You're still better off doing this server-side in the interim

I wouldn't trust the server operator with any free game with hardware and even more so with custom programs.

JohnJJohn:
Miz, it's not you, is it?

mastermiller01:
https://youtu.be/P0mJ3uSsE28

0x6273:
> I wouldn't trust the server operator with any free game with hardware and even more so with custom programs.

I don't see why you'd have to trust the server operator specifically? Anyone can write such a program. And thanks to the magic of open source software, anyone could check for themselves what the program does.

Derisionist:
> > I wouldn't trust the server operator with any free game with hardware and even more so with custom programs.
> 
> I don't see why you'd have to trust the server operator specifically? Anyone can write such a program. And thanks to the magic of open source software, anyone could check for themselves what the program does.

1. You can't assume every user is able to compile the program themselves
2. You can't assume the database wouldn't leak, so storing it on server is not an option
3. For external programs, you still need a trigger that comes from the game itself

0x6273:
> > > I wouldn't trust the server operator with any free game with hardware and even more so with custom programs.
> > 
> > 
> > I don't see why you'd have to trust the server operator specifically? Anyone can write such a program. And thanks to the magic of open source software, anyone could check for themselves what the program does.
> 
>     1. You can't assume every user is able to compile the program themselves

I don't see your point? Same could be said for any piece of software really.

>     2. You can't assume the database wouldn't leak, so storing it on server is not an option

Yeah, I agree that storing that stuff on the server is a bad idea.. The point of having a client-side program is this wouldn't be needed since it would be doing all the pishock api requests.

>     3. For external programs, you still need a trigger that comes from the game itself

Yeah the program could just read the triggers from the game log.

Derisionist:
> Same could be said for any piece of software really.

Shipping binaries requires trust in the vendor of said binary.

> Yeah the program could just read the triggers from the game log.

Just in - the game's log output is delayed and inconsistent from my testing. Also it feels like a cheap way out.

PJB3005:
I'd rather just have generic HTTP client support in RT (#6359), because that would enable this to be possible too afaict.

PR #6516, opened 2026-04-07
Add `SharedOnly` attribute and analyzer

### LINKED ISSUE #6506: Analyzer for NetManager.IsClient/IsServer on their respective sides.
It's strictly an error to use IsClient in a client assembly, or vice versa, as it's equivalent to `true`.

Tayrtahn:
Adds a new attribute, `SharedOnlyAttribute` that can be applied to methods, properties, and fields that should only be used in Shared assemblies. Attempting to use marked members in Client or Server assemblies will raise a warning.

This PR also applies the attribute to `INetManager.IsServer` and `INetManager.IsClient`.
Resolves https://github.com/space-wizards/RobustToolbox/issues/6506.

~~There are currently violations in Content that will need to be cleaned up before this can be merged:~~
```
Content.Client/FeedbackPopup/ClientFeedbackManager.cs(34,36): warning RA0048: IsClient should only be used in Shared assemblies [Content.Client/Content.Client.csproj]
Content.Client/FeedbackPopup/ClientFeedbackManager.cs(45,14): warning RA0048: IsClient should only be used in Shared assemblies [Content.Client/Content.Client.csproj]
Content.Server/FeedbackSystem/ServerFeedbackManager.cs(32,14): warning RA0048: IsServer should only be used in Shared assemblies [Content.Server/Content.Server.csproj]
Content.Server/FeedbackSystem/ServerFeedbackManager.cs(47,14): warning RA0048: IsServer should only be used in Shared assemblies [Content.Server/Content.Server.csproj]
Content.Server/FeedbackSystem/ServerFeedbackManager.cs(62,14): warning RA0048: IsServer should only be used in Shared assemblies [Content.Server/Content.Server.csproj]
Content.Server/FeedbackSystem/ServerFeedbackManager.cs(72,14): warning RA0048: IsServer should only be used in Shared assemblies [Content.Server/Content.Server.csproj]
Content.Server/Sound/EmitSoundSystem.cs(52,13): warning RA0048: IsClient should only be used in Shared assemblies [Content.Server/Content.Server.csproj]
```

### DIFF
--- a/Robust.Analyzers/SharedOnlyAnalyzer.cs
+++ b/Robust.Analyzers/SharedOnlyAnalyzer.cs
@@ -1,0 +1,94 @@
+#nullable enable
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+using Robust.Roslyn.Shared;
+
+namespace Robust.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class SharedOnlyAnalyzer : DiagnosticAnalyzer
+{
+    private const string SharedOnlyAttributeType = "Robust.Shared.Analyzers.SharedOnlyAttribute";
+
+    public static readonly DiagnosticDescriptor SharedOnlyRule = new(
+        Diagnostics.IdSharedOnly,
+        "Use is forbidden outside of Shared assemblies",
+        "{0} should only be used in Shared assemblies",
+        "Usage",
+        DiagnosticSeverity.Warning,
+        true
+    );
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
+    [
+        SharedOnlyRule,
+    ];
+
+    public override void Initialize(AnalysisContext context)
+    {
+        context.EnableConcurrentExecution();
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+
+        context.RegisterCompilationStartAction(ctx =>
+        {
+            // Don't do anything in Shared assemblies
+            if (ctx.Compilation.AssemblyName?.Contains("Shared") == true)
+                return;
+
+            // Get the type symbol for the attribute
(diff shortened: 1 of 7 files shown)


slarticodefast:
Shouldn't this make the tests fail if there are still violations in content? Or do they not compile in Release mode like the content tests do?

Tayrtahn:
> Shouldn't this make the tests fail if there are still violations in content? Or do they not compile in Release mode like the content tests do?

You'd think so, right? The same thing happened (or rather, didn't happen) with the `ProxyFor` analyzer PR, which also added new warnings. I think it's because the whole "treat warnings as errors" thing is part of Content's build config, which doesn't get used by the engine check that runs them here. We probably should change that so we don't get a nasty surprise in the future.

Tayrtahn:
Looking into it a little bit, I think the issue is actually that the warnings as errors option is only enabled for release builds of content, and the check against content here in engine uses a debug build (like you said!)

slarticodefast:
Can we also put this onto other attributes like `NetworkedComponent` and `AutoGenerateComponentState`? Because at the moment RT will just let you use those on the server and silently break networking for a few other random components as a result, which is basically impossible to debug unless you know already know what causes it.

Might be a little more complicated in the case of abstract shared components though.

And that should probably be an error, not a warning.

Tayrtahn:
I'd like to do that, but unfortunately this runs into our old friend "components with the same name in Client and Server but with no Shared base class".

It looks like `ThrusterComponent` is currently the only one of these that's networked (at least upstream), but we probably shouldn't break that functionality without making a formal decision and giving forks a heads up.

Robust.Analyzers/SharedOnlyAnalyzer.cs
```diff
@@ -0,0 +1,94 @@
...
+    public override void Initialize(AnalysisContext context)
+    {
+        context.EnableConcurrentExecution();
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+
+        context.RegisterCompilationStartAction(ctx =>
+        {
+            // Don't do anything in Shared assemblies
+            if (ctx.Compilation.AssemblyName?.Contains("Shared") == true)
+                return;
```
PJB3005, line 39: I'd flip this on its head so tests are still allowed to access these members.

PJB3005 (changes requested):

PR #6522, opened 2026-04-09
IHandleableEvent and ICancelableEvent Interfaces for events to use.

Princess-Cheeseballs:
## Description
Added IHandleableEvent and ICancelableEvent interfaces which are intended to be implemented by events in content and eventually engine to replace the HandleableEntityEventArgs and CancelableEntityEventArgs. 

As such those two classes have been marked obsolete

## Why?
Some systems in content care about events being handled to cancelled and there's not unified way to handle these events, this poses problems for potential construction system refactors and the like which need to handle a lot of events and check if they've been cancelled/handled/ect. or event relays which often explicitly check for handling/canceling to cut off the relay early. 

This is problematic as most new events are by ref structs due to their better performance, and they cannot inherit from these old classes. They can however implement interfaces, hence the need for such interfaces.

This PR only makes these interfaces and marks the old events obsolete as moving engine events to these interfaces would break content pretty badly, we should have at least an engine release or many with these available before we start breaking content by properly getting rid of the old args.

### DIFF
--- a/Robust.Shared/GameObjects/EntityEvents.cs
+++ b/Robust.Shared/GameObjects/EntityEvents.cs
@@ -5,14 +5,20 @@ using Robust.Shared.Serialization;
 
 namespace Robust.Shared.GameObjects
 {
-    public interface IEntityEventSubscriber { }
+    public interface IEntityEventSubscriber
+    {
+    }
 
     public delegate void EntityEventHandler<in T>(T ev);
+
     public delegate void EntityEventRefHandler<T>(ref T ev);
+
     public delegate void EntitySessionEventHandler<in T>(T msg, EntitySessionEventArgs args);
 
     [Serializable, NetSerializable]
-    public abstract class EntityEventArgs { }
+    public abstract class EntityEventArgs
+    {
+    }
 
     [Obsolete]
     [Serializable, NetSerializable]
@@ -44,47 +50,51 @@ namespace Robust.Shared.GameObjects
         public void Uncancel() => Cancelled = false;
     }
 
+    /// <summary>
+    /// An interface which allows an event to be "Handled" meaning a method has successfully responded to it in some way.
+    /// </summary>
     public interface IHandleableEvent
     {
         bool Handled { get; protected set; }
 
+        /// <summary>
+        /// Handles the event, call this when your method has correctly responded to the event implementing this interface.
+        /// </summary>
         public void Handle()
         {
             Handled = true;
         }
-
-        public void Unhandle()
-        {
-            Handled = false;
-        }
     }
 
+    /// <summary>
+    /// An interface which allows an event to be "Canceled" meaning a method has negatively responded to it in some way.
+    /// This is typically used for attempt events to prevent further code from being run.
+    /// </summary>
     public interface ICancelableEvent
     {
         bool Cancelled { get; protected set; }
 
+        /// <summary>
+        /// Cancels the event, call this when your method wishes to cancel the event implementing this interface.
+        /// </summary>
         public void Cancel()
         {
             Cancelled = true;
-        }
-
-        public void Uncancel()
-        {
-            Cancelled = false;
         }
     }
 
     /// <summary>
     /// Extension methods for Events so that inheritors of these event interfaces don't need to redefine their public methods.
+    /// These are filtered by struct and interfaces respectively because you should really be using ByRef structs for your events.
     /// </summary>
     public static class EventExtensions
     {
-        extension(IHandleableEvent ev)
+        extension<T>(T ev) where T : struct, IHandleableEvent
         {
             public void Handle() => ev.Handle();
         }
 
-        extension(ICancelableEvent ev)
+        extension<T>(T ev) where T : struct, IHandleableEvent
         {
             public void Cancel() => ev.Cancel();
         }
--- a/Robust.Shared/GameObjects/EntityEvents.cs
+++ b/Robust.Shared/GameObjects/EntityEvents.cs
@@ -94,7 +94,7 @@ namespace Robust.Shared.GameObjects
             public void Handle() => ev.Handle();
         }
 
-        extension<T>(T ev) where T : struct, IHandleableEvent
+        extension<T>(T ev) where T : struct, ICancelableEvent
         {
             public void Cancel() => ev.Cancel();
         }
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/EntityEvents.cs
```diff
@@ -41,6 +44,52 @@ public abstract class CancellableEntityEventArgs : EntityEventArgs
...
+            Cancelled = false;
+        }
+    }
+
+    /// <summary>
+    /// Extension methods for Events so that inheritors of these event interfaces don't need to redefine their public methods.
+    /// </summary>
+    public static class EventExtensions
+    {
+        extension(IHandleableEvent ev)
```
moonheart08: This should be generic over some T where T is `struct, IHandleableEvent`.

Robust.Shared/GameObjects/EntityEvents.cs
```diff
@@ -41,6 +44,52 @@ public abstract class CancellableEntityEventArgs : EntityEventArgs
...
+        {
+            public void Handle() => ev.Handle();
+        }
+
+        extension(ICancelableEvent ev)
```
moonheart08: ditto

Robust.Shared/GameObjects/EntityEvents.cs
```diff
@@ -41,6 +44,52 @@ public abstract class CancellableEntityEventArgs : EntityEventArgs
...
+
+        public void Cancel()
+        {
+            Cancelled = true;
+        }
+
+        public void Uncancel()
+        {
+            Cancelled = false;
+        }
```
moonheart08: While content uses these methods, it *shouldn't*. They're an antipattern and cases where you need them should engineer a more explicit mechanism for the precise problem they're solving.

Every occurance I can find in content is for object reuse or purely a no-op. If you want to reuse a cancellable event, have it be a `record` and use `with {}` to clone it.

moonheart08 (changes requested):

Robust.Shared/GameObjects/EntityEvents.cs
Princess-Cheeseballs: I can destroy it, currently it requires one to implement it explicitly but I suppose in that case this still isn't needed.

Robust.Shared/GameObjects/EntityEvents.cs
Princess-Cheeseballs: Why that instead specifically?

Tayrtahn: It avoids boxing/unboxing costs.

```cs
public void Foo(IBar bar) { }
```
This requires an object that can be cast to `IBar`. Since we're passing in a `struct` (a value type), we can't cast it directly and need to box it in an object (a reference type). That's slower and also allocates on the heap.


```cs
public void Foo<T>(T bar) where T : IBar { }
```
Generics are like an automated find-and-replace with the type arguments for type that gets used. So if you have pass in structs like: 
```cs
public struct Bar : IBar;
public struct Baz : IBar;
```
It'll compile those as something kinda like:
```cs
public void Foo_Bar(Bar bar) { }
public void Foo_Baz(Baz bar) { }
```
Which you'll note use the struct types directly, so there's no need to for boxing.

Robust.Shared/GameObjects/EntityEvents.cs
```diff
@@ -41,6 +50,56 @@ public abstract class CancellableEntityEventArgs : EntityEventArgs
...
+    /// These are filtered by struct and interfaces respectively because you should really be using ByRef structs for your events.
+    /// </summary>
+    public static class EventExtensions
+    {
+        extension<T>(T ev) where T : struct, IHandleableEvent
+        {
+            public void Handle() => ev.Handle();
+        }
+
+        extension<T>(T ev) where T : struct, IHandleableEvent
```
Tayrtahn:
```suggestion
        extension<T>(T ev) where T : struct, ICancelableEvent
```
I'm confused how this isn't a compiler error.

Edit: wait, is it infinitely recursive?

Princess-Cheeseballs: Seems to be infinitely recursive whoops.

Robust.Shared/GameObjects/EntityEvents.cs
```diff
@@ -1,18 +1,26 @@
...
+
     public delegate void EntitySessionEventHandler<in T>(T msg, EntitySessionEventArgs args);
 
     [Serializable, NetSerializable]
-    public abstract class EntityEventArgs { }
+    public abstract class EntityEventArgs
+    {
+    }
 
+    [Obsolete]
```
PJB3005: This needs an obsoletion reason specified

Robust.Shared/GameObjects/EntityEvents.cs
```diff
@@ -22,6 +30,7 @@ public abstract class HandledEntityEventArgs : EntityEventArgs
         public bool Handled { get; set; }
     }
 
+    [Obsolete]
```
PJB3005: This needs an obsoletion reason specified

Robust.Shared/GameObjects/EntityEvents.cs
```diff
@@ -1,18 +1,26 @@
 using System;
+using System.Reflection.Metadata;
```
PJB3005: Unused import?

Robust.Shared/GameObjects/EntityEvents.cs
```diff
@@ -41,6 +50,56 @@ public abstract class CancellableEntityEventArgs : EntityEventArgs
...
+    {
+        extension<T>(T ev) where T : struct, IHandleableEvent
+        {
+            public void Handle() => ev.Handle();
+        }
+
+        extension<T>(T ev) where T : struct, ICancelableEvent
+        {
+            public void Cancel() => ev.Cancel();
+        }
```
PJB3005: These need to take `ev` by ref. This doesn't work as written.

Robust.Shared/GameObjects/EntityEvents.cs
PJB3005, line 1: Revert all the unnecessary formatting changes in this file.

PJB3005 (changes requested):

PR #6523, opened 2026-04-09
Add exception handling to NetManager

mastermiller01:
When a client sends a malformed packet that (for any unforseeable reason, not just decryption) might throw an exception, currently it goes almost completely unhandled.
This PR changes it to catch and log the error, then kick the client and have an event for content to do admin alert, etc. with the channel after it's been kicked.

I have tested this with a well-behaved client and nothing bad happens to it.

### DIFF
--- a/Robust.Shared.IntegrationTests/GameObjects/EntityState_Tests.cs
+++ b/Robust.Shared.IntegrationTests/GameObjects/EntityState_Tests.cs
@@ -12,6 +12,7 @@ using Robust.Server.ServerStatus;
 using Robust.Shared.Asynchronous;
 using Robust.Shared.Configuration;
 using Robust.Shared.ContentPack;
+using Robust.Shared.Exceptions;
 using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
@@ -39,6 +40,7 @@ namespace Robust.UnitTesting.Shared.GameObjects
         {
             var container = new DependencyCollection();
             container.Register<ILogManager, LogManager>();
+            container.Register<IRuntimeLog, RuntimeLog>();
             container.Register<IConfigurationManager, ServerNetConfigurationManager>();
             container.Register<IConfigurationManagerInternal, ServerNetConfigurationManager>();
             container.Register<INetManager, NetManager>();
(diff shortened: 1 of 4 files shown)


mastermiller01:
I would like some help with solving the unit testing issue. Should I add a IRuntimeLog implementation for the assembly like it has for INetManager, or somehow register the Robust.Shared implementation?

moonheart08 (changes requested):
This is not a fix. Malformed packets should be rejected without exceptions being thrown and the client disconnected. Exceptions, admin alerts, logs, etc waste CPU time and give denial of service attempts extra beef.

If you or someone else has *crash logs* demonstrating a problem here, please DM it to me on discord @moonykay and I'll resolve this correctly.

mastermiller01:
This PR does not throw any exceptions.
To also use it for a DOS attack you'd have to constantly reconnect which is insanely obvious.
Blindly hoping that exceptions never happen without language support is a joke and is bound to repeat itself in the future as more things change.

JohnJJohn:
Is this gonna be merged any time soon? Both goob and trauma were getting ddossed because of this, traumas already merged this in their own robust toolbox but it would be far better if it was merged on the actual one

moonheart08:
> Is this gonna be merged any time soon? Both goob and trauma were getting ddossed because of this, traumas already merged this in their own robust toolbox but it would be far better if it was merged on the actual one

Are goob and trauma using the actual fix, or are they experiencing something new and novel they should be reporting? Because not reporting problems is precisely why I'm not a fan of this fix.

JohnJJohn:
> Are goob and trauma using the actual fix, or are they experiencing something new and novel they should be reporting? Because not reporting problems is precisely why I'm not a fan of this fix.

Trauma is using this fix and the ddos has stopped and is experiencing no abnormalities, goob doesn't have this and was getting ddossed a few days ago.....

moonheart08:
> > Are goob and trauma using the actual fix, or are they experiencing something new and novel they should be reporting? Because not reporting problems is precisely why I'm not a fan of this fix.
> 
> Trauma is using this fix and the ddos has stopped and is experiencing no abnormalities, goob doesn't have this and was getting ddossed a few days ago.....

Is trauma using this fix, or the upstream fix. Because that's my question. Obviously this works, but my point is it doesn't fix the underlying issue. The engine is not a "well it works :)" environment.

metalgearsloth:
1. We have Lidgren fixes getting merged yippee.
2. Agreed on the exception front; this is just catching them at a high-level when ideally it would be targeted. My PR still has them for the moment but it might be ideal that that gets merged first and a bigger rework around not using exceptions / handling it is done!

deltanedas:
even if you dont explicitly throw exceptions there can be things in CLR that do

Issue #6524, opened 2026-04-09
Calling both Dirty and DirtyField in the same game tick will only send the field delta


slarticodefast:
I found this one while debugging some mispredicts.
Networking the full component state should always take priority over field states, but currently the field state is prioritized for autonetworking.

For the reason this is happening look at the output of the source generator for autonetworking:

<img width="1030" height="134" alt="Image" src="https://github.com/user-attachments/assets/914745bf-e640-4dec-a386-7632d92e972a" />

If one of the fields was dirtied it will always send the field delta and never reach the code that creates a full state.
The code comment for `LastFieldUpdate` says `Track last time a field was dirtied. if the full component dirty exceeds this then we send a full state update.` but it does not seem to be doing that.

Princess-Cheeseballs:
This explains a lot.

slarticodefast:
Made a test #6525

PR #6525, opened 2026-04-09
Add test for DirtyField

slarticodefast:
Made a test to check if this is actually a bug and the reason for the mispredicts that I was seeing.
This will fail until #6524 is fixed.


slarticodefast:
Superceded by #6685

PR #6530, opened 2026-04-18
Render selected sprites after FOV

Pok27:
This PR adds the `RenderStage` field to `SpriteComponent` with `Default` and `Late` modes.

Sprites in the `Late` mode are rendered after FOV while preserving the existing draw order. Smoke uses this mode to remain visible over the FOV mask.

This will make effects like smoke look better.

before 
image

after
image


PJB3005:
> This will make effects like smoke look better.

I'm gonna be honest, that after screenshot *still* looks terrible.

mirrorcult:
yeah, im not really sure theres a noticeable difference in quality between the two there

Pok27:
The main problem with the smoke right now is the wall shadows, which I think can be turned off in the prototype. But it's still _better_
image
image

PJB3005:
I'm really not a fan of adding any additional complexity into our garbage renderer for a feature that still isn't actually acceptable.

Pok27:
@metalgearsloth Ready for review

Issue #6517, opened 2026-04-08
entities inside a container being deleted sometimes does not clean up its container


JohnJJohn:
Started happening 1-2 months ago
Entities that are deleted should be removed from the container

2026-04-08T14:28:01.3655274+02:00 [ERRO] system.status_effects: Found deleted entity 168203 inside effects container for Lavaland Planet (242/n242) while looking for WeatherAshfallLight!

![135655.png](https://github.com/user-attachments/assets/fcd91802-efe5-4431-94dd-882767445b36)

Tayrtahn:
I've been diving into this a bit, and it seems to be caused by a bug with inserting entities into containers, when the container is the map itself.

Specifically, when `SharedContainerSystem.Insert` calls `SharedTransformSystem.SetCoordinates`:
https://github.com/space-wizards/RobustToolbox/blob/3136118b5338ef2d9580178caf5c723e65eb76e7/Robust.Shared/Containers/SharedContainerSystem.Insert.cs#L108-L111

`SetCoordinates` calls `RaiseMoveEvent`, which winds up doing a grid traversal check. That check uses `IMapManager.TryFindGridAt` to check for a grid at the new coordinates, which it will then set as the new parent.

So instead of being parented to the map, the new entity winds up being parented to whatever grid happens to be located at (0, 0) on the map. `Insert` doesn't notice this, and still happily adds the entity to its internal contents list, so it still behaves as expected when interacting with the container. Later when the contained entity is deleted, it informs its parent entity of the deletion - normally, that's what tells the container to remove it from its internal list. However, since it's actually parented to the grid instead, the container doesn't get the memo and goes on blissfully unaware that one of its contained items has been deleted.

You can test this in dev easily by using `tp 0 0 1` to teleport to the map origin, then placing a few tiles to establish a grid there. If you then call `weatherset 1 WeatherAshfallLight`, you'll trip a debug assert in the container code.

moonheart08:
> I've been diving into this a bit, and it seems to be caused by a bug with inserting entities into containers, when the container is the map itself.
> 
> Specifically, when `SharedContainerSystem.Insert` calls `SharedTransformSystem.SetCoordinates`:
> 
[...]

PR #6532, opened 2026-04-18
Add parent check debug assert to `SharedContainerSystem.Insert`

Tayrtahn:
Adds an additional debug assert to `SharedContainerSystem.Insert` to verify that the inserted entity is correctly parented to the container entity.

In certain situations (see https://github.com/space-wizards/RobustToolbox/issues/6517), this can happen, and it causes some hard-to-track bugs. This extra assertion lets us catch the problem when it happens.


Tayrtahn:
Oh cool, looks like we already have tests that trigger that bug, and we didn't even know it.

PR #6538, opened 2026-04-25
Override color on sprite render

Rinary1:
Adds `colorOverride` property at render methods of `SpriteSystem.Render.cs`, so now we can override color on render to avoid things like this:

```cs
ent.sprite.Color = color;

_spriteSys.RenderSprite(ent, handle, sample.EyeRotation, sample.Rotation, sample.Position, null);

ent.sprite.Color = oldColor;
```

If we want to render sprite with another color.


PJB3005:
I feel like the semantics of this are too specific, and there's really no reason color in particular is treated as so much more important than all the other parameters on sprites.

PR #6539, opened 2026-04-25
Visibility Modifiers Layers

Pok27:
Adds `RefreshVisibilityModifiersEvent` so that systems can modify visibility layers without affecting the base layers

Required for https://github.com/space-wizards/space-station-14/pull/43706

### DIFF
--- a/Robust.Server/GameObjects/EntitySystems/VisibilitySystem.cs
+++ b/Robust.Server/GameObjects/EntitySystems/VisibilitySystem.cs
@@ -125,8 +125,7 @@ namespace Robust.Server.GameObjects
 
                 var childMask = mask;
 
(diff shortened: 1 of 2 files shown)


iaada (approved):
I think this is the best way to go about modifying the vismask. It enables us to not need to remember the original layer to reset it (fragile) when applying temporary effects, especially those defined through yaml.

PJB3005 (commented):
This seems like a really bad idea from a performance PoV? Now every time visibility has to be recalculated, it has to raise an event recursively.

metalgearsloth (commented):
Are you sure this doesn't cause pvs problems? Vismask updates alone are super expensive and having content being able to modify what is supposed to be more or less an engine flag when recursing seems super sketchy.

PR #6546, opened 2026-05-04
Add kinematic collide-and-slide helper.

rhailrake:
Adds a small `CollideAndSlide` helper on `RayCastSystem` for swept kinematic movement.
This builds on the existing shape cast path and clips the remaining translation along hit planes without changing physics solver behavior. It intentionally does not depenetrate initial overlaps, matching existing shape cast behavior.

### DIFF
--- a/Robust.Shared/Physics/Systems/RayCastSystem.Kinematic.cs
+++ b/Robust.Shared/Physics/Systems/RayCastSystem.Kinematic.cs
@@ -1,0 +1,279 @@
+using System;
+using System.Numerics;
+using Robust.Shared.Collections;
+using Robust.Shared.GameObjects;
+using Robust.Shared.Map;
+using Robust.Shared.Physics.Collision.Shapes;
+using Robust.Shared.Physics.Shapes;
+
+namespace Robust.Shared.Physics.Systems;
+
+public sealed partial class RayCastSystem
+{
+    private const int DefaultKinematicMoveIterations = 4;
+    private const float DefaultKinematicMoveSkin = 0f;
+    private const float KinematicMoveEpsilon = float.Epsilon;
+
+    private EntityQuery<FixturesComponent> _fixturesQuery;
+    private EntityQuery<TransformComponent> _xformQuery;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        _fixturesQuery = GetEntityQuery<FixturesComponent>();
+        _xformQuery = GetEntityQuery<TransformComponent>();
+    }
+
+    /// <summary>
+    /// Moves a shape through the physics broadphase and clips the remainder of the translation along hit planes.
+    /// </summary>
+    /// <remarks>
+    /// Does not depenetrate shapes that are already overlapping at <paramref name="origin"/>.
+    /// </remarks>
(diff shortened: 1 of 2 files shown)


rhailrake:
I believe this test failure is unrelated to my changes.

Robust.Shared/Physics/Systems/RayCastSystem.Kinematic.cs
```diff
@@ -0,0 +1,279 @@
...
+        {
+            PhysShapeCircle => true,
+            PolygonShape => true,
+            Polygon => true,
+            SlimPolygon => true,
+            _ => false,
+        };
+    }
+}
+
```
metalgearsloth, line 261: Won't this box them as some shapes have been ported to structs.

Robust.Shared/Physics/Systems/RayCastSystem.Kinematic.cs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    private bool TryCastShapeClosest(
+        MapId mapId,
+        IPhysShape shape,
+        Transform origin,
+        Vector2 translation,
+        QueryFilter filter,
+        out KinematicMoveHit hit)
+    {
```
metalgearsloth, line 200: Generics better as below r.e. boxing.

Robust.Shared/Physics/Systems/RayCastSystem.Kinematic.cs
```diff
@@ -0,0 +1,279 @@
...
+        hit = default;
+        var found = false;
+        Func<EntityUid, bool> isIgnored = IsIgnored;
+
+        foreach (var fixture in fixtures.Fixtures.Values)
+        {
+            if (!fixture.Hard ||
+                fixture.CollisionLayer == 0x0 ||
+                fixture.CollisionMask == 0x0 ||
+                !CanKinematicMoveCastShape(fixture.Shape))
```
metalgearsloth, line 163: This looks like a box2d port because box2d uses AND for layer / mask collisions but we still use godot's OR so are you really sure it should filter out if layer or mask is 0 atm?

Robust.Shared/Physics/Systems/RayCastSystem.Kinematic.cs
```diff
@@ -0,0 +1,279 @@
...
+using Robust.Shared.Physics.Collision.Shapes;
+using Robust.Shared.Physics.Shapes;
+
+namespace Robust.Shared.Physics.Systems;
+
+public sealed partial class RayCastSystem
+{
+    private const int DefaultKinematicMoveIterations = 4;
+    private const float DefaultKinematicMoveSkin = 0f;
+    private const float KinematicMoveEpsilon = float.Epsilon;
```
metalgearsloth, line 15: These really should not be constants as they look like they were built for single-purpose and should be CVars. All the other physics constants have purpose for being constants but even the solver iterations are configurable.

They also need documenting.

metalgearsloth:
Rest looks good fwiw.

metalgearsloth:
Box2D includes these now so I would prefer to port those to use ideally.

PR #6547, opened 2026-05-04
Fix escaped quotes in formatted message parameters.

### LINKED ISSUE #6537: There is no escaping for double quotes in formatted message attributes
- Have a string like `te"xt`
- Escape it through FormattedMessage.EscapeText
- Create a formatted message using it as an attribute, like `[tag="te"xt"]`
- Parsing throws an exception, or in the case of using permissive parsing, just displays the tag as text without parsing it, it does not matter how many slashes you put before the double quote

rhailrake:
Fixes #6537.

Quoted string parameters now parse escaped characters correctly.
This makes markup like `[tag="te\"xt"]` parse normally instead of failing or falling back to text in permissive parsing.

String parameter serialization now escapes quotes only inside parameter values, so `ToMarkup()` round-trips quoted parameter values correctly without changing plain text escaping.

### DIFF
--- a/Robust.Shared/Utility/MarkupNode.cs
+++ b/Robust.Shared/Utility/MarkupNode.cs
@@ -137,7 +137,7 @@ public readonly record struct MarkupParameter(string? StringValue = null, long? 
     public override string ToString()
     {
         if (StringValue != null)
-            return $"=\"{FormattedMessage.EscapeText(StringValue).Replace("\"", "\\\"")}\"";
+            return $"=\"{FormattedMessage.EscapeStringParameter(StringValue)}\"";
 
         if (LongValue.HasValue)
             return LongValue?.ToString().Insert(0, "=") ?? "";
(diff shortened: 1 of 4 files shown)


PJB3005 (commented):
I think I'd prefer if `"` was only escaped when inside a parameter, instead of in all cases.

rhailrake:
> I think I'd prefer if `"` was only escaped when inside a parameter, instead of in all cases.

Sure! Adjusted this so `\"` is only treated as an escape inside quoted string parameters. `EscapeText` no longer escapes plain text quotes, and I added coverage for both plain text quotes and quoted parameter round-tripping.

Robust.Shared/Utility/MarkupNode.cs
```diff
@@ -139,7 +137,7 @@ public bool TryGetColor([NotNullWhen(true)] out Color? value)
     public override string ToString()
     {
         if (StringValue != null)
-            return $"=\"{StringValue}\"";
+            return $"=\"{FormattedMessage.EscapeText(StringValue).Replace("\"", "\\\"")}\"";
```
PJB3005: This should have a dedicated escape function.

PJB3005 (changes requested):

PJB3005 (approved):

PR #6548, opened 2026-05-05
Drop failed runtime prototype uploads.

### LINKED ISSUE #6502: One bad prototype can cause a lot of issues
Title. If an admin uploads one bad prototype, all other uploaded prototypes will fail to serialize correctly for anyone new who connects or rejoins.

This can cause massive PVS issues that can only be resolved through a server restart, ideally these prototypes should just be dropped with an error if the server cannot parse them when an upload is attempted.

rhailrake:
Fixes #6502.

Runtime prototype uploads now fail as a transaction: bad uploads are dropped before being added to replay state or broadcast to clients.

If loading an upload partially mutates prototype state before failing, the prototype manager is reset and previously accepted uploads are reapplied. Strict reload validation is used for uploads so validation-only failures are rejected too, while normal prototype reload behavior remains unchanged.


Robust.Shared/Upload/SharedPrototypeLoadManager.cs
```diff
@@ -36,19 +36,50 @@ public virtual void Initialize()
...
-        _localizationManager.ReloadLocalizations();
+    protected bool TryLoadPrototypeData(string data)
+    {
+        try
+        {
+            LoadPrototypeData(data);
+        }
+        catch (Exception e)
+        {
+            _sawmill.Error($"Failed to load adminbus prototype data. Dropping upload.\n{e}");
```
metalgearsloth:
```suggestion
        {
            _sawmill.Error($"Failed to load prototype data. Dropping upload.\n{e}");
```

Robust.Shared/Upload/SharedPrototypeLoadManager.cs
```diff
@@ -36,19 +36,50 @@ public virtual void Initialize()
...
+    {
+        try
+        {
+            _prototypeManager.Reset();
+            if (LoadedPrototypes.Count != 0)
+                LoadPrototypeData(string.Join("\n\n", LoadedPrototypes));
+        }
+        catch (Exception e)
+        {
+            _sawmill.Error($"Failed to reload accepted adminbus prototype data.\n{e}");
```
metalgearsloth:
```suggestion
        {
            _sawmill.Error($"Failed to reload accepted prototype data.\n{e}");
```

Robust.Shared/Prototypes/PrototypeManager.cs
```diff
@@ -474,6 +498,31 @@ void AddToQueue(string id)
...
+            var errors = _serializationManager.ValidateNode(kind, validationMapping, context)
+                .GetErrors()
+                .ToArray();
+
+            if (errors.Length == 0)
+                return;
+
+            var errorText = string.Join("\n", errors.Select(x => x.ErrorReason));
+            throw new PrototypeLoadException($"Validation failed for {kind}({id})\n{errorText}");
+        }
```
metalgearsloth, line 525: I need to inspect this in more detail.

rhailrake: No problem. I also added a regression case for a partially valid upload to verify that the previously accepted prototype state is restored after a failure.

metalgearsloth: I think smug ended up doing something to remove types for perf reasons so I'll doublecheck where I saw it.

SlamBamActionman:
Heya, this PR seems to have broken a few prototype uploads that utilize components with Client-only fields. You can check more information here: https://github.com/space-wizards/RobustToolbox/issues/7061

It may also be related to an issue where certain prototypes are unable to reference entity prototypes in the same prototype upload file, acting as if they don't exist (despite being uploaded simultaneously).

PR #6549, opened 2026-05-05
[Dependency] source generator

PJB3005:
No more reflection, no more codegen at runtime

Also various changes to Roslyn helpers to make this easier to write.

Requires all types with dependencies to be partial and not have readonly dependency fields. An analyzer enforces this at warning level, the previous injection strategies have remained in the code *for now* as a fallback.

No fallback is available for [field: Dependency] properties, due to a Roslyn bug.

Code Fixes exist. We love Roslyn


VerinSenpai:
It's.. it's beautiful 😭

Robust.Analyzers/Generators/HasDependenciesGenerator.cs
```diff
@@ -0,0 +1,227 @@
...
+
+                typeInfo.WriteFooter(ref sb);
+
+                productionContext.AddSource(typeInfo.GetGeneratedFileName(), sb.ToString());
+            });
+    }
+
+    private static void WriteGetDependencies(ref IndentWriter sb, EquatableArray<FieldInfo> fields, bool isOverride)
+    {
+        sb.AppendLineIndented("return [");
```
moonheart08: I'm not sure the use of the new list initializer syntax is actually sandbox safe and until we *guarantee* that we should just avoid it.

Robust.Analyzers/HasDependenciesAnalyzer.cs
```diff
@@ -0,0 +1,144 @@
...
+        "Type has dependencies but is not partial",
+        "Type '{0}' has [Dependency] fields but is not partial. This will be required in the future.",
+        "Usage",
+        DiagnosticSeverity.Warning,
+        true);
+
+    public static readonly DiagnosticDescriptor DiagnosticNotPartialParent = new(
+        Diagnostics.IdHasDependenciesNotPartialParent,
+        "Type has dependencies but is not in a partial type",
+        "Type '{0}' has [Dependency] fields but is nested in a non-partial type. This will be illegal in the future.",
```
moonheart08:
```suggestion
        "Type '{0}' has [Dependency] fields but is nested in a non-partial type. The parent being partial will be required in the future.",
```

Robust.Shared/IoC/IHasDependencies.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Robust.Shared.IoC;
+
+/// <summary>
+/// Declares that a type has dependencies that can be injected via <see cref="IDependencyCollection"/>.
+/// </summary>
+/// <remarks>
+/// <para>
+/// This type should only be used by engine.
+/// The API may change in the future so content should not implement or use it itself.
+/// Automatic implementation via the source generator is the only supported method.
```
moonheart08:
```suggestion
/// Automatic implementation via the source generator is the only supported method, this is not ABI stable.
```

moonheart08 (dismissed):

notquitehadouken:
hooray, no more hitting f12 on something and being dragged down to the depths of autogenerated code

PR #6551, opened 2026-05-05
Fix/sensor logic

bababui328:
In the Goob build, there is an interactor, but it cannot interact with anchored objects.

This happens because the interactor’s collider is also anchored, meaning both objects are static and have awake = false, which is required for static objects. As a result, contacts between them are not processed. Additionally, when an object becomes anchored, all existing contacts are destroyed.

The main solution to this issue was to add a condition that still allows contact calculation when both objects are anchored, as long as one of them is a sensor (hard = false). In that case, the contact is still processed.


bababui328:
> In the Goob build, there is an interactor, but it cannot interact with anchored objects.
> 
> This happens because the interactor’s collider is also anchored, meaning both objects are static and have awake = false, which is required for static objects. As a result, contacts between them are not processed. Additionally, when an object becomes anchored, all existing contacts are destroyed.
> 
> The main solution to this issue was to add a condition that still allows contact calculation when both objects are anchored, as long as one of them is a sensor (hard = false). In that case, the contact is still processed.

Actually, the changes were made only in SharedPhysycSystem.Contact.cs, from line 487 to line 496, and in SharedBoardphaseSystem.cs, from line 512. The changes in these parts of the code should fix the issue with sensor behavior.

PJB3005:
This PR has completely broken diffs and cannot be reviewed at all.

PR #6553, opened 2026-05-05
Analyzer to detect nullable dependency fields

PJB3005:


Robust.Analyzers/DependencyNullableAnalyzer.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    public override void Initialize(AnalysisContext context)
+    {
+        context.EnableConcurrentExecution();
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+        context.RegisterCompilationStartAction(static ctx =>
+        {
+            var attr = ctx.Compilation.GetTypeByMetadataName(DependencyAttributeType);
+            if (attr == null)
+                return;
```
Tayrtahn, line 32:
```suggestion ;
            if (ctx.Compilation.GetTypeByMetadataName(DependencyAttributeType) is not { } attr)
                return;
```

Robust.Analyzers/DependencyNullableAnalyzer.cs
```diff
@@ -0,0 +1,62 @@
...
+    private static void CheckField(SymbolAnalysisContext ctx, INamedTypeSymbol attrSymbol)
+    {
+        if (ctx.Symbol is not IFieldSymbol symbol)
+            return;
+
+        if (!AttributeHelper.HasAttribute(symbol, DependencyAttributeType, out _))
+            return;
+
+        if (symbol.Type.NullableAnnotation == NullableAnnotation.Annotated)
+        {
```
Tayrtahn, line 47: Flip this around to a guard clause?

Robust.Analyzers/DependencyNullableAnalyzer.cs
PJB3005: I don't really think this is a readability improvement.

Robust.Analyzers/DependencyNullableAnalyzer.cs
PJB3005: Meh.

Tayrtahn (approved):
Yeah, this looks good.

Are there any existing violations in content? We'll need to clean them up before this can be merged.

PJB3005:
> Are there any existing violations in content? We'll need to clean them up before this can be merged.

Yes, but the analyzer is a warning so it can be immediately merged.

Tayrtahn:
Yeah, but we have TreatWarningsAsErrors in content now, so we'll need to either clean them up or add a temporary exception for this warning.

PR #6554, opened 2026-05-07
Typed Style Properties

Aeshus:
## About the PR

Added new `StylePropertyKey<T>` type to ensure style properties are correctly set to the right type, and a bunch of warnings to help enforce a better style (e.g. don't hard-code strings as StyleProperty inside of Prop<T>)

## Why

I've been doing a lot of stylesheet work recently and I've had a few times where I accidentally set a property to an invalid value, and it would just silently fail within `TryGetStyleProperty<T>`. Now, when writing stylesheet code, it'll give you better auto-completion for the value for a property as well as compile-time errors when you put an invalid value for a type.

## Technical details

The [Extract typed style properties](https://github.com/space-wizards/RobustToolbox/commit/875c547cdbaca8fc7fca437bead8b4c0bc3f8ccd) commit is the only one with substantive changes, the rest are all either style or fixing warnings.

`StylePropertyKey<T>` is really just a thin wrapper around a string with a phantom type.

It is possible to bypass the type checking using `StylePropertyKey<T>` by constructing untyped `StyleProperty` values directly, as older stylesheet code did this and removing support would cause a breaking change. The string-based key compatibility stuff produces warnings but can't provide type checking.

Memory-wise, it shouldn't be an issue as it's just a struct wrapped around a single string that it implicitly transforms between.

## Migration

It generally tends to be that most warnings are because a string is used as a key. Instead of hard-coding a string in the Stylesheet, define a `StylePropertyKey<T>` field in your `Control` class (or reuse another Control's) and use it. If you already had a string defined, just change the type.

Here's some examples of diffs to resolve some warnings in Content:

> warning CS0618: 'MutableSelector.Prop<T>(string, T)' is obsolete: 'Use Prop(StylePropertyKey<T>, T) with a typed style property key.'
```diff
            E<LineEdit>()
                .Class(LineEdit.StyleClassLineEditNotEditable)
-                .Prop("font-color", new Color(192, 192, 192)),
+                .Prop(Label.StylePropertyFontColor, new Color(192, 192, 192)),
            E<LineEdit>()
                .Pseudo(LineEdit.StylePseudoClassPlaceholder)
-                .Prop("font-color", Color.Gray),
+                .Prop(Label.StylePropertyFontColor, Color.Gray),
            E<TextEdit>()
                .Pseudo(TextEdit.StylePseudoClassPlaceholder)
-                .Prop("font-color", Color.Gray),
+                .Prop(Label.StylePropertyFontColor, Color.Gray),
```

> warning CS0618: 'Control.TryGetStyleProperty<T>(string, out T)' is obsolete: 'Use TryGetStyleProperty(StylePropertyKey<T>, out T) with a typed style property key.'
```diff
-        public const string StylePropertySeparation = "separation";
+        public static readonly StyleProperty<int> StylePropertySeparation = "separation";

        private const int DefaultSeparation = 0;

        private int ActualSeparation
        {
            get
            {
                if (TryGetStyleProperty(StylePropertySeparation, out int separation))
                {
                    return separation;
                }

                return SeparationOverride ?? DefaultSeparation;
            }
        }
```

## Breaking changes

Content (space-wizards/space-station-14) builds successfully without any changes and any new errors. The build will still succeed unless a type error was already present within the style code. 

**Changelog**

Added a new `StylePropertyKey<T>` class to be used to define style properties in controls. This provides type safety for styling, but can break builds if type errors are present.

Lots of new warnings for directly using strings instead of `StylePropertyKey<T>` as `StyleProperty` keys/names.

### DIFF
--- a/Robust.Client/UserInterface/Stylesheet.cs
+++ b/Robust.Client/UserInterface/Stylesheet.cs
@@ -98,7 +98,7 @@ namespace Robust.Client.UserInterface
     {
         private readonly List<StyleProperty> _props = new();
 
-        [Obsolete("Use Prop(StyleProperty<T>, T) with a typed style property key.")]
+        [Obsolete("Use Prop(StylePropertyKey<T>, T) with a typed style property key.")]
         public MutableSelector Prop<T>(string key, T value)
         {
             _props.Add(new StyleProperty(key, value));
@@ -321,7 +321,7 @@ namespace Robust.Client.UserInterface
             Name = name;
         }
 
-        [Obsolete("Use typed StyleProperty<T> key instead of string key.")]
+        [Obsolete("Use typed StylePropertyKey<T> instead of string key.")]
         public StyleProperty(string name, object? value)
             : this(name)
         {
--- a/Robust.Client/UserInterface/Stylesheet.cs
+++ b/Robust.Client/UserInterface/Stylesheet.cs
@@ -310,6 +310,9 @@ namespace Robust.Client.UserInterface
     [Virtual]
     public class StyleProperty
     {
+        public string Name { get; }
+        public object? Value { get; }
+
         protected StyleProperty(string name)
         {
             ArgumentNullException.ThrowIfNull(name);
@@ -329,9 +332,6 @@ namespace Robust.Client.UserInterface
             Name = property.Name;
             Value = value;
         }
-
-        public string Name { get; }
-        public object? Value { get; }
     }
 
     /// <summary>
--- a/Robust.Client/UserInterface/Stylesheet.cs
+++ b/Robust.Client/UserInterface/Stylesheet.cs
@@ -98,9 +98,16 @@ namespace Robust.Client.UserInterface
     {
         private readonly List<StyleProperty> _props = new();
 
-        public MutableSelector Prop(string key, object value)
+        [Obsolete("Use Prop(StyleProperty<T>, T) with a typed style property key.")]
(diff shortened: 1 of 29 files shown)


Robust.Client/UserInterface/Stylesheet.cs
```diff
@@ -298,18 +305,45 @@ public override string ToString()
     }
 
     /// <summary>
-    ///     A single property in a rule, with a name and an object value.
+    /// A single property in a rule, with a name and object value.
     /// </summary>
-    public readonly struct StyleProperty
+    [Virtual]
+    public class StyleProperty
```
PJB3005, line 313: Wait what? Are you re-using the type for stylesheet internals as the same one used to *define* properties?

PJB3005 (changes requested):

Robust.Client/UserInterface/Stylesheet.cs
Aeshus: Yeah I see what you mean

Aeshus: I think was trying to avoid using a struct + interface to try to turn the <T> to a generic type via inheritance (java on my brain... something something I was trying to make StyleProperty<T> so I could typecheck the manually constructed StyleProperties, but it wasn't possible), but that is just the wrong solution to this problem (given the constraints that I have since hit).

I just ended up making a special Key struct type and using an interface

PR #6555, opened 2026-05-08
fix sensor logic

bababui328:
The PR got rejected because the branch contained too many unrelated/follow-up commits, which made the diff very hard to review.
I’ll repeat what I said in the previous PR.
In the Goob build, there is an interactor, but it cannot interact with anchored objects.

This happens because the interactor’s collider is also anchored, meaning both objects are static and have awake = false, which is required for static objects. As a result, contacts between them are not processed. Additionally, when an object becomes anchored, all existing contacts are destroyed.

The main solution to this issue was to add a condition that still allows contact calculation when both objects are anchored, as long as one of them is a sensor (hard = false). In that case, the contact is still processed.

I hope this issue gets fixed. Sorry for my inexperience and for not noticing this earlier.


metalgearsloth:
To keep parity with box2d (which we should, IMO) contacts shouldn't be processed between static bodies whether sensors or not.

PR #6557, opened 2026-05-08
Test building content master against engine in release configuration

Tayrtahn:
Adds step to the `test-content` GH workflow to do a Release build of the latest content master using the PR'd changes to the engine. The existing check that builds and tests does so in debug configuration, which means that it misses potential problems that only show up in release builds.

Notably, this gives us a heads up when adding new warning types (through analyzers), since release builds use the `TreatWarningsAsErrors` option (I totally haven't [made this mistake](https://github.com/space-wizards/RobustToolbox/pull/6430)). This extra step lets us verify that content is prepared for such updates.


PJB3005 (changes requested):
Is this just a copy of `test-content.yml` that doesn't run tests? Wouldn't it be easier to just run the build twice in that workflow?

Tayrtahn:
Yeah, probably. I'll do that.

Tayrtahn:
In retrospect, this might be better as a separate action so it doesn't stop the tests from running when there's an unavoidable warning.

PR #6562, opened 2026-05-11
Add `SharedMapSystem.GetFilledTileCount`

Tayrtahn:
Content currently uses `_map.GetAllTiles(possibleTarget, comp).Count()` to get the total number of filled tiles on a grid ([here](https://github.com/space-wizards/space-station-14/blob/cff99716468bd8a8276d61ca4d095885a84b84dd/Content.Server/GameTicking/Rules/GameRuleSystem.Utility.cs#L98))

`GetAllTiles` does a 3-level loop (every chunk, every x coordinate, and every y coordinate) checking if the tile is empty and `yield return`ing each `TileRef`. `Count()` needs to iterate over every single result to get the total. But we don't need all that info; we just need the total number of filled tiles. And each `MapChunk` keeps a running tally in `FilledTiles`.

This PR adds a public method to `SharedMapSystem` that simply returns the sum of the `FilledTiles` fields in each chunk in the grid.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.Contracts;
 using System.Linq;
 using System.Numerics;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Systems/SharedMapSystem.Grid.cs
```diff
@@ -792,6 +793,18 @@ public IEnumerable<TileRef> GetAllTiles(EntityUid uid, MapGridComponent grid, bo
         }
     }
 
+    /// <summary>
+    /// Returns the total number of filled tiles on a grid
```
slarticodefast:
```suggestion
    /// Returns the total number of tiles on a grid.
```
Very minor nitpick because a grid is not really being filled, but can be extended to any size (unlike a chunk).
Looks good otherwise, and in-game test worked.

slarticodefast (approved):

PR #6565, opened 2026-05-12
BroadPhaseSystem & GridFixtureSystem clean up

TheShuEd:
file-scoping,
EntityQuery dependency injection
makes dependencies readonly 

Again, I'm not changing the underlying logic; I'm just removing outdated style and updating the code to meet modern standards

### DIFF
--- a/Robust.Client/Physics/BroadPhaseSystem.cs
+++ b/Robust.Client/Physics/BroadPhaseSystem.cs
@@ -1,13 +1,12 @@
 using Robust.Shared.Physics.Systems;
 
-namespace Robust.Client.Physics
+namespace Robust.Client.Physics;
+
+internal sealed class BroadPhaseSystem : SharedBroadphaseSystem
 {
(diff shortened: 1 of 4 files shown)


TheShuEd:
`GridSplitEvent` as a result of this change, it ended up inside the system namespace (which is correct, in theory).

But the content, of course, doesn't know this.

I don’t quite understand exactly how we should handle situations like this:

Changing the namespace in the content will also cause errors without changing the engine, and changing the engine will cause errors in the content. I can certainly create a PR with a fix for the content, but should these changes be merged simultaneously? That doesn’t feel right. How is this handled?

slarticodefast:
>  I can certainly create a PR with a fix for the content, but should these changes be merged simultaneously? That doesn’t feel right. How is this handled?

Make a content PR that fixes the namespace import and write `requires https://github.com/space-wizards/RobustToolbox/pull/6565` into the Pr description so that it uses this engine branch so that the tests can pass. It will then be merged along with the next engine release.

PR #6571, opened 2026-05-14
Add new BoxButton control

Aeshus:
## About the PR

A new `BoxButton` control, which is a `ContainerButton` that lays out its children like a `BoxContainer` without having to use an internal child BoxContainer.

## Why / Balance

A lot of buttons have trouble being used with the Parent/Child selectors as their restyle may need to go deeper than is automatically propagated.

This is also a common _enough_ pattern, as things like the `CollapsibleHeading`/`CheckBox` could be done as child classes of BoxButton rather than an extra internal `BoxContainer` (maybe both could be represented as a new `VisualButton` class that has both a Label and a TextureRect?).

Also, the `BoxContainer`'s methods are internal and thus aren't exposed to Content, meaning that a similar control can't be done in content. 

## Media

N/A

## Migration(s)

New control, N/A.

## Changes

Added new `BoxButton` control, which is a `ContainerButton` that lays out its children like a `BoxContainer` without having to use an internal child BoxContainer.

### DIFF
--- a/Robust.Client/UserInterface/Controls/BoxButton.cs
+++ b/Robust.Client/UserInterface/Controls/BoxButton.cs
@@ -54,37 +54,10 @@ public class BoxButton : ContainerButton
         var childBox = Vector2.Max(availableSize - boxSize, Vector2.Zero);
 
         var contents = Orientation == BoxContainer.LayoutOrientation.Vertical
-            ? MeasureItems<VerticalAxis>(childBox)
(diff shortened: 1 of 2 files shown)


PJB3005:
I'd really rather not add complexity like this to work around limitations in the styling engine.

Aeshus:
Yeah that's fair. I was thinking of just closing this anyways.

It would've caused a breaking change if this was used to fix anything in RT.

PR #6578, opened 2026-05-19
Fix swapped args in grid lookup

TemporalOroboros:
Noticed that these were swapped while I was poking around in MapManager. That probably shouldn't happen.

### DIFF
--- a/Robust.Shared/Map/MapManager.Queries.cs
+++ b/Robust.Shared/Map/MapManager.Queries.cs
@@ -53,7 +53,7 @@ internal partial class MapManager
     public void FindGridsIntersecting<T>(MapId mapId, T shape, Transform transform, GridCallback callback, bool approx = IMapManager.Approximate, bool includeMap = IMapManager.IncludeMap) where T : IPhysShape
     {
         if (_mapSystem.TryGetMap(mapId, out var mapEnt))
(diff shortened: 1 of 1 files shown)


Robust.Shared/Map/MapManager.Queries.cs
```diff
@@ -53,7 +53,7 @@ public void FindGridsIntersecting<T>(MapId mapId, T shape, Transform transform,
     public void FindGridsIntersecting<T>(MapId mapId, T shape, Transform transform, GridCallback callback, bool approx = IMapManager.Approximate, bool includeMap = IMapManager.IncludeMap) where T : IPhysShape
     {
         if (_mapSystem.TryGetMap(mapId, out var mapEnt))
-            FindGridsIntersecting(mapEnt.Value, shape, transform, callback, includeMap, approx);
+            FindGridsIntersecting(mapEnt.Value, shape, transform, callback, approx, includeMap);
```
metalgearsloth: Good catch, would be good to use named args here just to make it more explicit i.e. approx: approx, includeMap: includeMap

TemporalOroboros: Will do.

Tayrtahn (approved):

PR #6580, opened 2026-05-19
Removes all engine use of IMapManager

TemporalOroboros:
Converts all internal uses of IMapManager in the engine to use SharedMapSystem.
Shouldn't cause any breaking changes as I left anything publicly accessible well alone.

Includes #6579


ArtisticRoomba:
Superseded

PR #6581, opened 2026-05-19
Simplify generated code check in `AccessAnalyzer`

Tayrtahn:
`AccessAnalyzer` currently handles auto-generated code by examining the generated files normally and doing an attribute lookup on every access attempt within them to check for the `RobustAutoGeneratedAttribute`. This is slow and relies on all generated code remembering to apply the attribute.

Instead of doing all that, this PR changes the `GeneratedCodeAnalysisFlags` to `None` instead of `Analyze | ReportDiagnostics`.

On my machine, this reduced the execution time of `AccessAnalyzer` when packaging Content.Server from ~3.5 seconds (the slowest of our analyzers) to ~0.5 seconds.


metalgearsloth:
I don't understand analyzers but works on my machine, good to see you re-open it.

PR #6583, opened 2026-05-20
More Cursor Usage in Controls

Aeshus:
## About the PR

Sets the cursor on more controls, mainly being the addition of pointer cursors to button-like classes, not available cursors for disabled things, and more resize cursors for the slider/scrollbar/split container. 

I don't really have strong thoughts about this, but I think using the HResize/VResize for some controls like Slider/ScrollBar to better show the relative direction of action is better than just using a pointer. Maybe it would be better to have pointer before interaction and then switch to resize after dragging? Let me know if y'all have any thoughts.

## Why / Balance

Cursors are also a great way to show that UI is interactive than just the hover pseudoclasses, so I made sure they're used a lot.

## Media

https://github.com/user-attachments/assets/2e044a84-03d6-4a37-be10-96812c100fc0

## Migration(s)

This might cause problems if there's stuff in Content that is trying to override these cursors as the DefaultCursor stuff overrides the CustomCursor stuff. That fix is probably better for another PR and would involve breaking changes.

## Changes

Changed the cursors on interactive controls.

### DIFF
--- a/Robust.Client/UserInterface/Controls/ScrollBar.cs
+++ b/Robust.Client/UserInterface/Controls/ScrollBar.cs
@@ -130,11 +130,21 @@ namespace Robust.Client.UserInterface.Controls
 
         protected internal override void MouseMove(GUIMouseMoveEventArgs args)
         {
+            DefaultCursorShape = CursorShape.Arrow;
+
+            if (_isHovered || _grabData != null)
+            {
+                DefaultCursorShape = _orientation == OrientationMode.Horizontal
+                    ? CursorShape.HResize
+                    : CursorShape.VResize;
+            }
+
             if (_grabData == null)
             {
                 var box = _getGrabberBox();
                 _isHovered = box.Contains(args.RelativePixelPosition);
                 _updatePseudoClass();
+
                 return;
             }
 
--- a/Robust.Client/UserInterface/Controls/ScrollBar.cs
+++ b/Robust.Client/UserInterface/Controls/ScrollBar.cs
@@ -45,9 +45,7 @@ namespace Robust.Client.UserInterface.Controls
             ReservesSpace = true;
 
             _orientation = orientation;
-            DefaultCursorShape = orientation == OrientationMode.Horizontal
-                ? CursorShape.HResize
-                : CursorShape.VResize;
+            DefaultCursorShape = CursorShape.Pointer;
         }
 
         public bool IsAtEnd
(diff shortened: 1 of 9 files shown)


Robust.Client/UserInterface/Controls/ScrollBar.cs
```diff
@@ -45,6 +45,9 @@ protected ScrollBar(OrientationMode orientation)
             ReservesSpace = true;
 
             _orientation = orientation;
+            DefaultCursorShape = orientation == OrientationMode.Horizontal
+                ? CursorShape.HResize
+                : CursorShape.VResize;
```
PJB3005: I don't think a single program I know works like this.

PJB3005 (changes requested):

Robust.Client/UserInterface/Controls/Slider.cs
```diff
@@ -81,6 +81,7 @@ public StyleBox? GrabberStyleBoxOverride
         public Slider()
         {
             MouseFilter = MouseFilterMode.Stop;
+            DefaultCursorShape = CursorShape.HResize;
```
PJB3005: Ditto, weird

Robust.Client/UserInterface/Controls/ScrollBar.cs
Aeshus: I admit I was a little overzealous in some places :-)

I noted in the PR description that I thought it would be an interesting interaction maybe while dragging to help distinguish the direction of mouse motion, kinda like the custom cursor you see with browser autoscrolling features, and then leaving the DefaultCursorShape as pointer so you click and then drag (though I didn't push that version).

Though also tbh, I didn't look into how SDL does their fallbacks and it could cause issues with that when using an ostensibly resize cursor for something that isn't directly resizing (as it could fallback to a more semantically confusing cursor).

Robust.Client/UserInterface/Controls/Slider.cs
Aeshus: I was kinda thinking how some DAWS do some stuff like this when handling similar drag interactions, but I can see how this is confusing.

Aeshus: Should be addressed in 07bff39

Robust.Client/UserInterface/Controls/ScrollBar.cs
Aeshus: Addressed in 07bff39

PJB3005 (approved):

PR #6584, opened 2026-05-20
Removes IMapManager

TemporalOroboros:
Actually removes `IMapManager` from the codebase.

Includes #6580
Requires https://github.com/space-wizards/space-station-14/pull/44002

## Breaking Changes
- `IMapManager` is dead, RIP. If you were still using it all* of its content-facing functionality was ported to `SharedMapSystem` in #6579.


ArtisticRoomba (approved):
looks good to me when we're ready to do a major release

PR #6585, opened 2026-05-20
Cache dependencies in TypeSerializers

Tayrtahn:
This is part PR, part technical question: is there any reason why TypeSerializers shouldn't cache their dependencies in private fields?

`Read`, `Validate`, etc. are passed a `DependencyCollection` as a parameter, but doing the Resolve in those collections still isn't free, and presumably due to the sheer number of times those methods are called, it starts to add up.

In my basic testing, this change reduces the runtime of content tests by 30 seconds. @deltanedas (who provided me with the patch used in this PR) reports saving 3 minutes of runtime on a fork with more map testing.

I do note that (with the exception of EntitySerializer/Deserializer) TypeSerializers don't seem to use class fields. I'm wondering if there's a reason for that, or if that's just how things happen to be.


PJB3005:
This API needs to be refactored to remove the dependency collection parameter instead, if this is worth doing.

PR #6589, opened 2026-05-23
Add audio device switching API

metalgearsloth:
OpenAL-soft got updated at some point in my absence so this is back on the table.

### DIFF
--- a/Robust.Client/Audio/AudioManager.cs
+++ b/Robust.Client/Audio/AudioManager.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using OpenTK.Audio.OpenAL;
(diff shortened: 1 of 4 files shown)


Robust.Client/Audio/AudioManager.cs
```diff
@@ -143,11 +168,81 @@ private void InitializeAudio()
...
+            string.IsNullOrEmpty(requestedDevice) ? "<default>" : requestedDevice);
+
+        if (TryReopenAudioDevice(requestedDevice))
+        {
+            ReloadDeviceExtensions();
+            IsEfxSupported = HasAlDeviceExtension("ALC_EXT_EFX");
+            SetMasterGain(_cfg.GetCVar(CVars.AudioMasterVolume));
+            return;
+        }
+
```
metalgearsloth, line 200: I have no idea if hrtf needs re-applying but afaict it's context-based and not device-based but I could be wrong.

Robust.Client/Audio/AudioManager.cs
```diff
@@ -188,6 +283,31 @@ internal void LogError(string message)
...
+            if (method != null)
+                return (TDelegate?) method.MakeGenericMethod(typeof(TDelegate)).Invoke(null, [name]);
+
+            type = type.BaseType;
+        }
+
+        return null;
+    }
+
```
metalgearsloth, line 310: Evil hellcode happy to replace with whatever but OpenTK doesn't expose device switching that I could see now or when I checked a year ago.

PR #6593, opened 2026-05-25
Update .gitignore to ignore C# Dev Kit cache

slarticodefast:
Because Microsoft decided that it's a good idea to spam projects with tons of cached files that differ from system to system and are not even deterministic.

grafik

See https://github.com/github/gitignore/pull/4839


slarticodefast:
The other submodules will need the same changes to their gitignore files.
See the PRs linked above.

PR #6594, opened 2026-05-26
Change audio cvars to use InverseDistanceClamped by default

jessicamaybe:
`InverseDistanceClamped` is essentially the standard that every game uses and so it makes sense for it to be the engine default.

I also adjusted the z_offset to compensate for the change. Things should sound a lot better now.

Also see:
https://github.com/space-wizards/space-station-14/issues/5156


metalgearsloth:
> InverseDistanceClamped is essentially the standard that every game uses and so it makes sense for it to be the engine default.

We had it but at the time it sounded worse. It's also definitely not the default every game uses.

PR #6595, opened 2026-05-26
Add strict flag for keybinds to prevent key binding clashes

### LINKED ISSUE #6592: Impossible to type in ą as it deletes the whole input
Typing in ą (ctrl + alt + a) (alt gr + a) causes entire message to be deleted and replaced with the letter itself.

Issue only affecting a polish fork of the game where I've noted some people be able to do successfully write sentences with this letter and many also complaining that they can't.

Tested on wizden (commit 181a6aeff1900242e80c8dd0494869d519b86159)

OS: Windows 11

https://github.com/user-attachments/assets/fdde0d0a-078d-4342-bd6a-85f202323ab8

Elpersonn:
Adds a `strictModifiers` flag for all keybinds, making it so a keybind with this flag will work only when the specified modifier keys are pressed.

Closes https://github.com/space-wizards/RobustToolbox/issues/6592

### DIFF
--- a/Robust.Client/Input/IKeyBinding.cs
+++ b/Robust.Client/Input/IKeyBinding.cs
@@ -19,6 +19,8 @@ namespace Robust.Client.Input
         bool CanRepeat { get; }
         bool AllowSubCombs { get; }
 
+        bool Strict { get; }
+
         /// <summary>
         ///     For a <see cref="KeyBindingType.Command"/>-type binding,
         ///     whether the binding should activate if UI is focused.
--- a/Robust.Client/Input/IKeyBinding.cs
+++ b/Robust.Client/Input/IKeyBinding.cs
@@ -19,7 +19,7 @@ namespace Robust.Client.Input
         bool CanRepeat { get; }
         bool AllowSubCombs { get; }
 
-        bool Strict { get; }
+        bool StrictModifiers { get; }
 
         /// <summary>
         ///     For a <see cref="KeyBindingType.Command"/>-type binding,
--- a/Robust.Client/Input/IKeyBinding.cs
+++ b/Robust.Client/Input/IKeyBinding.cs
@@ -18,7 +18,6 @@ namespace Robust.Client.Input
         bool CanFocus { get; }
         bool CanRepeat { get; }
         bool AllowSubCombs { get; }
-
         bool StrictModifiers { get; }
 
         /// <summary>
--- a/Robust.Client/Input/InputManager.cs
+++ b/Robust.Client/Input/InputManager.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Numerics;
 using System.Runtime.InteropServices;
 using System.Text;
+using Avalonia.Data;
 using JetBrains.Annotations;
 using Robust.Client.UserInterface;
 using Robust.Shared.Collections;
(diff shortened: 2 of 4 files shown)


PJB3005 (changes requested):
You're never unsetting the flag here.

Honestly it'd probably be better to just figure out how to make the ctrl+a keybind do nothing when ctrl+alt+a is pressed. Maybe a flag to disallow the keybind to trigger if unecessary modifiers are active?

nikitosych (approved):
Merge this fucking PR

Robust.Client/Input/IKeyBinding.cs
```diff
@@ -19,6 +19,8 @@ public interface IKeyBinding
         bool CanRepeat { get; }
         bool AllowSubCombs { get; }
 
+        bool StrictModifiers { get; }
```
whatston3:
```suggestion
        bool StrictModifiers { get; }
```
It's a flag like the others, right?  Why not put them together?

Robust.Client/Input/InputManager.cs
```diff
@@ -252,7 +253,22 @@ public void KeyDown(KeyEventArgs args)
...
+                                    i != binding.Mod2 &&
+                                    i != binding.Mod3 && _keysPressed[(byte)i] == true)
+                                {
+                                    isStrict = false;
+                                    break;
+                                }
+                            }
+                            if (!isStrict)
+                                continue;
+                        }
```
whatston3, line 271: Why do this out here vs. writing another function like PackedMatchesPressedState?  You're doing something very similar, and it makes sense to me to separate the responsibilities - you wouldn't need a continue this way, so the flow becomes easier to read.

```cs
var packedMatch = binding.StrictModifiers ? PackedMatchesStrictPressedState(...) : PackedMatchesPressedState(...);
if (packedMatch)
    if ((matchedCombo == default || binding.PackedKeyCombo == matchedCombo) &&
       PackedContainsKey(binding.PackedKeyCombo, args.Key))
    {
    ...
```

Also, not sure if they expect these functions to work with any keys, if so maybe it'd be better to construct a strict key set and compare it directly with _keysPressed rather than using cherry-picked modifier keys, but I get it.

whatston3 (commented):
Seems straightforward enough.  A few thoughts, not sure how generic they would like the strict key set.

I do think it would make sense to put your main logic into a function like PackedMatchesPressedState.  The rest seems sane.

Robust.Client/Input/InputManager.cs
```diff
@@ -252,7 +253,22 @@ public void KeyDown(KeyEventArgs args)
                         PackedContainsKey(binding.PackedKeyCombo, args.Key))
                     {
                         matchedCombo = binding.PackedKeyCombo;
-
+                        if (binding.StrictModifiers)
+                        {
+                            var isStrict = true;
+                            for (Key i=Key.Control; i<Key.LSystem; i++)
```
metalgearsloth, line 259: Enum keys are not ordered like this assumes.

metalgearsloth (commented):
This will break stuff like me holding w to move up and then pressing ctrl-A to select all which I think should still be matching ctrl-A TBH.

Would it be better to just lock in keys to be treated as modifiers instead? So ctrl-A-W flags as ctrl-A but ctrl-shift-A doesn't or the likes?

PR #6596, opened 2026-05-26
Cache Component names in `CompIdx`

Tayrtahn:
Component names are now stored by `CompIdx` for fast, static lookup. This means that an `IComponentFactory` instance reference is no longer needed to do a component name lookup.

A new static `ComponentFactory.ComponentName<T>` method has been added for content to use. The non-static `IComponentFactory.GetComponentName<T>` has been marked obsolete and changed to call the static method.

Various uses of the old method in engine have been changed to use the new method, most notably `EntityPrototype`, where the single-parameter overload for `TryGetComponent<T>` is no longer marked obsolete, and the two-parameter overload (taking an `IComponentFactory` reference) now _is_ marked as obsolete. The few uses in engine have been cleaned up, but there are 90 more in content that will need (extremely easy) cleanup.

This should give a small improvement to performance, by removing the need to resolve `IComponentFactory` in a few places, and (very minimally) converting a dictionary lookup to a simple method call. 

The main benefit is the ability to use component names without needing an `IComponentFactory` instance. This simplifies test writing a bit, and makes the easiest way of writing a component check also the most performant way - there's no longer a need to locally cache component names when doing repeated lookups.

Currently, this is inefficient (because the name lookup needs to be done every iteration):
```cs
foreach (var entProto in entProtoList)
{
    if (entProto.TryGetComponent<MyComponent>(out var comp, _compFactory))
    {
        // etc.
    }
}
```
And this is more efficient, but a bit of a nuisance:
```cs
var compName = _compFactory.GetComponentName<MyComponent>();
foreach (var entProto in entProtoList)
{
    if (entProto.TryGetComponent(compName, out var comp))
    {
        // etc.
    }
}
```
But with this change, it can just be:
```cs
foreach (var entProto in entProtoList)
{
    if (entProto.TryGetComponent<MyComponent>(out var comp))
    {
        // etc.
    }
}
```

Also also, this enables content's `GameDataScrounger.EntitiesWithComponent` to be converted to a generic method instead of needing the target component's string name to be used as the parameter (which can't be validated).

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/SpriteSystem.Helpers.cs
+++ b/Robust.Client/GameObjects/EntitySystems/SpriteSystem.Helpers.cs
@@ -82,7 +82,7 @@ public sealed partial class SpriteSystem
     private IRsiStateLike GetPrototypeIconInternal(EntityPrototype prototype)
     {
         // IconComponent takes precedence. If it has a valid icon, return that. Otherwise, continue as normal.
-        if (prototype.TryGetComponent(out IconComponent? icon, _factory))
+        if (prototype.TryGetComponent(out IconComponent? icon))
             return GetIcon(icon);
 
         // If the prototype doesn't have a SpriteComponent, then there's nothing we can do but return the fallback.
@@ -108,7 +108,7 @@ public sealed partial class SpriteSystem
         var results = new List<IDirectionalTextureProvider>();
         noRot = false;
 
-        if (proto.TryGetComponent(out IconComponent? icon, _factory))
+        if (proto.TryGetComponent(out IconComponent? icon))
         {
             results.Add(GetIcon(icon));
             return results;
--- a/Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/SpriteSystem.cs
@@ -34,7 +34,6 @@ namespace Robust.Client.GameObjects
         [Dependency] private IPrototypeManager _proto = default!;
         [Dependency] private IResourceCache _resourceCache = default!;
         [Dependency] private ILogManager _logManager = default!;
-        [Dependency] private IComponentFactory _factory = default!;
 
         // Note that any new system dependencies have to be added to RobustUnitTest.BaseSetup()
         [Dependency] private SharedTransformSystem _xforms = default!;
--- a/Robust.Server/Placement/PlacementManager.cs
+++ b/Robust.Server/Placement/PlacementManager.cs
@@ -140,7 +140,7 @@ namespace Robust.Server.Placement
             {
                 // Replace existing entities if relevant.
                 if (msg.Replacement && _prototype.Index<EntityPrototype>(entityTemplateName).Components.TryGetValue(
-                        IComponentFactory.ComponentName<PlacementReplacementComponent>(), out var compRegistry))
+                        ComponentFactory.ComponentName<PlacementReplacementComponent>(), out var compRegistry))
                 {
                     var key = ((PlacementReplacementComponent)compRegistry.Component).Key;
                     var gridUid = _xformSystem.GetGrid(coordinates);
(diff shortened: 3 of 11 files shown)


PJB3005:
> This should give a small improvement to performance, by removing the need to resolve `IComponentFactory` in a few places, and (very minimally) converting a dictionary lookup to a simple method call.

This is not, at all, a correct description of what `CompIdx<T>` does. Have you actually measured the performance?

I'm personally very hesitant on this change.

PJB3005:
After further discussion I'm definitely convinced this change doesn't make much sense.

Tayrtahn:
Okay. Is the issue with the implementation or the concept?

I'm open to trying other implementations of the concept is solid - and I do think it would be useful to be able to get a component name from a component type without needing to spin up an entire game instance. The lack of type safety when using string identifiers for component types in tests just feels wrong.

Another possible solution would be making `ComponentFactory.CalculateComponentName` public instead of internal, but there's quite a bit of string manipulation, so I was looking for a faster approach.

> This is not, at all, a correct description of what `CompIdx<T>` does.

I would like to understand it better. Would you mind helping me out a bit? From what I can tell by looking at it, it uses a clever trick with a static generic class (`Store<T>`) to generate and store a unique, sequential int value for each component. The `Index<T>` method returns the value from `Store<T>`, and thanks to the magic of generic types, does so with a simple method invocation.

It's essentially the same trick used here: https://mariusgundersen.net/article/2019-type-dictionary-trick/
(Specifically this bit)



Those values are then used as indices to an array of component registrations inside `ComponentFactory` (I incorrectly identified this as a dictionary lookup before, sorry about that).

So if I'm understanding all that correctly (please correct me if I'm not), we have a handy way of getting a unique index number for any component type anywhere that the Type is known at compile time. The array that those numbers index however, only exists within a component factory instance, so the index can't be used to retrieve a value in a static context.

I figured shoving the entire registration into `CompIdx` would be nonsensical, but just adding the names there would be reasonable. Admittedly, it is a bit out of place to have within an "index". Maybe `ComponentFactory` could have a static read only array of the names, which could then be statically indexed with the values from `CompIdx`?

P.S. Please do let me know if my understanding of `CompIdx` is accurate. I'd love to add some documentation to that file for the sake of future engine maintainers.

PJB3005:
> I would like to understand it better.

.NET does not monomorphize generics for reference types. Unless (AFAIK) the call site is inlined, it is not "just a method call." The point of `CompIdx<T>` is that it *can* be inlined in most cases, in which case *the method call disappears entirely*. `CompIdx` is already a performance hack that I wrote *knowing* we would need to pay tech debt in the future. I do not like seeing it extended and I am entirely uncomfortable with the idea of component names being process-global like this.

The hack is acceptable for `CompIdx` int values because those are extremely high traffic, and aren't observeable outside ECS internals. The names? I still do not understand why we need these in such high quantities. You say "makes writing tests more convenient" yet I do not understand how passing an extra parameter or making more helpers is difficult and solves the problem?

> and I do think it would be useful to be able to get a component name from a component type without needing to spin up an entire game instance.

I do not understand what the use for this feature is.

> The lack of type safety when using string identifiers for component types in tests just feels wrong.

I do not understand how this PR fixes that. This is a completely unrelated API problem.

Tayrtahn:
Thanks for explaining that! I appreciate learning more about the innards of the engine. 

And yeah, I agree. I definitely don't want to expand our reliance on a hack. I might toss a comment into the compidx file so no one tries this kinda thing in the future. And if those calls aren't getting inlined, there's no point anyway.

Separately from all that though, I would like to get rid of the need to use string identifiers for components in tests. How would you feel about exposing `CalculateComponentName` to content for that purpose?

PJB3005:
> How would you feel about exposing `CalculateComponentName` to content for that purpose?

I still have no idea why they would need this. Why can't they pass in a `Type`?

Tayrtahn:
The use case is generating test case sources, where we want to run a test for each entity prototype that has a certain component. For example, checking every entity with a solution visuals component to make sure that sprites exist for all the fill levels. Test case data has to be accessible from a static context.

Currently, this is accomplished through GameDataScrounger, which is does a quick scan through the yaml and builds a dictionary of components -> entities. Because it's just scanning the yaml and not doing a full deserialization, it only knows the string names for components, and there's no good way to convert a name to a Type (or the reverse) in a static context.

I'm getting the feeling that the answer is to rewrite GameDataScrounger to do a real deserialization with proper type info. I got the sense that that would be too slow, but admittedly I'm going on hearsay there, and maybe improvements to the deserializer would fix that too. 

Lacking a short term solution, we'll stick to using string names in tests for the time being.

PR #6603, opened 2026-05-30
Remove NotImplementedException from DummySession.SendMessage

dffdff2423:
Hi!

My fork includes content that indirectly automatically calls this method after
some time passes. This causes issues with tests failing that should not when
our CI runs for too long. This method should not throw an exception but rather
do nothing when it is called.

Let me know if you want me to do the same for the other methods in this class
since I think it is very plausible other folks might run into similar issues.

Let me know if you guys also want a regression test and I can distill the issue down to a minimal reproducible form. Although, I think that would have to be content side if you want an exact replication of our issue.

### DIFF
--- a/Robust.Shared/Player/DummySession.cs
+++ b/Robust.Shared/Player/DummySession.cs
@@ -113,7 +113,7 @@ internal sealed class DummyChannel(DummySession session) : INetChannel
 
     public void SendMessage(NetMessage message)
     {
-        throw new NotImplementedException();
+        // Do nothing
     }
 
     public void Disconnect(string reason)


metalgearsloth:
> My fork includes content that indirectly automatically calls this method after
some time passes.

What's the test that's calling it?

dffdff2423:
It is related to our round restart vote system.

This has more details.
https://github.com/cosmatic-drift-14/cosmatic-drift/pull/896#issuecomment-4581277865 (I created an awful hack to fix it on master so you will need to check out and old version if you want to run it yourself)

It only shows up if you run the full test suite. My current theory is since the pair is reused some test eventually ticks the pair enough such that the timer for the round restart vote gets called on a later test. The restart vote ends up calling this method which causes the test to fail when it should not. A different test also fails on my hardware vs CI so I think it is hardware dependent.

This does not happen upstream since upstream does not have any automatic votes in the default test configuration

dffdff2423:
I think the content test failures below are unrelated by the way. I ran this PR against my fork and it was fine

PJB3005:
This really doesn't make sense given that all the other functions on this type are still throwing exceptions.

> My fork includes content that indirectly automatically calls this method after some time passes.

This would be fixed by changing your game code to use fake test time instead of real time. Not by changing the engine like this.

dffdff2423:
> This would be fixed by changing your game code to use fake test time instead of real time. Not by changing the engine like this.

My code is based on `SharedGameTicker::RoundDuration()` which refers to `IGameTiming` I had assumed that there was some reason why SS14 was not using test time already since this was a standard method. 

I had also assumed that since this was a dummy stub those exceptions were just being thrown since that is one of the defaults rider generates and it was safe to remove them.

Either way, I don't think I should have to pollute my content code with a test to determine weather or not we are running tests. That is exceptionally bad practice. Additionally, there is some bug in the test suite here where some information is retained between tests that should not be. I agree this might not be the best approach to fix it but I opted to open a PR to start a discussion on how best to approach a fix rather than just complain in discord.

PJB3005:
> My code is based on `SharedGameTicker::RoundDuration()` which refers to `IGameTiming` I had assumed that there was some reason why SS14 was not using test time already since this was a standard method.

Then `IGameTiming` should be fixed to not use real time for tests.

PR #6604, opened 2026-05-31
(Try to) Fix a null reference exception in ComponentTreeSystem.UpdateTreePositions

DrSmugleaf:
The error does not actually point to the correct line on local and it is one of the errors I keep seeing before a server crash on Afterlight
I have been staring at this method for an hour and I cannot see what else could possibly throw it
Also see https://github.com/space-wizards/RobustToolbox/pull/6366


metalgearsloth:
I think something upstream of this is blowing up but it's not really harmful so?

Don't think it warrants a RN.

PR #6606, opened 2026-06-01
Add AfterAppearanceChange event to AppearanceSystem

MLGTASTICa:
Adds an AfterAppearanceChange event + methods in VisualiserSystem.
Very usefull for cases where someone might want to duplicate a existing ent's appearance/layer data for rendering else where (like showing an APC entity in a mirror with the current state properly created)


slarticodefast:
Can't you just use the appearance data given in the AppearanceChangedEvent for the mirrored entity as well instead of copying sprite layers after applying the changes?

MLGTASTICa:
> Can't you just use the appearance data given in the AppearanceChangedEvent for the mirrored entity as well instead of copying sprite layers after applying the changes?

This is more generic and my use case is making item bundles that  can contain and render multiple entities at once both in world and inventoryUI

PR #6608, opened 2026-06-01
Add reusable component state caching to PVS serialization

StarlightHost:
There is not much code, so it is easier to read it and the comments.

Splitting component state construction into two separate steps, with deduplication and reuse of the serialized data.

This optimizes high pop.

### DIFF
--- a/Robust.Server/GameStates/PvsSystem.GetStates.cs
+++ b/Robust.Server/GameStates/PvsSystem.GetStates.cs
@@ -1,9 +1,12 @@
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using System.IO;
+using System.Threading;
 using Robust.Shared.GameObjects;
 using Robust.Shared.GameStates;
 using Robust.Shared.Player;
+using Robust.Shared.Serialization;
 using Robust.Shared.Timing;
 using Robust.Shared.Utility;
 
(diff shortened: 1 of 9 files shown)


StarlightHost:
I will reopen it once I test it in our build.

StarlightHost:
image
It did give us a performance gain, but not as much as I had hoped.
I will think more about the possibility of caching state between ticks if it has not changed.

StarlightHost:
Both rounds captured in the frame were nukies, one before and one after.

With 200 + online.

StarlightHost:
image
This is our low pop server, but the hit rate is still pretty good.

StarlightHost:
Turns out the UI component was not marked as user specific.
So I will hold off on putting this PR here for now.
Let us see what else we catch.

StarlightHost:
Okay, no other critical components with a missing user specific field have been found so far.
And the tests passed pretty well.

PJB3005:
What is the actual goal of this optimization? Reducing allocations? Or is the actual component state collection slow itself?

StarlightHost:
@PJB3005 Regarding your question, yes, the issue is that actual component state collection is slow.
With 270 players online, it causes spikes up to 30 ms by the end of the round, when everyone enters the shared PVS on the shuttle and CentComm, or in cases like nukies fighting a crowd in a tight area. Ghosts also contribute to this.
In general, this one modified piece of code consumes more tick time than all systems combined.

As for the PR itself, the result over several shifts is slightly below expectations. I am tuning the metrics, the caching itself, and also experimenting with caching between fromTick calls, provided that it is not a delta and the modification tick has not changed.
I will report back here about the experiments later.

For now, I do not consider it reasonable to merge this, but I would be happy to hear any additional ideas.

metalgearsloth:
If it's always faster wouldn't it be better to not have a CVar, unless it's just for testing purposes?

And given the fact this could probably cause some nightmarish hellbugs on live I'd think it probably needs a test?

Also should it end up being faster? I would've thought every player in different areas should be slower for PVS than everyone stacked as it has to check more entities. As only areas where every player is stacked AND you're sending a lot of states would be faster.

StarlightHost:
Added metrics, published to alpha.
Noticed a very interesting thing.
During spikes, when the state consumes an unreasonably large amount, the share of state builds rises to almost 95%.
I will try to optimize state collection specifically.
Will get back later with a new pr.

PJB3005:
> yes, the issue is that actual component state collection is slow.

Is it slow because of the amount of allocations or because we have to call `.GetComponentState()` at all?

Issue #4814, opened 2024-01-05
API for directly creating prototypes (with no originating YAML mapping)


moonheart08:
Prototype management itself already supports, at an API level, a prototype having no mapping. So the ability to generate additional prototypes during startup would be nice.
My current usecase for this is an improved materials system, which would generate stack prototypes and similar directly instead of requiring manual repetition.

MACMAN2003:
Another usecase for this would be procedurally generated reagents

PR #6609, opened 2026-06-02
Add methods for runtime creation and deletion of prototypes without YAML

MACMAN2003:
might fulfill #4814 probably?

Added two methods to IPrototypeManager for dynamically creating a prototype from a filled out MappingDataNode and deleting a prototype using <T> and it's id string.

Concerning the creation method, it is quite makeshift compared to what I would consider ideal but it was the simplest way that I could think of that did not involve insane levels of reflection.
Utilizing the creation method involves having to assemble a MappingDataNode and all of its DataNode children recursively, which is a little painful, but not as much as generating an entire yaml file in the code. It does involve having to use a debugger to see how the prototypes appear in memory.

The deletion method removes the provided id from the kind <T>. It's there as a way to remove prototypes if they are round-dependent OR as a safeguard against the attack of the 90000 rogue prototypes.

Both methods were tested and confirmed to work with Content.Shared.Chemistry.Reagent.ReagentPrototype


Robust.Shared/Prototypes/IPrototypeManager.cs
```diff
@@ -412,6 +412,37 @@ bool TryIndex(
...
+
+    /// <summary>
+    /// This will attempt to create a new prototype using the provided <paramref name="data"/>. <br/>
+    /// You <b>should not</b> use this under normal circumstances.
+    /// </summary>
+    /// <remarks>
+    /// To properly use this, you will need to make <paramref name="data"/> resemble how other prototypes of its kind appear in
+    /// _kinds' KindData.Results or KindData.RawResults. <br/>
+    /// You will need to utilize a debugger and a breakpoint if you want to get a reference of how the prototypes are
+    /// represented in _kinds.
```
PJB3005, line 365: Uh, what?

MACMAN2003: I can't think of a way to word it accurately and concisely.

Like if one prototype kind has a "foo" ValueDataNode and a "bar" SequenceDataNode in every instance then the MappingDataNode provided to the method needs to have both "foo" and "bar" nodes in its children.

MACMAN2003:
test fail is related

MACMAN2003:
this is required by https://github.com/AU-14/ColonialMarinesUniverse/pull/1120

PR #6610, opened 2026-06-02
Fix start of animations being inconsistent by tweaking AnimationTrackProperty.cs.

mika10095:
After trying and failing to properly fix it within C# or YAML or the animation keys themselves this 2 line engine PR is the solution.
Basically the first key of the animation is just inconsistent. It might get updated instantly in which case it looks like the fixed version or it might take a split second which causes the animation to do a harsh jump. This way the start of the animation will be on the first key which shouldnt break anything. 

# before

https://github.com/user-attachments/assets/be7ec757-cf51-439e-b073-58b976c77b08

# after
https://github.com/user-attachments/assets/167a9251-f37e-448b-8fd8-22647f449aac

### DIFF
--- a/Robust.Client.IntegrationTests/UserInterface/ControlTest.cs
+++ b/Robust.Client.IntegrationTests/UserInterface/ControlTest.cs
@@ -158,6 +158,7 @@ namespace Robust.UnitTesting.Client.UserInterface
                         Property = nameof(TestControl.Foo),
                         KeyFrames =
                         {
+                            new AnimationTrackProperty.KeyFrame(0f, 0f), // This is the common way to specify the first keyframe.
                             new AnimationTrackProperty.KeyFrame(1f, 1f),
                             new AnimationTrackProperty.KeyFrame(3f, 2f)
                         }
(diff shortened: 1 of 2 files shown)


ProPeperos:
wrong repo

mika10095:
@ProPeperos > wrong repo

Can you please explain why? I used footage from Project Carmine as thats where the problem can be easily reproduced.
Basically everywhere AnimationTrackProperty.KeyFrame() is used its used with a 0f length start key.
Like in LightFadeSystem.cs for example

> new AnimationTrackProperty.KeyFrame(Angle.Zero, 0),
                        new AnimationTrackProperty.KeyFrame(Angle.FromDegrees(120), third),
                        new AnimationTrackProperty.KeyFrame(Angle.FromDegrees(240), third),
                        new AnimationTrackProperty.KeyFrame(Angle.FromDegrees(360), third)

or ChasmFallingVisualsSytem.cs

> new AnimationTrackProperty.KeyFrame(component.OriginalScale, 0.0f),
                        new AnimationTrackProperty.KeyFrame(component.AnimationScale, length.Seconds),

Seemingly this should be the intended behaviour rather than the current one.
All this does is instantly snap it to that 0.0 second keyframe rather than needing it to update twice.

ProPeperos:
i guess i just got confused by the footage and and different PR formatting, sorry

PR #6614, opened 2026-06-03
Add PhysicsBodyStatusChangedEvent

tiders-shore:
Adds an event raised on entities after their PhysicsComponent's BodyStatus is changed.
SetBodyStatus raises it with this data:
- Physics component of the entity
- Current body status
- New body status that is about to be set

Component state handler for PhysicsComponentState also now uses SetBodyStatus instead of directly setting BodyStatus on the component, so the event is raised there too.

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
+++ b/Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
@@ -153,7 +153,7 @@ public partial class SharedPhysicsSystem
             SetSleepingAllowed(uid, component, newState.SleepingAllowed, dirty: false);
             SetFixedRotation(uid, newState.FixedRotation, body: component, dirty: false);
             SetCanCollide(uid, newState.CanCollide, body: component, dirty: false);
-            component.BodyStatus = newState.Status;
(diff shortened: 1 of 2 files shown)


Robust.Shared/Physics/Systems/SharedPhysicsSystem.Components.cs
```diff
@@ -549,7 +549,12 @@ public void SetBodyStatus(EntityUid uid, PhysicsComponent body, BodyStatus statu
         if (body.BodyStatus == status)
             return;
 
+        // Include old BodyStatus too
+        var ev = new PhysicsBodyStatusChangedEvent(body, body.BodyStatus, status);
+        RaiseLocalEvent(uid, ref ev);
+
         body.BodyStatus = status;
```
metalgearsloth: Normally for this stuff we'd raise the event after the change so any callers that check the status have the correct value.

tiders-shore: I agree with this approach, code has been changed to be as such

metalgearsloth:
top stuff mate.

PR #6615, opened 2026-06-04
Add API for getting current IClydeMonitor

metalgearsloth:
There's a way to set it but no way to get it.

Would've added a test but don't think there's a clean way to do it for runners and they use the headless stuff so.

Need this on content just for a silly options button to set display cap to refresh rate but can be useful for other stuff in future.


PowerfulBacon (approved):

Robust.Client/Graphics/IClyde.cs
```diff
@@ -37,6 +37,11 @@ public interface IClyde
         void SetWindowTitle(string title);
         void SetWindowMonitor(IClydeMonitor monitor);
 
+        /// <summary>
+        ///     Gets the monitor the main game window is currently on.
+        /// </summary>
+        IClydeMonitor? GetWindowMonitor();
```
PJB3005: Why is this API only available for the main window rather than all windows?

PJB3005 (changes requested):

PR #6618, opened 2026-06-05
Add ClydeWindow API to set borderless mode

PowerfulBacon:
Adds support for borderless windows in the Clyde window engine. A window can be set to borderless by calling the `SetWindowTitleBarVisible`.

A decision needs to be made as to whether I should rename this method to `SetWindowBorderMode` or something similar.

This is required for fancy TGUI window support in OpenDream:

image

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -39,7 +39,7 @@ END TEMPLATE-->
 
 ### New features
 
-*None yet*
+* Added the ability to set Clyde windows to borderless mode.
 
 ### Bugfixes
 
--- a/Robust.Client/Graphics/Clyde/Windowing/Sdl3.Window.cs
+++ b/Robust.Client/Graphics/Clyde/Windowing/Sdl3.Window.cs
@@ -415,6 +415,12 @@ internal partial class Clyde
             });
         }
 
+
+        private static void WinThreadWinSetTitle(CmdWinSetTitle cmd)
+        {
+                SDL.SDL_SetWindowTitle(cmd.Window, cmd.Title);
+        }
+
         public void WindowSetTitleBarVisible(WindowReg window, bool visible)
         {
             window.IsTitleBarVisible = visible;
@@ -423,11 +429,6 @@ internal partial class Clyde
                 Window = WinPtr(window),
                 Visible = visible,
             });
-        }
-
-        private static void WinThreadWinSetTitle(CmdWinSetTitle cmd)
-        {
-            SDL.SDL_SetWindowTitle(cmd.Window, cmd.Title);
         }
 
         private static void WinThreadWinSetTitleBarVisible(CmdWinSetTitleBarVisible cmd)
--- a/Robust.Client/Graphics/Clyde/Windowing/Sdl3.Window.cs
+++ b/Robust.Client/Graphics/Clyde/Windowing/Sdl3.Window.cs
@@ -418,20 +418,20 @@ internal partial class Clyde
 
         private static void WinThreadWinSetTitle(CmdWinSetTitle cmd)
         {
-                SDL.SDL_SetWindowTitle(cmd.Window, cmd.Title);
+            SDL.SDL_SetWindowTitle(cmd.Window, cmd.Title);
         }
 
-        public void WindowSetTitleBarVisible(WindowReg window, bool visible)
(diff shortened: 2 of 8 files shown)


Robust.Client/Graphics/Clyde/Windowing/Sdl3.Window.cs
```diff
@@ -415,11 +415,26 @@ public void WindowSetTitle(WindowReg window, string title)
...
         private static void WinThreadWinSetTitle(CmdWinSetTitle cmd)
         {
             SDL.SDL_SetWindowTitle(cmd.Window, cmd.Title);
         }
 
+        private static void WinThreadWinSetTitleBarVisible(CmdWinSetTitleBarVisible cmd)
+        {
+            SDL.SDL_SetWindowBordered(cmd.Window, cmd.Visible);
+        }
+
```
PowerfulBacon:
```suggestion

        private static void WinThreadWinSetTitle(CmdWinSetTitle cmd)
        {
                SDL.SDL_SetWindowTitle(cmd.Window, cmd.Title);
        }

        public void WindowSetTitleBarVisible(WindowReg window, bool visible)
        {
            window.IsTitleBarVisible = visible;
            SendCmd(new CmdWinSetTitleBarVisible
            {
                Window = WinPtr(window),
                Visible = visible,
            });
        }

        private static void WinThreadWinSetTitleBarVisible(CmdWinSetTitleBarVisible cmd)
        {
            SDL.SDL_SetWindowBordered(cmd.Window, cmd.Visible);
        }

```

ike709:
Incredibly based PR.

Robust.Client/Graphics/IClydeWindow.cs
```diff
@@ -20,6 +20,7 @@ public interface IClydeWindow : IDisposable
         bool IsFocused { get; }
         bool IsMinimized { get; }
         bool IsVisible { get; set; }
+        bool IsTitleBarVisible { get; set; }
```
PJB3005: This probably shouldn't be named this given it changes more than the title bar?

PJB3005 (changes requested):

Robust.Client/Graphics/IClydeWindow.cs
PowerfulBacon: Thanks for the review, I'll update the names of everything to be centered around the 'border' rather than the 'title bar' since thats how SDL names it.

PowerfulBacon: I have updated it to be `IsBordered` rather than `IsTitleBarVisible`, as that's the naming convention that SDL uses

PowerfulBacon:
We had a discussion on the SS14 discord, since wayland doesn't support setting the window position then we probably want to make it so that wayland windows cannot be made borderless. The question is whether or not that should be added in this PR, or in a PR that adds the ability to set window positions, since this PR's utility is a lot lower without the ability to set position at runtime.

PJB3005:
> We had a discussion on the SS14 discord, since wayland doesn't support setting the window position then we probably want to make it so that wayland windows cannot be made borderless. The question is whether or not that should be added in this PR, or in a PR that adds the ability to set window positions, since this PR's utility is a lot lower without the ability to set position at runtime.

For Wayland, a long-term solution would *probably* be to allow OD to integrate with window dragging implemented directly into the engine. Then we can re-hide the decorations.

Robust.Client/Graphics/WindowCreateParameters.cs
```diff
@@ -9,6 +9,7 @@ public sealed class WindowCreateParameters
         public string Title = "";
         public bool Maximized;
         public bool Visible = true;
+        public bool TitleBarVisible = true;
```
PJB3005, line 12: You forgot to rename this.

RELEASE-NOTES.md
```diff
@@ -39,7 +39,7 @@ END TEMPLATE-->
 
 ### New features
 
-*None yet*
+* Added the ability to set Clyde windows to borderless mode.
```
PJB3005, line 42: I'd avoid using the term "borderless" here because while it is correct, it might confuse people who are used to the "borderless fullscreen" terminology.

PJB3005 (changes requested):
I'm somewhat annoyed at the unwarranted style changes but *shrug

PR #6619, opened 2026-06-06
Fix DataRecord serialization

### LINKED ISSUE #6556: Deserializing record structs doesn't respect default arguments.
Produced on 275.2.0.

## Reproduction

In DataRecordTest.cs, add the following definitions:

```cs
    [DataRecord]
    public partial record struct OneByteOneDefaultIntRecordStruct(byte A, int B = 5);
```

```cs
    [Test]
    public void OneByteOneDefaultIntRecordStructTest()
    {
        var mapping = new MappingDataNode {{"a", "1"}};
        var val = Serialization.Read<OneByteOneDefaultIntRecordStruct>(mapping);

        Assert.Multiple(() =>
        {
            Assert.That(val.A, Is.EqualTo(1));
            Assert.That(val.B, Is.EqualTo(5));
        });
    }
```

Run the tests.  `OneByteOneDefaultIntRecordStructTest` will fail, as `val.B` has a value of 0.

ElectroJr:
DataRecord serialization is pretty broken at the moment. Default values are ignored for structs (see #6556) and simple records using floats or negative default values will cause `SerializationManager.CreateRecordInstantiator()` to generate invalid IL. E.g., neither of these work:
```cs
[DataRecord] public partial record Foo(float A);
[DataRecord] public partial record Bar(int A = -1);
```

I've tried to fix both of these issues, and added a bunch more tests,  but I'm not super confident that I've got all the edge cases and the code is somewhat convoluted. I'm also not that confident in writing IL code, and mostly worked backwards by looking at actual IL that got generated when manually calling record constructors.

Fixes #6556

### DIFF
--- a/Robust.Shared/Serialization/Manager/SerializationManager.Instantiation.cs
+++ b/Robust.Shared/Serialization/Manager/SerializationManager.Instantiation.cs
@@ -47,11 +47,21 @@ public partial class SerializationManager
         generator.Emit(OpCodes.Ret);
     }
 
+    /// <summary>
+    /// This generates IL code that will try to invoke a record's constructor by passing default values to any arguments.
+    /// </summary>
     private static void CreateRecordInstantiator(ILGenerator generator, Type type)
     {
         var constructors = type.GetConstructors();
         if (constructors.Length == 0)
-            throw new ArgumentException($"Could not find a constructor for record class {type}");
+        {
+            if (!type.IsValueType)
+                throw new ArgumentException($"Could not find a constructor for record class {type}");
+
+            // Handle constructorless data record struct by treating it like a normal struct
+            CreateValueTypeInstantiator(generator, type);
+            return;
+        }
 
         var constructor = constructors[0];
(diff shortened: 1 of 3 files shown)


Robust.Shared/Serialization/Manager/SerializationManager.Instantiation.cs
```diff
@@ -104,13 +164,13 @@ internal ISerializationManager.InstantiationDelegate<T> GetOrCreateInstantiator<
...
             {
-                CreateValueTypeInstantiator(generator, type);
+                CreateRecordInstantiator(generator, type);
             }
-            else if (isRecord)
+            else if (type.IsValueType)
             {
-                CreateRecordInstantiator(generator, type);
+                CreateValueTypeInstantiator(generator, type);
             }
```
ElectroJr, line 114: Switching these if blocks is what actually fixes #6556. Without this, the constructors of DataRecord structs are basically just ignored.

PR #6623, opened 2026-06-07
Add batch drawing APIs for WorldHandle

metalgearsloth:
Internally Clyde batches but an explicit API lets us skip the batch checks if we're drawing hundreds / thousands (cough decals) of the same quad in a row.

Same with caching the white clyde texture (afaict this never gets invalidated in engine so should be safe to do so?)


Pok27:
The branch has been deleted, the PR needs to be recreated

metalgearsloth:
> The branch has been deleted, the PR needs to be recreated

got merged in fed

PR #6627, opened 2026-06-08
Add chunk entities

### LINKED ISSUE #5839: Chunk entities
So atm decals + gas tile overlays have their own bespoke networking.

To quote electro:

Image

I would recommend starting with porting decals first as the main pain point is making sure the data still exists even if the pvschunk is empty of entities (and vice versa with making sure it's not accidentally cleaned up too early).

metalgearsloth:
Saves bespoke content code for chunk-based data. No idea if it's faster or not but much easier on content.

The pvs metadata change scares me but saved a separate hashset check. ClientGameStateManager might be slower but wasn't sure how to get around that.

image

Resolves https://github.com/space-wizards/RobustToolbox/issues/5839


Robust.Server/GameStates/PvsData.cs
```diff
@@ -188,7 +193,23 @@ internal struct PvsMetadata
...
+        get => (_flags & 0x80) != 0;
+        set => _flags = value ? (byte) (_flags | 0x80) : (byte) (_flags & ~0x80);
+    }
+
+    public EntityLifeStage LifeStage
+    {
+        get => (EntityLifeStage) (_flags & 0x7F);
+        set => _flags = (byte) (((byte) value & 0x7F) | (_flags & 0x80));
+    }
+
```
metalgearsloth, line 212: Evil blursed code but we do have a spare bit. Saves storing the data elsewhere.

PR #6632, opened 2026-06-11
Reduce TryParseEnum string allocs

metalgearsloth:
Added length check as well as an early-out.


metalgearsloth:
I just used these for tracking for my own reference but I have a notepad with my own PRs now!

PR #6634, opened 2026-06-12
Add StaggeredUpdateTracker for tick spreading

Illiux:
Add an entity system helper designed to be used where a component is updated on a regular interval greater than one tick, with updates intended to be spread across multiple ticks. This is intended to replace several places today where content records a next update field on a component, then iterates all entities every tick checking to see if it is time to update that component.

The implementation in this commit efficiently keeps a sorted schedule of updates and allows the entities due to update (and only those entities) to be enumerated in amortized constant time for each. Removed components are handled lazily, and new components are handled via a priority queue. The schedule itself is implemented as a ring buffer.

A limitation of the approach is that the update interval cannot vary across individual instances of tracked components.

### DIFF
--- a/Robust.Shared/GameObjects/EntitySystem.StaggeredUpdate.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.StaggeredUpdate.cs
@@ -36,6 +36,7 @@ public partial class EntitySystem
 public interface IStaggeredUpdate
 {
     static abstract TimeSpan UpdateInterval { get; }
+    static abstract TimeSpan MaxInitialDelay { get; }
 }
 
 public sealed class StaggeredUpdateTracker<TComp>
@@ -49,7 +50,6 @@ public sealed class StaggeredUpdateTracker<TComp>
     private readonly EntityQuery<MetaDataComponent> _metaQuery;
     private readonly IRobustRandom _rng;
     private readonly IGameTiming _timing;
-    private readonly TimeSpan _updateInterval;
     private readonly EntityEventRefHandler<TComp, MapInitEvent>? _mapInit;
 
     internal StaggeredUpdateTracker(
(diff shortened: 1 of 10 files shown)


MLGTASTICa:
Any chance for a implementation that doesn't register on the MapInitEvent?  if this does get merged i'd like to use it but i see that restriction being a issue (unless im wrong and engine/content have split  bus events that don't conflict)

Illiux:
What use case do you have in mind? The `GetStaggeredUpdateTracker` `mapInit` param is a chained MapInit handler so the idea is that systems that want to also hook it pass that handler into `GetStaggeredUpdateTracker` instead of to `SubscribeLocalEvent`.

If this were implemented as an abstract entity system it could also be made to work something like `VisualizerSystem` where you'd override a protected method to do your handler.

Do you have a use case that isn't covered by those? Where you'd for some reason want manual control over addition and removal instead of having it track component lifecycle?

Robust.Shared/GameObjects/EntitySystem.StaggeredUpdate.cs
```diff
@@ -0,0 +1,199 @@
...
+    internal void OnMapInit(Entity<TComp> ent, ref MapInitEvent args)
+    {
+        _mapInit?.Invoke(ent, ref args); // call a chained event handler if we have one
+
+        if (!_tracked.Add(ent.Owner)) return;
+
+        // randomize an offset from the current tick, up to interval
+        // we start from current tick + 1 because updates for the current tick may already have been processed
+        var when = _timing.CurTime + _timing.TickPeriod + _rng.Next(TimeSpan.Zero, _updateInterval);
+        _insertQueue.Enqueue(ent.Owner, when);
```
mirrorcult, line 86: this seems like it might stagger the initial update for components with a very long update interval (5 minutes or something) far more than i would expect -- really i would expect things to only get smeared over a handful of ticks surrounding the first update time. is the intention that very long update intervals just aren't used with this system really? since i expect most components are on the order of 1-5 seconds rather than something like that

Illiux: I didn't have that kind of thing in mind at all when I was implementing this. But I can pretty easily make the initial update delay tunable. I'll do that.

PR #6635, opened 2026-06-13
SpriteSystem helpers to retrieve LayerMap keys

MLGTASTICa:
Adds a helper to retrieve a SpriteComp's layerMap keys. Usefull for systems working with dynamic mappings

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/SpriteSystem.LayerMap.cs
+++ b/Robust.Client/GameObjects/EntitySystems/SpriteSystem.LayerMap.cs
@@ -195,14 +195,16 @@ public sealed partial class SpriteSystem
         return sprite.Comp.LayerMap[key];
     }
 
-    public bool LayerMapKeys(Entity<SpriteComponent?> sprite, [NotNullWhen(true)] out List<Object>? keys)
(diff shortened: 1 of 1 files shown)


Robust.Client/GameObjects/EntitySystems/SpriteSystem.LayerMap.cs
```diff
@@ -194,6 +195,17 @@ public int LayerMapGet(Entity<SpriteComponent?> sprite, string key)
         return sprite.Comp.LayerMap[key];
     }
 
+    public bool LayerMapKeys(Entity<SpriteComponent?> sprite, [NotNullWhen(true)] out List<Object>? keys)
```
PJB3005: It seems quite burdersome to have this method possibly fallible. Just return the list directly?

Also, does this need to return a list instead of just an array?

PJB3005 (changes requested):

Robust.Client/GameObjects/EntitySystems/SpriteSystem.LayerMap.cs
MLGTASTICa: Updated it , was following the model of the other functions

PR #6638, opened 2026-06-19
Add audio input (microphone), Opus codec, and offline speech-to-text primitives

issyman182:
## Summary

Exposes engine-level audio capabilities that sandboxed content cannot implement itself, because content assemblies may not reference native audio/codec libraries (OpenTK/OpenAL capture, Concentus, Whisper.net) or use native interop. These are the primitives needed to build features like in-game voice chat purely in content.

### Microphone capture
- `IAudioManager.GetAudioInputDevices()` enumerates capture devices.
- `IAudioManager.OpenAudioInput(deviceName, sampleRate, internalBufferSamples)` opens a microphone as a new `IAudioInputDevice` that produces mono 16-bit PCM, backed by OpenAL `ALC_EXT_CAPTURE`.
- `IAudioManager.GetAudioOutputDevices()` enumerates output devices.

### Opus codec
- `IAudioManager.CreateOpusEncoder()` / `CreateOpusDecoder()` return `IOpusEncoder` / `IOpusDecoder`, backed by the pure-managed [Concentus](https://www.nuget.org/packages/Concentus) library.

### Streaming playback
- `IBufferedAudioSource` is now `public` and `IAudioManager.CreateBufferedAudioSource()` is exposed (moved from the internal `IAudioInternal`), so PCM frames can be queued into a streaming source at runtime.

### Offline speech-to-text
- `IAudioManager.CreateSpeechTranscriber(modelPath, language)` returns a new `ISpeechTranscriber` that transcribes 16 kHz mono PCM into text using a **local** Whisper model, backed by [Whisper.net](https://www.nuget.org/packages/Whisper.net) (whisper.cpp). The model file is supplied by the caller; nothing is sent off-device.

`HeadlessAudioManager` implements all new members (no devices; the Opus codec works without an audio device, transcription returns null).

## New dependencies
- `Concentus` (pure-managed Opus) — referenced with `PrivateAssets="compile"`.
- `Whisper.net` + `Whisper.net.Runtime` (native whisper.cpp binaries) — referenced with `PrivateAssets="compile"`.

## Notes
The speech-to-text primitive is included here for completeness but pulls in native whisper.cpp runtime binaries; happy to split it into a separate PR (or drop it) if the added dependency is unwanted in core.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -12,6 +12,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="BenchmarkDotNet" Version="0.15.8" />
+    <PackageVersion Include="Concentus" Version="2.2.2" />
     <PackageVersion Include="DiscordRichPresence" Version="1.2.1.24" />
     <PackageVersion Include="JetBrains.Annotations" Version="2025.2.4" />
     <PackageVersion Include="JetBrains.Profiler.Api" Version="1.4.10" />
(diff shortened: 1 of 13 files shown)


PJB3005:
Is this PR AI-generated?

issyman182:
yes its AI generated. i want to implement proximity voice chat

PR #6647, opened 2026-06-25
fix warns associated with SpriteComponent (CS0618)

JrInventor05:
Replaced obsoleted SpriteComponent methods  from other systems to SpriteSystem methods.
Added new public method in SpriteSystem - CopyLayers. It's copy layer to provided entity.


metalgearsloth (changes requested):
Needs breaking change release note.

JrInventor05:
> Needs breaking change release note.

Alright, I'll do that. But I have a question: right now, the methods in SpriteSystem are split into two or three separate methods—identical in content but accepting either a string, an enum, or an int (though not always the int) as input. Should I combine them into a single method using `object`? After all, `MapLayers` are based on `object` to begin with, and some of the methods in `SpriteComponent` also work with `object`.

JrInventor05:
> Needs breaking change release note.

I'm having a bit of a brain fart—where should I put it? Here in the PR text, or in RELEASE-NOTES.md?

PR #6649, opened 2026-06-25
Merge Bouba RT into Kiki RT

slarticodefast:
DO NOT SQUASH

This merges all PRs from https://github.com/Space-Wizards-Federation/RobustToolbox, excluding some new-auth specific changes (see the last two commits).

I had to revert these rather than cherry-pick because PRs were not squash merged properly and they included release note changes, which caused conflicts.


slarticodefast:
Test fail is real.
Git seems to dislike the submodule changes in here. I have to think about how to fix this tomorrow.

DrSmugleaf (approved):

DrSmugleaf:
The test fail should be because the hash for the fix here is different than in the other repo I'll check on my pc tomorrow though

Rest is good

eoineoineoin:
> Test fail is real. Git seems to dislike the submodule changes in here. I have to think about how to fix this tomorrow.

Issue is the SpaceWizards.Lidgren.Network submodule points to 805b6e1c, which is the commit on Space-Wizards-Federation. The same commit went into Space-Wizards with a different commit message, so got a different commit ID. You should be able to checkout Lidgren.Network at 726dd552 and commit that to RT:

```
cd Lidgren.Network/Lidgren.Network
git checkout 726dd552
cd -
git add Lidgren.Network/Lidgren.Network
git commit
```

Suspect it worked on your computer because you have both space-wizards and space-wizards-federation hooked up as remotes simultaneously, so it was able to find the 805b6e1c commit.

deltanedas:
should probably rebase and squash the early prs that werent squash merged, its so ugly

slarticodefast:
Then we would also have to rebase every single open PR on the new repos

deltanedas:
its just 68 lines of changing some from pick to s?
`git rebase -i`

PR #6650, opened 2026-06-25
Add packages.lock.json via RestorePackagesWithLockFile

SlamBamActionman:
This PR adds 
`<RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>` to Directory.Build.props, generating package lock files for the various nuget packages used. 

This is a (possibly) temporary measure to ensure that the nuget packages remain locked to hash, until we can publish our own version of the Robust and SpaceWizards packages, as those are currently not under our control.

Will need evaluation from the Package People - this does lock other packages unrelated to those two mentioned above.

### DIFF
--- a/Avalonia.Base/packages.lock.json
+++ b/Avalonia.Base/packages.lock.json
@@ -1,0 +1,6 @@
+{
+  "version": 1,
+  "dependencies": {
+    "net10.0": {}
+  }
+}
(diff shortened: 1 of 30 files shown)


slarticodefast:
best merge #6649 first to prevent merge conflicts

FairlySadPanda:
Easiest PR in RT history be like 
image

This looks right to me, I'm glad that NuGet's tooling has made this painless.

metalgearsloth:
I can confirm there's probably no trojan here TM.

SlamBamActionman:
We have confirmed ownership of the nuget packages, so is this PR still necessary?

SlamBamActionman:
Probably just gonna kill this one since we got the package ownership now 🙏

PR #6657, opened 2026-06-26
Cache PostShader render targets in Clyde

lermal:
PostShader sprites previously reused a single render target for the whole DrawEntities pass. If a large sprite was drawn first, every subsequent PostShader entity paid for a shader pass over that oversized buffer. Targets were also created and disposed every frame.

- Add a small pooled cache (power-of-two buckets, max 16 RTs) reused across frames
- Rent and return a render target per PostShader sprite instead of growing one RT for the pass
- Blit only the actual sprite region when the pooled RT is larger than needed
- Clear the pool on Clyde shutdown


metalgearsloth:
Good stuff!!!

PR #6658, opened 2026-06-26
Add cvars for new lidgren stuff

### LINKED ISSUE #34: Cannot connect multiple clients quickly on dev
If you connect with multiple clients quickly (Try doing this with 4 clients 😭) #27 trolls and doesn't allow it. This was so confusing for like 2 hours I just assumed my code was bad somehow

Image

@deltanedas stop the trolling!!

deltanedas:
- added cvars for my new config fields which update when changed. technically client can use them too but it doesnt matter (in soviet russia, server ddos YOU!)
- renamed ConnectionTimeout cvar field to NetConnectionTimeout for consistency
- update NetConnectionTimeout in the config when its changed, it works and no reason not to
- made it apply to server too, maybe you think 25 seconds for a single udp packet to arrive is far too long and want to shorten it to like 5. this also means unauthenticated DOS attempts have less time to do anything before being disconnected 

basically did this so if a server owner finds limits i use to be unreasonable (realistically just for the mythical ss14 lan parties) they can change them whenever with  no friction since they apply immediately

see https://github.com/space-wizards/SpaceWizards.Lidgren.Network/pull/27

resolves https://github.com/space-wizards/SpaceWizards.Lidgren.Network/issues/34 since it's now possible to raise the limits for dev

### DIFF
--- a/Robust.Shared/CVars.cs
+++ b/Robust.Shared/CVars.cs
@@ -36,6 +36,44 @@ namespace Robust.Shared
             CVarDef.Create("net.max_connections", 256, CVar.ARCHIVE | CVar.REPLICATED | CVar.SERVER);
 
         /// <summary>
+        /// How many seconds after the last message from the server or a client before we consider it timed out.
+        /// </summary>
+        public static readonly CVarDef<float> NetConnectionTimeout =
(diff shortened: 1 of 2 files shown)


deltanedas:
@metalgearsloth merge

metalgearsloth:
Yes master.

PR #6660, opened 2026-06-27
Allow foreach with EntityQueryEnumerator and AllEntityQueryEnumerator

DrSmugleaf:
Test attached
Returns Entity\<T\>


metalgearsloth:
Sniped me to it I started doing this with some mapmanager stuff.

PR #6662, opened 2026-06-27
Optimise Box2i

metalgearsloth:
Similar change to Box2.

### DIFF
--- a/Robust.Shared.Maths/Box2i.cs
+++ b/Robust.Shared.Maths/Box2i.cs
@@ -13,56 +13,196 @@ namespace Robust.Shared.Maths
     {
         public static Box2i Empty => new();
 
-        [FieldOffset(sizeof(int) * 0)] public int Left;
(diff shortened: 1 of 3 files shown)


Robust.Shared.Maths/Box2i.cs
```diff
@@ -5,293 +5,514 @@
...
 
-        public Box2i(Vector2i bottomLeft, Vector2i topRight)
+    public int Left
+    {
+        readonly get => _left;
+        set
         {
-            Unsafe.SkipInit(out this);
+            if (value > _right)
+                throw new ArgumentOutOfRangeException(nameof(value), value, "Left cannot be greater than Right.");
```
ArtisticRoomba, line 30: could use `ArgumentOutOfRangeException` helpers but not really the end of the world

ArtisticRoomba (approved):

PR #6668, opened 2026-06-27
Box2 + Box2Rotated performance and fixes

metalgearsloth:
See https://github.com/Space-Wizards-Federation/RobustToolbox/pull/39


AJCM-git (approved):
only thing i suss out are the ordering of the  box2 we pass in the hlr but i trust in the process

PR #6670, opened 2026-06-27
add auth server and engine information to json apis and replays

deltanedas:
combined old swf prs, atomic commits

## engine_type

ill make my life a bit easier
its for deciding which cdn and signing key to use for multiengine launcher, doesnt do anything for you since you dont use it
uses the string engine_kind in /status and /info's build object, as well as replays

think of it like fork_id for organizing server manifests

## auth_servers

lets the launcher know what auth servers an account is allowed to have to connect
currently it adds `"auth_servers": ["<auth server cvar value>"]` to the auth object when fetching /info. it is not present in /status.
adding multiple servers e.g. if using multiauth proxy (in which case youd want to change it anyway or people cant join as the launcher would cry about not having localhost as an auth server) is trivial later on and the json "protocol" will support it to boot with no changes on launchers (assuming theyre not shitcode hoping its only ever 1 url)

"hey you need to use a different launcher / sign in on this auth server" in the launcher is infinitely better than waiting for game to start then just get "lol authentication failed"

### DIFF
--- a/Robust.Server/ServerStatus/StatusHost.Handlers.cs
+++ b/Robust.Server/ServerStatus/StatusHost.Handlers.cs
@@ -88,9 +88,12 @@ namespace Robust.Server.ServerStatus
                 buildInfo = GetExternalBuildInfo();
             }
 
+            var authServers = new JsonArray();
+            authServers.Add(_cfg.GetCVar(CVars.AuthServer));
             var authInfo = new JsonObject
             {
                 ["mode"] = _netManager.Auth.ToString(),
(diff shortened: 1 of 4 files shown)


deltanedas:
test fail fake

deltanedas:
@metalgearsloth chud

metalgearsloth:
> @metalgearsloth chud

lol outside my wheelhouse pom

deltanedas:
then nothing for you to say is bad, goidamerge

PR #6671, opened 2026-06-27
set SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS to false

rabbi-internet:
Before this change, when the main game window loses focus in fullscreen mode it minimizes. After this change, when the main game window loses focus in fullscreen mode it does not minimize. It will flicker black briefly if the desktop video mode is different from the games (which it always will be) which is expected behavior. I've never known anyone that wants the game to minimize in this case, but if this is desired behavior we can close this PR.

Tested on Windows 10. SDL3 should work the same on wayland/x11/win11 in this case, but I haven't tested it.


deltanedas:
should probably be a cvar for this individual

image

metalgearsloth (changes requested):
Agree with delta definitely CVar material.

Also write a release note please (should be breaking change due to changing the default).

PR #6674, opened 2026-06-27
Serializer improvement and cleanup

deltanedas:
- all type serializers can now use `[Dependency]`, made them all use it instead of resolve
- added `BaseTypeSerializer` currently just for Log sawmill. no more copy pasting the same getsawmill slop
  in future i want to kill the SerializationManager + IDependencyCollection args from all methods but thats much bigger
- added SerializationManager arg to SerializerProvider so it can create serializers with dependecies and log sawmill.
- changed sound specifiers to not have redundant custom serializers set lol
  changed sound collection specifier to use protoid not string
- **removed** all unused type serializers, obsoleted avoidable PrototypeIdX serializers since nothing has to use them. they can all die after someone makes a AbstractProtoId and changes every prototype to use that instead of the custom serializer slop :)

aside from shitcode le bad, it will also improve test performance by not spamming ioc resolves for every fucking call to everything.
i have used this style for ages using ??= resolve instead of dependency and its perfectly fine

requires https://github.com/space-wizards/space-station-14/pull/44364 for content to survive


deltanedas:
all test pass :face_holding_back_tears:

metalgearsloth (approved):
I have pinged smug for 2nd set of eyes but LGTM.

PR #6678, opened 2026-06-28
Modular Resources (second attempt)

Roudenn:
Reopening of the #6331

## About this PR
This PR introduces Modular resource loading. It allows forks to mount entirely separate physical directories (e.g., `ModResources/`) alongside the standard `Resources/` folder. The engine treats these as native VFS roots without forcing them to live inside the standard directory structure.

## Why
Currently, downstreams are forced to mix their custom assets into the monolithic `Resources/` directory. To keep files organized, resorting to the "Underscore Hack"--prefixing folders with `_` (e.g., `_Goobstation`, `_Corvax`) just to group files together. As the result, features are split across multiple of `Resources/` "root" directories instead of being contained in a single module, which is less than ideal to maintain.
Also this PR allows for the resources to live in a separate submodule, which improves modularity of the engine significantly.

### Resource Manifest changes
`ResourceManifestData` now holds a `resources` dictionary, with where key is a path in a Virtual File System (what you have to specify in `ResPath`s to get the files inside), and value is the real path of the folder relative to the content root.

**Example usage:**
```YAML
# manifest.yml
resources:
  FooStation: Modules/FooStation/Resources
```
```YAML
# Example of accessing a sprite in a module folder
- type: Sprite
  sprite: /FooStation/Textures/Objects/some_sprite.rsi
  state: goober
```

Why not just append the new resource folders to the main one? This will cause files to start colliding, meaning that ones will override others, which is very bad and super inconvenient.

### Prototypes, Locale, ResourceCache, and Hot Reloading changes (and shenanigans)
Because modular resource folders are separate roots, and because of how the searching algorithms works in these managers, to discover resources automatically from modules they have to get the `ResourceManifestData` and iterate through all virtual roots and get all files in them too.

Content code might also need to do this in some places. Maybe this can be replaced with some cool generic API method later, but that would require standartising a lot of magical constants together, which is out of scope of this PR.

### Absolute/Relative resource paths logic
In some cases (mainly textures), when the ResPath is relative, it automatically serializes to have the `/Textures/` prefix. This format doesn't work with modular folders, so all paths to modular resources have to be absolute (see example 2 sections above)

### Packaging
All modular resources get copied into the main `/Resources` folder on packaging - that way the VFS structure matches both on Debug and Release builds.

### Tests for `ResourceManifestData` loading
Added 2 tests:
- First loads a very simple modular resource folder example (with 1 prototype, locale and texture file) and checks that all of the resources are loaded and registered properly;
- Second ckecks that a manifest with an invalid path pointing to an outside directory will throw `InvalidOperationException` on launch.

For the 2 tests above, a new option was added to both server and client launch options called `ManifestOverride` - it allows to change the file name of the manifest to search. Also an `EngineManifests` folder was added containing 2 manifest examples.

## Tested

*   [x] Regular resources folder still loads correctly (Backward compatibility)
*   [x] Modular textures and rsis load via `SpriteComponent`
*   [x] Maps load from modular directories
*   [x] Locale load from modular directories
*   [x] Packaging release builds preserve the folder structure and run without crashing
*   [x] Hot Reloading detects changes in modular folders
*   [x] Verified the practical use case on the example of Carpmosia's repository (YAML linter + in-game tests)
*   [ ] Add modular ResourceManager API to prevent copy-pasting the manifest everywhere
*   [ ] Test on a fully packaged build (both server and client through the launcher)
*   [ ] Test on a fully packaged build on MacOS and Linux working (Can't test it myself, they SHOULD work but better to double-check)


Roudenn:
Hot reloading doesn't work, as well as locale strings

Roudenn:
All major parts of the modular Resource folders should be fixed now
Ideally this would be tested with a real fork, but I can't find a fork that has the latest version of RT and I don't feel like backporting and at the same time restructuring an entire build by hand

Also I'm not sure if it's even possible to write IntegrationTests for this feature because of how fundamental it is. Judging by the integration tests in SS14 and by manual tests, it doesn't break anything on Master, so I guess this is ready to merge at least for forks to test

Prototypes, textures, locale and other files load properly, though because of how its structured, the discovery of modular resource folders has to be done manually in some cases, which may require some content changes for features like Changelog. But it's fairly minor and can be done later

SnappingOpossum:
> All major parts of the modular Resource folders should be fixed now Ideally this would be tested with a real fork, but I can't find a fork that has the latest version of RT and I don't feel like backporting and at the same time restructuring an entire build by hand

We have https://github.com/carpmosia/carpmosia/tree/early-merge on 283.1.0, that should at least build and pass tests out the box

Roudenn:
> > All major parts of the modular Resource folders should be fixed now Ideally this would be tested with a real fork, but I can't find a fork that has the latest version of RT and I don't feel like backporting and at the same time restructuring an entire build by hand
> 
> We have https://github.com/carpmosia/carpmosia/tree/early-merge on 283.1.0, that should at least build and pass tests out the box

Thanks I'll try to manually port everything to modular resources here and see if it works

Robust.Shared/ContentPack/ResourceManifestData.cs
```diff
@@ -1,44 +1,61 @@
...
+
+    public static ResourceManifestData LoadFromFile(string filePath)
+    {
+        if (!File.Exists(filePath))
+            return Default;
+
+        using var stream = File.OpenRead(filePath);
+        return Parse(stream);
+    }
+
```
metalgearsloth, line 52: You have 0 sanitisation here.

metalgearsloth (changes requested):
Given the fact I've just found another path escape in this codebase I'm going to require this to have adequate test coverage to avoid more and this absolutely cannot use path or file directly and must use IVirtualDirProvider.

Roudenn:
> We have https://github.com/carpmosia/carpmosia/tree/early-merge on 283.1.0, that should at least build and pass tests out the box

Ported all of the main resources to a modular folder and everything works properly, YAML linter runs without any errors and all of the content is in place

Robust.Shared/Serialization/TypeSerializers/Implementations/MapPathSerializer.cs
```diff
@@ -0,0 +1,45 @@
...
+using Robust.Shared.Serialization.Manager;
+using Robust.Shared.Serialization.Markdown;
+using Robust.Shared.Serialization.Markdown.Validation;
+using Robust.Shared.Serialization.Markdown.Value;
+using Robust.Shared.Serialization.TypeSerializers.Interfaces;
+using Robust.Shared.Utility;
+
+namespace Robust.Shared.Serialization.TypeSerializers.Implementations;
+
+public sealed class MapPathSerializer : ITypeSerializer<ResPath, ValueDataNode>
```
Roudenn: Not sure if that is needed in the engine?

Robust.Shared/ContentPack/ResourceManifestData.cs
```diff
@@ -77,22 +98,48 @@ public static ResourceManifestData LoadResourceManifest(IResourceManager res)
...
+            var diskPath = child.Value.AsString();
 
-            var sequence = (YamlSequenceNode)node;
-            var array = new string[sequence.Children.Count];
-            for (var i = 0; i < array.Length; i++)
-            {
-                array[i] = sequence[i].AsString();
-            }
+            // Sanitize the path right away so it never escapes the root.
+            var resPath = new ResPath(diskPath).Clean().ToString();
```
Roudenn: Not sure if `ResPath.Clean()` is enough to prevent the path from escaping, or maybe it should just be sanitized in `ProgramShared` using PathHelpers since it's the only place where the real path actually used instead of the virtual one

Robust.Shared/Serialization/TypeSerializers/Implementations/ResPathSerializer.cs
```diff
@@ -33,8 +32,6 @@ public ValidationNode Validate(ISerializationManager serializationManager, Value
             path = SpriteSpecifierSerializer.TextureRoot / path;
         }
 
-        path = path.ToRootedPath();
```
SnappingOpossum, line 36: This means that a path that starts with `Textures` and no slash will no longer get rooted.

Robust.Shared/ProgramShared.cs
```diff
@@ -84,18 +80,30 @@ internal static void DoMounts(
...
 #endif
+        var manifest = ResourceManifestData.LoadResourceManifest(res);
+        if (manifest.ModularResources != null)
+        {
+            foreach (var (vfsPath, diskPath) in manifest.ModularResources)
+            {
+                var virtualPath = new ResPath($"/{vfsPath}/");
+                var cleanDiskPath = diskPath.Trim('/', '\\');
+#if FULL_RELEASE
+                res.MountContentDirectory($@"{diskPath}/", virtualPath);
```
Veritable-Calamity: Packaging puts the module under `Resources/Foo`, but this still mounts its original `Modules/Foo/Resources` directory. This was reproduced: the control manifest starts successfully, while the module manifest throws `DirectoryNotFoundException`, despite the module files being present in the package.

Veritable-Calamity, line 99: Packaging moves the module files into `Resources/Foo`, but this still tries to mount `Modules/Foo/Resources`. That source directory isn't included in the package. Testing with the release server confirmed that this throws `DirectoryNotFoundException` during startup. Release builds should use the module files already included under `Resources`.

Robust.Shared/ContentPack/ResourceManifestData.cs
```diff
@@ -1,44 +1,69 @@
...
+    }
+
+    public static ResourceManifestData LoadFromFile(string baseDir, ResPath resPath)
+    {
+        var safePath = PathHelpers.SafeGetResourcePath(baseDir, resPath);
+
+        if (!File.Exists(safePath))
+            return Default;
+
+        using var stream = File.OpenRead(safePath);
```
Veritable-Calamity, line 49: This method is public, so sandboxed content can call it and choose any `baseDir`. The path check doesn't protect game resources when the caller chooses the directory being checked against. Please keep this file-reading helper internal or restrict it to directories chosen by the engine.

Veritable-Calamity: This is potentially a [Path Traversal Vulnerability](https://cwe.mitre.org/data/definitions/35.html) waiting to happen.

Robust.Packaging/RobustSharedPackaging.cs
```diff
@@ -82,10 +84,36 @@ public static Task DoResourceCopy(
...
 
-    private static void CopyDirIntoZip(string directory, string basePath, AssetPass pass)
+    public static Task DoModularResourceCopy(string contentDir, AssetPass pass, HashSet<string> ignoreSet)
+    {
+        var manifest = ResourceManifestData.LoadFromFile(contentDir, new ResPath($"Resources/{ResourceManifestData.DefaultManifestName}"));
+        if (manifest.ModularResources == null) return Task.CompletedTask;
+
+        foreach (var (vfsPath, diskName) in manifest.ModularResources)
+        {
+            var modPath = Path.Combine(contentDir, diskName);
```
Veritable-Calamity, line 94: If `diskName` is absolute, `Path.Combine` ignores `contentDir`. The manifest check only rejects `..`, so absolute paths pass. Testing the full client packaging pipeline confirmed that a file outside the content directory ended up in the ZIP. Please reject absolute paths and check that the final source path stays inside the content directory.

Veritable-Calamity: This may introduce untrusted code or files into the package, allowing unsafe or malicious code to execute.

Robust.Client/Utility/ReloadManager.cs
```diff
@@ -136,11 +138,32 @@ void OnWatch(object sender, FileSystemEventArgs args)
...
+    private ResPath ResolveModularPath(ResPath relative, string rootIter)
+    {
+        var rootName = new DirectoryInfo(rootIter).Name;
+
+        if (_manifest.ModularResources == null)
+            return relative;
+
+        foreach (var (vfsPath, diskPath) in _manifest.ModularResources)
+        {
+            if (diskPath != rootName)
```
Veritable-Calamity, line 161: For `Foo: Modules/Foo/Resources`, this compares `Modules/Foo/Resources` with just `Resources`, so it never matches. A real file-watcher test queued the changed file without its `/Foo` prefix, and the reload handler skipped it. Please match the full mounted directory and use its virtual prefix.

Robust.Packaging/RobustSharedPackaging.cs
```diff
@@ -82,10 +84,36 @@ public static Task DoResourceCopy(
...
+        string directory,
+        string basePath,
+        AssetPass pass,
+        IReadOnlySet<string>? ignoreSet = null)
     {
         foreach (var file in Directory.EnumerateFiles(directory, "*.*", SearchOption.AllDirectories))
         {
+            var filename = Path.GetFileName(file);
+            if (ignoreSet != null && ignoreSet.Contains(filename))
+                continue;
```
Veritable-Calamity, line 115: This only checks filenames, but the ignore list also contains folder names like `Maps` and `ConfigPresets`. Testing confirmed that the client ZIP excluded the normal `Maps` folder but included `Foo/Maps/module.yml`. Please skip ignored folders and their contents, as the existing resource-copy code does.

Robust.Packaging/RobustSharedPackaging.cs
```diff
@@ -82,10 +84,36 @@ public static Task DoResourceCopy(
...
+        if (manifest.ModularResources == null) return Task.CompletedTask;
+
+        foreach (var (vfsPath, diskName) in manifest.ModularResources)
+        {
+            var modPath = Path.Combine(contentDir, diskName);
+            if (!Directory.Exists(modPath))
+                continue;
+
+            var zipRoot = vfsPath.TrimStart('/');
+            CopyDirIntoZip(modPath, zipRoot, pass, ignoreSet);
```
Veritable-Calamity, line 99: The manifest key isn't checked before becoming a ZIP path. Trimming leading slashes still allows `../outside`, which produced `../outside/sentinel.txt` in a completed package during testing. .NET refused to extract that package, and no write outside the destination was found. Please reject parent-directory segments before writing these paths into the ZIP.

Veritable-Calamity (changes requested):
Testing confirmed a filesystem sandbox bypass: sandboxed content can use `LoadFromFile` to read supported YAML fields from files outside the mounted resources. Since the caller chooses `baseDir`, the current path check doesn’t prevent this. RCE has not been demonstrated, but the confirmed file access **needs** to be fixed before merging.

Testing also reproduced the startup, hot-reload, and packaging problems described in the inline comments. Please address these and add tests, especially for absolute source paths and parent-directory segments in ZIP paths.

I used solely the packaged server startup for these reproductions.

Veritable-Calamity:
Refer to [MITRE](https://cwe.mitre.org/data/definitions/35.html) [CWE](https://cwe.mitre.org/data/definitions/180.html) [Mitigations ](https://cwe.mitre.org/data/definitions/174.html) for further guidance.

PR #6682, opened 2026-06-28
Add DictionaryEquals helper

metalgearsloth:
IDK if there's an easier way to do this but .Equals doesn't do what you expect.

### DIFF
--- a/Robust.Shared/Utility/CollectionExtensions.cs
+++ b/Robust.Shared/Utility/CollectionExtensions.cs
@@ -56,6 +56,17 @@ namespace Robust.Shared.Utility
             return dict;
         }
 
+        /// <summary>
+        /// Compares the entries inside of 2 dictionaries to check equality.
+        /// </summary>
+        /// <remarks>
+        /// The base Equals implementation checks references hence this.
+        /// </remarks>
+        /// <param name="self"></param>
+        /// <param name="other"></param>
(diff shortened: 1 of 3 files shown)


Robust.Shared/Utility/CollectionExtensions.cs
```diff
@@ -56,6 +56,24 @@ public static Dictionary<TKey, TValue> ShallowClone<TKey, TValue>(this Dictionar
             return dict;
         }
 
+        public static bool DictionaryEquals<TKey, TValue>(
```
slarticodefast, line 70: add documentation and mention how it is different from Dictionary.Equals

Robust.Shared/Utility/CollectionExtensions.cs
```diff
@@ -56,6 +56,24 @@ public static Dictionary<TKey, TValue> ShallowClone<TKey, TValue>(this Dictionar
...
+            this IReadOnlyDictionary<TKey, TValue> self,
+            IReadOnlyDictionary<TKey, TValue> other)
+            where TKey : notnull
+        {
+            if (self.Count != other.Count)
```
slarticodefast, line 75: you can also add an early return by comparing the reference

slarticodefast (changes requested):

PR #6686, opened 2026-06-29
Basic Tracy support

HeyBanditoz:
## Intro
This PR adds basic Tracy support to the RT engine.

The existing profiler system is untouched, both will work together.

Technically it creates a new partial ProfManager class and adds support to the already existing `GroupGuard` and `ValueGuard`
structs so adding new profilers calls will cause it to show up on both.

Download the Tracy client from their [Github releases.](https://github.com/wolfpld/tracy/releases/tag/v0.13.1) Only v0.13.1 is supported.

### Screenshots (open in new window!)
Client example:
tracy-profiler_sgumhySNiI

Server example:
tracy-profiler_GFgqmspz7I

## To use
Enable the `prof.tracy.enabled` cvar, `cvar prof.tracy.enabled true`, this can be done on either the client or server.
Both are supported, but I mostly focused on the client for this PR. You can run Tracy against the server though and see
rough system overview per tick! And you can get averages in the statistics or frame graph tab, too.

Afterwards, the game will hitch as Tracy initializes. You can then connect with Tracy and start seeing frame data!

CPU usage of the system, managed dotnet heap, allocation rate, and Gen0, Gen1 and Gen2 collections, along with the number
of entities the client knows about.

You'll see everything profiled with `_prof.Group("...") {}`, which isn't too much, but...

## What if I want to profile my code?
1. Add `ProfManager` as a dependency in your system:
    * `[Dependency] private ProfManager _prof = default!;`
2. Call `_prof.Group("my slow block")` to start a profiler zone.
    * This returns an `IDisposable` so you'll want to do `using (var group = _prof.Group("block") {...}` 
	or `using var group = _prof.Group("block")` to profile the entire scope.
3. That's it! If you want to add more details, you can call `group.TrazyZone?.EmitText("extra details here!");`
   and they will show up as custom text in your zone in Tracy.

### Example
This diff:
```diff
diff --git a/Robust.Client/Physics/PhysicsSystem.cs b/Robust.Client/Physics/PhysicsSystem.cs
index 6552ef272..27f239b0c 100644
--- a/Robust.Client/Physics/PhysicsSystem.cs
+++ b/Robust.Client/Physics/PhysicsSystem.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Threading;
 using JetBrains.Annotations;
 using Robust.Shared.Collections;
 using Robust.Shared.GameObjects;
@@ -7,6 +8,7 @@ using Robust.Shared.Physics;
 using Robust.Shared.Physics.Components;
 using Robust.Shared.Physics.Dynamics;
 using Robust.Shared.Physics.Systems;
+using Robust.Shared.Profiling;
 using Robust.Shared.Timing;
 
 namespace Robust.Client.Physics
@@ -17,9 +19,15 @@ namespace Robust.Client.Physics
         [Dependency] private IGameTiming _gameTiming = default!;
         [Dependency] private SharedTransformSystem _transform = default!;
         [Dependency] private SharedBroadphaseSystem _broadphase = default!;
+        [Dependency] private ProfManager _prof = default!;
 
         public override void Update(float frameTime)
         {
+            using (var group = _prof.Group("My Evil Sleep"))
+            {
+                Thread.Sleep(1);
+                group.TracyZone?.EmitText("Look at me, I'm evil and slept!!");
+            }
             UpdateIsPredicted();
             SimulateWorld(frameTime, _gameTiming.InPrediction);
         }
```
will produce a Tracy zone that looks like this:
tracy-profiler_sv7JS8ZqXB
	
## What still could to be done
According to the [Tracy manual,](https://github.com/wolfpld/tracy/releases/download/v0.13.1/tracy.pdf), support for profiling OpenGL calls is supported (p. 40). I have not the first idea on how to implement this in RobustToolbox, especially since we use OpenGL wrappers. I dunno graphics programming :(

An example of what that looks like in the profiler is available in the [Tracy demo,](https://tracy.nereid.pl/) though.

There's also support for sending screenshots in p. 30 with example OpenGL code.

Also, documentation for Tracy in RT, for wherever it lives now! Not sure where that's at.

Sending GC stats, start/end time as its own thread?

Auto instrument C# method calls to Tracy

### Extra
Also, thanks to [amylizzle](https://github.com/OpenDreamProject/OpenDream/pull/2127) for inspiration from OpenDream!

### DIFF
--- a/Robust.Shared/CVars.cs
+++ b/Robust.Shared/CVars.cs
@@ -1752,6 +1752,11 @@ namespace Robust.Shared
         public static readonly CVarDef<bool> ProfEnabled = CVarDef.Create("prof.enabled", false);
 
         /// <summary>
+        /// Enables the Tracy profiling system.
+        /// </summary>
+        public static readonly CVarDef<bool> TracyProfEnabled = CVarDef.Create("prof.tracy.enabled", false);
+
+        /// <summary>
         /// Event log buffer size for the profiling system.
         /// </summary>
         public static readonly CVarDef<int> ProfBufferSize = CVarDef.Create("prof.buffer_size", 8192);
--- a/Robust.Shared/CVars.cs
+++ b/Robust.Shared/CVars.cs
@@ -1752,8 +1752,13 @@ namespace Robust.Shared
         public static readonly CVarDef<bool> ProfEnabled = CVarDef.Create("prof.enabled", false);
 
         /// <summary>
-        /// Enables the Tracy profiling system.
+        /// Enables the Tracy profiling system. Tracing will stay enabled for the entire runtime of the program even if
+        /// you turn this cvar off.
         /// </summary>
+        /// <remarks>
+        /// By default, this will listen for Tracy connections on all interfaces! Set the <c>TRACY_ONLY_LOCALHOST</c>
+        /// env var to 1 if you want to restrict to localhost.
+        /// </remarks>
         public static readonly CVarDef<bool> TracyProfEnabled = CVarDef.Create("prof.tracy.enabled", false);
 
         /// <summary>
--- a/Robust.Shared/Profiling/ProfManager.Tracy.cs
+++ b/Robust.Shared/Profiling/ProfManager.Tracy.cs
@@ -1,5 +1,6 @@
 using System;
 using bottlenoselabs.C2CS.Runtime;
+using Robust.Shared.Maths;
 using static Tracy.PInvoke;
 
 namespace Robust.Shared.Profiling;
@@ -27,7 +28,7 @@ public sealed partial class ProfManager
     private CString _tracyPlotGen2;
     private CString _tracyEntityCount;
 
-    public unsafe void EmitFrameImage(void* image, ushort width, ushort height, byte offset, bool flip)
+    internal unsafe void EmitFrameImage(void* image, ushort width, ushort height, byte offset, bool flip)
     {
         if (!IsTracyEnabled)
             return;
@@ -35,7 +36,7 @@ public sealed partial class ProfManager
         TracyEmitFrameImage(image, width, height, offset, flip ? 1 : 0);
     }
 
-    public void EmitMemoryPlots(int gcGen0, int gcGen1, int gcGen2)
+    internal void EmitMemoryPlots(int gcGen0, int gcGen1, int gcGen2)
     {
         if (!IsTracyEnabled)
             return;
@@ -55,7 +56,7 @@ public sealed partial class ProfManager
 
     }
 
-    public void EmitEntities(int entityCount)
+    internal void EmitEntities(int entityCount)
     {
         if (!IsTracyEnabled)
             return;
@@ -101,7 +102,7 @@ public sealed partial class ProfManager
     /// Creates a <seealso cref="CString"/> for use by Tracy. Also returns the
     /// length of the string for interop convenience.
     /// </summary>
-    public static CString GetCString(string? fromString, out ulong cLength)
+    internal static CString GetCString(string? fromString, out ulong cLength)
     {
         if (fromString == null)
         {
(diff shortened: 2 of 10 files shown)


Mervill (commented):
- Don't worry about the graphics debugging, like you noticed it's complicated by the wrapper layer. It's a red herring for now.
- Next direction to take this (in my mind) is to get really detailed GC info. See [here](https://learn.microsoft.com/en-us/dotnet/api/system.gc.registerforfullgcnotification?view=net-10.0) for some idea of where to start.

Mervill:
- When Tracy is enabled it will listen for connections on all network interfaces and not just `localhost` we might want to surface that better / we should make sure we're ok with that before we merge.

Mervill (approved):
Very good step toward much better performance tooling. I've wanted to adopt Tracy for a while and this gets that process rolling.

HeyBanditoz:
ty for looking!

___

> Next direction to take this (in my mind) is to get really detailed GC info. See [here](https://learn.microsoft.com/en-us/dotnet/api/system.gc.registerforfullgcnotification?view=net-10.0) for some idea of where to start.

That one is weird since it seems to _start_ an event at 99% memory occupied, instead of when GC starts? I guess it would be hard to run code during that time, but I don't know much about dotnet's GC.

> When Tracy is enabled it will listen for connections on all network interfaces and not just localhost we might want to surface that better / we should make sure we're ok with that before we merge.

Hmm, good point. From the docs: `By default, the Tracy client will listen on all network interfaces. If you want to restrict it to only listening on the localhost interface, define the TRACY_ONLY_LOCALHOST macro at compile-time, or set the TRACY_ONLY_LOCALHOST environment variable to 1 at runtime.`

Should we document that env var somewhere?

HeyBanditoz:
Also it'd be cool to auto instrument method calls but I dunno if that's possible. It's supported in BYOND's 3rd party impl of Tracy and looks like this: 
tracy-profiler_Jur0WF5cmQ

Robust.Shared/CVars.cs
```diff
@@ -1751,6 +1751,11 @@ protected CVars()
         /// </summary>
         public static readonly CVarDef<bool> ProfEnabled = CVarDef.Create("prof.enabled", false);
 
+        /// <summary>
+        /// Enables the Tracy profiling system.
+        /// </summary>
+        public static readonly CVarDef<bool> TracyProfEnabled = CVarDef.Create("prof.tracy.enabled", false);
+
```
Mervill, line 1763: You could add the bit about being exposed to the network as a `<remarks>` section here.

HeyBanditoz: done!

metalgearsloth (commented):
C# side looks fine however I have not audited Tracy-CSharp

Robust.Shared/Profiling/ProfManager.Tracy.cs
```diff
@@ -0,0 +1,161 @@
...
+        TracyEmitPlotInt(_tracyPlotManagedHeap, GC.GetTotalMemory(false));
+        TracyEmitPlotInt(_tracyPlotAllocRate, allocated - _tracyLastAllocatedBytes);
+        TracyEmitPlotInt(_tracyPlotGen0, gcGen0);
+        TracyEmitPlotInt(_tracyPlotGen1, gcGen1);
+        TracyEmitPlotInt(_tracyPlotGen2, gcGen2);
+
+        _tracyLastAllocatedBytes = allocated;
+
+    }
+
```
metalgearsloth, line 58: These should not be public.

Robust.Shared/Profiling/ProfManager.Tracy.cs
```diff
@@ -0,0 +1,161 @@
...
+    {
+        using var textStr = ProfManager.GetCString(text, out var textLn);
+        TracyEmitZoneText(Context, textStr, textLn);
+    }
+
+    public void Dispose()
+    {
+        TracyEmitZoneEnd(Context);
+    }
+}
```
metalgearsloth, line 151: Same here, make internal and content should not be able to call these.

metalgearsloth (changes requested):
Forgot to flag this.

Robust.Shared/Profiling/ProfManager.Tracy.cs
HeyBanditoz: Tightened up. I kept `EmitText` but moved it up to the `Guard`.

metalgearsloth:
We put top minds on this and I didn't see any sandboxing issuesTM.

PR #6689, opened 2026-06-29
QOL TimeoutCvars Allow TOOLS build to be breakpoint friendly

JohnKieser:
<h3>Changes</h3>
Replaced Cvar ConnectionTimeout with ConnectionTimeoutClient and ConnectionTimeoutServer in Robust.Shared/CVars.cs

Also added TOOLS clause to preprocessor directive in line 703 of NetManager.cs, so DEBUG and TOOLS builds can benefit from an arbitrarily large timeout deadline. 

Other changes are just linter auto-adding/removing spaces for style reasons.

<h3>Why</h3>

I recently learned that people do not like the Debug build. I really liked it since it didn't disconnect the client when I used breakpoints. I figured out how it does this, I had a hunch it just ignored timeout or something, and so made the detailed changes.


Robust.Shared/Network/NetManager.cs
```diff
@@ -881,7 +882,7 @@ private void HandleDisconnect(NetPeerData peer, NetConnection connection, string
             try
             {
 #endif
-                OnDisconnected(channel, reason);
+            OnDisconnected(channel, reason);
```
metalgearsloth:
```suggestion
                OnDisconnected(channel, reason);
```

PR #6690, opened 2026-06-29
Changed Resolve to log a Warning instead of an Error

ShepardToTheStars:
~~**Compare to #6691**~~. This PR changes functionality of logMissing slightly by logging to a different output stream.

logMissing can fail tests because it logs to the stderr stream instead of stdout.

#### Reasoning
`logMissing` is very misleading. I expect something logged as an Info or Warning level log. This intermittently fails tests by logging to stderr as well, so clarifying that this is an Error-level log may help.
For example, a recently failed test of ours fails because:

```
  1) SERVER: 8.449s [ERRO] system.entity_storage: Can't resolve "Content.Shared.Storage.Components.EntityStorageComponent" on entity space carp (790) (1034204/n1034204, MobCarp)!
   at System.Environment.get_StackTrace()
   at Content.Shared.Storage.EntitySystems.SharedEntityStorageSystem.CanOpen(EntityUid user, EntityUid target, Boolean silent, EntityStorageComponent component) in /home/runner/work/Delta-v/Delta-v/Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs:line 418
   at Content.Shared.Storage.EntitySystems.SharedEntityStorageSystem.TryOpenStorage(EntityUid user, EntityUid target, Boolean silent) in /home/runner/work/Delta-v/Delta-v/Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs:line 390
   at Content.Server.NPC.HTN.PrimitiveTasks.Operators.Combat.Melee.EscapeOperator.Plan(NPCBlackboard blackboard, CancellationToken cancelToken) in /home/runner/work/Delta-v/Delta-v/Content.Server/NPC/HTN/PrimitiveTasks/Operators/Combat/EscapeOperator.cs:line 60
```

The `EscapeOperator` class contains a call to Resolve but since the entity does not have the component, it logs and error and just completely fails the test, even though that is expected behavior. We could just set `logMissing` to false in this case, **but** there may be cases where we want to log missing components *without* having to set `logMissing` to false just to get tests to not fail randomly.

### DIFF
--- a/Robust.Shared/GameObjects/EntitySystem.Resolve.cs
+++ b/Robust.Shared/GameObjects/EntitySystem.Resolve.cs
@@ -29,7 +29,7 @@ namespace Robust.Shared.GameObjects
 
             if (logMissing && !found)
             {
-                Log.Error($"Can't resolve \"{typeof(TComp)}\" on entity {ToPrettyString(uid)}!\n{Environment.StackTrace}");
+                Log.Warning($"Can't resolve \"{typeof(TComp)}\" on entity {ToPrettyString(uid)}!\n{Environment.StackTrace}");
             }
 
             return found;


EmoGarbage404:
I feel like this is just a content issue with people not considering whether or not something should log an error or not. Logging to info instead of error is pretty much useless. If the resolve failing doesnt matter, then it should be marked as such with the optional param. Making it output to info instead is just papering over the issue so that tests dont fail, ignoring the actual problem.

ShepardToTheStars:
> Making it output to info instead is just papering over the issue so that tests dont fail, ignoring the actual problem.

Counterpoint: If a test needs to assert that resolve doesn't log errors, it should be an explicit assertion in the test that the component was successfully resolved. Otherwise, its just the illusion of a problem, or the default developer behavior will just be to always set it to false and sweep the problem under the rug because waiting ~20 minutes to find out that you have to restart your tests because of an intermittent failure is not exactly a good use of anyone's time.

Or the function calling `Resolve` should log an error to stderr and cause its specific test to fail.

EmoGarbage404:
Its important to distinguish and raise issues when functions are being called on incorrect entities. It's singlehandedly caught numerous bugs with wrong entities being passed. How is it rugsweeping to cognizantly configure usages instead of literally broadly causing all of these to not error solely for the purpose of not wanting test fails.

deltanedas:
bad pr

slarticodefast:
I agree with Emo on this.
The test fails are intended, and if they only happen randomly during a round or in heisentests then that is a symptom of our bad test coverage.

ShepardToTheStars:
@slarticodefast @deltanedas @EmoGarbage404 

So, the issue is access to the types. I'll give you an *arbitrary* example. Say, I have a component and two systems. One system governs the component (which is how things look nowadays). 

Again, this is an example, not how things actually are. But its a decent way of portraying the issue here.

Let's assume we have `GhostComponent`, `GhostSystem`, and another system that uses `GhostSystem` - `GhostAbilitySystem`. This is mostly psuedocode.

```c#
[RegisterComponent]
[Access(typeof(SharedGhostSystem))]
public partial class GhostComponent : Component
{
    [DataField]
    public int SomeProperty = 0;
}
```

And you have SharedGhostSystem, which governs GhostComponent and access to its properties.
```c#
public sealed class SharedGhostSystem : EntitySystem {
    public void TrySetSomeProperty(Entity<GhostComponent?> ent, int value)
    {
        if (!Resolve(ent.Owner, ref ent.Comp))
            return false;
        
        ent.Comp.SomeProperty = value;
        return true;
    }
}
```

If another system wants to interact with `GhostComponent`, it has to do it through `SharedGhostSystem`, which alright, great. I agree with this. But now the issue is when you want to **test** the system that uses the `SharedGhostSystem`. In this example, we'll use uh... it doesn't matter - `GhostAbilitySystem`.

```c#
public sealed class GhostAbilitySystem : EntitySystem {
    [Dependency] private readonly SharedGhostSystem _ghost = default!;
    public bool DoSomething(Entity<GhostComponent?> ent)
    {
        /* other logic */
        return _ghost.TrySetSomeProperty(ent, 2);
    }
}
```

If I write tests around `GhostAbilitySystem`, I cannot test anything that could recover from failure, because for any entities that do not have `SharedGhostComponent`,  `_ghost.SetSomeProperty(ent, 2);` will log an error to the error stream, even if I code `GhostAbilitySystem` in a way that says "Okay, if they don't have ghost component, don't do anything."

Now, I need to write my own `GhostAbilitySystem` like so.

```c#
public sealed class GhostAbilitySystem : EntitySystem {
    [Dependency] private readonly SharedGhostSystem _ghost = default!;
    public bool DoSomething(Entity<GhostComponent?> ent)
    {
        if (!Resolve(ent.Owner, ref ent.Comp, false)) // Or just null check GhostComponent in general
            return false;
            
        /* other logic */
        return _ghost.TrySetSomeProperty(ent, 2);
    }
}
```

Which is just another needless call to `Resolve` to appease unit tests specifically. If `Resolve` threw an exception, I'd understand, but it just logs an error.

The game will run and will log just fine, its just the tests that will fail. I can't govern the system that calls `GhostComponent` the way I want to, so my only recourse  is to modify the systems it calls and tell it to ignore failed resolutions, or remove the access limitations from the component, which is just... bad.

If `Resolve` should fail, it should be an explicit fail. Don't get me wrong - I think its good to be cautious with game state issues. But I don't want to be modifying somewhat-unrelated systems just to appease the test gods. I want resolution fails to matter.

deltanedas:
you can just add a logMissing bool to the apis and pass it to Resolve, plenty of things do this

ShepardToTheStars:
> you can just add a logMissing bool to the apis and pass it to Resolve, plenty of things do this 

Many WizDen system APIs do not give a parameter to override the `Resolve` writing to stderr. (and maybe that's the issue)

metalgearsloth:
> > you can just add a logMissing bool to the apis and pass it to Resolve, plenty of things do this
> 
> Many WizDen system APIs do not give a parameter to override the `Resolve` writing to stderr. (and maybe that's the issue)

Anyone taking a resolve and logging an error is explicitly stating the component is required, it's just a convenience to not have to do it self.

If you want it to actually be optional use trygetcomponent.

ShepardToTheStars:
> Anyone taking a resolve and logging an error is explicitly stating the component is required

Then why would the system have a `Entity<NullableComponent?>` in the first place? If the component is required in the first place, it should not be nullable.

> If you want it to actually be optional use trygetcomponent.

That works to get the component, but access prevents me from doing anything with it. Either I remove the access restriction, or I modify the upstream system, both of which suck.

mirrorcult:
> > Anyone taking a resolve and logging an error is explicitly stating the component is required
> 
> Then why would the system have a `Entity<NullableComponent?>` in the first place? If the component is required in the first place, it should not be nullable.
> 

for the convenience of the caller to not have to resolve it themselves if they know the entity is already supposed to have it

mirrorcult:
http://docs.spacestation14.com/en/general-development/codebase-info/conventions.html#public-api-method-signature

mirrorcult:
if there are issues with specific `Resolve` calls in system APIs that seem like they would make more sense to not error if the component is missing, then they should be changed, rather than every error silenced

ShepardToTheStars:
> if there are issues with specific `Resolve` calls in system APIs that seem like they would make more sense to not error if the component is missing, then they should be changed, rather than every error silenced

I suppose my question to this is; why is that paradigm not inverted? Resolve already returns a boolean if the resolution was success. I don't think I've seen a system that hasn't checked the result of the function.

If and when the component is important enough to log that it wasnt found to stderr, why wouldn't that be something the governing system would say, "Yeah, this is something thats basically an error when running tests." (And only tests)

Debug assertions seem like the way to go if you make sure to fail when required (but in dev and tests), so logging to stderr to ensure tests fail seems so odd when there are better ways to do it.

SnappingOpossum:
I doesn't just ensure tests fail it allows the game to continue running unlike the debug assertions, and unlike those it still gives info in release builds since those still show error logs.

PR #6691, opened 2026-06-29
Clarify logMissing will log to the Error stream

ShepardToTheStars:
**Compare to #6690**. This suggestion does not change functionality, but instead adds clarity.

Adds clarity that resolving missing components will result in an error to the error stream. Changes no functionality really, but provides clarity to developers in IDE intellisense functionality that not setting it to `false` will log errors to the console (and can fail tests).

#### Reasoning
`logMissing` is very misleading. I expect something logged as an Info or Warning level log. This intermittently fails tests by logging to stderr as well. This changes it to still log, but log to stdout rather than stderr.


ShepardToTheStars:
Oh, right, since `logMissing` is an optional param, this may be more of a breaking change than I realize if others have used `logMissing: true` to specify the parameter. I'll probably close this since the intention of this PR over the other was to be unobtrusive but may reopen if #6690 is not merged.

PR #6692, opened 2026-06-29
Auto-generate prototype variants at runtime

chromiumboy:
Currently, if you wish to have two prototypes that are identical except for a single value, you must manually code a child prototype in YAML that inherits from a parent and update the value that needs to be adjusted. This can result in a substantial amount of YAML boilerplate code that becomes difficult to follow over time.

This PR introduces a method of auto-generating collections of multiple prototypes from a single YAML entry that only exist at runtime. This variantization is quite straightforward to implement, works on any prototype, and only requires YAML coding. To begin, simply substitute the prototype ID in the YAML file with
```
!type:CreateVariants
    values: [ <PrototypeId1>, <PrototypeId2>, <PrototypeId3>, ... ]
```
where there is one value in the array for each prototype to be in the auto-generated collection. Other values in the YAML file can be replaced in exactly the same manner. Linear collections (e.g., lists and arrays) can be replaced by
```
!type:CreateVariants
    sequences: [ [<VariantValue1a>, <VariantValue1b>, ...], [<VariantValue2a>, <VariantValue2b>, ... ], ... ]
```
If more complex replacements are desired (e.g., replacing an entire dictionary), you should use the traditional method of inheritance to do so.

Here is example YAML code that will auto-generate three different prototypes, each with its own distinct set of values, at runtime:

```
- type: entity
  id: !type:CreateVariants
    values: [ TestEntityA, TestEntityB, TestEntityC ]
  components:
  - type: Test
    floatValue: !type:CreateVariants
      values: [ 0.5, 1, 1.5 ]
    enumValue: !type:CreateVariants
      values: [ 1, 2, 3 ]
    stringArray: !type:CreateVariants
      sequences: [ [ string-1 ], [ string-1, string-2 ], [ string-1, string-2, string-3 ] ]
  - type: Sprite
    sprite: path/to/overlay.rsi
    layers:
    - sprite: !type:CreateVariants
        values: [ path/to/base.rsi, path/to/base_alt1.rsi, path/to/base_alt2.rsi ]
      state: baseState
      map: ["firstLayer"]
    - state: overlayState
      map: ["secondLayer"]
```

The second prototype to be generated will have the following values:

```
- type: entity
  id: TestEntityB
  components:
  - type: Test
    floatValue: 1
    enumValue: 2
    stringArray: [ string-1, string-2 ]
  - type: Sprite
    sprite: path/to/overlay.rsi
    layers:
    - sprite: path/to/base_alt1.rsi
      state: baseState
      map: ["firstLayer"]
    - state: overlayState
      map: ["secondLayer"]
```

Integration tests for the prototype variant generation have been included.


mirrorcult:
super dope stuff

ArtisticRoomba:
I smell pipe layers code solutions :trollface:

Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
```diff
@@ -197,30 +232,91 @@ public void LoadFile(ResPath file, bool overwrite = false, Dictionary<Type, Hash
...
+            {
+                variantData = new Dictionary<string, ExtractedMappingData>();
+                var variantCollection = new List<string>();
+                
+                // We need to generate a collection of prototype variants.
+                // Extract the IDs of the variants to generate as a sequence of strings.
+                // The number of extracted strings (minus one) is the number of clones to generate.
+                if (mappingNode.TryGet<SequenceDataNode>(VariantValuesFieldAttribute.Name, out var sequenceNode))
+                {
+                    for (int i = 1; i < sequenceNode.Count(); i++)
```
metalgearsloth, line 251: Isn't there some length / count property? As the LINQ one iterates it to determine count.

Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
```diff
@@ -197,30 +232,91 @@ public void LoadFile(ResPath file, bool overwrite = false, Dictionary<Type, Hash
...
+                    {
+                        // Clone the data node, then recursively search through it for any CreateVariants nodes.
+                        // Replace these nodes with data appropriate for the current variant index.
+                        var clonedNode = dataNode.Copy();
+                        RecursivelySearchForVariantNodes(clonedNode, i);
```
metalgearsloth, line 256: I have another PR avoiding deep copies due to the 100MB extra RAM it causes but this should be fine for now.

Robust.Shared/Prototypes/PrototypeManager.YamlLoad.cs
metalgearsloth, line 251: Is it possible to run ensurecapacity on the collections if we have some idea of parent count?

Robust.Shared/Prototypes/PrototypeManager.Variantization.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        foreach (var collectionMember in collectionVariants)
+        {
+            kindCollection[collectionMember] = collectionVariants;
+        }
+
+        _variantCollections[kind] = kindCollection;
+    }
+
```
metalgearsloth, line 106: EnsureCapacity here, also do you even need to re-assign the collection to the list?

Just use collectionsmarshal.getvaluerefordefault so you get the 1 dictionary access instead of 2.

Robust.Shared/Prototypes/PrototypeManager.Variantization.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// </summary>
+    /// <param name="dataNode">The current data node being searched.</param>
+    /// <param name="variantIndex">The current variantization index.</param>
+    private void RecursivelySearchForVariantNodes(DataNode dataNode, int variantIndex)
+    {
+        switch (dataNode)
+        {
+            case MappingDataNode mappingNode:
+                foreach (var (childName, childNode) in mappingNode.Children.ToDictionary())
+                {
```
metalgearsloth, line 30: Is the .ToDictionary necessary?

Robust.Shared/Prototypes/PrototypeManager.Variantization.cs
```diff
@@ -0,0 +1,137 @@
...
+
+namespace Robust.Shared.Prototypes;
+
+public partial class PrototypeManager
+{
+    /// <summary>
+    /// A dictionary that maps a prototype to a list of all variants of that prototype.
+    /// </summary>
+    private readonly Dictionary<Type, Dictionary<string, List<string>>> _variantCollections = new();
+
```
metalgearsloth, line 18: Is it not possible to frozendictionary this? As prototypemanager is the single most expensive thing for entity spawning atm any gains we can get are good.

Robust.Shared/Prototypes/IPrototypeManager.cs
```diff
@@ -564,6 +564,14 @@ void ReloadPrototypes(
...
     FrozenDictionary<ProtoId<EntityCategoryPrototype>, IReadOnlyList<EntityPrototype>> Categories { get; }
+
+    /// <summary>
+    /// Tries to get the list of all associated variants for a given prototype. 
+    /// </summary>
+    /// <param name="collectionMember">The prototype being indexed.</param>
+    /// <param name="collectionVariants">The collection of variants this prototype belongs to.</param>
+    /// <returns>Returns true if the prototype is part of a variant collection, false otherwise.</returns>
+    bool TryGetVariantCollection<T>(ProtoId<T> collectionMember, [NotNullWhen(true)] out List<ProtoId<T>>? collectionVariants) where T : class, IPrototype;
 }
```
metalgearsloth, line 584: Do we enforce prototypes to be classes elsewhere?

metalgearsloth (changes requested):

Robust.Shared/Prototypes/IPrototypeManager.cs
chromiumboy: No clue, TBH. I was following the pattern used in the rest of the functions listed in IPrototypeManager...

Robust.Shared/Prototypes/PrototypeManager.Variantization.cs
chromiumboy: Yep, it'll error out otherwise, saying that the collection is being modified while iterating

PR #6696, opened 2026-06-30
Make entity spawning 2x faster

metalgearsloth:
This has a high chance of breaking something but in tests we trust.

image

So right now serialization copying is the majority of entity spawning.

To copy a component it does up to 3 array accesses to get the copier and then to call it. On something with 20 datafields like a gun that's a lot of overhead compared to assigning an Angle over and over, so now we just do direct assignment unless we need to fallback to createcopy (e.g. customtypeserializers like paused fields or ref types that aren't collections). For basic primitives and valuetypes (at least where they themselves only contain valuetypes) we should be able to get away with cheatingTM.


metalgearsloth:
Think it can be rolled into serv5

PR #6697, opened 2026-06-30
Add `EntitySystem.PostInitialize()`

Roudenn:
## About this PR
This PR adds a new method to EntitySystem and its interface - `PostInitialize()`.
It is called right after all systems were initialized, and EntitySystemManager was marked as initialized.

## Why
`PostInitialize()` can be useful for some core systems that have to setup something while also using other systems.
Right now the only example of this in SS14 is server's `GameTicker`, but it can also be useful for other systems too.

## Technical details
`EntitySystemManager.Initialize` now caches the resolved systems into `_resolvedSystem` list. After the manager is marked as initialized, the list of resolved systems gets iterated through and `IEntitySystem.PostInitialized` method gets called on all of them.

### DIFF
--- a/Robust.Shared.IntegrationTests/GameObjects/EntitySystemManagerOrderTest.cs
+++ b/Robust.Shared.IntegrationTests/GameObjects/EntitySystemManagerOrderTest.cs
@@ -34,6 +34,8 @@ namespace Robust.UnitTesting.Shared.GameObjects
             public virtual IEnumerable<Type> UpdatesBefore => Enumerable.Empty<Type>();
             public bool UpdatesOutsidePrediction => true;
             public void Initialize() { }
+            public void PostInitialize() { }
+
             public void Shutdown() { }
 
             public void Update(float frameTime)
--- a/Robust.Shared.IntegrationTests/GameObjects/EntitySystemManager_Tests.cs
+++ b/Robust.Shared.IntegrationTests/GameObjects/EntitySystemManager_Tests.cs
@@ -18,6 +18,7 @@ namespace Robust.UnitTesting.Shared.GameObjects
             public virtual IEnumerable<Type> UpdatesBefore => Enumerable.Empty<Type>();
             public bool UpdatesOutsidePrediction => true;
             public void Initialize() { }
+            public void PostInitialize() { }
             public void Shutdown() { }
             public void Update(float frameTime) { }
             public void FrameUpdate(float frameTime) { }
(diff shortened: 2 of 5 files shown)


metalgearsloth:
I really don't think systems should have this kind of dependency chain with init and gameticker is very exceptional due to managing rounds outside of gamesim which should ultimately be on a manager.

Roudenn:
> I really don't think systems should have this kind of dependency chain with init and gameticker is very exceptional due to managing rounds outside of gamesim which should ultimately be on a manager.

There are open PRs that would benifit from the PostInitialize():

https://github.com/space-wizards/space-station-14/pull/44418
https://github.com/space-wizards/space-station-14/pull/44346
https://github.com/space-wizards/space-station-14/pull/44345

All of them implement a Dictionary that re-routs to other EntitySystem handlers in order to execute code related to some Type. Not sure how it's actually called but I'll refer to it as Static Handler approach.

This approach is much faster than using for example an event bus, and can be made even faster (~10% from my benchmarks) when also making it a `FrozenDictionary`.
But here's the catch, the dictionary can only be frozen after all systems were registred, which is impossible at the level of systems themselves.

So one of these has to be true:
1. You have to implement a manager for each mechanic that uses the Static Handler method (for node groups, devices, entity effects, and potentially many other performance-heavy mechanics)
2. You have to just ignore the performance boost and cope with having a normal `Dictionary` at all times
3. You call a PostInitialize method to freeze the custom subscription dictionaries.

If you ever need to PostInitialize a system on the client, you have to subscribe to a PlayerManager C# event manually, which is super inconvenient. So if the option 3 is true, then this PR is neccessary.

I consider the first option just wrong because managers are intended for out-of-sim mechanics, so creating a manager for each mechanic with a StaticHandler feature will just be tedious and also will clog up the `EntryPoint`s for no good reason.

If the option 2 is true, then this method may still be useful for other games on RobustToolbox. I don't see an issue with adding a "final" layer like that, since I can't imagine it being misused in any actually bad way.
And if the reason to not implement PostInitialize is just architecture concerns, they no longer hold when practicallity comes in, since Option 1 is objectively bad, and allowing a pretty odd but rarely useful method for performance reasons can be valid if it saves even just a millisecond of the next tick calculation on the Server.

metalgearsloth:
https://github.com/space-wizards/space-station-14/pull/44345 is done on initialize and has 0 postinit behavior. It could also use events and be fine as right now it's just trying to avoid the eventbus entirely.
I did not read the 20k diff on the other PRs but I imagine they are the same.

metalgearsloth:
Talked with some other maints and don't think this is warranted atm. 

If all 3 end up merged as is then maybe we can revisit it but they don't seem to require this over using other patterns and adding more ordering shenanigans could cause headaches (i.e. systems postinit requiring other systems postinit).

PR #6702, opened 2026-07-01
Sourcegen eventsub fixes

metalgearsloth:
- Forgot to check for implicituse when I checked it out.
- Typo because eyesight.


ArtisticRoomba (approved):
its gold

PR #6705, opened 2026-07-01
Add chunk entities

metalgearsloth:
Saves bespoke content code for chunk-based data. No idea if it's faster or not but much easier on content. I completely zoned if the delta compstate change was load-bearing for chunked content or not.

The clientgamestate change to check for chunkents is kinda stinky but wasn't sure how to make it better.


Robust.Shared/GameStates/ChunkEntitySystem.cs
```diff
@@ -0,0 +1,319 @@
...
+            Root = root,
+            Chunk = chunk,
+        };
+        AddComp(uid, comp);
+
+        return (uid, comp);
+    }
+
+    /// <summary>
+    /// Returns a chunk for the specified position if one exists. Assumes normalized chunk origin i.e. position divided by ChunkSize.
```
ArtisticRoomba, line 96: Need debug checks/tests for checking that chunksize is properly divisible/origin normalized

Robust.Client/GameStates/ClientGameStateManager.cs
```diff
@@ -839,11 +854,21 @@ private void ApplyEntityStates(GameState curState, GameState? nextState)
...
                 if (isEnteringPvs)
                 {
+                    if (_detachedChunkEntities.TryGetValue(es.NetEntity, out var detachedTick))
+                    {
+                        if (curState.ToSequence <= detachedTick)
+                            continue;
+
+                        _detachedChunkEntities.Remove(es.NetEntity);
+                        isEnteringPvs = false;
+                    }
```
ArtisticRoomba: Feels kinda gross special-casing this though it's not like `ContainerSystem` is any better.

Robust.Shared/GameStates/ChunkEntitySystem.cs
```diff
@@ -0,0 +1,327 @@
...
+
+/// <summary>
+/// Manages nullspace entities that are treated as members of map/grid PVS chunks.
+/// </summary>
+public sealed partial class ChunkEntitySystem : EntitySystem
+{
+    public const int ChunkSize = MapGridComponent.DefaultChunkSize;
+    private static readonly EntProtoId ChunkEntityPrototype = "ChunkEntity";
+
+    [Dependency] private EntityQuery<MetaDataComponent> _metaQuery = default!;
```
ArtisticRoomba: Initializers not required

ArtisticRoomba (changes requested):
Would be nice if there were docs for the public API that shows client how to append info onto the chunks just to get them started.

Joe Biden I don't know PVS code.

Robust.Shared/GameStates/ChunkEntitySystem.cs
metalgearsloth: Is that even possible normalized ones; if it were non-normalized you could just divide by chunksize but idk if that works here (and having it normalized is probably better if we get double-precision worldpos).

PR #6709, opened 2026-07-02
Fix `EntitySystemSubscriptionsGeneratorAttributes` xmldocs

Tayrtahn:
Slight touchup to the xmldoc formatting on these attributes so they display correctly (tested in vscode).

- Adds `<summary>` and `<remarks>` tags. Without at least one, no documentation is shown on hover.
- Converts `<ul>` and `<li>` to `<list>` and `<item>` respectively. Vscode does not recognize the former elements and displays the list as a single line with no formatting. 
- Swapped `<c>` tags on method references with `<see>` tags pointing to the referenced methods. This makes it easier to control-click through to see more documentation.


slarticodefast (approved):
Good stuff

Issue #3933, opened 2023-04-15
MoveEvent subscription needs its own grafana category


metalgearsloth:
The physics solver itself isn't super slow but right now a bunch of systems that sub to it just get their performance slapped under physics.

PR #6712, opened 2026-07-03
Tracy: Add physics update profiling zones

rebaserHEAD:
Adds profiling zones to the physics update so it isn't one opaque block in Tracy / the in-game profiler. Splits it into PreSolve, Broadphase, Collide, Solve and PostSolve, with Solve further split into island construction vs the solve and Collide's manifold job as its own zone.

Also attributes each VirtualController's before/after-solve cost to the controller by name instead of it all landing under physics, which is #3933.

All main-thread `_prof.Group`, no behaviour change.


Robust.Shared/Physics/Systems/SharedPhysicsSystem.Island.cs
```diff
@@ -304,6 +304,24 @@ private void ClearForces()
...
     private void Solve(float frameTime, float dtRatio, float invDt, bool prediction)
+    {
+        List<IslandData> islands;
+        using (_prof.Group("Build Islands"))
+        {
+            islands = BuildIslands(prediction);
+        }
+
+        SolveIslands(islands, frameTime, dtRatio, invDt, prediction);
+
```
metalgearsloth, line 318: It kind of mixes between wrapping methods in _prof.Group or putting it in the method, would be better to be more consistent with the approach.

Robust.Shared/Physics/Controllers/VirtualController.cs
```diff
@@ -15,9 +16,13 @@ public abstract partial class VirtualController : EntitySystem
     {
         [Dependency] protected SharedPhysicsSystem PhysicsSystem = default!;
         [Dependency] protected SharedTransformSystem TransformSystem = default!;
+        [Dependency] private ProfManager _prof = default!;
 
         private static readonly Stopwatch Stopwatch = new();
 
+        // Cached alongside the histogram labels to skip the virtual GetType() call on the hot path.
```
metalgearsloth: "hot"

metalgearsloth (changes requested):

Robust.Shared/Physics/Systems/SharedPhysicsSystem.Island.cs
rebaserHEAD: Ope. Probably should have fixed this as well. I'll get it sorted, baus.

rebaserHEAD: Sorted.

Robust.Shared/Physics/Controllers/VirtualController.cs
rebaserHEAD: Yeaaah. I removed the comment lol. Turns out not that hot.

Issue #6701, opened 2026-07-01
Engine needs more performance logging


metalgearsloth:
See https://github.com/space-wizards/RobustToolbox/pull/6686 now that tracy is supported.

Our logging is very insufficient, especially around clientgamestatemanager, rendering, and PVS, all could do with more so we have a better breakdown of what's going on.

PR #6713, opened 2026-07-03
Tracy: Add PVS game state profiling zones

rebaserHEAD:
Adds profiling zones to the PVS game state send pipeline, which was mostly dark before (only `PvsSystem.Update` showed up, not the actual `SendGameStates` work). Server side of #6701.

Adds a top-level PVS zone plus one per phase matching the existing Prometheus histogram labels, and splits each session's serialize work into computing the visible set and serializing it.

The per-session serialize/send runs in `Parallel.For` and the prof buffer is single-threaded, so those use a new `BeginThreadZone` helper that emits a Tracy-only zone without touching the buffer. No behaviour change.

### DIFF
--- a/Robust.Shared/Profiling/ProfManager.Tracy.cs
+++ b/Robust.Shared/Profiling/ProfManager.Tracy.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Runtime.CompilerServices;
 using bottlenoselabs.C2CS.Runtime;
 using Robust.Shared.Maths;
 using static Tracy.PInvoke;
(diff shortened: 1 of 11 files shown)


Robust.Shared/Profiling/ProfManager.Tracy.cs
```diff
@@ -123,6 +124,25 @@ private static TracyProfilerZone BeginTracyZone(string name, int lineNumber, Col
...
         return new TracyProfilerZone(context);
     }
+
+    /// <summary>
+    /// Begins a Tracy-only zone that is safe to call from any thread. Unlike <see cref="Group"/> this does NOT
+    /// write to the profiling buffer, so it will only show up in Tracy and not the in-game profiler. Use it
+    /// inside parallel regions (e.g. parallel job bodies) where the single-threaded prof buffer would
+    /// otherwise be corrupted. Returns null when Tracy is disabled.
+    /// </summary>
+    internal TracyProfilerZone? BeginThreadZone(
```
HeyBanditoz: I think it'd be better API-wise if which ProfManager method to use wasn't a concern of the caller but is handled for us in the GroupGuard struct ctor instead. Something like (pseudocode)
```cs
if (Thread.current() == "main")
    _sampler = ProfSampler.StartNew();
// Tracy stuff...
```
might work.

rebaserHEAD: Yeah that tracks, Let me clean it up right quick.

rebaserHEAD: Just committed it, Should be much better

Robust.Shared/Profiling/ProfManager.cs
```diff
@@ -37,6 +41,7 @@ public sealed partial class ProfManager
 
     internal void Initialize()
     {
+        _owningThread = Thread.CurrentThread;
```
metalgearsloth: ProfManager isn't necessarily guaranteed to be initialized on each thread due to dependency collections FYI. Sometimes code just runs basic ops over arrays and has no dependencies to them.

rebaserHEAD: Yeah, good point. Moved the stamp to the game loop frame instead of Initialize, and made an unset owner allow the write so the no-game-loop paths still work.

rebaserHEAD:
Sorry, I'm new to upstream contributing. Is there anything else I need to do here that I've missed?

PR #6718, opened 2026-07-04
Fix: entity localization parent fallback

codebykoi:
When I working with a locale files for my fork, I found it troublesome when working with many different .ftl files, and my builds always were giving me errors about missing .desc and etc, not allowing me to run properly.

this PR its basically a QoL fixes when working with locales.

Keep in mind that its my first ever PR to RobustToolbox, so I may dont know some specific details about the codebase, but I hope this one works for everyone.

### DIFF
--- a/Robust.Shared/Prototypes/EntityPrototype.cs
+++ b/Robust.Shared/Prototypes/EntityPrototype.cs
@@ -36,6 +36,7 @@ namespace Robust.Shared.Prototypes
         private const int DEFAULT_RANGE = 200;
 
         [DataField("loc")]
+        [NeverPushInheritance]
         private Dictionary<string, string>? _locPropertiesSet;
 
         /// <summary>
@@ -51,12 +52,15 @@ namespace Robust.Shared.Prototypes
         /// </summary>
         /// <seealso cref="Name"/>
         [DataField("name")]
+        [NeverPushInheritance]
         public string? SetName { get; private set; }
 
         [DataField("description")]
+        [NeverPushInheritance]
         public string? SetDesc { get; private set; }
 
         [DataField("suffix")]
+        [NeverPushInheritance]
         public string? SetSuffix { get; private set; }
 
         [DataField("categories"), Access(typeof(PrototypeManager))]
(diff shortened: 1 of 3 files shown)


codebykoi:
found another bug, fixing it rn

codebykoi:
alright, added small tests, and added NeverPushInheritance attribute to the fields of EntityPrototype.
Funny enough, the comments were saying they are not supposed to be inherited way before I did the fix

codebykoi:
Test was failing because of the RCD, the recent commit should fixed the edge case when entity had no name at all, it just fallbacks to protoId instead

codebykoi:
I have no idea why test failed on RandomSpecies test. I literalyl have nothing to do with it.

Robust.Shared/Prototypes/EntityPrototype.cs
```diff
@@ -51,12 +52,15 @@ public sealed partial class EntityPrototype : IPrototype, IInheritingPrototype,
         /// </summary>
         /// <seealso cref="Name"/>
         [DataField("name")]
+        [NeverPushInheritance]
```
slarticodefast, line 55: won't this completely break almost all inherited entity prototype names, descriptions and suffixes in the entire content repo?

codebykoi: uh, doesnt break for me. At least on my locale, I can run it on en-US again. one second

codebykoi: > won't this completely break almost all inherited entity prototype names, descroptions and suffixes in the entire content repo?

Just checked in local build, the en-US entity names act the same as before

codebykoi:
and here we go, all tests passes

Robust.Shared/Prototypes/EntityPrototype.cs
codebykoi, line 34: its literally says here that fields should not be inherited

Robust.Shared/Prototypes/EntityPrototype.cs
slarticodefast: I mean for entities with inherited names etc, because this would stop them from being inherited

codebykoi: > I mean for entities with inherited names etc, because this would stop them from being inherited

No it doesnt

Content Client_zPAXNmpvCA
Cursor_kw65datjL1

slarticodefast: huh, nevermind then

PR #6719, opened 2026-07-04
Multiple post shaders support.

### LINKED ISSUE #5105: Support for multiple post shaders per sprite
At the moment I'm making one thing that requires a post shader to work properly. Unfortunately, the post shader is usually busy with the selection shader. I would like to be able to display several shaders.

Aviu00:
Closes #5105

Adds support for multiple post shaders per sprite. This will cause breaking changes, content pr will follow when this one leaves draft state.
This is almost finished (not entirely), but review would be appreciated. Note that I had to do some weird tricks to make this work.

It works by creating 2 render targets with same size per sprite (currently it caches all render targets. TODO smaller sprites should reuse rt crated for larger sprites). It draws sprite contents on first rt, then loops through all shader instances, copying rt contents to intermediate rt, then clearing first rt and drawing intermediate rt contents onto it with shader applied.
Before there was a trick to offset the viewport while rendering sprite to make sure lighting application works, this new implementation does the same thing but with multiple render targets and ping pong rendering this leads to some issues: We are drawing shaders from small rt onto small rt, before we were drawing  shaders from small rt onto large viewport. This means before we used normal viewport when drawing shaders, now we are using offseted viewport. This leads to fragcoord being offseted for shaders, I fixed it by manually offseting it back by creating new FragCoordOffset uniform for base fragment shader.

Another problem is lighting application. It (1) renders light/shadows over and over on each shader iteration, and (2) renders shadows again when drawing from rt onto viewport. It is fixed by 1: overriding shader blend mode to None, 2: disabling lightingReady temporarily (see my comments in the code)

Post shaders are stored in SortedDictionary<SpriteShaderKey, SpriteShaderData>. It is sorted by shader render order defined in shaderkey and hashed by unique id stored in shaderkey. This can be yamlable, see media. TODO allow yaml to specify id for key (currently it always generates random one form yaml) so that entsys can access it if needed.

Also made BeforePostShaderRenderEvent byref

<details>
<summary> Media </summary>

Outline + water transparency shaders:

2026-07-04-102740_1920x1080_scrot

Some more showcase with 3 custom shaders + outline, different render order defined in yaml:

https://github.com/user-attachments/assets/715d0e94-1686-46d4-b9e9-9066e350870f

</details>

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -202,34 +202,10 @@ namespace Robust.Client.GameObjects
         [ViewVariables(VVAccess.ReadWrite)] internal bool _inertUpdateQueued;
 
         /// <summary>
-        ///     Shader instance to use when drawing the final sprite to the world.
-        /// </summary>
-        [ViewVariables(VVAccess.ReadWrite)]
-        public ShaderInstance? PostShader
-        {
-            get;
-            // This will get obsoleted, but I only want to mark it as obsolete when multi-shader support is added, so
-            // that people can use the appropriate method and don't migrate to an incorrect new method that wont
-            // be obsoleted.
-            set;
-        }
-
-        /// <summary>
-        ///     Whether to pass the screen texture to the <see cref="PostShader"/>.
-        /// </summary>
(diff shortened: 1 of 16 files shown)


Aviu00:
#6817

PR #6720, opened 2026-07-04
Add methods to support generating color palettes

mqole:
adds methods to `Color.cs` that support generating:
- complementary colors
- split complementary colors
- triadic palettes

image

(ANALOGOUS NOT INCLUDED)
EDIT: ANALOGOUS NOW INCLUDED

this pr is required for https://github.com/space-wizards/space-station-14/pull/38528.

### DIFF
--- a/Robust.Shared.Maths/Color.cs
+++ b/Robust.Shared.Maths/Color.cs
@@ -589,69 +589,6 @@ namespace Robust.Shared.Maths
             return new Vector4(hue, saturation, max, rgb.A);
         }
 
-        /// <summary>
-        ///     Generates a list of triadic complementary colors
-        /// </summary>
-        public static Color[] GetTriadicComplementaries(Color color)
-        {
-            return GetComplementaryColors(color, 0.120f);
-        }
-
-        /// <summary>
-        ///     Generates a list of split complementary colors
-        /// </summary>
-        public static Color[] GetSplitComplementaries(Color color)
-        {
-            return GetComplementaryColors(color, 0.150f);
-        }
-
-        /// <summary>
-        ///     Generates a list containing the base color and two copies of a single complementary color
-        /// </summary>
-        public static Color[] GetOneComplementary(Color color)
-        {
-            return GetComplementaryColors(color, 0.180f);
-        }
-
-        /// <summary>
-        ///    Generates a complementary colour palette for a provided
-        ///    colour by rotating a set amount of degrees around the
-        ///    colour wheel, and then varying the value and saturation
-        ///    slightly.
-        /// </summary>
-        /// <returns>
-        ///     A list of 3 colors.
-        /// </returns>
-        private static Color[] GetComplementaryColors(Color color, float angle)
-        {
-            var hsl = ToHsl(color);
-            var random = new Random();
-            // sorry about how messy these are, but to get all random values we need to reroll for positive and negative HSL.
-            // since we want to rotate x degrees around the colour wheel, we need to do so in both directions- doing x + x degrees will give us the wrong hue!
-
-            var hVal = hsl.X + angle;
-            hVal -= MathF.Floor(hVal);
-            var positiveHSL = new Vector4(
-                hVal,
-                MathHelper.Clamp01(hsl.Y + random.Next(-20 / 100, 0)),
-                MathHelper.Clamp01(hsl.Z + random.Next(-15 / 100, 16/ 100)),
-                hsl.W);
-
-            var hVal1 = hsl.X - angle;
-            hVal1 += hVal1 <= 0f ? hVal1 + 0.360f : hVal1;
-            var negativeHSL = new Vector4(
-                hVal1,
-                MathHelper.Clamp01(hsl.Y + random.Next(-20 / 100, 0)),
-                MathHelper.Clamp01(hsl.Z + random.Next(-15 / 100, 16 / 100)),
-                hsl.W);
-
-            var c0 = FromHsl(positiveHSL);
-            var c1 = FromHsl(negativeHSL);
-
-            var palette = new Color[] { color, c0, c1 };
-            return palette;
-        }
-
         #region Oklab/Oklch
         /*
 
--- a/Robust.Shared.Maths/Color.cs
+++ b/Robust.Shared.Maths/Color.cs
@@ -589,6 +589,69 @@ namespace Robust.Shared.Maths
             return new Vector4(hue, saturation, max, rgb.A);
         }
 
+        /// <summary>
+        ///     Generates a list of triadic complementary colors
+        /// </summary>
+        public static List<Color> GetTriadicComplementaries(Color color)
+        {
+            return GetComplementaryColors(color, 0.120f);
+        }
+
+        /// <summary>
+        ///     Generates a list of split complementary colors
+        /// </summary>
+        public static List<Color> GetSplitComplementaries(Color color)
+        {
+            return GetComplementaryColors(color, 0.150f);
+        }
+
+        /// <summary>
+        ///     Generates a list containing the base color and two copies of a single complementary color
+        /// </summary>
+        public static List<Color> GetOneComplementary(Color color)
+        {
+            return GetComplementaryColors(color, 0.180f);
+        }
+
+        /// <summary>
+        ///    Generates a complementary colour palette for a provided
+        ///    colour by rotating a set amount of degrees around the
+        ///    colour wheel, and then varying the value and saturation
+        ///    slightly.
+        /// </summary>
+        /// <returns>
+        ///     A list of 3 colors.
+        /// </returns>
+        private static List<Color> GetComplementaryColors(Color color, float angle)
+        {
+            var hsl = ToHsl(color);
+            var random = new Random();
+            // sorry about how messy these are, but to get all random values we need to reroll for positive and negative HSL.
+            // since we want to rotate x degrees around the colour wheel, we need to do so in both directions- doing x + x degrees will give us the wrong hue!
+
+            var hVal = hsl.X + angle;
+            hVal -= MathF.Floor(hVal);
+            var positiveHSL = new Vector4(
+                hVal,
+                MathHelper.Clamp01(hsl.Y + random.Next(-20 / 100, 0)),
+                MathHelper.Clamp01(hsl.Z + random.Next(-15 / 100, 16/ 100)),
+                hsl.W);
+
+            var hVal1 = hsl.X - angle;
+            hVal1 += hVal1 <= 0f ? hVal1 + 0.360f : hVal1;
+            var negativeHSL = new Vector4(
+                hVal1,
+                MathHelper.Clamp01(hsl.Y + random.Next(-20 / 100, 0)),
+                MathHelper.Clamp01(hsl.Z + random.Next(-15 / 100, 16 / 100)),
+                hsl.W);
+
+            var c0 = FromHsl(positiveHSL);
+            var c1 = FromHsl(negativeHSL);
+
+            var palette = new List<Color> { color, c0, c1 };
+            return palette;
+        }
+
         #region Oklab/Oklch
         /*
 
--- a/Robust.Shared.Maths/Color.cs
+++ b/Robust.Shared.Maths/Color.cs
@@ -592,7 +592,7 @@ namespace Robust.Shared.Maths
         /// <summary>
         ///     Generates a list of triadic complementary colors
         /// </summary>
-        public static List<Color> GetTriadicComplementaries(Color color)
+        public static Color[] GetTriadicComplementaries(Color color)
         {
             return GetComplementaryColors(color, 0.120f);
         }
@@ -600,7 +600,7 @@ namespace Robust.Shared.Maths
         /// <summary>
         ///     Generates a list of split complementary colors
         /// </summary>
-        public static List<Color> GetSplitComplementaries(Color color)
+        public static Color[] GetSplitComplementaries(Color color)
         {
             return GetComplementaryColors(color, 0.150f);
         }
@@ -608,7 +608,7 @@ namespace Robust.Shared.Maths
         /// <summary>
         ///     Generates a list containing the base color and two copies of a single complementary color
         /// </summary>
-        public static List<Color> GetOneComplementary(Color color)
+        public static Color[] GetOneComplementary(Color color)
         {
             return GetComplementaryColors(color, 0.180f);
         }
@@ -622,7 +622,7 @@ namespace Robust.Shared.Maths
         /// <returns>
         ///     A list of 3 colors.
         /// </returns>
-        private static List<Color> GetComplementaryColors(Color color, float angle)
+        private static Color[] GetComplementaryColors(Color color, float angle)
         {
             var hsl = ToHsl(color);
             var random = new Random();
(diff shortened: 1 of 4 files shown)


AJCM-git (changes requested):
return a color array

Robust.Shared.Maths/Color.cs
```diff
@@ -589,6 +589,69 @@ public static Vector4 ToHsv(Color rgb)
...
+        ///    colour wheel, and then varying the value and saturation
+        ///    slightly.
+        /// </summary>
+        /// <returns>
+        ///     A list of 3 colors.
+        /// </returns>
+        private static List<Color> GetComplementaryColors(Color color, float angle)
+        {
+            var hsl = ToHsl(color);
+            var random = new Random();
```
AJCM-git: After thinking some more about it, pulling in robustrandom will be necessary as we dont wanna depend on dotnet's version anywhere

Robust.Shared/ColorExtensions/ColorExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// <summary>
+    ///    Generates a complementary colour palette for a provided
+    ///    colour by rotating a set amount of degrees around the
+    ///    colour wheel, and then varying the value and saturation
+    ///    slightly.
+    /// </summary>
+    /// <returns>
+    ///     A list of 3 colors.
+    /// </returns>
+    private static Color[] GetComplementaryColors(Color color, float angle)
```
dffdff2423: Very minor nit but I don't see a reason to make this private?

dffdff2423 (approved):

Robust.Shared/Utility/ColorExtensions.cs
AJCM-git, line 1: Move this to shared.utility

Robust.Shared/ColorExtensions/ColorExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Robust.Shared.ColorExtensions;
+
+public static class ColorExtensions
+{
+
+    /// <summary>
+    ///     Generates a list of triadic complementary colors
+    /// </summary>
+    public static Color[] GetTriadicComplementaries(Color color)
```
dffdff2423: I completely missed this. These methods need to be marked properly to be used as extension methods
```suggestion
    public static Color[] GetTriadicComplementaries(this Color color)
```
Should be similar for all the later ones.

dffdff2423 (changes requested):
See below.

I think you should also maybe add a basic unit test or two but that would be something up to the maintainers.

Robust.Shared/ColorExtensions/ColorExtensions.cs
AJCM-git:
```suggestion
    public static Color[] GetTriadicComplementaries(this Color color)
```
and you can do 
var color = new Color();
color.GetTriadicComplementaries()

Same for other public methods

Robust.Shared/ColorExtensions/ColorExtensions.cs
AJCM-git: cant this be public?

AJCM-git (changes requested):

Robust.Shared/ColorExtensions/ColorExtensions.cs
mqole: missed that one!

AJCM-git (approved):

Robust.Shared/Utility/ColorExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// <summary>
+    ///     Generates a list containing the base color and two copies of a single complementary color
+    /// </summary>
+    public static Color[] GetOneComplementary(this Color color)
+    {
+        return GetComplementaryColors(color, 0.180f);
+    }
+
+    /// <summary>
+    ///    Generates a complementary colour palette for a provided
```
metalgearsloth:
```suggestion
    ///    Generates a complementary color palette for a provided
```

I also spell it this way but american codebase.

Robust.Shared/Utility/ColorExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+    ///    colour by rotating a set amount of degrees around the
+    ///    colour wheel, and then varying the value and saturation
+    ///    slightly.
+    /// </summary>
+    /// <returns>
+    ///     A list of 3 colors.
+    /// </returns>
+    public static Color[] GetComplementaryColors(Color color, float angle)
```
metalgearsloth: Is it possible to use the Angle struct for this? The performance diff should be negligible but makes it more standard.

metalgearsloth (changes requested):
Could also really do with a test.

Robust.Shared/Utility/ColorExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+        var hVal = hsl.X + angle;
+        hVal -= MathF.Floor(hVal);
+        var positiveHSL = new Vector4(
+            hVal,
+            MathHelper.Clamp01(hsl.Y + random.Next(-20 / 100, 0)),
+            MathHelper.Clamp01(hsl.Z + random.Next(-15 / 100, 16/ 100)),
+            hsl.W);
+
+        var hVal1 = hsl.X - angle;
+        hVal1 += hVal1 <= 0f ? hVal1 + 0.360f : hVal1;
```
whatston3:
```suggestion
        hVal1 += hVal1 <= 0f ? hVal1 + 1.0f : hVal1;
```
Should wrap around at 1 (hue being a value in [0-1]), not 0.36, no?

Robust.Shared/Utility/ColorExtensions.cs
whatston3: Also not an angle, it's just delta hue :(
Might make sense to rename it if not using actual angles.

Robust.Shared/Utility/ColorExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+
+public static class ColorExtensions
+{
+
+    /// <summary>
+    ///     Generates a list of triadic complementary colors
+    /// </summary>
+    public static Color[] GetTriadicComplementaries(this Color color)
+    {
+        return GetComplementaryColors(color, 0.120f);
```
whatston3:
```suggestion
        return GetComplementaryColors(color, 0.33333f);
```
+/- 1/3. 120 degrees, 0.333 over hue (or even better, define them all as constants at the top of your file from their respective angles :))

Robust.Shared/Utility/ColorExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /// <summary>
+    ///     Generates a list of split complementary colors
+    /// </summary>
+    public static Color[] GetSplitComplementaries(this Color color)
+    {
+        return GetComplementaryColors(color, 0.150f);
```
whatston3:
```suggestion
        return GetComplementaryColors(color, 0.41667f);
```
+/- 5/12. 150 degrees, 0.4166... over hue.

Robust.Shared/Utility/ColorExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /// <summary>
+    ///     Generates a list containing the base color and two copies of a single complementary color
+    /// </summary>
+    public static Color[] GetOneComplementary(this Color color)
+    {
+        return GetComplementaryColors(color, 0.180f);
```
whatston3:
```suggestion
        return GetComplementaryColors(color, 0.5f);
```
+/- 1/2. 180 degrees.

[Pardon the comment spam]

Robust.Shared/Utility/ColorExtensions.cs
mqole: i am saying unkind things about america in my mind.......................

Robust.Shared/Utility/ColorExtensions.cs
mqole: cheers for the specifics, i figured there was a better way to do it but maths is scary

mqole:
i think i did the angles properly, will need someone to check working.
EDIT: i do not think i did the angles properly

no idea how to write integration test. tried looking for examples in the codebase as well as slam's unmerged tutorial, but the tutorial is more geared towards content and all the codebase examples in robust i found were either 7 years old, inconsistently formatted to each other, or both.

EDIT 2: never mind im basically smart

Robust.Shared/Utility/ColorExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+using Robust.Shared.Random;
+using Robust.Shared.Maths;
+
+namespace Robust.Shared.Utility;
+
+public static class ColorExtensions
+{
+    private static readonly float TriadicHueDelta = 0.33333f; // +/- 1/3. 120 degrees, 0.333 over hue
+    private static readonly float SplitComplementaryHueDelta = 0.41667f; // +/- 5/12. 150 degrees, 0.4166... over hue
+    private static readonly float ComplementaryHueDelta = 0.5f; // +/- 1/2. 180 degrees
```
whatston3:
```suggestion
    // Hue deltas, converting from degrees to [0-1]
    public static readonly float TriadicHueDelta = 120.0f / 360.0f;
    public static readonly float SplitComplementaryHueDelta = 150.0f / 360.0f;
    public static readonly float ComplementaryHueDelta = 180.0f / 360.0f;
```
I think it was casting the result of integer division when you tried previously.  This should work, should be more precise, and seems more readable to my eyes.

If they're public, then you can use them in testing, though that might be relying on implementation details to test the implementation.

Robust.Shared/Utility/ColorExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+        var hVal = hsl.X + hueDelta;
+        hVal -= MathF.Floor(hVal);
+        var positiveHSL = new Vector4(
+            hVal,
+            MathHelper.Clamp01(hsl.Y + random.Next(-20 / 100, 0)),
+            MathHelper.Clamp01(hsl.Z + random.Next(-15 / 100, 16 / 100)),
+            hsl.W);
+
+        var hVal1 = hsl.X - hueDelta;
+        hVal1 += hVal1 <= 0f ? 1f : 0f;
```
whatston3:
```suggestion
        if (hval1 < 0f)
            hVal1 += 1f;
```
I know it's a one-liner as is, but this seems easier to read in my eyes.
Those that know better may disagree.

Robust.Shared.Tests/Utility/ColorExtensionsTest.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        var palette = ColorExtensions.GetSplitComplementaries(Color.Red);
+
+        using (Assert.EnterMultipleScope())
+        {
+            Assert.That(palette, Has.Length.EqualTo(3));
+            Assert.That(MathHelper.CloseToPercent(palette[0], Color.Red));
+
+            Assert.That(Color.ToHsl(palette[1]).X, Is.EqualTo(0.41667f));
+            Assert.That(Color.ToHsl(palette[2]).X, Is.InRange(0.58333f, 0.58334f)); // TODO FIX THIS
```
whatston3: This is true for all of the cases: couldn't you check MathHelper.CloseToPercent instead of Is.EqualTo/Is.InRange?

Also, because you're starting from a hue of 0 (which you could also maybe check), your values should be predictably at your constants - at `ColorExtensions.XHueDelta` and `1f - ColorExtensions.XHueDelta`.

```suggestion
            var firstHue = Color.ToHsl(palette[1]).X;
            var secondHue = Color.ToHsl(palette[2]).X;
            Assert.That(MathHelper.CloseToPercent(firstHue, ColorExtensions.SplitComplementaryHueDelta));
            Assert.That(MathHelper.CloseToPercent(secondHue, 1f - ColorExtensions.SplitComplementaryHueDelta));
```

Not great at testing best practices, so take this with a grain of salt.

whatston3 (changes requested):
A few more thoughts.

I think the constants based on the angles you had before should work better (slightly more precise).  The values you're testing could be based off of your constants, too.

Haven't tested the suggestions, all webedit ops, so do take care.

Robust.Shared.Tests/Utility/ColorExtensionsTest.cs
mqole: ive literally never used assert before so theyre better practices than i'd know :p

mqole:
image
what i love to see!

whatston3 (approved):
Reads fine to me, pardon the annoyance.

[Also, didn't say it yet but this is a very nice PR.  Good stuff.]

Robust.Shared/Utility/ColorExtensions.cs
```diff
@@ -7,10 +7,19 @@ namespace Robust.Shared.Utility;
 
 public static class ColorExtensions
 {
+    public static readonly float AnalogousHueDelta = 45f / 360f; // +/- 1/8. 45 degrees, 0.125 over hue
```
whatston3: Not to be a dingus, but between the diagram in the PR and this https://www.sessions.edu/color-calculator/, it looks more like 30 degrees.

If 45 looks great though, :+1:

mqole: i dont think it has an exact value (unlike the other 3 methods), but sure 😝

Robust.Shared.Tests/Utility/ColorExtensionsTest.cs
```diff
@@ -0,0 +1,75 @@
+using NUnit.Framework;
+using Robust.Shared.Maths;
+using Robust.Shared.Utility;
+
+namespace Robust.Shared.Tests.Utility;
+
+[TestFixture]
```
Tayrtahn: You can omit this attribute - modern NUnit versions automatically infer it for classes with test methods.

Robust.Shared.Tests/Utility/ColorExtensionsTest.cs
```diff
@@ -0,0 +1,75 @@
...
+        var palette = ColorExtensions.GetAnalogousComplementaries(Color.Red);
+
+        using (Assert.EnterMultipleScope())
+        {
+            Assert.That(palette, Has.Length.EqualTo(3));
+            Assert.That(MathHelper.CloseToPercent(palette[0], Color.Red));
+
+            Assert.That(Color.ToHsl(palette[0]).X, Is.Zero);
+            Assert.That(MathHelper.CloseToPercent(Color.ToHsl(palette[1]).X, ColorExtensions.AnalogousHueDelta));
+            Assert.That(MathHelper.CloseToPercent(Color.ToHsl(palette[2]).X, 1f - ColorExtensions.AnalogousHueDelta));
```
Tayrtahn: I know it goes against all your programming instincts, but you generally should hardcode your expected results in unit tests. Since you're testing `ColorExtensions`, you shouldn't rely on values from `ColorExtensions` to calculate the expected values - if the delta value gets set to some incorrect value, the test would still pass, which kinda defeats the purpose! I would just hardcode the expected color values.

whatston3: My apologies here, then.

mqole: makes sense, good to know!

Tayrtahn (approved):

dffdff2423 (approved):

PR #5663, opened 2025-02-07
Clean up uses of obsolete/EC TransformComponent properties/methods

TemporalOroboros:
This should clear out the last of the references to stuff like the `Anchored` setter or the various `WorldThing` getters.

Also adds a basic `SharedTransformSystem` API for `LocalMatrix`, `InvLocalMatrix` and `NoLocalRotation`.

PR #6721, opened 2026-07-04
Cleanup: Warning resolution/suppression

whatston3:
## About the PR
This PR resolves a handful of simpler compiler warnings.  Picks up from #5663 in spirit, though slightly broader in the scope of warnings resolved, focused _only_ on resolving warnings, and with no new API calls.

Warning count on complete solution compilation went from 201 to 151 against master at the time of opening.

Wink and a nod to https://github.com/space-wizards/space-station-14/issues/33279

## Technical details

The lion's share of the warnings involved TransformSystem setters for LocalRotation.

A few instances of warnings were suppressed with pragma blocks, though I've tried to explain why in the majority.

Uses of Anchored setter in AnchoredSystemTests left as-is, wasn't particularly confident in suppressing them with a pragma.

Tests passed locally before PR.

image

## Breaking changes

* Obsoleted TileDefManager.GetVariantTile versions using System.Random.

### DIFF
--- a/Robust.Client.WebView/Cef/WebViewManagerCef.Control.cs
+++ b/Robust.Client.WebView/Cef/WebViewManagerCef.Control.cs
@@ -7,6 +7,7 @@ using Robust.Client.UserInterface;
 using Robust.Shared.IoC;
 using Robust.Shared.Log;
 using Robust.Shared.Maths;
+using Robust.Shared.Prototypes;
 using Robust.Shared.Utility;
 using SixLabors.ImageSharp.PixelFormats;
 using Xilium.CefGlue;
@@ -19,16 +20,18 @@ namespace Robust.Client.WebView.Cef
     {
         private readonly List<ControlImpl> _activeControls = new();
 
+        private static readonly ProtoId<ShaderPrototype> _bgraShader = "bgra";
+
         public IWebViewControlImpl MakeControlImpl(WebViewControl owner)
         {
-            var shader = _prototypeManager.Index<ShaderPrototype>("bgra");
+            var shader = _prototypeManager.Index(_bgraShader);
             var shaderInstance = shader.Instance();
             var impl =  new ControlImpl(this, owner, shaderInstance);
             _dependencyCollection.InjectDependencies(impl);
             return impl;
         }
 
-        private sealed class ControlImpl : IWebViewControlImpl
+        private sealed partial class ControlImpl : IWebViewControlImpl
         {
             private static readonly Dictionary<Key, CefKeyCodes.ChromiumKeyboardCode> KeyMap = new()
             {
--- a/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
 using System.Numerics;
 using JetBrains.Annotations;
 using Robust.Shared.Animations;
@@ -8,7 +7,6 @@ using Robust.Shared.GameStates;
 using Robust.Shared.IoC;
 using Robust.Shared.Map;
 using Robust.Shared.Maths;
-using Robust.Shared.Physics;
 using Robust.Shared.Serialization.Manager.Attributes;
 using Robust.Shared.Timing;
 using Robust.Shared.Utility;
@@ -135,8 +133,10 @@ namespace Robust.Shared.GameObjects
             get => _noLocalRotation;
             set
             {
+#pragma warning disable CS0618
                 if (value)
                     LocalRotation = Angle.Zero;
+#pragma warning restore CS0618
 
                 _noLocalRotation = value;
                 _entMan.Dirty(Owner, this);
@@ -528,7 +528,9 @@ namespace Robust.Shared.GameObjects
 
         public string GetDebugString()
         {
+#pragma warning disable CS0618 // Accessing world coords in debug function.
             return $"pos/rot/wpos/wrot: {Coordinates}/{LocalRotation}/{WorldPosition}/{WorldRotation}";
+#pragma warning restore CS0618
         }
     }
 
--- a/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
+++ b/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
@@ -131,7 +131,7 @@ namespace Robust.Shared.GameObjects
         public bool NoLocalRotation
         {
             get => _noLocalRotation;
-            [Obsolete("Use SharedTransformSystem.SetNoLocalRotation")]
+            [Obsolete("Use SharedTransformSystem.SetNoLocalRotation() instead")]
             set
             {
                 if (value)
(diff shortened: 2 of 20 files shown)


Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -447,7 +447,9 @@ public void SetLocalRotationNoLerp(EntityUid uid, Angle value, TransformComponen
         if (!XformQuery.Resolve(uid, ref xform))
             return;
 
+#pragma warning disable CS0618
```
ArtisticRoomba: CC @metalgearsloth but these should probably stay unsurpressed and the methods deprecated since people should be using xformsys.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
```diff
@@ -887,7 +889,9 @@ internal void OnHandleState(EntityUid uid, TransformComponent xform, ref Compone
             }
             else
             {
+#pragma warning disable CS0618 // AnchorEntity/Unanchored can't be used from uninitialized states.
```
ArtisticRoomba, line 892: Ditto to an extent.

Robust.Client.WebView/Cef/WebViewManagerCef.Control.cs
```diff
@@ -19,16 +20,18 @@ internal partial class WebViewManagerCef
...
         {
-            var shader = _prototypeManager.Index<ShaderPrototype>("bgra");
+            var shader = _prototypeManager.Index(_bgraShader);
             var shaderInstance = shader.Instance();
             var impl =  new ControlImpl(this, owner, shaderInstance);
             _dependencyCollection.InjectDependencies(impl);
             return impl;
         }
 
-        private sealed class ControlImpl : IWebViewControlImpl
```
ArtisticRoomba, line 31: I don't know why you made this change

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -135,8 +133,10 @@ public bool NoLocalRotation
             get => _noLocalRotation;
             set
             {
+#pragma warning disable CS0618
```
ArtisticRoomba: CC @metalgearsloth

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
```diff
@@ -528,7 +528,9 @@ public void RebuildMatrices()
 
         public string GetDebugString()
```
ArtisticRoomba, line 529: Might want to just move/dupe this into system and deprecate this method instead.

ArtisticRoomba (changes requested):

ArtisticRoomba:
Yeah Sloth got back to me and he said don't supress xformcomp ext. deprecations. Everything else looks good other than those things that I saw.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
whatston3: This _is_ that function.  The internal set needs to be suppressed the same way that SetLocalPosition is.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
whatston3: AnchorEntity/Unanchor (typo...) genuinely don't work on uninitialized entities though.  If they should, that's a good change, but that seems a touch out of scope.

This is the cause of the flailing here: https://github.com/space-wizards/space-station-14/pull/44494 and the test failures on [63d9da3](https://github.com/space-wizards/space-station-14/pull/44494/commits/63d9da3d53f9cc05b1418c836c646cb19a0bc8f8) and [430714c](https://github.com/space-wizards/space-station-14/pull/44494/commits/430714cdf348e5e03c308829e1a8c9b3fb4af8ac)

Robust.Client.WebView/Cef/WebViewManagerCef.Control.cs
whatston3: ~~Offhand I believe it was suggested to be unused.  Will revert.~~  [Sorry, should've looked at the file view]

Image

It's a warning.  It's resolving that warning.

whatston3:
> Yeah Sloth got back to me and he said don't supress xformcomp ext. deprecations. Everything else looks good other than those things that I saw.

Still firmly disagree with the one internal to SetLocalRotation.  SetLocalPosition does the same thing, and it should, because you can't call SetLocalRotation from SetLocalRotation.

Otherwise I'll make those changes.  Cheers.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
ArtisticRoomba: Ah, :trollface:. In this instance and from what I've been told already I would keep the warning unsurpressed and mark a TODO that the xformcomp code needs to be brought into sys instead of the sys itself piggybacking off of it.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
ArtisticRoomba: It's probably best to unsurpress the warning and leave a TODO that says that the behavior in that anchor state set is reliant on being able to use it in an uninitted state. Definitely a out of scope fix for this PR but I don't really want to silence the issue.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
whatston3: Then should the SetLocalPosition warning be unsuppressed as well?

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
whatston3: Obsoleting this function instead vs. SharedTransformSystem.NoLocalRotation.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
whatston3: > It's probably best to unsurpress the warning and leave a TODO that says that the behavior in that anchor state set is reliant on being able to use it in an uninitted state. Definitely a out of scope fix for this PR but I don't really want to silence the issue.

Makes more sense to me to prefer to just open two issues - one for the Anchor/Unanchor functions and another for maybe an Access attribute on TransformComponent?

Zero warnings means zero warnings.

[Edit: issues created, #6738, #6739]

Robust.Client.WebView/Cef/WebViewManagerCef.Control.cs
ArtisticRoomba: I see, silly me. On my local I reverted this change and no compiler warning came up so I questioned why.

Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs
whatston3: Should be done.

whatston3: Retrospectively, this is a bit silly - this is only used at the moment (at least in SS14) via the IComponentDebug interface in the `entfo` command.

Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs
metalgearsloth: I have a PR that nukes these and fixes the system API so probably just leave it.

PR #6723, opened 2026-07-05
Chat Message Spacing Offset

Alecksohs:
## About the PR
This PR adds a public message spacing offset variable to the Output Panel.

## Technical details

The current chatbox renders using not controls but a bespoke custom drawpass with Rich Text Entries. This has an issue of pixel rounding on spacing, where at smaller font sizes, you will have issues regarding pixel spacing. A message spacing offset is added to the previous GetLineSeparation to add the float to the spacing. It's rather simple.

## Breaking changes

None.

### DIFF
--- a/Robust.Client/UserInterface/Controls/OutputPanel.cs
+++ b/Robust.Client/UserInterface/Controls/OutputPanel.cs
@@ -23,6 +23,8 @@ namespace Robust.Client.UserInterface.Controls
 
         public const string StylePropertyStyleBox = "stylebox";
 
+        public float LineSpacing { get; set; }
+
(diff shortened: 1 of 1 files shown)


metalgearsloth:
image

Does this fix this

Alecksohs:
This will not 'fix' the fundemental issues if the renderer's spacing being odd due to pixel alignment. It just allows you to, for every message, increase the total spacing.

chavonadelal:
It seems to me that our work overlaps in some way, but I am surprised by the insignificance of your changes

https://github.com/space-wizards/RobustToolbox/pull/6393

PR #6726, opened 2026-07-06
Workflow standardisation + pruning

metalgearsloth:
See https://github.com/space-wizards/space-station-14/pull/44532


metalgearsloth:
See if it works as intended worst case we reverty.

PR #6733, opened 2026-07-06
refactor: xml-doc + helper method for audio device display names

Fildrance:
If audio device get method is invoked and OS have other default codepage then UTF-8 - it can show stuff in wrong encoding. For example having ru windows i got stuff in 'cp 1251' and dotnet just tried to display it as UTF-8 string, which resulted in this

image

To fix this we need method that decodes things and additional encoding providers (as by default only following are available in dotnet - "UTF-8, UTF-16, UTF-32, ASCII, ISO-8859-1").

As method will use Encoding.GetString / Encoding.GetBytes which are sandboxed, so it would be just easier to put them in RT.

Also added xml-doc to related code.


Fildrance:
I have sneaky suspicion that it can be brought into OpenAL but i am not sure it would worth it.

metalgearsloth:
Add Release Note new feature or I die.

Fildrance:
> Add Release Note new feature or I die.

It is done, master.

PR #6740, opened 2026-07-06
Add missing EntitySystemSubscriptionsGenerator.targets into Server.csproj

whatston3:
Content's server-side SubscribeLocalEvent attributes won't work without it.

As the simplest test, as-is, all attribute uses in Content.Server are flagged as unused by the IDE.


kontakt (approved):
Looks like an oversight from #6227.
Project builds and binary testing with breakpoints on `public sealed partial class PointLightSystem : SharedPointLightSystem` shows that the attribute does not work before and does work after this change.

ArtisticRoomba (approved):
We're so good at this development thing

PR #6749, opened 2026-07-07
CVar for code source info

Fildrance:
Bug report feature would need full info on sources that were used to run current instance of application, at least server side.

Added script to put more sources info into cvar, and related cvars (well documented i feel). 
build.hash already exists so no cvar for it.

As alternative to adding cvar data to `server_config.toml` - i could add `build.json`. But it seems like its generation with it was removed some time ago, and i lack knowledge of why, so not touching it :|

Also i am not editing client-side toml, relying on syncing kinda, but it is worth considering (for cases when somehow client will be reporting things and having separate version... not sure how we could end up in usch situation T_T).

Sample of scirpt run result:

```

[build]
# Auto-generated by CI pipeline. Do not edit manually.
fork_url = "https://github.com/Fildrance/space-station-14.git"
hash = "710a6bf61a50fb3446b45f74cd150fa55a016528"
branch = "feature/save-commit-id-in-config-on-release-build"
robust_toolbox_url = "https://github.com/space-wizards/RobustToolbox.git"
robust_toolbox_hash = "b9c96f60033ecac8cd711fc5157e9a262a66ad39"
robust_toolbox_branch = "feature/inject-build-info-into-cvar-on-publish"
```


Fildrance:
After merge we will just going to need to add 
```
      - name: Inject build info
        run: RobustToolbox/Tools/inject_build_info.py
        env:
          GITHUB_SERVER_URL: ${{ github.server_url }}
          GITHUB_REPOSITORY: ${{ github.repository }}
          GITHUB_SHA: ${{ github.sha }}
          GITHUB_REF_NAME: ${{ github.ref_name }}
```
after `Build Packaging` and before `Package server` steps
into 
* publish-testing.yml
* publish.yml
* test-packaging.yml

And ideally add display of this data somewhere in client UI :>

Simyon264:
This should be handled by RobustCDN. build.json is also currently made by RobustCDN.
https://github.com/space-wizards/Robust.Cdn/blob/0b2b814e4f16b0474f341bb0f041bd7ae9f5b34d/Robust.Cdn/Controllers/ForkPublishController.cs#L114

Fildrance:
@Simyon264 okay i found related code in CDN and changed it, changed RT code to only get more variables from build.json

Can you point me to GHA that is posting to RobustCDN so i can add new arguments there?

Simyon264:
It's publish.yml and publish-testing.yml iirc.

PR #6751, opened 2026-07-07
Send serialized PVS states asynchronously

metalgearsloth:
There was an old-ahh TODO for this.

Only bug I saw was CurTick not getting cached for when the server moves into the next tick and it trips a debug assert.


ArtisticRoomba:
`ParallelRobustJob` stays winning

metalgearsloth:
I broke my stacked branchies.

PR #6753, opened 2026-07-07
Make TableContainer virtual and public

whatston3:
Turning TableContainer into a public, virtual class to expose it content-side.

SS14 content-side PR: https://github.com/space-wizards/space-station-14/pull/44548

Part of https://github.com/space-wizards/space-station-14/issues/44537 resolution.


Princess-Cheeseballs (approved):
Why even port this to engine and then make it internal. What was even the point.
Thanks for the cleanup.

PR #6755, opened 2026-07-07
File Dialog: Expose file name of the selected file

AJCM-git:
Cleans up FileDialog and adds functions to get the name of the file from the dialog. 
Previously we only exposed the file stream so content had no way of knowing the name of the file, now we also expose a sanitized filename

No breaking changes

### DIFF
--- a/Robust.Client/UserInterface/FileDialogManager.cs
+++ b/Robust.Client/UserInterface/FileDialogManager.cs
@@ -4,7 +4,6 @@ using System.Threading.Tasks;
 using Robust.Client.Graphics;
 using Robust.Shared.Console;
 using Robust.Shared.IoC;
-using Robust.Shared.Utility;
 
 namespace Robust.Client.UserInterface;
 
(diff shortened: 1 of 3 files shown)


Robust.Client/UserInterface/FileDialogManager.cs
```diff
@@ -4,33 +4,67 @@
...
+        FileDialogFilters? filters = null,
+        FileAccess access = FileAccess.ReadWrite,
+        FileShare? share = null)
+    {
+        var name = await Prompt(false, filters);
 
-        return File.Open(name, FileMode.Open, access, realShare);
+        if (name == null) return (null, null);
+
+        var path = new ResPath(name);
```
Ataman: This throws for paths like 
`C:\Users\someuser\Desktop\midis\somemidi.mid`

Ataman (commented):
Just tested this in my instrument PR. `FileDialogManager.GetFileAndName()` currently throws a `DebugAssertException` inside the `ResPath` constructor after selecting a MIDI file from my Desktop somewhere. I'm guessing `ResPath` should only be used inside the sandboxed environment? If this is correct, `FileDialogManager` should probably return strings instead in all places, not just in `FileDialogManager.GetFileAndName()`.

Robust.Client/UserInterface/FileDialogManager.cs
AJCM-git: the woes of testing in linux, ugh

PR #6758, opened 2026-07-08
Multi-window: Steam Overlay Memory Leak

AprilCatStreams:
Steam does not behave well when multiple double buffered windows exist from one parent process, this PR fixes the memory leak by disabling double buffering on all secondary windows.

Known Issues:
- If you `shift+tab` to open the overlay while having a pop-up focused the overlay will draw to the main window (at the wrong size) but leave the inputs (close interaction, buttons, etc) on the pop-up


metalgearsloth:
Gonna test this live and merge if it's all good.

PR #6764, opened 2026-07-08
Skip log formatting if level not enabled

metalgearsloth:
Shows up on test profile and I imagine a lot of logs are debug that we never need to create msgs for.


kontakt:
Nice!

deltanedas:
nooo my logging with side effects in the format string got broken :trollface:

PR #6766, opened 2026-07-08
Cache GetAllChildren

metalgearsloth:
1/3 of client test runner, I think mostly from serialization reflection, but for example entityconditions would do hundreds of lookups that we can just dictionary + array for much faster.


metalgearsloth:
I also think this is volatile enough we can't just 1-and-done a frozendictionary unless we somehow inferred most cached types that would hit it.

PR #6772, opened 2026-07-09
Color.cs Cleanup

sleepyyapril:
Color.TryFromHex was returning a bool you had to manually check. we're based now

Content side: https://github.com/space-wizards/space-station-14/pull/44581


Robust.Shared.Maths/Color.cs
```diff
@@ -933,69 +933,73 @@ public static Color InterpolateBetween(Color α, Color β, float λ)
             return new(Vector4.Lerp(α.RGBA, β.RGBA, λ));
         }
 
-        public static Color? TryFromHex(ReadOnlySpan<char> hexColor)
+        public static bool TryFromHex(ReadOnlySpan<char> hexColor, out Color color)
```
ArtisticRoomba, line 1053: previous public API should remain as an overload to prevent breaking changes

ArtisticRoomba, line 1053: API docs pwetty please

ArtisticRoomba (changes requested):

Robust.Shared.Maths/Color.cs
sleepyyapril: oh right i forgot about them

sleepyyapril: sending me down the rabbit hole...

AJCM-git (approved):
lgtm

ArtisticRoomba (approved):
booyeah! booyeah! booyeah!

PR #6778, opened 2026-07-10
Move EventBus benchmarks to the Engine.

Princess-Cheeseballs:
Title.
Moves the content EventBus benchmarks into Engine. These should be deleted in Content post this getting merged:tm:

Why? Why were they in content to begin with?

image

Here's my gamer stats so you can prove your CPU superiority.


Princess-Cheeseballs:
I ran the wrong benchmark to test this because I changed files and the key changed from 9 to 8 fuuuck

Princess-Cheeseballs:
Alright fixed it

PR #6784, opened 2026-07-12
EntitySystem.Spawn: defer MapInitEvent in SpawnNextTo/InContainer/...OrDrop

whatston3:
This PR changes the `EntityManager.SpawnX` functions so that the MapInitEvent is fired when the entity is on its target map, not nullspace.

## Why?

Will resolve this issue: https://github.com/space-wizards/space-station-14/issues/41502

Without this, these Spawn functions are more or less useless for spawning spawners (that tend to run on MapInit).

Gives a stronger, more consistent guarantee of when the MapInitEvent will be run between Spawn commands.

## Media

A demonstration of the issue in https://github.com/space-wizards/space-station-14/issues/41502 being resolved, which uses SpawnNextToOrDrop:

https://github.com/user-attachments/assets/21f24c12-558f-4ae0-801c-f37b36a24616

### DIFF
--- a/Robust.Shared/GameObjects/EntityManager.Spawn.cs
+++ b/Robust.Shared/GameObjects/EntityManager.Spawn.cs
@@ -7,7 +7,6 @@ using System.Runtime.CompilerServices;
 using Robust.Shared.Collections;
 using Robust.Shared.Containers;
 using Robust.Shared.Maths;
-using Robust.Shared.Serialization;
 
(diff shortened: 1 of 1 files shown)


SnappingOpossum:
Similar to #6192 in what it accomplishes.

whatston3:
> Similar to #6192 in what it accomplishes.

Wasn't noted in the linked issue when it should've been.

whatston3:
death

PR #6786, opened 2026-07-12
Check for deleted contacts

neomoth:
While making hand throwing predicted on Starlight, ran into this bit of code causing this exception to print in the client console:
```
[ERRO] runtime: Caught exception in "ResetPredictedEntities"
System.Collections.Generic.KeyNotFoundException: Entity 0 does not have a component of type Robust.Shared.GameObjects.TransformComponent
   at Robust.Client.Physics.PhysicsSystem.UpdateIsTouching(List`1 toUpdate) in /home/neomoth/src/space-station-14/RobustToolbox/Robust.Client/Physics/PhysicsSystem.Predict.cs:line 209
   at Robust.Client.GameStates.ClientGameStateManager.ResetPredictedEntities() in /home/neomoth/src/space-station-14/RobustToolbox/Robust.Client/GameStates/ClientGameStateManager.cs:line 695
   at Robust.Client.GameStates.ClientGameStateManager.ApplyGameState() in /home/neomoth/src/space-station-14/RobustToolbox/Robust.Client/GameStates/ClientGameStateManager.cs:line 357
```
and when checking with the debugger the contact point was flagged as deleted, so really there's no reason to be processing the contact point at all, especially if it's just going to flood the client console with exceptions.

Tested game after applying change, nothing seems to be affected other than avoiding the exception altogether, so as far as I can tell this shouldn't cause any issues.

### DIFF
--- a/Robust.Client/Physics/PhysicsSystem.Predict.cs
+++ b/Robust.Client/Physics/PhysicsSystem.Predict.cs
@@ -132,7 +132,10 @@ public sealed partial class PhysicsSystem
 
             // Don't evaluate deleted contacts.
             if ((contact.Flags & (ContactFlags.Deleting | ContactFlags.Deleted)) != 0x0)
+            {
+                contact.IsTouching = false;
                 continue;
+            }
 
             if (!bodyA.CanCollide || !bodyB.CanCollide)
             {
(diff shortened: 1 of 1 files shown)


metalgearsloth:
Do you know why the deleted contact was there?

neomoth:
> Do you know why the deleted contact was there?

Seemingly due to a predicted physics object embedding into a wall?
Check this PR for the changes I made that wound up causing this behavior to even occur: https://github.com/ss14Starlight/space-station-14/pull/5049

PR #6797, opened 2026-07-16
Fix serialization normalize performance

metalgearsloth:
Seems to be faster on my machine but will check github.


metalgearsloth:
Runners so noisy who even KNOWS but I got more coming.

PR #6804, opened 2026-07-16
sprite geometry clipping for water effect

Peptide90:
This is a PR necessary for my downstream fork. It allows the clipping of sprite geometry and in our case is an improvement over the FloorOccluder component to allow hiding parts of the players sprite without the need for rendering a sprite over the player.

Here is an example in action where we have a bobbing effect where the sprite cut is moving up and down.

https://github.com/user-attachments/assets/5530a999-2664-49a4-af65-885ee2a02ef7

### DIFF
--- a/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
+++ b/Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
@@ -215,6 +215,14 @@ namespace Robust.Client.GameObjects
         }
 
         /// <summary>
+        /// Fraction of every normal sprite layer cropped from its bottom edge before rendering.
+        /// This is applied before layer and post shaders, allowing effects such as water immersion
+        /// without consuming the sprite's single post-shader slot.
+        /// </summary>
+        [ViewVariables(VVAccess.ReadWrite)]
+        public float BottomClipFraction { get; set; }
+
+        /// <summary>
         ///     Whether to pass the screen texture to the <see cref="PostShader"/>.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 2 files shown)


metalgearsloth:
This looks like it'd be better off as a shader as the clipping is arbitrary?

Peptide90:
> This looks like it'd be better off as a shader as the clipping is arbitrary?

Hey MGS, there is already a water clipping shader but we only support one post shader currently, so water clipping as a post-shader competes with effects such as outlines, invisibility, status effects, tinting, or anything else using PostShader.

Geometry clipping directly crops each layer’s quad and texture region, giving consistent proportions compared to doing it in content. I think this was the simplest way I could do this unless we decide to support post shader chaining.

metalgearsloth:
> I think this was the simplest way I could do this unless we decide to support post shader chaining.

PostShader is better as this targeted at only 1 fork. You wouldn't see this kind of thing in a normal game engine like godot or unity which is what RT should be compared to.

Peptide90:
> > I think this was the simplest way I could do this unless we decide to support post shader chaining.
> 
> PostShader is better as this targeted at only 1 fork. You wouldn't see this kind of thing in a normal game engine like godot or unity which is what RT should be compared to.

So is post shader chaining what you want me to do or do I just use a regular post shader and forego the other shaders?

metalgearsloth:
Chaining would be better I would think.

deltanedas:
related to post shaders, there is 0 api for rendering a sprite with a post shader. its all done by clyde internally so you cant do anything in overlays etc

PR #6805, opened 2026-07-16
Fix sprite jittering

metalgearsloth:
This has annoyed me for FIVE YEARS since I coded arbitrary grid rotations.

So the issue is that we transform every single sprite into world terms and then back into the viewport. The problem with this is we get rounding issues as the game world is only 32x32 px so you will get slightly different rounded positions due to floating point maths!

To get around this we just clamp the grid positions to 32x32. We still keep floating-point maths for transforms + physics and only touch the render paths.


AftrLite:
LGTM 👍

PR #6811, opened 2026-07-17
Make rect deletion area align with camera rotation instead of world

B-Kirill:
Should be more convenient to use on grids.

Before:
before

After:
after


AJCM-git (approved):
lgtm

PR #6814, opened 2026-07-17
Add default location for open dialog

metalgearsloth:
Useful if we open character exports / playtime transfers. This is entirely for the prompt and never returns a file itself so should be sandbox safe. Content would just do like "/playtime-exports" for example.


AJCM-git (approved):
code looks good

PR #6816, opened 2026-07-17
Move release build to end of content test action

Tayrtahn:
Moves the attempt to build against content using the release configuration to the last step of the workflow, so that tests are still run to look for more serious errors when an unavoidable warning exists in content.


metalgearsloth:
Awesome thank you!

PR #6817, opened 2026-07-18
Add multiple post-shader support

metalgearsloth:
This should also fix the performance issue of cycling the render target constantly somewhat.

I also fixed post-shaders stacking their alpha so any transparent entities having it multiplied every time. The most obvious of this on content is ghost outlines messing with their transparency.

So the main issue internally is post-shaders stepping on each other's toes as they get rendered. To get around this we just use an intermediate that gets rendered into and swap between them.

Originally I had the 1 postrendertarget texture and drew a cutout of that but a bunch of content shaders assume they have exact bounds RN so changing that is more of a breaking change.

This should also stop the postshader churn where it recycles it every frame.


Robust.Client/Graphics/Clyde/Clyde.HLR.cs
```diff
@@ -334,93 +361,73 @@ private void DrawEntities(Viewport viewport, Box2Rotated worldBounds, Box2 world
...
+                            roundedPos = (Vector2i) entry.SpriteScreenBB.Center;
+                            var flippedPos = new Vector2i(roundedPos.X, screenSize.Y - roundedPos.Y);
+                            flippedPos -= entityPostRenderTarget.Size / 2;
+                            _renderHandle.Viewport(Box2i.FromDimensions(-flippedPos, screenSize));
+
+                            if (PostShadersRaiseEvent(postShaders))
+                            {
+                                _entityManager.EventBus.RaiseLocalEvent(entry.Uid,
+                                    new BeforePostShaderRenderEvent(entry.Sprite, viewport),
+                                    false);
```
Aviu00: Shouldn't this raise specific even for every post shader, passing their id, so that content can actually modify specific shader paramters based on it?
See #6719

Robust.Client/GameObjects/Components/Renderable/SpriteComponent.cs
```diff
@@ -1768,6 +1791,129 @@ public CopyToShaderParameters(CopyToShaderParameters toClone) : this(toClone.Lay
             }
         }
 
+        /// <summary>
+        ///     A post-shader currently applied to this sprite.
+        /// </summary>
+        public sealed class PostShaderEntry
```
Aviu00: Any chance to make it yamlable?

Robust.Client/Graphics/Clyde/Clyde.HLR.cs
Aviu00: Or content can just... Access shaders form sprite comp? I guess its fine then

metalgearsloth: Can do that.

deltanedas:
please make this an api method like RenderSprite is or entity views in UI etc have no way to do post shaders without literally copy pasting everything from engine

metalgearsloth:
> please make this an api method like RenderSprite is or entity views in UI etc have no way to do post shaders without literally copy pasting everything from engine

example!!!

deltanedas:
evil thing raidercode does for thermal vision "overlay" of shittiest hack known to man where it replaces the shader for every layer of a sprite (completely breaks dwarves btw) then calls RenderEntity and resets it
in an overlay...

deltanedas:
see also this ai slop that will get broken by literally any layer with a shader because UseShader gets overriden in RenderSprite

https://github.com/Goob-Station/Goob-Station/pull/6920/files#diff-c7fba322886990b65ca48e513c299789b4db1d57fb09d4248701ef7a92c2ebb7

metalgearsloth:
DrawEntity is super scrunkly but I have added postshaders as an optional arg.

Robust.Client/GameObjects/EntitySystems/SpriteSystem.Component.cs
```diff
@@ -108,4 +117,271 @@ public void QueueUpdateIsInert(Entity<SpriteComponent> sprite)
...
+        var afterArray = ToDependencyArray(args.After);
+
+        // IDs are stable handles for systems to replace/remove their own post-shader without disturbing others.
+        for (var i = 0; i < sprite.PostShaders.Count; i++)
+        {
+            var existing = sprite.PostShaders[i];
+            if (existing.Id != args.Id)
+                continue;
+
+            existing.Prototype = default!;
```
whatston3, line 232: Should this be `new(string.Empty);`, or should `TryIndex(ProtoId<T>)` work with null values inside the ProtoId?

deltanedas: currently it throws because TryIndex doesnt check for null ids which is silly
this happens for e.g. components where you dont specify an id in yml too

PR #6819, opened 2026-07-19
EntityTimers

aristophanivan:
## About the PR

Adds component-owned entity timers to RobustToolbox.

`EntityTimerSystem` provides a shared scheduler for delayed and repeating ECS behavior without requiring individual systems to run `Update()` loops over every entity with a timer-like component.

Multiple components on the same entity can own multiple independently identified timers.

## Why / Balance

Systems frequently use `EntitySystem.Update()` and `EntityQueryEnumerator` only to compare stored deadlines against `IGameTiming.CurTime`. This makes tick cost depend on the total number of matching entities even when few or no timers are ready.

The new system keeps timers in priority queues and processes only elapsed entries. It also centralizes timer cancellation, entity pause, prediction, and component lifecycle behavior.

This is an engine feature and has no direct effect on game balance.

## Technical details

- Added:
  - `EntityTimerSystem`
  - `ClientEntityTimerSystem`
  - `ServerEntityTimerSystem`
  - `EntityTimerId`
  - `EntityTimerFlags`
  - `EntityTimerInfo`
  - `EntityTimerEvent`
- A timer is identified by:
  - its owning `EntityUid`;
  - its owning component type;
  - its `EntityTimerId`.
- Components may own multiple timers.
- Different components on the same entity may use identical timer IDs.
- Scheduling the same timer key again atomically replaces the previous timer.
- Timers are stored in priority queues ordered by deadline and insertion sequence.
- A dictionary provides direct lookup, replacement, and cancellation.
- Generations allow stale priority queue entries to be removed lazily.
- Equal-deadline timers are dispatched deterministically.
- Timers scheduled from an event handler with an elapsed deadline cannot execute recursively during the same update.
- Repeating timers:
  - preserve their original phase;
  - report skipped intervals through `ElapsedCount`;
  - are rescheduled before dispatch so handlers can safely cancel or replace them.
- `EntityTimerInfo.Remaining` exposes the remaining duration for cooldown and UI consumers.
- One-shot timers are removed before event dispatch.
- Timer events are raised directly on the component that owns the timer.
- Removing a component cancels only timers owned by that component.
- Deleting or flushing an entity cancels all of its timers.
- Entity pause suspends timers by default and preserves their remaining duration.
- `IgnoreEntityPause` allows a timer to continue using simulation time while its entity is paused.
- `UpdatesOutsidePrediction` controls whether a client timer is processed outside predicted updates.
- Client state application temporarily prevents timer dispatch.
- The server always processes authoritative simulation timers.
- Timer dispatch occurs before ordinary `EntitySystem.Update()` calls.
- Timer state is runtime-only. Serializable or predicted deadlines remain owned by components and must be registered again after initialization or state application.

## Test plan

Built:

- `Robust.Shared`
- `Robust.Client`
- `Robust.Server`

Added and ran integration tests covering:

- multiple timers on one entity;
- multiple timers owned by one component;
- identical IDs owned by different components;
- timer replacement;
- cancellation and lookup;
- remaining-time calculation;
- deterministic equal-deadline ordering;
- reentrant scheduling;
- repeating timers;
- skipped interval reporting;
- cancellation from an event handler;
- entity pause and unpause;
- `IgnoreEntityPause`;
- component removal;
- entity deletion;
- entity flush;
- server authoritative dispatch;
- dispatch before ordinary entity-system updates;
- client prediction;
- updates outside prediction;
- past prediction;
- state application.

Results:

- Shared integration tests: all passed.
- Client entity-timer integration tests: all passed.

## Media

No media is required. This PR introduces an engine API and does not contain player-visible content.

## Requirements

- [X] I have tested this pull request and written instructions on how to test it
- [X] I have added media to this PR or it does not require an in-game showcase.

## Breaking changes

A new engine API is available through `EntityTimerSystem`.

Timer deadlines that must survive serialization, networking, or prediction rollback must remain stored in their owning components. Such timers must be registered again with `SetTimerAt()` after component initialization, loading, or state application.

The timer system does not serialize or roll back its internal priority queues.

No existing engine API was removed.

## Changelog

No changelog entry is required because this is an internal engine feature without player-visible behavior.


metalgearsloth:
While I personally don't care if an LLM is used this one clearly does not understand the codebase, and I don't think it's a productive use of my time feeding the review back into your LLM or for you to not be understanding the output.

PR #6828, opened 2026-07-20
Allow passing a reason to the shutdown command

Atakku:
Title.

Passed down the trimmed substringed argStr into `IBaseServer.Shutdown`, `BaseServer.Shutdown` already handles null/empty input, so effectively no change unless an argument is provided


Princess-Cheeseballs (approved):
Shrimple PR I suppose

PR #6830, opened 2026-07-21
SharedBroadphaseSystem.FindPairs: exit early if passed broadphase UID is invalid

whatston3:
Preventing a KeyNotFoundException from a component lookup on an invalid EntityUid.

Think this is sane - it probably shouldn't have an AABB to pass into QueryBroadphase if there's no valid map.

If you'd like something deeper, I'll try my best, but I'm still not very familiar with the engine.

<details>
    <summary>Call tracing:</summary>

BroadphaseContactJob.Execute tries to get an entity's map UID, defaulting to EntityUid.Invalid.
    
https://github.com/space-wizards/RobustToolbox/blob/e89668c21282f1b5017929ebc5c12b97e6453b93/Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs#L632-L638
    
It then eventually calls FindPairs with mapUid set to EntityUid.Invalid.
    
https://github.com/space-wizards/RobustToolbox/blob/e89668c21282f1b5017929ebc5c12b97e6453b93/Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs#L663
    
Inside of FindPairs, you run GetInvWorldMatrix on broadphase (EntityUid.Invalid)
    
https://github.com/space-wizards/RobustToolbox/blob/e89668c21282f1b5017929ebc5c12b97e6453b93/Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs#L421
    
Then GetInvWorldMatrix tries to get a TransformQuery on the passed entity.
    
https://github.com/space-wizards/RobustToolbox/blob/e89668c21282f1b5017929ebc5c12b97e6453b93/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs#L1351
    
And this results in a KeyNotFoundException in EntityQuery.GetComponent
    
https://github.com/space-wizards/RobustToolbox/blob/e89668c21282f1b5017929ebc5c12b97e6453b93/Robust.Shared/GameObjects/EntityManager.Components.cs#L1826-L1832

</details>

### DIFF
--- a/Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs
+++ b/Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs
@@ -396,7 +396,7 @@ namespace Robust.Shared.Physics.Systems
             DebugTools.Assert(proxy.Body.CanCollide);
 
             // Broadphase can't intersect with entities on itself so skip.
-            if (proxy.Entity == broadphase || !_xformQuery.TryGetComponent(proxy.Entity, out var xform))
+            if (proxy.Entity == broadphase || broadphase == EntityUid.Invalid || !_xformQuery.TryGetComponent(proxy.Entity, out var xform))
             {
                 return;
             }


metalgearsloth (changes requested):
It should definitely not be getting this far with an invalid broadphase and this path is hot so any check is significantly expensive.

metalgearsloth:
> BroadphaseContactJob.Execute tries to get an entity's map UID, defaulting to EntityUid.Invalid.

If they're in nullspace they shouldn't be on movebuffer so need to track down why that happens.

whatston3:
> If they're in nullspace they shouldn't be on movebuffer so need to track down why that happens.

Best I have at the moment is here: https://github.com/space-wizards/space-station-14/pull/44733#issuecomment-5039996965

edit:
The reproduction steps I have involve lights anchoring themselves or unanchoring themmselves and parenting themselves to a wall to respect the anchored state of the wall they're on.  The lights themselves shouldn't be in nullspace, nor should the walls, but this only seems to happen when the entities move in/out of PVS.  Will look into it more when I can, but any input would be greatly welcome.

metalgearsloth:
> > If they're in nullspace they shouldn't be on movebuffer so need to track down why that happens.
> 
> Best I have at the moment is here: [space-wizards/space-station-14#44733 (comment)](https://github.com/space-wizards/space-station-14/pull/44733#issuecomment-5039996965)
> 
> It's an issue with lights anchoring/unanchoring and parenting themselves on an anchor event. The lights themselves shouldn't be in nullspace, nor should the walls, but it does happen when moving in/out of PVS. Will look into it more when I can, but any input would be greatly welcome.

Anything touching transforms during anchoring is omegacursed and ideally engine would throw if someone tried.

Likely movebuffer stuff needs auditing but I can do it in the next couple weeks if required, I just need to stop delaying the illumination PR review.

metalgearsloth:
See linked PR!!!

whatston3:
Still seems to produce the issue on [ss14#44733](https://github.com/space-wizards/space-station-14/pull/44733)

I can get around this by changing SharedBroadPhaseSystem's Execute to start with the following:
```cs
            public void Execute(int index)
            {
                var proxy = MoveBuffer[index];
                var proxyXform = XformQuery.GetComponent(proxy.Entity);
                if (proxyXform.MapUid is not { } mapUid)
                    return;

                var broadphaseUid = proxyXform.Broadphase?.Uid;
                var worldAABB = TransformSys.GetWorldMatrix(broadphaseUid!.Value).TransformBox(proxy.AABB);

                var broadphaseExpand = System.GetBroadphaseExpand(proxy.Body, FrameTime);
```

The null coersion on MapUid just blows it up inside FindPairs on master:

```cs
var mapUid = XformQuery.GetComponent(proxy.Entity).MapUid ?? EntityUid.Invalid;
```

This isn't a real entity, it has no transform.

You're handling the Nullspace condition by assigning it to EntityUid.Invalid, which makes no sense as it won't ever have a TransformComponent.  Why not early return?

For a bit more context: the proxy's entity _is_ valid, but it seems to exist in nullspace.  This seems to be happening on PVS popout as well, possibly with light reparenting to an off-tile entity, for what it's worth.

PR #6831, opened 2026-07-21
Add polygon occluders

metalgearsloth:
Temporarily lost my mind trying to figure out wall lights but it was a content side bug. Needs 2nd pair of eyes to make sure lighting is the same.

Edge face code is similar we just need it to use tris not quads. Edge support is done via vector2 keys (face ordering on neighbor occluders will differ so need it to be order-agnostic). I think this is better than doing cardinal checks, or at least more flexible.

Will be more expensive, especially around handling embedded lights, but I don't think this part of light render is that badTM. If it is I can revisit it.

Future work is making the GPU winding be CCW so we don't have to reverse it. I just used CCW for the datafield so it's consistent with physics.

image

BC around occluder API.

Content PR is https://github.com/space-wizards/space-station-14/pull/44843


ArtisticRoomba:
Sauce alert

PR #6832, opened 2026-07-22
Text Outline

Pok27:
Added support for text outlines in `Font`, `DrawingHandleScreen`, `Label`, `RichTextLabel`, and overlays.
For https://github.com/space-wizards/space-station-14/pull/44848
image

### DIFF
--- a/Robust.Client/Graphics/Font.cs
+++ b/Robust.Client/Graphics/Font.cs
@@ -179,24 +179,27 @@ namespace Robust.Client.Graphics
             var glyphPosition = baseline + new Vector2(metrics.Value.BearingX, -metrics.Value.BearingY);
             if (outline is { Thickness: > 0 } settings)
             {
-                // Draw the outline by rendering the glyph around its original position.
-                var outlineOffset = settings.Thickness * scale;
-                foreach (var offset in OutlineOffsets)
+                // Sample concentric rings so the outline remains smooth at small sizes without gaps at larger sizes.
+                var outlineRadius = settings.Thickness * scale;
+                var ringCount = (int) MathF.Ceiling(outlineRadius);
+
+                for (var ring = 1; ring <= ringCount; ring++)
                 {
-                    DrawGlyph(handle, texture, glyphPosition + offset * outlineOffset, settings.Color);
+                    var radius = MathF.Min(ring, outlineRadius);
+                    var segmentCount = Math.Max(1, (int) MathF.Ceiling(MathHelper.TwoPi * radius));
+
+                    for (var segment = 0; segment < segmentCount; segment++)
+                    {
+                        var angle = segment / (float) segmentCount * MathHelper.TwoPi;
+                        var offset = new Vector2(MathF.Sin(angle), MathF.Cos(angle)) * radius;
+                        DrawGlyph(handle, texture, glyphPosition + offset, settings.Color);
+                    }
                 }
             }
 
             DrawGlyph(handle, texture, glyphPosition, color);
             return metrics.Value.Advance;
         }
-
-        private static readonly Vector2[] OutlineOffsets =
-        [
-            new(-1, -1), new(0, -1), new(1, -1), // Top
-            new(-1, 0), new(1, 0), // Sides
-            new(-1, 1), new(0, 1), new(1, 1), // Bottom
-        ];
 
         private static void DrawGlyph(DrawingHandleBase handle, Texture texture, Vector2 position, Color color)
         {
--- a/Robust.Client/Graphics/Font.cs
+++ b/Robust.Client/Graphics/Font.cs
@@ -177,18 +177,11 @@ namespace Robust.Client.Graphics
             }
 
             var glyphPosition = baseline + new Vector2(metrics.Value.BearingX, -metrics.Value.BearingY);
-            if (outline is { Thickness: > 0 } settings)
+            if (outline is { Thickness: > 0 } settings &&
+                Handle.GetOutlinedChar(rune, scale, settings.Thickness) is { } outlinedGlyph)
             {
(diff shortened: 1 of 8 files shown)


Robust.Client/Graphics/Font.cs
```diff
@@ -137,14 +176,33 @@ public override float DrawChar(DrawingHandleBase handle, Rune rune, Vector2 base
...
+                {
+                    for (var angle = 0f; angle < MathHelper.TwoPi; angle += 1f / radius)
+                    {
+                        var offset = new Vector2(MathF.Sin(angle), MathF.Cos(angle)) * radius;
+                        DrawGlyph(handle, texture, glyphPosition + offset, settings.Color);
+                    }
+                }
+            }
+
+            DrawGlyph(handle, texture, glyphPosition, color);
```
metalgearsloth, line 187: Isn't this going to make text rendering significantly slower? e.g. 4x calls makes the entire text render 4x slower. Wouldn't it be better to do it at a text level instead so at least the overhead is manageable?

sowelipililimute (changes requested):
this is not the correct way to render outlined text, this should be making use of freetype's glyph stroker https://freetype.org/freetype2/docs/reference/ft2-glyph_stroker.html#ft_glyph_strokeborder in order to make an outlined glyph that also gets cached like the filled glyphs do

metalgearsloth:
> this is not the correct way to render outlined text, this should be making use of freetype's glyph stroker [freetype.org/freetype2/docs/reference/ft2-glyph_stroker.html#ft_glyph_strokeborder](https://freetype.org/freetype2/docs/reference/ft2-glyph_stroker.html#ft_glyph_strokeborder) in order to make an outlined glyph that also gets cached like the filled glyphs do

incredibly suss name but sounds better knowing we have vector fonts!

Pok27:
I'll see what I can do

Robust.Client/Graphics/FontManager.cs
```diff
@@ -322,6 +348,24 @@ public void ClearSizeData()
...
 
+            public OutlinedGlyph? GetOutlinedChar(Rune codePoint, float scale, float thickness)
+            {
+                var glyph = GetGlyph(codePoint);
+                if (glyph == 0)
+                    return null;
+
+                var radius = (int) MathF.Round(thickness * scale * 64f);
+                if (radius <= 0)
+                    return null;
```
metalgearsloth, line 362: You should be able to check <= 0 above before doing MathF.Round

Robust.Client/Graphics/FontManager.cs
```diff
@@ -400,6 +444,7 @@ public ScaledFontData(int ascent, int descent, int height, int lineHeight)
 
             public readonly List<OwnedTexture> AtlasTextures = new();
             public readonly Dictionary<uint, GlyphInfo> GlyphInfos = new();
+            public readonly Dictionary<(uint Glyph, int Radius), OutlinedGlyphInfo> OutlinedGlyphInfos = new();
             public readonly int Ascent;
```
metalgearsloth, line 451: If this is constant please use frozendictionary if you can't use an array. Ideally this gets built once whenever the font is retrieved rather than on first render as well.

Robust.Client/Graphics/FontManager.cs
```diff
@@ -142,102 +142,128 @@ private GlyphInfo EnsureGlyphCached(FontInstanceHandle instance, ScaledFontData
...
-                            (0, 0));
+            var face = instance.FaceHandle.Face;
+            face.SetCharSize(0, instance.Size, 0, (uint) (_baseFontDpi * scale));
+            face.LoadGlyph(glyph, LoadFlags.Default, LoadTarget.Normal);
 
-                        break;
-                    }
+            using var sourceGlyph = face.Glyph.GetGlyph();
+            using var stroker = new Stroker(_library);
+            stroker.Set(radius, StrokerLineCap.Round, StrokerLineJoin.Round, Fixed16Dot16.FromInt32(0));
```
metalgearsloth, line 167: Is stroker supposed to be disposed like this or should it be kept on fontmanager?

metalgearsloth (changes requested):

Robust.Client/Graphics/FontManager.cs
Pok27: It is only needed while creating a single outlined glyph, so keeping it local with `using` gives it the appropriate lifetime

Robust.Client/Graphics/FontManager.cs
Pok27: glyphs and outlined glyphs are cached as they are requested, pre-building them when the font is retrieved would require rasterizing every glyph, for every scale and outline radius, including glyphs that may never be rendered

Robust.Client/Graphics/FontManager.cs
metalgearsloth: How expensive is it to setup though? E.g. HttpClient is disposable but ideally kept around for the application lifetime. I can try to do some digging on it.

Robust.Client/Graphics/FontManager.cs
metalgearsloth: If it shows up on a profiler I might come back to it then though dictionaries are too slow to use on hotpaths most of the time.

Robust.Client/Graphics/FontManager.cs
sowelipililimute: it's basically just buffer allocation, you need to call rewind if you use the same instance for more than one stroke https://freetype.org/freetype2/docs/reference/ft2-glyph_stroker.html#ft_stroker_rewind

metalgearsloth:
Sure!

mirrorcult:
Extreme beautiful lovely!!!!!!!!!!!

PR #6834, opened 2026-07-23
Add a field to exclude auto states from replays

DrSmugleaf:
My RMC replays are nearing 1 gigabyte each, please help me
Test attached
`[AutoGenerateComponentState(excludeReplays: true)]` and your replays will no longer store the entire bee movie script 300 times per tick

Planning to use it for the tactical map / area info that gets sent to players through their components

To my knowledge SessionSpecific and SendOnlyToOwner aren't useful since those explicitly check if it's not a replay state (null player)

You can also technically use this in a custom comp state handler if you are a terrorist and love making yourself and other people suffer by writing those out manually


Moredread:
I haven't looked too deeply into the code, but wouldn't that mean that if you add it to the tactical map in RMC-14, the map would be gone completely from the replay? I think it is worth keeping it, even if it can not be accessed via the current client, e.g. as a global copy, so one can patch it in later. Not recording it would mean it being gone forever.

I think an explicit warning in the docs would help so that people consider that this means complete exclusion from replays _permanently_, and it means, that it can't be recovered in the future. (I'm not familiar enough with the engine to give a precise diagnosis, my main worry is, being interested in replays being accessible in the more distant future, that state should not be dropped that players might not be able to access today, and devs accidentally dropping it because they think it not being worth it. The current design by accident includes state, that was not explicitly hooked up to the replays, and for preservation losing that side effect would be a bit sad.)

deltanedas:
who would win in a ram eating competition: 1T param llm vs trying to load 1 rmc replay

DrSmugleaf:
> I haven't looked too deeply into the code, but wouldn't that mean that if you add it to the tactical map in RMC-14, the map would be gone completely from the replay? I think it is worth keeping it, even if it can not be accessed via the current client, e.g. as a global copy, so one can patch it in later. Not recording it would mean it being gone forever.

You would need a separate thing for that that returns a separate state for the replay without saving everyone's tacmaps into it and while still using auto states, at the moment RMC replays store the tac map for every player instead of just once
Also the only way that you can see the tac map in an RMC replay right now is if the player you are spectating happens to open it or if you manage to open it yourself otherwise
Admins already have a history tracker of the last x tacmaps so for that specifically it's not needed by them

metalgearsloth:
> who would win in a ram eating competition: 1T param llm vs trying to load 1 rmc replay

checkpoints some day after recall's prs gets merged this century.

Robust.Server/GameStates/PvsSystem.GetStates.cs
```diff
@@ -68,12 +71,13 @@ private EntityState GetEntityState(ICommonSession? player, EntityUid entityUid,
...
 
-    private IComponentState? ComponentState(EntityUid uid, IComponent comp, ushort netId, ref ComponentGetState stateEv)
+    private IComponentState? ComponentState(EntityUid uid, IComponent comp, ushort netId, ref ComponentGetState stateEv, out bool excludeReplays)
     {
         DebugTools.Assert(comp.NetSyncEnabled, $"Attempting to get component state for an un-synced component: {comp.GetType()}");
         stateEv.State = null;
         _getStateHandlers![netId]?.Invoke(uid, comp, ref Unsafe.As<ComponentGetState, EntityEventBus.Unit>(ref stateEv));
         var state = stateEv.State;
+        excludeReplays = stateEv.ExcludeReplays;
         return state;
```
metalgearsloth:
```suggestion
// Reset the ComponentGetState data.
stateEv.State = null;
stateEv.ExcludeReplays = false;
_getStateHandlers![netId]?.Invoke(uid, comp, ref Unsafe.As<ComponentGetState, EntityEventBus.Unit>(ref stateEv));
var state = stateEv.State;
        excludeReplays = stateEv.ExcludeReplays;
return state;
```

PR #6838, opened 2026-07-23
Fix some movebuffer holes with nullspace

metalgearsloth:
Some cases of proxies getting removed but not getting purged from movebuffer, alternative to https://github.com/space-wizards/RobustToolbox/pull/6830


metalgearsloth:
Will try this but physics + transform needs some love whenever I get around to physics again in a couple months.

PR #6840, opened 2026-07-24
Hotfix: pass last state in SharedUserInterfaceSystem.EnsureClientBui

whatston3:
Fixes issues described in https://github.com/space-wizards/RobustToolbox/pull/6829#issuecomment-5063702223

I goofed.  This sends null state when the last available state _might not be null!_

Proof of the pudding on ss14 master + a few compilation fixes for v284.

https://github.com/user-attachments/assets/8d2f36e5-66df-444e-8fa6-5313b187f6a8

@deltanedas sorry about that - see video attached, test for yourself

### DIFF
--- a/Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
+++ b/Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
@@ -92,20 +92,20 @@ public abstract partial class SharedUserInterfaceSystem : EntitySystem
     }
 
     /// <summary>
-    /// Enqueues a BUI command to be processed in the next Update call.
+    /// Enqueues a BUI command to be processed in the next Update call,
+    /// getting the last available state for the given UserInterfaceComponent at the last state.
     /// </summary>
-    private void AddQueued(BoundUserInterface bui, QueuedUpdate type, BoundUserInterfaceState? state = null)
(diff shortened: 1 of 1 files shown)


Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
```diff
@@ -554,7 +554,8 @@ private void EnsureClientBui(Entity<UserInterfaceComponent> entity, Enum key, In
         if (!open)
             return;
 
-        AddQueued(boundUserInterface, QueuedUpdate.Open);
+        var state = entity.Comp.States.GetValueOrDefault(key);
+        AddQueued(boundUserInterface, QueuedUpdate.Open, state);
```
Princess-Cheeseballs: Might want an API like "AddQueuedState" that does the boilerplate automatically and use that in both places where we do this.

Princess-Cheeseballs (changes requested):

Robust.Shared/GameObjects/Systems/SharedUserInterfaceSystem.cs
whatston3: Changes should be applied, and I commented the two AddQueued... methods.

[Tested again, still functional.]

Princess-Cheeseballs:
I'll test and merge after I'm done doing staging stuff

Princess-Cheeseballs (approved):
Yummy! More tags released!

PR #6841, opened 2026-07-24
Expose cvar for changing doppler factor + lower default doppler factor

### LINKED ISSUE #6597: Feature request: Add CVAR to disable audio doppler effect
Would be nice to have

mirrorcult:
fixes #6597

lowers the default doppler factor because with our physics velocities it was pretty common to hear dopplering in situations where shit really should not be dopplering like just moving around as an aghost. this sounded pretty bad. it can still occur with very fast moving grids and things like that which is when it should be happening to begin with

basically see comment on the cvar

https://github.com/user-attachments/assets/cc77db63-1592-41e5-ab7f-70ec02b34a88

### DIFF
--- a/Robust.Shared/CVars.cs
+++ b/Robust.Shared/CVars.cs
@@ -1370,15 +1370,11 @@ namespace Robust.Shared
         /// and (if more than 1) exaggerates the effect. Setting this to 0 will disable the doppler effect.
         /// </summary>
         /// <remarks>
-        /// The default value for OpenAL is 1, but here we use 0.25, which was chosen sort of arbitrarily to better tune the effect
-        /// to velocity ranges actually common in RT projects. In SS14, with a value of 1, it was common to hear noticeable dopplering
-        /// even while just moving around as a ghost, which sounded pretty bad.
-        ///
         /// This is replicated rather than client-only because it makes more sense to control this on a server-level depending on
         /// what the game in question requires: servers with fast-moving grids may want to de-emphasize doppler even more than usual, etc.
         /// </remarks>
         public static readonly CVarDef<float> AudioDopplerFactor =
-            CVarDef.Create("audio.doppler_factor", 0.25f, CVar.REPLICATED | CVar.ARCHIVE);
+            CVarDef.Create("audio.doppler_factor", 1f, CVar.REPLICATED | CVar.ARCHIVE);
 
         /// <summary>
         /// Whether to enable HRTF (head-related transfer function) support for positional audio.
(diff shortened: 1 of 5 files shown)


Robust.Shared/CVars.cs
```diff
@@ -1373,6 +1365,21 @@ protected CVars()
...
+        /// <remarks>
+        /// The default value for OpenAL is 1, but here we use 0.25, which was chosen sort of arbitrarily to better tune the effect
+        /// to velocity ranges actually common in RT projects. In SS14, with a value of 1, it was common to hear noticeable dopplering
+        /// even while just moving around as a ghost, which sounded pretty bad.
+        ///
+        /// This is replicated rather than client-only because it makes more sense to control this on a server-level depending on
+        /// what the game in question requires: servers with fast-moving grids may want to de-emphasize doppler even more than usual, etc.
+        /// </remarks>
+        public static readonly CVarDef<float> AudioDopplerFactor =
+            CVarDef.Create("audio.doppler_factor", 0.25f, CVar.REPLICATED | CVar.ARCHIVE);
```
metalgearsloth: Preferable not to hardcode ss14 into engine please!

metalgearsloth (changes requested):

Robust.Shared/CVars.cs
mirrorcult: are you talking about the comment or the fact that the doppler factor is lowered? it being lowered would be relevant even for non-SS14 games on RT which would presumably be using relatively similar velocity numbers.

mirrorcult: also this isnt hardcoded--its a cvar dude

metalgearsloth: If you show me an engine with a default lower than 1 I'll merge it.

mirrorcult: https://github.com/enginmanap/limonEngine/blob/563401584fb56f3bbe116e8d31123550989d1e92/src/ALHelper.cpp#L23

what are we doing here man

metalgearsloth: I understand your frustration but 0.25 is as arbitrary out of the hat as you'd understand. The spec has 1.0 as default as do 3 major engines I checked, and where possible I think we should go with what the spec has as default as it's more predictable!

mirrorcult: 1.0 is as arbitrary a value as 0.25 for our purposes since openal doesnt require velocity to be in some universal unit. if you want it to be 1.0 no matter what thats fine by me but i figured people that aren't me and aren't aware that this is even a changeable value might enjoy a saner default

metalgearsloth:
Thanks mirror! I'll write the release note :]

PR #6854, opened 2026-07-25
Update to player toolshed command

neomoth:
As per the comment, Toolshed can just, inherently parse ICommonSession, but due to how the system works `player:imm` IS still needed since you can't just start a command with a session name. That being said, there's no real reason to parse a username string when you can just, pass it an ICommonSession directly and then return the exact thing you passed in, which then allows it to be piped into something else as per usual.
The main benefit here is being able to use autocompletion like basically any other instance of needing to pass in ICommonSession.

### DIFF
--- a/Robust.Server/Toolshed/Commands/Players/PlayersCommand.cs
+++ b/Robust.Server/Toolshed/Commands/Players/PlayersCommand.cs
@@ -34,26 +34,17 @@ public sealed partial class PlayerCommand : ToolshedCommand
         return ctx.Session!;
     }
 
+    /*
+     * Pass in ICommonSession to get the... passed... ICommonSession...
+     * Look that seems really stupid (because it is) but the Toolshed parser doesn't let you just type in a session
+     * name at the start of a command, so immediate IS required in order to pipe to other commands. Luckily, Toolshed
+     * DOES have a parser for ICommonSession that even comes with autocomplete. So as insane as it looks here to just
+     * have a function that returns the exact thing you're passing to it, it's more efficient (and convenient)
+     * than parsing a username string. Toolshed is fucking weird, man.
+     */
     [CommandImplementation("imm")]
-    public ICommonSession Immediate(IInvocationContext ctx, string username)
-    {
-        _playerManager.TryGetSessionByUsername(username, out var session);
-
-        if (session is null)
-        {
-            if (Guid.TryParse(username, out var guid))
-            {
-                _playerManager.TryGetSessionById(new NetUserId(guid), out session);
-            }
-        }
-
-        if (session is null)
-        {
-            ctx.ReportError(new NoSuchPlayerError(username));
-        }
-
-        return session!;
-    }
+    public ICommonSession Immediate(IInvocationContext ctx, ICommonSession session) =>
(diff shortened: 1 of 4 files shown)


AJCM-git (approved):

Robust.Server/Toolshed/Commands/Players/PlayersCommand.cs
```diff
@@ -34,26 +34,17 @@ public ICommonSession Self(IInvocationContext ctx)
...
-    public ICommonSession Immediate(IInvocationContext ctx, string username)
-    {
-        _playerManager.TryGetSessionByUsername(username, out var session);
-
-        if (session is null)
-        {
-            if (Guid.TryParse(username, out var guid))
-            {
-                _playerManager.TryGetSessionById(new NetUserId(guid), out session);
-            }
```
ElectroJr, line 47: The `SessionTypeParser` doesn't currently support guids. Just in case someone expects this command to work with guids, you should add support for that to avoid breaking changes.

Robust.Server/Toolshed/Commands/Players/PlayersCommand.cs
```diff
@@ -34,26 +34,17 @@ public ICommonSession Self(IInvocationContext ctx)
         return ctx.Session!;
     }
 
+    /*
+     * Pass in ICommonSession to get the... passed... ICommonSession...
+     * Look that seems really stupid (because it is) but the Toolshed parser doesn't let you just type in a session
+     * name at the start of a command, so immediate IS required in order to pipe to other commands. Luckily, Toolshed
```
ElectroJr, line 40: I've thought about making it so that commands that need a piped input automatically also work without a pipe and instead treat it like a normal positional argument, which would get rid of that weirdness, but I don't know if there is some reason it wasn't done like that in the first place.

ElectroJr (changes requested):

Robust.Server/Toolshed/Commands/Players/PlayersCommand.cs
neomoth: Actually yeah I suppose looking at it I managed to overlook that part of the command entirely. I will add that to the parser but I will not provide autocomplete for it as to avoid leaking GUIDs since I don't think theres any ways of doing that normally without having access to the server files and database (i haven't checked but i know on Starlight at least we need to ask Darkrell for any user's guid as we cant seem to access that information ourselves)

Robust.Server/Toolshed/Commands/Players/PlayersCommand.cs
neomoth: That is a damn good question honestly though that feels a little out of scope for what I'm trying to do with this PR

Robust.Server/Toolshed/Commands/Players/PlayersCommand.cs
neomoth: Actually turns out we can just query https://auth.spacestation14.com/api/query/name?name=UsernameHere to get GUID so autocomplete it is

ElectroJr (approved):

PR #6857, opened 2026-07-26
Support ordered multi-registrar component events

Inconnu1337:
Directed subscriptions were capped at one handler per component & event pair
engine-wide, so a system could never subscribe to a pair another system had
already claimed. Now several can, as long as each declares an order against the
others with before/after the first to subscribe may omit it and runs first.

A single system subscribing to the same pair twice still throws, as does stacking
on [ComponentEvent]; neither can express an order.

Also adds UnsubscribeLocalEvent<TComp, TEvent>(Type owner) so a system shutting
down only drops its own handler instead of everyone's

### DIFF
--- a/Robust.Shared.IntegrationTests/GameObjects/EntityEventBusTests.MultiRegistrar.cs
+++ b/Robust.Shared.IntegrationTests/GameObjects/EntityEventBusTests.MultiRegistrar.cs
@@ -1,0 +1,320 @@
+using System;
+using System.Collections.Generic;
+using NUnit.Framework;
+using Robust.Shared.GameObjects;
+using Robust.Shared.IoC;
+using Robust.Shared.Map;
+using Robust.Shared.Reflection;
+using Robust.UnitTesting.Server;
+
+namespace Robust.UnitTesting.Shared.GameObjects
+{
+    internal sealed partial class EntityEventBusTests
+    {
+        private static (EntityEventBus Bus, EntityUid Uid, Action Attach) MultiRegistrarFactory()
+        {
+            var sim = RobustServerSimulation
+                .NewSimulation()
+                .RegisterComponents(f => f.RegisterClass<OrderAComponent>())
+                .InitializeInstance();
+
+            var entMan = sim.Resolve<EntityManager>();
+            var uid = entMan.Spawn();
+            var comp = entMan.AddComponent<OrderAComponent>(uid);
+            var bus = entMan.EventBusInternal;
+            bus.ClearSubscriptions();
+
+            var reg = sim.Resolve<IComponentFactory>().GetRegistration(CompIdx.Index<OrderAComponent>());
+
+            void Attach()
+            {
+                bus.LockSubscriptions();
+                bus.OnEntityAdded(uid);
+                bus.OnComponentAdded(new AddedComponentEventArgs(new ComponentEventArgs(comp, uid), reg));
+            }
+
+            return (bus, uid, Attach);
+        }
+
+        [Test]
+        public void MultipleRegistrarsSubscribeToSameCompEvent()
+        {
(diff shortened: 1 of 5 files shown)


metalgearsloth (changes requested):
Needs benchmarks to justify even having this as it's rare that systems even need this.

Inconnu1337:
> Needs benchmarks to justify even having this as it's rare that systems even need this.

I took the chain walk back out of the hot path.

Stacked subscriptions have to declare `before`/`after` against each other, which marks the
event ordered, so chains longer than one are only ever dispatched through
`EntCollectOrdered`. `EntDispatch` keeps its single call, with a
`[Conditional("DEBUG")]` assert as a tripwire in case that invariant is ever relaxed
extra handlers would otherwise silently stop running.

`RaiseComponentEvent` and `GetNetCompEventHandlers` were never in scope: `_compEventSubs`
still maps straight to `DirectedEventHandler`, and registrar tracking is dropped in
`LockSubscriptions`.

`RaiseEventBenchmark`, .NET 10.0.9, X64 RyuJIT x86-64-v3, idle machine:

| Method           | master     | changes     | Δ     |
|------------------|-----------:|-----------:|------:|
| RaiseEvent1      |  77.20 ns  |  73.85 ns  | -4.3% |
| RaiseEvent2      | 161.95 ns  | 168.03 ns  | +3.8% |
| RaiseEvent4      | 243.60 ns  | 242.28 ns  | -0.5% |
| RaiseEvent8      | 399.06 ns  | 390.38 ns  | -2.2% |
| RaiseEvent16     | 721.65 ns  | 764.55 ns  | +5.9% |
| RaiseCompEvent   |  20.28 ns  |  20.21 ns  | -0.4% |
| RaiseICompEvent  |  41.34 ns  |  37.38 ns  | -9.6% |
| RaiseNetEvent    |   6.56 ns  |   6.57 ns  | +0.2% |
| RaiseCSharpEvent |   4.50 ns  |   4.93 ns  | +9.6% |

The control benchmarks suggest this harness can't resolve the question.
`RaiseCSharpEvent` is a plain delegate invoke that never touches the bus, and it moves
9.6% between the two runs. `RaiseICompEvent`, also on an untouched path, moves -9.6% the
other way. Over four runs:

| | changes #1 | master #1 | changes #2 | master #2 | spread |
|---|---:|---:|---:|---:|---:|
| RaiseCSharpEvent | 4.977 | 4.930 | 4.929 | 4.498 | 10.6% |
| RaiseEvent1      | 83.15 | 76.46 | 73.85 | 77.20 | 12.6% |

That code is identical in both branches. BDN's error bars cover variance within a
process; these are two separately built binaries, so JIT output and code layout differ.
Between-build noise sits an order of magnitude above anything this diff could do.

Costs worth naming:

- `DirectedRegistration` grows by two references. One object per subscription, at startup.
- `IsOrdered` lives on `EventData`, so stacking marks the whole event type ordered rather
  than the pair a system stacking on a hot event moves it onto `RaiseLocalOrdered`
  everywhere. Pre-existing for any `before`/`after` use; making `IsOrdered` per-pair would
  fix it, separate change

metalgearsloth:
> > Needs benchmarks to justify even having this as it's rare that systems even need this.
> 
> I took the chain walk back out of the hot path.
> 
> Stacked subscriptions have to declare `before`/`after` against each other, which marks the event ordered, so chains longer than one are only ever dispatched through `EntCollectOrdered`. `EntDispatch` keeps its single call, with a `[Conditional("DEBUG")]` assert as a tripwire in case that invariant is ever relaxed extra handlers would otherwise silently stop running.

I'm skeptical the niche usecases of this are worth materially making the eventbus slower, especially as this will take up much more of eventbus performance when the trycomp is removed, and please add benchmarks to verify.

PR #6862, opened 2026-07-28
Fix count usages

kontakt:
These are popping as errors in updated nunit as they're comparing the function signature to the int. 
Throws "These can never be equal" errors.


kontakt:
*shakes fist at XAML*

metalgearsloth (approved):

PR #6864, opened 2026-07-28
[Toolshed] Array parsing

neomoth:
Adds the ability to parse arrays through `List<T>`.
Features mostly intuitive autocompletion, errors for when too few or too many parameters, and the autocompletion will also only provide `,` or `]` depending on if you're within the length constraints (if any).
minimum and maximum can be set by prefixing the parameter with the `[ListLength]` attribute, specifying `MinLength` and `MaxLength` to set their respective values.
Example:
```cs
[CommandImplementation]
public void CommandMethod(IInvocationContext ctx, [ListLength(MinValue=5)] List<string> stringArray){ ... }
```
`IInvocationContext` parameter can be placed anywhere in the parameter list, and this accounts for that.
`ParserContext` now exposes a few extra things to make this possible, such as the current `CommandArgument` and the `MethodInfo` for the command method, in order to be able to check attributes.
Also adds `valarr` as a command, more or less a duplicate of `val` but for constructing a `ValueArray` of a given type for storing as a variable or something.
As of the latest commit in this PR, ALSO now makes optional values suck less. Instead of needing `[Optional] [DefaultValue(value)]`, which didn't support defaults for custom types, now you can use `[Optional] OptionalValue<T>`.


ElectroJr:
I only had a brief look so far but have two questions:

> Adds the ability to parse arrays through ValueArray<T>.

Why introduce a new type instead of just adding support for T[] or List<T>?

> As of the latest commit in this PR, ALSO now makes optional values suck less. Instead of needing [Optional] [DefaultValue(value)], which didn't support defaults for custom types, now you can use [Optional] OptionalValue<T>.

I'm not sure I understand, can you give an example of a case where the current optional arguments fail?

neomoth:
> I only had a brief look so far but have two questions:
> 
> > Adds the ability to parse arrays through ValueArray<T>.
> 
> Why introduce a new type instead of just adding support for T[] or List<T>?
> 
> > As of the latest commit in this PR, ALSO now makes optional values suck less. Instead of needing [Optional] [DefaultValue(value)], which didn't support defaults for custom types, now you can use [Optional] OptionalValue<T>.
> 
> I'm not sure I understand, can you give an example of a case where the current optional arguments fail?

As for why not to use `T[]`, I tried that but it seems toolshed will not recognize it. It wants very strict typing, at least for the base type, as in it doesnt seem to care about generics, but if you put `string[]` as an argument it will explicitly look for a `string[]` parser, while doing something like `ArrayValue<string>` seems to circumvent this, as it looks for `ArrayValue<>` at which point I can parse the generic type passed to it, e.g. `ArrayValue<string>` ~~I remember there being SOME kind of reason I did not opt for `List<>` despite this though, but I might go back and check if I can get it to work with that.~~
Edit: Updated PR to use `List<T>` as that seems to work just fine. Yay!

Current optional values DO work, but only for primitives. If you write a custom parser for a specific thing that you want auto complete for, lets say a parser that shows all active vote IDs as an example, if I wanted to for whatever reason make this an optional value with some sort of default, there is no way to do that with `[DefaultParameterValue]` as far as I could tell, as I could not get it to accept anything but a constant value for a primitive, such as a string or a number. I could *not* get it to for example, accept a default vote ID, or to pass null or an empty array in place of `ValueArray<>`. You also still can't pass null as there's no nullable parser, so my solution here was to have a parameter telling if it was defined by the player in the command or if it was omitted or skipped. Thats the other main benefit, being able to skip a parameter. Before there was no way to skip one, and now there is, via passing `_`, so you can have an optional parameter not necessarily be at the end of the command and it would still work.

I made a vote command as a test for both of these features since that was the whole reason I wanted to implement them, to avoid redundant commands depending on the amount of options or which parameters you wanted to include, since I had noticed that there were some parameters you could not set with `customvote` such as the duration or whether to show votes, as well as there being no way to specify who to show the vote to (e.g. what if I wanted to do a ghost-only vote? or give the members of command but nobody else a vote? (admins would always be shown the vote regardless))
Below is the method signature for this command as well as the syntax you could use in-game with the two features combined to show how it cuts down on needing to do something like `vote:customwithduration` or whatever:
```cs
[CommandImplementation("custom")]
public void CreateCustomVote(IInvocationContext ctx, string title, [ListLength(MinLength = 2, MaxLength = 9)] List<string> options, [Optional] OptionalValue<float> duration, bool showVotes)
```
Default duration and showvotes: `> vote:custom "Title text" ["option one", "option two"]`
Default showvotes: `> vote:custom "Title text" ["option one", "option two"] 20`
Default duration but specifying showvotes: `> vote:custom "Title text" ["option one", "option two"] _ true`

As for the whole thing with `[DefaultParameterValue]`, as an example of what would *not* work, lets say I want a parameterless `vote:cancel` to cancel all active votes instead of needing a `vote:cancel` and `vote:cancelall` command for this. As I stated before, I have a custom parser for finding and displaying all active vote IDs as autocomplete options (internally it just parses as an integer its mainly the autocomplete that is custom here but because toolshed is dumb I need to make a whole `VoteId` record and an entire parser for it just for this), if I want the value to be optional, I cannot use `[DefaultParameterValue]` since it won't accept anything that would produce a default `VoteId`. You may think I can just, not include the attribute, but not doing so like the following:
```cs
[CommandImplementation("cancel")]
public void CancelVote(IInvocationContext ctx, [Optional] VoteId id)
```
would just make it return as `Missing` when parsing it, which causes the entire command to fail to parse.

ElectroJr:
> I could not get it to accept anything but a constant value for a primitive, such as a string or a number. I could not get it to for example, accept a default vote ID, or to pass null or an empty array in place of ValueArray<>.

That is a C# limitation, not a toolshed limitation, and I don't see how your implementation gets around that. The way you'd usually handle this is to just use null as the default and assign the desired default values if null gets passed. I.e., something like this works fine once you implement array parsing
```cs
[ToolshedCommand]
public sealed class FooCommand : ToolshedCommand
{
    public record struct CustomStruct(string Data);

    private static readonly string[] DefaultOptions = ["default"];
    private static readonly CustomStruct DefaultStruct  = new() {Data = "default"};

    [CommandImplementation("foo")]
    public void Foo(
        string title = "default title",
        string[]? options = null,
        float duration = 123,
        bool showVotes = false,
        CustomStruct? data= default)
    {
        options ??= DefaultOptions;
        data ??= DefaultStruct;
		// Run the actual command
    }
}
```

I assume you were planning to do something similar, but checking `OptionalValue<T>.Assigned` instead of checking for null, but I don't see why that'd be needed when you could just do null checks and have access to all the nice features like `??` or `??=`


>  Before there was no way to skip one, and now there is, via passing _

Yes it does definitely add that which wasn't there before. But if that does get added, then I feel like that should be standardized so that all optional arguments are skippable, because otherwise you start getting inconsistencies, where some commands have optional args that can be skipped while others can't. So for the sake of consistency I'd rather just implement that for all arguments. Though I think a better idea would be to just add tab completion for optional arguments so that you can "skip" the argument by just hitting tab and having it auto-fill the default value, which would also get around the issue of what to do if you actually want to pass a literal "_" to an optional argument.

Either way, its probably better to separate that out into a different PR to get array/list parsing merged faster.

>  its mainly the autocomplete that is custom here but because toolshed is dumb I need to make a whole VoteId record and an entire parser for it just for this)

You don't need to create a new struct for this. If you want custom type parsing for an existing type like int, you can just create a class that inherits from `CustomTypeParser<int>`, or in your case `CustomCompletionParser<int>` (which uses the default parser for the type and just exists to modify completion suggestions). You then specify that the command's argument should use via the `CommandArgument` attribute. Its still some effort, but I don't think there's really any way to make custom completion suggestions easier.

For example, the `createnode` command has a prototype id argument: `[CommandArgument(typeof(XenoEffectParser))] ProtoId<EntityPrototype> effect`. Instead of using the default protoId parser which would suggest any entity prototype, it uses the `XenoEffectParser : CustomCompletionParser<ProtoId<EntityPrototype>>` which is basically just the normal protoid parser but with customized auto completion suggestions.

I know toolshed docs are pretty lacking and the custom completion parsers aren't even mentioned, but at least general custom parsers are mentioned briefly in https://docs.spacestation14.com/en/robust-toolbox/toolshed/development.html

neomoth:
> > I could not get it to accept anything but a constant value for a primitive, such as a string or a number. I could not get it to for example, accept a default vote ID, or to pass null or an empty array in place of ValueArray<>.
> 
> That is a C# limitation, not a toolshed limitation, and I don't see how your implementation gets around that. The way you'd usually handle this is to just use null as the default and assign the desired default values if null gets passed. I.e., something like this works fine once you implement array parsing
> ```cs
> [ToolshedCommand]
> public sealed class FooCommand : ToolshedCommand
> {
>     public record struct CustomStruct(string Data);
> 
>     private static readonly string[] DefaultOptions = ["default"];
>     private static readonly CustomStruct DefaultStruct  = new() {Data = "default"};
> 
>     [CommandImplementation("foo")]
>     public void Foo(
>         string title = "default title",
>         string[]? options = null,
>         float duration = 123,
>         bool showVotes = false,
>         CustomStruct? data= default)
>     {
>         options ??= DefaultOptions;
>         data ??= DefaultStruct;
> 		// Run the actual command
>     }
> }
> ```
> 
> I assume you were planning to do something similar, but checking `OptionalValue<T>.Assigned` instead of checking for null, but I don't see why that'd be needed when you could just do null checks and have access to all the nice features like `??` or `??=`
> 
> 
> >  Before there was no way to skip one, and now there is, via passing _
> 
> Yes it does definitely add that which wasn't there before. But if that does get added, then I feel like that should be standardized so that all optional arguments are skippable, because otherwise you start getting inconsistencies, where some commands have optional args that can be skipped while others can't. So for the sake of consistency I'd rather just implement that for all arguments. Though I think a better idea would be to just add tab completion for optional arguments so that you can "skip" the argument by just hitting tab and having it auto-fill the default value, which would also get around the issue of what to do if you actually want to pass a literal "_" to an optional argument.
> 
> Either way, its probably better to separate that out into a different PR to get array/list parsing merged faster.
> 
> >  its mainly the autocomplete that is custom here but because toolshed is dumb I need to make a whole VoteId record and an entire parser for it just for this)
> 
> You don't need to create a new struct for this. If you want custom type parsing for an existing type like int, you can just create a class that inherits from `CustomTypeParser<int>`, or in your case `CustomCompletionParser<int>` (which uses the default parser for the type and just exists to modify completion suggestions). You then specify that the command's argument should use via the `CommandArgument` attribute. Its still some effort, but I don't think there's really any way to make custom completion suggestions easier.
> 
> For example, the `createnode` command has a prototype id argument: `[CommandArgument(typeof(XenoEffectParser))] ProtoId<EntityPrototype> effect`. Instead of using the default protoId parser which would suggest any entity prototype, it uses the `XenoEffectParser : CustomCompletionParser<ProtoId<EntityPrototype>>` which is basically just the normal protoid parser but with customized auto completion suggestions.
> 
> I know toolshed docs are pretty lacking and the custom completion parsers aren't even mentioned, but at least general custom parsers are mentioned briefly in https://docs.spacestation14.com/en/robust-toolbox/toolshed/development.html

Hm. I'll look into that, and yeah I suppose I can pull optional stuff out into its own PR. Maybe I could still get away with _ to skip stuff in place of typing null though, in said PR. Either way I'll make another push later today separating the two things and probably work on the optional stuff some more.

ElectroJr:
I've opened a PR to your branch (https://github.com/neomoth/RobustToolbox/pull/1) that adds array parsing, some unit tests, and a couple of other things. E.g., the current way that the min/max length attributes are obtained isn't correct if the method has non-command-argument arguments.

neomoth:
@ElectroJr approved and merged on my branch, everything looks good methinks.

ElectroJr (approved):

PR #6865, opened 2026-07-28
Expose relative mouse mode on windows

rhailrake:
Exposes relative mouse mode through `IClydeWindow`.


metalgearsloth (approved):
Just need to see tests passing either when freebsd fixed or screenshot or w/e

PR #6867, opened 2026-07-28
Generic Event Subscriptions auto-generator

Roudenn:
## About this PR
Added `[SubscribeRelayEvent]` attribute for subscription methods in `EntitySystem`s. This attribute is applied to generic event subscriptions that follow a specific format, and this 

## Why
Right now to make a relay for some events you have to do it completely manually. Also there is no generic way to do it with the normal `[SubscribeLocalEvent]` attribute, so you always have to override the `Initialize()` method.
And finally, in case of very general systems specifying relays for each new event quickly becomes... Hard to support.

[Example fix on SS14's InventorySystem](https://github.com/space-wizards/space-station-14/pull/45015)

This attribute allows to significantly reduce tons of already existing boilerplate code by auto-generating this stuff in a partial class with other attribute generated event subscriptions. This hides the super-evil looking code, and saves a lot of time and nerves.

## Technical concept
When generating a method wtih any entity system subscription attribute, it gets the type parameter requirements and tries filter them on all `[GenericEvent]` types. For each remaining type that passes all checks, a Local event subscription gets created. For each selected type if it also has type parameters, it runs the same process an picks the available types with `[GenericEvent]` attribute in all possible combinations.

## Technical details
Entity subscription generator first gets all types with `[GenericEvent]` attribute (for optimization reasons, so we don't need to filter the inheritance of ALL types). When parsing the subscription methods it stores the info amout type parameter constraints if there are any. When generating a subscription with type parameters, it goes through all `[GenericEvent]` types to fill the type parameters using the Cartesian product while still respecting type constraints. Each end result is then replaced in the final string either fully if the type parameter is just standing here alone, or with a Regex expression if it's inside of the brackets (that way if there are many type parameters, they will still be replaced).

### Naming
`[GenericEvent]` doesn't actually fully describe what it does in its name. It makes the type available to be put into an event with type parameters when it meets some constraints, but it can be any type, not just an event that can be raised on the event bus.

### Tests
I've made a test for this feature, but I've failed to properly setup a Generator test because it apparently requires some unknown super-duper shenanigans. If you are about to ask me for tests, first help me to set them up because this is complicated for some reason.

## TODO list
- [x] Add support for struct events
- [x] Allow nested generic event arguments (with arbitrary depth) and respect other special constraints
- [ ] Add support for Network and All subscriptions (?) Seems pretty useless so I'll add this only if requested
- [x] Test this on some real SS14 code

## Disclaimer
I don't actually know how code generators work!! I've spent an entire day on researching on just how to make this feature, so I'm sorry if you see some horrible code right here, this is my first time on Roslyn lands (I honestly don't like it here but my the need for this feature was stronger).
Though the code itself works perfectly fine and it seems pretty fast.


Princess-Cheeseballs:
@Jezithyr this conflicts with some of your eventbus plans, but I feel like the attribute is probably fine since it could be repurposed later on.

Roudenn:
Changing the name to Generic Event subscriptions and adding support for all subscription types because this PR basically just adds support for type parameters in event subscriptions (by automatically generating the code)

Roudenn:
Should be working, but before merge a unit test is likely required to test for some weird cases (like multiple type parameters on the same method)

Roudenn:
Would appreciate if someone would help me to fix the generator test for templates, otherwise the PR is ready

Roudenn:
Draft until I figure out how to make `[GenericEvent]` check inheritance instead of being applied to literally everything

Robust.Shared.EntitySystemSubscriptionsGenerator/KnownTypes.cs
```diff
@@ -54,5 +56,5 @@ public enum SubscriptionType
 {
     All,
     Network,
-    Local,
+    Local
```
Centronias, line 59: Trailing commas are good because when a new enum variant is added, it doesn't require modifying the previous line, why remove it?

PR #6869, opened 2026-07-29
Make disabled checkboxes greyed out

SlamBamActionman:
Checkboxes currently do not have any different visuals for when they're disabled or not, so this PR adds it:

Enabled vs disabled:
image


SlamBamActionman:
Test fail is UI but appears unrelated (occurs on other PRs)

PR #6870, opened 2026-07-29
Fix sRGB framebuffer in Windows compat mode

metalgearsloth:


Robust.Client/Graphics/Clyde/Windowing/Sdl3.Window.cs
kontakt, line 179: This is just disabling ES on non-windows platforms?

metalgearsloth: opengles support for srgb is inconsistent, and incorrect handling of this causes the dark screen bug, but I am open to other solutions.

kontakt: But this will set `GLAttr.SDL_GL_FRAMEBUFFER_SRGB_CAPABLE` to '1' on linux, even in compatibility mode. This line shouldn't affect a windows usage in any case.

metalgearsloth: I'll take whatever fixes it but srgb handling has been broken since forever so settling for whatever bandaids work.

kontakt: Right, but I don't think making it so that Linux is always marked as SRGB capable, even in compat, will help anything

metalgearsloth: Angle is windows-only which is likely the cause and has its own srgb handling.

PR #6871, opened 2026-07-29
try microoptimise physics island solving

deltanedas:
- reuse list for islands and the island joints
  island joints is only used within a loop so its fine, Step is never ran in parallel so the islands list is fine
- fix Cleanup, it was clearing _islandSet twice and didnt clear _bodyStack
- use a span instead of ToArray when actually solving the islands

the joint/error lists in islanddata could be pooled in the future but im not sure

high client cpu usage from prediction ticks mostly from Solve Islands prof section made me look at this

havent properly profiled it but it might help

@metalgearsloth if you have the time can you bench it and see if it actually improved anything (lower cpu use or less pressure on the GC)


metalgearsloth:
I ain't benchmarking micro-opts for physics just merge master to see if it works.

deltanedas:
@metalgearsloth master merged

PR #6872, opened 2026-07-30
Fix grid fragmentation during drag-fill placement

XxRaay:
Fix grid fragmentation during drag-fill placement:

PlaceNewTile in PlacementManager.cs now checks _lastPlacement before falling back to TryFindGridAt.
Added a 2-second timeout on _lastPlacement, so tiles placed during one drag-fill merge into the same grid, but won't merge with an old grid if placed later.
Use WorldToTile for tile position calc so coordinate conversion is correct regardless of the entity the original coords were relative to.


metalgearsloth:
Kinda janky, much better if we just add the tilerefs as a list in the message sorry, I might PR a placementsystem soon!!!

PR #6880, opened 2026-07-31
Fix RobustIntegrationTest not raising a OnConnecting event for dummy sessions

DrSmugleaf:
Guess who had to debug tests for hours with them randomly locking up for fun half-way through
This causes any code ever expecting a player row to exist for example to just die
The data used is the same RobustIntegrationTest uses elsewhere (loopback, port 1212, empty hwid, guest assigned)


metalgearsloth:
Good stuff!!!

PR #6882, opened 2026-07-31
Strip nullable annotations from component types when generating event subscriptions

### LINKED ISSUE #6873: Event subscription code generator generates bad code for nullable component parameters
This is oddly specific, but it's an issue I ran into.

Using an event subscription attribute on a method with a nullable component type as a parameter causes the generator to produce code that raises a compiler error.

This can be demonstrated by adding `[SubscribeLocalEvent]` to `SharedCryoPod.OnEmagged`, which has this signature:
```cs
protected void OnEmagged(EntityUid uid, CryoPodComponent? cryoPodComponent, ref GotEmaggedEvent args)
```
This generates the following method invocation:
```cs
SubscribeLocalEvent<Content.Shared.Medical.Cryogenics.CryoPodComponent?, Content.Shared.Emag.Systems.GotEmaggedEvent>(OnEmagged);
```
which raises this error:
```
The annotation for nullable reference types should only be used in code within a '#nullable' annotations context. Auto-generated code requires an explicit '#nullable' directive in source.
```

Now, there's really no reason for `CryoPodComponent` to be nullable in that event handler, so this case is easy enough to fix. But it would probably be good if either the generator could handle this (probably by just removing the nullable annotation? Or adding `#nullable enable` in the generator?) or the analyzer could flag this as a problem.

Tayrtahn:
The EntitySystem event subscription generator now removes nullable annotations (`?`) from component types when generating `ComponentEventHandler`-based subscriptions.

So:
```cs
protected void OnEmagged(EntityUid uid, CryoPodComponent? cryoPodComponent, ref GotEmaggedEvent args)
```
now generates:
```cs
SubscribeLocalEvent<Content.Shared.Medical.Cryogenics.CryoPodComponent, Content.Shared.Emag.Systems.GotEmaggedEvent>(OnEmagged);
```
instead of:
```cs
SubscribeLocalEvent<Content.Shared.Medical.Cryogenics.CryoPodComponent?, Content.Shared.Emag.Systems.GotEmaggedEvent>(OnEmagged);
```

Fixes https://github.com/space-wizards/RobustToolbox/issues/6873.

### DIFF
--- a/Robust.Shared.EntitySystemSubscriptionsGenerator/EntitySystemSubscriptionGenerator.cs
+++ b/Robust.Shared.EntitySystemSubscriptionsGenerator/EntitySystemSubscriptionGenerator.cs
@@ -210,7 +210,7 @@ using JetBrains.Annotations;
             !TypeSymbolHelper.ImplementsInterface(componentType, IComponentTypeName))
             return null;
 
-        return [componentType.ToString(), eventType.ToString()];
+        return [componentType.WithNullableAnnotation(NullableAnnotation.NotAnnotated).ToString(), eventType.ToString()];
     }
 
     private static SubscriptionInfo? TryParseSubscription(


deltanedas:
this shouldnt be a thing in content anyway it will never be null to begin with

Tayrtahn:
> this shouldnt be a thing in content anyway it will never be null to begin with

Oh yeah, I agree! I've found one instance upstream of it being used in an almost-justifiable way, though one might argue that it's event handler abuse:
https://github.com/space-wizards/space-station-14/blob/bbe57ec6e0823f69504389864623505041700d91/Content.Client/Chemistry/EntitySystems/ChemistryGuideDataSystem.cs#L32-L50
(The event is marked as nullable so that it can be called both by the event bus and manually in the system using `null` as the parameter)
(I'm now realizing that this PR doesn't even fix this particular example, which is fun)

deltanedas:
still abuse it should just have an actual api, dont enable this evil practice

Tayrtahn:
Yeah, thinking about it more, I agree that this isn't the right approach. I fixed the violations in content with https://github.com/space-wizards/space-station-14/pull/45070.

Maybe we can throw together an analyzer that catches these, or we can add it to the generator's validation analyzer.

PR #6885, opened 2026-08-01
Fix cross-compile CI for FreeBSD

Partmedia:
This reverts commit 3149158dde678cf7b0d255c605d22976a2e2fccc (#6848), then runs the actual build part of publish_client_builds.py.

FreeBSD was never intended to be a direct build target. Instead, it was supposed to be cross-compiled and shipped as a portable assembly so that signed engine builds would be available via the official launcher.

**This build will still fail, but it won't fail for the wrong reasons.**

### DIFF
--- a/.github/workflows/build-all-configurations.yml
+++ b/.github/workflows/build-all-configurations.yml
@@ -16,40 +16,9 @@ jobs:
   build:
     strategy:
       matrix:
-        include:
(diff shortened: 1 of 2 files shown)


metalgearsloth:
lol timing

metalgearsloth:
We're so fucking back.

PR #6887, opened 2026-08-01
 Make ComponentTreeEntry faster

AJCM-git:
It didnt implement GetHashCode and one of the most expensive operations was the hashset lookup

I estimate its like ~30% faster or something by roughly comparing the usages of QueryAabb in an overlay before and after, though there may be some noise
Before:
image

After:
image


ArtisticRoomba:
thats gore

PR #6891, opened 2026-08-01
Do not serialize components pending removal

### LINKED ISSUE #6890: Components pending deferred removal can still be saved
Components queued for deferred removal can still be serialized if an entity or map is saved before the end of the tick.
I encountered this with `ActiveInputMoverComponent`:
1. During map loading, entities are started before their paused state is restored.
2. `ActiveInputMoverComponent` is added because the entity still appears unpaused.
3. Restoring the paused state queues the component for deferred removal.
4. The map is saved immediately afterwards, before deferred removals are processed, so the component is written to YAML.
The relevant code is `EntitySerializer.SerializeComponents()`:
https://github.com/space-wizards/RobustToolbox/blob/master/Robust.Shared/EntitySerialization/EntitySerializer.cs

It iterates over `EntMan.GetComponentsInternal(uid)` and only skips components registered as `Unsaved`. It does not check whether a component has already been marked for removal.
Found while working on:
https://github.com/space-wizards/space-station-14/pull/44836

AndrewFenriz:
Prevents components queued for deferred removal from being serialized before the end of the tick.

This fixes the SaveLoadSave failure exposed by space-wizards/space-station-14#44831 and replaces the content-side workaround attempted in space-wizards/space-station-14#44836.

Resolves #6890.

### DIFF
--- a/Robust.Shared.IntegrationTests/EntitySerialization/PendingComponentRemovalSerializationTest.cs
+++ b/Robust.Shared.IntegrationTests/EntitySerialization/PendingComponentRemovalSerializationTest.cs
@@ -1,127 +1,49 @@
 using NUnit.Framework;
 using Robust.Shared.EntitySerialization.Systems;
 using Robust.Shared.GameObjects;
-using Robust.Shared.IoC;
 using Robust.Shared.Map;
-using Robust.Shared.Reflection;
-using Robust.Shared.Serialization.Markdown.Mapping;
-using Robust.Shared.Serialization.Markdown.Sequence;
-using Robust.Shared.Serialization.Markdown.Value;
+using Robust.Shared.Utility;
 
 namespace Robust.UnitTesting.Shared.EntitySerialization;
 
 [TestFixture]
-internal sealed partial class PendingComponentRemovalSerializationTest : RobustIntegrationTest
+internal sealed class PendingComponentRemovalSerializationTest : RobustIntegrationTest
 {
-    private const string InputComponentName = "PendingRemovalInput";
-    private const string ActiveComponentName = "PendingRemovalActive";
-    private const string PrototypeId = "PendingRemovalSerializationPrototype";
-    private const string TestPrototype = $"""
-        - type: entity
-          id: {PrototypeId}
-          components:
-          - type: {InputComponentName}
-        """;
-
     [Test]
-    public async Task PendingComponentAfterPausedMapLoadIsNotSerialized()
+    public async Task PendingComponentIsNotSerialized()
     {
-        var options = new ServerIntegrationOptions
-        {
-            Pool = false,
-            ExtraPrototypes = TestPrototype,
-        };
-        options.BeforeRegisterComponents += () =>
-        {
-            var factory = IoCManager.Resolve<IComponentFactory>();
-            factory.RegisterClass<PendingRemovalInputComponent>();
-            factory.RegisterClass<PendingRemovalActiveComponent>();
-        };
-        options.BeforeStart += () =>
-        {
-            var systemManager = IoCManager.Resolve<IEntitySystemManager>();
-            systemManager.LoadExtraSystemType<PendingRemovalSerializationSystem>();
-        };
-
-        var server = StartServer(options);
+        var server = StartServer();
         await server.WaitIdleAsync();
 
-        var mapSystem = server.System<SharedMapSystem>();
+        var entMan = server.EntMan;
         var loader = server.System<MapLoaderSystem>();
-        var testSystem = server.System<PendingRemovalSerializationSystem>();
-        var source = string.Empty;
+        var mapSystem = server.System<SharedMapSystem>();
+        var path = new ResPath($"{nameof(PendingComponentIsNotSerialized)}.yml");
+        MapId mapId = default;
 
-        await server.WaitAssertion(() =>
+        await server.WaitPost(() =>
         {
-            var map = mapSystem.CreateMap(out var mapId, runMapInit: false);
-            var uid = server.EntMan.SpawnEntity(PrototypeId, new MapCoordinates(0, 0, mapId));
+            mapSystem.CreateMap(out mapId);
+            var uid = entMan.SpawnEntity(null, new MapCoordinates(0, 0, mapId));
+            var component = entMan.AddComponent<PendingRemovalTestComponent>(uid);
 
-            // Pre-map-init entities are paused, so the active component is absent from the original save.
-            Assert.That(server.EntMan.GetComponent<MetaDataComponent>(uid).EntityPaused, Is.True);
-            Assert.That(server.EntMan.HasComponent<PendingRemovalActiveComponent>(uid), Is.False);
-
-            using var writer = new StringWriter();
-            Assert.That(loader.TrySaveMap(map, writer), Is.True);
-            source = writer.ToString();
-
-            server.EntMan.DeleteEntity(map);
+            // Deferred removals are processed at the end of the tick. Saving before then used to serialize the
+            // component even though it had already been queued for removal.
+            entMan.RemoveComponentDeferred(uid, component);
+            Assert.That(entMan.Count<PendingRemovalTestComponent>(), Is.EqualTo(1));
+            Assert.That(loader.TrySaveMap(mapId, path), Is.True);
         });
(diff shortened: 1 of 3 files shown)


AndrewFenriz:
Tested with the changes from #44831, the test passed successfully. Yaay?

AJCM-git (changes requested):
Make a test replicating the situation in content, shouldnt be too difficult. Of course double check it fails before the patch

AndrewFenriz:
> Make a test replicating the situation in content, shouldnt be too difficult. Of course double check it fails before the patch

I examined the existing tests and got a basic understanding of how they are structured (and did a bit of copy-pasting). As expected, the test failed before the fix:

```
Assert.That(HasSerializedComponent(serialized, ActiveComponentName), Is.False)
  Expected: False
  But was:  True
   at Robust.UnitTesting.Shared.EntitySerialization.PendingComponentRemovalSerializationTest.PendingComponentAfterPausedMapLoadIsNotSerialized() in C:\Users\kasto\RiderProjects\SS14\RobustToolbox\Robust.Shared.IntegrationTests\EntitySerialization\PendingComponentRemovalSerializationTest.cs:line 85

   at Robust.UnitTesting.Shared.EntitySerialization.PendingComponentRemovalSerializationTest.PendingComponentAfterPausedMapLoadIsNotSerialized() in C:\Users\kasto\RiderProjects\SS14\RobustToolbox\Robust.Shared.IntegrationTests\EntitySerialization\PendingComponentRemovalSerializationTest.cs:line 85
   at NUnit.Framework.Internal.TaskAwaitAdapter.GenericAdapter`1.BlockUntilCompleted()
   at NUnit.Framework.Internal.MessagePumpStrategy.NoMessagePumpStrategy.WaitForCompletion(AwaitAdapter awaiter)
   at NUnit.Framework.Internal.AsyncToSyncAdapter.Await[TResult](TestExecutionContext context, Func`1 invoke)
   at NUnit.Framework.Internal.AsyncToSyncAdapter.Await(TestExecutionContext context, Func`1 invoke)
   at NUnit.Framework.Internal.Commands.TestMethodCommand.RunTestMethod(TestExecutionContext context)
   at NUnit.Framework.Internal.Commands.TestMethodCommand.Execute(TestExecutionContext context)
   at NUnit.Framework.Internal.Execution.SimpleWorkItem.<>c__DisplayClass3_0.<PerformWork>b__0()
   at NUnit.Framework.Internal.ContextUtils.<>c__DisplayClass1_0`1.<DoIsolated>b__0(Object _)
```

Robust.Shared.IntegrationTests/EntitySerialization/PendingComponentRemovalSerializationTest.cs
```diff
@@ -0,0 +1,127 @@
...
+            .SelectMany(group => group.Get<SequenceDataNode>("entities").Cast<MappingDataNode>())
+            .Any(entity => entity.TryGet<SequenceDataNode>("components", out var components)
+                && components.Cast<MappingDataNode>()
+                    .Any(component => component.Get<ValueDataNode>("type").Value == componentName));
+    }
+
+    [Reflect(false)]
+    private sealed partial class PendingRemovalSerializationSystem : EntitySystem
+    {
+        public int ActiveComponentsAdded { get; private set; }
```
AJCM-git: i think you can use entman.count for this

Robust.Shared.IntegrationTests/EntitySerialization/PendingComponentRemovalSerializationTest.cs
```diff
@@ -0,0 +1,127 @@
...
+        public int ActiveComponentsQueuedForRemoval { get; private set; }
```
AJCM-git: maybe can use count for this as well? removal is at the end of the tick after all

Robust.Shared.IntegrationTests/EntitySerialization/PendingComponentRemovalSerializationTest.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        var mapSystem = server.System<SharedMapSystem>();
+        var loader = server.System<MapLoaderSystem>();
+        var testSystem = server.System<PendingRemovalSerializationSystem>();
+        var source = string.Empty;
+
+        await server.WaitAssertion(() =>
+        {
+            var map = mapSystem.CreateMap(out var mapId, runMapInit: false);
+            var uid = server.EntMan.SpawnEntity(PrototypeId, new MapCoordinates(0, 0, mapId));
```
AJCM-git: make an entman var

Robust.Shared.IntegrationTests/EntitySerialization/PendingComponentRemovalSerializationTest.cs
```diff
@@ -0,0 +1,127 @@
...
+            systemManager.LoadExtraSystemType<PendingRemovalSerializationSystem>();
+        };
+
+        var server = StartServer(options);
+        await server.WaitIdleAsync();
+
+        var mapSystem = server.System<SharedMapSystem>();
+        var loader = server.System<MapLoaderSystem>();
+        var testSystem = server.System<PendingRemovalSerializationSystem>();
+        var source = string.Empty;
```
AJCM-git: You can probably handle the map like AlwaysPushSerializationTest does, looks like a little less jank

Robust.Shared.IntegrationTests/EntitySerialization/PendingComponentRemovalSerializationTest.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            Pool = false,
+            ExtraPrototypes = TestPrototype,
+        };
+        options.BeforeRegisterComponents += () =>
+        {
+            var factory = IoCManager.Resolve<IComponentFactory>();
+            factory.RegisterClass<PendingRemovalInputComponent>();
+            factory.RegisterClass<PendingRemovalActiveComponent>();
+        };
```
AJCM-git: delete after previous review

Robust.Shared.IntegrationTests/EntitySerialization/PendingComponentRemovalSerializationTest.cs
```diff
@@ -0,0 +1,127 @@
...
+            RemCompDeferred<PendingRemovalActiveComponent>(entity);
+            ActiveComponentsQueuedForRemoval++;
+        }
+    }
+
+    [ComponentProtoName(InputComponentName)]
+    private sealed partial class PendingRemovalInputComponent : Component;
+
+    [ComponentProtoName(ActiveComponentName)]
+    private sealed partial class PendingRemovalActiveComponent : Component;
```
AJCM-git: registercomp and use the default comp name. make it internal. PendingRemovalInputComponent is a bad name because this test has nothing to do with input, same for field name

Robust.Shared.IntegrationTests/EntitySerialization/PendingComponentRemovalSerializationTest.cs
```diff
@@ -0,0 +1,127 @@
...
+        await server.WaitAssertion(() =>
+        {
+            using var reader = new StringReader(source);
+            Assert.That(loader.TryLoadMap(reader, nameof(PendingComponentAfterPausedMapLoadIsNotSerialized),
+                out var loadedMap, out _), Is.True);
+
+            // Loading starts the entity before restoring its paused state. This temporarily adds the active component,
+            // then EntityPausedEvent queues it for deferred removal.
+            Assert.That(testSystem.ActiveComponentsAdded, Is.EqualTo(1));
+            Assert.That(testSystem.ActiveComponentsQueuedForRemoval, Is.EqualTo(1));
```
AJCM-git: I think this is another issue entirely, so maaaybe It would be better to not put a system in this test and get rid of PendingRemovalInputComponent and just have 1 component that you add manually, defer the deletion and then save the map, this should narrow the focus of the test while still failing before the patch right?

AJCM-git (changes requested):
prob needs a couple comments describing the original issue in the test, also a few more comments. You can use server.metadata() to get the component btw

AJCM-git (approved):

PR #6897, opened 2026-08-02
Intern prototype components

metalgearsloth:
- The majority of component data is similar, especially when children inherit parent prototypes.
- Component data is theoretically treated as read-only
- Can drop headless client from 400MB to 350MB of usage (the other 10MB is some other pending change).

image


DrSmugleaf (commented):
[YamlDotNet 17.0.0 and above](https://github.com/aaubry/YamlDotNet/releases/tag/v17.0.0) interns mapping keys now but not anything else, not sure how much that alone saves

metalgearsloth:
> [YamlDotNet 17.0.0 and above](https://github.com/aaubry/YamlDotNet/releases/tag/v17.0.0) interns mapping keys now but not anything else, not sure how much that alone saves

Client said 399MB and normally it's 403MB not sure if rounding error or actual savings.

PR #6900, opened 2026-08-02
Revert UI queueing (#6789, #6829, #6840)

whatston3:
Reverts #6789, #6829, #6840.

It's a good concept, a nice feature, but it breaks on edge cases and needs a test suite for correctness.

Not giving up on it, and I think it's good conceptually, but _man_.


Princess-Cheeseballs (approved):
Queued Updates are pretty peak but yeah they often break if you're not careful. Worth pursuing in the future:tm:

whatston3:
Sorry for the saga, I'll try to make sure the next one's right.

Princess-Cheeseballs:
> Sorry for the saga, I'll try to make sure the next one's right.

No need to apologize, this is a normal part of the learning process. Now you know how to do better.

PR #6901, opened 2026-08-02
Reduce memory usage substantially

metalgearsloth:
- Remove unnecessary mappings after resolving prototypes. The downside is we make TryGetMapping much slower (which is used rarely and on slowpaths) but we also significantly reduce memory usage.
- Shallow copy parent nodes and pretend prototype datanodes are readonly until we get a better read-only API (because they really, really should be). If a child changes 1 component it shouldn't deep-copy 1 morbillion nodes.
- Remove 2 ints per datanode for 99.99% of them and use a scrunkly table for the remainders.
- Use array for sequencedatanode storage where we can get away with it otherwise fall back to list.

Client uses 400MB atm
image

Down to 290MB
image

Unfortunately the actual non-headless client uses a significant amount of memory for textures + audio (for obvious reasons) but for headless client this helps a lot.


deltanedas:
are xaml proxies even needed for headless it sounds like that can be skipped

metalgearsloth:
Ideally you'd run headless clients on release where (hopefully) xamlxproxymanager shouldn't be running, but that should also be disabled on debug or have a CVar for it yeah.

PR #6905, opened 2026-08-03
Add SCREEN_UV to shaders

metalgearsloth:
Postshaders are being drawn without any knowledge of the screen texture and the stealth shader on content needs to know its screen coords unfortunately.


metalgearsloth:
Normally I'd wait but content wants it bad so yell at me later.

PR #6908, opened 2026-08-03
Fix WordWrap bugs

eoineoineoin:
When testing https://github.com/space-wizards/space-station-14/pull/45085, I hit an assert in WordWrap; it's caused by the same issue which causes the strange wrapping described in https://github.com/space-wizards/RobustToolbox/pull/6393#issuecomment-4960932950

Reproing the assert is fairly straightforward; build content in debug and spawn a "paper". Empty paper has placeholder text "This page intentionally left blank" - resize the paper window so it's around the width of the word "intentionally". Maybe wiggle the mouse around a little bit; you'll hit the assert eventually. I've added some new tests, and running them at revision c0568fda5967d79f8d80a1327645990c2c1222dd will also repro the assert.

Reproing the wrapping also straightforward. Write on the paper with:

> asdf asdf aVeryLongWordThatNeedsToBeSplitMultipleTimes asdf asdf
 
Then resize the paper window, note the weird line breaks at locations that don't really make sense:

https://github.com/user-attachments/assets/d5c3e4dc-95ba-420e-8957-4cbba30942c3

Both of these problems were coming from the same issue - the WordWrap had a field `ForceSplitData`, which stored the index of the first rune which causes the current line to overflow. As best I can tell (paging Mr. Chesterson), this was trying to handle the case where, if a word needed splitting on account of being longer than a single line, the start of that word could be placed on the current line, optimizing vertical space. i.e.

```
asdf asdf aVer
yLongWordThatN
eedsTo...
```

but this `ForceSplitData` conflicted with the "split at a word boundary" code that handles the rest of the splitting. The assert was being hit because we were seeing two conflicting split locations - one at the `ForceSplitData` and one at the location where the full word no longer fits on a single line.

That kind of split above is also different to how every other gui text tool wraps (at least, the ones I've tested.) - those tools would *not* try to optimize vertical space and would instead lay the text out like:

```
asdf asdf
aVeryLongWordT
hatNeedsTo...
```

This PR just removes that `ForceSplitData` entirely, which makes the wrapping behave a lot more sane (and not hit asserts):

https://github.com/user-attachments/assets/43170601-000c-4a28-b4d1-0aa93cf9cc5b

### DIFF
--- a/Robust.Client.Tests/UserInterface/WordWrapTest.cs
+++ b/Robust.Client.Tests/UserInterface/WordWrapTest.cs
@@ -1,0 +1,89 @@
+using NUnit.Framework;
+using Robust.Client.Graphics;
+using Robust.Client.UserInterface;
+using System.Text;
+
+namespace Robust.Client.Tests.UserInterface;
+
+[Parallelizable(ParallelScope.All)]
+public sealed class WordWrapTest
+{
+    private List<int> GenerateBreaks(string s, int maxWidth)
+    {
+        var breaksOut = new List<int>();
+        var wrapper = new WordWrap(maxSizeX: maxWidth);
+
+        // For simplicity, assume every character has the same width, except for some special ones
+        var charMetrics = new CharMetrics (bearingX: 0, bearingY: 0, advance: 10, width: 10, height: 10);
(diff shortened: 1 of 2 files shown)


metalgearsloth (approved):

metalgearsloth:
Awesome good stuff

PR #6904, opened 2026-08-02
New Feature: ``before`` and ``after`` support for subscriptions through attributes

Winkarst-cpu:
Fixes #6706.

## Example subscription
```cs
[SubscribeLocalEvent([typeof(EmptyOnMachineDeconstructSystem), typeof(ItemSlotsSystem)])]
private void OnMachineDeconstructed(Entity<IdCardConsoleComponent> entity, ref MachineDeconstructedEvent args)
```

## Generated code
```cs

// <auto-generated />

using Robust.Shared.GameObjects;
using JetBrains.Annotations;

#pragma warning disable CS0618 // Type or member is obsolete: event handlers will have their own obsoletion warnings, dont need to dupe them

namespace Content.Server.Access.Systems;

public partial class IdCardConsoleSystem
{
    /// <inheritdoc />
    [MustCallBase]
    public override void AutoSubscriptions()
    {
        base.AutoSubscriptions();

        SubscribeLocalEvent<Content.Shared.Access.Components.IdCardConsoleComponent, Content.Shared.Construction.MachineDeconstructedEvent>(OnMachineDeconstructed, [typeof(Content.Server.Containers.EmptyOnMachineDeconstructSystem), typeof(Content.Shared.Containers.ItemSlots.ItemSlotsSystem)], null);

    }
}
```

PR #6909, opened 2026-08-03
Add a Code Fix for converting EntitySystem event subscriptions to use attributes

Tayrtahn:
Adds an analyzer and code fix to detect instances of SubscribeXEvent method calls and replace them with the event subscription attributes added by https://github.com/space-wizards/RobustToolbox/pull/6227.

There are quite a few limitations on which instances can be replaced automatically, but this currently covers 4794 of them in engine and content, which I think is pretty good.

The analyzer detects SubscribeXEvent method invocations that are eligible for automatic conversion and marks them with an Info diagnostic:
Screenshot 2026-08-03 at 1 07 09 AM
(They show as `...` in vscode, and I believe as a green squiggle in Rider)

Clicking Quick Fix or the lightbulb presents the option for the code fix:
Screenshot 2026-08-03 at 1 07 37 AM

The code fix:
- Deletes the method invocation.
- Adds the appropriate `[SubscribeXEvent]` attribute to the event handler method.
- Adds `partial` to the class if it doesn't already have it.
- Adds `using Robust.Shared.Analyzers` if needed.

This works even if the event handler is in a different file (for systems spread across multiple partial files).

If this deletes all the subscribe invocations from a method, it _will_ leave behind an empty method or one that just calls its base. I investigated trying to automatically remove the method in those cases, but it wound up being quite complicated and I don't think it's worth trying to sort it all out.

There are a bunch of cases where the subscribe invocation can't be directly converted to use an attribute. The analyzer will not mark these cases; they will need manual conversion:
- Subscriptions that are not at the top level of the containing method, such as those within an if statement, or a while loop. There's no way to conditionally apply the attribute, so these are skipped.
- Subscriptions in classes that have any conditional preprocessor directives (`#if`, `#else`, etc). These are too complicated to try to parse and work out how to automatically convert, and there are (thankfully) very few of them.
- Event handlers that aren't explicitly defined class methods, such as subscriptions using lambda expressions (`SubscribeLocalEvent<MyComp, MyEvent>((u, c, ref _) => MyMethod(u, c, e));`)
- ~Subscriptions with `before` and `after` parameters are currently unsupported by the attributes, so the fixer ignores them. [If support is added](https://github.com/space-wizards/RobustToolbox/pull/6904), I'll probably update this to work with it.~ Support added! Code fix updated!
- Subscriptions using a generic type parameter: `SubscribeLocalEvent<TComp, MyEvent>(MyMethod);`
- Event handlers that are generic methods: `private MyMethod<T>(Entity<MyComp> entity, ref T args) { }`
- Event handlers with abstract event types: `private MyMethod(Entity<MyComp> entity, ref BaseMyEvent args) { }` (where `BaseMyEvent` is an abstract type)

With these restrictions, I was able to blindly run Fix All on the full engine+content solution and produce code that compiles correctly. If downstream content runs into any additional edge cases, please let me know!


Winkarst-cpu:
#6904 was merged. Can you add support for it?

Tayrtahn:
Support added for conversions with `before` and `after`!

Input:
Screenshot 2026-08-05 at 11 22 26 PM
Converted:
Screenshot 2026-08-05 at 11 22 50 PM

PR #6915, opened 2026-08-04
Make SerializationManager.TryGetTypeNodeSerializer 65 times faster

DrSmugleaf:
Also fixes serializers not getting PostInject called, and makes the IsArray check faster
This is the PR with the 30 thousand long array
Can probably remove the copier-specific array with this


ArtisticRoomba:
65 times the detail

PR #6918, opened 2026-08-05
IOC fixes

metalgearsloth:
Just the array length + 1 fix and reverse lookup building.


metalgearsloth:
Okay I was gonna leave this but https://github.com/space-wizards/RobustToolbox/pull/6920 got satanic hellbug.

PR #6920, opened 2026-08-05
Optimise full state updates

metalgearsloth:
Collectionsmarshal once is faster than 2 dictionary indexes, do it in two spots.


DrSmugleaf (approved):

Robust.Client/GameStates/GameStateProcessor.cs
```diff
@@ -215,30 +215,34 @@ public void UpdateFullRep(GameState state, bool cloneDelta = false)
...
-                    _lastStateFullRep.Add(entityState.NetEntity, compData);
-                }
+                ref var compDataRef = ref CollectionsMarshal.GetValueRefOrAddDefault(
+                    _lastStateFullRep,
+                    entityState.NetEntity,
+                    out var compDataExists);
 
+                if (!compDataExists)
+                    compDataRef = new();
+
```
metalgearsloth, line 228:
```suggestion
                if (!compDataExists)
                {
                    var componentCount = entityState.NetComponents?.Count ?? entityState.ComponentChanges.Span.Length;
                    compDataRef = new(componentCount);
                }

```

PR #6923, opened 2026-08-06
Add partial prototypes, for defining and modifying prototype data with the same ID in different files

DrSmugleaf:
# Content has a system to do this automatically with a yml file in Resources/PartialPrototypes
Content PR https://github.com/space-wizards/space-station-14/pull/45453

Doc bible attached

For downstreams of downstreams, you can specify the load order:
```csharp
PartialDirectory(
    new ResPath("/Prototypes/_ForkOne"),
    new ResPath("/Prototypes/_ForkTwo")
);
```

Or
```csharp
PartialDirectory(new ResPath("/Prototypes/_ForkOne"));
PartialDirectory(new ResPath("/Prototypes/_ForkTwo"));
```

Non partial directories are loaded before any partial ones.

# Examples:
Add three loadouts and remove one:
```yml
- type: loadoutGroup
  id: MailTechnicianJumpsuit
  loadouts:
  - CargoTechnicianJumpsuitAlt
  - CargoTechnicianJumpsuitFormal
  - CargoTechnicianJumpskirtFormal
  - !Remove MailTechnicianJumpsuitSenior
```

Remove LanguageKnowledgeComponent and remove spawned from Butcherable:
```yml
- type: entity
  id: BaseMobHuman
  components:
  - !Remove type: LanguageKnowledge
  - type: Butcherable
    spawned: !Remove
```

To make it not add the prototype if the original one does not exist or is later removed:
```yml
- type: !PartialOnly entity
  id: BaseMobHuman
```

To reorder an existing module and insert a new one at specific indexes:
```yml
- type: !PartialOnly borgType
  id: MyBorgType
  defaultModules:
  - !Index:0 BorgModuleNew
  - !Remove BorgModuleExisting
  - !Index:-1 BorgModuleExisting
```

Combined with variants, it can apply to more than one prototype:
```yml
- type: !PartialOnly listing
  id: !type:CreateVariants
  values:
  - VendorPlushieHuman
  - VendorPlushieMoth
  - VendorPlushieVulp
  conditions: !Remove
```


Robust.Server/Prototypes/ServerPrototypeManager.cs
```diff
@@ -53,6 +54,7 @@ private void HandleReloadPrototypes(MsgReloadPrototypes msg)
 
         public override void LoadDefaultPrototypes(Dictionary<Type, HashSet<string>>? changed = null)
         {
+            PartialDirectory(new ResPath("/Prototypes/_Afterlight/"));
```
kontakt: Afterlight string

Robust.Shared/Prototypes/IPrototypeManager.cs
```diff
@@ -561,6 +561,94 @@ void ReloadPrototypes(
     /// </summary>
     bool TryGetRandom<T>(IRobustRandom random, [NotNullWhen(true)] out IPrototype? prototype) where T : class, IPrototype;
 
+    /// <summary>
+    ///     Like <see cref="PartialDirectory"/>, but only for a single file.
```
kontakt: Convention is no indent on xmldoc.

deltanedas:
should have a distinct syntax (probably partialType instead of type) so its easy to rg, as well as always being an error if a non-partial one doesnt exist by default imo. otherwise the lazy way of doing it for 99% of usecases results in broken prototypes that might survive yml linter

DrSmugleaf:
> should have a distinct syntax (probably partialType instead of type) so its easy to rg, as well as always being an error if a non-partial one doesnt exist by default imo. otherwise the lazy way of doing it for 99% of usecases results in broken prototypes that might survive yml linter

One of the big things this is trying to fix is porting YML between downstreams being mighty annoying, if it always threw an error that would still be almost as annoying
I originally had it throw an error, I could make that configurable if that's wanted, but without it it makes it easier to lint away if you wanted to, if it threw an exception you would need to catch that in the linter, or a test

deltanedas:
with no error:
- you have a useless broken prototype with no warning

with an error:
- it tells you exactly which prototypes dont actually exist so you can just remove them or investigate getting them

DrSmugleaf:
> with no error:
> 
> * you have a useless broken prototype with no warning
> 
> with an error:
> 
> * it tells you exactly which prototypes dont actually exist so you can just remove them or investigate getting them

PartialOnly does not leave behind a prototype if the original doesn't exist, there's no broken prototype behind, the only thing you get is an additional mapping
I don't know if you haven't had contributors port YML from one downstream to another before but they usually have to make half a dozen commits trying to fix linter errors or irrelevant prototypes that get parsed as actual prototypes later

Robust.Server/Prototypes/ServerPrototypeManager.cs
DrSmugleaf: Fixed in [`fae32f4` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6923/commits/fae32f448a0a6910060f38912f74514f43345f1a)

Robust.Shared/Prototypes/IPrototypeManager.cs
DrSmugleaf: Fixed in [`3c30602` (this PR)](https://github.com/space-wizards/RobustToolbox/pull/6923/commits/3c30602d6e16f884a403392dfa896a4932d6eec9)

DrSmugleaf:
Also the important thing is that the error should not be in engine, that makes it very hard to work around in a downstream where you don't want it to error, and borderline impossible to allow it for live admin uploads
Content.YAMLLinter is in content and you can put that there as an explicit check

metalgearsloth (approved):

DrSmugleaf:
image

sleepyyapril:
peak

PR #6924, opened 2026-08-06
Make ToolshedCommand not a downgrade of LocalizedCommands

DrSmugleaf:
Content PR https://github.com/space-wizards/space-station-14/pull/45579

This shit proper pissing me off
Same as LocalizedCommands, only with two attributes since you can't add a property to a method
Localized strings still take priority over CommandDescriptionAttribute and CommandHelpAttribute
The localization key is returned if neither exist, same as before


ElectroJr:
The engine `AllCommandsHaveDescriptions` test should be updated to check for the attribute, otherwise LGTM. The equivalent content test can just be updated whenever the feature gets used.

also release notes

DrSmugleaf:
> The engine `AllCommandsHaveDescriptions` test should be updated to check for the attribute, otherwise LGTM. The equivalent content test can just be updated whenever the feature gets used.
> 
> also release notes

Fixed in https://github.com/space-wizards/RobustToolbox/pull/6924/commits/f0ba3a466fc9ff97c403102711aae04cdf017bfb and https://github.com/space-wizards/RobustToolbox/pull/6924/commits/63abc7b12462a89d29c16599474e30d711dfee13

ElectroJr (approved):

PR #6925, opened 2026-08-06
Make CVars allow EntProtoId and EntProtoId<T> and automatically validate them

DrSmugleaf:
Spicy meatball PR 
Tired of typing them as strings and then having to validate them in Content manually
This PR does not add support for other prototype types, because I live a happy life downstream with the only prototype kind that matters being EntityPrototype

Content PR (not needed to merge this) https://github.com/space-wizards/space-station-14/pull/45169

Example output:
```
EntProtoId CVar failed validation. No entity prototype found with id abcd1234.
EntProtoId CVar failed validation. Entity prototype 'ActionFireStarter' does not have a component of type 'Buckle'.
```

### DIFF
--- a/Robust.Shared/Configuration/ConfigurationCommands.cs
+++ b/Robust.Shared/Configuration/ConfigurationCommands.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using Robust.Shared.Console;
 using Robust.Shared.IoC;
 using Robust.Shared.Localization;
+using Robust.Shared.Prototypes;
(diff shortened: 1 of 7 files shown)


deltanedas:
this should be using type serializers not hardcododing entprotoid like this

DrSmugleaf:
> this should be using type serializers not hardcododing entprotoid like this

I have bad news about what the entire configuration manager does

deltanedas:
so

PR #6930, opened 2026-08-06
Update yamldotnet

metalgearsloth:
17.0.0 has interniong


DrSmugleaf (approved):
Saves me having to do it in the YAML source generator PR

PR #6932, opened 2026-08-06
Fix "spawn tiles" window UIBox2 errors

eoineoineoin:
Like https://github.com/space-wizards/space-station-14/pull/45085, but now for engine!

If you open the sandbox menu and hit the "spawn tiles" button, you'll see UIBox2 failing validation. Some of the tile textures had multiple variants, and the ItemList had not enough room for displaying text. 8969a8f17d7591b365326ae0c88229c2219b82dd just added a sanity Math.Min to make that piece robust.

The tile menu displays all the variants of the tile, which, IMO looks bad:

TilesBefore

The ItemList supports showing only a subset of a texture, but this feature was unused and didn't work. afea191f8121b756236df0d493389321800b4120 fixes that and displays only the first tile variant in that UI:

TilesAfter

A few tile types have the same name and same variant0 (see the "asteroid sand" in that screenshot), so probably need to do something there? Animate switching between the variants? Just give them different names? An explicit icon in the tile prototype? Open to suggestions.

### DIFF
--- a/Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
+++ b/Robust.Client/UserInterface/Controllers/Implementations/TileSpawningUIController.cs
@@ -1,6 +1,5 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
-using System.Drawing;
 using System.Linq;
 using Robust.Client.Graphics;
 using Robust.Client.Placement;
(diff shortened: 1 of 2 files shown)


deltanedas:
it should show the id on hover like entity spawn menu at least

PR #6937, opened 2026-08-07
Make FOV slightly blue

metalgearsloth:
Less harsh than black but it's so hard to tell on screenshots.

image

### DIFF
--- a/Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
+++ b/Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
@@ -998,7 +998,7 @@ namespace Robust.Client.Graphics.Clyde
             fovShader.SetUniformTextureMaybe(UniIMainTexture, TextureUnit.Texture0);
 
             if (!Color.TryParse(_cfg.GetCVar(CVars.RenderFOVColor), out var color))
-                color = Color.Black;
+                color = Color.FromHex("#080B10");
 
             fovShader.SetUniformMaybe("occludeColor", color);
             FovSetTransformAndBlit(viewport, eye.Position.Position, fovShader);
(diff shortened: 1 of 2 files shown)


Pok27:
I don't like that 😥

DogZeroX:
I like that 😁

Pok27:
It doesn't look very good with the black stripes
image

ScarKy0:
Maybe could be a clientside cvar?

metalgearsloth:
> It doesn't look very good with the black stripes image

image

Not sure how flooring shadows to it looks but will consult with kazne.

I also wanted to try fog of war stuff but that's a huge redesign and effort so unlikely to happen.

Robust.Client/Graphics/Clyde/Clyde.LightRendering.cs
```diff
@@ -998,7 +998,7 @@ private void ApplyFovToBuffer(Viewport viewport, IEye eye)
             fovShader.SetUniformTextureMaybe(UniIMainTexture, TextureUnit.Texture0);
 
             if (!Color.TryParse(_cfg.GetCVar(CVars.RenderFOVColor), out var color))
-                color = Color.Black;
+                color = Color.FromHex("#080B10");
```
whatston3, line 1001: Could this string be defined as a constant in cvars?

mirrorcult:
is there a reason this is being changed in engine instead of just altering `RenderFOVColor` in a config preset upstream?

metalgearsloth:
> is there a reason this is being changed in engine instead of just altering `RenderFOVColor` in a config preset upstream?

image

But it hinged upon the artistic side though I don't think it's getting merged anyway.

PR #6938, opened 2026-08-07
Move RA0056 check to the EntitySystemSubscriptionGeneratorErrorAnalyzer

### LINKED ISSUE #6771: RA0056 Analyzer does not work
Image

It should give you an error if you have a `SubscribeLocalEventAttribute` in a non-partial system, but it compiles fine.

Image

If it inherits a shared system that is partial it seems to just not generate any code at all.

And non-inherited systems will compile fine in Debug, but in Release it gives the following error,

Image

```
CSC : error CS8785: Generator 'EntitySystemSubscriptionGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'NullReferenceException' with message 'Object reference not set to an instance of an object.'.
```

Winkarst-cpu:
Fixes #6771.

Screenshot_20260807_171617

Breaking changes:
Removed PartialTypeInfo.CheckPartialDiagnostic since it was obsolete and unused.
Removed PartialTypeInfo.SyntaxLocation since it became unused.

### DIFF
--- a/Robust.Roslyn.Shared/PartialTypeHelper.cs
+++ b/Robust.Roslyn.Shared/PartialTypeHelper.cs
@@ -16,7 +16,6 @@ public sealed record PartialTypeInfo(
     string? Namespace,
     EquatableArray<PartialTypeInfo.NestedPart> Parts,
     bool IsValid,
-    Location SyntaxLocation,
     bool IsSealed)
 {
     public string Name => Parts[^1].Name;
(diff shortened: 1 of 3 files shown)


Tayrtahn:
Am I right in thinking that this removes the only use of `PartialTypeInfo.Location`? If so, it would probably be good to remove that field from `PartialTypeInfo` so instances can be compared without needing to remember to use the custom comparer (since forgetting to use it _very_ costly!)

PR #6940, opened 2026-08-08
Fix occluder perf

metalgearsloth:
Caches embedded lights and fixed allocs

### DIFF
--- a/Robust.Client/GameObjects/EntitySystems/ClientOccluderSystem.cs
+++ b/Robust.Client/GameObjects/EntitySystems/ClientOccluderSystem.cs
@@ -2,6 +2,7 @@ using JetBrains.Annotations;
 using Robust.Shared.Maths;
 using Robust.Shared.GameObjects;
 using Robust.Shared.IoC;
+using Robust.Shared.Map;
 using Robust.Shared.Physics;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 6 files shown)


Robust.Client/GameObjects/EntitySystems/ClientOccluderSystem.cs
```diff
@@ -105,24 +118,128 @@ private void OnShutdown(EntityUid uid, OccluderComponent comp, ComponentShutdown
...
+        }
+
+        var worldBounds = XformSystem.GetWorldMatrix(treeXform, _xformQuery)
+            .TransformBox(treeBounds.Enlarged(SharedOccluderEdgeTolerance));
+
+        try
+        {
+            foreach (var lightUid in _embeddedLightCache)
+            {
+                if (!_pointLightQuery.TryGetComponent(lightUid, out var light) ||
```
DrSmugleaf, line 218: If this is already properly getting removed on ComponentRemove would it not be safe to store the component so you don't need to TryGet it here?

DrSmugleaf (approved):

Robust.Client/GameObjects/EntitySystems/ClientOccluderSystem.cs
metalgearsloth: Yeah.

TBH I may revisit this so gonna hold off on merging for now but handling all of the edge cases to drop shared edges is not great on client atm.

PR #6941, opened 2026-08-08
Grid-fixture refactor

metalgearsloth:
- Reduce unnecessary work.
- Speedup
- Re-centre split grids due to lerping problems (I'm pretty sure I'm the only person who's noticed this).
- Avoid running RegenerateCollision if the chunk's shape remains the same.
- Subby attribute and dependencies.

Wrote it before chents but uhh could probably use it at some point though might need larger changes due to clientside prediction.


metalgearsloth:
image

In test coverage we trust.

PR #6947, opened 2026-08-08
Fix AlwaysPushInheritance read writes resulting in inconsistent ordering for sequences

DrSmugleaf:
Might cause large diffs on maps would have to test
Previously it would do
parent: 1, 2, 3
child: 4, 5, 6
result: 4, 5, 6, 1, 2, 3

Now it always does
result: 1, 2, 3, 4, 5, 6

This is consistent with the documentation we already have, but changes the behaviour
AlwaysPushInheritanceAttribute states:
```
    /// <summary>
    ///     When inheriting a field from a parent, always <b>merge</b> the two fields when such behavior exists.
    ///     This is unlike the normal behavior where the child's field will always <b>overwrite</b> the parent's.
    ///     <br/>
    ///     Merging is done at a YAML level by merging mappings and sequences recursively.
    /// </summary>
    /// <example>
    ///     <code>
    ///     - id: Parent
    ///       myField: [Foo, Bar]
    ///     <br/>
    ///     - id: Child
    ///       parents: [Parent]
    ///       myField: [Baz, Qux]
    ///     </code>
    ///     Which, when deserialized and assuming myField is marked with AlwaysPushInheritance, will result in data that
    ///     looks like this:
    ///     <code>
    ///     - id: Child
    ///       myField: [Foo, Bar, Baz, Qux]
    ///     </code>
    ///     compared to the default behavior:
    ///     <code>
    ///     - id: Child
    ///       myField: [Baz, Qux]
    ///     </code>
    /// </example>
```

Mind you that wasn't written by Paul, so maybe he intended for it to end up the other way around, either way a test has been added to ensure consistency

The actual fix is 2 lines, the test is the rest of the diff

### DIFF
--- a/Robust.Shared/Serialization/Manager/SerializationManager.Composition.cs
+++ b/Robust.Shared/Serialization/Manager/SerializationManager.Composition.cs
@@ -152,9 +152,10 @@ public partial class SerializationManager
         // I have NFI what this comment means.
 
         var result = child.Copy();
+        var i = 0;
         foreach (var entry in parent)
         {
-            result.Add(entry.Copy());
+            result.Insert(i++, entry.Copy());
         }
 
         return result;
--- a/Robust.Shared.IntegrationTests/Serialization/AlwaysPushInheritanceTest.cs
+++ b/Robust.Shared.IntegrationTests/Serialization/AlwaysPushInheritanceTest.cs
@@ -28,17 +28,19 @@ public sealed partial class AlwaysPushInheritanceTest : OurSerializationTest
             [listTag] = new SequenceDataNode("3", "4", "5")
         };
 
-        var combined = Serialization.PushComposition<AlwaysPushInheritanceTestDefinition, MappingDataNode>(parent, child);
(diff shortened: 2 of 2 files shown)


DrSmugleaf:
@Princess-Cheeseballs

Princess-Cheeseballs:
Funnily enough fixes a different issue, but the issue we were having was that we needed a way for inheritance to automatically sort a list. I believe this can be done with an ITypeInheritanceSerializer but I need to look into it.

Ideally we should have in engine a serializer which can sort lists in ascending and descending order based on the type.

Robust.Shared/Serialization/Manager/SerializationManager.Composition.cs
```diff
@@ -152,9 +152,10 @@ private SequenceDataNode PushInheritanceSequence(SequenceDataNode child, Sequenc
         // I have NFI what this comment means.
 
         var result = child.Copy();
+        var i = 0;
         foreach (var entry in parent)
         {
-            result.Add(entry.Copy());
+            result.Insert(i++, entry.Copy());
         }
```
ElectroJr, line 159: Wouldn't it be better to copy the parent and then add child entries to avoid inserts?

PR #6953, opened 2026-08-09
Spawn Tiles menu improvements

eoineoineoin:
As requested https://github.com/space-wizards/RobustToolbox/pull/6932#issuecomment-5209360609, display tooltips for item IDs in the spawn tile menu. ItemList.Item had a `TooltipText` and `TooltipEnabled, but these weren't actually implemented, so I've done that. Because of the way ItemList doesn't create child Controls per-item, these don't behave _exactly_ like regular tooltips, but hopefully close enough.

I've also animated the tiles variants, so users can see differences between them.

https://github.com/user-attachments/assets/73eccb19-7eea-4a0c-b795-093e5929d647


ProPeperos:
its nice, although i preferred when i was able to see the whole sprite stripe, was faster to look up if the sprite has a lot of variation or not
also the tiles name should be improved anyway, having 4 "asteroid sand" is kind of ridiculous

PR #6954, opened 2026-08-09
Revert "Fix sRGB framebuffer in Linux GLES compatibility mode"

metalgearsloth:
Reverts space-wizards/RobustToolbox#6846


metalgearsloth:
Broke more than it fixed.

PR #6955, opened 2026-08-09
Add physics queries to entitylookup

metalgearsloth:
Sanity. Right now they're on sharedphysicssystem but a lot of content just uses entitylookup and then immediately wants fixture data. This just makes it actually sane and not bleed performance like crazy.

Added callback versions because the existing lookup should get it + used struct args because they are based and I should've used them in the first place to minimise API breakage.

### DIFF
--- a/Robust.Shared/GameObjects/Systems/EntityLookupSystem.Physics.cs
+++ b/Robust.Shared/GameObjects/Systems/EntityLookupSystem.Physics.cs
@@ -1,0 +1,406 @@
+using System.Collections.Generic;
+using Robust.Shared.Map;
+using Robust.Shared.Map.Components;
+using Robust.Shared.Maths;
+using Robust.Shared.Physics;
+using Robust.Shared.Physics.Collision;
+using Robust.Shared.Physics.Collision.Shapes;
+using Robust.Shared.Physics.Dynamics;
+using Robust.Shared.Physics.Shapes;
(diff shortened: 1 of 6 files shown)


Robust.Shared/GameObjects/Systems/EntityLookupSystem.Physics.cs
```diff
@@ -0,0 +1,406 @@
...
+    public bool Approximate { get; init; }
+    public bool IgnoreShapeSkin { get; init; }
+
+    public FixtureQueryArgs(QueryFilter filter, bool approximate = false, bool ignoreShapeSkin = false)
+    {
+        Filter = filter;
+        Approximate = approximate;
+        IgnoreShapeSkin = ignoreShapeSkin;
+    }
+}
```
DrSmugleaf: Should this not be readonly? Or do you want to make it possible to add mutable fields later
Also you should be able to replace it with an implicit constructor, you can still init set the properties using curly braces

DrSmugleaf (changes requested):

Robust.Shared/GameObjects/Systems/EntityLookupSystem.Physics.cs
metalgearsloth: Wasn't sure if callers would want to re-use / mutate it but can easily make it readonly for current use cases.

PR #6956, opened 2026-08-09
MapLoaderSystem compression support, new commands for saving/loading full game state

### LINKED ISSUE #2406: Full server snapshot/reload

Roudenn:
## Overview
Added support for ZStd compression for files saved and loaded by MapLoaderSystem, using the new file extension: `.rtsave`.
Added `savegame` and `loadgame` commands that can save and load all entities to a save file, by calling an already existing method in the maploader system, and `convertsavefile` that converts YAML files into RTSAVE and backwards, which is useful for debugging.

On more technical changes, MapLoaderSystem and EntitySerializer now have a little more logging. `BeforeEntityReadEvent` now also includes an (expected) file category, so Content can apply conditional migrations depending on the fily type.

## Why
Fixes #2406 since now save files are very cheap memory-wise and have official support, starting from here content can eventually implement full persistence

ZStd compression can provide up to 97% file size reduction, which can be used for packaged server builds on Content, or even in the repository itself if you're crazy and don't care about readability.
Also it's useful for servers that support persistent rounds since SS14's serializers are sometimes not efficient for post map-init entities, so file compression saves a lot of disk space and makes them practically free.

## Test plan

New file format support should be covered by MapLoaderTest

New commands can be easily tested in-game by just running them on a devmap

## Media
Convert command works correctly in both ways, and the file is compressed correctly:

image

image

Demonstration of the new commands:

https://github.com/user-attachments/assets/0c08882a-2737-4549-9907-9935edc86163

## Breaking Changes

- `MapLoaderSystem` now supports compression, all files with the new `.rtsave` extension will be compressed automatically using ZStd. The level of compression can be changed with the new `MapSavesCompressLevel` CVar.
- `BeforeEntityReadEvent` now has an `ExpectedCategory` field that specifies the category of this save.
- Added `savegame` and `loadgame` commands that save and load the full game state.


Robust.Shared/GameSaves/GameSavesSystem.cs
```diff
@@ -0,0 +1,157 @@
...
+            BitConverter.TryWriteBytes(buf.AsSpan(0, 4), uncompressed.Count);
+
+            // Compress the data
+            var compressedLength = _zstdContext.Compress2(
+                buf.AsSpan(4, bound),
+                uncompressed.AsSpan());
+
+            var filePath = Path.Combine(_resourceManager.UserData.RootDir, path.Filename + Extension);
+
+            using var fileStream = new FileStream(filePath, FileMode.Create, FileAccess.Write);
```
metalgearsloth: This is a sandbox break as it has no validation on inputs and accepts arbitrary path.combine.

Robust.Shared/GameSaves/GameSavesSystem.cs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /// <summary>
+    /// Compresses a YAML data node using ZSTD compression.
+    /// </summary>
+    /// <param name="path">Path to a file.</param>
+    /// <param name="data">Mapping data node to compress in the specified path.</param>
+    private void WriteCompressedZstd(ResPath path, MappingDataNode data)
+    {
+        var uncompressedStream = new MemoryStream();
```
metalgearsloth: Never disposed?

metalgearsloth (changes requested):
Needs tests please.

Also not really sure why compression isn't rolled into existing maploader instead of a bespoke system.

Roudenn:
> Also not really sure why compression isn't rolled into existing maploader instead of a bespoke system.

Compression of pre-init maps is useless since while uncompressed they're humanly readable on GitHub, which is extremely important to review small and medium map changes to ensure that someone didn't add a debug gibstick in a secret stash somewhere. I think map compression will only be viable after a map editor is added, then all map changes can be viewed from the program conveniently

The `GameSavesSystem` does a different thing compared to `MapLoaderSystem`, it manages the post map-init saves of *the entire state* of the game, while MapLoader works mostly with grids and maps.
Content will use game saves in an entirely different context - after startup and before shutdown of the server, or during the game to save and load the full state, so I think splitting it into a different system is justified

Also this system may include more functionality later, such as Autosaves or automatic save loading on startup and shutdown, which won't fit too well into `MapLoaderSystem`

Robust.Shared/EntitySerialization/Systems/MapLoaderSystem.cs
```diff
@@ -57,21 +78,79 @@ private StreamWriter GetWriterForPath(ResPath path)
...
 
-    private void Write(ResPath path, MappingDataNode data)
+    /// <summary>
+    /// Compresses a YAML data node using ZSTD compression.
+    /// </summary>
+    /// <param name="path">Path to a file.</param>
+    /// <param name="data">Mapping data node to compress into the specified path.</param>
+    private void WriteCompressedZstd(ResPath path, MappingDataNode data)
+    {
+        using var uncompressedStream = new MemoryStream(Encoding.UTF8.GetBytes(data.ToString()));
```
Roudenn, line 108: I've tried to do that with a IRobustSerializer, but it seems to return a prefix with an arbitrary byte length depending on how large the file is, so I've resorted to using Encoding.UTF8.GetBytes instead

Robust.Shared/EntitySerialization/Systems/MapLoaderSystem.cs
```diff
@@ -57,21 +78,79 @@ private StreamWriter GetWriterForPath(ResPath path)
...
+
+        if (!uncompressedStream.TryGetBuffer(out var uncompressed))
+        {
+            uncompressed = new ArraySegment<byte>(uncompressedStream.ToArray());
+        }
```
Roudenn, line 113: When I debugged this it looks like this never fails, is this fail-safe redundant here?

Robust.Shared/EntitySerialization/Systems/MapLoaderSystem.cs
```diff
@@ -57,21 +78,79 @@ private StreamWriter GetWriterForPath(ResPath path)
...
-        Write(writer, data);
+        if (path.Extension == SaveExtension)
+        {
+            WriteCompressedZstd(path, data);
+        }
+        else
+        {
+            using var writer = _resourceManager.UserData.OpenWriteText(path);
+            WriteYaml(writer, data);
+        }
```
Roudenn: Is it okay to run compression when a special file format is specified? Maybe it should be a boolean passed as a parameter everywhere instead

Roudenn (commented):
Technically the full entity serialization should be incompatable with adminbus prototype loading, so I wonder if having adminbus should be controlled by a CVar that if enabled disallows full entity serialization

Roudenn:
Reworked the PR entirely to work around maploader, now uses safe methods for file reading and writing, and also included a test to confirm that the new file format works as intended

PR #6957, opened 2026-08-09
Add netserializer aware delta states

### LINKED ISSUE #5117: Transforms should use delta states
99% of the time it's a localpos / localrot update so you can cut down on what gets sent + what the client needs to do a lot.

### LINKED ISSUE #6860: Serializer should be delta state aware
ATM we use discrete states for autonetworking but ideally we just have the 1 delta state that the serializer just handles so we can get easy deltas.

metalgearsloth:
Resolves https://github.com/space-wizards/RobustToolbox/issues/5117
Resolves https://github.com/space-wizards/RobustToolbox/issues/6860

Does NOT resolve the 1-frame lerp mispredict on transform parent change (walking off or on a grid) but I think I'm the only one who notices.

Not sourcegen because moving netserializer to sourcegen was an experience in benchmarking and would be better to do it globally (and in some cases dynamic typing is still required), so you get IL copypasta for now.

Was going to make deltas the default but will split that out to make it less unmergeable.

Transform and physics are special-cased due to prediction but content should be able to get away with the componentnetworkgenerator 99% of the time and not have to worry about this.


Robust.Shared/Serialization/NetComponentDeltaStateSerializer.cs
```diff
@@ -0,0 +1,164 @@
...
+namespace Robust.Shared.Serialization;
+
+/// <summary>
+/// Generated delta states store ONE <see cref="IAutoGeneratedComponentDeltaState.ChangedFields"/> bitmask and ONE
+/// field per networked component field. This serializer writes the mask first, then only writes fields whose bits are
+/// set. This allows a single generated delta state to represent any dirty-field combination without serializing the
+/// unchanged fields.
+/// If you misuse this your networked states will probably explode!
+/// </summary>
+internal sealed class NetComponentDeltaStateSerializer : IDynamicTypeSerializer
```
metalgearsloth, line 20: It's actually not THAT bad because we leave unsafe.skipinits performance on the table (so all the data gets defaulted) but it is possible to footgun yourself with silently getting invalid data.

Robust.Shared/Serialization/NetComponentDeltaStateSerializer.cs
```diff
@@ -0,0 +1,164 @@
...
+            il.Emit(OpCodes.Ldarg_0);
+            il.Emit(OpCodes.Ldarg_1);
+            il.Emit(OpCodes.Ldarg_2);
+            il.Emit(OpCodes.Ldfld, field);
+            il.Emit(OpCodes.Call, writeField);
+
+            il.MarkLabel(endField);
+        }
+
+        il.Emit(OpCodes.Ret);
```
Tayrtahn, line 77: Oof. Could you throw some comments in here explaining what these do? Maybe even a C# translation?

metalgearsloth: yesssir, in rider it gives detailed descriptions for the opcodes but it's just args + method calls.

PR #6962, opened 2026-08-10
Fix LayerSetShader to accept null shader input

SlamBamActionman:
When rendering, all "regular" sprites are rendered with the layer's shader property set to `null`. 

If a sprite gets a shader set, you can revert it to `null` via the method:
`LayerSetShader(int layer, ShaderInstance? shader, string? prototype = null)`.

For some reason the `object layerKey` version of this method did not allow `null`, so this PR fixes that.

---

I left
`LayerSetShader(int layer, string shaderName)`
`LayerSetShader(object layerKey, string shaderName)`
the same since making the string nullable would have them effectively identical to the other two.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -39,7 +39,7 @@ END TEMPLATE-->
 
 ### New features
 
-*None yet*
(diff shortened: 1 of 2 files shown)


metalgearsloth (changes requested):
Release note missing.

PR #6967, opened 2026-08-11
SharedBroadphaseSystem.Execute: early return with invalid map.

whatston3:
A slightly different tack on #6830.  Early return in SharedBroadphaseSystem.Execute if MapUid is null.

Suppose your proxy has a valid entity, but an invalid MapUid.  You get a value of EntityUid.Invalid as a result.

You then call SharedMapSystem.FindGridsIntersecting, which gives you nothing (null map, no GridTreeComponent!).

After this, you call FindPairs.  You get the TransformComponent of your proxy's entity, all good.  You get your current broadphase, works fine.  You check the broadphase's owner, it's not EntityUid.Invalid - okay, then you get your world matrix, and trigger a debug assertion.

Not great!  Should FindPairs do anything in nullspace?  I would assume not.  If not, why continue if your map is null?


Robust.Shared/Physics/Systems/SharedBroadphaseSystem.cs
```diff
@@ -632,10 +632,12 @@ private record struct BroadphaseContactJob() : IParallelRobustJob
...
                 var proxy = MoveBuffer[index];
-                var broadphaseUid = XformQuery.GetComponent(proxy.Entity).Broadphase?.Uid;
-                var worldAABB = TransformSys.GetWorldMatrix(broadphaseUid!.Value).TransformBox(proxy.AABB);
+                var proxyXform = XformQuery.GetComponent(proxy.Entity);
+                if (proxyXform.MapUid is not { } mapUid)
+                    return;
 
-                var mapUid = XformQuery.GetComponent(proxy.Entity).MapUid ?? EntityUid.Invalid;
+                var broadphaseUid = proxyXform.Broadphase?.Uid;
+                var worldAABB = TransformSys.GetWorldMatrix(broadphaseUid!.Value).TransformBox(proxy.AABB);
```
metalgearsloth, line 640: Is it client or server because either way we're bandaiding nullspace movebuffer entities.

whatston3: Client.

[I also don't quite know how to get at the tumor that's under the bandaid.]

metalgearsloth (commented):
I added https://github.com/space-wizards/RobustToolbox/pull/6980

If possible I'd prefer some log on release outside of the parallel context if we can or something else that will yell, as long as we can avoid any additional trycomps.

whatston3:
> I added #6980

I'll try running that on the wallmount branch and see what happens.

> If possible I'd prefer some log on release outside of the parallel context if we can or something else that will yell, as long as we can avoid any additional trycomps.

Add a string (array?) to the job, set it on failure, then check if it's null or not after `_parallel.ProcessNow` returns?  I'm not quite up to your speed.

metalgearsloth:
Right now the exception is making it all the way to the loop where it's silently being swallowed when ideally the code blows up well before that point, preferably as close to the source as possible, which would be when the caller is mutating transforms on PVS detachment.

PR #6969, opened 2026-08-12
Kill prototypeid serializers

deltanedas:
- removed every PrototypeIdXSerializer except for Abstract...Array one since all inheriting prototypes currently have to use it
- removed ValidatePrototypeId nothing uses it anymore

any forks that still use them will have to replace them with ProtoId/EntProtoId until it compiles. RIP mining station

https://github.com/user-attachments/assets/294e019b-5aa8-4184-bf0b-0c836084dd2b


deltanedas:
holy sniped

PR #6977, opened 2026-08-13
Optimize directed EventBus component dispatch

metalgearsloth:
Use array lookup over trygetcomponents.

Have some works in the pipe but this one at least gives a loud debug "DO NOT MUTATE ENTITY WHILE EVENTBUS RUNNING" assert.

Master:
image

Branch:
image


Robust.Shared/GameObjects/EntityEventBus.Directed.cs
```diff
@@ -553,25 +568,85 @@ private void EntDispatch(EntityUid euid, Type eventType, ref Unit args)
...
+                }
+                finally
+                {
+#if DEBUG
+                    eventTable.DirectedDispatchDepth--;
+#endif
+                }
+
+                DebugTools.AssertEqual(eventTable.Version, version);
+                return;
```
Princess-Cheeseballs: Should have this scream a reason as well so people don't get confused as to what the assert is when they trigger it.

Princess-Cheeseballs (changes requested):

Robust.Shared/GameObjects/EntityEventBus.Directed.cs
metalgearsloth: Going to remove it for now as solutions + containers REALLY want it and I don't want to scopecreep with commandbuffers.

Robust.Shared/GameObjects/EntityEventBus.Directed.cs
```diff
@@ -381,6 +382,17 @@ public void OnComponentRemoved(in RemovedComponentEventArgs e)
...
+        internal void ComponentSetMutating(EntityUid euid)
+        {
+            if (!_entEventTables.TryGetValue(euid, out var eventTable))
+                return;
+
+#if DEBUG
+            DebugTools.Assert(eventTable.DirectedDispatchDepth == 0);
+#endif
+            eventTable.Version++;
+        }
```
ElectroJr: Why not just increment the version in `EntityEventBus.EntAddComponent()` and `EntRemoveComponent()` instead of having a separate method that has to be called explicitly? It also avoids the extra dict lookups.

Robust.Shared/GameObjects/EntityEventBus.Directed.cs
ElectroJr: Not sure how commandbuffers will work, but at least currently there's no expectation that raising events won't have side effects like adding/removing components and a few parts of the game either rely on being able to do that or are just very likely to inadvertently trigger these asserts. So removing the asserts for now is probably best.

Robust.Shared/GameObjects/EntityEventBus.Directed.cs
metalgearsloth: Can probably do that, I had much more destructive changes down the line but made it easier to review by splitting.

PR #6986, opened 2026-08-14
Fix StyleBoxTexture using negative bounds UIBox2

metalgearsloth:
Admin menu calls first draw with 0,0 bounds so the margin makes it negative; we just clamp it.


whatston3 (commented):
Looks good.

PR #6988, opened 2026-08-14
Make Audio entitites saveable

Roudenn:
Title, not exactly sure why they were marked as unsaveable in the first place, please tell me if you now the reason
Required for https://github.com/space-wizards/space-station-14/pull/45293

### DIFF
--- a/Resources/EnginePrototypes/Audio/audio_entities.yml
+++ b/Resources/EnginePrototypes/Audio/audio_entities.yml
@@ -2,8 +2,7 @@
   id: Audio
   name: Audio
   description: Audio entity used by engine
-  save: false # TODO PERSISTENCE what about looping or long sounds?
   categories: [ HideSpawnMenu ]
   components:
-    - type: Transform
-      gridTraversal: false
+  - type: Transform
+    gridTraversal: false


metalgearsloth:
> Title, not exactly sure why they were marked as unsaveable in the first place, please tell me if you now the reason

Many 1-shot audio entities should not be saved, e.g. a door opening sound is definitely in that category, or a welding sound, but jukebox probably should.

Roudenn:
> Many 1-shot audio entities should not be saved, e.g. a door opening sound is definitely in that category, or a welding sound, but jukebox probably should.

Don't really see an issue with saving short sounds, it doesn't bloat the save file by any noticeable margin (especially compared to other problems with EntitySerializer that cause it to make post-init save files 2 times bigger than they should be, DestructibleComponent is the most noticeable example right now).

And cancelling the saving of specifically short sounds creates many issues with code that stores short audio entities
Also the code would beed to somehow define what a "short" sound is, which is a whole other deal

So far from my testing, just saving all entities for a full game state is much easier than manually excluding "unnecessary" entities from a save on a technical level

metalgearsloth:
> Don't really see an issue with saving short sounds

You don't see an issue if a mapper misclicks potentially thousands of audio entities are saved?

Roudenn:
> You don't see an issue if a mapper misclicks potentially thousands of audio entities are saved?

Thank for clarifying the problem, I'll make some sort of a system to remove specific entities from a file while saving it sometime later (BeforeEntityWriteEvent or maybe an entire migration system for the engine)

Issue #6739, opened 2026-07-06
TransformSystem.AnchorEntity doesn't work on uninitialized entities.


whatston3:
TransformSystem.AnchorEntity() can result in errors when used on uninitialized entities.  If the recommendation for using the obsolete TransformComponent.Anchored.Set is to use TransformSystem.AnchorEntity, it really should be a drop-in replacement.

I attempted this in https://github.com/space-wizards/space-station-14/pull/44494, the following test failure (run on 430714c in that branch) was caused by this drop-in change: https://github.com/space-wizards/space-station-14/actions/runs/28694740963/job/85102329507

Anchored.Set acknowledges it as-is (see L377 in the snippet below), and Unanchor respects the lifecycle, why isn't this done inside of AnchorEntity?
https://github.com/space-wizards/RobustToolbox/blob/a1d3d16e43f8f38ff1f116988db0995f831535cf/Robust.Shared/GameObjects/Components/Transform/TransformComponent.cs#L367-L391

PR #6990, opened 2026-08-14
Anchor API sanity

metalgearsloth:
- Removes the property
- Anchoring now works as expected so anchoring twice doesn't silently corrupt.
- Kills ReAnchorEvent because any perf gains are not worth the bugs.

PR from the other repo.

### DIFF
--- a/Robust.Shared.IntegrationTests/GameObjects/Systems/AnchoredSystemTests.cs
+++ b/Robust.Shared.IntegrationTests/GameObjects/Systems/AnchoredSystemTests.cs
@@ -1,3 +1,5 @@
+using System.Collections.Generic;
+using System.Linq;
 using System.Numerics;
 using NUnit.Framework;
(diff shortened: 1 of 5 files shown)


whatston3:
Does this resolve https://github.com/space-wizards/RobustToolbox/issues/6739?

Robust.Shared.IntegrationTests/GameObjects/Systems/AnchoredSystemTests.cs
Tayrtahn, line 1: I'm excited to see these warnings finally go away!

DrSmugleaf (commented):
I'm on my (half broken) phone so checking would take a while but does this add a VV path for anchored? Or make setting the field directly work
A lot of admins toggle anchored with VV so removing that for them would be very bad
Could just keep the setter around as a private setter maybe

PR #6995, opened 2026-08-14
Wake cross-grid joints when grids move

### LINKED ISSUE #6381: Cross-grid joints do not wake their objects when grids move
Title! This is an issue we've faced while addressing reviews for space-wizards/space-station-14#42409.

We're currently unsure if it's specifically cross-grid joints attached to relays that're the issue, or if it's all cross-grid joints that're affected. But either way, this can lead to odd behavior like this:

https://github.com/user-attachments/assets/d3d30c96-935b-4634-b791-67e7b106c87f

When a grid moves, joints that affect objects outside of that grid remain asleep, which is a fairly noticeable bug in the physics simulation.

We've been able to resolve the bug through a hack that forces both objects to wake when they're both on different grids, but this is non-ideal for likely obvious reasons.

rhailrake:
Fixes #6381.

Bodies connected by an enabled joint across different grids are now prevented from sleeping. This is updated when joints are enabled, disabled, or removed, when relays change, and when either body moves between grids.


metalgearsloth (changes requested):
This should be handled with disabling sleeping for these entities rather than adding a large overhead.

rhailrake:
> This should be handled with disabling sleeping for these entities rather than adding a large overhead.

Makes sense. I've changed this to disable sleeping for these entities instead.

PR #7000, opened 2026-08-15
Entity Relations

### LINKED ISSUE #6152: We need WeakEntityReference or another solution to solve PVS error spam

Roudenn:
Added Entity Relations - a generic system to automatically clear references to entities that are becoming invalid.
Fixes #6152
Related content PR (not required for merge): https://github.com/space-wizards/space-station-14/pull/45411
Related Docs PR: https://github.com/space-wizards/docs/pull/670

Ready for review!

### Main usage loop
1. `EntityRelation` struct is a wrapper around `EntityUid?` - to write to it you have to use an `EntityManager` method (or an EntitySystem proxy). The entity inside if not null is always guaranteed to have at least a single link with another entity.
2. When you write an entity into a relation, an `EntityRelationComponent` gets added to both entities, which contains `List<EntityRelation>` to entities that relate to it - that way a link is formed between 2 entities using a single universal marker component.
3. When an entity with `EntityRelationComponent` gets deleted, it raises an event `EntityRelationDeleteEvent` for all its `EntityRelation`s to handle
4. Systems can handle the event manually, or component can autogenerate a system using `[AutoGenerateEntityRelations]` and `[AutoRelationField]` on a `EntityRelation` (or a list/set/dictionary) to set it to null (or remove from the collection).

### Component shutdown
`[AutoGenerateEntityRelations]` generates a `ComponentShutdown` event by default for an assignmed component. This can be disabled by typing `shutdownEvent: false` as a parameter into the attribute, then any system can manually handle the event for the shutdown.
All components have a static method called `ClearComponentRelations` - it can be called at the end of the shutdown subscription to clear all relations in other components to this entity manually. Alternatively you can simply delete the related entity (for example stop an audio) to remove the relaton automatically.

When `EntityRelationsComponent` is itself shutting down while the entity isn't terminating, it will call a `EntityRelationShutdownEvent` on the entity to clear all of its relations and set them to null values. That is another safety feature that guarantees that the EntityRelation references are always valid even in such weird scenarios.

### Manual handling
To handle an `EntityRelation` manually, for example if you use some non-default data structure to store it, or to do something else when removing the relations, you have to subscribe to these events and do the actions manually:
- `EntityRelationDeleteEvent` - compare all relations in the component to the removed relation to replace it with an `EntityRelation.Null` value if found
- `EntityRelationShutdownEvent` to set **all** EntityRelations to null in the component
- `ComponentShutdown` to call the `ClearRelation` function on each relation.

### Networking
By default all data in `EntityRelationsComponent` is fully networked, but you can use the `dirty` bool parameter in EntityManager's Relation methods to do something with relations without sending the changes right away. This is useful for manual networking setups where you want to manually send the data only to a specific client, or for performance where you have a lot of relation interactions happening in a row and you want to network the component only at the very end.

### Other parameters in `EntityRelationsComponent`
- **`RemoveOnEmpty`** removes the component if the list of relations becomes empty. It's true by default to retain consistency and minimize changes in entities caused by supporting relations, since an entity that never had any relations will look like an entity that had all relations removed from it. You might want to disable this if your code does a lot of relation operations with a risk of accidentally removing all relations from the list, causing the component to be constantly removed and added in the process.

### Supported types

The autogenerator for relations currently supports the following types:
- `EntityRelation`
- `EntityRelation?`
- `List<EntityRelation>`
- `HashSet<EntityRelation>`
- `Dictionary<EntityRelation, T>`
- `Dictionary<T, EntityRelation>`

The reason why I didn't add more is because `ComponentNetworkGenerator` is not flexible at all and is almost entirely hardcoded, so adding more types while making them supported by `[AutoGenerateComponentState]` is impossible.

In general a helper that recursively goes through all fields in a data type to be able to automatically generate code for all fields of a certain type would be super useful for auto-generators and would allow for support of **any** struct, class, or a collection.

## Technical details

`EntityRelation` struct was added to `ComponentNetworkGenerator`, `EntitySerializer` and `EntityDeserializer`, where it's automatically converted into `NetEntity?` and `EntityUid?` respectively.

`EntityRelationsGenerator` automatically generates systems for components with [AutoGenerateEntityRelations], handling `EntityRelationDeleteEvent` for that component. Fields with `[AutoRelationField]` get set to null when a corresponding relation is found in them.

`EntityManager` now has public methods to manipulate relations, the most important one being `SetRelation`. It is used for linking an entity to another owner entity, and storing the relation in a field inside of the owner's component.
`ClearRelation` can be used to remove a specific link between 2 entities, and if there are no relations anymore in one of them, the component is automatically removed (for consistency so that when there's a component, it means that the entity is linked to other entities)
Other API methods for relations are Network methods for converting collections of `EntityRelation` into `NetEntity`.


Robust.Serialization.Generator/EntityRelationsGenerator.cs
```diff
@@ -0,0 +1,306 @@
...
+                        {
+                            collection = true;
+                        }
+                        else
+                        {
+                            invalid = true;
+                        }
+                    }
+
+                    // If any pause field has [AutoNetworkedField], automatically mark it to dirty on unpause.
```
Tayrtahn: That copy-paste left a little residue. 🙂

Robust.Serialization.Generator/EntityRelationsGenerator.cs
```diff
@@ -0,0 +1,334 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Robust.Roslyn.Shared;
+using Robust.Roslyn.Shared.Helpers;
+
+namespace Robust.Serialization.Generator;
+
+/// <summary>
+/// Automatically generates implementations for handling timer unpausing.
+/// </summary>
```
Tayrtahn, line 13: Here too

Robust.Serialization.Generator/EntityRelationsGenerator.cs
```diff
@@ -0,0 +1,334 @@
...
+public sealed class EntityRelationsGenerator : IIncrementalGenerator
+{
+    private const string AutoGenerateComponentPauseAttributeName = "Robust.Shared.Analyzers.AutoGenerateEntityRelationsAttribute";
+    private const string AutoPausedFieldAttributeName = "Robust.Shared.Analyzers.AutoRelationFieldAttribute";
+    private const string AutoNetworkFieldAttributeName = "Robust.Shared.Analyzers.AutoNetworkedFieldAttribute";
+    // ReSharper disable once InconsistentNaming
+    private const string IComponentTypeName = "Robust.Shared.GameObjects.IComponent";
+
+    private static readonly DiagnosticDescriptor NotComponentDiagnostic = new(
+        Diagnostics.IdComponentPauseNotComponent,
```
Tayrtahn: Add new diagnostic IDs for these instead of piggybacking on the autopause ones.

Robust.Serialization.Generator/EntityRelationsGenerator.cs
```diff
@@ -0,0 +1,334 @@
...
+                    dirty = dirtyBool;
+
+                var shutdownSub = false;
+                if (syntaxContext.Attributes[0].ConstructorArguments[1].Value is bool shutdownSubBool)
+                    shutdownSub = shutdownSubBool;
+
+                var fieldBuilder = ImmutableArray.CreateBuilder<FieldInfo>();
+                foreach (var member in symbol.GetMembers())
+                {
+                    if (!AttributeHelper.HasAttribute(member, AutoPausedFieldAttributeName, out var _))
```
Tayrtahn: Wrong attribute

Robust.Serialization.Generator/EntityRelationsGenerator.cs
```diff
@@ -0,0 +1,334 @@
...
+        PartialTypeInfo PartialTypeInfo,
+        EquatableArray<FieldInfo> Fields,
+        bool Dirty,
+        bool ShutdownEvent,
+        bool NotComponent,
+        Location Location);
+
+    public sealed record FieldInfo(string Name, bool Nullable, bool Invalid, bool DictionaryKey, bool DictionaryValue, bool Collection, Location Location);
+
+    public sealed record AllFieldInfo(string Name, string ParentDisplayName, Location Location);
```
Tayrtahn: I don't have a lot of experience with generators, but since `Location` is a reference type, I believe passing around these types breaks caching for the incremental generator, which could be a major performance issue. Is it possible to remove these `Location` fields?

Robust.Serialization.Generator/EntityRelationsGenerator.cs
```diff
@@ -0,0 +1,334 @@
...
+                    relationBuilder.AppendLine($"        ent.Comp.{field.Name}.Remove(args.Relation);");
+
+                    shutdownBuilder.AppendLine($"        entMan.ClearRelation(ent.Owner, ent.Comp.{field.Name});");
+                }
+                else if (field.DictionaryValue)
+                {
+                    relationBuilder.AppendLine($$"""
+                                foreach (var (key, value) in ent.Comp.{{field.Name}})
+                                {
+                                    if (ent.Comp.{{field.Name}}[key] == args.Relation)
```
Tayrtahn: Isn't `ent.Comp.{field.Name}[key]` the same as `value`?

Robust.Shared/Analyzers/EntityRelationsGeneratorAttributes.cs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/// <summary>
+///     Mark a field or property to automatically handle setting an entity relation to null when the related entity is deleted.
+/// </summary>
+/// <remarks>
+///     The type of the field must be <see cref="EntityRelation"/>,
+///     an <see cref="IEnumerable{T}"/> listing EntityRelations,
+///     or a Dictionary with EntityRelation as a key or a value.
+///     For all other use cases handle the
```
Tayrtahn: Unfinished comment

Robust.Shared/Analyzers/EntityRelationsGeneratorAttributes.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <summary>
+    ///     Whether the generated code should automatically call
+    ///     <see cref="IEntityManager.Dirty(EntityUid,IComponent,MetaDataComponent)"/> after resetting the related entity.
+    ///     This is automatically inferred for fields marked <see cref="AutoNetworkedFieldAttribute"/>.
+    /// </summary>
+    public readonly bool Dirty;
+
+    /// <summary>
+    ///     Whenever the generated code should subscribe to the <see cref="ComponentShutdown"/> event.
+    ///     in order to clear the related links.
```
Tayrtahn:
```suggestion
    ///     Whether the generated code should subscribe to the <see cref="ComponentShutdown"/> event
    ///     in order to clear the related links.
```

Robust.Shared/GameObjects/EntityManager.Relations.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        var relationComp = _relationsQuery.Comp(relation.Entity.Value);
+
+        ent.Comp.Relations.Remove(relation);
+        if (ent.Comp.Relations.Count == 0 && ent.Comp.LifeStage < ComponentLifeStage.Stopping)
+            RemoveComponent(ent.Owner, ent.Comp);
+
+        relationComp.Relations.Remove(new EntityRelation(ent.Owner));
+        if (relationComp.Relations.Count == 0 && relationComp.LifeStage < ComponentLifeStage.Stopping)
+            RemoveComponent(relation.Entity.Value, relationComp);
```
Tayrtahn: Maybe make a helper method to handle this "remove an item and if the list is empty, remove the component" logic.

Tayrtahn (commented):
This definitely needs some integration tests to show everything works as expected! 

Here are a couple of snags that I ran into when I tried to do something like this myself:
- Imagine entity A has a component containing a relation to entity B. If entity A removes just that component (without the entity terminating), what happens? If this requires subscribing to the shutdown event for the component, does this prevent other systems (like the main one for the component) from subscribing to shutdown?
- Ideally we would have a way to make networking optional or even player-specific. Otherwise a modified client will be able to detect any entities with relations to a given entity - for example, a homing missile tracking the local player. This is very tricky to balance with the desire for prediction, so I recommend further discussion before trying to implement anything here.

Roudenn:
> Here are a couple of snags that I ran into when I tried to do something like this myself:
> 
> * Imagine entity A has a component containing a relation to entity B. If entity A removes just that component (without the entity terminating), what happens? If this requires subscribing to the shutdown event for the component, does this prevent other systems (like the main one for the component) from subscribing to shutdown?

This implementation auto-generates a system that subscribes to `ComponentShutdown` automatically, but it can be turned off  by adding `(shutdownEvent: false)` as a constructor parameter to the `[AutoGenerateEntityRelations]` attribute - this should be used in cases where other code has to subscribe to `ComponentShutdown`, and to still be able to clear the relations automatically without much poilerplate, components with `[AutoGenerateEntityRelations]` also have a public method to clear all relations.

I've took that way because Sloth said that adding a new event just to shutdown the relations will make deletion performance even worse, and at the same time clearing the relations on shutdown of a certain component is required to make `EntityRelationsComponent` consistent (so it can be removed from entities that are no longer related by anything), so I've decided to just force all developers to deal with that manually if they encounter this issue and gave them convenient tools to do it. Sad but unless eventbus will somehow have 0 performance cost, handling the component shutdown properly will remain a responsibility of the developers.

> * Ideally we would have a way to make networking optional or even player-specific. Otherwise a modified client will be able to detect any entities with relations to a given entity - for example, a homing missile tracking the local player. This is very tricky to balance with the desire for prediction, so I recommend further discussion before trying to implement anything here.

This is an issue that isn't that urgent compared to the error spams that still appear on live servers and it can be solved later. The only thing that needs to be checked here is whenever the system is flexible enough to still be backwards compatiable when a generic conditional networking system comes out.

Without standartising all conditional networking usecases, adding that system to EntityRelations will be impossible, because even if you make a bare-bones version that supports a couple features, Content may need something more, and then they're still screwed because the bare-bones conditional networking system for EntityRelations isn't developed enough.

For now my opinion is that solving error spams is far more important than caring about possible minor info cheating opportunities (especially when a lot of them already exist in Content still unfixed).

Robust.Serialization.Generator/EntityRelationsGenerator.cs
Roudenn: This field is only used for Diagnostics reports, I'll just move this into a separate code analyzer and it should be fine.

metalgearsloth:
> Imagine entity A has a component containing a relation to entity B. If entity A removes just that component (without the entity terminating), what happens? If this requires subscribing to the shutdown event for the component, does this prevent other systems (like the main one for the component) from subscribing to shutdown?

ComponentRemove is mostly unused / misused but it might cause too much of a blast radius on it. The only component that really needs it is transform and all the others just want ordered startups. If the choice is performance or QOL just because it's component lifecycle related I'm choosing performance.

Roudenn:
Added the base test for EntityRelations functionality

Still working on splitting the diagnostics from the relations codegen and making a codegen test

Tayrtahn:
> This implementation auto-generates a system that subscribes to `ComponentShutdown` automatically, but it can be turned off  by adding `(shutdownEvent: false)` as a constructor parameter to the `[AutoGenerateEntityRelations]` attribute - this should be used in cases where other code has to subscribe to `ComponentShutdown`, and to still be able to clear the relations automatically without much poilerplate, components with `[AutoGenerateEntityRelations]` also have a public method to clear all relations.

Sounds good to me! Maybe we can have an analyzer that detects manual Shutdown handlers and makes sure they call that clear method - I'd be happy to look into that.

Roudenn:
Added a test for the code generator and split its analyzers into a proper `DiagnosticsAnalyzer` class, but I couldn't figure out how to properly make a test for the analyzer unfortunately

Besides the analyzer test, the only other thing left is an integration test for EntityRelation networking

(well technically there's also map saving but it should be trivial enough to just save a post-init map and ensure it gets saved and loaded correctly)

Roudenn:
I solved the issue with `EntityRelationsComponent` networking by adding a bool `dirty` optional parameter to the `SetRelation` method that's true by default, now content can control what data is sent using this parameter and specifying Dirty in a constructor of `AutoGenerateEntityRelations`

Robust.Serialization.Generator/EntityRelationsAnalyzer.cs
```diff
@@ -0,0 +1,145 @@
...
+            case { Name: "Nullable", TypeArguments: [{ Name: "EntityRelation" }] }:
+            case { Name: "List", TypeArguments: [{ Name: "EntityRelation" }] }:
+            case { Name: "HashSet", TypeArguments: [{ Name: "EntityRelation" }] }:
+            case { Name: "Dictionary", TypeArguments: [{ Name: "EntityRelation" }, _] }:
+            case { Name: "Dictionary", TypeArguments: [_, { Name: "EntityRelation" }] }:
+                return;
+        }
+
+        context.ReportDiagnostic(
+            Diagnostic.Create(WrongTypeAttributeDiagnostic, symbol.Locations[0], symbol.Name));
```
Fildrance: error text is about EntityRelation type but we accept either EntityRelation or collection of it.

Robust.Shared/EntitySerialization/EntityDeserializer.cs
```diff
@@ -1294,4 +1295,34 @@ DataNode ITypeWriter<MapId>.Write(
     }
 
     #endregion
+
+    public ValidationNode Validate(
```
Fildrance: new methods are still part of TypeSerializer, so should be inside region above

Fildrance: specify type of interface from which it implements, like other similar methods

Robust.Shared/GameObjects/Components/EntityRelationsComponent.cs
```diff
@@ -0,0 +1,36 @@
...
+    public List<EntityRelation> Relations = new();
+
+    /// <summary>
+    /// If true, will remove the component when <see cref="Relations"/> list becomes empty.
+    /// Useful for conditional networking or performance sensitive EntityRelation interactions.
+    /// </summary>
+    /// <remarks>
+    /// This field is always networked to the client.
+    /// </remarks>
+    [DataField, AutoNetworkedField]
```
Fildrance, line 31: does this need to be networked? really?

Fildrance: And remarks block seems a bit... excessive?

Robust.Shared/GameObjects/Components/EntityRelationsComponent.cs
```diff
@@ -0,0 +1,36 @@
...
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState(fieldDeltas: true)]
+public sealed partial class EntityRelationsComponent : Component
+{
+    /// <summary>
+    /// A list of currently active entity relations.
+    /// </summary>
+    /// <remarks>
+    /// This type is a list since the entity may be related
+    /// by 2 different components on the same owner, and to handle that safely
+    /// the relation is listed multiple times.
```
Fildrance:
```suggestion
    /// Not HashSet because one entity can relate by
    /// 2+ different components on the same owner, and to handle that safely
    /// the relation is listed multiple times.
```

Robust.Shared/EntitySerialization/EntitySerializer.cs
```diff
@@ -1154,5 +1155,35 @@ DataNode ITypeWriter<MapId>.Write(
         return new ValueDataNode("invalid");
     }
 
+    ValidationNode ITypeValidator<EntityRelation, ValueDataNode>.Validate(
```
Fildrance, line 1158: same, mention interface.

Robust.Shared/EntitySerialization/EntitySerializer.cs
```diff
@@ -1154,5 +1155,35 @@ DataNode ITypeWriter<MapId>.Write(
...
+
+    EntityRelation ITypeReader<EntityRelation, ValueDataNode>.Read(
+        ISerializationManager seri,
+        ValueDataNode node,
+        IDependencyCollection deps,
+        SerializationHookContext hookCtx,
+        ISerializationContext? ctx,
+        ISerializationManager.InstantiationDelegate<EntityRelation>? instanceProvider)
+    {
+        return new EntityRelation(seri.Read<EntityUid?>(node, ctx));
```
Fildrance, line 1175: Compared to what ITypeReader<EntityUid, ValueDataNode> does, this looks... quite barebones, tbh.
same goes for EntityDeserializer. Would like @metalgearsloth knowledge to shine upon if thats okay =w=

Robust.Shared/GameObjects/Systems/EntityRelationsSystem.cs
```diff
@@ -0,0 +1,60 @@
...
+    [Dependency] private EntityQuery<EntityRelationsComponent> _relationsQuery = default!;
+
+    /// <summary>
+    /// Sets whether the component should be removed when it's empty or not.
+    /// </summary>
+    /// <param name="ent">The target entity with relations.</param>
+    /// <param name="value">The value to set. If true,
+    /// the component will be removed when this entity isn't related to anything.</param>
+    [PublicAPI]
+    public void SetRemoveOnEmpty(Entity<EntityRelationsComponent?> ent, bool value)
```
Fildrance, line 23: i am not fully grasping how it should be used? Also... if we change it, shouldn't we check if we need to remove it right now?...

Robust.Shared/Analyzers/EntityRelationsGeneratorAttributes.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public readonly bool Dirty;
+
+    /// <summary>
+    ///     Whether the generated code should subscribe to the <see cref="ComponentShutdown"/> event.
+    ///     in order to clear the related links.
+    /// </summary>
+    public readonly bool ShutdownEvent;
+
+    public AutoGenerateEntityRelationsAttribute(bool dirty = false, bool shutdownEvent = true)
```
Fildrance: primary c-tor?

Robust.Shared/Analyzers/EntityRelationsGeneratorAttributes.cs
```diff
@@ -0,0 +1,51 @@
...
+///     (<see cref="EntityUnpausedEvent"/>).
+/// </remarks>
+[AttributeUsage(AttributeTargets.Class, Inherited = false)]
+[BaseTypeRequired(typeof(IComponent))]
+public sealed class AutoGenerateEntityRelationsAttribute : Attribute
+{
+    /// <summary>
+    ///     Whether the generated code should automatically call
+    ///     <see cref="IEntityManager.Dirty(EntityUid,IComponent,MetaDataComponent)"/> after resetting the related entity.
+    ///     This is automatically inferred for fields marked <see cref="AutoNetworkedFieldAttribute"/>.
```
Fildrance: maybe small remark why it can be useful to turn OFF? I guess for server-side components for example? :|

Robust.Shared/Analyzers/EntityRelationsGeneratorAttributes.cs
```diff
@@ -0,0 +1,51 @@
...
+using JetBrains.Annotations;
+using Robust.Shared.GameObjects;
+
+namespace Robust.Shared.Analyzers;
+
+/// <summary>
+///     Indicate that a <see cref="Component"/> should automatically handle setting entity relation fields to null when the related entity is deleted.
+/// </summary>
+/// <remarks>
+///     When this attribute is set on a <see cref="Component"/>, an <see cref="EntitySystem"/> will automatically be
```
Fildrance: copy-paste?

Robust.Shared/Analyzers/EntityRelationsGeneratorAttributes.cs
```diff
@@ -0,0 +1,51 @@
...
+        ShutdownEvent = shutdownEvent;
+    }
+}
+
+/// <summary>
+///     Mark a field or property to automatically handle setting an entity relation to null when the related entity is deleted.
+/// </summary>
+/// <remarks>
+///     The type of the field must be <see cref="EntityRelation"/>,
+///     an <see cref="IEnumerable{T}"/> listing EntityRelations,
```
Fildrance: i might have missed something but we are not supporting IEnumerable, not of ANY derived kind at least - List or Hashset most likely?

Robust.Serialization.Generator/EntityRelationsGenerator.cs
```diff
@@ -0,0 +1,290 @@
...
+                if (field.Nullable)
+                {
+                    relationBuilder.AppendLine($"""
+                                if (ent.Comp.{field.Name}.HasValue && ent.Comp.{field.Name}.Value == args.Relation)
+                                    ent.Comp.{field.Name} = null;
+                        """);
+
+                    clearBuilder.AppendLine($"        ent.Comp.{field.Name} = null;");
+
+                    shutdownBuilder.AppendLine($"        entMan.ClearRelation(ent.Owner, ref ent.Comp.{field.Name}, false);");
```
Fildrance: is it dirty set to false? Why is it not respecting attribute thingy?

Robust.Serialization.Generator/EntityRelationsGenerator.cs
```diff
@@ -0,0 +1,290 @@
...
+                {
+                    relationBuilder.AppendLine($"        ent.Comp.{field.Name}.Remove(args.Relation);");
+
+                    clearBuilder.AppendLine($"        ent.Comp.{field.Name}.Clear();");
+
+                    shutdownBuilder.AppendLine($"        entMan.ClearRelations(ent.Owner, ent.Comp.{field.Name}, false);");
+                }
+                else if (field.DictionaryValue)
+                {
+                    relationBuilder.AppendLine($$"""
```
Fildrance: this seem to handle only case when relation is value... and i think we somewhere declared that we support it at any position?
Also shouldn't we remove value on key, instead of setitng it to empty value?

Robust.Serialization.Generator/EntityRelationsGenerator.cs
```diff
@@ -0,0 +1,290 @@
...
+                """);
+
+            info.PartialTypeInfo.WriteHeader(result);
+
+            result.AppendLine($$"""
+
+                {
+                [RobustAutoGenerated]
+                [global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
+                public sealed partial class {{info.PartialTypeInfo.Name}}_AutoRelationsSystem : EntitySystem
```
Fildrance: i understand that it is quite handy to have multiple specific systems generated, but one system would be much easier to track perf after using our usual tools. for @metalgearsloth to consider ofc...

Robust.Shared/GameObjects/EntityManager.Network.cs
```diff
@@ -384,6 +419,31 @@ public void EnsureEntityList<T>(List<NetEntity> netEntities, EntityUid callerEnt
         }
     }
 
+    /// <inheritdoc />
+    public List<EntityRelation> EnsureEntityListRelation<T>(List<NetEntity> netEntities, EntityUid callerEntity)
```
Fildrance, line 423: can we have it all be overloads (same name different input and output arguments)? would probably make thigns a little bit easier coz you dont strictly need different code generated for calling it then

Robust.Shared/GameObjects/EntityManager.Relations.cs
```diff
@@ -0,0 +1,583 @@
...
+public partial class EntityManager
+{
+    /// <inheritdoc/>
+    public void SetRelation(
+        Entity<EntityRelationsComponent?> owner,
+        ref EntityRelation relation,
+        Entity<EntityRelationsComponent?>? entity,
+        bool dirty = true)
+    {
+        DebugTools.Assert(owner.Owner != entity,
```
Fildrance, line 16: i am not too sure if thats invalid... in general sense some cases can accept any entity, including itself, don't they? Not sure..

Robust.Shared/GameObjects/EntityManager.Relations.cs
```diff
@@ -0,0 +1,583 @@
...
+            $"Entity {ToPrettyString(owner.Owner)} attempted to set an {nameof(EntityRelation)} to itself!");
+
+        if (!entity.HasValue)
+        {
+            relation.Entity = null;
```
Fildrance, line 24: hmm, so this method is not meant to set entity to null? Because it does not call for dirty... do we HAVE to call clear for that?

Robust.Shared/GameObjects/EntityManager.Relations.cs
```diff
@@ -0,0 +1,583 @@
...
+            return;
+        }
+
+        var ent = entity.Value;
+
+        // TODO add EnsureComp methods to EntityQuery
```
Fildrance, line 30: is this for separate PR? then we should not forget to create issue for it upon merging

Robust.Shared/GameObjects/EntityManager.Relations.cs
```diff
@@ -0,0 +1,583 @@
...
+        if (!_relationsQuery.Resolve(ent.Owner, ref ent.Comp, false))
+            EnsureComponent<EntityRelationsComponent>(ent.Owner, out ent.Comp);
+
+        if (!_relationsQuery.Resolve(owner.Owner, ref owner.Comp, false))
+            EnsureComponent<EntityRelationsComponent>(owner.Owner, out owner.Comp);
+
+        if (relation.Entity != null)
+            ClearRelation(owner, ref relation, false);
```
Fildrance: will this lead to not dirty-ing comp of entity that we are removing?

Robust.Shared/GameObjects/EntityManager.Relations.cs
```diff
@@ -0,0 +1,583 @@
...
+                continue;
+
+            DirtyRelations((relation.Entity.Value, entityRelations));
+        }
+
+        DirtyRelations(owner!);
+    }
+
+    /// <inheritdoc/>
+    public void AddRelations(Entity<EntityRelationsComponent?> owner, HashSet<EntityRelation> relations, bool dirty = true, params Entity<EntityRelationsComponent?>?[] entities)
```
Fildrance: line breaks?

Robust.Shared/GameObjects/IEntityManager.Network.cs
```diff
@@ -126,11 +131,21 @@ public bool TryGetNetEntity(EntityUid? uid, [NotNullWhen(true)] out NetEntity? n
     /// </summary>
     public HashSet<NetEntity> GetNetEntitySet(HashSet<EntityUid> entities);
 
+    /// <summary>
+    /// HashSet version of <see cref="GetNetEntity"/>
```
Fildrance, line 139: this does not specify any method specifically, and leaves xml-doc related warning, which is not ideal. samge goes for other added things.

Robust.Shared/GameObjects/IEntityManager.Network.cs
```diff
@@ -61,6 +61,11 @@ public bool TryGetNetEntity(EntityUid? uid, [NotNullWhen(true)] out NetEntity? n
     /// </summary>
     public EntityUid? EnsureEntity<T>(NetEntity? nEntity, EntityUid callerEntity);
 
+    /// <summary>
+    /// Tries to get a corresponding <see cref="EntityRelation"/> if it exists and nEntity is not null.
```
Fildrance, line 65: add that it returns EntityRelation.Null, or it will contradict with method signature a bit

Fildrance (commented):
Was not looking at tests yet - maybe its my fault, and thats why i misunderstood some of stuff. But at least i tried to have full graps of how it works and tried to find problems!
Code looks good and handles a lot of our usual cases, but i left some open questions and remarks - maybe Sloth would look at them and think they are not of concern :D then its fine...
I really need to take a look at tests but there is a lot of code to chew in one go >_>

Robust.Shared/GameObjects/IEntityManager.Network.cs
Roudenn: Funny thing is that it's like that on all network methods for whatever reason :godo:

Robust.Shared/GameObjects/Components/EntityRelationsComponent.cs
Roudenn: > does this need to be networked? really?

Yes because it makes component removal predicted and therefore the same code runs the same both on server and the client

Robust.Shared/EntitySerialization/EntitySerializer.cs
Roudenn: Validation could have some sanity checks to ensure that if the relation stores an entity it also always has `EntityRelationsComponent`, but I'm not sure if it's that important to log during serialization, it's something that should be catched right when it appears

Robust.Shared/GameObjects/Systems/EntityRelationsSystem.cs
Roudenn: This can be set to false in order to prevent the component from being automatically removed, this can be useful for code that has some possibly dangerous EntityRelation interactions that may lead to the list of relations becoming empty, therefore if the component was previously cached in code it will become invalid, which is really bad to do

Robust.Shared/Analyzers/EntityRelationsGeneratorAttributes.cs
Roudenn: It's already turned off by default and gets enabled by `[AutoNetworkedField]`, but you probably want to turn this on manually if you subscribe to component states manually

Robust.Serialization.Generator/EntityRelationsGenerator.cs
Roudenn: Shutdown static auto-generated method is intended to be called during ComponentShutdown, which afaik is already predicted/raised on the client for `[NetworkedComponent]`s.

Someone please re-check this theory, but I think the shutdown method should technically *always* run the same way both on server and client, so dirtying this at all is unnecessary.

Robust.Serialization.Generator/EntityRelationsGenerator.cs
Roudenn: If you want to delete the key-value pair on entity deletion, you just need to set the entity relation as a key

Otherwise values are just values, so removing the key together with the value relation feels like overreacting to me

Robust.Serialization.Generator/EntityRelationsGenerator.cs
Roudenn: Probably a TODO for a future refactor of all code generators at once, currently it's absolutely painful to write them and unifying all system generators into a single one will result in a huge unreadable mess right now

Robust.Shared/GameObjects/EntityManager.Network.cs
Roudenn: It conflicts with the `List<EntityUid>` overload, so no it's not possible

Robust.Shared/GameObjects/EntityManager.Relations.cs
Roudenn: When I think about it that's partially true, sometimes when you make a universal component you may want to store a reference to the entity itself inside the component to do some special behaviour on the entity itself... But on the other hand having a self-relation isn't something that makes sense, because it adds unecessary overhead in some cases

Would like to hear an opinion of an RT maintainer on that

Robust.Shared/GameObjects/EntityManager.Relations.cs
Roudenn: Thanks for spotting an error, setting a relation to a null entity basically means replacing it, therefore a clear relation method always has to be called to handle that

Robust.Shared/GameObjects/EntityManager.Relations.cs
Roudenn: Yes, I'll make an issue on that right now

Roudenn: #7046

Roudenn:
The main issue that remains with this PR is that there are edge cases where when a relation is deleted, instead of simply setting it to null, the entire component or even the entire entity must be removed as well. Basically this means that there should be categories to the entity relation implementation where they have different strength levels (for example `EntityRelationStrength.` `Null`, `RemComp` and `Delete`.
The convenience issue is that `EntityRelation` struct will remain to be nullable regardless of the relation strength and whenever it can ever actually be null, though I think it's not that important.

I will try to implement that feature later, but the PR should still be ready for review as is because there aren't many cases where stronger relations are needed

PR #7007, opened 2026-08-16
Change PrototypeListContainer to VirtualListContainer

metalgearsloth:
1. Having a virtual control thing is good for large lists but calling it prototypelist is weird when it has nothing to do with prototypes.
2. Some fixes I need to make 1,000 entity context menus open in a performant way.


metalgearsloth:
Urgh forgot mappingprotolist

deltanedas:
being called prototype is probably copying qt or similar toolkits

PR #7015, opened 2026-08-17
Fix client predicteddel running on detaching entities

metalgearsloth:
The issue is that the client will restore the entity in a weird state so we just need to block those situations entirely. Once it's detached we should just ignore it entirely.

This should defensively guard against https://github.com/space-wizards/space-station-14/pull/45410 from the engine side.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -43,7 +43,7 @@ END TEMPLATE-->
 
 ### Bugfixes
 
-*None yet*
+* ClientGameStateManager will now drop predicted deletions if they occur while an entity is detached to nullspace due to PVS.
 
 ### Other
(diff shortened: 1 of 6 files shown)


DrSmugleaf (commented):
Would this mess up deleting as it comes back into pvs range? Otherwise unless entity lookup is fixable I guess this works

metalgearsloth:
> Would this mess up deleting as it comes back into pvs range? Otherwise unless entity lookup is fixable I guess this works

Should only affect predicteddel as deletions get run at the very end of applyentitystates separately.

DrSmugleaf (approved):
Should be fine, even if you mispredict a delete the entity is getting detached anyways

PR #7017, opened 2026-08-17
Refactoring Audio Manager

Rinary1:
## Introduction

This PR aims to refactor the Audio Manager, my main idea was in adding compatibility for devices without any active audio device, i.e. instead of getting exception you can play without speakers or headphones, It really annoyed me that I couldn't run the game without an active output device, since my headphones often run out of battery and without them I couldn't even test the in-game changes. But it turned out that doing this correctly is much more difficult; during the refactoring, I also found and tried to fix several critical issues.

## What was actually changed and why

I'll place here a description of the items in as much detail as possible.

1. The first thing I encountered that any operation with AL runs even if we aren't initialized, which means if we don't have any device - we will get an exception, which is wrong, it lies because AL.GenSource() without context silently returns 0, so you're getting "Failed to generate source. Too many simultaneous audio streams?" with white space at the end. So I added `_initialized` flag(but since I was doing this in an older version, I had to rewrite it to use _audioInitialized) and added an initialization check to every public method that performs AL calls
2. Then I found that default OpenTK prefers to use System's OpenAL32.dll instead of package variant, what does that mean? 
It means that we're playing in a gamble, because OpenAL32.dll might be corrupted, the wrong version, or just not right, I noticed this because I wanted to add proper hot-plug support to an older version, but OpenAL wouldn't provide me with a "No Output" device, this also may create unexpected behavior such as problems with Audio(hissing, other artifacts) or just Silence, I've already received complaints about this issue on my fork, I attribute this specifically to that, so I added `PreloadOpenAl`.
3. This PR also adds device state update in FrameUpdate, which should immediately swap manager to newly connected audio device.
4. Next comes a pure refactor, so I won't go into all that detail, I'll just write down what I did - organized all the methods by regions, placed public methods to public part and some private methods into private.
5. Some of already existed audio swap logic was adapted for hot-plug support.
6. Implemented all AttenuationGains
7. Covered system with more tests(since added more things that require testing)
8. Changed `SwitchesAudioDevice` test to test more than just CVars.
9. Fixed ClydeHandle collision, when manager creates stream, it uses buffer.Count as identifier, but if we're removing stream, new stream may use buffer which was used by removed stream.
10. Fixed several other issues found during the refactoring, in short: Shutdown now sets _audioInitialized to false, so we will drop future AL calls, also unregisters OnReload in _reload, fixed/implemented/changed some TODOs.

## What tests were conducted

- Launched the game without an audio output device connected.
[...]

PR #7021, opened 2026-08-18
Fix pause when audio disposed

### LINKED ISSUE #2012: Calling StopPlaying on audio twice shouldn't crash
There's no way to check if a source is disposed from content so if you try to stop a stream twice it crashes.

Rinary1:
Part of #7017, dividing it on small patches, this change just Fixes #2012, by adding Playing check on pause operation and replaces StartPlaying() call with `Playing = true;` so it won't check Playing for the second time.

### DIFF
--- a/Robust.Client/Audio/Sources/BaseAudioSource.cs
+++ b/Robust.Client/Audio/Sources/BaseAudioSource.cs
@@ -47,7 +47,7 @@ public abstract class BaseAudioSource : IAudioSource
     /// <inheritdoc />
     public void Pause()
     {
-        if (_isDisposed())
+        if (!Playing)
             return;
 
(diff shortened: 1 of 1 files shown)


metalgearsloth (changes requested):
Can you elaborate on this one more? All of the calls have _checkDisposed already for playing so if anything restart and pause should get it for consistency.

Rinary1:
> Can you elaborate on this one more? All of the calls have _checkDisposed already for playing so if anything restart and pause should get it for consistency.

I missed dispose check in Playing, pushed fix, now that change only adds playing check in pause method and replaces StartPlaying() call in Restart with Playing=true, so it won't check Playing second time.

PR #7022, opened 2026-08-18
Audio resilent against device init failure

### LINKED ISSUE #2180: Clyde should be resilient against audio device init failure

### LINKED ISSUE #6414: Client crash when no audio devices are available
Client crashes when no audio playback devices are available. Audio engine appears to proceed regardless, initiating a crash upon buffering audio resources.

### Specific cause
`AL.GenBuffer` always returns 0 without a proper audio context. See: [opentk/LearnOpenTK#86](https://redirect.github.com/opentk/LearnOpenTK/issues/86)
https://github.com/space-wizards/RobustToolbox/blob/fe1648d290d05c66c752bc8bca752b18a6014210/Robust.Client/Audio/AudioManager.Public.cs#L105

This causes the audio sample buffer dictionary to have duplicate keys upon loading the next audio resource.
https://github.com/space-wizards/RobustToolbox/blob/fe1648d290d05c66c752bc8bca752b18a6014210/Robust.Client/Audio/AudioManager.Public.cs#L136

---

[client.stdout.log](https://github.com/user-attachments/files/25143908/client.stdout.log) on Lizard.

Rinary1:
Part of #7017, dividing it on small patches, this change just Fixes #2180 and Fixes #6414, by adding init checks in public methods.

However, there is no hot reload, so you'll still need to restart the game to enable audio.

### DIFF
--- a/Robust.Client/Audio/AudioManager.Public.cs
+++ b/Robust.Client/Audio/AudioManager.Public.cs
@@ -69,18 +69,21 @@ internal partial class AudioManager
     /// <inheritdoc/>
     public void SetVelocity(Vector2 velocity)
     {
+        if (!_audioInitialized) return;
         AL.Listener(ALListener3f.Velocity, velocity.X, velocity.Y, 0f);
     }
 
     /// <inheritdoc/>
     public void SetPosition(Vector2 position)
     {
+        if (!_audioInitialized) return;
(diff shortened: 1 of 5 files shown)


metalgearsloth (changes requested):
Not a big fan of just adding initialize checks everywhere, it would be better to just use headlessaudiomanager if audio blows up on game init. Likewise I wouldn't want this handling on clyde.

PR #7024, opened 2026-08-18
Use packaged OpenAL

Rinary1:
Part of #7017, dividing it on small patches, this change forces game to use packaged OpenAL instead of system ones by preloading it, because OpenAL prefers to use system one.

Why?

General Stability and Predictability: I've seen players complain about sound issues more than once; I attribute this to a corrupted OpenAL library installed on their systems, but we can't force everyone to reinstall the library or the app.
Also the same version and the same implementation will always be used, for example when I tried to add hot-plug it used my OpenAL, which was not OpenAL Soft.

### DIFF
--- a/Robust.Client/Audio/AudioManager.cs
+++ b/Robust.Client/Audio/AudioManager.cs
@@ -1,8 +1,10 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Threading;
 using OpenTK.Audio.OpenAL;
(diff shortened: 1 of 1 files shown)


metalgearsloth (changes requested):
1. Anyone with corrupted install needs logging for it and provide exceptions.
2. Needs comments + info log for loading it.
3. It would be better to globally use bundled rather than having more subtle platform inconsistencies, no?

Rinary1:
> 3. It would be better to globally use bundled rather than having more subtle platform inconsistencies, no?

yeah, you're right, but I'm not linux user and don't have macOS, so I can't test it by myself and be sure that it won't break mac or linux. If someone is ready to test this - I can do this

PR #7025, opened 2026-08-18
Fix audio source collision

Rinary1:
Part of #7017, dividing it on small patches, this PR changes buffer registration, to fix collision, also adds check in `CreateAudioSource` to be sure that buffer exists.

### DIFF
--- a/Robust.Client/Audio/AudioManager.Public.cs
+++ b/Robust.Client/Audio/AudioManager.Public.cs
@@ -330,7 +330,7 @@ internal partial class AudioManager
     {
         if (!_audioSampleBuffers.TryGetValue(stream.BufferId, out var sample))
         {
-            OpenALSawmill.Warning($"Audio stream '{stream.Name}' has no backing buffer, skipping.");
+            OpenALSawmill.Error($"Audio stream '{stream.Name}' has no backing buffer, skipping.");
             return null;
         }
 
--- a/Robust.Client/Audio/AudioManager.Public.cs
+++ b/Robust.Client/Audio/AudioManager.Public.cs
@@ -149,9 +149,8 @@ internal partial class AudioManager
 
         _checkAlError();
 
-        var handle = new ClydeHandle(_audioSampleBuffers.Count);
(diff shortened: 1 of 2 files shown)


Robust.Client/Audio/AudioManager.Public.cs
```diff
@@ -339,9 +336,14 @@ internal void RemoveBufferedAudioSource(int handle)
...
-        // TODO: This really shouldn't be indexing based on the ClydeHandle...
-        AL.Source(source, ALSourcei.Buffer, _audioSampleBuffers[stream.BufferId].BufferHandle);
+        if (!_audioSampleBuffers.TryGetValue(stream.BufferId, out var sample))
+        {
+            OpenALSawmill.Warning($"Audio stream '{stream.Name}' has no backing buffer, skipping.");
+            AL.DeleteSource(source);
+            return null;
+        }
+
+        AL.Source(source, ALSourcei.Buffer, sample.BufferHandle);
```
metalgearsloth, line 346: Wouldn't it make more sense to run this before getting a source

Rinary1: > Wouldn't it make more sense to run this before getting a source

oh, you're right, pushed patch with fix, now it won't generate source and delete it if there's no sample.

Robust.Client/Audio/AudioManager.Public.cs
```diff
@@ -331,6 +328,12 @@ internal void RemoveBufferedAudioSource(int handle)
 
     public IAudioSource? CreateAudioSource(AudioStream stream)
     {
+        if (!_audioSampleBuffers.TryGetValue(stream.BufferId, out var sample))
+        {
+            OpenALSawmill.Warning($"Audio stream '{stream.Name}' has no backing buffer, skipping.");
```
metalgearsloth:
```suggestion
            OpenALSawmill.Error($"Audio stream '{stream.Name}' has no backing buffer, skipping.");
```

PR #7027, opened 2026-08-20
Move static colors to partial struct and add two colors

sleepyyapril:
Added:
Gluon Grey (known as #1B1B1E everywhere in content. RGB: 27, 27, 30, 255)
Noble Black (known as #202025 everywhere in content. RGB: 32, 32, 37, 255)

Moved static colors to a separate partial struct.


whatston3:
black brighter than grey
it's so over

sleepyyapril:
> black brighter than grey it's so over

im just going off some random site that said they have predefined names

sleepyyapril:
> black brighter than grey it's so over

would you rather Muted Deep Blue Canyon and Muted Deep Blue Opal

whatston3:
names are fine, IMO ideally there should be some sort of doc, because the rest of the system colours are MS Color values  [Even some sort of "Robust" reference would maybe be good but this is really nitpicky.]
References to a slight blue tint probably good, but whatever the names are I'll cope :+1:

[One question I do have is whether this is best suited for RT or content - it seems to all be content-side, and projects apart from SS14 might want different UI theming, not that an extra constant is a huge deal.]

Tayrtahn:
I suppose it doesn't really _hurt_ anything having more static colors in engine, but it might be more appropriate to define these as extension properties in content.

Edit: I just noticed the frozen dictionaries. That makes things a bit trickier. It might be best to do this here in the engine then.

sleepyyapril:
> I suppose it doesn't really _hurt_ anything having more static colors in engine, but it might be more appropriate to define these as extension properties in content.
> 
> Edit: I just noticed the frozen dictionaries. That makes things a bit trickier. It might be best to do this here in the engine then.

Yeah thats why I did it. Not sure if that's the best solution imho

PR #7031, opened 2026-08-22
Reset overlay draw state around callbacks

metalgearsloth:
Yeah I shoulda done this a while ago, it's an annoying footgun and the perf difference is immaterial.


DrSmugleaf (approved):
We have the technology

PR #7035, opened 2026-08-22
Fix generated generics

metalgearsloth:


DrSmugleaf (commented):
This shouldn't do anything since a data definition of a generic type is skipped, check the logs during serialization init
There should probably just be a better error that that doesn't work, unless we want to figure out generic data definitions actually being registered

PR #7036, opened 2026-08-22
Deep copy in compnetworkgenerator

metalgearsloth:
Not true deep copying but at least fixes immediate IRobustCloneable on collections.


Robust.Shared.CompNetworkGenerator/ComponentNetworkGenerator.cs
```diff
@@ -172,43 +172,197 @@ public class ComponentNetworkGenerator : ISourceGenerator
...
+                return type.WithNullableAnnotation(NullableAnnotation.None);
+            }
+
+            string GetCopyValueExpression(ITypeSymbol type, string value)
+            {
+                if (!NeedsCopyValue(type))
+                    return value;
+
+                if (type is INamedTypeSymbol named &&
+                    named.OriginalDefinition.ToDisplayString() == "System.Nullable<T>")
```
Tayrtahn, line 355: I think this check might be slightly cheaper/better like:
```cs
if (type is INamedTypeSymbol namedType && namedType.ConstructedFrom.SpecialType == SpecialType.System_Nullable_T)
```

DrSmugleaf (approved):
The world if serialization was fast enough to use here

PR #7039, opened 2026-08-24
Adds Custom to RSI Licenses

Redrover1760:
Allows forks to use Custom licensing for RSI

without having to edit engine files

I understand wanting to block this for standard wizden servers, however, testfailing on all servers is bad and and if its still desired it should be moved to content or something

### DIFF
--- a/Schemas/rsi.json
+++ b/Schemas/rsi.json
@@ -83,7 +83,8 @@
                 "CC-BY-NC-4.0",
                 "CC-BY-NC-SA-3.0",
                 "CC-BY-NC-SA-4.0",
-                "CC0-1.0"
+                "CC0-1.0",
+                "Custom"
             ],
             "examples": [
                 "CC-BY-SA-3.0"


ike709:
The RSI specification doesn't actually explicitly support `Custom` like the RGA spec does.

[Current RSI](https://github.com/space-wizards/docs/blob/master/src/en/specifications/robust-station-image.md) pertinent text:
> `license` | Required. A valid [SPDX License Identifier](https://spdx.org/licenses/) applying to this work.

[Current RGA](https://github.com/space-wizards/docs/blob/master/src/en/specifications/robust-generic-attribution.md#yaml) pertinent text:
> `license` | A valid [SPDX License Identifier](https://spdx.org/licenses/) applying to all files within an entry. If a license does not have a valid SPDX identifier, `Custom` may be used but a link to the license should be provided in the `copyright` field.

Should probably copy the RGA version of the text over to the RSI specification document?

PR #7040, opened 2026-08-24
RGA accepts a list of sources

iaada:
I was working on an audio file and in it I used two sounds from different sources. I want to attribute them both, but the current schema only accepts a single source. With my limited understanding of the schema validation (and help from Whatstone) I believe this will work to make my `source:` list validate without any breaking changes.

### DIFF
--- a/Schemas/rga.yml
+++ b/Schemas/rga.yml
@@ -6,7 +6,7 @@ attribution:
   files: list(str())
   license: license()
   copyright: str()
-  source: url()
+  source: list(url())
 
 # Example
 # - files: ["deprecated.png"]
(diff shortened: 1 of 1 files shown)


metalgearsloth:
Doesn't this muddy which file comes from which source?

deltanedas:
presumably youd only do that if every file listed uses all sources, similar case to listing multiple files for a directory permalink. i imagine youd only  list filea actually taken from there

iaada:
Yeah in the case of my linked PR (and the one in the example), both sounds listed in the `files` field were made using both sounds from `source` mixed together.

ike709:
This is a breaking change from the [specification](https://github.com/space-wizards/docs/blob/master/src/en/specifications/robust-generic-attribution.md) so at the very least that should also be updated in conjunction with this PR.

iaada:
Made a PR for the spec update https://github.com/space-wizards/docs/pull/673

PR #7041, opened 2026-08-24
Allow content to implement new BaseContainer types

DrSmugleaf:
Same as before with SerializedType but with less horrible performance now that ReflectionManager isn't (is less of) a sith holocron


Robust.Shared.IntegrationTests/GameObjects/ContainerTests.cs
```diff
@@ -365,5 +367,159 @@ await server.WaitAssertion(() =>
...
+             await server.WaitAssertion(() =>
+             {
+                 // Setup
+                 var mapId = MapId.Nullspace;
+                 var mapPos = MapCoordinates.Nullspace;
+
+                 sEntManager.System<SharedMapSystem>().CreateMap(out mapId);
+                 mapPos = new MapCoordinates(new Vector2(0, 0), mapId);
+
+                 sParentId = sEntManager.SpawnEntity(null, mapPos);
```
SnappingOpossum: You can call `Spawn` instead for the same result.

```suggestion
                 sParentId = sEntManager.Spawn(null, mapPos);
```

PR #7044, opened 2026-08-24
Fix YML linter with partials, game not loading with non-partial partials, and ambiguous sequence operations on partial prototypes

DrSmugleaf:
Fixes so good I had to re-code it from scratch 3 times because I lost my changes 2 times
- Fixes partial component addition/removals getting very confused and sometimes overwriting components
- Fixes it being a pain in the arse to modify long sequences but its still sort of unwieldy
- Fixes the game not loading if you partially lload a non-partial prototype
- Fixes the YML linter to work with partials, and added a partialNoOriginalError parameter to ValidateDirectory that when set to true will error when a non PartialOnly partial definition is found without an original to reference. This is set to false by default. This can be useful if you are not mixing partials and non-partials in the same directory
- Fixes not copying the original YML that is kept in PartialOriginals, resulting in duplicated data after applying partials
- Fixes YML hot reload with prototype partials
- Fixes inconsistent partial ordering if the same prototype is modified more than once in the same directory. Now it is ordered by the file path's name
- Added per-step verbose logging for prototype partials that logs every step of applying a partial prototype, if PrototypeManager's sawmill is set to show verbose logs.

The world if no dicts acting as lists
Also adds 2 more tests for the war

---

This
```yml
list:
- "g": 7
  "b": !Remove
- "h": 8
  "e": !Remove
```

Becomes
```yml
list:
- !CombineIndex:0 "g": 7
  "b": !Remove
- !CombineIndex:1 "h": 8
  "e": !Remove
```

# Breaking changes
- Using partial prototypes, use the tag !CombineIndex to combine a node in a sequence. This fixes a bug where modifying sequences (lists) in partial prototypes was buggy and inconsistent. If you are adding to a sequence or removing from it you don't need to change anything. If you were trying to modify existing sequence elements, you will need to replace - "a": 1 with - !CombineIndex:0 "a": 1 for example.

### DIFF
--- a/Robust.Shared/Prototypes/IPrototypeManager.cs
+++ b/Robust.Shared/Prototypes/IPrototypeManager.cs
@@ -429,7 +429,7 @@ public interface IPrototypeManager
     /// <param name="path">The directory containing the yaml files that need validating.</param>
     /// <param name="prototypes">The prototypes ids that were present in the directory.</param>
     /// <param name="partialNoOriginalError">
-    /// Whether a partial existing without an original to modify is an error ot not.
+    /// Whether a partial existing without an original to modify is an error or not.
     /// !PartialOnly prototypes will not error, even if this is set to true.
     /// </param>
     /// <returns>A dictionary containing sets of errors for each file that failed validation.</returns>
--- a/Robust.Shared/Prototypes/IPrototypeManager.cs
+++ b/Robust.Shared/Prototypes/IPrototypeManager.cs
@@ -417,7 +417,7 @@ public interface IPrototypeManager
     /// </summary>
     /// <param name="path">The directory containing the yaml files that need validating.</param>
     /// <param name="partialNoOriginalError">
-    /// Whether a partial existing without an original to modify is an error ot not.
+    /// Whether a partial existing without an original to modify is an error or not.
     /// !PartialOnly prototypes will not error, even if this is set to true.
     /// </param>
     /// <returns>A dictionary containing sets of errors for each file that failed validation.</returns>
--- a/Robust.Shared/Prototypes/IPrototypeManager.cs
+++ b/Robust.Shared/Prototypes/IPrototypeManager.cs
@@ -416,17 +416,27 @@ public interface IPrototypeManager
     /// Validate all prototypes defined in yaml files contained in the given directory.
     /// </summary>
     /// <param name="path">The directory containing the yaml files that need validating.</param>
+    /// <param name="partialNoOriginalError">
+    /// Whether a partial existing without an original to modify is an error ot not.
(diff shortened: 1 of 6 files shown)


Robust.Shared/Prototypes/IPrototypeManager.cs
```diff
@@ -416,17 +416,27 @@ bool TryGetInstances<T>([NotNullWhen(true)] out FrozenDictionary<string, T>? ins
     /// Validate all prototypes defined in yaml files contained in the given directory.
     /// </summary>
     /// <param name="path">The directory containing the yaml files that need validating.</param>
+    /// <param name="partialNoOriginalError">
+    /// Whether a partial existing without an original to modify is an error ot not.
```
metalgearsloth:
```suggestion
    /// Whether a partial existing without an original to modify is an error or not.
```

Robust.Shared/Prototypes/IPrototypeManager.cs
```diff
@@ -416,17 +416,27 @@ bool TryGetInstances<T>([NotNullWhen(true)] out FrozenDictionary<string, T>? ins
...
-    Dictionary<string, HashSet<ErrorNode>> ValidateDirectory(ResPath path);
+    Dictionary<string, HashSet<ErrorNode>> ValidateDirectory(ResPath path, bool partialNoOriginalError = false);
 
     /// <summary>
     /// Validate all prototypes defined in yaml files contained in the given directory.
     /// </summary>
     /// <param name="path">The directory containing the yaml files that need validating.</param>
     /// <param name="prototypes">The prototypes ids that were present in the directory.</param>
+    /// <param name="partialNoOriginalError">
+    /// Whether a partial existing without an original to modify is an error ot not.
```
metalgearsloth:
```suggestion
    /// Whether a partial existing without an original to modify is an error or not.
```

Robust.Shared/Prototypes/PrototypeManager.YamlValidate.cs
```diff
@@ -82,8 +85,16 @@ public Dictionary<string, HashSet<ErrorNode>> ValidateDirectory(ResPath path,
                 }
             }
 
+            var partial = IsFilePartial(resourcePath, out var index);
+            var partialResults = new List<ExtractedMappingData>(0);
```
metalgearsloth: Doesn't dotnet already pre-allocate these to 0 size, but this also adds on the capacity check?

DrSmugleaf: Fixed in https://github.com/space-wizards/RobustToolbox/pull/7044/changes/db586641f3b29adf108a5f81546db537583aab94

PR #7049, opened 2026-08-29
Expose MouseMoveEventArgs

Aidenkrz:
please speed i need this


Robust.Client/Graphics/Clyde/ClydeHeadless.cs
```diff
@@ -59,7 +59,7 @@ public ClydeHeadless()
 
         public event Action<TextEnteredEventArgs>? TextEntered { add { } remove { } }
         public event Action<TextEditingEventArgs>? TextEditing { add { } remove { } }
-        public event Action<MouseMoveEventArgs>? MouseMove { add { } remove { } }
+        public event Action<MouseMoveEventArgs> MouseMove { add { } remove { } }
```
metalgearsloth, line 62: Why was it made non-nullable? There's no guarantee when it's invoked that it has any subscribers.

Issue #7043, opened 2026-08-24
Combine deletion + predicted deletion


metalgearsloth:
We can reasonably infer which one to use (unlike spawning) so in this case just combine them.

PR #7059, opened 2026-09-01
Combine predicted deletion

maciejwalendziuk:
- Client `DeleteEntity` and `QueueDeleteEntity` now predict deletions of networked entities by detaching them to nullspace which is what `PredictedDeleteEntity` and `PredictedQueueDeleteEntity` did. They're obsolete aliases now along with the `PredictedDel` and `PredictedQueueDel`
- Removed the client's separate predicted deletion queue
- Fixed `TryQueueDeleteEntity` ignoring the predicted deletion queues and returning true for entities that were already going away
- Fixed predicted deletion state not being cleared on PVS reentry which left entities stuck detached
- Fixed a queued deletion still running after it had been cancelled. The queue only checked the set on insert but didn't on dequeue

Client-side `Del` and `QueueDel` on a networked entity used to log an error and do nothing - now it detaches. The obsoletion warnings cover the `PredictedDel` -> `Del` rename but not this so anything that was quietly relying on the noop will start visibly deleting for a tick

I hope that is what [metalgearsloth](https://github.com/metalgearsloth) meant. 
___
Solves: #7043

### DIFF
--- a/Robust.Client/GameStates/ClientGameStateManager.cs
+++ b/Robust.Client/GameStates/ClientGameStateManager.cs
@@ -1752,7 +1752,7 @@ namespace Robust.Client.GameStates
             if (!_entities.IsClientSide(uid) && _entities.TryGetComponent(uid, out TransformComponent? xform))
                 _recursiveRemoveState(meta.NetEntity, xform, _entities.GetEntityQuery<MetaDataComponent>(), _entities.GetEntityQuery<TransformComponent>());
 
-            // Set ApplyingState to true to avoid logging errors about predicting the deletion of networked entities.
+            // DeleteEntity predicts the deletion of networked entities.
             using (_timing.StartStateApplicationArea())
             {
                 _entities.DeleteEntity(uid);
--- a/Robust.Client/GameStates/ClientGameStateManager.cs
+++ b/Robust.Client/GameStates/ClientGameStateManager.cs
@@ -910,6 +910,7 @@ namespace Robust.Client.GameStates
                     DebugTools.Assert(!_toApply.ContainsKey(uid));
 
                     meta.Flags &= ~MetaDataFlags.Detached;
+                    _entities.ClearPredictedDeletion(uid);
                     if (isEnteringPvs)
                         enteringPvs++;
                 }
--- a/Robust.Shared.IntegrationTests/GameState/PredictedDeletionTests.cs
+++ b/Robust.Shared.IntegrationTests/GameState/PredictedDeletionTests.cs
@@ -81,16 +81,14 @@ internal sealed partial class PredictedDeletionTests : RobustIntegrationTest
 
         await client.WaitPost(() =>
         {
-            var ent = new Entity<MetaDataComponent?, TransformComponent?>(clientTarget, null, null);
-
-            client.EntMan.PredictedDeleteEntity(ent);
+            client.EntMan.DeleteEntity(clientTarget);
             var xform = client.EntMan.GetComponent<TransformComponent>(clientTarget);
             var firstDetachTick = xform.LastModifiedTick;
 
             Assert.That(((ClientEntityManager) client.EntMan).IsPredictedDetached(clientTarget), Is.True);
             Assert.That(client.EntMan.IsQueuedForDeletion(clientTarget), Is.True);
 
-            client.EntMan.PredictedDeleteEntity(ent);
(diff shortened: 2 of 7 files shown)


Robust.Shared/GameObjects/EntityManager.cs
```diff
@@ -302,10 +302,14 @@ private void UpdateTickHistogram(Histogram? histogram)
...
 
-        internal virtual void ProcessQueueudDeletions()
+        internal void ProcessQueueudDeletions()
         {
             while (QueuedDeletions.TryDequeue(out var uid))
             {
+                // The deletion may have been canceled since it was queued.
+                if (!QueuedDeletionsSet.Remove(uid))
+                    continue;
+
```
metalgearsloth, line 312: This would already be modifying the queue which in itself should not be touched so remove? At best make it a debug assert but these should always match 1-1.

Robust.Shared/GameObjects/EntitySystem.Proxy.cs
```diff
@@ -936,66 +936,66 @@ protected void QueueDel(EntityUid? uid)
 
     /// <inheritdoc cref="IEntityManager.DeleteEntity(EntityUid?)" />
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    [ProxyFor(typeof(EntityManager), nameof(EntityManager.PredictedDeleteEntity))]
+    [Obsolete("Use Del")]
     protected void PredictedDel(Entity<MetaDataComponent?, TransformComponent?> ent)
```
metalgearsloth, line 940: The ProxyFor is still useful to keep the interim I would think, just direct it at the new method?

Robust.Client/GameStates/ClientGameStateManager.cs
```diff
@@ -910,6 +910,7 @@ private void ApplyEntityStates(GameState curState, GameState? nextState)
                     DebugTools.Assert(!_toApply.ContainsKey(uid));
 
                     meta.Flags &= ~MetaDataFlags.Detached;
+                    _entities.ClearPredictedDeletion(uid);
                     if (isEnteringPvs)
```
metalgearsloth, line 914: If anything is touching entities outside of PVS it should really be an assert instead.

Robust.Shared.IntegrationTests/GameState/PredictedDeletionTests.cs
```diff
@@ -81,23 +81,72 @@ public async Task RepeatedPredictedDeletionDoesNotQueueOrDetachRepeatedly()
...
+                entMan.QueueDeleteEntity(clientTarget);
+            }
+            finally
+            {
+                entMan.EntityQueueDeleted -= OnQueueDeleted;
+            }
+
+            // Physics relies on this to purge contacts before the entity gets detached.
+            Assert.That(raised, Is.EqualTo(clientTarget));
+        });
```
metalgearsloth, line 124: Suss?

Robust.Shared/GameObjects/EntityManager.cs
maciejwalendziuk: they are intentionally not 1-1 anymore. ClearPredictedDeletion cancels a queued deletion by removing it from the set (since we cant remove from a queue) and this check discards the stale queue entry. otherwise a rolled back deletion redetaches the entity after state application restores it - assert would fire on every rollback

Robust.Shared/GameObjects/EntitySystem.Proxy.cs
maciejwalendziuk: Del already has ProxyFor(DeleteEntity) and the analyzer will just suggest the first matching proxy so pointing it at DeleteEntity could make DeleteEntity calls warn "use PredictedDel". PredictedDeleteEntity callers still get the obsolete warning pointing at DeleteEntity. I could be missing something though...

Robust.Client/GameStates/ClientGameStateManager.cs
maciejwalendziuk: its the rollback for predicted deletions. PredictedDetachNetworkedEntity sets Detached flag and when prediction reset doesnt run (for example prediction is disabled) reentry here is the only restore path. assert would fire on every one of those

metalgearsloth: That still doesn't explain why a nullspace entity would ever be predicted deleted?

PR #7062, opened 2026-09-03
Allow prototype upload validation to pass fields unknown to Server

### LINKED ISSUE #7061: Prototype validator drops valid prototypes if they have components with client-only fields
When prototype uploading, the server performs a validation check on the prototype. Part of that is checking the fields of the provided components. 
This becomes an issue when the component is split into different Client/Server components, such as LightBehaviourComponent. It has fields which only exists on Client, so when the server checks for the provided fields, it sees a field it doesn't recognize and fails the prototype. 

This is unfortunately not as simple as avoiding using components that are split (though I would advise against going down that route), as this issue is encountered even in inheritance. In the example below, I am simply inheriting the existing FlashlightSeclite entity, which itself inherits from FlashlightLantern which is where the LightBehaviourComponent comes from.

Image

Appears to have been introduced in https://github.com/space-wizards/RobustToolbox/pull/6548

SlamBamActionman:
Resolves #7061 

With the recently added ValidatePrototype function added to PrototypeManager, it now pre-validates the prototypes and fails the upload if it detects errors. Since this runs on the Server, the validation has no knowledge of components with Client-only fields (i.e. when the component is split up in Client/Server component versions). This causes the validation to get errors and reject valid prototypes. 

While ideally there shouldn't be a component split (or some better support for it), since that is a longterm issue this PR fixes it by checking that the ErrorNode has AlwaysRelevant set to true, which should filter out unrecognized fields. 

This does mean that incorrect/non-existent fields can pass, but afaik they get discarded anyways so it shouldn't mess up the game, just at worst make the prototype not behave as the uploader expects.

### DIFF
--- a/Robust.Shared/Prototypes/PrototypeManager.cs
+++ b/Robust.Shared/Prototypes/PrototypeManager.cs
@@ -516,6 +516,7 @@ namespace Robust.Shared.Prototypes
 
             var errors = _serializationManager.ValidateNode(kind, validationMapping, context)
                 .GetErrors()
+                .Where(x => x.AlwaysRelevant)
                 .ToArray();
 
             if (errors.Length == 0)


Tayrtahn (approved):

DIMMoon1:
@metalgearsloth @AJCM-git 
Please, this is necessary.

Svist666s:
silence

FriendAboba:
Still waiting for it

Svist666s:
2 weeks waiting it

Princess-Cheeseballs (approved):

Princess-Cheeseballs:
Please do not merge beg on Github, it doesn't make the engine releases go any faster.

Veritable-Calamity (approved):
My approval doesn't mean much, but have the 👍

metalgearsloth:
Having looked at it ideally we'd have some other way to differentiate client vs server vs actual mistakes but this is a decent hotfix (it's not exactly a bandaid but not 100% comprehensive).

PR #7064, opened 2026-09-03
Allow Color serialization from named colors in palette prototypes.

whatston3:
## About the PR

This PR adds Color serialization from palette colors.  These are specified in YAML in the form `<PALETTE_NAME>.<COLOR_NAME>`, e.g. `Paper.bounty`.

Palettes are specified in a palette prototype (refer to https://github.com/space-wizards/space-station-14/pull/45826 for examples)

Adds a PaletteManager to handle caching and looking up these values, with a few convenience methods to get a random color out of a particular palette.

If added, could provide an alternative for https://github.com/space-wizards/RobustToolbox/pull/7027 where it's used as a content-side palette.

## Why / Balance

One of the changes brainstormed from https://github.com/space-wizards/space-station-14/pull/45540.

Being able to reference the same Color from completely different YAML files based on the same palette, without needing anchors or parenting seems useful.

## Technical details

Adds a new PaletteManager (placed into Robust.Shared.ColorNaming, name not ideal but still fits).  This is registered as a new dependency.  PaletteManager holds both a named reference to specific colors (allowing for direct lookup from YAML), and a reference from a palette to an unnamed list of Colors inside.

Adds a PalettePrototype to store named references to Colors (implementation largely taken from SS14's ColorPalettePrototype) with a localized name.

PalettePrototypes serialize with a high priority (arbitrarily set to 1000 to allow lower/higher values).  They have a PostDeserialize handler that registers their Colors with the PaletteManager - this currently builds a dictionary as it reads the values on initial serialization (frozen by Initialize), then expects to receive updates on the prototype reload event.

Probably an error - if changes were add/update, but removals were handled by the reload event, probably makes more sense.

## Test plan

https://github.com/space-wizards/space-station-14/pull/45540 is a sample of this in use, and a decent sandbox for testing.

Should write a small integration test that adds a known set of colors in two serialization passes, testing the parsed Color values of a test entity prototype - adding a color, changing a color, removing a color.


Robust.Shared/ColorNaming/PalettePrototype.cs
```diff
@@ -0,0 +1,41 @@
...
+    [DataField(required: true)]
+    public Dictionary<string, Color> Colors { get; private set; } = null!;
+
+    /// <summary>
+    /// Post-deserialize callback, adds the named colors into the palette manager.
+    /// </summary>
+    void ISerializationHooks.AfterDeserialization()
+    {
+        var paletteMan = IoCManager.Resolve<IPaletteManagerInternal>();
+        paletteMan.Add(this);
```
Fildrance, line 39: You made it because ProtoMan is not filling up with proper stages? Its a bit odd but neat hack...

Robust.Shared/ColorNaming/PalettePrototype.cs
```diff
@@ -0,0 +1,41 @@
...
+using Robust.Shared.Serialization;
+using Robust.Shared.Serialization.Manager.Attributes;
+
+namespace Robust.Shared.ColorNaming;
+
+/// <summary>
+/// A prototype for storing arbitrary, named references to colors.
+/// </summary>
+[Prototype(loadPriority: 1000)]
+public sealed partial class PalettePrototype : IPrototype, ISerializationHooks
```
Fildrance: are we going to Obsolete ColorPalettePrototype of content with this? I really hope so! :>

Robust.Shared/ColorNaming/IPaletteManager.cs
```diff
@@ -0,0 +1,78 @@
...
+    bool TryGetPaletteColors(ProtoId<PalettePrototype> palette, List<Color> colors);
+
+    /// <summary>
+    /// Return a <see cref="List{Color}"/> of all colors within a given prototype.
+    /// Order of colors is not guaranteed.
+    /// </summary>
+    /// <exception cref="KeyNotFoundException">
+    /// Thrown if the kind of prototype is not registered.
+    /// </exception>
+    Color PickRandomColor(ProtoId<PalettePrototype> palette);
```
Fildrance, line 68: This gives not really obvious prediction problem - better put it where all randoms live, in SharedRandomExtensions so it can use manually created IRobustRandom

Robust.Shared/ColorNaming/IPaletteManager.cs
```diff
@@ -0,0 +1,78 @@
...
+    bool TryGetQualifiedColor(string name, [NotNullWhen(true)] out Color? color);
+
+    /// <summary>
+    /// Clears and fills <paramref name="colors"/> of all colors within a given palette.
+    /// Order of colors is not guaranteed.
+    /// </summary>
+    /// <exception cref="KeyNotFoundException">
+    /// Thrown if the kind of prototype is not registered.
+    /// </exception>
+    void GetPaletteColors(ProtoId<PalettePrototype> palette, List<Color> colors);
```
Fildrance, line 51: proviging List here feels odd. I know there are a lot of places where you would reuse List to not allocate new stuff but not giving an option to just get colors without extra body moves is just odd. Also using List implies that it can be changes, which would be very bad idea - i personally think it all should be IReadOnlyList stuff. And it probably does not make much sense to use this overload, because we should not change it anyways so no need for re-using same list. Its nice API things for optimizations, but this service does not need that behaviour.

Robust.Shared/ColorNaming/PaletteManager.cs
```diff
@@ -0,0 +1,153 @@
...
+        foreach (var (name, color) in proto.Colors)
+        {
+            _colorsByQualifiedName.Add(proto.ID + "." + name, color);
+        }
+    }
+
+    /// <summary>
+    /// Prototype reload handler, refreshes all references to palette colors.
+    /// </summary>
+    private void OnPrototypesReloaded(PrototypesReloadedEventArgs args)
```
Fildrance, line 130: Interesting side-note, which i am not sure how to make public knowledge.

When prototypes of colors will be reloaded, its usages won't be updated. I personally do not think its this PRs problem, but it is a bit puzzling for future scenarios...

Robust.Shared/ColorNaming/PaletteManager.cs
```diff
@@ -0,0 +1,153 @@
...
+    public PaletteManager()
+    {
+        _colorsByPalette = new();
+        _colorsByQualifiedName = new();
+    }
+
+    [PublicAPI]
+    public void Initialize()
+    {
+        _dictsFrozen = true;
```
Fildrance, line 38: aaaand where is call ToFrozen?

Robust.Shared/Serialization/TypeSerializers/Implementations/ColorSerializer.cs
```diff
@@ -34,7 +39,7 @@ public ValidationNode Validate(
         IDependencyCollection dependencies,
         ISerializationContext? context = null)
     {
-        return Color.TryFromName(node.Value, out _) || Color.TryFromHex(node.Value, out _)
+        return Color.TryFromName(node.Value, out _) || _paletteMan.TryGetQualifiedColor(node.Value, out _) || Color.TryFromHex(node.Value, out _)
```
Fildrance, line 42: maybe extract variable and do linebreaks for each condition? :3

Fildrance (changes requested):
Big thanks for this work, it looks nice and it can provide a lot of flexibility in future.
Major victory for crew!
Small nits and open quesitons... updating other prototypes that reference palette is clearly out of scope but i have no idea how to propelry communicate it to RT users... also please add changelog entry, they are in file and you have to do the thing manually.

Robust.Shared/ColorNaming/PaletteManager.cs
whatston3: It doesn't exist - it'd have to be done between deserializing all of the palettes, but before handling lower priority prototypes.  AFAIK there isn't a nice way to do that right now.

Robust.Shared/ColorNaming/PalettePrototype.cs
whatston3: Pretty much!  It would also work with per-prototype callbacks (likely a better solution, since then we can freeze the dictionaries and know that everything has been parsed and loaded)

Robust.Shared/ColorNaming/IPaletteManager.cs
whatston3: Got rid of them, RobustRandom.Pick + (Try)GetPaletteColors should be sufficient.

Robust.Shared/ColorNaming/IPaletteManager.cs
whatston3: Swapped over, works a lot cleaner.  Thanks!

Robust.Shared/ColorNaming/PaletteManager.cs
whatston3: Wrote a little snippet in remarks for the PalettePrototype.  It's not great, but it's something.

Robust.Shared/Serialization/TypeSerializers/Implementations/ColorSerializer.cs
whatston3: Made a change here, let me know if the update is sufficient.

whatston3:
Updated the comments in PaletteManager, I got _real_ lazy there.  Had a look at your points, sorry it took a while - https://github.com/space-wizards/RobustToolbox/pull/7064#discussion_r4054447322 I think _this_ might be interesting, it'd be a step in removing ISerializationHooks altogether, but I've done nothing there so far.

On the whole, this is still a lot hackier than I'd like, and I am sorry for that.

image

Still functional on the head of https://github.com/space-wizards/space-station-14/pull/45826 as of time of writing.

Robust.Shared/Serialization/TypeSerializers/Implementations/ColorSerializer.cs
```diff
@@ -39,9 +39,14 @@ public ValidationNode Validate(
...
-            ? new ValidatedValueNode(node)
-            : new ErrorNode(node, "Failed parsing Color.");
+        if (Color.TryFromName(node.Value, out _)
+            || _paletteMan.TryGetQualifiedColor(node.Value, out _)
+            || Color.TryFromHex(node.Value, out _))
+        {
+            return new ValidatedValueNode(node);
+        }
+
+        return new ErrorNode(node, "Failed parsing Color.");
```
Fildrance: maybe something like
```
        var isValid =  Color.TryFromName(node.Value, out _) 
                              || _paletteMan.TryGetQualifiedColor(node.Value, out _) 
                              || Color.TryFromHex(node.Value, out _);
        return isValid
            ? new ValidatedValueNode(node)
            : new ErrorNode(node, "Failed parsing Color.");
```

Robust.Shared/ColorNaming/PaletteManager.cs
```diff
@@ -60,53 +60,21 @@ public bool TryGetQualifiedColor(string name, [NotNullWhen(true)] out Color? col
...
-    [PublicAPI]
-    public Color PickRandomColor(ProtoId<PalettePrototype> palette)
-    {
-        return _random.Pick(_colorsByPalette[palette]);
+        return _colorsByPalette[palette];
     }
 
     [PublicAPI]
-    public bool TryPickRandomColor(ProtoId<PalettePrototype> palette, [NotNullWhen(true)] out Color? color)
+    public bool TryGetPaletteColors(ProtoId<PalettePrototype> palette, out IReadOnlyList<Color> colors)
```
Fildrance: nullable and NotNullWhenAttribute (on interface i suppose)

Fildrance (changes requested):
A little bit more, otherwise its pretty well cooked, now we wait for engine maintainer :3

Fildrance (approved):
Peak

PR #7067, opened 2026-09-06
Transform lerp rework

metalgearsloth:
Okay so this fixes SOME stuff on master but the last big piece is not handling packet loss because if we go from A -> B -> C and never get B it still doesn't lerp quite well, but that should be buildable off of this as we already buffer states by 1, I just didn't want to scopecreep this into oblivion. This is kind of more similar to source with a 100ms position correction but we don't have the extrap they have or the packet loss interp.

The ACTUAL fixes:
- Cross-parent lerps now work instead of being dropped. I needed this for z-levels.
- Transforms no longer get moved constantly in lerps saving expensive updates and we just use renderpose structs instead. Downside is we need to expand viewport queries as their trees never update.
- Local / networked mispredicts get smoothed out rather than instantly being corrected. If packets are dropped there will still be a noticeable lurch for now but we can fix that after.
- Made lerp values into CVars.
- Moving in a straight line now actually averages out to the actual velocity (4.5m/s for mob movement in ss14) instead of slightly changing, it's just hard to notice unless you dump the positions every frame and derive velocity.
- Tick timing gets frozen for every prediction which may cause some framepacing issues so this one is a bit susser, I mostly just wanted it stable for lerps.
- Added snap methods so we can just turn lerps off instead of working around it (containers mostly).

Realistically, in general the game should look marginally smoother, but it can't fix the actual game frame pacing (COUGH PVS POPIN PERF).


Robust.Shared.Tests/Timing/GameTiming_Test.cs
```diff
@@ -117,7 +117,146 @@ public void OutSimCurTimeTest()
...
+
+            foreach (var progress in new[] { 0.4f, 0.6f, 0.8f, 1f })
+            {
+                SetRealTickProgress(gameTiming, progress);
+                var phase = gameTiming.TickPhase;
+
+                Assert.That(phase, Is.GreaterThanOrEqualTo(previousPhase).Within(0.00001f));
+                previousPhase = phase;
+            }
+        }        /// <summary>
```
TheShuEd, line 259:
```suggestion
        }        
        
        /// <summary>
```

Robust.Client/GameStates/ClientGameStateManager.cs
```diff
@@ -231,6 +236,7 @@ public void Reset()
             _detachedChunkEntities.Clear();
         }
 
+
```
TheShuEd, line 245:
```suggestion
```

TheShuEd (commented):
This is maximum what i can review here lol

also, the TickTimingAdjustment getter returns the requested value, but the tick timing is calculated based on the frozen value. Is it just me, or would any third-party consumer reading TickTimingAdjustment in the middle of a tick see a value that doesn't correspond to the current tick period?

PR #7068, opened 2026-09-06
Add a test for clearing and adding a parent with a partial prototype

DrSmugleaf:


deltanedas:
by the way you cant change a parent if it uses the single value syntax instead of an array

PR #7072, opened 2026-09-07
Revert "try microoptimise physics island solving (#6871)"

Princess-Cheeseballs:
This reverts commit 3ee56ea02722dd97bb7cd00d1eacca8da82aa8e8.

Has an issue where it can cause an exception loop on either client or server which causes physics to stop completely. Don't have time to debug this so it's just gonna get reverted.


deltanedas:
move the clear to above where it adds to the list instead of below
alternatively fix whatever is adding deleted entities to islands joints

Issue #5648, opened 2025-02-01
BUIs with different names can collide.


whatston3:
SharedUserInterfaceSystem uses ReflectionManager.LooseGetType to match BUIs.  If you had multiple related BoundUserInterfaces with different prefixes - a lookup call for one without a prefix will match to the first one it can find, even between assemblies.  Ideally the type name should be complete, so a call for "ABC" will match "ABC" and not "XYZABC", regardless of what assembly it's in.

This was the cause of this PR:
https://github.com/new-frontiers-14/frontier-station-14/pull/2831 - when the namespace of an assembly with three BUIs (StationBankATM, WithdrawOnlyBankATM, and BankATM) changed, the interface that opened when requesting BankATM was actually StationBankATM.

whatston3:
Will be fixed by #7073.

PR #7073, opened 2026-09-07
fix: ReflectionManager.TryLooseGetType now doesnt matches part for full name

CREAsTIVE:
Before `ReflectionManager.TryLooseGetType` matched part of name, i. e. type `ShuttleConsoleBoundUserInterface` may be matched as `TradeShuttleConsoleBoundUserInterface` effectively replacing existing one without any warnings or something.

~~Initially i matched with `.Split('.').TakeLast(...).SequenceEqual(...)`, but that allocates too much so i just went with checking for existence of `.` or `+` (for nested types, i. e. `A.B.C+D`)~~ I just use already written method for enums

Related issues: https://github.com/space-wizards/RobustToolbox/issues/5648


CREAsTIVE:
Requires merge of https://github.com/space-wizards/space-station-14/pull/45913 first (to pass tests). All tests pass on my machine and game works perfectly fine with that fix

PR #7074, opened 2026-09-07
Add a parameter to FileDialogManager.SaveFile to append an extension.

whatston3:
Title.

Adds an optional parameter called `appendExtension` to `FileDialogManager.SaveFile`, defaulting to false (preserving existing behaviour - shouldn't break anything as-is).  When set to true, if a file is selected for saving, the extension is added to the file if the entered file name:

1. Doesn't exist,
2. Doesn't have an extension already,
3. Wasn't selected with a wildcard filter.

`IFileDialogManagerImplementation.OpenFile`, `SaveFile` edited to receive the filter index to get the expected file extension to append to the file.

## To test:

1. Checkout [2026-09-07-append-extension](https://github.com/space-wizards/space-station-14/compare/master...whatston3:space-station-14:2026-09-07-append-extension?expand=1) off of my [ss14 repo](https://github.com/whatston3/space-station-14/tree/2026-09-07-append-extension)
2. Build and run (might need to add me as an RT remote to get the submodule, but it's the head of this PR.
3. Run `golobby`, enter the character menu.
5. Export a character to "bababooey", you should write to "bababooey.yml".
6. Export a character to "plonk.yml", you should write to "plonk.yml".
7. Export a character to "funny.txt", you should write to "funny.txt".
8. Rename "funny.txt" to "funny" and save to "funny" - you should rewrite over "funny", and _not_ create a file called "funny.yml".


Robust.Client/UserInterface/FileDialogManager.cs
```diff
@@ -69,15 +69,31 @@ private static FileShare Validate(FileAccess access, FileShare? share)
...
+                && !Path.HasExtension(name)
+                && !File.Exists(name))
+            {
+                var firstExtension = filters.Groups[filter].Extensions[0];
+                if (firstExtension != "*")
+                {
+                    name += "." + firstExtension;
+                }
+            }
+
```
metalgearsloth, line 96: Am I high or is this possible to sandbox break because filedialogfilters are entirely user input and we're manually constructing the path here?

whatston3: The extension groups shouldn't be user input - in the existing test cases, that should be "yml" and "csv".
I mean, they _could_ be, but why would you do that?

[If it is a problem, and if it can be fixed, I'm all ears.]

RELEASE-NOTES.md
```diff
@@ -42,6 +42,7 @@ END TEMPLATE-->
 ### New features
 
 * `Control.Restyle` has been made public for manually restyling controls on edge cases.
+* `FileDialogManager.SaveFile` now has an optional `appendExtension` parameter.  When true, it adds the selected extension (if not a wildcard) if saving to a new file.
```
metalgearsloth: The API is changing so this is explicitly a breaking change.

Robust.Client/UserInterface/FileDialogManager.cs
metalgearsloth: I don't think we have any existing restrictions around file formats, e.g. if some fork wants to use their own format, or wants their own tigher whitelist, which I don't know if we want to support or not. it may become a problem when video resources get implemented so maybe better to just allow any extensions to be used for filters.

Mackamuir:
Epic

PR #7076, opened 2026-09-08
make island solver more robust

deltanedas:
- clear the joints list before adding to it not after using it, so if it throws there aren't deleted entities in it permanently breaking physics because of the GetComponent calls. this doesn't stop whatever is failing to clean up deleted entities from island joints.
- use TryComp so it doesnt throw anyway

closes #7072


Princess-Cheeseballs (approved):
Yummers

Princess-Cheeseballs:
Remind me to backport this to the current engine version tomorrow if I forget to do it today.

deltanedas:
no

Moredread:
The client case from space-wizards/space-station-14#45725 is still reproducible on v289.0.3 (content 16fc277605, engine 36905986f), and it got worse: the throw moved from BuildIslands into the solver and now repeats every predicted tick until the client is restarted, with client CPU and memory climbing over time (ten minutes on my machine: process CPU 47% -> 85%, RSS +90 MB/min, fps 165 -> 130 on the dev map. I think earlier today I saw 11 fps, but could not reproduce it).

Repro: pull something (a chair, a machine or another mob), disconnect and reconnect while still pulling. Likely not on every relog, I estimate ~1/3-1/2, but did not nail this down, so you might need to relog multiple times to trigger it. Once it starts, stopping the pull did not seem to clear it; a relog does.

Client log, one exception per frame (~170/s at 165 fps, ~3400 lines/s, 300 MB in ten minutes):

    System.AggregateException ---> System.Collections.Generic.KeyNotFoundException: The given key '1' was not present in the dictionary.
       at Robust.Shared.Physics.Dynamics.Joints.DistanceJoint.InitVelocityConstraints(...) in DistanceJoint.cs:line 298
       at Robust.Shared.Physics.Systems.SharedPhysicsSystem.SolveIsland(...) in SharedPhysicsSystem.Island.cs:line 826
       at ...<SolveIslands>b__1 in SharedPhysicsSystem.Island.cs:line 697   (Parallel.For)
       at ...Solve in SharedPhysicsSystem.Island.cs:line 318
       at ...SimulateWorld in SharedPhysicsSystem.cs:line 307

What happens, as far as I can tell: after the reconnect the client holds two Joint objects for the pull joint, one with an unresolved end (entity 0). This PR skips that copy in BuildIslands, so the other copy now reaches the solver and fails the IslandIndex lookup. Solve() has no cleanup on the exception path, so the island list and _awakeBodyList that #6871 turned into fields keep their state, the same island is re-solved and re-thrown every tick, and the awake list grows every tick.

Same client setup on v288.1.0 (before #6871): one or two throws, then no additional log lines. The prediction for the pulled object dies, but in contrast to now, letting go seems to be enough to resolve it, without a relog. On 289.0.0-0.2 it was the permanent BuildIslands throw this PR fixed, visible client side e.g. as space-wizards/space-station-14#45725, no physics prediction on the client until the causing joint is cleared and the player relogs.

This PR fixed one of the two lists #6871 moved to fields; the other one still needs the same treatment. A try/finally around Solve() that resets the per-tick state bounds the damage: nothing accumulates and physics recovers the moment the bad joint copy is gone. It is not enough on its own, though: with only that, the valid copy still throws in the solver every tick while the pull lasts, so the player stays frozen until letting go; that needs a guard at the solver's IslandIndex lookup as well, and the pull itself only works once the client repairs the unresolved copy.

None of this stops malformed joints from occurring, it limits what they can do; the guard is what keeps the player moving when one shows up.

Until that lands I would reopen #7072 and revert #6871 together with this PR. Before #6871 the same state cost one or two throws and a dead pull, and the player kept moving; that was the behaviour for years, and it is better than any 289.x state for the client. I also can not exclude that the surface #6871 opened is wider than what I looked at, I mainly focused on the client side regression.
 
Disclosure: the code analysis, integration tests comparing the behaviour across engine versions and the draft for this post were done with help of an AI assistant (Claude). I post this as a comment and not a PR because of this.

deltanedas:
W claude

joints having invalid entities isnt the fault of this anyways

PR #7077, opened 2026-09-09
Add lock to RegisterVariantCollection

Mackamuir:
Adds a lock to RegisterVariantCollection to prevent multiple threads writing to the same dictionary, corrupting it.

Recently had a [testfail](https://github.com/ss14Starlight/space-station-14/actions/runs/34268389448/job/102203478537#step:9:1034) on Starlight: `ArgumentNullException`, pretty sure this is the cause but I couldn't repro locally since race condition.


metalgearsloth:
This looks pretty suss as any number of methods above it will explode without thread-safety?

Mackamuir:
> This looks pretty suss as any number of methods above it will explode without thread-safety?

Doubt it, each worker only edits node from the file it parsed, kindData is shared and unfrozenVariants is the only shared thing extractMapping writes to.

Honestly the chances of this happening are pretty low anyways, I was more curious and wanted to look into it. will just close and if it comes up again later for whatever reason someone can see here.

PR #7078, opened 2026-09-09
Return Ray.Intersects() exit as out var

JiyuSoftware:
I thought I needed to do this to get https://github.com/space-wizards/space-station-14/pull/45954 working, but I ended up not needing it. I still think this is a good change I think so I committed it anyways.

## What I did:

I added a second `Intersects()` function to Ray.cs with a fourth param which is the ray exit out var that could be used for future more advanced collision checks or whatever.

This value is literally free and costs no extra resources, it is already calculated as `tmax` inside of the function and is then discarded, so why not return it so it can be used elsewhere.

## Also added:

* Previously missing docblocks.
* Added a tiny performance optimisation in `SharedPhysicsSystem` by using the new previously discarded exit value instead of casting another ray to calculate it as a demonstration to show you why this feature isn't actually completely useless!


metalgearsloth:
RaycastSystem obsoleted this.

JiyuSoftware:
> RaycastSystem obsoleted this.

Yes, but this is still in the code and it is a tiny performance optimization + an extra public function that might be useful downstream.

You can take it or leave it.

As I said before, I made this because I thought I was going to need this but I didn't end up needing it.

PR #7080, opened 2026-09-11
Make OutputPanel correctly orphan controls when modifying messages

mycobiota:
OutputPanel provides a SetMessage method that can be used to modify already existing chat messages, but when doing so it does not call the RemoveControls method on the RichTextEntry that is being modified, and as a result any controls in the message like icons or command links will linger, and end up hovering in weird positions.
RichTextEntry explicitly states that if it's ever removed or modified without removing the owning control then RemoveControls should be called.
[RMC already encountered this issue and was able to work around it by orphaning the problematic control directly.](https://github.com/RMC-14/RMC-14/pull/10203)


mycobiota:
here's another example of the issue from funky station's rebase onto wizden, once again in relation to chat stacking, where ghost follow buttons end up lingering
image

PR #7081, opened 2026-09-12
DebugConsole: Don't invoke OnTextChanged on scrolling through command history.

### LINKED ISSUE #45991: Command Recall (Up Arrow) Gets Stuck On Fill Lookup

_No response_

### Description

In summary, when an admin cycles through their command list, previously the up arrow used to skip past whatever additional information the command was able to accept.

Currently, the command always gets stuck on any enterable value.

### Reproduction

Run some commands, include something like TP and addgamerule, griddrag, and any other common command you may use as an admin which can accept, but won't always need further data.

Press up and try to cycle back to the first entered command you wish to reuse (ghost exec scripts are super common here). Find that the up button instead is cycling through autofill for the commands after the target that are looking for further input.

### Additional Context

This started happening about a week or so ago, roughly when Lizard was transitioned to the testing server.

whatston3:
Sets InvokeOnHistorySelect to [false] in the DebugConsole.

Fixes https://github.com/space-wizards/space-station-14/issues/45991, restores (desirable!) prior behaviour when selecting between messages in the debug history.

Change of behaviour introduced in #7011.

Also removes a no-op OnTextChanged handler and its related, commented out variable.

Changes as viewed in SS14:

https://github.com/user-attachments/assets/979c4cf0-f9f3-4616-b459-5dc863abe86e


metalgearsloth (approved):

metalgearsloth:
This was also annoying me and I shoulda spotted it earlier.

PR #7086, opened 2026-09-15
Self Merge, prevent ultra giga PVS death when reconnecting while having a joint.

Princess-Cheeseballs:
Title.
Clear the islands first in case that shit throws an exception so we don't keep throwing the exception one gorillion times.


Princess-Cheeseballs:
@deltanedas you have about 20 minutes to judge this PR

deltanedas:
yeah its good
i dont think Cleanup should be clearing any of these tbh, it should always be done ahead of time

PR #7087, opened 2026-09-15
Add z-levels

metalgearsloth:
https://www.youtube.com/watch?v=VKhpE-oNoGY

Okay so here's the deal. This is ported vaguely from https://github.com/crystallpunk-14/crystalledge/

The main problem was lerping didn't work properly for 2 reasons:
1. Cross-parent lerps don't work on master.
2. After fixing the above you still need to lerp between z-levels due to the projection offset.

For xform lerping we have https://github.com/space-wizards/RobustToolbox/pull/7067 which this is built off of, all cool looks good lerping looks great, but combining this with z-levels is tortium. The main issue being z-level position is its own data that needs lerping too and also needs to integrate with this. We also need to handle cases like "teleporting across maps doesn't lerp but teleporting across z-levels does".

Oh yeah also lerping the render alpha + lighting is annoying.

Anyway if I were to split this PR the rendering could potentially be split out alongside the grid-sync, but the physics is half of it which is a bit harder to do. Also fixed chunk edges but I'll split that out.

The CE zlevel values + the render shader is on the content PR as I wasn't entirely sure how biased the engine side should be so opted to leave it out for now, it's just the physics values are biased.

All of the breaking changes are contained on the transform side around lerping but there will be a slight perf overhead, but less-so than what occlusion is doing atm.

Github doesn't want to upload the video sadly.

NOT included is ramps / ladders, it's only rendering + grid-sync + falling physics + lerp fixes as the scope is already unreviewable.


JohnJJohn:
Dear god

deltanedas:
how much of this was claude

Strol20:
Z level before surgery?

Svetoslav228:
Finally something actually happening.

Robust.Shared/Physics/Systems/ZLevelPhysicsSystem.Clearance.cs
```diff
@@ -0,0 +1,302 @@
...
+                LayerBits = bodyFixture.CollisionLayer,
+                MaskBits = bodyFixture.CollisionMask,
+                Flags = QueryFlags.Static,
+            },
+            Approximate: false,
+            IgnoreShapeSkin: true);
+
+    private bool TryGetVolumeTopHeight(
+        EntityUid provider,
+        ZLevelHighGroundComponent highGround,
```
TheShuEd, line 201: unused

TheShuEd (commented):
single thing that i found

notquitehadouken:
media never

notquitehadouken:
oh right this is engine we don't have media here

MayasMischief:
media now (even though its engine so it'd be useless) it'd be funny

PR #7089, opened 2026-09-16
Rate limit full game state requests

acts-1631:
A connected client can repeatedly send `MsgStateRequestFull`. `ServerGameStateManager.HandleFullStateRequest` forwards every request to `PvsSystem.OnClientRequestFull`, which clears the session’s PVS history and forces a complete reliable state rebuild. There was no per-session throttling, allowing repeated requests to consume server CPU, memory, and reliable bandwidth.

This adds a one-second per-session cooldown and coalesces requests while a full state is already pending. Normal recovery requests still work, but a client can no longer continuously force expensive full-state rebuilds.


deltanedas:
rebase your branch

metalgearsloth (changes requested):
RealTime will be better than CurTime hereas it's not affected by paused. And as delta said.

acts-1631:
Updated

PR #7090, opened 2026-09-16
Adds AfterLighting overlay space

jessicamaybe:
Adds support for drawing to lighting after all of the other lighting shaders have been drawn.

Need this for some lighting stuff I wanna do as some of the lighting effects were getting in the way


metalgearsloth (commented):
Is this for tile lighting or something else? Ideally either:
A) you draw before the lightblur is applied so it's consistent
B) you just draw to the light target afterwards in worldspacebelowworld which runs at the same time.

Or am I missing something...

jessicamaybe:
> Is this for tile lighting or something else? Ideally either: A) you draw before the lightblur is applied so it's consistent B) you just draw to the light target afterwards in worldspacebelowworld which runs at the same time.
> 
> Or am I missing something...

Well I didn't even notice that `WorldSpaceBelowWorld` is basically in the same place. That solves my problem, I'll close this. Thank you!

(also yes this is for my weird tile lighting, i basically wanna just roll my own lighting instead of engines)

PR #7091, opened 2026-09-17
Add missing proxies for predicted spawning methods

SnappingOpossum:
Not really sure why these were even missing before.
Added the four proxy methods with direct inheritdocs to what they're proxies for. Release note part could probably be shorter, and I'm not too sure I should have this in the `New features` section or `Other`.


Princess-Cheeseballs (approved):
Truth supernova.

PR #7093, opened 2026-09-17
Fix predicteddel rollback

metalgearsloth:
Was getting this on the lerp branch with 2-handed items like plants.


metalgearsloth:
Merge after https://github.com/space-wizards/RobustToolbox/pull/7059 (who am I kidding it's probably gonna be a sefly)

PR #7095, opened 2026-09-18
Fix more stale proxyytree clears

metalgearsloth:
Thought I got this but emisse just got another. Also added full broadphase clears which are faster than individually removing each node entry. Also avoids MoveProxy for micro-changes because Box2D expands the AABBs slightly.


Princess-Cheeseballs (approved):
Code makes sense and it handled by horrific testing choices well so should be fine. Feel free to merge whenever.

